PR #15, opened 2023-09-25
Update Readme

SmittieC:


bderenzi (approved):
I'll take your word for it on the `virtualenv` update. I use `pyenv` as an alternative.

SmittieC:
> I'll take your word for it on the `virtualenv` update. I use `pyenv` as an alternative.

yeah in this case `virtualenv` is just another way of doing it. I found the `python -m venv` command to be quicker. I guess eveyone has their way of doing it

Issue #6, opened 2023-09-25
Add channel to data export and chat summaries


SmittieC:
- [x] Currently channel is not recorded (web / telegram / WhatsApp) on chat summaries 
- [x] WhatsApp and Telegram chat transcripts are marked as "anonymous". We should be able to get the phone number for WhatsApp and at least the Telegram ID
- [ ] Voice notes are not 'noted' in any way. I think it would be good to mark what was a voice note transcription

SmittieC:
Only the last point. Currently we don't know which message was transcribed when looking at the chat history. Maybe we should tag those messages and use that as indicators in the export. It's been so long since this was requested....

snopoke:
Remainder tracked in #1208

PR #19, opened 2023-09-27
Show channel on Experiment Review page

Charl1996:
This PR addresses the first point on [this issue](https://github.com/dimagi/open-chat-studio/issues/6), namely to display the channel name on the _Experiment Review_ page. See screenshow below.

image

In addition the platform name is also included when exporting the chat data.

### DIFF
--- a/apps/channels/models.py
+++ b/apps/channels/models.py
@@ -10,10 +10,21 @@ from telebot import TeleBot, apihelper
 from apps.experiments.models import Experiment, ExperimentSession
 from apps.utils.models import BaseModel
 from apps.web.meta import absolute_url
+from apps.channels.const import (
+    WEB,
+    TELEGRAM,
+    WHATSAPP,
+    PLATFORM_DISPLAY_NAME,
+)
 
 
 class ExperimentChannel(BaseModel):
-    PLATFORM = (("telegram", "telegram"), ("web", "web"), ("whatsapp", "whatsapp"))
+    PLATFORM = (
+        (TELEGRAM, TELEGRAM),
+        (WEB, WEB),
+        (WHATSAPP, WHATSAPP)
+    )
+
     name = models.CharField(max_length=40, help_text="The name of this channel")
     experiment = models.ForeignKey(Experiment, on_delete=models.CASCADE, null=True, blank=True)
     active = models.BooleanField(default=True)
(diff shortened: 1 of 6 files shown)


templates/experiments/components/experiment_details.html
```diff
@@ -25,6 +25,12 @@ <h3 class="text-base font-semibold leading-7 text-gray-900">Experiment details</
         <dt class="text-sm font-medium leading-6 text-gray-900">Experiment</dt>
         <dd class="mt-1 text-sm leading-6 text-gray-700 sm:col-span-2 sm:mt-0">{{ experiment_session.experiment.name }}</dd>
       </div>
+      <div class="px-4 py-6 sm:grid sm:grid-cols-3 sm:gap-4 sm:px-0">
+        <dt class="text-sm font-medium leading-6 text-gray-900">Platform</dt>
```
czue, line 29:
```suggestion
        <dt class="text-sm font-medium leading-6 text-gray-900 dark:text-gray-100">Platform</dt>
```

templates/experiments/components/experiment_details.html
```diff
@@ -25,6 +25,12 @@ <h3 class="text-base font-semibold leading-7 text-gray-900">Experiment details</
         <dt class="text-sm font-medium leading-6 text-gray-900">Experiment</dt>
```
czue, line 25:
```suggestion
        <dt class="text-sm font-medium leading-6 text-gray-900 dark:text-gray-100">Experiment</dt>
```

apps/channels/models.py
```diff
@@ -33,6 +35,10 @@ def save(self, *args, **kwargs):
             token = self.extra_data.get("bot_token", "")
             logging.error(f"Unable set Telegram webhook with token '{token}'")
 
+    @property
+    def channel_display_name(self):
+        return PLATFORM_DISPLAY_NAME[self.platform]
```
czue, line 40: Can you not just use the `choices` and the built in `get_platform_display` for this? https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model.get_FOO_display

czue (commented):
some dark mode suggestions (haven't tested) and one clarifying question / suggestion

apps/channels/models.py
Charl1996: @czue Yeah, that was the direction what I initially took, but decided to rather decouple the display text from the internal slugs with this `PLATFORM_DISPLAY_NAME` dict. 

Granted, the `choices` display text is currently the same as their slugs, so maybe the question needs to be asked whether or not we want the display text (take for example whatsapp) to be "whatsapp" or "WhatsApp"?

I know it's a very small thing, but in my opinion rather get it right now than having to fix it later on. Thoughts?

czue: the second value is intended to be the display name. so i'd think we can update the choices dict to use the display names we want and then follow convention rather than re-implement the same thing in a different way.

Charl1996:
Closing this to clean up the commits

apps/channels/models.py
Charl1996: @czue Good point.

PR #20, opened 2023-09-28
Cs/show channel name on summary

Charl1996:
This PR addresses the first point on [this issue](https://github.com/dimagi/open-chat-studio/issues/6), namely to display the channel name on the _Experiment Review_ page. See screenshow below.

image

In addition the platform name is also included when exporting the chat data.


Charl1996:
@czue I didn't address the [dark mode suggestions](https://github.com/dimagi/open-chat-studio/pull/19#pullrequestreview-1646927338) in this PR; I think it's probably better to address that separately?

apps/channels/models.py
```diff
@@ -11,9 +11,14 @@
 from apps.utils.models import BaseModel
 from apps.web.meta import absolute_url
 
+WEB = "web"
```
SmittieC, line 14: These names are the same as
```python
PLATFORM = (("telegram", "telegram"), ("web", "web"), ("whatsapp", "whatsapp"))
```

Any particular reason why you're creating these constants then? It would make sense to do this if it were used somewhere else in the code maybe, but personally I don't see reason to do it?

apps/experiments/models.py
```diff
@@ -322,6 +322,9 @@ def get_invite_url(self) -> str:
             )
         )
 
+    def get_platform_name(self) -> str:
+        return self.channel_session.experiment_channel.platform_display
```
SmittieC, line 326: You should be able to do
```python
return self.channel_session.experiment_channel.get_platform_display()
```
then you can remove this:
```python
    @property
    def platform_display(self):
        return ExperimentChannel.get_platform_display(self)
```

Charl1996: Ah, you're right. Missed that option.

apps/channels/models.py
Charl1996: I get anxious seeing strings lying around in code, so I like to tie it to constants.

SmittieC: I don't have any beef with this

SmittieC (approved):
Looks fine to me

apps/channels/models.py
```diff
@@ -11,9 +11,14 @@
...
+TELEGRAM = "telegram"
+WHATSAPP = "whatsapp"
+
 
 class ExperimentChannel(BaseModel):
-    PLATFORM = (("telegram", "telegram"), ("web", "web"), ("whatsapp", "whatsapp"))
+    PLATFORM = ((TELEGRAM, TELEGRAM), (WEB, WEB), (WHATSAPP, WHATSAPP))
```
czue, line 20:
```suggestion
    PLATFORM = ((TELEGRAM, "Telegram"), (WEB, "Web"), (WHATSAPP, "WhatsApp"))
```

is there a reason you chose not to add the capitalized display names?

czue:
> @czue I didn't address the https://github.com/dimagi/open-chat-studio/pull/19#pullrequestreview-1646927338 in this PR; I think it's probably better to address that separately?

When I look at that screenshot I don't see something that I would be comfortable shipping, but I suppose it's up to you. Is it not just a matter of applying classes similar to the ones I suggested in the other PR?

bderenzi:
What is the argument for not adding the CSS to support dark mode in this PR? I just happened to miss this particular page when I went through and did fixed up dark mode on the rest of the site. I'm +1 on adding the fix here before merging.

Charl1996:
Turns out I had to remove the `text-gray-` class from the elements in order to make the dark mode appear correctly...is this expected?

It's fine now in dark and light mode.

Dark mode
image

Light mode
image

bderenzi:
> Turns out I had to remove the `text-gray-` class from the elements in order to make the dark mode appear correctly...is this expected?

There are two ways to solve it: either add a light colour that will be used in dark mode (e.g., Cory's original suggestion `dark:text-gray-100`) or to remove the colour assigned so that it goes back to the default colour and correctly toggles with dark mode.

I'm fine with either solution. I think in other places we've used the approach Cory mentioned so that colours are explicit.

Charl1996:
Any preferences, @czue ? I'd like to keep it uniform with the rest of the code, but not sure how strict you are with this particular way of doing? (doing something like `dark:<style>` doesn't work for me for some reason, but before I investigate further I just want your take if it's worth doing so)

czue:
@Charl1996 No strong feelings. I was under the impression that the color classe theses were added for emphasis (e.g. they are a bit darker/lighter than the default), but that might not be true? To be honest, they were likely lifted directly from tailwind UI so I would have to check how they compare to the default values.

In general, not using any color classes seems simpler if it works and looks good, so maybe let's go with that?

czue (approved):

PR #21, opened 2023-09-28
Small refactor to load synthetic voices using a migration

SmittieC:
## Removed
The management command to create synthetic voice records

## Added
- A migration to create synthetic voice records
- A folder named `preload_data` in the migrations directory in the `experiments` app

I have mixed feelings about the data file (aws_voices.json) and where it's located, not 100% why. For now I'm leaving it there to see if anyone shares my feelings.


snopoke (approved):
The file location seems fine to me.

PR #31, opened 2023-10-06
Real time token counting

bderenzi:
Limitations: This currently assumes GPT-3.5-turbo or GPT-4 for BPE using cl100k_base.

See also [this note](https://github.com/latitudegames/GPT-3-Encoder/issues/40#issuecomment-1565789175) about why it won't match [OpenAI's online tokenizer](https://platform.openai.com/tokenizer), which hasn't been updated to the new tokenization scheme from.

Also, this stuff is rather imprecise. The number of input tokens seems to be system prompt token count + user token count + 15?

Last thing: I didn't choose the JS library carefully and seems like there are a lot. Seems likely there is one that is smaller/faster/better.

### DIFF
--- a/assets/javascript/site.js
+++ b/assets/javascript/site.js
@@ -2,4 +2,5 @@
 // HTMX setup: https://htmx.org/docs/#installing
 import 'htmx.org';
 import './htmx';
-import './alpine'
+import './alpine';
+import './tiktoken';
--- a/assets/javascript/tiktoken.js
+++ b/assets/javascript/tiktoken.js
@@ -5,8 +5,8 @@ const { encodingForModel } = require('js-tiktoken');
 const encoding = encodingForModel("gpt-3.5-turbo");
 
 // Function to count tokens
-function countTokens(string) {
-    const tokens = encoding.encode(string);
+function countTokens(stringToCount) {
+    const tokens = encoding.encode(stringToCount);
     return tokens.length;
 }
 
--- a/assets/javascript/tiktoken.js
+++ b/assets/javascript/tiktoken.js
@@ -1,0 +1,14 @@
+// Import the necessary parts of the js-tiktoken library
+const { encodingForModel } = require('js-tiktoken');
+
+// Get the encoding details for gpt-3.5-turbo and gpt-4
+const encoding = encodingForModel("gpt-3.5-turbo");
+
+// Function to count tokens
+function countTokens(string) {
+    const tokens = encoding.encode(string);
+    return tokens.length;
+}
+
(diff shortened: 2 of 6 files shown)


SmittieC:
> Also, this stuff is rather imprecise. The number of input tokens seems to be system prompt token count + user token count + 15?

How did you determine the system prompt token count? Also, what does the user token refer to?

We ultimately want to know how many tokens the final constructed prompt is going to be that the LLM is seeing right? To my knowlegde, this includes

1. the "framework"'s token count that LangChain builds for the variable data (source material, prompt, chat history)
2. The source material, prompt and chat history's token count

We should be able to get the token count for the framework that LangChain constructs, so if we can create a call to get that, then we should be able to simply count the source material, chat history and user prompt separately on the FE (I'm 90% sure about this).

bderenzi:
> Maybe the extra 15 tokens were due to some internal LangChain thing?

I think the extra 15 tokens are either LangChain or OpenAI itself. My understanding is that the way the models work is they have to collapse all the messages down into a single chunk of text that is fed to the model. For example, [here's what it looks like with LLMA](https://huggingface.co/TheBloke/Llama-2-13B-chat-GPTQ/discussions/5). I'm guessing OpenAI doesn't charge for that additional formatting (and I'd expect the extra tokens to grow linearly instead of be constant if it was the OpenAI formatting), but if LangChain is doing it's own formatting ahead of time, that could be it.

> We ultimately want to know how many tokens the final constructed prompt is going to be that the LLM is seeing right?

It's a good question. The main motivator for this particular feature was to get a sense of how many tokens the system prompt itself uses. This came from the work we're doing in Malawi and recognising that the token-density for low-resource languages is so much higher than English. 

It's a good point that at best this is an approximation because it doesn't take into account any source material that might be subbed in. I think there's a larger update where we want to support source material editing directly in the prompt builder and we'd be able to do estimates there.

I don't think we want to include chat history since we already get that from the conversation token count (which can also be a bit confusing). 


> The number of input tokens seems to be system prompt token count + user token count + 15?

By this I just meant that when I did some testing, the number of input tokens returned by the OpenAI API seemed to be equal to the number of system prompt tokens counted + the number of tokens for the user message I had input + this constant 15, which is possibly from LangChain

snopoke (commented):
This seems OK to me assuming it's purely for estimation purposes. I think the only way to get an accurate count would be to send it to the API and read the token count out of the response.

bderenzi:
Yeah, definitely just for estimation. I can add a little tilde before the number to make it clear that it's approximate?

PR #34, opened 2023-10-09
Add task to send a message to the user

SmittieC:
Added [django-celery-beat](https://pypi.org/project/django-celery-beat/) to add periodic tasks (this new app adds 20 migrations). This will allow the admin user to schedule a bot message to be send to a range of users by adding the following as keyword arguments to the Periodic Task:

image

The 'is_bot_instruction' determines whether the specified message should be sent to the bot first, or if this message should go to the users directly. Adding a UI on top of this is to be determined.

Note: A future PR will also replace the `FutureMessage` model with the `PeriodicTask` moddel, so we don't have to handle the scheduling etc.

### DIFF
--- a/apps/chat/tasks.py
+++ b/apps/chat/tasks.py
@@ -6,6 +6,7 @@ import pytz
 from celery.app import shared_task
 from django.db.models import OuterRef, Subquery
 
+from apps.channels.models import ChannelSession
 from apps.chat.bots import get_bot_from_experiment
 from apps.chat.exceptions import ExperimentChannelRepurposedException
 from apps.chat.message_handlers import MessageHandler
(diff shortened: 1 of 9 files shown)


czue (approved):
Didn't review too closely but looks good to me. Are there any migration concerns with the database celery scheduling or that will be seamless?

heroku.yml
```diff
@@ -20,5 +20,5 @@ run:
     image: django
   worker:
     command:
-      - celery -A gpt_playground worker -l INFO --beat --concurrency 2
+      - celery -A gpt_playground worker -l INFO --beat --concurrency 2 --scheduler django_celery_beat.schedulers:DatabaseScheduler
```
snopoke, line 23: FYI you can use the `CELERY_BEAT_SCHEDULER` setting to set this globally which will prevent having to set it in each place

snopoke (approved):

apps/chat/tasks.py
```diff
@@ -26,6 +27,36 @@ def periodic_tasks(self):
         _check_future_messages()
 
 
+@shared_task
+def send_bot_message_to_users(*args, **kwargs):
```
snopoke, line 31: why use `*args, **kwargs` and not normal arguments?

snopoke (approved):

apps/chat/tasks.py
SmittieC: Initially when I tested it it didn't work, so I just used args + kwargs and forgot about it tbh. I checked again and it worked now. I think I misspelled a parameter previously. Thanks for raising this

SmittieC:
> Are there any migration concerns with the database celery scheduling or that will be seamless?

I don't foresee any reason why there should be? I tested this against a copy of prod's data and everything was fine. Are you thinking about something in particular?

heroku.yml
SmittieC: Ah thanks!

czue:
> Are you thinking about something in particular?

Oh no, I just wasn't sure if there were existing scheduled tasks that needed to migrated to use the new scheduler.

SmittieC:
> > Are you thinking about something in particular?
> 
> Oh no, I just wasn't sure if there were existing scheduled tasks that needed to migrated to use the new scheduler.

The only scheduled task would be [apps.chat.tasks.periodic_tasks](https://github.com/dimagi/open-chat-studio/blob/ff1c1a7bb09ca83031ddc0da69a2e02fc3ec3d71/gpt_playground/celery.py#L21-L28), but after deployment this should be picked up by the new scheduler. Other than this, nothing else

snopoke (approved):

Issue #16, opened 2023-09-25
Fix "smart" reminder


SmittieC:
The implementation of "Reminders" isn't working very well. Time and again the bot seems to halluncinate times / timezones. A rough suspicion is that the parameters might be too complicated for the LLM to figure out sometimes. Maybe could
1. Inject text in the prompt to tell the bot to only invoke the tool when it is absolutely certain that it has all the parameters for the method
2. Change the tool's parameters to make it easier for the bot to figure out.
3. Change the feature?

SmittieC:
The mentioned PR makes this more robust and messages are not being scheduled in the past anymore

PR #35, opened 2023-10-09
Replace `FutureMessage` with PeriodicTask

SmittieC:
Since [this PR](https://github.com/dimagi/open-chat-studio/pull/34) introduces django-celery-beat that makes scheduling periodic tasks easy, we can remove the `FutureMessage` model in favour of the `PeriodicTask` one. Doing this, we don't have to worry about scheduling and can actually remove a bunch of code. (There are currently no `FutureMessage`'s to port to `PeriodicTask`s, so we should be fine simply dropping the table

I also improved the bot's ability to schedule messages. See [this issue](https://github.com/dimagi/open-chat-studio/issues/16). By splitting the "reminder" tool into "one-off" and "recurring" reminders, the LLM is now much better at providing correct parameters to these functions/tools. I noticed it had some difficulty sometimes with the single reminder tool.

Another issue was that the LLM did not always invoke the "current-datetime" tool before trying to set the reminder, causing it to schedule them in the past, which is obviously wrong. By sending the current datetime as a system message during local testing, I got a 100% success rate on the scheduled datetimes, which is promising.


apps/chat/agent/tools.py
```diff
@@ -32,44 +37,55 @@ def action(*args, **kwargs):
...
+        interval_schedule, _created = IntervalSchedule.objects.get_or_create(every=every, period=period)
+        channel_session = ChannelSession.objects.filter(experiment_session=self.experiment_session).first()
+        PeriodicTask.objects.create(
+            name=f"reminder-{self.experiment_session.id}-{uuid.uuid4()}",
+            task=BOT_MESSAGE_FOR_USER_TASK,
+            kwargs=json.dumps({"chat_ids": [channel_session.external_chat_id], "message": message}),
+            start_time=datetime_due,
+            expires=datetime_end,
+            interval=interval_schedule,
+        )
```
snopoke, line 64: this could be extracted to a function which would make it easier to read

apps/chat/agent/tools.py
```diff
@@ -32,44 +37,55 @@ def action(*args, **kwargs):
...
-            interval_minutes=interval_minutes if repeating else 0,
-            end_date=end_date,
+        channel_session = ChannelSession.objects.filter(experiment_session=self.experiment_session).first()
+        PeriodicTask.objects.create(
+            name=f"reminder-{self.experiment_session.id}-{uuid.uuid4()}",
+            task=BOT_MESSAGE_FOR_USER_TASK,
+            kwargs=json.dumps({"chat_ids": [channel_session.external_chat_id], "message": message}),
+            clocked=ClockedSchedule.objects.create(clocked_time=datetime_due),
+            one_off=True,
         )
```
snopoke, line 87: similar to above - could extract to a function

apps/chat/conversation.py
```diff
@@ -39,9 +43,13 @@ def __init__(
         if experiment_session and experiment_session.experiment.tools_enabled:
             self.executer = AgentExecuter(llm=llm, memory=memory, experiment_session=experiment_session)
             # Insert the messages here
+            current_date_message = SystemMessagePromptTemplate.from_template("{current_date}")
             self.executer.agent.prompt = OpenAIFunctionsAgent.create_prompt(
                 system_message=prompt_to_use,
-                extra_prompt_messages=[MessagesPlaceholder(variable_name="history")],
+                extra_prompt_messages=[
+                    MessagesPlaceholder(variable_name="history"),
+                    current_date_message.format(current_date=str(current_datetime)),
```
snopoke, line 51: How does this work with history? Do we end up with multiple 'current_dates' in the chat history?

SmittieC: Good question. Only the content of the user message / bot response get's saved to the chat history. See [this](https://github.com/dimagi/open-chat-studio/blob/95dff42ed292a391df3f420b77144aa4889846a6/apps/chat/bots.py#L118-L127).

apps/chat/agent/tools.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/35/commits/64bf25d03de4c2ca992939257055abd0e457a4e3

apps/chat/agent/tools.py
```diff
@@ -77,15 +74,25 @@ def action(
...
-        )
+        periodic_task_kwargs = {
+            "clocked": ClockedSchedule.objects.create(clocked_time=datetime_due),
+            "one_off": True,
+        }
+        create_periodic_task(self.experiment_session, message=message, kwargs=periodic_task_kwargs)
         return "Success"
 
 
+def create_periodic_task(experiment_session: ExperimentSession, message: str, kwargs: dict):
```
snopoke, line 85: I'm not suggesting a change, just making note that another option here is to accept **kwargs and pass those through instead of accepting a dict. It can make it easier for the caller since they don't have to construct a dict.

snopoke (approved):

apps/chat/agent/tools.py
SmittieC: Hmm..good point. I'll make the update, it's small enough

SmittieC: https://github.com/dimagi/open-chat-studio/pull/35/commits/738e80e08f9bb5e307cf3cdf329483c249f5af64

Issue #29, opened 2023-10-05
Remove default consent form


bderenzi:
We should probably remove the hardcoded default consent form since it reference's Dimagi names and contact details:

https://github.com/dimagi/open-chat-studio/blob/1c6de115cdea269126cb2346cf74e4d6e94569bb/templates/experiments/consent/consent_default.md

Having a default is still a good idea, but should probably be configured somewhere in the DB instead of in the repo.

SmittieC:
The user can create their own form and override the default one for the experiment.
![image](https://github.com/dimagi/open-chat-studio/assets/64970009/a58fa1d1-47d6-4e16-b36c-cdd4f54193e7)

SmittieC:
Oh I misunderstood the initial issue then. Coolios!

PR #41, opened 2023-10-11
default consent

snopoke:
Resolves https://github.com/dimagi/open-chat-studio/issues/29

This removes the Dimagi consent content and creates a default consent model instance in the DB which will be used for all experiments if not otherwise specified when creating the experiment.

The content is very minimal (open to suggestions for improving it).

We will have to manually update the default content on any running instance after deploy.


apps/experiments/models.py
```diff
@@ -224,11 +229,9 @@ class Experiment(BaseModel):
     public_id = models.UUIDField(default=uuid.uuid4, unique=True)
     consent_form = models.ForeignKey(
         ConsentForm,
-        null=True,
-        blank=True,
```
snopoke, line 228: note that this is now a required field

apps/experiments/models.py
```diff
@@ -122,10 +123,18 @@ class ConsentForm(BaseModel):
 
     name = models.CharField(max_length=50)
     consent_text = models.TextField(help_text="Custom markdown text")
+    is_default = models.BooleanField(default=False)
```
SmittieC, line 126: Should this not be unique as well?

snopoke: I could but in the future I think this will be per tenant so then it won't be unique on a different axis. I'm fine leaving it as is for now. It's a readonly field in the admin as well.

snopoke: I've also updated the model field by adding `editble=False` (https://docs.djangoproject.com/en/4.2/ref/models/fields/#editable)

SmittieC: Cool!

SmittieC (approved):
Nice!

apps/experiments/models.py
```diff
@@ -123,7 +123,7 @@ class ConsentForm(BaseModel):
 
     name = models.CharField(max_length=50)
     consent_text = models.TextField(help_text="Custom markdown text")
-    is_default = models.BooleanField(default=False)
+    is_default = models.BooleanField(default=False, editable=False)
```
SmittieC, line 126: This requires a migration?

apps/experiments/migrations/0033_default_consent.py
```diff
@@ -0,0 +1,65 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("experiments", "0032_load_synthetic_voices"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="consentform",
+            name="is_default",
+            field=models.BooleanField(default=False),
```
snopoke, line 51:
```suggestion
            field=models.BooleanField(default=False, editable=False),
```

PR #43, opened 2023-10-11
crud scaffolding and views for safety layer

snopoke:
Peek 2023-10-11 16-37


templates/experiments/experiment_create.html
```diff
@@ -0,0 +1,14 @@
+{% extends 'web/app/app_base.html' %}
+{% load form_tags %}
```
SmittieC: For my curiosity, where does `form_tags` come from?

snopoke: `apps/web/templatetags/form_tags.py` - it's a Pegasus thing.

I'm considering trying out It would be nice to use crispy forms but

SmittieC: Ah there it is, thanks!

> I'm considering trying out https://github.com/django-crispy-forms/crispy-tailwind

I have no reservations against it

SmittieC (approved):
This looks awesome!

Issue #17, opened 2023-09-25
Ability to reset conversation


SmittieC:
We want to be able to send a command (/start or /reset) to the bot. This should then start a new experiment/channel session for this experiment and user.

PR #45, opened 2023-10-12
Reset conversations

SmittieC:
Issue #17

The flow:

image

- Resetting conversations simply creates new experiment sessions which has no chat history.
- The reset command is not saved in the chat history
- Webchats are not able to reset conversations, since they have the luxury of simply creating a new one

### DIFF
--- a/apps/channels/models.py
+++ b/apps/channels/models.py
@@ -5,7 +5,7 @@ from django.conf import settings
 from django.db import models
 from django.db.models import JSONField
 from django.urls import reverse
-from telebot import TeleBot, apihelper
+from telebot import TeleBot, apihelper, types
 
 from apps.experiments.models import Experiment, ExperimentSession
 from apps.utils.models import BaseModel
@@ -13,6 +13,8 @@ from apps.web.meta import absolute_url
 
 
 class ExperimentChannel(BaseModel):
+    RESET_COMMAND = "/reset"
+
     PLATFORM = (("telegram", "telegram"), ("web", "web"), ("whatsapp", "whatsapp"))
     name = models.CharField(max_length=40, help_text="The name of this channel")
     experiment = models.ForeignKey(Experiment, on_delete=models.CASCADE, null=True, blank=True)
@@ -45,6 +47,7 @@ def _set_telegram_webhook(experiment_channel: ExperimentChannel):
         webhook_url = None
 
     tele_bot.set_webhook(webhook_url, secret_token=settings.TELEGRAM_SECRET_TOKEN)
+    tele_bot.set_my_commands(commands=[types.BotCommand(ExperimentChannel.RESET_COMMAND, "Restart chat")])
 
 
 # TODO: Remove this model
--- a/apps/channels/models.py
+++ b/apps/channels/models.py
@@ -40,7 +40,7 @@ def _set_telegram_webhook(experiment_channel: ExperimentChannel):
     """
     Set the webhook at Telegram to allow message forwarding to this platform
     """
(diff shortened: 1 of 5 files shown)


apps/chat/message_handlers.py
```diff
@@ -19,6 +20,8 @@
 from apps.chat.exceptions import AudioSynthesizeException, MessageHandlerException
 from apps.experiments.models import ExperimentSession, SessionStatus
 
+RESET_COMMAND = "/ocs-reset"
```
snopoke, line 23: Why not just `/reset`? Might be nice to support multiple options which could make sense to different users (and different languages).

e.g. `/clean-slate`, `/startover`, `/newchat`

SmittieC: > Why not just /reset?

I don't want this to interfere with "native" commands. On telegram, anything starting with / is a command, so I think prepending "ocs-" makes it a bit more specific to OCS. I don't have strong feelings on this though and can be easily persuaded

> Might be nice to support multiple options which could make sense to different users

I like this


> and different languages

Hmm this might be trickier

apps/channels/models.py
```diff
@@ -5,14 +5,16 @@
...
-from telebot import TeleBot, apihelper
+from telebot import TeleBot, apihelper, types
 
 from apps.experiments.models import Experiment, ExperimentSession
 from apps.utils.models import BaseModel
 from apps.web.meta import absolute_url
 
 
 class ExperimentChannel(BaseModel):
+    RESET_COMMAND = "/reset"
```
SmittieC, line 16: Going back on my previous comment about commands: I did a bit of digging and `/reset` is not a telegram command. Bot's by default don't have commands, so we can 1. use this command and 2. register it as a bot command, so when the user types "/" (or click on the new Menu button), the list of available commands with a description will be shown. Example:
image

apps/channels/models.py
```diff
@@ -40,7 +40,7 @@ def _set_telegram_webhook(experiment_channel: ExperimentChannel):
     """
     Set the webhook at Telegram to allow message forwarding to this platform
     """
-    tele_bot = TeleBot(experiment_channel.extra_data.get("bot_token", ""), parse_mode=None)
+    tele_bot = TeleBot(experiment_channel.extra_data.get("bot_token", ""), threaded=False)
```
SmittieC, line 43: Threaded=True will start threads that handles certain events happening inside TeleBot. We don't use that, so it's safe to not start the threads rather.

apps/channels/models.py
SmittieC: cc @bderenzi

snopoke (approved):

SmittieC:
rebased

snopoke (approved):

SmittieC:
After the fix, I tested it again against a copy of prod's data

apps/channels/models.py
bderenzi: Cool, I'm happy with `/reset`. And +1 on being able to expand it in the future to have additional aliases. I'm not sure `/reset` will be the most obvious for young people in Kenya and certainly wouldn't expect that to be true in Senegal.

SmittieC: We could maintain a list of reset commands in different languages (maintenance burden), or ask the bot to tell us if the user wants to reset the chat (token burden).

bderenzi: I think I'm pretty against asking the bot to tell us. I'd say let's defer any further action until it becomes a problem.

SmittieC: agreed with both.

PR #48, opened 2023-10-12
Add support for Azure TTS

bderenzi:
One thing to note is that I'm writing a temporary file and reading it (then deleting it). There's probably a better pattern for the clean up when there's an exception?

I really just want this to be able to share some Swahili-based voice bots with colleagues to see how well it works.


apps/experiments/migrations/0035_alter_syntheticvoice_language_code_and_more.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 4.2 on 2023-10-11 15:30
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
snopoke, line 6: this can be combined with migration 0033 (just delete both and re-run `makemigrations`)

apps/experiments/migrations/0034_update_AWS_synthetic_voices.py
```diff
@@ -0,0 +1,46 @@
+# Generated by Django 4.2 on 2023-10-11 15:18
+import os, json
+from django.db import migrations
+
+
+def load_language_code(apps, schema_editor):
```
snopoke, line 6: Since these have already been loaded I think you can just run `SyntheticVoice.objects.update(service="AWS")`

apps/experiments/models.py
```diff
@@ -150,25 +150,38 @@ class SyntheticVoice(BaseModel):
         ("male (child)", "Male (Child)"),
         ("female (child)", "Female (Child)"),
     )
+
+    SERVICES = (
+        ("AWS", "AWS"),
+        ("azure", "Azure"),
```
SmittieC, line 156: Small nit: azure -> Azure

apps/experiments/migrations/preload_data/aws_voices.json
```diff
@@ -5,651 +5,744 @@
...
+        "service": "AWS"
       },
       {
         "language": "Arabic (Gulf)",
         "language_code": "ar-AE",
         "name": "Hala",
         "gender": "Female",
-        "neural": true
+        "neural": true,
+        "service": "AWS"
```
SmittieC, line 17: Nit: Technically we don't have to add the "service" field to each entry, since we know this file contains onyl AWS voices. We can add the service field when loading them

apps/experiments/migrations/0034_update_AWS_synthetic_voices.py
SmittieC: As per my understanding the issue is that the language_code isn't loaded on the existing aws voices, so we need to 1. fetch the correct synthetic voice entry using the other fields (name, language, gender etc) and 2. update it's specific language code. This relates to the reason why the two migrations cannot be merged i.e. the language code should first be populated before locking it down to null=False.

If we really want only one migration, I'm wondering if we shouldn't add a default nonsense language code, and update it in a subsequent data migration? It feels like a bad approach though. Not sure if there are other better approaches to handling this @snopoke ?

apps/experiments/migrations/0036_load_azure_synthetic_voices.py
```diff
@@ -0,0 +1,39 @@
...
+
+    voices_created = 0
+    for voice in voice_data:
+        _, created = SyntheticVoice.objects.get_or_create(
+            name=voice["name"],
+            language=voice["language"],
+            language_code=voice["language_code"],
+            gender=voice["gender"],
+            neural=voice["neural"],
+            service=voice["service"],
```
SmittieC, line 23: See [this](https://github.com/dimagi/open-chat-studio/pull/48#discussion_r1356563704). Feel free to address if you want

apps/experiments/migrations/0034_update_AWS_synthetic_voices.py
snopoke: Ah right, language code... To keep it in one migration you can add it as a nullable field and then add in the 'not null' constraint.

apps/channels/audio.py
```diff
@@ -17,6 +20,60 @@ def get_transcript(audio: BytesIO) -> str:
...
+
+            return BytesIO(file_content), duration_seconds
+        else:
+            raise AudioSynthesizeException(f"Azure audio synthesis failed with reason: {result.reason}. Req")
+
+    except Exception as e:
+        raise AudioSynthesizeException(f"Unable to synthesize audio with Azure: {e}")
+
+
+def AWS_synthesize_voice(text: str, synthetic_voice: SyntheticVoice) -> Tuple[BytesIO, float]:
```
SmittieC, line 85: typically function names are in snake case

apps/experiments/migrations/0035_alter_syntheticvoice_language_code_and_more.py
bderenzi: I didn't realise the rebase broke the migrations as well. I think I've fixed it all in #a29797e

apps/channels/audio.py
```diff
@@ -17,6 +20,60 @@ def get_transcript(audio: BytesIO) -> str:
...
+def azure_synthesize_voice(text: str, synthetic_voice: SyntheticVoice) -> Tuple[BytesIO, float]:
+    """
+    Calls Azure's cognitive speech services to convert the text to speech using the synthetic_voice
+    """
+    try:
+        speech_config = speechsdk.SpeechConfig(
+            subscription=settings.AZURE_SUBSCRIPTION_KEY, region=settings.AZURE_REGION
+        )
+
+        temp_file = tempfile.NamedTemporaryFile(delete=False, suffix=".wav")
```
SmittieC, line 40: why the `delete=False`? my understanding from looking at the code is that speechsdk writes to the temp_file, but if `delete=True`, that would remove the temp_file again? Is that what's happening? :thinking:

apps/experiments/migrations/0034_update_AWS_synthetic_voices.py
bderenzi: As opposed to doing it with the default that doesn't get preserved when I add the field? The issue there is that I can't make it a required field without the default.

apps/experiments/models.py
bderenzi: https://github.com/dimagi/open-chat-studio/pull/48/commits/a29797eba6fa3aebf71bbfc8b38fbc9895561606

apps/channels/audio.py
```diff
@@ -17,6 +20,60 @@ def get_transcript(audio: BytesIO) -> str:
...
+        audio_config = speechsdk.audio.AudioConfig(filename=temp_file.name)
+
+        # Setup voice font (Azure refers to synthetic voices as voice fonts)
+        speech_config.speech_synthesis_voice_name = f"{synthetic_voice.language_code}-{synthetic_voice.name}"
+
+        # Create a speech synthesizer
+        synthesizer = speechsdk.SpeechSynthesizer(speech_config=speech_config, audio_config=audio_config)
+
+        # Synthesize the text
+        result = synthesizer.speak_text_async(text).get()
```
SmittieC, line 52: Should this operation not be blocking i.e. use `speak_text` instead?

apps/experiments/migrations/preload_data/aws_voices.json
bderenzi: Good point. One advantage of keeping it is that it stays with the object and/or we could merge the list of objects together in the future. It does use more characters, though. (I'm going to leave it unless you feel strongly)

apps/channels/audio.py
bderenzi: I wonder if we should add that check to the pre-commit formatting stuff? Anyway, https://github.com/dimagi/open-chat-studio/pull/48/commits/3758fbb3f1294522021fad16d44f0096057d867a

SmittieC: Not sure if the lemon would be worth the squeeze? But do feel free to do it if you're up for it. I think a review will quickly sniff em' stray names out :smile:

apps/channels/audio.py
SmittieC: You can try setting delete=True (the default) and see if it's still working. If it is, we can chuck the `os.unlink(temp_file.name)` pieces

bderenzi: According to my coding partner, if you have `delete=True`, then the file is deleted as soon as it is closed. Since I'm re-opening the file to read out the contents and pass it to `pydub`, then I need to not have it automatically deleted. I didn't completely understand. Looking at the [docs](https://docs.python.org/3/library/tempfile.html) now:

> If delete is true (the default) and delete_on_close is true (the default), the file is deleted as soon as it is closed. If delete is true and delete_on_close is false, the file is deleted on context manager exit only, or else when the [file-like object](https://docs.python.org/3/glossary.html#term-file-like-object) is finalized. Deletion is not always guaranteed in this case (see [object.__del__()](https://docs.python.org/3/reference/datamodel.html#object.__del__)). If delete is false, the value of delete_on_close is ignored.

I don't really understand the context manager, so defer to you all about how to interpret this. My coding partner thinks an alternative approach (which might be cleaner and remove all of the clean up code) would be:

```
with tempfile.NamedTemporaryFile(delete=True, delete_on_close=False) as tmp:
    # Work with the temp file
    pass
# Temp file is deleted here when the with block is exited
```

I assume that an exception generated within a `with` block would also trigger the `__exit__()` and clean up the temp file?

apps/channels/audio.py
bderenzi: No idea. I changed it though https://github.com/dimagi/open-chat-studio/pull/48/commits/28a18f1a98efbdae682b2ee8a3b6bd02a170e5d3

apps/channels/audio.py
SmittieC: > I assume that an exception generated within a with block would also trigger the __exit__() and clean up the temp file?

yes. if possible, I like the idea to work with everything inside the context manager. Another approach is to use BytesIO?

apps/experiments/migrations/0035_alter_syntheticvoice_language_code_and_more.py
bderenzi: (it's still broken because `main` changed again. Fixed in https://github.com/dimagi/open-chat-studio/pull/48/commits/2ffb9f64f08378d402aa442c68c2a309562e71a4

apps/experiments/migrations/preload_data/aws_voices.json
SmittieC: No strong feelings about this. Good to leave it as is

apps/experiments/migrations/0036_update_AWS_synthetic_voices.py
```diff
@@ -0,0 +1,47 @@
...
+        synth_voice, created = SyntheticVoice.objects.get_or_create(
+            name=voice["name"],
+            language=voice["language"],
+            gender=voice["gender"],
+            neural=voice["neural"],
+            service=voice["service"],
+        )
+
+        synth_voice.language_code=voice["language_code"]
+        synth_voice.save()
```
snopoke, line 27: You can do this more easily with 'update_or_create':
```suggestion
        synth_voice, created = SyntheticVoice.objects.update_or_create(
            name=voice["name"],
            gender=voice["gender"],
            neural=voice["neural"],
            defaults={
                "language": voice["language"],
                "service": voice["service"]
            }
        )
```

gpt_playground/settings.py
```diff
@@ -425,3 +425,6 @@
     "class": "border-b border-gray-200 hover:bg-gray-100",
     "id": lambda record: f"record-{record.id}",
 }
+
+AZURE_SUBSCRIPTION_KEY = os.environ.get("AZURE_SUBSCRIPTION_KEY")
+AZURE_REGION = os.environ.get("AZURE_REGION")
```
snopoke, line 430: this should be using 'environ' so that it also supports local .env files:

```suggestion
AZURE_SUBSCRIPTION_KEY = env("AZURE_SUBSCRIPTION_KEY", default="")
AZURE_REGION = env("AZURE_REGION", default="")
```

It would be good to also put these in `.env.example` so that it's kept up to date.

snopoke (changes requested):
Small changes but otherwise looks good

gpt_playground/settings.py
bderenzi: Added them to `.env.example`.

As for the change here, should we switch everything else to use `env()`? All of the other settings currently use `os.environ.get()` like mine. Or maybe I misunderstood?

snopoke: > As for the change here, should we switch everything else to use `env()`? All of the other settings currently use `os.environ.get()` like mine. Or maybe I misunderstood?

Some use `env` and some use `os.environ`. No need to change the others here. I'll raise a separate PR for that.

apps/experiments/migrations/0039_update_AWS_synthetic_voices.py
```diff
@@ -0,0 +1,47 @@
...
+    voices_created = 0
+    voices_edited = 0
+    for voice in voice_data:
+        _, created = SyntheticVoice.objects.update_or_create(
+            name=voice["name"],
+            language=voice["language"],
+            gender=voice["gender"],
+            neural=voice["neural"],
+            service=voice["service"],
+            defaults={
```
snopoke, line 24: this should be in the `defaults` since it's a new field:
```suggestion
            defaults={
                "service": voice["service"],
```

bderenzi:
Another rebase.

SmittieC:
@bderenzi I think you have to run the `inv requirements` command again. When I run it on this branch, it updates the requirements.txt file with greenlet

bderenzi:
Interesting. I just reran it and the only thing it did for me was to:
1. Change the comment at the top about how/when it was run
2. Add some stuff about an extra index url with Nvidia, which is a hold over from the early days when there were some GPU requirements with the indexing code. 

`greenlet` isn't installed on my system.

gpt_playground/settings.py
bderenzi: I forgot to mention, but I did this in https://github.com/dimagi/open-chat-studio/pull/48/commits/44e1ef231537c1a8b05c2bc60580551c3342adf6

apps/experiments/migrations/0039_update_AWS_synthetic_voices.py
bderenzi: Also should have mentioned this was done in https://github.com/dimagi/open-chat-studio/pull/48/commits/28e0bf3ff7a2a3849a56526089598781a88ee31c

snopoke (approved):

SmittieC:
> Interesting. I just reran it and the only thing it did for me was to:
> 
> 1. Change the comment at the top about how/when it was run
> 2. Add some stuff about an extra index url with Nvidia, which is a hold over from the early days when there were some GPU requirements with the indexing code.
> 
> `greenlet` isn't installed on my system.

Interesting. Then it's maybe just my virtual env. Good to go then!

bderenzi:
I did one more rebase (no conflicts) to deal with the merge conflict.

Let me know if I should rather pull than rebase? It probably doesn't matter since we weren't collaborating directly on this PR?

bderenzi:
Manual migrations are a headache. This wouldn't be such a slog if I could just delete and recreate them.

snopoke:
> I did one more rebase (no conflicts) to get it ready to merge with main
> 
> Let me know if I should rather pull than rebase? It probably doesn't matter since we weren't collaborating directly on this PR?

Rebasing does make it harder to review since all the commit hashes change. Ideally only rebase if you need to amend change the commit history for some reason.

PR #49, opened 2023-10-12
more crud

snopoke:
I think this covers the rest of the models except `NoActivityMessageConfig`


SmittieC (approved):
Looks excellent! :rocket:

Issue #24, opened 2023-09-28
Multi-tenancy support


bderenzi:
We would like to add multi-tenancy support. There is a partially implemented version of this called Teams.

czue:
There are some notes on this in [this doc](https://docs.google.com/document/d/1ELpv0TNkEjJfmxepHqThEJbHyFMzxXNeZWBLdhMw7Ws/edit#_

PR #50, opened 2023-10-12
Multi-tenancy

snopoke:
Resolves: https://github.com/dimagi/open-chat-studio/issues/24

Best reviewed by commit

* Add team to all necessary models
* All existing data owned by Team(1)
* All users added to Team(1)
* View queries updated to filter by team

**Safety Story**
I have tested this locally using a dump of the 'dimagi' instance DB and confirming that the migrations run as expected.

After running the migration I inspected all pages and tested a lot of features to ensure that they still work.

**TODO**

* remove remaining redirects to admin UI
  * main experiment page
  * prompt builder save


SmittieC:
From a quick glance it looks like the `ExperimentChannel` is missing a team? (I'm grabbing coffee now for a more in-depht review)

snopoke:
> Awsome stuff! From a quick glance it looks like the `ExperimentChannel` is missing a team?

That's correct but it is always associated with an `Experiment`. The same for `ChatMessage` which is always associated with a `Chat`.

We could add team to them but it doesn't seem necessary at this point.

SmittieC:
> > Awsome stuff! From a quick glance it looks like the `ExperimentChannel` is missing a team?
> 
> That's correct but it is always associated with an `Experiment`. The same for `ChatMessage` which is always associated with a `Chat`.
> 
> We could add team to them but it doesn't seem necessary at this point.

Yeah that's fair

apps/utils/teams_migration.py
```diff
@@ -0,0 +1,15 @@
...
+    return Team.objects.order_by("id").first()
+
+
+def assign_model_to_team_migration(model_name, delete_if_no_team=False):
+    def _migration_op(apps, schema_editor):
+        model_cls = apps.get_model(model_name)
+        default_team = _get_default_team(apps)
+        if default_team:
+            model_cls.objects.update(team=default_team)
+        elif delete_if_no_team:
```
SmittieC, line 12: I'm struggling to understand when/why this is needed?

snopoke: During initial setup (and during tests) there is are no teams so we delete the global default and then when new teams are created the team default get's created.

SmittieC: Got it. So the code should not enter this branch in prod then

SmittieC (approved):
Looks good to me, although I noticed two things:
1. An administrator of a team is not able to add/edit experiments (only the superuser can do this).
2. when creating a new experiment in say, team 2, I can (as a superuser) see and link a prompt from team 1 to this experiment

snopoke:
> 1. An administrator of a team is not able to add/edit experiments (only the superuser can do this).

I kept this as it is now since most users won't be administrators. Improving permissions is a whole other set of changes.

> 2. when creating a new experiment in say, team 2, I can (as a superuser) see and link a prompt from team 1 to this experiment.

ah - this is an issue which I'll fix in this PR

snopoke:
> > 2. when creating a new experiment in say, team 2, I can (as a superuser) see and link a prompt from team 1 to this experiment.
> 
> ah - this is an issue which I'll fix in this PR

Resolved in 263be36be53e10ee2ecc109325246782a794c7ff

snopoke:
@SmittieC I found an issue with Participants which already had a team before. I've added a new migration to handle it:

https://github.com/dimagi/open-chat-studio/pull/50/commits/4484351e4f82b20aa4ebc4502a25a5c601ebd7e3

snopoke:
I'm going to wait till Monday to roll this out

apps/experiments/migrations/0039_add_users_to_default_team.py
```diff
@@ -0,0 +1,24 @@
...
+from django.conf import settings
+from django.db import migrations
+
+
+def add_users_to_default_team(apps, schema_editor):
+    Team = apps.get_model("teams", "Team")
+    User = apps.get_model(settings.AUTH_USER_MODEL)
+    Membership = apps.get_model("teams", "Membership")
+    default_team = Team.objects.order("id").first()
+    for user in User.objects.all():
```
czue: I don't totally understand this logic / why this is necessary. If Users were part of a different team would it not make sense to explicitly have them added to the team manually, as opposed to doing it automatically?

czue (commented):
Didn't review in depth but looks good overall. Two big picture thoughts/questions (not actionable or blockers, just writing them down):

- Do you think there is a better way to handle team-based filtering (and setting on create?) that uses something like middleware + threadlocals to always ensure the proper filters get put in place? I'm not sure exactly what that would look like, but it just struck me that it would be nice to not have to handle it explicitly everywhere if there was a way to do that.
- Related - tn terms of testing, I think the largest potential remaining risk is that some functionality accidentally "leaks" across teams. In your manual testing did you also create a new team and confirm that none of the existing Dimagi stuff showed up in all the common flows? Note: this is a lower-risk issue if we don't expose multi-tenancy to anyone outside Dimagi until we do further testing in production.

apps/experiments/migrations/0039_add_users_to_default_team.py
snopoke: Prior to this PR all the data was visible in every team. After this PR it will only be visible in the 'default team' so all existing users who are not already part of that team will lose access. Inviting them could be done manually but this has the same effect.

snopoke:
> * Do you think there is a better way to handle team-based filtering (and setting on create?) that uses something like middleware + threadlocals to always ensure the proper filters get put in place? I'm not sure exactly what that would look like, but it just struck me that it would be nice to not have to handle it explicitly everywhere if there was a way to do that.

That approach does appear nice but there are a lot of edge cases. If we did want to do that we could try out this project rather than implement it ourselves: https://github.com/citusdata/django-multitenant

> * Related - tn terms of testing, I think the largest potential remaining risk is that some functionality accidentally "leaks" across teams. In your manual testing did you also create a new team and confirm that none of the existing Dimagi stuff showed up in all the common flows? Note: this is a lower-risk issue if we don't expose multi-tenancy to anyone outside Dimagi until we do further testing in production.

I did create a new team and verified most of the views. I do think this is relatively low risk from that point of view. The higher risk right now is whether there are any links out there which point to a team other than the 'default team' but I've confirmed with our users that there aren't.

PR #52, opened 2023-10-13
[Proposal] Make it easier to test telegram

SmittieC:
See [this wiki](https://github.com/dimagi/open-chat-studio/wiki/Experiment-Channels#note-to-developers) for the steps to test telegram channels in the dev environment. It's a pain. I propose the following implementation that adds two invoke commands. If `runserver ` is run with the `--public` flag, it will use ngrok to get a public url for the dev machine and run the server with that URL. It will give this url to external services when registering webhooks (in Telegram's case) so they can actually talk back to the server.

Suggestions for other approaches are super welcome


snopoke (approved):
seems reasonable to me

PR #55, opened 2023-10-13
set default for 'SITE_URL_ROOT'

snopoke:


snopoke:
Merging since current settings are broken unless the env var is specified.

snopoke:
Merging since the settings are broken unless the env var is specified

PR #56, opened 2023-10-16
production docker setup

snopoke:
* [multi-stage build](https://www.linkedin.com/pulse/harnessing-power-docker-multi-stage-builds-enhancing-de-rinaldis/)
  * faster
  * smaller final image (1GB vs 1.9GB)
* don't use root user

I have run tested the build locally.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 media
 static/css
 static/js
+static_root
 node_modules
 npm-debug.log
 .env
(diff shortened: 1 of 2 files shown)


czue (approved):
This is cool. Is it considered a security issue to use the root user?

snopoke:
> This is cool. Is it considered a security issue to use the root user?

Yes because it means the container is run using the root user on the host system: https://www.howtogeek.com/devops/why-processes-in-docker-containers-shouldnt-run-as-root/

PR #58, opened 2023-10-17
display and CRUD for experiment channels

snopoke:
* Display channels on the experiment details page
* Ability to add, update and delete channels

Peek 2023-10-17 17-13

### DIFF
--- a/apps/channels/admin.py
+++ b/apps/channels/admin.py
@@ -16,6 +16,7 @@ class ExperimentChannelAdmin(admin.ModelAdmin):
     )
     search_fields = ("name",)
     list_filter = (
+        "platform",
         "created_at",
         "updated_at",
     )
@@ -30,7 +31,3 @@ class ExperimentChannelAdmin(admin.ModelAdmin):
 
     def get_changeform_initial_data(self, request: HttpRequest) -> Dict[str, str]:
         return {"extra_data": {"bot_token": "your token here"}}
-
-    def get_queryset(self, request):
-        queryset = super().get_queryset(request)
-        return queryset.exclude(platform="web")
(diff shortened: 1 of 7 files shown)


apps/channels/admin.py
```diff
@@ -30,7 +31,3 @@ def team(self, obj):
 
     def get_changeform_initial_data(self, request: HttpRequest) -> Dict[str, str]:
         return {"extra_data": {"bot_token": "your token here"}}
```
SmittieC, line 33: Out of curiosity, what was the thinking of including them again?

apps/channels/models.py
```diff
@@ -12,16 +12,42 @@
...
+            cls.WHATSAPP,
+        ]
+
+    def form(self):
+        from apps.channels.forms import ChannelForm
+
+        return ChannelForm(initial={"platform": self})
+
+    def extra_form(self, *args, **kwargs):
+        from apps.channels.forms import TelegramChannelForm, WhatsappChannelForm
```
SmittieC, line 33: Can this be moved to the top of the file, or were there circular import errors?

snopoke: It can't because forms imports from models

apps/experiments/views/experiment.py
```diff
@@ -145,17 +145,60 @@ def single_experiment_home(request, team_slug: str, experiment_id: int):
...
+        ExperimentChannel, id=channel_id, experiment_id=experiment_id, experiment__team__slug=team_slug
+    )
+    if request.POST.get("action") == "delete":
+        channel.delete()
+    else:
+        form = channel.form(data=request.POST)
+        if form.is_valid():
+            extra_form = channel.extra_form(data=request.POST)
+            config_data = {}
+            if extra_form and extra_form.is_valid():
```
SmittieC, line 196: would `extra_form` not always exist here?

apps/channels/admin.py
snopoke: This could actually be removed now

SmittieC (approved):

PR #60, opened 2023-10-17
remove remaining links to admin

snopoke:
* remove the link on the experiments page
* add prompt CRUD views
* redirect to prompt create view from prompt builder instead of to admin


SmittieC (approved):
:rocket:

apps/experiments/views/prompt.py
```diff
@@ -4,17 +4,107 @@
...
+            "title": "Prompts",
+            "new_object_url": reverse("experiments:prompt_new", args=[team_slug]),
+            "table_url": reverse("experiments:prompt_table", args=[team_slug]),
+        },
+    )
+
+
+class PromptTableView(SingleTableView):
+    model = Prompt
+    paginate_by = 25
```
SmittieC, line 41: Out of curiosity, does this handle pagination for us?

snopoke: It's handled by [django-tables2](https://django-tables2.readthedocs.io/en/latest/pages/pagination.html). You can see what it's doing in `django_tables2.views.TableMixinBase.get_table_pagination`.

PR #62, opened 2023-10-18
azure cognitive audio dependencies

snopoke:
https://learn.microsoft.com/en-us/azure/ai-services/speech-service/quickstarts/setup-platform

### DIFF
--- a/apps/chat/channels.py
+++ b/apps/chat/channels.py
@@ -188,7 +188,7 @@ class ChannelBase:
             voice_audio, duration = audio.synthesize_voice(text, synthetic_voice=self.experiment.synthetic_voice)
             self.send_voice_to_user(voice_audio, duration)
         except AudioSynthesizeException as e:
-            logging.warning(e)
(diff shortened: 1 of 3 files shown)


apps/chat/channels.py
```diff
@@ -188,7 +188,7 @@ def _reply_voice_message(self, text: str):
             voice_audio, duration = audio.synthesize_voice(text, synthetic_voice=self.experiment.synthetic_voice)
             self.send_voice_to_user(voice_audio, duration)
         except AudioSynthesizeException as e:
-            logging.warning(e)
+            logging.exception(e)
```
bderenzi, line 191: This is cool. Code will keep executing, but it'll log an exception to sentry so we see it?

We want the fallback state of sending as text if the audio fails (at least for now)

snopoke: > This is cool. Code will keep executing, but it'll log an exception to sentry so we see it?

That's correct

PR #64, opened 2023-10-18
change experiment dropdown menu to a toolbar

snopoke:
When I made the dropdown I had intended to change it to a button bar but forgot about it before pushing the changes.

I do think this is a better UI.
Peek 2023-10-18 16-02


bderenzi (approved):
looks good

PR #65, opened 2023-10-18
fix channels

snopoke:
@bderenzi I think the issue was that I made the 'bot_token' field a 'SlugField' but tokens also contain ':' failed the regex validation.

I've removed that validation and also improved the views so that errors are handled properly


bderenzi (approved):
Great, thanks!

PR #72, opened 2023-10-19
data model and crud views for 'LLM Providers'

snopoke:
This PR has more in it because I wanted to try and abstract out the way we do 'config' forms where a model has a type and a config dict. Depending on the type we expect different fields in the config dict. Apart from this new model, `ExperimentChannel` also follows this pattern.

Note this does not connect up the API keys with anything in the backend yet. I'm planning to do that in a separate PR.

Peek 2023-10-20 15-28


SmittieC (approved):
Took me a while to get through, but nothing from my side really. Looks good!

PR #74, opened 2023-10-23
Fix bot sending messages to users

SmittieC:
The function that sends messages from the bot to a range of users does take the specific bot into consideration. Consequently, it will send the message to **all** chat with the specified chat id.

To reproduce the issue:
1. Create two telegram bots
2. Create two experiments (one experiment should have tools enabled) and link bot 1 to experiment 1 and bot 2 to experiment
3. Talk to both
4. In the experiment with tools enabled, ask the bot to send to a ping message in the near future (say, 30 seconds from now)

This PR fixes that by restricting the target audience of the message to a specific bot / experiment. This PR also fixes chat downloads that doesn't seem to work after the channel sessions were moved to the experiment session table

### DIFF
--- a/apps/chat/tasks.py
+++ b/apps/chat/tasks.py
@@ -1,6 +1,7 @@
 import logging
 from datetime import datetime, timedelta
 from typing import List
+from uuid import UUID
 
 import pytz
 from celery.app import shared_task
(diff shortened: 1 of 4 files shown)


apps/chat/tasks.py
```diff
@@ -28,15 +29,17 @@ def periodic_tasks(self):
 
 
 @shared_task
-def send_bot_message_to_users(message: str, chat_ids: List[str], is_bot_instruction: bool):
+def send_bot_message_to_users(message: str, chat_ids: List[str], is_bot_instruction: bool, experiment_public_id: UUID):
     """This sends `message` to the sessions related to `chat_ids` as the bot.
```
SmittieC, line 33: Why the experiment's public id? The feature that Delta's using, she'll need to manually create these periodic tasks and she only has the experiment's public ID at hand. It's sort of to accomodate that feature.

snopoke (approved):

PR #78, opened 2023-10-24
Use per-team LLM Provider configuration for LLM interactions

snopoke:
Review by commit.

With this change the global OpenAI API key will be unused and all interactions will use the API key configured for the experiment.


apps/teams/context_processors.py
```diff
@@ -1,4 +1,33 @@
...
+        "notices": get_team_notices(request),
     }
+
+
+def get_team_notices(request):
+    team = getattr(request, "team", None)
+    if not team:
+        return []
+
+    return filter(None, [notice(request, team) for notice in NOTICES])
```
SmittieC, line 16: Why the `filter(None, ...)`? Wouldn't this just return the iterable?

apps/experiments/migrations/0047_bootstrap_llm_provider.py
```diff
@@ -0,0 +1,28 @@
...
+    Team = apps.get_model("teams", "Team")
+    LlmProvider = apps.get_model("llm_providers", "LlmProvider")
+    Experiment = apps.get_model("experiments", "Experiment")
+
+    for team in Team.objects.all():
+        provider = LlmProvider.objects.filter(team=team).first()
+        if not provider:
+            continue
+
+        Experiment.objects.filter(team=team).update(llm_provider=provider)
```
SmittieC, line 16: Is my understanding correct in that there will exist no `LlmProvider` records on prod yet, and that when this code is deployed, we will need to create one and link them to the experiments after these changes?

SmittieC: If so, that would mean that this migration should be run _after_ that step. Maybe I'm missing something

apps/teams/context_processors.py
snopoke: It removes any that are 'falsy'

apps/experiments/migrations/0047_bootstrap_llm_provider.py
snopoke: Correct. I've already created the LlmProvider on prod

apps/teams/context_processors.py
SmittieC: Cool, did not know that!

apps/experiments/migrations/0047_bootstrap_llm_provider.py
SmittieC: Coolios

SmittieC (approved):
Looks good to me

PR #87, opened 2023-10-26
move LLM Provider to new app

snopoke:
I should have done this originally but 🤷🏻‍♂️ 

This moves the code for LLM providers to a new app called 'service_providers' in preparation for adding code for 'VoiceProviders'. My plan is to use generic views etc for these classes to prevent duplicating a lot of code.

This change isn't strictly necessary but given the similarities of the data it would be nice to have them in the same app.

I've tested this migration locally and it works fine:

* Add new table
* Copy data to new table
* Add new column to experiment to reference new table
* Populate new column in experiment
* Remove old column
* Update all references to old column and model

This preserves the old model and table for now for the same of backward compatibility during deploy.

I have tested this with a prod DB backup.


apps/experiments/models.py
```diff
@@ -231,6 +231,9 @@ class Experiment(BaseTeamModel):
     llm_provider = models.ForeignKey(
         "llm_providers.LlmProvider", on_delete=models.SET_NULL, null=True, blank=True, verbose_name="LLM Provider"
     )
+    llm_provider_new = models.ForeignKey(
```
SmittieC, line 234: The plan is to remove the existing `llm_provider` app and then rename `llm_provider_new` to `llm_provider` here?

apps/service_providers/migrations/0003_migrate_experiments_to_new_model.py
```diff
@@ -0,0 +1,24 @@
...
+    for experiment in Experiment.objects.all():
+        experiment.llm_provider_new_id = LlmProvider.objects.filter(old_id=experiment.llm_provider_id).first().id
+        experiment.save()
+
+
+class Migration(migrations.Migration):
+    initial = True
+
+    dependencies = [
+        ("service_providers", "0002_copy_llm_providers"),
```
SmittieC, line 19: Interesting that this migration doesn't depend on `0048_experiment_llm_provider_new.py` in experiments, since it requires the `llm_provider_new` to exist on experiments?

SmittieC (approved):
Looks pretty straight forward. Just one comment/question

PR #88, opened 2023-10-26
Voice providers part 1

snopoke:
Data models and views for managing voice providers for a team.

PR on top of https://github.com/dimagi/open-chat-studio/pull/87


apps/service_providers/tables.py
```diff
@@ -1,23 +1,32 @@
...
 
-    class Meta:
-        model = LlmProvider
-        fields = (
-            "type",
-            "name",
-        )
-        row_attrs = settings.DJANGO_TABLES2_ROW_ATTRS
-        orderable = False
+    class_name = model.__name__ + "AutogeneratedTable"
```
SmittieC, line 16: This is cool

apps/service_providers/utils.py
```diff
@@ -1,36 +1,88 @@
...
-        form.fields["type"].disabled = True
+        form.fields[provider.secondary_key_field].disabled = True
 
     return form
 
 
-def formfield_for_dbfield(db_field, **kwargs):
-    if db_field.name == "type":
-        return forms.TypedChoiceField(empty_value=None, choices=LlmProviderType.choices)
+def formfield_for_dbfield(db_field, provider, **kwargs):
```
SmittieC, line 84: nit: can we add a type hints here?

apps/service_providers/views.py
```diff
@@ -3,35 +3,45 @@
...
-from .tables import LlmProviderTable
-from .utils import get_llm_config_form
+from .utils import ServiceProvider, get_service_provider_config_form
 
 
-class LlmProviderTableView(SingleTableView):
+class ServiceProviderMixin:
+    @property
+    def provider_type(self) -> ServiceProvider:
+        type_ = self.kwargs["provider_type"]
```
SmittieC, line 12: Where does `self.kwargs` come from?

SmittieC (approved):
This was brain stretching, thanks. I wonder if a small `Readme` in this app on how to add more service models might not be a good investment. I can see us adding for instance a transcription service as well in the future

snopoke:
> This was brain stretching, thanks. I wonder if a small `Readme` in this app on how to add more service models/views might not be a good investment. I can see us adding for instance a transcription service as well in the future

I noticed that we already do transcription, I was planning to reuse the LlmProvider for now.

apps/service_providers/views.py
snopoke: They are the standard Django view kwargs.

PR #96, opened 2023-10-31
save chat messages before calling API to get correct 'creation' date

snopoke:
Best reviewed per commit


bderenzi (approved):
This looks great to me. I remember dealing with the `history` in the prompt builder, but looks like that all is well accounted for.

PR #108, opened 2023-10-31
allows signups when accepting invitations

### LINKED ISSUE #107: Fix user invitation workflow
When accepting an invitation the user must be able to signup, even if public signups are closed.

snopoke:
Resolves #107


bderenzi (approved):
That was all that was required? wow.

PR #110, opened 2023-11-01
add taskbadger integration for monitoring celery tasks

snopoke:
This will give us even more visibility into what's going on with the tasks.


SmittieC (approved):
Exciting! I presume there's a dashboard that we can look at?

snopoke:
> Exciting! I presume there's a dashboard that we can look at?

I sent you an invite.

PR #114, opened 2023-11-03
catch exception and return 404 response

snopoke:
For some reason the Http404 exception is not
being handled when raised from middleware

### DIFF
--- a/apps/teams/helpers.py
+++ b/apps/teams/helpers.py
@@ -26,10 +26,7 @@ def get_next_unique_team_slug(team_name: str) -> str:
 def get_team_for_request(request, view_kwargs):
     team_slug = view_kwargs.get("team_slug", None)
     if team_slug:
-        try:
-            return Team.objects.get(slug=team_slug)
-        except Team.DoesNotExist:
-            return None
+        return Team.objects.filter(slug=team_slug).first()
 
     if not request.user.is_authenticated:
         return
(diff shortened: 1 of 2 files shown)


apps/teams/middleware.py
```diff
@@ -28,5 +29,11 @@ def _get_team_membership(request):
 
 class TeamsMiddleware(MiddlewareMixin):
     def process_view(self, request, view_func, view_args, view_kwargs):
-        request.team = SimpleLazyObject(lambda: _get_team(request, view_kwargs))
+        try:
+            request.team = _get_team(request, view_kwargs)
+        except Http404:
+            request.team = None
+            request.team_membership = None
+            return HttpResponseNotFound()
```
SmittieC, line 37: Hmm, this is interesting. What does it show if not the 404 page? Returning `HttpResponseNotFound` puts the onus on us to render the HTML

snopoke: Good point. Alternative approach is to return `None` for the team and let other code like `login_and_team_required` handle the it: [246a50d](https://github.com/dimagi/open-chat-studio/pull/114/commits/246a50d2b22d8420993aff415fcba0494169251d)

apps/teams/helpers.py
```diff
@@ -26,7 +26,10 @@ def get_next_unique_team_slug(team_name: str) -> str:
 def get_team_for_request(request, view_kwargs):
     team_slug = view_kwargs.get("team_slug", None)
     if team_slug:
-        return get_object_or_404(Team, slug=team_slug)
+        try:
```
SmittieC, line 29: Nit: You can just do this then (up to you if you want to update)
```python
return Team.objects.filter(slug=team_slug).first()
```

SmittieC (approved):

PR #115, opened 2023-11-03
Facebook Integration

SmittieC:
Best going through it commit-wise. I'm still figuring out how to make the Meta App that links a Facebook page to OCS live, but I added a Readme in the channels app that explains setting it up for testing.

**The plan is to merge this once we can have a live page/app**


apps/channels/models.py
```diff
@@ -34,18 +37,24 @@ def form(self):
...
             case self.TELEGRAM:
-                return TelegramChannelForm(*args, **kwargs)
+                return forms.TelegramChannelForm(*args, **kwargs)
             case self.WHATSAPP:
-                return WhatsappChannelForm(*args, **kwargs)
+                return forms.WhatsappChannelForm(*args, **kwargs)
+            case self.FACEBOOK:
+                initial = kwargs.get("initial", {})
+                if "verify_token" not in initial:
+                    initial["verify_token"] = str(uuid.uuid4())
```
snopoke, line 50: nit: you can do this with `initial.setdefault("verify_token", str(uuid.uuid4())`

apps/channels/views.py
```diff
@@ -24,3 +25,21 @@ def new_whatsapp_message(request):
     message_data = json.dumps(request.POST.dict())
     tasks.handle_whatsapp_message.delay(message_data)
     return HttpResponse()
+
+
+@csrf_exempt
+def new_facebook_message(request: HttpRequest):
```
snopoke, line 31: Since this has to be setup for each bot, what do you think about adding a `team_slug` to the URL path which would allow us to further filter the channel query.

apps/channels/views.py
```diff
@@ -24,3 +25,21 @@ def new_whatsapp_message(request):
...
+    # https://developers.facebook.com/docs/messenger-platform/webhooks#:~:text=Validating%20Verification%20Requests
+    if request.method == "GET":
+        challenge = request.GET["hub.challenge"]
+        verify_token_exists = ExperimentChannel.objects.filter(
+            platform=ChannelPlatform.FACEBOOK, extra_data__verify_token=request.GET["hub.verify_token"]
+        ).exists()
```
snopoke, line 37: this seems like a good candidate for an manager method and a unit test:

```python
ExperimentChannel.objects.filter_by_extras(extra_key, extra_value, platform, team)
```

snopoke (commented):
This looks great.

apps/channels/views.py
SmittieC: Good idea!

apps/channels/models.py
```diff
@@ -34,18 +37,30 @@ def form(self):
...
+                return forms.WhatsappChannelForm(*args, **kwargs)
+            case self.FACEBOOK:
+                initial = kwargs.get("initial", {})
+                initial.setdefault("verify_token", str(uuid.uuid4()))
+                kwargs["initial"] = initial
+                return forms.FacebookChannelForm(*args, **kwargs)
+
+
+class ExperimentChannelObjectManager(models.Manager):
+    def filter_extras(self, key: str, value: str, platform: ChannelPlatform, team: str):
```
SmittieC, line 55: We can keep the platform and team separate from this method call. I'm in two minds. I think we can change it later if need be

apps/channels/views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/115/commits/c668e30e2a35daf1c3bfe487cf0df07ed2a8d113

apps/channels/views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/115/commits/44d16329299546cbcd99cab4dc2dcfcf654496b8

apps/channels/urls.py
```diff
@@ -7,5 +7,5 @@
 urlpatterns = [
     path("telegram/<uuid:channel_external_id>", views.new_telegram_message, name="new_telegram_message"),
     path("whatsapp/incoming_message", views.new_whatsapp_message, name="new_whatsapp_message"),
-    path("facebook/incoming_message", views.new_facebook_message, name="new_facebook_message"),
+    path("facebook/<slug:team>/incoming_message", views.new_facebook_message, name="new_facebook_message"),
```
snopoke, line 10: You could call this `team_slug` to stay consistent with other places. Also the teams middleware expects that name.

apps/channels/models.py
snopoke: nit: if team and platform and required I'd put them first in the args list. Otherwise make them optional.

I'm having a hard time articulating why I think that's better. Perhaps because they are like constant filters and the key / value are the dynamic ones? I guess it's also analogous to placing the big rocks first: team -> platform -> key+value

apps/channels/tests/model_manager_test.py
```diff
@@ -0,0 +1,52 @@
...
+from django.test import TestCase
+
+from apps.channels.models import ChannelPlatform, ExperimentChannel
+from apps.experiments.models import ConsentForm, Experiment, Prompt
+from apps.service_providers.models import LlmProvider
+from apps.teams.models import Team
+from apps.users.models import CustomUser
+
+
+class TestExperimentChannelObjectManager(TestCase):
```
snopoke, line 11: no need to change this now since you've already written it but I'd encourage you to look into `pytest`. This is a good article to start with: https://djangostars.com/blog/django-pytest-testing/

apps/channels/models.py
```diff
@@ -34,18 +37,30 @@ def form(self):
...
+        extra_data_filter = Q(extra_data__contains={key: value})
+        return super().get_queryset().filter(extra_data_filter).filter(experiment__team__slug=team, platform=platform)
```
snopoke, line 57: nit: you don't need to call `super().get_queryset()`
```suggestion
        return self.filter(extra_data_filter).filter(experiment__team__slug=team, platform=platform)
```

apps/channels/tasks.py
```diff
@@ -27,3 +27,30 @@ def handle_whatsapp_message(message_data: str):
...
+        content_type = attachment["type"]
+
+    message = FacebookMessage(
+        user_id=message["sender"]["id"],
+        page_id=page_id,
+        message_text=message["message"].get("text", ""),
+        media_url=media_url,
+        content_type=content_type,
+    )
+    experiment_channel = ExperimentChannel.objects.filter(extra_data__contains={"page_id": message.page_id}).first()
```
snopoke, line 52: could this also use the `filter_extras` manager method?

apps/channels/forms.py
```diff
@@ -21,3 +21,11 @@ class TelegramChannelForm(forms.Form):
 
 class WhatsappChannelForm(forms.Form):
     number = forms.CharField(label="Number", max_length=100)
+
+
+class FacebookChannelForm(forms.Form):
```
snopoke, line 26: You could add the webhook URL into the form as a disabled field with a help text giving some context:

```
webook_url = forms.Charfield(label="Webhook URL", disabled=True,
    help_text="Copy this into the Facebook setup form.")
```

apps/channels/tasks.py
SmittieC: Ah thanks

SmittieC: https://github.com/dimagi/open-chat-studio/pull/115/commits/ce2298bedab43b56ec9d9aeb6e4fb6a1a217534e

apps/channels/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/115/commits/ce2298bedab43b56ec9d9aeb6e4fb6a1a217534e

apps/channels/urls.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/115/commits/ce2298bedab43b56ec9d9aeb6e4fb6a1a217534e

apps/channels/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/115/commits/ce2298bedab43b56ec9d9aeb6e4fb6a1a217534e

apps/channels/forms.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/115/commits/20913cded1b4f1f385497242a7365ab47f6b602e

snopoke:
@SmittieC ping me when this is ready for review again.

SmittieC:
@snopoke This should be ready for final review now

snopoke (approved):

PR #122, opened 2023-11-07
mark data migrations as `elidable=True`

snopoke:
Preparing to squash some migrations

### DIFF
--- a/apps/experiments/migrations/0050_bootstrap_voice_providers.py
+++ b/apps/experiments/migrations/0050_bootstrap_voice_providers.py
@@ -40,5 +40,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(bootstrap_voice_providers)
+        migrations.RunPython(bootstrap_voice_providers, elidable=True)
     ]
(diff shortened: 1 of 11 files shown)


SmittieC (approved):

apps/experiments/migrations/0050_bootstrap_voice_providers.py
```diff
@@ -40,5 +40,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(bootstrap_voice_providers)
+        migrations.RunPython(bootstrap_voice_providers, elidable=True)
```
SmittieC, line 43: Hmm, should this be marked as `elidable`?

snopoke: ah, no. Thanks

snopoke: Actually this one is fine since it's a one-off migration that's not required on new installs.

SmittieC: My bad, you're right! I think I confused this with the data migrations that loaded the voices.

PR #124, opened 2023-11-07
initial permissions framework

### LINKED ISSUE #123: Build permissions backend

snopoke:
Resolves #123 

* add a link between team membership and Django auth groups
* custom auth backend to get permissions from current team membership
* initial set of groups with permissions

This does not update any UI's or do any permission checking


apps/teams/backends.py
```diff
@@ -0,0 +1,128 @@
...
+            AppPermSetDef("channels", ALL),
+        ],
+    ),
+    GroupDef(
+        "Chat Viewer",
+        [
+            AppPermSetDef("chat", [VIEW]),
+        ],
+    ),
+]
```
snopoke, line 112: I put these in as initial groups, very open to other suggestions

apps/utils/factories/team.py
```diff
@@ -19,6 +19,13 @@ class Meta:
     team = factory.SubFactory(TeamFactory)
     role = "admin"
 
+    @factory.post_generation
```
SmittieC, line 22: These factories are only used in tests right?

apps/teams/tests/test_backends.py
```diff
@@ -0,0 +1,67 @@
...
+    return group
+
+
+@pytest.fixture(autouse=True)
+def use_team_backend(settings):
+    settings.AUTHENTICATION_BACKENDS = ["apps.teams.backends.TeamBackend"]
+
+
+@pytest.mark.django_db()
+def test_team_backend_no_current_team(group1, group2):
```
SmittieC, line 42: The usge of the context variable made this test a bit unintuitive. My first thought was that the user should have had permissions, since
1. the user did have team membership, and
2. the membership did belong to groups with permissions.

Keeping in mind that the context variable was not set, this test makes sense. Might be good to add a comment about that

snopoke: 362252fa0543eb9177242fcf450882f6f475c2f2

apps/utils/factories/team.py
SmittieC: Thoughts on having an `apps/utils/tests/factories` file and keep the test related utils in there?

snopoke: Could do. I don't think the location matters too much.

apps/teams/backends.py
SmittieC: These look fine to me. Is there a specific need to split the "Chat Viewer" and "Experiment Admin" at the moment?

apps/utils/factories/team.py
SmittieC: I usually just like to keep test related things in some sort of test/ folder, but it's fine. Agreed that it doesn't matter too much

SmittieC (approved):
nice:+1:

apps/teams/backends.py
snopoke: Not at the moment but I do think it's useful in general.

Issue #127, opened 2023-11-08
Better validation for experiments


snopoke:
Use langchain models to parse prompts for input variables and make sure they are present. 

e.g: https://dimagi.sentry.io/issues/4613304606/events/68611602d0b94743bb568779577dc01c/

This error is because the prompt has `{source_material}` input defined but there is no input material linked to the experiment.

PR #128, opened 2023-11-09
[proposal] Set source material to empty string if not present

SmittieC:
This is a proposed solution to #127. If this is fine, I'll add a test. I think it makes sense to do this, because if the user specifies in the prompt that there is source material, but doesn't link any, then it should be empty. I'm open to other suggestions though

### DIFF
--- a/apps/chat/conversation.py
+++ b/apps/chat/conversation.py
@@ -34,7 +34,9 @@ class Conversation:
         UTC = pytz.timezone("UTC")
         current_datetime = datetime.now().astimezone(UTC)
         prompt_to_use = SystemMessagePromptTemplate.from_template(prompt_str)
-        if source_material:
+        if "source_material" in prompt_to_use.input_variables:
+            if not source_material:
+                source_material = ""
             try:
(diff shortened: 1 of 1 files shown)


apps/chat/conversation.py
```diff
@@ -52,6 +54,7 @@ def __init__(
                 ],
             )
         else:
+            print(f"using prompt_to_use: {prompt_to_use}\n\n\n---------------\n\n\n")
```
snopoke, line 57: print...

snopoke (commented):
This seems OK though I still think we should at least warn the user when saving an experiment in this state.

apps/chat/conversation.py
SmittieC: haha whoops..

snopoke (approved):

SmittieC:
I decided to rather go with something like this, so we can tell the user they need to link source material to the experiment when the prompt expects source material.

image

I'll create the PR for this soon

SmittieC:
New approach: https://github.com/dimagi/open-chat-studio/pull/132

Issue #103, opened 2023-10-31
Move Twilio keys to per tenant


snopoke:
## Relevant settings
* `TWILIO_ACCOUNT_SID`
* `TWILIO_AUTH_TOKEN`

## Suggested Approach

Create a new `MessagingProvider` model and configuration. Twilio would be the only option currently

See https://github.com/dimagi/open-chat-studio/blob/main/apps/service_providers/README.md

SmittieC:
Closing, since the work is done and only needs deploying

PR #130, opened 2023-11-09
Messaging Service Provider

SmittieC:
For #103 
- Added a MessagingService and Twilio as a ServiceType
- ChannelForm will now optionally have a `messaging provider` input that only shows when there are providers supporting the platform in question.

Peek 2023-11-13 17-10


apps/channels/tests/test_forms.py
```diff
@@ -0,0 +1,26 @@
...
+    message_provider = MessagingProviderFactory(type=MessagingProviderType("twilio"), team=team)
+    MessagingProviderFactory(type=MessagingProviderType("twilio"))
+
+    form = ChannelForm(initial={"platform": ChannelPlatform(platform)}, team=team)
+    widget = form.fields["messaging_provider"].widget
+    form_queryset = form.fields["messaging_provider"].queryset
+    if platform == "whatsapp":
+        assert isinstance(widget, Select)
+    else:
+        assert isinstance(widget, HiddenInput)
```
snopoke, line 23: instead of having an if statement here you could pass the expected type in to the test function:

```python
@pytest.mark.parametrize("platform,widget", [
    ("whatsapp", Select),
    ("telegram", HiddenInput),
])
def test_channel_form_reveals_provider_types(team, platform, widget):
```

apps/service_providers/messaging_service.py
```diff
@@ -0,0 +1,50 @@
...
+import requests
+from twilio.rest import Client
+
+from apps.channels import audio
+from apps.channels.models import ChannelPlatform
+
+
+class MessagingService(pydantic.BaseModel):
+    _type: str
+    _supported_platforms: list
```
snopoke, line 13: General question. What do you think about moving away from pydantic? I think we could use dataclasses as easily.

apps/service_providers/tests/test_messaging_providers.py
```diff
@@ -0,0 +1,70 @@
...
+    assert form.is_valid()
+    form.cleaned_data.pop(config_key)
+    _test_messaging_provider_error(MessagingProviderType.twilio, data=form.cleaned_data)
+
+
+@pytest.mark.parametrize(
+    "platform,expected_provider_types",
+    [
+        ("whatsapp", ["twilio"]),
+        ("telegram", []),
```
snopoke, line 44: Is the plan to add a provider for telegram later?

snopoke (approved):

apps/service_providers/messaging_service.py
SmittieC: I have no reservations. We're using pydantic to define the tools' schemata for agents (in the `schemas.py` file). If that can work with dataclasses, I don't see why not. I suspect it wouldn't be an issue

apps/service_providers/tests/test_messaging_providers.py
SmittieC: Oh, no. This was just to test that no providers should be returned for platforms that doesn't have any providers

apps/channels/tests/test_forms.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/130/commits/c395dd250a2e7941fef29fbd3869faa253cff18e

snopoke (approved):

SmittieC:
Two things:
1. When this is deployed, I need to (quickly) create the messaging provider and link that to the existing experiment channel. I checked again, there's only 1 whatsapp experiment channel, so doing it manually should be fine.
2. I'll need to merge + deploy this somewhere where it wouldn't interfere with normal operations.

PR #131, opened 2023-11-09
partial groups rollout

snopoke:
* update team & use views to show groups instead of legacy roles
* bootstrap groups for all existing team memberships
* some permission changes

Before continuing I'm going to cleanup some code that we aren't using e.g. team APIs


apps/teams/helpers.py
```diff
@@ -57,6 +56,5 @@ def create_default_team_for_user(user: CustomUser, team_name: str = None):
     team_name = team_name or get_default_team_name_for_user(user)
     slug = get_next_unique_team_slug(team_name)
     team = Team.objects.create(name=team_name, slug=slug)
-    team.members.add(user, through_defaults={"role": roles.ROLE_ADMIN})
-    team.save()
+    make_user_team_owner(team, user)
```
SmittieC, line 59: Not a biggy, but what are your thoughts on making the `make_user_team_owner` and creating the team more coupled? Maybe calling `make_user_team_owner` inside the save function after the super().save call or something. Up to you

apps/teams/models.py
```diff
@@ -61,11 +54,41 @@ class Membership(BaseModel):
...
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    role = models.CharField(max_length=100, choices=roles.ROLE_CHOICES)
+
     def __str__(self):
         return f"{self.user}: {self.team}"
 
-    def is_admin(self) -> bool:
-        return self.role == roles.ROLE_ADMIN
+    def is_team_admin(self) -> bool:
+        return self.has_perms(["teams.change_team", "teams.delete_team"])
```
SmittieC, line 91: Can't we ssimply check if the membership belongs to the `TEAM_ADMIN_GROUP` group?

apps/teams/tests/test_member_management.py
```diff
@@ -134,9 +139,13 @@ def test_only_admin_cant_leave(self):
...
-        self._change_role(c, self.admin_membership2, roles.ROLE_MEMBER)
+        self._change_role(c, self.admin_membership2, self.member_groups)
 
         # confirm it doesn't work
         response = self._remove_member(c, self.admin_membership)
         self.assertNotEqual(200, response.status_code)
         self.assertTrue(Membership.objects.filter(pk=self.admin_membership.pk).exists())
+
+    def _check_groups(self, membership, expected_group_ids):
+        membership = Membership.objects.get(id=membership.id)
```
SmittieC, line 150: just out of curiosity, no need to really change anything: is there a reason why you don't just use the membership object from the params directly instead of doing `Membership.objects.get(id=membership.id)`?

apps/teams/helpers.py
snopoke: I kept it separate for the sake of tests since there are places where we create 2 'owners' for the same team.

I could create another utility which is used here that does both.

apps/teams/models.py
snopoke: we could use that as a proxy but I thought it would be better to always check the permissions so that if we ever change the groups it still works.

Also this is how I found the issue with creating groups in the `post_migrate` signal.

apps/teams/tests/test_member_management.py
snopoke: I need to reload it from the DB to clear the M2M cache.

I'll add a comment.

apps/teams/helpers.py
SmittieC: Fair, thanks for clarifying

SmittieC (approved):
:rocket:

PR #132, opened 2023-11-13
Validate source material presence

SmittieC:
This is a proposed solution to https://github.com/dimagi/open-chat-studio/issues/127 and a replacement for [this approach](https://github.com/dimagi/open-chat-studio/pull/128). This "new" way notifies users when they try to update an experiement that requires source material (in the prompt), but forgot to add any. An error message will show when they try to save:

Screenshot from 2023-11-13 11-18-44

### DIFF
--- a/apps/experiments/views/experiment.py
+++ b/apps/experiments/views/experiment.py
@@ -11,7 +11,7 @@ from django.contrib.auth.decorators import user_passes_test
 from django.contrib.postgres.search import SearchVector
 from django.core.exceptions import ValidationError
 from django.http import Http404, HttpResponse, HttpResponseRedirect
-from django.shortcuts import get_object_or_404, redirect
+from django.shortcuts import get_object_or_404, redirect, render
 from django.template.response import TemplateResponse
 from django.urls import reverse
 from django.utils import timezone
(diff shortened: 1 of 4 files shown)


snopoke (approved):

apps/experiments/views/experiment.py
```diff
@@ -64,7 +64,23 @@ def get_queryset(self):
...
+class ExperimentViewMixin:
+    def get_form(self):
+        form = super().get_form()
+        _apply_related_model_querysets(self.request.team, form)
+        _apply_voice_provider_alpine_attrs(form)
+        return form
+
+    def form_valid(self, form):
+        form.instance.team = self.request.team
+        form.instance.owner = self.request.user
```
snopoke, line 76: Not a blocker but this will change the 'owner' to whichever user is updating the experiment.

At some point we should probably just create a custom form class for experiments.

SmittieC: Ah good catch. For the sake of running the "same code" as before, I made this update: https://github.com/dimagi/open-chat-studio/pull/132/commits/8a12501d57a0220ff25a494abf81581f995075a5

PR #149, opened 2023-11-24
Notify user when they try to format the user input in the prompt string

SmittieC:
This is in response to [this sentry error](https://dimagi.sentry.io/issues/4664550002/?alert_rule_id=14063552&alert_timestamp=1700802704064&alert_type=email&environment=production&notification_uuid=eaaaec44-1a6d-49fc-9887-d96b70d4117c&project=4505001320316928&referrer=alert_email). The user tried to format the user input in the prompt string, instead of using the input formatter.

This change shows an error message when the user includes the `{input}` key in the prompt string.

image

While this change is preventative, it's still not very clear how to use the `{input}` key to format the user input I think. Since we're using `{source_material}` in the prompt string, I can imagine it being a bit confusing to have to use the `{input}` key in the input formatter when the user doesn't have any context on how these keys/fields are being used.

We should rethink the help text or add a short section at the top giving some context or an example? Not 100% sure what would be best. Suggestions welcome cc @bderenzi @snopoke

### DIFF
--- a/apps/experiments/views/prompt.py
+++ b/apps/experiments/views/prompt.py
@@ -14,6 +14,7 @@ from django.utils import timezone
 from django.views.decorators.http import require_POST
 from django.views.generic import CreateView, UpdateView
 from django_tables2 import SingleTableView
+from langchain.prompts import PromptTemplate
 
 from apps.experiments.helpers import get_real_user_or_none
 from apps.experiments.models import Prompt, PromptBuilderHistory, SourceMaterial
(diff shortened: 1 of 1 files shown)


apps/experiments/views/prompt.py
```diff
@@ -54,7 +55,18 @@ def get_queryset(self):
         return query_set
 
 
-class CreatePrompt(CreateView):
+class PromptViewMixin:
+    def form_valid(self, form):
+        if "{input}" in form.data["prompt"]:
```
snopoke, line 60: What about using the langchain PromptTemplate class to get the input variables. Seems like it would be more robust.

```
input_variables = PromptTemplate.from_template(form.data["prompt"]).input_variables
if input_variables - allowed_variables:
    ...
```
```

SmittieC: https://github.com/dimagi/open-chat-studio/pull/149/commits/19c753af562df0b16feebb61bcd75479df3b30dd

image

snopoke (approved):

PR #151, opened 2023-11-28
Bot to present the consent form

### LINKED ISSUE #18: External channels to support asking for user consent in experiment runs
In a typical field study / trial, users are provided with a link or a publicly accessible link. When they follow this link, they are initially required to review a Google Form and provide their consent. Afterward, on the original link, they are expected to check a box indicating their consent. Only when they have confirmed their consent or indicated that they have read the information are they allowed to proceed to the chat. We are currently looking to implement a comparable workflow in external channels (Telegram / Whatsapp etc), where the bot will inquire if they have read the information and granted consent to proceed.

SmittieC:
Resolves #18 

This adds a new field on the `ConsentForm` table that allows experiment admins to tell users which words/phrases they should use in order to give consent. Example (the default consent):

image

For external channels, the chat flow will be as follows:
- The user initiates the chat by sending a message
- The bot sends the consent form (along with the newly added text) to the user
- Unless the user sends a word / phrase that is specified as a confirmation word/phrase, the bot will not say anything (we _can_ keep on responding with the consent form, but this will bloat the conversation history. Unless we make this message of type **system**)
- When the user gives consent, one of two things will happen
- 1. If the experiment has a seed message, the bot will send that
- 2. If the experiment does not have a seed message, the bot will keep quiet.

### Open for discussion:

> If the experiment does not have a seed message, the bot will keep quiet.

One thing I'm in favour of is to maybe let the bot respond to the first message that the user sent. For telegram + whatsapp I'm pretty sure we can specify which message its responding to, so it will look quite natural. Thoughts?


apps/experiments/migrations/0055_consentform_conversational_consent.py
```diff
@@ -0,0 +1,20 @@
...
+    dependencies = [
+        ("experiments", "0054_consentform_capture_identifier_and_more"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="consentform",
+            name="conversational_consent",
+            field=models.CharField(
+                default="Use any of the following to accept: {yes,accept,I accept}",
```
SmittieC, line 16: I'm not very happy with the approach I took, but couldn't think of a better one. Ultimately, we need a way to tell users "respond with the following word/phrase(s) to accept" in multiple languages, and only when the consent form is given through conversation. Other suggestions are super welcome

snopoke: With telegram you can create [buttons](https://core.telegram.org/api/bots/buttons). Whatsapp too but they need to go through an [approval process](https://www.freshworks.com/live-chat-software/customer-engagement/whatsapp-interactive-buttons-blog/).

Do we even know the user's language by this point? If we do we could use the LLM to produce the text with a prompt like:

> Write a short message to determine if a user agrees to continue they must type "1". Write in French.

snopoke: Also thinking that it will be simpler if we always make the user type "1" to agree.

SmittieC: > Do we even know the user's language by this point? If we do we could use the LLM to produce the text with a prompt like

Depends on the message they used to initiate the conversation with, so I wouldn't really rely on that. For example, "hi" is quite language agnostic.

What we _can_ do, come to think of it, is to use the same language in which the consent form is written, since we're presenting users with the consent form in anycase :thinking:

SmittieC: I think let's try to generate "Type '1' if you agree" in the language that the consent form is written in and see how that goes. If that doesn't work, we can expose a new field to allow bot creators to manually specify this sentence.

apps/chat/channels.py
```diff
@@ -185,28 +187,23 @@ def _chat_initiated(self):
...
+        ]
 
     def _user_gave_consent(self) -> bool:
-        """Match the user's input to the configured "acceptance" keywords"""
-        consent_text = self.experiment.consent_form.conversational_consent
-        pattern = r"\{(.*?)\}"
-        matches = re.findall(pattern, consent_text)
-        consent_keywords = matches[0].split(",")
-        return self.message_text in [word.lstrip() for word in consent_keywords]
+        return self.message_text == USER_CONSENT_TEXT
```
snopoke, line 206: should probably do `self.message_text.strip()`

snopoke (approved):
minor change but otherwise looks good

apps/chat/channels.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/151/commits/e2af2e254ee356789eac5782538076d6709185fd

SmittieC:
I'll merge this on Monday when I'm back. Don't want to accidentally let this go into prod and causing issues that someone else has to fix

PR #152, opened 2023-11-30
Data analysis pipelines

snopoke:
Data analysis pipelines.

See https://github.com/dimagi/open-chat-studio/blob/sk/pipeline/apps/analysis/README.md for some basic docs and overview.

ocs_analysis


apps/analysis/steps/loaders.py
```diff
@@ -0,0 +1,82 @@
+import json
+from functools import cached_property
+
+import pandas as pd
```
SmittieC, line 4: the `pandas` requirement is needed

apps/analysis/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+    llm_service = run.analysis.llm_provider.get_llm_service()
+    pipeline_context = PipelineContext(llm_service, logger=Logger(log_stream), params=run.params)
+
+    try:
+        yield pipeline_context
+        run.status = RunStatus.SUCCESS
+    except Exception as e:
+        run.status = RunStatus.ERROR
+        run.error = repr(e)
+        log.exception("Error running analysis")
```
SmittieC, line 32: Suggestion: What if we `log.exception(e)` (so we see the exception details in Sentry) and rather set `run.error` to "Error running analysis"? I get that we'd want to tell the user what happened, but 1. would they understand exception messages? and 2. exposing exceptions doesn't sit quit right....or did you have some specific usecase in mind?

snopoke: `log.exception` will already capture the full exception details, no need to pass in the exception instance.

> I get that we'd want to tell the user what happened, but 1. would they understand exception messages? and 2. exposing exceptions doesn't sit quit right....or did you have some specific usecase in mind?

Absolutely. This is a temporary setup for now while we're still building.

apps/analysis/models.py
```diff
@@ -0,0 +1,133 @@
...
+        if seconds is None:
+            return ""
+
+        if seconds < 60:
+            return str(seconds) + "s"
+        return timedelta(seconds=math.ceil(seconds))
+
+
+class RunGroup(BaseRun):
+    team = models.ForeignKey(Team, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
snopoke: Yea, it didn't seem necessary to have team in `AnalysisRun`

apps/analysis/pipelines.py
```diff
@@ -0,0 +1,80 @@
...
+from .steps.filters import TimeseriesFilter
+from .steps.processors import AssistantStep, LlmCompletionStep
+from .steps.splitters import TimeseriesSplitter
+
+LLM_PIPE = "llm"
+
+TEXT_DATA_PIPE = "text_data"
+
+SOURCE_PIPELINES = {
+    TEXT_DATA_PIPE: Pipeline(
```
SmittieC, line 14: nit: the other keys in this dict aren't constants like this one

apps/analysis/pipelines.py
```diff
@@ -0,0 +1,80 @@
...
+            TimeseriesSplitter(),
+        ],
+        "Split WhatsApp Data",
+        "Load whatsApp data from a file and split it into chunks by date.",
+    ),
+}
+
+
+PIPELINES = {
+    LLM_PIPE: Pipeline(
```
SmittieC, line 43: > nit: the other keys in this dict aren't constants like this one

Same here. Was there a reason for this?

apps/analysis/log.py
```diff
@@ -0,0 +1,82 @@
+import dataclasses
```
SmittieC, line 1: This looks a lot like python's logger. Was it easier / quiker to implement this than to try to piggy-back on the logger's functionality?

apps/analysis/models.py
```diff
@@ -0,0 +1,133 @@
...
+class ResourceMetadata(pydantic.BaseModel):
+    type: str
+    format: str
+    data_schema: dict
+
+
+class Resource(BaseTeamModel):
+    name = models.CharField(max_length=255)
+    type = models.CharField(max_length=128, choices=ResourceType.choices)
+    file = models.FileField()
```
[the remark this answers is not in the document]
snopoke: https://github.com/dimagi/open-chat-studio/pull/160

apps/analysis/pipelines.py
snopoke: These were the ones that I accessed in tests so I made them contstants. I'm going to refactor this code later anyway.

apps/analysis/log.py
snopoke: Originally this was going more but can probably remove it now.

apps/analysis/serializers.py
```diff
@@ -0,0 +1,106 @@
...
+    serializer = get_serializer(data)
+    metadata = serializer.get_metadata(data)
+    time_suffix = timezone.now().replace(tzinfo=None).isoformat(timespec="milliseconds")
+    resource = Resource(
+        team=team,
+        name=f"{name} {time_suffix}",
+        type=metadata.format,
+        metadata=metadata.model_dump(),
+    )
+    with tempfile.TemporaryFile(mode="w+") as file:
```
SmittieC, line 22: Should this not also be saved on S3?

SmittieC: Right! For some reason I thought that only saved a reference to the file. Which is technically true, but I forgot that it actually saves the file as well. Cool cool

templates/analysis/components/group_progress_inner.html
```diff
@@ -2,7 +2,11 @@
 <span class="loading loading-dots loading-md"></span>
 {% endif %}
 {% for run in runs reversed %}
-  <p class="my-2 font-medium">Step {{ forloop.revcounter }} ({{ run.get_status_display }})</p>
+  {% if forloop.first %}
+  <p class="my-2 font-medium">Source Pipeline ({{ run.get_status_display }})</p>
+  {% else %}
+  <p class="my-2 font-medium">Analysis Pipeline {{ forloop.revcounter-0 }} ({{ run.get_status_display }})</p>
```
SmittieC, line 8: is `forloop.revcounter-0` correct? I get an error where python's not able to parse the -0 part. Did you mean to do `forloop.revcounter0` perhaps?

apps/analysis/steps/parsers.py
```diff
@@ -0,0 +1,45 @@
...
+
+        01/01/2021, 00:00 - System Message
+        01/01/2021, 00:00 - Sender: Message which can
+        go across multiple lines
+    """
+
+    input_type = str
+    output_type = pd.DataFrame
+
+    def run(self, params: Params, data: str) -> tuple[pd.DataFrame, dict]:
```
SmittieC, line 23: Have you had a look at the [whatstk](https://pypi.org/project/whatstk/) library?

snopoke: Yes I did look at that but it seemed way overkill for just the parsing. They also take quite a complex approach to parsing which didn't seem necessary (unless there are multiple chat formats - I'm assuming there's only one).

SmittieC (approved):
Looks good to me :+1:  I admittedly didn't go through the processors.py + splitters.py files _too much_. The core.py file looks pretty solid. In fact, everything looks pretty solid for a feature introduction..nice work! :guitar: The fact that there are so many tests gives me confidence, thanks. I guess we can release it "to the masses" and see what happens :smile:

Issue #144, opened 2023-11-22
Audting spec


snopoke:
Spec out what audting we should be doing and how to expose the data to users.

Look at the following projects:
* https://github.com/jazzband/django-auditlog
* https://github.com/dimagi/django-field-audit

SmittieC:
The [spec](https://docs.google.com/document/d/11DS5j0jFjSBEvJxm35BRXZwajyJuivlLK62NqGlctsk/edit)

SmittieC:
### Models to audit
- [ ] CustomUser
- [ ] Team
- [ ] Membership
- [ ] Experiment
- [ ] SourceMaterial
- [ ] SafetyLayer
- [ ] NoActivityMessageConfig
- [ ] ConsentForm
- [ ] LlmProvider
- [ ] VoiceProvider
- [ ] MessagingProvider
- [ ] ChannelPlatform

PR #167, opened 2023-12-12
Audit logs

SmittieC:
See #144. On the issue I added a list of models that needs auditing. This PR includes those that are ticked. We can ship this so long if you think it's fine @snopoke, then in a subsequent PR I'll add the others. I'm having some issue with `Member` table not creating logs for `group` updates that I want to figure out before continuing.

You'll see I created cusom object managers so that we can audit [special methods](https://github.com/dimagi/django-field-audit#auditing-special-queryset-writes).

### DIFF
--- a/apps/experiments/model_audit_fields.py
+++ b/apps/experiments/model_audit_fields.py
@@ -23,3 +23,4 @@ EXPERIMENT_FIELDS = [
 ]
 
 SOURCE_MATERIAL_FIELDS = ["owner", "topic", "description", "material", "team"]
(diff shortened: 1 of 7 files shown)


apps/experiments/model_audit_fields.py
```diff
@@ -0,0 +1,35 @@
+PROMPT_FIELDS = ["owner", "name", "description", "prompt", "input_formatter", "team"]
```
SmittieC, line 1: I don't like this. Will think of other better ways

snopoke (approved):

PR #169, opened 2023-12-12
split static and dynamic config for analysis pipelines

snopoke:
This allows steps to provide separate forms for static configuration and dynamic configuration. Any fields in the dynamic config with the same name will be populated with values from the static config which allows users to override defaults at runtime.

This also refactors the table actions to make them more configurable in terms of enabled state and whether they should be displayed or not.


templates/analysis/analysis_form.html
```diff
@@ -1,21 +1,48 @@
...
+        </a>
+      </div>
+    </div>
+  {% endif %}
+{% endblock title_bar_end %}
+{% block pre_form %}
+  {% if analysis.needs_configuration %}
+    <div class="alert mt-2">
+      <i class="fa-solid fa-circle-info"></i>
+      <span>This analysis requires static <a class="link" href="{% url "analysis:configure" request.team.slug analysis.id %}">configuration</a></span>
```
SmittieC, line 25: Just wondering, would it be obvious to the user what "static config" refers to?

apps/analysis/views.py
```diff
@@ -45,6 +48,38 @@ def analysis_details(request, team_slug: str, pk: int):
     )
 
 
+@login_and_team_required
+@permission_required("analysis.change_analysis", raise_exception=True)
+def analysis_configure(request, team_slug: str, pk: int):
```
SmittieC, line 53: No need to change this, but using a view class would make this a bit cleaner imo

SmittieC (approved):
Looks good to me. Just one question and one comment that's up to you if you want to address it

apps/analysis/views.py
snopoke: Can you say more? How would a view class make it clearer?

templates/analysis/analysis_form.html
snopoke: Maybe not but it's what I came up with - happy to change it if you have suggestions. Making this more of a wizard with some help text would improve things too.

snopoke:
I'm going to merge. Will make any changes from feedback in a future PR.

snopoke:
I'm going to merge. Will make any changes from feedback in a future PR.

snopoke:
I'm going to merge. Will make any changes from feedback in a future PR.

snopoke:
I'm going to merge. Will make any changes from feedback in a future PR.

PR #172, opened 2023-12-13
refactor resource handling in preparation for parallel runs and agents

snopoke:
This refactors how resources are created during pipeline runs. Each step can now manage their own resources and resources are included in the result context.

This means that later steps have access to resources created in previous steps and will allow the assistant step to create resources from files generated by the assistant.

This also paves the way for parallel runs which require persistence between tasks.

🐡 (Best reviewed by commit)


SmittieC (approved):
Looks fine to me

PR #190, opened 2024-01-11
Email users on safety layer violation

SmittieC:
Email addresses to notify is specified on the experiment:

image


apps/chat/bots.py
```diff
@@ -119,6 +125,8 @@ def _get_response(self, input_str: str):
         # human safety layers
         for safety_bot in self.safety_bots:
             if safety_bot.filter_human_messages() and not safety_bot.is_safe(input_str):
+                if self.session_id:
```
snopoke, line 128: When would there be no session ID?

snopoke (commented):
One question but otherwise looks good

snopoke (approved):

apps/chat/bots.py
SmittieC: When using the prompt builder. In the `get_prompt_builder_response_task` task we don't give the `TopicBot` a session_id

SmittieC: Added https://github.com/dimagi/open-chat-studio/pull/190/commits/ec92e5880a825585cba03427800dd1e1b4b9ceef

PR #194, opened 2024-01-17
Present user with survey link

SmittieC:
The bot will now, if there is a `pre_survey` on the experiment, present the user with the link to that and ask them to confirm that they filled it in. I'm not 100% sure how we'll do the `post_survey`, since we wouldn't know when the conversation is done. One way could be to add another action (like the /reset one), that the user must take to indicate that they are done with the conversation.

image


apps/experiments/migrations/0058_survey_confirmation_text.py
```diff
@@ -0,0 +1,20 @@
...
+    dependencies = [
+        ("experiments", "0057_experiment_safety_violation_notification_emails"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="survey",
+            name="confirmation_text",
+            field=models.TextField(
+                default="Before starting the experiment, we ask that you complete a short survey. Please click on thesurvey link, fill it out, and, when you have finished, respond with '1' to let us know that you've completed it. Survey link: {survey_link}",
```
snopoke, line 16:
```suggestion
                default="Before starting the experiment, we ask that you complete a short survey. Please click on the survey link, fill it out, and, when you have finished, respond with '1' to let us know that you've completed it. Survey link: {survey_link}",
```

snopoke (approved):

apps/experiments/migrations/0058_survey_confirmation_text.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/194/commits/9fdf7cfca1f92984385af6036063963144f49c0b

PR #200, opened 2024-01-19
use llm models from selected llm provider in prompt builder

### LINKED ISSUE #163: Prompt builder support for available models from the LLM Provider
Write now the prompt builder is hardcoded with gpt-4 and 3.5 turbo. We should have it instead pull from the available models for a LLM provider for the current team. 

When we do that, we should probably move it back to the starting tab since it'll be important to show. Before the answer was always `gpt-4`, but probably the correct default now would be gpt-4 turbo. and there's at least one project that I think is going to try to use Anthropic/claude2.1.

(note the prompt builder is probably due for a larger overhaul, but I think we can put it off a bit longer)

snopoke:
Resolves #163


SmittieC (approved):
nice

PR #206, opened 2024-01-24
compress history when over token limit

### LINKED ISSUE #141: History limits at the experiment level
GPT-4 turbo has a context size of 128k. If a conversation went on long enough unchecked it would cost $1280 at current pricing just to read the input tokens.

I think we should limit context size, configured at an experiment level, to prevent runaway costs. We could choose to use LangChain's history compression or not. I'd vote for a simple thing for now because this is an edge case we just want to protect against. 

OpenAI's Assistant API is stateful and does something to compress history in the background. It's not clear to me exactly how this works (I couldn't find anything in the docs, but didn't dig too hard). Despite that, this may be appropriate for longer deployments so we may not want to invest too much effort into how we compress history just yet.

Note: we've hit the 8k limit numerous times operating in Chichewa, which is a token-dense language. If the gpt-4 limit of 8k hadn't existed those conversations could have continued in an increasingly expensive manner.

snopoke:
Allows specifying the token limit in the experiment configuration. When this limit is reached the message history will be compressed. The current compression method will keep the last 10 messages and replace the rest with a summary.

This includes a migration which will change the max token limit for all existing experiments to 8192 tokens.

Resolves #141


apps/chat/models.py
```diff
@@ -64,9 +77,18 @@ def created_at_datetime(self):
         return quote(self.created_at.isoformat())
 
     def to_langchain_dict(self) -> dict:
+        return self._get_langchain_dict(self.content, self.message_type)
+
+    def summary_to_langchain_dict(self) -> dict:
+        return self._get_langchain_dict(self.summary, ChatMessageType.SYSTEM)
```
SmittieC, line 83: since this is calling the `_get_langchain_dict`, is there any reason to keep `_get_langchain_dict` around?

apps/experiments/models.py
```diff
@@ -356,6 +355,11 @@ class Experiment(BaseTeamModel):
         null=True,
         blank=True,
     )
+    max_token_limit = models.IntegerField(
+        default=0,
```
SmittieC, line 359: this default also needs to be updated to "8192"

apps/chat/bots.py
```diff
@@ -163,6 +165,42 @@ def _save_message_to_history(self, message: str, type_: ChatMessageType):
...
+            log.exception("Unable to compress history")
+            return self.chat.get_langchain_messages_until_summary()
+
+
+def compress_chat_history(chat: Chat, llm: BaseLanguageModel, max_token_limit: int, keep_history_len: int = 10):
+    """Compresses the chat history to be less than max_token_limit tokens long. This will summarize the history
+    if necessary and save the summary to the DB.
+    """
+    history = chat.get_langchain_messages_until_summary()
+    if max_token_limit <= 0 or not history:
```
SmittieC, line 182: Should we not rather prevent the user from entering a token limit smaller than 0?

SmittieC (commented):
Nice. Just a few comments

apps/chat/bots.py
```diff
@@ -163,7 +166,12 @@ def _save_message_to_history(self, message: str, type_: ChatMessageType):
             )
 
     def _get_optimized_history(self):
-        return compress_chat_history(self.chat, self.llm, self.max_token_limit)
+        try:
+            return compress_chat_history(self.chat, self.llm, self.max_token_limit)
+        except (NameError, ImportError, ValueError):
```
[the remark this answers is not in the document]
snopoke: Different LLMs use different methods. For now I think we are covered by tiktoken but I added this here for safety. We'll still get notified by the logging but it won't crash the chat.

apps/experiments/models.py
snopoke: I kept this as 0 so that new experiments can set it to what they want. I would like set the default on the UI based on the model being used but that was too complicated for this PR.

apps/chat/bots.py
snopoke: I guess I could make the model field a PositiveSmallIntegerField but it's not really a big deal. The `<= 0` was just me being defensive.

snopoke: 883d60f

apps/chat/bots.py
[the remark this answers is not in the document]
SmittieC: Got it, thanks

apps/experiments/models.py
SmittieC: Wouldn't the discrepancy between this default value and the one in the migration cause a new migration to be generated that updates the default value then?

apps/chat/bots.py
SmittieC: :+1:

apps/experiments/models.py
snopoke: You are correct. For some reason I thought you could do this 🤷🏻‍♂️ 1f1a369

SmittieC (approved):
Exciting!

PR #215, opened 2024-01-26
More auditing

SmittieC:
Addresses #144

For a smooth experience, go through the PR commit-wise.

All went well, except for the `Membership`model. I had to update the tests, since `team.members.add(...)` does not play well with the auditing library, which is a suboptimality. For now, simply creating a `Membership` object will suffice.

I ran into another issue when I tried to audit the `CustomUser` object. The audit library struggles with datetime serialization. I'll look into it. I have an idea where the issue occurs

### DIFF
--- a/apps/experiments/model_audit_fields.py
+++ b/apps/experiments/model_audit_fields.py
@@ -44,3 +44,5 @@ EXPERIMENT_CHANNEL_FIELDS = [
     "platform",
     "messaging_provider",
 ]
+
+NO_ACTIVITY_CONFIG_FIELDS = ["message_for_bot", "name", "max_pings", "ping_after"]
--- a/apps/experiments/model_audit_fields.py
+++ b/apps/experiments/model_audit_fields.py
@@ -33,3 +33,6 @@ CONSENT_FORM_FIELDS = [
     "confirmation_text",
     "team",
 ]
+
+TEAM_FIELDS = ["name", "slug", "members"]
+MEMBERSHIP_FIELDS = ["team", "user", "role"]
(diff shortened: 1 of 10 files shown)


apps/experiments/model_audit_fields.py
```diff
@@ -33,3 +33,6 @@
     "confirmation_text",
     "team",
 ]
+
+TEAM_FIELDS = ["name", "slug", "members"]
+MEMBERSHIP_FIELDS = ["team", "user", "role"]
```
snopoke, line 40: A bit odd to have these in the experiments app. Maybe move them to the `teams` app?

apps/experiments/model_audit_fields.py
```diff
@@ -49,3 +51,6 @@
 MESSAGING_PROVIDER_FIELDS = ["type", "name", "team"]
 VOICE_PROVIDER_FIELDS = ["type", "name", "team"]
 LLM_PROVIDER_FIELDS = ["team", "type", "name", "llm_models"]
+
+# The auditing library struggles with dates. Let's ignore them for now
+CUSTOM_USER_FIELDS = [f.attname for f in AbstractUser._meta.fields if f.attname not in ["last_login", "date_joined"]]
```
snopoke, line 56: Probably no need to audit these fields anyway.

snopoke (approved):

apps/experiments/model_audit_fields.py
SmittieC: Lol, fair

SmittieC: https://github.com/dimagi/open-chat-studio/pull/215/commits/6ed76de0e0efc6cd0806de2429245e535b91fa5d

PR #222, opened 2024-01-31
Fix for "unhashable type: 'LlmProviderTypes'" error

snopoke:


apps/service_providers/utils.py
```diff
@@ -65,7 +65,7 @@ def get_service_provider_config_form(provider: ServiceProvider, data=None, insta
     return TypeSelectForm(
         primary=_get_main_form(provider, data=data.copy() if data else None, instance=instance),
         secondary={
-            subtype: subtype.form_cls(data=data.copy() if data else None, initial=initial_config)
+            str(subtype): subtype.form_cls(data=data.copy() if data else None, initial=initial_config)
```
snopoke, line 68: This is the actual fix.

I also looked at other usages of the class and don't think there are any more places that will cause issues.

SmittieC (approved):
Nice

PR #223, opened 2024-01-31
Turn IO integration

### LINKED ISSUE #204: Integrate with Turn.io
Allow clients that have Whatsapp numbers registered with Turn.io to use them with OCS.

https://whatsapp.turn.io/docs/category/turn-api

SmittieC:
Resolves #204 

### Demo

Peek 2024-01-31 12-56

Things to note:
- Users have to manually set the wehbook URL, which is why we show them the url they should use
- Voice doesn't work at this moment. 

Left for me to do:
- Tests
- See if I can't get voice working. The [docs tells me](https://whatsapp.turn.io/docs/api/webhooks) that I should get a link to the media being sent, but I simply don't get it. I'll look a bit deeper and/or reach out to them about this
- A bit of cleanup


apps/channels/views.py
```diff
@@ -27,6 +27,20 @@ def new_whatsapp_message(request):
...
+@csrf_exempt
+def new_turn_message(request, experiment_id: uuid):
+    message_data = json.loads(request.body.decode("utf-8"))
+    if "statuses" in message_data:
+        # Ignore status updates
+        return HttpResponse()
+
+    try:
+        tasks.handle_turn_message.delay(experiment_id=experiment_id, message_data=message_data)
+    except Exception as e:
```
snopoke, line 39: Is the error handling necessary?

apps/channels/models.py
```diff
@@ -43,11 +43,24 @@ def form(self, team: Team):
...
             case self.TELEGRAM:
                 return forms.TelegramChannelForm(*args, **kwargs)
             case self.WHATSAPP:
+                if channel and channel.messaging_provider.type == MessagingProviderType.turnio:
+                    webhook_url = absolute_url(
+                        reverse("channels:new_turn_message", kwargs={"experiment_id": channel.experiment.public_id}),
+                        is_secure=True,
+                    )
+                    initial = kwargs.get("initial", {})
+                    initial.setdefault("verify_token", str(uuid.uuid4()))
```
snopoke, line 60: what is `verify_token`? I don't see it used anywhere else. Maybe just a copy paste?

apps/channels/models.py
```diff
@@ -43,11 +43,24 @@ def form(self, team: Team):
...
+                    initial.setdefault("webook_url", webhook_url)
+                    kwargs["initial"] = initial
```
snopoke, line 62: What about moving this to the form class itself? It seams reasonable to me to pass the channel to the form constructor for these forms.

You could then also reuse it in the view instead of having mostly the same code there.

snopoke (approved):

apps/channels/models.py
SmittieC: Yes, a copy and paste :sweat_smile:

proteusvacuum (approved):
LGTM! For incoming media messages, I remember also having issues receiving them through Turn, I don't remember if I ever found a solution.

For outgoing media messages, this wasn't ever implemented in `turn-python`,  as it wasn't needed in CommCareHQ. You might have to do that and push a new version.

apps/channels/views.py
SmittieC: No. It was used for testing. I'll remove

SmittieC:
> LGTM! For incoming media messages, I remember also having issues receiving them through Turn, I don't remember if I ever found a solution.
> 
> For outgoing media messages, this wasn't ever implemented in `turn-python`, as it wasn't needed in CommCareHQ. You might have to do that and push a new version.

Good to know, thanks! I think I'll reach out to someone and ask them what's going on.

apps/channels/tests/test_whatsapp_integration.py
```diff
@@ -37,17 +43,98 @@ def audio_message():
...
+
+
+class TestTwilio:
     def test_parse_text_message(self):
         incoming_message = TwilioMessages.text_message()
         whatsapp_message = WhatsappMessage.model_validate(json.loads(incoming_message))
-        self.assertEqual(whatsapp_message.chat_id, whatsapp_message.from_number)
-        self.assertEqual(whatsapp_message.content_type, MESSAGE_TYPES.TEXT)
-        self.assertEqual(whatsapp_message.media_url, None)
+        assert whatsapp_message.chat_id == whatsapp_message.from_number
```
SmittieC, line 95: this was for pytest

apps/experiments/views/experiment.py
```diff
@@ -291,7 +292,11 @@ def create_channel(request, team_slug: str, experiment_id: int):
         except ChannelAlreadyUtilizedException as exception:
             messages.error(request, exception.html_message)
             return redirect("experiments:single_experiment_home", team_slug, experiment_id)
-
+        if platform == ChannelPlatform.WHATSAPP:
+            webhook_url = absolute_url(
+                reverse("channels:new_turn_message", kwargs={"experiment_id": experiment.public_id}), is_secure=True
+            )
+            messages.info(request, f"Use the following URL when setting up the webhook in Turn.io: {webhook_url}")
```
snopoke, line 299: What do you think about putting this in the `extra_form`? Then the view code would become:

```python
if extra_form:
    message = extra_form.get_message(experiment)
    if message:
        messages.info(request, message)
```

snopoke (approved):

apps/experiments/views/experiment.py
SmittieC: like this? https://github.com/dimagi/open-chat-studio/pull/223/commits/19e6b04b03dd9d05556acf9759d536f4cb04e4ad

snopoke (approved):

PR #227, opened 2024-02-02
Make 'assistant experiments' work

### LINKED ISSUE #203: Integration Experiments with OpenAI Assistants v1
https://docs.google.com/document/d/1K_6LrmTf5RxHy_ZkaquTGEoQOTXPx8FWCuPlTf6vMSc/edit

* Basic support
* No files input / output

snopoke:
This connects up the OpenAI assistant API to experiments that have an assistant configured.

The PR includes quite a lot of refactoring so best reviewed by commit.

Resolves #203


apps/chat/channels.py
```diff
@@ -315,15 +314,18 @@ def _get_llm_response(self, text: str) -> str:
...
+        experiment_bot = TopicBot.from_experiment_session(self.experiment_session)
         answer = experiment_bot.process_input(message)
         self.experiment_session.no_activity_ping_count = 0
         self.experiment_session.save()
         return answer
 
     def _add_message_to_history(self, message: str, message_type: ChatMessageType):
-        topic_bot = get_bot_from_experiment(self.experiment, self.experiment_session.chat)
-        topic_bot._save_message_to_history(message, message_type)
+        ChatMessage.objects.create(
```
SmittieC, line 325: :+1:

apps/chat/bots.py
```diff
@@ -22,13 +17,23 @@ def create_conversation(
...
 ) -> Conversation:
     try:
-        return Conversation(
-            prompt_str=prompt_str,
-            source_material=source_material,
-            memory=ConversationBufferMemory(return_messages=True),
-            llm=llm,
-            experiment_session=experiment_session,
-        )
+        if experiment_session and experiment_session.experiment.assistant:
```
SmittieC, line 20: Would the `if experiment_session` still be needed after you split the prompt builder out? From what I could see, we're always creating the experiment session before sending the user message to the LLM, so I think we can safely assume that the session will exist here.

SmittieC (approved):
Love the refactors! Just one question, otherwise it looks good!

apps/chat/bots.py
snopoke: This function (`create_conversation`) is still used by the prompt builder task so we still need the check here for now.

But there are more refactors coming 😉

PR #228, opened 2024-02-02
add tabs to experiment form when assistants is enabled

snopoke:
This divides up the UI a bit to make it clearer what applies to the current configuration.

More changes are needed though since some fields are required which don't need to be e.g. prompt (not needed if an assistant is selected)
localhost_8000_a_dimagi_experiments_new_


SmittieC (approved):
Looks great!

PR #230, opened 2024-02-06
Add a pipeline step for loading application data

snopoke:
This adds a step to pipelines that can pull an app JSON from the commcare API. This is for a specific bot currently but it should also pave the way for more future integrations.

In particular this adds a way to store auth configurations which can be reused.


apps/analysis/tasks.py
```diff
@@ -89,7 +89,7 @@ def run_analysis(run_group_id: int):
     elif group.is_complete:
         return
 
-    with RunStatusContext(group):
+    with RunStatusContext(group, bubble_errors=False):
```
SmittieC, line 92: Why this change?

snopoke: This is the outermost context manager so if it bubbles errors it will cause the celery task to fail which isn't what we want.

SmittieC (approved):
Looks fine to me

PR #233, opened 2024-02-07
Prompt builder UI tweaks

### LINKED ISSUE #232: Prompt builder UI improvements
‘Tabs’ not displaying properly
Message input height is off
Collapse input formatting
Show source material
Able to expand prompt / source material to larger size, maybe ⅔ screen width

snopoke:
Resolves #232 

See individual commits.

image


bderenzi (approved):
nice! The input textareas have a whole different look now, is that right? Initially they were the invisible-until-you-hover-of-click look that OpenAI had. I don't think it's worth cycles to restore that if we have something else that works now.

bderenzi:
Awesome -- that looks great!

PR #242, opened 2024-02-09
remove prompt model and references

snopoke:
Part 2 of https://github.com/dimagi/open-chat-studio/pull/229


SmittieC (approved):
Looking good

PR #245, opened 2024-02-09
validate optional experiment fields

snopoke:


apps/experiments/views/experiment.py
```diff
@@ -143,8 +145,19 @@ def __init__(self, request, *args, **kwargs):
 
     def clean(self):
         cleaned_data = super().clean()
-        if not cleaned_data["prompt_text"] and not cleaned_data["assistant"]:
-            raise forms.ValidationError("Prompt text is required unless you select an OpenAI Assistant")
+        print(cleaned_data)
```
SmittieC, line 148: I presume this should not be in here?

SmittieC (approved):

PR #248, opened 2024-02-12
switch from isort+black to ruff

snopoke:
https://docs.astral.sh/ruff/

Why:
* Much better [performance](https://docs.astral.sh/ruff/)
* One tool instead of 2
* Many more [rules](https://docs.astral.sh/ruff/rules/) supported out of the box

See individual commits for specific format / lint fixes.

I've also added a note at the end of the REAMDE: https://github.com/dimagi/open-chat-studio/blob/6c6886f4db2a6754cc9ac55f12664bd469d3ba52/README.md#linting

### DIFF
--- a/apps/channels/tests/test_whatsapp_integration.py
+++ b/apps/channels/tests/test_whatsapp_integration.py
@@ -29,7 +29,7 @@ def turnio_whatsapp_channel(turn_io_provider):
 
 
 @pytest.fixture()
-def twilio_provider():
+def twilio_provider(db):
     return MessagingProviderFactory(
         name="twilio", type=MessagingProviderType.twilio, config={"auth_token": "123", "account_sid": "123"}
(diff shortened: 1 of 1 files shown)


apps/analysis/serializers.py
```diff
@@ -92,7 +91,7 @@ def get_metadata(self, data: Any) -> ResourceMetadata:
         )
 
     def get_summary(self, data: Any) -> str:
-        if type(data) == str:
+        if isinstance(data, str):
```
SmittieC, line 94: Did the linter do this?

SmittieC: Oh wait, [I see it did](https://docs.astral.sh/ruff/rules/type-comparison/). Nice

SmittieC (approved):
:+1:

PR #255, opened 2024-02-15
Handle unsupported message types

SmittieC:
Addresses [this issue](https://dimagi.sentry.io/share/issue/e35a652364334717bb13ae8f46160430/). We simply return when someone sends an unsupported message type. A better approach probably is to ask the bot to formulate a good response here.


snopoke (approved):

SmittieC:
[This change](https://github.com/dimagi/open-chat-studio/pull/261) nullifies this PR

PR #257, opened 2024-02-15
Support for OpenAI Assistant files

### LINKED ISSUE #256: Support file download for files created by assistants

### LINKED ISSUE #258: Support file upload when creating assistants

snopoke:
Resolves #258

assistant_files


apps/files/apps.py
```diff
@@ -0,0 +1,21 @@
...
+    label = "files"
+
+    def ready(self):
+        from .models import File
+
+        for field in File._meta.get_fields():
+            if not field.many_to_many:
+                continue
+
+            def delete_orphaned_files(sender, m2m_field=field, **kwargs):
```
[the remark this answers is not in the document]
snopoke: There's currently no way to share files between assistants except using Django Admin or the a Django shell.

I used a many-to-many relation because I expect the files model to be used in other places as well and I didn't want to put a generic foreign key on `Files`.

I guess I could add a check here to make sure that there are no other references.

apps/assistants/sync.py
```diff
@@ -18,38 +47,118 @@ def push_assistant_to_openai(assistant: OpenAiAssistant):
...
+        purpose="assistants",
+    )
+    file.external_id = openai_file.id
+    file.external_source = "openai"
+    file.save()
+
+
+@wrap_openai_errors
+def delete_file_from_openai(assistant: OpenAiAssistant, file: File):
+    if not file.external_id or file.external_source != "openai":
```
SmittieC, line 66: I'm just trying to wrap my head around the usage of `external_source`. Looks like it's only ever used for this check here. This makes me wonder if the `external_id` would not be enough safeguard (assuming its some sort of uuid that is)? Unless we want to cater for cases where the external id is a plain integer accors multiple external sources, in which case one can accidentely remove a file from the wrong external source.

Is this the intented usage of the `external_source` field?

apps/assistants/sync.py
```diff
@@ -18,38 +47,118 @@ def push_assistant_to_openai(assistant: OpenAiAssistant):
...
+        return
+
+    client = assistant.llm_provider.get_llm_service().get_raw_client()
+    client.beta.assistants.files.delete(assistant_id=assistant.assistant_id, file_id=file.external_id)
+    try:
+        client.files.delete(file.external_id)
+    except openai.NotFoundError:
+        pass
+    file.external_id = ""
```
SmittieC, line 75: Is is necessary to clear these fields out, since the record have already been deleted by this time?

apps/files/apps.py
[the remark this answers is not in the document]
SmittieC: Got it. Up to you, as long as we know about this potentially causing problems down the line

apps/assistants/sync.py
snopoke: It's not strictly necessary since this only happens during deletion but I left it in since this function doesn't actually do the deletion so it seemed logical to reset the fields. This function could get used in other places where we keep the local file.

apps/assistants/sync.py
snopoke: This is designed to cater for multiple places where files may be pushed to in future e.g. if we use an external vector DB or push files to S3 or something else. It's a bit of over engineering at this point.

apps/assistants/views.py
```diff
@@ -115,8 +155,12 @@ class DeleteOpenAiAssistant(LoginAndTeamRequiredMixin, View, PermissionRequiredM
     @transaction.atomic()
     def delete(self, request, team_slug: str, pk: int):
         assistant = get_object_or_404(OpenAiAssistant, team=request.team, pk=pk)
+        try:
+            delete_openai_assistant(assistant)
+        except OpenAiSyncError as e:
+            messages.error(request, f"Error deleting assistant from OpenAI: {e}")
+            return HttpResponse(status=500)
```
SmittieC, line 162: Just confirming this wouldn't crash the page?

snopoke: no, it just displays the error message

templates/files/partials/file_list.html
```diff
@@ -0,0 +1,26 @@
+{# Template fragment for display a file list with ability to add and remove #}
+{# Required template vars: #}
+{#  - upload_url: URL to upload a file #}
+{#  - See also "files/partials/file_item.html" #}
+<div x-data="{}">
```
SmittieC, line 5: Should this be here?

templates/files/partials/file_formset.html
```diff
@@ -0,0 +1,27 @@
+<div x-data="{ count: 0, get zeroCount() { return this.count-1 } }">
+  <div class="flex flex-row justify-between mb-4">
+    <h3 class="text-lg font-medium">FILES</h3>
+    <button type="button" class="btn btn-sm"
+            @click="count++; $nextTick(() => { document.getElementById('id_files-' + zeroCount + '-file').click() })">
+      <i class="fa-solid fa-paperclip"></i> Add
+    </button>
+  </div>
+  <input type="hidden" name="files-TOTAL_FORMS" x-bind:value="count" id="id_files-TOTAL_FORMS">
```
SmittieC, line 9: Out of curiosity, why's `TOTAL_FORMS` in caps?

templates/files/partials/file_list.html
snopoke: Yes, it is a signal to Alpine to process the element: https://alpinejs.dev/directives/data#data-less-components

templates/files/partials/file_formset.html
SmittieC: Oh I see it's a django thing

apps/assistants/views.py
```diff
@@ -90,11 +98,38 @@ class CreateOpenAiAssistant(BaseOpenAiAssistantView, CreateView):
     button_text = "Create"
     permission_required = "assistants.add_openaiassistant"
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if "file_formset" not in context:
+            context["file_formset"] = self._get_file_formset()
+        return context
+
+    def _get_file_formset(self):
```
SmittieC, line 107: nit: since this is calling 1 function, you might as well call `get_file_formset` directly

SmittieC (approved):
Looks good to me :+1:

apps/assistants/views.py
snopoke: I'm going to leave this since this is called in 2 places and it does simplify a bit by not requiring you to pass the request

PR #261, opened 2024-02-16
Respond to unsupported message types

SmittieC:
This update nullifies [this PR](https://github.com/dimagi/open-chat-studio/pull/255).  **Please note** that this branch is based off of that branch, which is why the first commit in this branch is basically that branch's commit. It's just easier making this PR to main instead of that branch first, and then again to main after it's merged.

We now intead of just simply returning when the user sends an unsupported message type, respond to them. We also save the fact that the user tried to send such a message as a system message so users can see it in the chat transcript.

I also added a telegram message data class. The reason behind this is to have as much parity with other channels as possible.  (I might in a future PR refactor the message data structures to have similar fields, but we'll see). I also moved a few things around in the tests. Best to go through everything commit-wise. If you find me doing something stupid as you go through the commits, best to check later commits first to see if I didn't fix that. If I didn't, do let me know :laughing:


apps/channels/tests/test_whatsapp_integration.py
```diff
@@ -193,10 +192,6 @@ def test_parse_messages(self, message, message_type):
             assert whatsapp_message.content_type == MESSAGE_TYPES.VOICE
             assert whatsapp_message.media_url == "http://example.com/media"
 
-    def test_unsupported_message_type_exception_raised(self):
-        with pytest.raises(UnsupportedMessageTypeException):
-            TwilioMessage.model_validate(json.loads(TwilioMessages.image_message()))
```
snopoke: What about updating one of the other tests should be updated to include unsupported message types to make sure that no exception is raised

apps/channels/datamodels.py
```diff
@@ -18,6 +18,30 @@ def content(self) -> str:
         return self.message_text
 
 
+class TelegramMessage(BaseModel):
+    chat_id: int = Field()
```
snopoke, line 22: Isn't having `Field()` the same as making it equal to `None`?

snopoke (approved):
This looks great. Couple small comments.

apps/channels/tests/test_whatsapp_integration.py
snopoke: I think this was done in a later commit

SmittieC: Yeah, apologies, I did not spend too much time on my commit history. Should have cleaned it up a bit

apps/channels/datamodels.py
SmittieC: No. It will still expet you to specify it, and will not allow the field to be `None`. Having `= Field()` is a bit redundant though

snopoke (approved):

apps/channels/datamodels.py
snopoke: What about the other instances of the same thing?

SmittieC: :man_facepalming: https://github.com/dimagi/open-chat-studio/pull/261/commits/9a7c7e3456a98516d6e1d4d863b56440bb81135a

PR #262, opened 2024-02-16
Fix: deleting experiment channel updates

SmittieC:
- Deleting an experiment channel will not remove the experiment session anymore
- Removing an experiment channel from the UI only sets it to inactive. This is so that we can still query information from the experiment channel (like `get_platform_display` for instance). The channel will stay deactivated. When the user adds the channel again (so with exactly the same information), a new one will be created.


apps/channels/models.py
```diff
@@ -166,6 +172,10 @@ def webhook_url(self) -> str:
             is_secure=True,
         )
 
+    def soft_delete(self):
+        self.active = False
```
SmittieC, line 176: It doesn't look like the `active` field is being used for anything else. We could rename it to `deleted` also if need be.

snopoke: Makes sense to rename it to `deleted` to avoid confusion

snopoke (approved):

snopoke (approved):

PR #264, opened 2024-02-16
don't delete files unless they are truly orphaned

snopoke:
Followup from https://github.com/dimagi/open-chat-studio/pull/257#discussion_r1492111003

This was way more effort than expected but I think we can use some of the code for other things too e.g. displaying related objects before deletion


SmittieC (approved):
LGTM

PR #270, opened 2024-02-20
no activity testing and refactor

snopoke:
* refactor code a bit + tests
* use django-celery-beat to run periodic tasks directly
* use django-redis locks for locking

PR on top of #269


apps/chat/tasks.py
```diff
@@ -111,6 +111,11 @@ def _get_sessions_to_ping():
         Chat.objects.filter(pk__in=Subquery(subquery))
         .exclude(experiment_session__status__in=STATUSES_FOR_COMPLETE_CHATS)
         .exclude(experiment_session__experiment__no_activity_config=None)
+        .exclude(
```
SmittieC, line 117: :clap:

apps/chat/migrations/0010_celery_periodic.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 4.1.7 on 2023-04-05 06:47
+
+from django.db import migrations
+
+
+def create_celery_schedules(apps, schema_editor):
```
SmittieC, line 6: Out of curiosity, is there a specific reason why you're doing it this way instead of the way it was?

SmittieC: Cool, thanks! Another benefit is the ability to control this via the admin page if we wanted to.

SmittieC (approved):
Very nice :tophat:

PR #271, opened 2024-02-20
show input formatter for all experiment types

snopoke:
See #269

### DIFF
--- a/templates/experiments/experiment_form.html
+++ b/templates/experiments/experiment_form.html
@@ -25,7 +25,7 @@
       <div role="tablist" class="tabs tabs-bordered mt-4">
         <input type="radio" name="bot_type_tabs" role="tab" class="tab" aria-label="Use LLM Directly" {% if not object.assistant_id %}checked{% endif %}/>
         <div role="tabpanel" class="tab-content pt-4">
-          {% render_form_fields form "llm_provider" "llm" "temperature" "prompt_text" "input_formatter" "tools_enabled" "source_material" %}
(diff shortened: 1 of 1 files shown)


templates/experiments/experiment_form.html
```diff
@@ -25,7 +25,7 @@ <h3 class="font-semibold text-xl mt-4">Language Model Configuration</h3>
       <div role="tablist" class="tabs tabs-bordered mt-4">
         <input type="radio" name="bot_type_tabs" role="tab" class="tab" aria-label="Use LLM Directly" {% if not object.assistant_id %}checked{% endif %}/>
         <div role="tabpanel" class="tab-content pt-4">
-          {% render_form_fields form "llm_provider" "llm" "temperature" "prompt_text" "input_formatter" "tools_enabled" "source_material" %}
+          {% render_form_fields form "llm_provider" "llm" "temperature" "prompt_text" "tools_enabled" "source_material" %}
```
SmittieC, line 28: wait, am I going crazy? [This line](https://github.com/dimagi/open-chat-studio/blob/19a96be179439923095db94df28dcd22a2981e70/apps/web/templatetags/form_tags.py#L11) would be executed, meaning that "input_formatter" should be there?

snopoke: It was there, I was just moving it from the 'LLM' tab to underneath the tabs so that you can also see it when you configure an assistant.

SmittieC: Ohh, I see! Should have looked more properly. Thanks!

PR #274, opened 2024-02-21
pipeline prefactor on the way to migrating to langchain runnables

snopoke:
Some incremental changes to move closer to being able to make pipelines into langchain runnables.

Motivation:
* runnables handle pipelining (no separate pipeline implementation required)
* branching, parallel, merge all supported out the box
* they support dynamic configuration (no need for `PipelineContext`)

Review by commit.


SmittieC (approved):
Nice

PR #278, opened 2024-02-22
implement access control for chats

### LINKED ISSUE #168: Login and viewing transcripts
To recreate:

1. Get a link for a transcript. In my testing I used a Telegram transcript, but shouldn't matter.
2. Log out of OCS.
3. You can still view the transcript even though you're not logged in. Expected behaviour: it should fail.

This doesn't happen in incognito mode, so it's something about having been logged in at some point? I don't know.

My first guess would be to add:
```
@permission_required("experiments.XXX", raise_exception=True)
```
(where XXX is the `ChatViewer` permission) as a decorator for:
https://github.com/dimagi/open-chat-studio/blob/4346dd8030b9925ca712b52566ffbb513d0f16bc/apps/experiments/views/experiment.py#L669

snopoke:
Resolves #168 

This uses a signed cookie to verify access to chat sessions for web requests that are not authenticated.

Access is provided for 6 months. If the cookie is not present or expired the user will receive a 404 page.

Access is granted to authenticated users if they have the 'chat.view_chat' permission.

### DIFF
--- a/apps/experiments/decorators.py
+++ b/apps/experiments/decorators.py
@@ -79,7 +79,9 @@ def verify_session_access_cookie(view):
 
     @wraps(view)
     def _inner(request, *args, **kwargs):
-        if request.user.is_authenticated and request.user.has_perm("chat.view_chat"):
+        if request.user.is_authenticated and (
+            request.experiment_session.user_id == request.user.id or request.user.has_perm("chat.view_chat")
+        ):
             return view(request, *args, **kwargs)
 
         try:
--- a/apps/experiments/decorators.py
+++ b/apps/experiments/decorators.py
@@ -1,11 +1,18 @@
 from functools import wraps
 
 from django.contrib import messages
-from django.http import HttpResponseRedirect
+from django.core import signing
+from django.http import Http404, HttpResponseRedirect
 from django.shortcuts import get_object_or_404
 from django.urls import reverse
 
 from apps.experiments.models import Experiment, ExperimentSession, SessionStatus
+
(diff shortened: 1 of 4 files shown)


apps/experiments/decorators.py
```diff
@@ -30,6 +37,83 @@ def decorated_view(request, team_slug: str, experiment_id: str, session_id: str)
...
+            return view(request, *args, **kwargs)
+
+        try:
+            access_value = signing.get_cookie_signer(salt=CHAT_SESSION_ACCESS_SALT).unsign_object(
+                request.COOKIES[CHAT_SESSION_ACCESS_COOKIE], max_age=MAX_AGE
+            )
+        except (signing.BadSignature, KeyError):
+            raise Http404()
+
+        # access_data = json.loads(access_value)
```
snopoke, line 94:
```suggestion
```

apps/experiments/decorators.py
```diff
@@ -30,6 +37,82 @@ def decorated_view(request, team_slug: str, experiment_id: str, session_id: str)
...
+
+        try:
+            access_value = signing.get_cookie_signer(salt=CHAT_SESSION_ACCESS_SALT).unsign_object(
+                request.COOKIES[CHAT_SESSION_ACCESS_COOKIE], max_age=MAX_AGE
+            )
+        except (signing.BadSignature, KeyError):
+            raise Http404()
+
+        access_data = access_value
+        if not _validate_access_cookie_data(request.experiment_session, access_data):
```
snopoke, line 95:
```suggestion
        if not _validate_access_cookie_data(request.experiment_session, access_value):
```

Issue #268, opened 2024-02-20
Cancel pending LLM requests when a new message comes in


bderenzi:
A natural way to use a messaging platform is to send many short messages quickly. The LLMs are set up to take in a conversation history + user message and generate a reply. So, currently if we send many messages in a row it starts a response for each incoming message, which won't contain any previous responses (as they are pending) and potentially do not account for any corrections in subsequent messages.

I propose the following straightforward change:
- If a new user message is received and there is no pending LLM response, start an LLM response [current behavior]
- If a new user message is received and there is a pending LLM response, first cancel that pending LLM response, then start a new LLM response with the updated history (multiple user messages in a row).

This seems straightforward for the channel versions. It might be more work for the front end of the web interface, although we could disable the send button there until there is a response returned to avoid this situation (this is what ChatGPT does on the web and in their mobile app--they also stream the response and let you press the stop button to interrupt at any time).

snopoke:
This will need some re-design with pipelines.

snopoke:
This needs to be re-considered for pipelines

PR #280, opened 2024-02-23
support cancellation of generation

snopoke:
Part 1 of #268

(best reviewed by commit)

This updates the bot runnables to stream their outputs and detect if the output should be cancelled or not.

The implementation is slightly different for the different runnable types based on how they function:

* simple runnable
  * Just stream the tokens and check after each token if cancellation is necessary
* agent runnable
  * stream the steps and check after each step if cancellation is necessary. Agents don't stream individual tokens, only steps
* assistant runnable
  * let OpenAI tell us if another run is running and then cancel that run. Also handle the state when a run get's cancelled during processing.

Setting the cancellation status has been left for a future PR. The current mechanism uses the 'chat.metadata' dict to store a cancellation status.

## Safety

This is pretty well covered by tests and I have tested it locally. The cancellation checks are only executed every 1s so it won't add much load to the DB.

### DIFF
--- a/apps/service_providers/llm_service/runnables.py
+++ b/apps/service_providers/llm_service/runnables.py
@@ -1,8 +1,12 @@
+import logging
+import re
 from abc import ABC
 from datetime import datetime
 from operator import itemgetter
+from time import sleep
 from typing import Any, Literal
 
+import openai
 import pytz
 from langchain.agents import AgentExecutor, create_openai_tools_agent
 from langchain.agents.openai_assistant.base import OpenAIAssistantFinish
@@ -24,6 +28,12 @@ from apps.chat.agent.tools import get_tools
 from apps.chat.conversation import compress_chat_history
 from apps.chat.models import ChatMessage, ChatMessageType
 from apps.experiments.models import Experiment, ExperimentSession
+
+logger = logging.getLogger(__name__)
+
+
+class GenerationError(Exception):
+    pass
 
 
 class GenerationCancelled(Exception):
@@ -251,8 +261,6 @@ class AssistantExperimentRunnable(BaseExperimentRunnable):
         config["callbacks"] = config["callbacks"] or []
         config["callbacks"].append(callback)
 
-        assistant_runnable = RunnableLambda(self.format_input) | self.experiment.assistant.get_assistant()
-
         input_dict = {"content": input}
 
         if config.get("configurable", {}).get("save_input_to_history", True):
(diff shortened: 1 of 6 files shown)


apps/service_providers/llm_service/runnables.py
```diff
@@ -25,6 +30,17 @@
 from apps.chat.models import ChatMessage, ChatMessageType
 from apps.experiments.models import Experiment, ExperimentSession
 
+logger = logging.getLogger(__name__)
+
+
+class GenerationError(Exception):
```
SmittieC, line 36: Nit: Exceptions to be moved to their own file

apps/service_providers/llm_service/runnables.py
```diff
@@ -227,11 +280,51 @@ def invoke(self, input: str, config: RunnableConfig | None = None) -> ChainOutpu
...
+        if error_thread_id != thread_id:
+            raise GenerationError(f"Thread ID mismatch: {error_thread_id} != {thread_id}", exc)
+
+        self._cancel_run(assistant, thread_id, run_id)
+
+    def _cancel_run(self, assistant, thread_id, run_id):
+        logger.info("Cancelling run %s in thread %s", run_id, thread_id)
+        run = assistant.client.beta.threads.runs.cancel(thread_id=thread_id, run_id=run_id)
+        cancelling = run.status == "cancelling"
+        while cancelling:
```
SmittieC, line 326: Wondering if we should not include a timeout here, so incase something happens at openai and the status doesn't get updated, we don't go into an infinte loop. Thoughts?

SmittieC (approved):
Looks fine to me. Just two comments

SmittieC:
I'm curious if you considered just cancelling the celery task that's being used to get the response from the LLM and what your thoughts on that was? I _think_ we should be able to do the following:
1. store the task id that's used to get an LLM response in the chat metadata.
2. When the chat is cancelled, we can look up the current task id, stop that one and start the new task.

Curious what your thoughts are?

snopoke:
> I'm curious if you considered just cancelling the celery task that's being used to get the response from the LLM and what your thoughts on that was? I _think_ we should be able to do the following:
> 
> 1. store the task id that's used to get an LLM response in the chat metadata.
> 2. When the chat is cancelled, we can look up the current task id, stop that one and start the new task.
> 
> Curious what your thoughts are?

I did consider that but there are a few downsides:
1. it requires more plumbing (storing task ID's etc)
2. it's not recommended by Celery except for tasks that are hung
3. in some cases (e.g. assistants) we want to take some action when a task is cancelled (cancel the OpenAI run)

PR #284, opened 2024-02-27
change polling to avoid getting duplicate messages

snopoke:
There was a race condition where the poll would execute at the same time as a new message was added so the message would get added twice, once from htmx and once from the poll.

This is a minimal change to prevent that by cancelling the poll before an HTMX request is sent and then starting it again after.

### DIFF
--- a/templates/experiments/chat/chat_message_response.html
+++ b/templates/experiments/chat/chat_message_response.html
@@ -17,7 +17,7 @@
         Please try again, and wait a few minutes if this keeps happening.
       </p>
     {% else %}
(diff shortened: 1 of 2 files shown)


templates/experiments/chat/chat_message_response.html
```diff
@@ -17,7 +17,7 @@
         Please try again, and wait a few minutes if this keeps happening.
       </p>
     {% else %}
-      <span class="loading-dots">Thinking</span>
+      <span class="loading loading-dots loading-sm"></span>
```
snopoke, line 20: also a fix for the loading indicator

PR #287, opened 2024-02-27
Events 

proteusvacuum:
This PR implements an events system that can be built upon:
https://docs.google.com/document/d/1LDS23qVW9XpclVYV4WkS37b_8v6uairMbRG55taJL-s/edit#heading=h.uxp8b9jha1ya

The actions currently aren't very interesting, but there is a framework in the codebase that they can be added to later.

[Screencast from 2024-03-18 22-01-08.webm](https://github.com/dimagi/open-chat-studio/assets/146896/e3efd548-12ba-4b72-8e97-7d4e7596bc8c)


apps/events/models.py
```diff
@@ -0,0 +1,109 @@
...
+        open_experiment_sessions = self.experiment.sessions.filter(ended_at=None)
+        for session in open_experiment_sessions:
+            try:
+                if self.stats.get(session=session).trigger_count >= self.total_num_triggers:
+                    continue
+            except TriggerStats.DoesNotExist:
+                # This trigger didn't run yet
+                pass
+            last_human_message = session.chat.messages.filter(message_type=ChatMessageType.HUMAN).last()
+            if last_human_message and last_human_message.created_at < trigger_time:
```
snopoke, line 60: Could all this logic be done in the DB? Maybe similar to https://github.com/dimagi/open-chat-studio/pull/288?

apps/events/models.py
```diff
@@ -0,0 +1,109 @@
...
+            TriggerStats.objects.create(trigger=self, session=session, trigger_count=1, last_triggered=now)
+
+    def _end_conversation(self, session):
+        if self.end_conversation and self.stats.get(session=session).trigger_count >= self.total_num_triggers:
+            session.ended_at = datetime.now().astimezone(UTC)
+            session.status = SessionStatus.PENDING_REVIEW
+            session.save()
+
+
+class TriggerStats(BaseModel):
```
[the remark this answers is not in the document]
proteusvacuum: I like this, then I can also use it to store more information about each run (e.g. logs).

apps/events/models.py
```diff
@@ -0,0 +1,109 @@
...
+class TimeoutTrigger(BaseTrigger):
+    experiment = models.ForeignKey(Experiment, on_delete=models.CASCADE, related_name="timeout_triggers")
+    delay = models.PositiveIntegerField(
+        help_text="The amount of time in seconds to fire this trigger.",
+    )
+    total_num_triggers = models.IntegerField(
+        default=1,
+        help_text="The number of times to fire this trigger",
+    )
+    end_conversation = models.BooleanField(
```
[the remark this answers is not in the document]
proteusvacuum: Hmm, yes, maybe this warrants some discussion? 
For the `end_conversation` specifically - I agree that it is an "action", but it is also a side-effect of the final action occurring. I suppose this could be modeled with a separate Trigger and Action pair, where the trigger is that the final timeout has occurred.

apps/events/models.py
```diff
@@ -0,0 +1,109 @@
...
+    SUMMARIZE = "summarize"  # requires prompt
+
+
+class EventAction(BaseModel):
+    action_type = models.CharField(choices=EventActionType.choices)
+    params = models.JSONField(null=True, blank=True)  # The parameters for the specific action
+
+
+class BaseTrigger(BaseModel):
+    action = models.OneToOneField(EventAction, on_delete=models.CASCADE, related_name="action")
```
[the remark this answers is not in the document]
proteusvacuum: I had imagined that in the UI you would always define these together, i.e. something like:
"if (trigger) then do (action)"
I was having trouble thinking of when having a library of pre-configured actions would be useful where you couldn't just copy the definition from another event pair. 
I'd be happy to discuss if you have other things in mind though!

apps/events/models.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class EventActionType(models.TextChoices):
+    LOG = "log"  # Prints the last message
+    END_CONVERSATION = "end_conversation"  # Ends the conversation
+    SUMMARIZE = "summarize"  # Summarize the conversation
+
+
+class EventAction(BaseModel):
+    action_type = models.CharField(choices=EventActionType.choices)
```
snopoke, line 24: I'm torn about putting the choices in here. 
`<rant>`
It really irritates me that any time you change the Django makes a new migration file.
`</rant>`

apps/events/models.py
```diff
@@ -0,0 +1,165 @@
...
+
+    class Meta:
+        indexes = [
+            models.Index(fields=["content_type", "object_id"]),
+        ]
+
+
+class StaticTriggerType(models.TextChoices):
+    CONVERSATION_END = "conversation_end"
+    LAST_TIMEOUT = "last_timeout"
```
snopoke, line 53: Could also add `safety_layer_triggered`, `conversation_start`, `new_participant` (maybe `human_message`, `bot_message`)

apps/events/models.py
```diff
@@ -0,0 +1,165 @@
...
+            .values("id")[:1]
+        )
+        log_count_for_last_message = (
+            EventLog.objects.filter(
+                session=OuterRef("pk"),
+                chat_message_id=Subquery(last_human_message_id),
+                status=EventLogStatusChoices.SUCCESS,
+            )
+            .annotate(
+                count=Func(F("chat_message_id"), function="Count")
```
snopoke, line 115: you should be able to use `django.db.models.Count`
```suggestion
                count=Count("chat_message_id")
```

apps/events/models.py
```diff
@@ -0,0 +1,165 @@
...
+                last_human_message_created_at__isnull=False,
+            )  # The last message was received before the trigger time
+            .filter(
+                Q(log_count__lt=self.total_num_triggers) | Q(log_count__isnull=True)
+            )  # There were either no tries yet, or fewer tries than the required number for this message
+        )
+        return sessions.all()
+
+    def fire(self, session):
+        last_human_message = session.chat.messages.filter(message_type=ChatMessageType.HUMAN).last()
```
snopoke, line 140: I haven't checked where this get's called but doing these long chains can lead to a bunch of extra DB queries. It might be better to just query `ChatMessage` directly using `session.chat_id`

apps/experiments/models.py
```diff
@@ -486,6 +487,19 @@ def is_stale(self) -> bool:
...
         return self.status == SessionStatus.COMPLETE
 
-    def update_status(self, new_status: SessionStatus):
+    def update_status(self, new_status: SessionStatus, commit: bool = True):
         self.status = new_status
-        self.save()
+        if commit:
+            self.save()
+
+    def end(self, commit: bool = True, propagate: bool = True):
```
snopoke, line 495: nit: docstring for `propagate` would be useful

snopoke (commented):
Looking good

apps/events/models.py
proteusvacuum: Totally. Would you like me to add these here already?

apps/events/models.py
proteusvacuum: I added a comment about this on the next line. Unfortunately, we can't use `Count` because it breaks the subquery.

apps/events/models.py
proteusvacuum: Great point. I'll update it.

apps/events/models.py
snopoke: Right, I thought we'd used it before but I see we didn't do a subquery, we did an "id__in" query

proteusvacuum:
image

apps/events/models.py
```diff
@@ -0,0 +1,171 @@
...
+                last_human_message_created_at__isnull=False,
+            )  # The last message was received before the trigger time
+            .filter(
+                Q(log_count__lt=self.total_num_triggers) | Q(log_count__isnull=True)
+            )  # There were either no tries yet, or fewer tries than the required number for this message
+        )
+        return sessions.all()
+
+    def fire(self, session):
+        last_human_message = ChatMessage.objects.filter(chat_id=session.chat_id).last()
```
snopoke, line 146: Is this meant to be limited to human messages or just the last message? Variable says 'human' but there's no filtering.

apps/events/models.py
```diff
@@ -0,0 +1,171 @@
...
+                chat_message_id=Subquery(last_human_message_id),
+                status=EventLogStatusChoices.SUCCESS,
+            )
+            .annotate(
+                count=Func(F("chat_message_id"), function="Count")
+            )  # We don't use Count here because otherwise Django wants to do a group_by, which messes up the subquery: https://stackoverflow.com/a/69031027
+            .values("count")
+        )
+
+        sessions = (
```
snopoke, line 126: Perhaps for a future PR but I think if we want to replace the 'no activity pings' with this we'll need to be able to fire only for session where the last message was from a bot i.e. waiting for human response.

apps/events/views.py
```diff
@@ -0,0 +1,123 @@
...
+    return _create_event_view(StaticTriggerForm, request, team_slug, experiment_id)
+
+
+def _create_event_view(trigger_form_class, request, team_slug: str, experiment_id: str):
+    if request.method == "POST":
+        action_form = EventActionForm(request.POST)
+        if action_form.is_valid():
+            saved_action = action_form.save(experiment_id=experiment_id)
+            trigger_form = trigger_form_class(request.POST)
+            if trigger_form.is_valid():
```
snopoke, line 24: What happens if the trigger form isn't valid but the action form is? Can the user fix the errors and re-save or will that create 2 actions?

apps/events/views.py
```diff
@@ -0,0 +1,123 @@
...
+    model_class = {
+        "static": StaticTrigger,
+        "timeout": TimeoutTrigger,
+    }[trigger_type]
+
+    trigger = get_object_or_404(model_class, id=trigger_id, experiment_id=experiment_id)
+    if trigger_type == "timeout":
+        trigger_text = f"No response for {seconds_to_human(trigger.delay)}"
+    else:
+        trigger_text = StaticTriggerType(trigger.type).label
```
snopoke, line 114: nit: you could use Django's builtin `get_x_display`:
```suggestion
        trigger_text = trigger.get_type_display()
```

gpt_playground/settings.py
```diff
@@ -472,7 +473,7 @@
 # text-neutral text-error (pipeline run table styles here for tailwind build)
 DJANGO_TABLES2_ROW_ATTRS = {
     "class": "border-b border-gray-200 hover:bg-gray-100",
-    "id": lambda record: f"record-{record.id}",
+    "id": lambda record: f"record-{record.id}" if hasattr(record, "id") else "",
```
snopoke, line 476: when does this fail?

snopoke (approved):
Looks great. Some comments but nothing blocking

apps/events/models.py
proteusvacuum: Ah, great point. I think I forgot to add this back in when changing the query here: https://github.com/dimagi/open-chat-studio/pull/287/commits/896013224e777d5c955fa8d5c5e80c4acb9c080c

apps/events/models.py
proteusvacuum: Sounds good. Out of curiosity, would there ever be a time the last message _wasn't_ from a bot? Doesn't the bot always have the last word?

apps/events/views.py
proteusvacuum: Great point, will update

gpt_playground/settings.py
proteusvacuum: It fails because the records I was passing in to the table were dicts, not objects. I've updated this too.

apps/events/views.py
```diff
@@ -0,0 +1,120 @@
+from django.http import HttpResponseRedirect
+from django.shortcuts import get_object_or_404, render
+from django.urls import reverse
+
+from apps.events.forms import EventActionForm, StaticTriggerForm, TimeoutTriggerForm
+from apps.events.models import StaticTrigger, TimeoutTrigger
+from apps.utils.time import seconds_to_human
+
+
+def create_timeout_event_view(request, team_slug: str, experiment_id: str):
```
SmittieC, line 10: We need "login_and_team_required" decorators for these views?

apps/events/models.py
snopoke: That probably is true

apps/events/tables.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+        return format_html(
+            """
+            <a class="btn btn-sm btn-outline btn-primary" href="{}">View Logs</a>
+            <a class="btn btn-sm btn-outline btn-primary" href="{}">Edit</a>
+            <a class="btn btn-sm btn-outline btn-primary" href="{}">Delete</a>""",
+            view_log_url,
+            edit_url,
+            delete_url,
+        )
```
snopoke, line 45: It would be better to have this HTML in a template file so that it get's picked up by the tailwind post-processor.

I think you could just do `get_template("template name").render(context)`

apps/events/templates/events/manage_event.html
snopoke, line 1: I just noticed this now. The convention in this project is to put all templates in the root `templates` folder. So this should move to `templates/events/manage_event.html`.

Same for the other templates.

apps/events/views.py
snopoke: +1 for adding auth and permission decorators
```suggestion
@login_and_team_required
@permission_required("events.add_timeouttrigger")
def create_timeout_event_view(request, team_slug: str, experiment_id: str):
```

snopoke (approved):
Nothing blocking

PR #289, opened 2024-02-28
switch from decorator to normal function

snopoke:
https://dimagi.sentry.io/share/issue/39c915bc3de24cc4b8805cacf876fea4/


apps/experiments/views/experiment.py
```diff
@@ -639,7 +639,6 @@ def send_invitation(request, team_slug: str, experiment_id: str, session_id: str
     )
 
 
-@set_session_access_cookie
```
snopoke, line 642: this only worked in one of the 2 views

PR #299, opened 2024-03-02
Allow duplicate analysis steps

bderenzi:
I think I understand the analysis code slightly better now. A couple changes were necessary to make way for pipelines with duplicate steps:
1. We were using class name as a key in a dict, so it would only ever acknowledge a single instance of any given class type
2. We need to use `prefix` since we're mashing forms together. If we ever had a field with the same name across different steps (e.g., if we're duplicating a step type) it will clash and just use the final value. 

This doesn't solve all the problems: there is still the issue that duplicate pipeline steps all have the same labels in the config screen (e.g., `form_name = "LLM Completion Parameters"` is used for all `LlmCompletionStepParamsForm`).


snopoke (commented):
This is great! It's been on my [list](https://github.com/orgs/dimagi/projects/3/views/1?pane=issue&itemId=47586171) for a while too.

We also need to tell each step what it's step ID is so that the param loading works: https://github.com/dimagi/open-chat-studio/blob/main/apps/analysis/core.py#L273

We can do that using a similar strategy to what you've got and passing the offset and index here: https://github.com/dimagi/open-chat-studio/blob/5ec03950780600c2c0da444a0d8548242c8c7209/apps/analysis/pipelines.py#L24-L25

bderenzi:
Great- thanks. I meant to follow up to say that it doesn't work yet. :) 

The param loading is the issue I ran into. I updated the `BaseStep` to accept an unique name and added it where you said, but I still couldn't get it to work. I'll have another look this afternoon if I can get some time. I might just push what I have if I can't get it as I suspect you'll be able to get through it much more quickly.

And, just to check: loops should be fine, right? I think we just need a unique step name so that params are saved/loaded correctly. Then we have the UUID id for different instances if we have a list to process, right?

snopoke:
@bderenzi this is working now

snopoke (approved):

PR #300, opened 2024-03-05
UI and model changes for uploading files for experiments

snopoke:
This PR re-works the experiment form view to make it easier to interact with. The form has many fields and it requires lots of scrolling to interact with. This replaces scrolling with tabs for each 'section' which I think is much easier to use and also presents a less overwhelming view to the users.

This PR also adds support for attaching files to experiments (under the 'source material' tab). The files are not currently used but future PRs will add RAG functionality to experiments.

The files UI is hidden a feature flag.

(This PR also removes usages of the 'assistants' flag which has now been released.

Peek 2024-03-05 11-54


templates/experiments/experiment_form.html
```diff
@@ -16,35 +19,62 @@
...
+      <div role="tabpanel" class="tab-content">
+        {% render_form_fields form "assistant" %}
+      </div>
+      <input type="radio" name="main_tabs" role="tab" class="tab" aria-label="Safety">
+      <div role="tabpanel" class="tab-content">
+        {% render_form_fields form "safety_layers" "safety_violation_notification_emails" "input_formatter" "max_token_limit" %}
+      </div>
+      <input type="radio" name="main_tabs" role="tab" class="tab" aria-label="Consent">
+      <div role="tabpanel" class="tab-content">
+        {% render_form_fields form "conversational_consent_enabled" %}
```
SmittieC, line 49: Would it not make more sense to show the "consent" form here instead of at the top?

templates/experiments/experiment_form.html
```diff
@@ -16,35 +19,62 @@
...
-          {% render_form_fields form "llm_provider" "llm" "temperature" "prompt_text" "tools_enabled" "source_material" %}
-        </div>
+  <div>
+    {% render_form_fields form "name" "description" "consent_form" %}
+    <div class="divider"></div>
+    {% render_form_fields form "type" %}
+    {% if not form.fields.assistant.choices %}
+      <div x-show="type === 'assistant'">
+        <p class="text-error">{% blocktranslate %}
+          No assistants have been created. Please create an assistant before creating an 'assistant' experiment.
```
SmittieC, line 29: Nit: How about this (in this or a future PR): "There are no assistants available to use. <a>Click here</a> to create one."

apps/experiments/views/experiment.py
```diff
@@ -231,24 +247,59 @@ class CreateExperiment(BaseExperimentView, CreateView):
     button_title = "Create"
     permission_required = "experiments.add_experiment"
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if "file_formset" not in context:
```
SmittieC, line 252: when would "file_formset" already be in the context?

SmittieC: Oh wait I think it would be when there's some error and we return the invalid form. Nevermind, I should have read further

templates/experiments/experiment_form.html
snopoke: That is where I had it initially but since it's required, if you leave it out there's not indication that it is the field which is preventing you from submitting since it's hidden behind a tab.

templates/files/partials/file_list.html
```diff
@@ -9,15 +9,15 @@ <h3 class="text-lg font-medium">FILES</h3>
       <i class="fa-solid fa-paperclip"></i> Add
     </button>
   </div>
-  <form id="fileUploadForm" enctype="multipart/form-data"
-        hx-post="{{ upload_url }}"
-        hx-trigger="change"
-        hx-swap="afterbegin"
-        hx-target="#fileList"
+  <div id="fileUploadForm" hx-encoding="multipart/form-data"
```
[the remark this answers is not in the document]
snopoke: Since this is using htmx to submit the data it doesn't need to be a form.

SmittieC (approved):
This is cool

PR #306, opened 2024-03-06
A few misc changes resulting from Sentry errors

snopoke:


gpt_playground/settings.py
```diff
@@ -270,7 +270,10 @@
 if AWS_ACCESS_KEY_ID:
     AWS_SECRET_ACCESS_KEY = env("AWS_SECRET_ACCESS_KEY")
     AWS_S3_REGION = env("AWS_S3_REGION", default=None)
-    WHATSAPP_S3_AUDIO_BUCKET = env("WHATSAPP_AWS_AUDIO_BUCKET", default="ocs-whatsapp-voice")
+    WHATSAPP_S3_AUDIO_BUCKET = env("WHATSAPP_S3_AUDIO_BUCKET", default="")
```
SmittieC, line 273: Can we not just drop support for `WHATSAPP_AWS_AUDIO_BUCKET`?

snopoke: Do you mean use the 'AWS_PUBLIC_STORAGE_BUCKET_NAME'?

SmittieC: We did discuss this offline, but just putting it here for visibility. We're fetching the value for `WHATSAPP_S3_AUDIO_BUCKET` from `WHATSAPP_AWS_AUDIO_BUCKET` if `WHATSAPP_S3_AUDIO_BUCKET` is not specified. I'm suggesting we just drop support for `WHATSAPP_AWS_AUDIO_BUCKET`.

You then suggested that we rather show a temporary warning message, after which we'll drop support to give people chance to update their environment.

snopoke: updated

SmittieC (approved):

PR #308, opened 2024-03-06
Audio fix:Convert synthesized audio to a supported format

SmittieC:
This addresses the issue where Azure's synthesized audio is in wav format, so when this voice provider is used with the Twilio messaging provider - which supports only mp3 - stuff breaks.

This PR adds a `SynthesizedAudio` dataclass to hold info pertaining to synthesized audio. This info is used to convert the audio to a supported format for the messaging service. We could have simply converted the synthesized audio from Azure to mp3 before returning it to the `BaseChannel` class, but it doesn't feel very robust, since then we have to remember that all synthesized audio must be in mp3 format (or document it and trust that people will read and rember this detail). Another approach could be to check if it's mp3 just before returning and raise an error if it's not, but blehg. I instead opted for the approach of returning the synthesized audio info (the raw audio, duration and format) so that each messaging provider can convert is as needed.

Tested this locally and it works

### DIFF
--- a/apps/channels/tests/test_whatsapp_integration.py
+++ b/apps/channels/tests/test_whatsapp_integration.py
@@ -96,40 +96,6 @@ class TestTwilio:
             # elif message_type == "audio": TODO: Figure out why this is not passing in the github workflows
             #     s3_client_mock.generate_presigned_url.assert_called()
 
-    @pytest.mark.usefixtures("_twilio_whatsapp_channel")
-    @pytest.mark.parametrize(
-        "synthesized_audio_format",
-        ["wav", "mp3"],
-    )
-    @patch("apps.channels.audio.convert_audio")
-    @patch("apps.chat.channels.ChannelBase._get_voice_transcript")
-    @patch("apps.service_providers.speech_service.SpeechService.synthesize_voice")
-    @patch("apps.chat.channels.WhatsappChannel._get_llm_response")
-    def test_convert_audio_to_supported_format(
-        self,
-        get_llm_response_mock,
-        synthesize_voice_mock,
-        _get_voice_transcript,
-        convert_audio_mock,
-        synthesized_audio_format,
-    ):
-        """Test that the synthesized audio is converted to mp3 before being sent to Twilio"""
-        synthesize_voice_mock.return_value = SynthesizedAudio(
-            audio=BytesIO(b"123"), duration=10, format=synthesized_audio_format
-        )
-        _get_voice_transcript.return_value = "Hi"
-        convert_audio_mock.return_value = BytesIO(b"123")
-
-        with patch("apps.service_providers.messaging_service.TwilioService.s3_client"), patch(
-            "apps.service_providers.messaging_service.TwilioService.client"
-        ):
-            get_llm_response_mock.return_value = "Hi"
-            handle_twilio_message(message_data=twilio_messages.audio_message())
-            if synthesized_audio_format == "mp3":
-                convert_audio_mock.assert_not_called()
-            else:
-                convert_audio_mock.assert_called()
-
 
 class TestTurnio:
     @pytest.mark.parametrize(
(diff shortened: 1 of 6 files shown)


apps/channels/tests/test_whatsapp_integration.py
```diff
@@ -96,6 +96,40 @@ def test_twilio_uses_whatsapp_channel_implementation(
             # elif message_type == "audio": TODO: Figure out why this is not passing in the github workflows
             #     s3_client_mock.generate_presigned_url.assert_called()
 
+    @pytest.mark.usefixtures("_twilio_whatsapp_channel")
```
SmittieC, line 99: Technically these tests are overkill, since it tests a single if-statement.

apps/service_providers/messaging_service.py
```diff
@@ -106,11 +113,11 @@ def client(self) -> TurnClient:
     def send_whatsapp_text_message(self, message: str, from_number: str, to_number):
         self.client.messages.send_text(to_number, message)
 
-    def send_whatsapp_voice_message(self, voice_audio: BytesIO, duration: int, from_number: str, to_number: str):
-        from apps.channels.audio import convert_audio
-
+    def send_whatsapp_voice_message(self, synthetic_voice: SynthesizedAudio, from_number: str, to_number: str):
         # OGG must use the opus codec: https://whatsapp.turn.io/docs/api/media#uploading-media
-        voice_audio = convert_audio(voice_audio, target_format="ogg", source_format="mp3", codec="libopus")
```
snopoke, line 113: Why are we converting to ogg here if the source format is already mp3?

apps/service_providers/messaging_service.py
```diff
@@ -60,7 +61,13 @@ def s3_client(self):
     def send_whatsapp_text_message(self, message: str, from_number: str, to_number):
         self.client.messages.create(from_=f"whatsapp:{from_number}", body=message, to=f"whatsapp:{to_number}")
 
-    def send_whatsapp_voice_message(self, voice_audio: BytesIO, duration: int, from_number: str, to_number):
+    def send_whatsapp_voice_message(self, synthetic_voice: SynthesizedAudio, from_number: str, to_number):
+        voice_audio = synthetic_voice.audio
+        if synthetic_voice.format != "mp3":
+            voice_audio = audio.convert_audio(
+                synthetic_voice.audio, target_format="mp3", source_format=synthetic_voice.format
+            )
```
snopoke, line 69: what about moving this to the SynthesizedAudio class so that you can call:

```python
required_formats = [...]  # sorted by preference (if file is not already in a format in the list it will be converted to the first format listed)
synthetic_voice.get_audio_bytes(supported_formats=required_formats)
```

The conversion will happen in the `SynthesizedAudio` class if necessary.

(Supported formats appear to be ACC, MP4, AMR, MPEG, OGG codec = OPUS)

SmittieC: https://github.com/dimagi/open-chat-studio/pull/308/commits/989dc8475607c28c8b598a537eec9d4e67ab22c1

apps/service_providers/messaging_service.py
SmittieC: Just putting what we discussed here: TurnIO expects ogg (amongst other formats. I only got ogg working for now). We always convert to ogg to make sure it used the "libopus" codec.

apps/service_providers/tests/test_synthesized_audio.py
```diff
@@ -0,0 +1,23 @@
...
+from unittest import mock
+
+import pytest
+
+from apps.service_providers.speech_service import SynthesizedAudio
+
+
+@pytest.mark.parametrize(
+    ("source_format", "target_format", "codec", "conversion_expected"),
+    [("mp3", "ogg", None, True), ("mp3", "mp3", None, False), ("mp3", "ogg", "libopus", True)],
```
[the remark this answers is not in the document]
SmittieC: You're right, thanks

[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/308/commits/da059c2a1f3bb84570165d7d39ad1b659e050f4f

PR #311, opened 2024-03-08
update some field lengths in the DB

snopoke:
The 2nd commit isn't really necessary but I thought I'd just go through them all while I'm making migrations.

As a rule of thumb I've stuck to 128 chars for name fields where we expect the user to type it and 255 if it's autogenerated.

For the `llm` field (model name) I've made it 255 just to be on the safe side.


apps/chat/models.py
```diff
@@ -18,9 +18,9 @@ class Chat(BaseTeamModel):
     class MetadataKeys(StrEnum):
         OPENAI_THREAD_ID = "openai_thread_id"
 
-    # tbd what goes in here
     user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, null=True, blank=True)
-    name = models.CharField(max_length=100, default="Unnamed Chat")
+    # must match or be greater than experiment name field
+    name = models.CharField(max_length=128, default="Unnamed Chat")
```
SmittieC, line 23: We don't use this name anywhere, do we?

SmittieC (approved):

apps/chat/models.py
snopoke: Only in Django admin currently

PR #315, opened 2024-03-12
Remove extra newlines in LLMs response

SmittieC:
A very naive way of fixing the weird spacing we get when the LLM returns a list. There are typically two newline characters between the items, where the first one gets converted into a \<li\> or \<p\> tag and the second into a \<br\> tag when we call `linebreaksbr`. If we simply remove the `linebreaksbr` call, then the paragraphs gets squashed which doesn't look very good. Any recommendations welcome.

## Before
b_1
b_2

## After
a_1
a_2

### DIFF
--- a/apps/chat/templatetags/chat_tags.py
+++ b/apps/chat/templatetags/chat_tags.py
@@ -11,4 +11,10 @@ register = template.Library()
 def render_markdown(text):
     if not text:
         return ""
-    return linebreaksbr(mark_safe(markdown.markdown(text, extensions=[FencedCodeExtension()])))
+    text = markdown.markdown(text, extensions=[FencedCodeExtension()])
+    text = text.replace("</li>\n<li>", "</li><li>")
+    text = text.replace("<ol>\n<li>", "<ol><li>")
+    text = text.replace("</li>\n</ol>", "</li></ol>")
+    text = text.replace("<li>\n<p>", "<li><p>")
+    text = text.replace("</p>\n</li>", "</p></li>")
+    return linebreaksbr(mark_safe(text))


apps/chat/templatetags/chat_tags.py
```diff
@@ -11,4 +11,10 @@
 def render_markdown(text):
     if not text:
         return ""
-    return linebreaksbr(mark_safe(markdown.markdown(text, extensions=[FencedCodeExtension()])))
+    text = markdown.markdown(text, extensions=[FencedCodeExtension()])
+    text = text.replace("</li>\n<li>", "</li><li>")
+    text = text.replace("<ol>\n<li>", "<ol><li>")
+    text = text.replace("</li>\n</ol>", "</li></ol>")
+    text = text.replace("<li>\n<p>", "<li><p>")
+    text = text.replace("</p>\n</li>", "</p></li>")
```
snopoke, line 19: What about using a regex to match all these cases? `<(.*?)>\\n<(.*?)>`

SmittieC: Tried this. There is an exception to the `<(.*?)>\\n<(.*?)>` rule, which is two p tags. A newline between a closing and opening p tag looks very natural, so we don't want to match that. Will have to play around with regex a bit to see how we can do this. I'm not 100% sure if this is the only exception to the rule, but I think it is.

SmittieC:
@snopoke can I merge?

apps/chat/templatetags/chat_tags.py
snopoke: @SmittieC could you add a comment with the outcome of our discussion for future reference. Specifically what elements the regex catches that we don't want it to.

snopoke (approved):

apps/chat/templatetags/chat_tags.py
SmittieC: Sure! So the cases I added which should be excluded are the only ones (so far) to be addressed. There are cases that should not be changed, like \<p\>\\n\<\p\>. This means we cannot simply use regex to replace all instances where a newline character is present between html tags.

PR #317, opened 2024-03-13
take summary length into account when compression chat history

snopoke:
The previous implementation compressed the history to below the max length but then added the summary on top of that.

This update continues to loop until the history is exhausted or the total length (including the summary) is less than the max.

It is still possible to get a message > max if the summary alone is greater but that seems unlikely.

Resolves https://dimagi.sentry.io/share/issue/dcccf4ff11de4c94b371e996dfcdc850/

### DIFF
--- a/apps/chat/conversation.py
+++ b/apps/chat/conversation.py
@@ -115,10 +115,12 @@ def compress_chat_history(
     """
     history = chat.get_langchain_messages_until_summary()
     if max_token_limit <= 0 or not history:
+        log.info("Skipping chat history compression")
         return history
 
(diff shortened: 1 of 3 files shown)


apps/chat/conversation.py
```diff
@@ -17,6 +17,8 @@
 
 from apps.chat.models import Chat, ChatMessage, ChatMessageType
 
+INITIAL_SUMMARY_TOKENS_ESTIMATE = 20
```
SmittieC, line 20: This is just random right?

SmittieC (approved):
Cool

PR #318, opened 2024-03-14
Chat annotations - PR 1

SmittieC:
[Spec](https://docs.google.com/document/d/19lxITyXjASGXNoZSqN0KR4BHNfouiL_q1pW3FynAuOY/edit#heading=h.4u2212lqaqrg)

- I used [django-taggit](https://django-taggit.readthedocs.io/en/latest/)  and customed the `Tag` and its through models to suit our needs. (Turns out it's much easier to just do this than to try and implement this ourselves).
- The `Chat` model supports tags now.
- I added a `tag_multiselect` component (html + JS, which uses Tom Select underneath)

## Screenshots
image

(The white doesn't look the best in dark mode. I couldn't find a nice way to update the background color)
image


templates/experiments/experiment_form.html
```diff
@@ -2,7 +2,6 @@
 {% load i18n %}
 {% load waffle_tags %}
 {% load form_tags %}
-{% load waffle_tags %}
```
SmittieC, line 5: this is already loaded on line 3

templates/generic/tag_multiselect.html
```diff
@@ -0,0 +1,115 @@
+<div x-data="dropdown()" x-init="loadOptions()">
+  <input name="values" type="hidden" x-bind:value="selectedValues()">
+
+  <div id="linkTagDiv" data-url="{% url 'annotations:link_tag' object_id=object_id %}"></div>
+  <div id="unlinkTagDiv" data-url="{% url 'annotations:unlink_tag' object_id=object_id %}"></div>
```
snopoke, line 5: Are these used for anything other than storing the links? If not you could just put the links directly into the JS.

templates/generic/tag_multiselect.html
```diff
@@ -0,0 +1,115 @@
...
+      options: [],
+      selected: [],
+      show: false,
+      open() { this.show = true },
+      close() { this.show = false },
+      isOpen() { return this.show === true },
+      select(index, event) {
+        if (!this.options[index].selected) {
+          this.options[index].selected = true;
+          this.options[index].element = event.target;
```
snopoke, line 77: what is `.element` used for?

apps/annotations/admin.py
```diff
@@ -0,0 +1,15 @@
+from django.contrib import admin
+
+from .models import CustomTaggedItem, Tag
+
+
+@admin.register(Tag)
+class ChatAdmin(admin.ModelAdmin):
```
snopoke, line 7:
```suggestion
class TagAdmin(admin.ModelAdmin):
```

apps/annotations/models.py
```diff
@@ -0,0 +1,20 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+from django.utils.translation import pgettext_lazy
+from taggit.models import TagBase, TaggedItem
+
+from apps.teams.models import BaseTeamModel
+
+
+class Tag(TagBase, BaseTeamModel):
+    name = models.CharField(verbose_name=pgettext_lazy("A tag name", "name"), max_length=100)
```
snopoke, line 10: could we add a 'created_by' field here for auditing?

apps/annotations/views.py
```diff
@@ -0,0 +1,97 @@
...
+        "button_text": "Create",
+        "active_tab": "tags",
+    }
+
+    def get_success_url(self):
+        return reverse("experiments:tag_home", args=[self.request.team.slug])
+
+    def form_valid(self, form):
+        form.instance.team = self.request.team
+        form.instance.owner = self.request.user
```
snopoke, line 43: I don't see an `owner` field on tags. But see comment above about adding a `created_by` field

apps/annotations/models.py
```diff
@@ -0,0 +1,20 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+from django.utils.translation import pgettext_lazy
+from taggit.models import TagBase, TaggedItem
+
+from apps.teams.models import BaseTeamModel
+
+
+class Tag(TagBase, BaseTeamModel):
```
snopoke, line 9: Maybe getting ahead of myself but `tag_color` would be a nice to have (#scope-creep)

apps/annotations/views.py
```diff
@@ -0,0 +1,97 @@
...
+    model = Tag
+    paginate_by = 25
+    table_class = TagTable
+    template_name = "table/single_table.html"
+
+    def get_queryset(self):
+        return Tag.objects.filter(team=self.request.team)
+
+
+class UnlinkTag(LoginAndTeamRequiredMixin, View):
```
snopoke, line 82: D-oes linking / unlinking get recorded anywhere e.g. the audit logs? If not, that's a something to make not of for future work.

templates/generic/tag_multiselect.html
```diff
@@ -0,0 +1,115 @@
...
+              <div class="flex justify-center items-center m-1 font-medium py-1 px-1 rounded bg-gray-100 hover:bg-gray-100">
+                <div class="text-xs font-normal leading-none max-w-full flex-initial" x-model="options[option]" x-text="options[option].text"></div>
+                <div class="flex flex-auto flex-row-reverse">
+                  <div x-on:click.stop="remove(index,option)">
+                    <svg class="fill-current h-4 w-4 " role="button" viewBox="0 0 20 20">
+                      <path d="M14.348,14.849c-0.469,0.469-1.229,0.469-1.697,0L10,11.819l-2.651,3.029c-0.469,0.469-1.229,0.469-1.697,0
+                               c-0.469-0.469-0.469-1.229,0-1.697l2.758-3.15L5.651,6.849c-0.469-0.469-0.469-1.228,0-1.697s1.228-0.469,1.697,0L10,8.183
+                               l2.651-3.031c0.469-0.469,1.228-0.469,1.697,0s0.469,1.229,0,1.697l-2.758,3.152l2.758,3.15
+                               C14.817,13.62,14.817,14.38,14.348,14.849z" />
+                    </svg>
```
snopoke, line 23: nit: use `<i class="fa-solid fa-xmark"></i>`

templates/generic/tag_multiselect.html
snopoke, line 1: this is quite specific to tags so should probably live in the `annotations` template folder

snopoke: Also, you may want to consider a library for this. I think the best non-JQuery version is [Tom Select](https://tom-select.js.org/). I've used it before and found it to be pretty good.

It will also be handy for other places in the UI e.g. LLM provider models input

snopoke (commented):
Looking good. Some general comments but nothing blocking.

apps/experiments/urls.py
```diff
@@ -124,3 +124,4 @@
 urlpatterns.extend(make_crud_urls(views, "Survey", "survey"))
 urlpatterns.extend(make_crud_urls(views, "ConsentForm", "consent"))
 urlpatterns.extend(make_crud_urls(views, "NoActivityMessageConfig", "no_activity"))
+urlpatterns.extend(make_crud_urls(views, "Tag", "tag"))
```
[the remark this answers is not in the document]
SmittieC: I'm yet to move it :+1:

templates/generic/tag_multiselect.html
SmittieC: Do you mean inject it into the JS script like this?
```JS
<script>
let url = {% url 'annotations:link_tag' %}
</script>
```

snopoke: yes, though with quotes to make it a string

SmittieC: Tried that, it's not rendering properly. See https://github.com/dimagi/open-chat-studio/pull/318/commits/602e57f308287fe5f68002075d30c25e769f84c4. I extracted the JS code into a dedicated .js file so it's easy to reuse in the future. For this though, I put the urls on the script tag as data attributes, since we cannot render stuff in that JS file like we do with html templates. I'd be glad if there is a way to still do it the "template way", but I couldn't find something in my brief search

templates/generic/tag_multiselect.html
SmittieC: Not applicable anymore, since we're using tom select now

apps/experiments/urls.py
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/318/commits/640db42e1179f890c81bfbbef268c69369a6a736

apps/annotations/views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/318/commits/0831828bc95552dcb11e0239eac0edca89340606

apps/annotations/views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/318/commits/0831828bc95552dcb11e0239eac0edca89340606

templates/generic/tag_multiselect.html
SmittieC: This PR's size is getting a bit out of hand. I'm just going to add some tests to finalize this PR

templates/experiments/components/experiment_chat.html
```diff
@@ -30,13 +32,15 @@
...
               </td>
               <td class="px-3 py-4 text-sm text-gray-500">{{ message.content|render_markdown }}</td>
-              {% flag "chat_annotations" %}
-                <td scope="col" class="w-1/4 overflow-y-visible">
-                  <div class="my-2 mr-2">
-                    {% include "generic/tag_multiselect.html" with object_tags=message.get_linked_tags object_info=message.object_info %}
-                  </div>
-                </td>
-              {% endflag %}
+              {% if taggible|default:False %}
```
SmittieC, line 35: This is to prevent showing the tag component when the user ends the experiment chat and is taken to the transcript view

templates/generic/tag_multiselect.html
snopoke: using the data tags seems like a good compromise to me

apps/annotations/Readme.md
```diff
@@ -0,0 +1,19 @@
...
+### Include the `tag_multiselect.html` template in your template:
+
+```
+{% include "generic/tag_multiselect.html" with object_tags=object_variable %}
+```
+### Load the script
+
+```html
+<script
+    id="tag-multiselect"
```
snopoke, line 11: Add comment that the ID must be kept the same
```suggestion
    id="tag-multiselect"  (you must use this ID)
```

apps/annotations/migrations/0002_alter_tag_options.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('annotations', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
```
snopoke, line 13: would be good to merge these migrations into a single one

apps/annotations/models.py
```diff
@@ -0,0 +1,64 @@
...
+    created_by = models.ForeignKey("users.CustomUser", on_delete=models.DO_NOTHING)
+
+    class Meta:
+        verbose_name = _("Tag")
+        verbose_name_plural = _("Tags")
+        unique_together = ("team", "name")
+        ordering = ["name"]
+
+
+@audit_fields(
```
snopoke, line 29: Do you know how easy it is to query these? e.g. if we were showing a list of changes to a 'chat' object and wanted to include changes to tags?

apps/annotations/tables.py
```diff
@@ -0,0 +1,28 @@
...
+
+class TagTable(tables.Table):
+    actions = columns.TemplateColumn(
+        template_name="generic/crud_actions_column.html",
+        extra_context={
+            "actions": [
+                actions.edit_action(url_name="annotations:tag_edit"),
+                actions.delete_action(
+                    url_name="annotations:tag_delete",
+                    confirm_message="Continuing with this action will remove this tag from any tagged entity",
```
snopoke, line 17: 👍🏻

static/tag_multiselect.js
```diff
@@ -0,0 +1,27 @@
...
+  return function() {
+    let postData = {swap: 'none', values: {"tag_name": arguments[0], "object_info": objectInfo}};
+    htmx.ajax('POST', linkTagUrl, postData);
+  };
+};
+
+var removeTag = function(name, objectInfo) {
+  return function() {
+    let postData = {swap: 'none', values: {"tag_name": arguments[0], "object_info": objectInfo}};
+    htmx.ajax('POST', unlinkTagUrl, postData);
```
snopoke, line 15: Is there a good reason to use htmx rather than a vanilla `fetch` if we're not doing any swapping?

templates/experiments/components/experiment_chat.html
snopoke: Could this be replaced with a permission check? If the user is logged in a has permissions then it seems OK to show the UI at that stage in the workflow.

static/tag_multiselect.js
SmittieC: Not much except that htmx handles the csrf token for us it seems. When using fetch, I need to manually get the csrf token and inject it into the request headers

apps/annotations/models.py
SmittieC: Easy
```python
from field_audit.models import AuditEvent
AuditEvent.objects.filter(...)
```

apps/annotations/models.py
```diff
@@ -33,8 +33,23 @@ class Meta:
     "object_id",
     "content_type",
 )
-class CustomTaggedItem(TaggedItem, BaseTeamModel):
+class CustomTaggedItem(GenericTaggedItemBase, BaseTeamModel):
     user = models.ForeignKey("users.CustomUser", on_delete=models.DO_NOTHING)
+    tag = models.ForeignKey(Tag, related_name="%(app_label)s_%(class)s_items", on_delete=models.CASCADE)
```
SmittieC, line 38: All I can say is...oops :man_facepalming:

apps/annotations/models.py
snopoke: Sure, but I mean if I give you a chat object and and want to get back a list of tag changes for that chat?

SmittieC: Right. So this is what we have from the audit log:

image

We can use the object's app + model name to query the audit event records, then we can pluck the username from the `change_context` and find the user that did it (or just look at `delta` if we only care about what changed).

SmittieC:
@snopoke bump :smile_cat:

snopoke (approved):

PR #331, opened 2024-03-22
Fix: Consider message type when checking what the user said

SmittieC:
Fixes [this issue](https://dimagi.sentry.io/issues/5086818703/?project=4505001320316928&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=7d&stream_index=5). The issue is due to the code looking at `channel.message_text` during the pre-conversation flow, thereby assuming that it will always be populated, but when a voice message is sent, it is `None`. We now call a method that extracts the user's query on demand.

Note: We don't support voice interactions for the pre-conversation flow, but this is an issue for another time/PR. At least it doesn't break now

### DIFF
--- a/apps/channels/tests/test_base_channel_behavior.py
+++ b/apps/channels/tests/test_base_channel_behavior.py
@@ -320,6 +320,7 @@ def test_reply_with_text_when_synthetic_voice_not_specified(
     send_text_to_user.assert_called()
 
 
+@pytest.mark.django_db()
 @pytest.mark.parametrize(
     ("message_func", "message_type"),
     [(telegram_messages.audio_message, "voice"), (telegram_messages.text_message, "text")],
@@ -327,7 +328,7 @@ def test_reply_with_text_when_synthetic_voice_not_specified(
 @patch("apps.chat.channels.TelegramChannel.send_text_to_user")
 @patch("apps.chat.channels.TelegramChannel._get_llm_response")
 def test_user_query_extracted_for_pre_conversation_flow(
-    _get_llm_response, send_text_to_user_mock, message_func, message_type, db
+    _get_llm_response, send_text_to_user_mock, message_func, message_type
 ):
     """The user query need to be available during the pre-conversation flow. Simply looking at `message_text` for
     this is erroneous, since it will not be available when the user sends a voice message.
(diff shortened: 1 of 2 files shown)


apps/chat/channels.py
```diff
@@ -217,7 +227,7 @@ def _handle_pre_conversation_requirements(self):
         (Status==PENDING_PRE_SURVEY) user indicated that they took the survey -> sett status to ACTIVE
         """
         # We manually add the message to the history here, since this doesn't follow the normal flow
-        self._add_message_to_history(self.message_text, ChatMessageType.HUMAN)
```
snopoke, line 220: Is `message_text` used anywhere now? Should all references be switched?

SmittieC: yes, `message_text` is only used inside the `_extract_user_query` method

apps/channels/tests/test_base_channel_behavior.py
```diff
@@ -318,3 +318,38 @@ def test_reply_with_text_when_synthetic_voice_not_specified(
...
+
+
+@pytest.mark.parametrize(
+    ("message_func", "message_type"),
+    [(telegram_messages.audio_message, "voice"), (telegram_messages.text_message, "text")],
+)
+@patch("apps.chat.channels.TelegramChannel.send_text_to_user")
+@patch("apps.chat.channels.TelegramChannel._get_llm_response")
+def test_user_query_extracted_for_pre_conversation_flow(
+    _get_llm_response, send_text_to_user_mock, message_func, message_type, db
```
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/331/commits/c905e62f02312b941e5f04f3eb27ca2363a02941

apps/chat/channels.py
snopoke: Can we rename it to `get_message_text` to be symmetrical with `get_message_audio` (and make it a method).

Side note: it seems like a lot of the subclasses of `ChannelBase` have the same implementation for some of these functions like `message_content_type`. Feels like we could make it cleaner. Maybe there should be a base message type which exposes a consistent interface rather that having the channel manage it?

snopoke (approved):

PR #333, opened 2024-03-22
Fix: Do not call .output on the string response

SmittieC:
Fixes [this issue](https://dimagi.sentry.io/share/issue/415c7c67fa254b22b2bb6f3456813566/). I was able to reproduce the issue with the updates I made to the test.


SmittieC:
Lol, fixed [here](https://github.com/dimagi/open-chat-studio/pull/332/files)

PR #335, opened 2024-03-22
Initial vector db implementation

snopoke:
This adds a vector DB implementation based off [langchain pgvector](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/vectorstores/pgvector.py) but backed by a custom Django model.

This can not be merged to main yet since our current production DB does not support the pgvector extension.

## Upgrade note
If you want to update without loosing all your existing data in postgres you can follow the instructions here: https://gist.github.com/snopoke/80e9a2e88ec493b7d4efae2578339f00


apps/vectordb/tests/test_vectorstore.py
```diff
@@ -0,0 +1,143 @@
...
+def test_vectorstore(experiment):
+    """Test end to end construction and search."""
+    texts = ["foo", "bar", "baz"]
+    docsearch = PGVector.from_texts(
+        texts=texts,
+        embedding=DeterministicFakeEmbedding(size=ADA_TOKEN_COUNT),
+        experiment=experiment,
+    )
+    output = docsearch.similarity_search("foo", k=1)
+    _remove_meta_fields(output)
```
SmittieC, line 31: Why's the meta fields removed here?

apps/vectordb/tests/test_vectorstore.py
```diff
@@ -0,0 +1,143 @@
...
+        experiment=experiment,
+        distance_strategy=DistanceStrategy.COSINE,
+    )
+    output = docsearch.similarity_search_with_score("foo", k=3)
+    _remove_meta_fields(output)
+    # ordering here is deterministic but random due to fake embeddings
+    assert output == [
+        (Document(page_content="foo", metadata={"page": "0"}), 0.0),
+        (Document(page_content="baz", metadata={"page": "2"}), 0.9290842232061864),
+        (Document(page_content="bar", metadata={"page": "1"}), 1.0246011368123038),
```
SmittieC, line 118: Interesting that this is larger than 1. I presume these scores refer to angles then?

apps/vectordb/tests/test_vectorstore.py
snopoke: yes, just for testing so that we can compare the results without having to compare the metadata

apps/vectordb/tests/test_vectorstore.py
snopoke: I was a bit surprised by this too but then I checked and pgvector returns the cosine distance and not the cosine similarity (https://github.com/pgvector/pgvector/blob/d7354a86a8496a4432b9d27e2eaf5658274ed1f4/src/vector.c#L696)

### tl;dr

Similarity is the angle which ranges from -1 to 1
Distance is 1 - similarity which ranges from 0 to 2

See https://medium.com/@milana.shxanukova15/cosine-distance-and-cosine-similarity-a5da0e4d9ded for a good article on the difference.

SmittieC: That makes sense, thanks!

PR #340, opened 2024-03-25
OpenAI TTS Support

### LINKED ISSUE #296: OpenAI TTS support
Can we add [OpenAI TTS](https://platform.openai.com/docs/guides/text-to-speech) as a voice provider? I'm not sure how we want to do it since it will share an API key with an LLM provider.

SmittieC:
Resolves #296. Since OpenAI automatically detects the language, I left the language section of the `SyntheticVoice` records empty. OpenAI doesn't specify voice genders, so I guessed it, but we can remove this if we want.


apps/experiments/migrations/0072_load_openai_voices.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 4.2.7 on 2024-03-25 08:50
+
+from django.db import migrations
+import os
+import json
+
+def load_synthetic_voices(apps, schema_editor):
```
snopoke, line 7: would be great to abstract this and the other 2 methods (in 0032 and 0044) to a utility function

apps/service_providers/speech_service.py
```diff
@@ -147,3 +148,33 @@ def transcribe_audio(self, audio: BytesIO) -> str:
...
+        """
+        response = self._client.audio.speech.create(model="tts-1", voice=synthetic_voice.name, input=text)
+        audio_data = response.read()
+
+        audio_segment = AudioSegment.from_file(BytesIO(audio_data), format="mp3")
+        duration_seconds = len(audio_segment) / 1000  # Convert milliseconds to seconds
+        return SynthesizedAudio(audio=BytesIO(audio_data), duration=duration_seconds, format="mp3")
+
+    def transcribe_audio(self, audio: BytesIO) -> str:
+        transcript = self._client.audio.transcriptions.create(
```
snopoke, line 176: we should probably have some error handling here

snopoke (approved):

apps/service_providers/speech_service.py
SmittieC: I added https://github.com/dimagi/open-chat-studio/pull/340/commits/399845285b39eaa084af458b5adc4899756b9816. I think if we catch + raise an `AudioTranscriptionException` where we call it from the channel.py file, we should be fine. I'm wondering, we _can_ provide the user with feedback, like we do when the user sends an unsupported message. In this case, we'll just tell the user "listen, what you sent us isn't working. Please send something better". Obviously the bot will formulate an appropriate response here...

apps/experiments/migrations/0032_load_synthetic_voices.py
```diff
@@ -35,5 +20,5 @@ class Migration(migrations.Migration):
         ("experiments", "0031_promptbuilderhistory"),
     ]
 
-    operations = [migrations.RunPython(load_synthetic_voices, drop_synthetic_voices)]
+    operations = [migrations.RunPython(drop_synthetic_voices, drop_synthetic_voices)]
```
snopoke, line 23: you have `drop_synthetic_voices` twice now.

apps/experiments/migrations/0032_load_synthetic_voices.py
```diff
@@ -1,30 +1,15 @@
 # Generated by Django 4.2 on 2023-09-28 07:13
 import os
-import json
 from django.db import migrations
+from apps.utils import migration_utils
 
 
-def load_synthetic_voices(apps, schema_editor):
+def drop_synthetic_voices(apps, schema_editor):
```
snopoke, line 7: This should be `load_synthetic_voices`

apps/service_providers/speech_service.py
snopoke: One nice thing about doing the generic error handling in the service is that users of the service only need to catch the service exceptions and not base Exception.

apps/experiments/migrations/0043_update_AWS_synthetic_voices.py
```diff
@@ -1,37 +1,15 @@
...
 from django.db import migrations
+from apps.utils import migration_utils
 
 
 def load_language_code(apps, schema_editor):
     SyntheticVoice = apps.get_model("experiments", "SyntheticVoice")
-    voice_data = {}
     current_directory = os.path.dirname(os.path.abspath(__file__))
     file_path = os.path.join(current_directory, "preload_data/aws_voices.json")
+    migration_utils.create_synthetic_voices_from_file(SyntheticVoice, file_path=file_path)
```
SmittieC, line 11: I forgot to use the utility here. Context: We created AWS voices, but then later wanted to add `language_code` and `service` to the model as well. We then reran the migration to load the voices, only this time with `SyntheticVoice.objects.update_or_create(...)`, which is why I update the util function to use `update_or_create` rather than `get_or_create`. The json file is the latest one (so it has the concept of "service" and "language code", so no need to accomodate versions of this file)

apps/service_providers/speech_service.py
```diff
@@ -36,10 +37,7 @@ class SpeechService(pydantic.BaseModel):
 
     def synthesize_voice(self, text: str, synthetic_voice: SyntheticVoice) -> SynthesizedAudio:
         assert synthetic_voice.service == self._type
-        try:
-            return self._synthesize_voice(text, synthetic_voice)
-        except Exception as e:
-            raise AudioSynthesizeException(f"Unable to synthesize audio with {self._type}: {e}") from e
```
snopoke, line 42: how come this error handling was removed?

apps/service_providers/speech_service.py
snopoke: Yes, I mean doing something like this: https://github.com/dimagi/open-chat-studio/blob/main/apps/service_providers/speech_service.py#L39-L42

Where we catch and re-raise as an app specific exception. The benefit is that callers only have to care about that one exception and not any others.

apps/experiments/migrations/0043_update_AWS_synthetic_voices.py
SmittieC: Oh sorry I forgot to mention, I realized that we can't do that. Reason being that we added the `SyntheticVoice` in [migration 0023](https://github.com/dimagi/open-chat-studio/blob/main/apps/experiments/migrations/0023_syntheticvoice_experiment_synthetic_voice.py). We loaded the AWS voices in [migration 32](https://github.com/dimagi/open-chat-studio/blob/main/apps/experiments/migrations/0032_load_synthetic_voices.py), and then altered the synthetic voice model in [migration 42](https://github.com/dimagi/open-chat-studio/blob/main/apps/experiments/migrations/0042_alter_syntheticvoice_unique_together_and_more.py), meaning that even though the first migration ran, it will ignore the `service` and `language_code` fields from the json file. In the second migration where we load the voices, all those records will be updated, so we sort of need this one to also run :+1:

apps/service_providers/speech_service.py
SmittieC: Yeah that makes sense

SmittieC: https://github.com/dimagi/open-chat-studio/pull/340/commits/ef9c0ff55216740f577bcfc159b3c1801f981139

apps/experiments/migrations/0032_load_synthetic_voices.py
snopoke: did you change this - looks like it is still 'drop' instead of 'load'

SmittieC: I did, and double checked it :raised_eyebrow:, but....it's gone.. :open_mouth:

apps/utils/migration_utils.py
```diff
@@ -24,13 +24,20 @@ def create_synthetic_voices_from_file(SyntheticVoiceModel, file_path: str):
...
         voice_data = json.load(json_file)
 
-    model_fields = [f.name for f in SyntheticVoiceModel._meta.get_fields()]
-
     voices_created = 0
     service = voice_data["service"]
-    for voice_data in voice_data["voices"]:
-        voice_data = {**voice_data, "service": service}
-        voice_fields = {key: value for key, value in voice_data.items() if key in model_fields}
+    for voice in voice_data["voices"]:
```
SmittieC, line 29: Why am I doing this? When this method is first used, not all fields exists on the `SyntheticVoiceModel` model, so we now use only the fields that do exist. The `0043_update_AWS_synthetic_voices` migration will update the missing fields

SmittieC:
@snopoke There's just one more update I had to make, because there was a conflict with migration file names

snopoke (approved):

PR #341, opened 2024-03-27
Support FB Messenger through Twilio

SmittieC:
This adds support for FB Messenger through Twilio. This way, users don't need to create a Meta app, they can simply link their page to twilio and point it to the server running OCS. I removed the elements related to the previous Messenger implementation, so we just need to update our FB integration. There's only one.

Note: Twilio does not support Instagram, so we need to find another way to integrate that


apps/channels/datamodels.py
```diff
@@ -46,48 +47,47 @@ def parse(update_obj) -> "TelegramMessage":
...
     @staticmethod
     def parse(message_data: dict) -> "TwilioMessage":
+        prefix_channel_map = {"messenger": ChannelPlatform.FACEBOOK, "whatsapp": ChannelPlatform.WHATSAPP}
+        prefix = message_data["From"].split(":")[0]
         content_type = message_data.get("MediaContentType0")
         return TwilioMessage(
-            from_number=message_data["From"],
-            to_number=message_data["To"],
+            from_=message_data["From"].split(f"{prefix}:")[1],
+            to=message_data["To"].split(f"{prefix}:")[1],
```
snopoke, line 85: Rather than `.split()[1]` you should use `.removeprefix(...)` which is simpler and safer

apps/channels/tests/test_facebook_integration.py
```diff
@@ -1,93 +1,87 @@
...
+            (twilio_messages.Messenger.audio_message(), "audio"),
+        ],
+    )
+    @patch("apps.service_providers.speech_service.SpeechService.synthesize_voice")
+    @patch("apps.chat.channels.ChannelBase._get_voice_transcript")
+    @patch("apps.service_providers.messaging_service.TwilioService.send_text_message")
+    @patch("apps.chat.channels.FacebookMessengerChannel._get_llm_response")
+    def test_twilio_uses_facebook_channel_implementation(
+        self,
+        get_llm_response_mock,
```
snopoke, line 69: Should this also assert that `get_llm_response_mock` get's called?

apps/service_providers/messaging_service.py
```diff
@@ -76,20 +78,30 @@ def send_whatsapp_voice_message(self, synthetic_voice: SynthesizedAudio, from_nu
...
+        public_url = self._upload_audio_file(synthetic_voice)
+        self.client.messages.create(from_=f"{prefix}:{from_}", to=f"{prefix}:{to}", media_url=[public_url])
 
     def get_message_audio(self, message: TwilioMessage) -> BytesIO:
         auth = (self.account_sid, self.auth_token)
-        ogg_audio = BytesIO(requests.get(message.media_url, auth=auth).content)
-        return audio.convert_audio(ogg_audio, target_format="wav", source_format="ogg")
+        response = requests.get(message.media_url, auth=auth)
+        # Example header: {'Content-Type': 'audio/ogg'}
+        content_type = response.headers["Content-Type"].split("/")[1]
```
snopoke, line 103: what other content types might we expect?

templates/web/components/messages.html
```diff
@@ -19,7 +19,7 @@
           Toast.fire({
             icon: '{{ message.tags }}',
             title: '{{ message.tags|capfirst }}',
-            text: '{{ message }}'
+            text: '{{ message|escape }}'
```
snopoke, line 22: curious what you observed that resulted in this

snopoke (approved):
Some small comments but nothing blocking. Does this mean we've given up on the direct FB integration?

apps/channels/tests/test_facebook_integration.py
SmittieC: We don't really have to, since it must have been called when `send_text_message` was called. I'll in anycase just update the code to assert stuff when voice was used. I see this is missing

apps/service_providers/messaging_service.py
SmittieC: So when we get a voice message from whatsapp through twilio, its `audio/ogg`, but when we get it from Messenger through twilio, its `video/mp4`.

templates/web/components/messages.html
SmittieC: Oh this was a quick thing I tried, because I saw this:
Screenshot from 2024-04-02 13-56-00

I added the `escape` filter but sort of forgot about this and didn't test it again. Not only should I not have escaped it (should have used `|safe` or something), it didn't fix the issue. I'll just remove the escape filter and fix this in a separate PR. Thanks for asking.

apps/channels/datamodels.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/341/commits/42cae86d8f3cc0bf5a9638afe61f24b16d3653b8

snopoke (approved):

snopoke:
```
FAILED apps/channels/tests/test_facebook_integration.py::TestTwilio::test_twilio_uses_facebook_channel_implementation[{"SmsMessageSid": "DDDDDDDDDDDDDdd", "NumMedia": "0", "ProfileName": "Chris Smit", "SmsSid": "CCCCCCCCCCCCCCCCCCCCCCCCCC", "WaId": "27456897512", "SmsStatus": "received", "Body": "Dobroye utro", "To": "messenger:14155238886", "NumSegments": "1", "ReferralNumMedia": "0", "MessageSid": "BBBBBBBBBB", "AccountSid": "AAAAAAAAAAAAA", "From": "messenger:27456897512", "ApiVersion": "2010-04-01", "MediaContentType0": "audio/ogg", "MediaUrl0": "http://example.com/media"}-audio] - AssertionError: Expected 'send_voice_message' to have been called.
```

apps/channels/tests/test_facebook_integration.py
```diff
@@ -60,6 +60,7 @@ def test_parse_messages(self, message, message_type):
             (twilio_messages.Messenger.audio_message(), "audio"),
         ],
     )
+    @patch("apps.service_providers.messaging_service.settings.AWS_ACCESS_KEY_ID", "123")
```
SmittieC, line 63: This fixes that failing test. I couldn't get [this method](https://stackoverflow.com/questions/31582750/python-mock-patch-os-environ-and-return-value) to work, so resorting to this alternative.

snopoke: Did you try [@override_settings](https://docs.djangoproject.com/en/5.0/topics/testing/tools/#django.test.override_settings)?

SmittieC: I did not (I might have searched for the wrong thing!), but I did [use it here](https://github.com/dimagi/open-chat-studio/pull/350/commits/58b16d59073b5bd30121c4e2fb66bce6cec9c578) and it works! Thanks Simon!

PR #342, opened 2024-03-28
Minor improvements

SmittieC:
- Fix: The seed message is now sent to the bot to generate a response to the user during the pre-conversation flow (previously the seed message was sent directly to the user, which is wrong)
- I updated some tests to use `@pytest.mark.django_db()` instead of the `db` fixture.
- I made table rows clickable (it bothers me a little that it's not :sweat_smile: ). The object being listed need an `get_absolute_url` method to be define the url to the the user to when they click on the row. If there isn't one, nothing will happen. If we want to have this feature, we'll need to remove the edit button from some objects


templates/table/single_table.html
```diff
@@ -1,2 +1,19 @@
+{% load static %}
 {% load render_table from django_tables2 %}
 {% render_table table "table/tailwind_js_pagination.html" %}
+<script>
+    function initializeRowClickHandlers() {
+        document.querySelectorAll('.clickable').forEach(function(element) {
+            element.addEventListener('click', function() {
+                let editUrl = this.getAttribute('data-redirect-url');
+                window.location.href = editUrl;
```
proteusvacuum, line 9: what happens if this is `""`? Might be worth just adding the `clickable` class when there is a url.

gpt_playground/settings.py
```diff
@@ -475,8 +475,10 @@
 }
 # text-neutral text-error (pipeline run table styles here for tailwind build)
 DJANGO_TABLES2_ROW_ATTRS = {
-    "class": "border-b border-gray-200 hover:bg-gray-100",
+    "class": "border-b border-gray-200 hover:bg-gray-100 clickable",
     "id": lambda record: f"record-{record.id}",
+    "data-redirect-url": lambda record: record.get_absolute_url() if hasattr(record, "get_absolute_url") else "",
+    "style": "cursor: pointer",
```
proteusvacuum, line 481: What if you added this in CSS to the clickable class? That way those items without a URL wouldn't look like they were clickable (if you also take the suggestion in my comment below)

SmittieC: Good idea :+1:

SmittieC: I used a class from tailwind instead: https://github.com/dimagi/open-chat-studio/pull/342/commits/2b33cd1029bfc89105c834332af4b4c3e1f8b5dd

apps/chat/channels.py
```diff
@@ -252,7 +252,8 @@ def start_conversation(self):
         self.experiment_session.update_status(SessionStatus.ACTIVE)
         # This is technically the start of the conversation
         if self.experiment.seed_message:
-            self._send_message_as_bot(self.experiment.seed_message)
+            bot_response = self._generate_response_for_user(self.experiment.seed_message)
+            self.send_text_to_user(bot_response)
```
snopoke, line 262: should this not use `new_bot_message`?

templates/table/single_table.html
snopoke: +1 to not doing anything if the 'editUrl' is empty

gpt_playground/settings.py
```diff
@@ -475,8 +475,9 @@
 }
 # text-neutral text-error (pipeline run table styles here for tailwind build)
 DJANGO_TABLES2_ROW_ATTRS = {
-    "class": "border-b border-gray-200 hover:bg-gray-100",
+    "class": "border-b border-gray-200 hover:bg-gray-100 hover:cursor-pointer",
```
snopoke, line 478: Using a dynamic selector we can make this class only apply when the data attribute is present and not empty:

`data-[redirect-url]:[&:not([data-redirect-url=""])]:hover:cursor-pointer`

See:
* https://stackoverflow.com/questions/17248915/css-attribute-selector-for-existing-but-non-empty-attributes
* https://github.com/tailwindlabs/tailwindcss/discussions/11822#discussioncomment-6717861
* https://tailwindcss.com/docs/hover-focus-and-other-states#using-arbitrary-variants

apps/chat/channels.py
SmittieC: Correct, it should!

gpt_playground/settings.py
SmittieC: Hmm, I could not get this working unfortunately. The hover class isn't being applied. See https://github.com/dimagi/open-chat-studio/pull/342/commits/93ad98eefbdfa6c4929d48200f8bbc874a93fb17 for the change to select only elements where the redirect url is actually populated

snopoke (approved):

PR #344, opened 2024-04-01
Ability to comment on chat messages

SmittieC:
Users can now comment on individual chat messages. They can only create and delete their own comments, but not edit it (yet?) or delete other user's comments. When the user clicks on a message row, it will now expand and show all related comments. I wanted to use a sidebar for the comments rather, but couldn't get it to work.

## Screenshot
Screenshot from 2024-04-01 10-28-29

FYI the comments will be improved a bit in a next PR


apps/annotations/views/comment_views.py
```diff
@@ -0,0 +1,42 @@
...
+        object_info = json.loads(request.POST["object_info"])
+        object_id = object_info["id"]
+        content_type = get_object_or_404(ContentType, app_label=object_info["app"], model=object_info["model_name"])
+        chat_message = content_type.get_object_for_this_type(id=object_id)
+        try:
+            UserComment.add_for_model(
+                chat_message, comment=request.POST["comment"], added_by=request.user, team=request.team
+            )
+        except FieldDoesNotExist:
+            messages.error(request, "Unable to add comment to this entity")
```
snopoke, line 28: nit: include content type name in error message:
```suggestion
            messages.error(request, f"Unable to add comment to '{content_type.name}' objects.")
```

apps/annotations/views/comment_views.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class UnlinkComment(LoginAndTeamRequiredMixin, View, PermissionRequiredMixin):
+    permission_required = "annotations.delete_usercomment"
+
+    def post(self, request, team_slug: str):
+        object_info = json.loads(request.POST["object_info"])
+        object_id = object_info["id"]
+        content_type = get_object_or_404(ContentType, app_label=object_info["app"], model=object_info["model_name"])
+        chat_message = content_type.get_object_for_this_type(id=object_id)
```
snopoke, line 40: nit: rename var (same as above)

apps/annotations/views/comment_views.py
```diff
@@ -0,0 +1,42 @@
...
+        UserComment.objects.get(id=request.POST["comment_id"], team__slug=team_slug).delete()
```
snopoke, line 41: nit: use `get_object_or_404` instead of plain `.get`

apps/chat/models.py
```diff
@@ -6,12 +6,12 @@
...
 from langchain.schema import BaseMessage, messages_from_dict
 
-from apps.annotations.models import BaseTaggedModel
+from apps.annotations.models import TaggedModelMixin, UserCommentsMixin
 from apps.teams.models import BaseTeamModel
 from apps.utils.models import BaseModel
 
 
-class Chat(BaseTeamModel, BaseTaggedModel):
+class Chat(BaseTeamModel, TaggedModelMixin):
```
snopoke, line 14: Just confirming that we plan to add comment support to chats later.

apps/experiments/export.py
```diff
@@ -10,9 +10,28 @@ def _parse_tags(tags: list[Tag]) -> str:
     return ", ".join([t.name for t in tags])
 
 
+def _parse_comments(user_comments: list[UserComment]) -> str:
+    """Parses `user_comments` into a single string that looks like this:
```
snopoke, line 14: I think `parse` is the wrong term here. It should be `format`: `_format_comments`

apps/experiments/export.py
```diff
@@ -10,9 +10,28 @@ def _parse_tags(tags: list[Tag]) -> str:
...
+    <username_1>: "user 1's comment" | <username_2>: "user 2's comment" | <username_1>: "user 1's comment"
+    """
+    comment_str = ""
+    for idx, comment in enumerate(user_comments):
+        if idx > 0:
+            comment_str = f'{comment_str} | <{comment.user.username}>: "{comment.comment}"'
+        else:
+            comment_str = f'<{comment.user.username}>: "{comment.comment}"'
+    return comment_str
```
snopoke, line 23: What about this:
```python
" | ".join([str(comment) for comment in user_comments])
```

Where `comment.__str__` does the single comment formatting: `'<{self.user.username}>: "{self.comment}"'`

apps/experiments/export.py
```diff
@@ -10,9 +10,28 @@ def _parse_tags(tags: list[Tag]) -> str:
...
 def experiment_to_message_export_rows(experiment: Experiment, filter_tags: list[str] = []):
     queryset = experiment.sessions.prefetch_related(
-        "chat", "chat__messages", "participant", "experiment_channel", "chat__tags", "chat__messages__tags"
+        "chat",
+        "chat__messages",
+        "participant",
+        "experiment_channel",
+        "chat__tags",
+        "chat__messages__tags",
+        "chat__messages__comments",
```
snopoke, line 34: probably also add `chat__messages__comments__user` otherwise we'll be doing a DB query per comment to load the username

apps/annotations/models.py
```diff
@@ -52,16 +54,7 @@ class Meta:
...
-        abstract = True
-
-    tags = TaggableManager(through=CustomTaggedItem)
-
-    def add_tags(self, tags: list[str], team: Team, added_by: CustomUser):
-        for tag in tags:
-            self.tags.add(tag, through_defaults={"team": team, "user": added_by})
-
+class AnnotationMixin:
     @property
```
snopoke, line 58: make this `cached_property` since it's called in a loop in the template

apps/annotations/models.py
```diff
@@ -74,7 +67,46 @@ def object_info(self):
...
+class UserCommentsMixin(models.Model, AnnotationMixin):
+    comments = GenericRelation(UserComment)
+
+    class Meta:
+        abstract = True
+        ordering = ["created_at"]
+
+    @property
+    def get_user_comments(self) -> UserComment:
+        return self.comments.all()
```
snopoke, line 112: maybe prefetch the comment user field?

snopoke (commented):
Looking good

apps/annotations/views/comment_views.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class LinkComment(LoginAndTeamRequiredMixin, View, PermissionRequiredMixin):
+    permission_required = "annotations.add_usercomment"
+
+    def post(self, request, team_slug: str):
+        object_info = json.loads(request.POST["object_info"])
+        object_id = object_info["id"]
+        content_type = get_object_or_404(ContentType, app_label=object_info["app"], model=object_info["model_name"])
+        chat_message = content_type.get_object_for_this_type(id=object_id)
```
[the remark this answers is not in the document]
SmittieC: Good catch thanks

apps/annotations/urls.py
```diff
@@ -6,8 +6,10 @@
 app_name = "annotations"
 
 urlpatterns = [
-    path("link/", views.LinkTag.as_view(), name="link_tag"),
-    path("unlink/", views.UnlinkTag.as_view(), name="unlink_tag"),
+    path("link/tag/", views.LinkTag.as_view(), name="link_tag"),
+    path("unlink/tag", views.UnlinkTag.as_view(), name="unlink_tag"),
+    path("link/comment/", views.LinkComment.as_view(), name="link_comment"),
+    path("unlink/comment", views.UnlinkComment.as_view(), name="unlink_comment"),
```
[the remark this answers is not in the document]
SmittieC: You're right, `add/remove` does make more sense here :+1:

apps/annotations/models.py
```diff
@@ -74,7 +67,46 @@ def object_info(self):
...
+
+class UserCommentsMixin(models.Model, AnnotationMixin):
+    comments = GenericRelation(UserComment)
+
+    class Meta:
+        abstract = True
+        ordering = ["created_at"]
+
+    @property
+    def get_user_comments(self) -> UserComment:
```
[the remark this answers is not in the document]
SmittieC: That's a fair mental model

apps/experiments/export.py
SmittieC: Ah, I used it wrong all the time, thanks! :sweat_smile:

apps/experiments/export.py
SmittieC: I like this, yeah

apps/annotations/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/344/commits/dfbc5fef42f3eef9e71b42152d7c9280c18d68f5

apps/experiments/export.py
```diff
@@ -10,9 +10,23 @@ def _parse_tags(tags: list[Tag]) -> str:
     return ", ".join([t.name for t in tags])
 
 
+def _parse_comments(user_comments: list[UserComment]) -> str:
```
snopoke, line 13: nit: also rename the function
```suggestion
def _format_comments(user_comments: list[UserComment]) -> str:
```

snopoke (approved):

PR #345, opened 2024-04-01
Ability to add comments to chats

SmittieC:
We can now add comments to chats as well. I also made the comment UI a bit nicer.

## Screenshots
Screenshot from 2024-04-01 15-27-10
Screenshot from 2024-04-01 15-27-21

### Still to do
- The number showing the number of comments doesn't update when you add/delete a comment. I must still look into this


templates/experiments/components/experiment_chat.html
```diff
@@ -42,15 +42,17 @@
                 </td>
               {% endif %}
               {% if perms.annotations.add_usercomment %}
-                <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500"><i class="fa-regular fa-comment"></i> {{message.get_user_comments|length}}</td>
+                <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">
```
SmittieC, line 45: Nicer / more succinct approaches welcome!

snopoke: You could use [hx-swap-oob](https://htmx.org/attributes/hx-swap-oob/) in the 'add_comment' view response to update the comment count.

I used this in pipelines: https://github.com/dimagi/open-chat-studio/blob/ccda7fcc79c3fb103ed098e951be57bffa6fd36a/templates/analysis/components/group_progress_inner.html#L2

SmittieC: Of course! Thanks

snopoke (approved):

Issue #338, opened 2024-03-24
Add error logging to EventLog and add count of errors to experiment view page


proteusvacuum:
We should add error logging to the EventLog model whenever an event encounters an error.
Once that is done, add a count of errors on the list of events at the bottom of the experiment page.

proteusvacuum:
https://github.com/dimagi/open-chat-studio/pull/346

PR #346, opened 2024-04-01
Add log to EventLog model

proteusvacuum:
Fixes [#338]


templates/events/view_logs.html
```diff
@@ -37,6 +38,7 @@ <h1 class="pg-title">{{ title }}</h1>
                                 </a>
                             </td>
                             <td>{{ event_log.chat_message.content }}</td>
+                            <td>{{ event_log.log }}</td>
```
SmittieC, line 41: What's the motivation for showing this to the user? They'll just see a stacktrace, which wouldn't be helpful I'd guess? Also, what about sensitive data that might be revealed? :eyes:

snopoke: For errors they would see the result of `str(e)` which is generally just the exception class + message (no stack trace).

For normal events they would see the log message like 'Session ended'.

snopoke (approved):

PR #348, opened 2024-04-02
Add EventLog UI to the Experiment Session page

### LINKED ISSUE #337: Add EventLog UI to the Experiment Session page
We want to add the UI for Events / Triggers to the experiment session page. 
There is the `_logs_view` helper in `events/views.py` that could be useful here.

This table should only show when the user has permission to view it (i.e. it should only show to admins, not to users at the end of their session).

proteusvacuum:
Adds event logs to the bottom of the experiment session page. 
Fixes #337

image


SmittieC (approved):
Pretty cool!

snopoke (approved):
Perhaps for a separate PR, but what do you think about using tabs in the experiment view?
* My Sessions
* All Sessions
* Events

proteusvacuum:
I like the idea of tabs, that page is pretty long now. I made a new task: https://github.com/dimagi/open-chat-studio/issues/352

PR #349, opened 2024-04-03
Add no activity ping functionality to events framework

### LINKED ISSUE #339: Migrate no-activity pings to use events system

proteusvacuum:
Contains the commits from https://github.com/dimagi/open-chat-studio/pull/348, the commits from this PR start at "[Add some tests and functionality from no activity pings](https://github.com/dimagi/open-chat-studio/commit/26c7058df87bc50fb95f354df94ae02c770d101f)"

This adds the functionality from No Activity Pings using the events framework. 
I also re-wrote the TimeoutTrigger tests using `freeze_time`, as it made it much clearer what was happening. 

The one check that TimeoutTriggers don't currently do is to check whether the last message was from the bot. I was under the impression that the bot will always have the last word, but I'd be happy to add that in here.

One thing to do here is to improve the UI around the action parameters, as we've mentioned before.

[Screencast from 2024-04-02 22-30-20.webm](https://github.com/dimagi/open-chat-studio/assets/146896/90c65651-274c-499f-a7a5-93da4b04689c)

(Note the demo is a bit contrived, because we actually check for timeouts every 10 seconds, which is why it isn't triggering right away). 

Fixes #339


apps/events/tests/test_timeout_trigger.py
```diff
@@ -204,3 +226,67 @@ def test_new_human_message_resets_count(session):
...
+        (SessionStatus.SETUP, True),
+        (SessionStatus.PENDING, True),
+        (SessionStatus.PENDING_PRE_SURVEY, True),
+        (SessionStatus.ACTIVE, True),
+        (SessionStatus.PENDING_REVIEW, False),
+        (SessionStatus.COMPLETE, False),
+        (SessionStatus.UNKNOWN, False),
+    },
+)
+def test_not_triggered_complete_chats(status, matches, session):
```
SmittieC, line 244: Nit: `test_not_triggered_for_complete_chats`

apps/events/models.py
```diff
@@ -10,9 +10,11 @@
 from apps.chat.models import ChatMessage, ChatMessageType
 from apps.events.actions import end_conversation, log, summarize_conversation
 from apps.experiments.models import Experiment, ExperimentSession
+from apps.experiments.models import Experiment, ExperimentSession, SessionStatus
 from apps.utils.models import BaseModel
 
 ACTION_FUNCTIONS = {"log": log, "end_conversation": end_conversation, "summarize": summarize_conversation}
+STATUSES_FOR_COMPLETE_CHATS = [SessionStatus.PENDING_REVIEW, SessionStatus.COMPLETE, SessionStatus.UNKNOWN]
```
SmittieC, line 15: In [chat/tasks.py](https://github.com/dimagi/open-chat-studio/blob/9976a4d49bdb4925711de57417a543332790a628/apps/chat/tasks.py#L22) there is also a `STATUSES_FOR_COMPLETE_CHATS` Just wondering if we should read from there (or a more central place)?

apps/events/models.py
```diff
@@ -8,19 +8,25 @@
...
+    "send_message_to_bot": send_message_to_bot,
+    "summarize": summarize_conversation,
+}
+
 
 class EventActionType(models.TextChoices):
     LOG = ("log", "Log the last message")
     END_CONVERSATION = ("end_conversation", "End the conversation")
     SUMMARIZE = ("summarize", "Summarize the conversation")
+    SEND_MESSAGE_TO_BOT = ("send_message_to_bot", "Send a message to the bot")
```
snopoke, line 29: I'm a bit unsure about this name. What about "Prompt the bot to message the user"?

snopoke (approved):

apps/events/models.py
proteusvacuum: I think we might remove it from there if we are going to remove the no-activity-ping functionality, which is why I copied it here.

apps/events/models.py
proteusvacuum: Sure! You can send the bot any message, which is why I named it this way, but happy to change it.

snopoke: Sure you can send any message to the bot but the end result is that the response from the bot get's sent to the user right?

PR #350, opened 2024-04-03
Update `new_bot_message` implementation + new test

SmittieC:
I noticed that the `new_bot_message` mimicks what the `send_text_to_user` method does. This comes from early days when we didn't support differnet message types etc. What this method should do, is call `_send_message_to_user` which will take the bot message and, depending on some config, send either a text or voice message to the user. This is what I did. Also, each channel doesn't have to implement this method anymore, since it will call `_send_message_to_user` which will call implemented methods ultimately (see the `test_new_bot_message` test case).

Another notable change: A test that will catch new channel type implementations that is missing from the `ChannelBase.from_experiment_session` method. The `FacebookMessengerChannel` type was, so this came in handy.

### DIFF
--- a/apps/channels/tests/test_base_channel_behavior.py
+++ b/apps/channels/tests/test_base_channel_behavior.py
@@ -7,8 +7,8 @@ from unittest.mock import Mock, PropertyMock, patch
 
 import pytest
 
-from apps.channels.models import ExperimentChannel
-from apps.chat.channels import TelegramChannel
+from apps.channels.models import ChannelPlatform, ExperimentChannel
+from apps.chat.channels import ChannelBase, TelegramChannel
 from apps.chat.models import ChatMessageType
 from apps.experiments.models import ExperimentSession, SessionStatus, VoiceResponseBehaviours
 from apps.utils.factories.channels import ExperimentChannelFactory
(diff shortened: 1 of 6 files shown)


apps/channels/tests/test_base_channel_behavior.py
```diff
@@ -354,3 +354,19 @@ def test_user_query_extracted_for_pre_conversation_flow(
                 _get_voice_transcript.assert_called()
             elif message_type == "text":
                 message_text_mock.assert_called()
+
+
+@pytest.mark.django_db()
+@pytest.mark.parametrize("platform", [platform for platform, _ in ChannelPlatform.choices])
+def test_all_channels_can_be_instantiated_from_a_session(platform, twilio_provider):
```
snopoke, line 361: 👍🏻

apps/chat/channels.py
```diff
@@ -103,6 +103,8 @@ def initialize(self):
 
     @property
     def chat_id(self) -> int:
+        if self.experiment_session and self.experiment_session.external_chat_id:
+            return self.experiment_session.external_chat_id
```
snopoke, line 107: what's the advantage of doing this?

SmittieC: So we can instantiate a channel (TelegramChannel, WhatsappChannel etc) in two ways:
1. WhatsappChannel(...)
2. ChannelBase.from_experiment_session(experiment_session)

Consider the second case: When the channel is freshly instantiated, `channel.message` will be `None`, so if we try to send a message to the user, then calling `channel.chat_id` will cause an error, because it calls `get_chat_id_from_message`, which tries to get the `chat_id` from the user's message, but we don't have that. We do have the experiment session which have this info though, so we might as well look there first.

You might be wondering "what about the first case then?". It is perfectly valid to do something like `WhatsappChannel(experiment_channel=experiment_channel)` where `experiment_channel` will be without an experiment session. This is in fact the case when a user first sends a message. In this case, we'll fall back to the user message to get the chat id.

If we want to send a message to the user and use `WhatsappChannel(experiment_channel=experiment_channel)`, then it will fail, since then it wouldn't be able to determine the `chat_id`. We then have to pass in an `experiment_session` instance as well.

I hope I explained this correctly.

P.S. The reason why this is change is made only now is because the `new_bot_message` method was implemented by each channel and it looked at `self.experiment_session.external_chat_id` to determine the `chat_id`.

snopoke (approved):

PR #353, opened 2024-04-03
Use tripe quotes rather than multiline string for row classes

SmittieC:
The multiline string didn't work properly when running `npm run build`. Triple quotes seem to be doing just fine though

### DIFF
--- a/gpt_playground/settings.py
+++ b/gpt_playground/settings.py
@@ -475,10 +475,10 @@ DJANGO_TABLES2_TABLE_ATTRS = {
 }
 # text-neutral text-error (pipeline run table styles here for tailwind build)
 DJANGO_TABLES2_ROW_ATTRS = {
-    "class": (
-        "border-b border-gray-200 hover:bg-gray-100"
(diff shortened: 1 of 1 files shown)


snopoke (approved):

proteusvacuum:
I think this is because you are missing a space in the multiline case. You could have also just added a space at the end of the first string:
```
>>> x = (
... "foo"
... "bar"
... )
>>> x
'foobar'
>>> x = """
... foo
... bar
... """
>>> x
'\nfoo\nbar\n'
>>> 
```

SmittieC:
> I think this is because you are missing a space in the multiline case. You could have also just added a space at the end of the first string:
> 
> ```
> >>> x = (
> ... "foo"
> ... "bar"
> ... )
> >>> x
> 'foobar'
> >>> x = """
> ... foo
> ... bar
> ... """
> >>> x
> '\nfoo\nbar\n'
> >>> 
> ```

Good catch! I think I'll leave it as-is though, it looks better with triple quotes (at least to me).

PR #355, opened 2024-04-04
Create tags on the fly (and other fixes)

SmittieC:
Users can now create tags on-the-fly from within the multiselect widget. I went this route instead of navigating to the Tags page, because I think this will make for a better user experience.

### Fixes include:
- Prevent an experiment from being deleted when source_material or surveys are removed
- Better to see this one (I wanted to delete the tag, but it took me to the edit page).

#### The issue (that has been fixed)
https://github.com/dimagi/open-chat-studio/assets/64970009/702a1055-ef52-480e-aa1b-a2a87e068289


snopoke (approved):

snopoke (commented):
Will this still show the tags even if the user can't add / link new tags?

snopoke:
> > Will this still show the tags even if the user can't add / link new tags?
> 
> Yes and no. Let me explain: If the user cannot add/link tags, then it doesn't make sense to show the `tag_multiselect.js` widget to them in the session view. If they have view permissions, they'll be able to see it in the session list view though (but this is unrelated to the changes I made in the PR. That multiselect widget gets included [here](https://github.com/dimagi/open-chat-studio/blob/87fccdc7dd9866f9edfc536ba5a42901d4755ffb/templates/experiments/single_experiment_home.html#L167C1-L176C22). At the bottom of that file there's a custom `TomSelect` instance, so it's not using the `tag_multiselect.js` script).

OK thanks for the explanation. Just to confirm, if they don't have the 'add' permissions they won't be able to see existing tags that have been linked in the detail view.

SmittieC:
> > > Will this still show the tags even if the user can't add / link new tags?
> > 
> > 
> > Yes and no. Let me explain: If the user cannot add/link tags, then it doesn't make sense to show the `tag_multiselect.js` widget to them in the session view. If they have view permissions, they'll be able to see it in the session list view though (but this is unrelated to the changes I made in the PR. That multiselect widget gets included [here](https://github.com/dimagi/open-chat-studio/blob/87fccdc7dd9866f9edfc536ba5a42901d4755ffb/templates/experiments/single_experiment_home.html#L167C1-L176C22). At the bottom of that file there's a custom `TomSelect` instance, so it's not using the `tag_multiselect.js` script).
> 
> OK thanks for the explanation. Just to confirm, if they don't have the 'add' permissions they won't be able to see existing tags that have been linked in the detail view.

Correct. They will see neither the multiselect widget, nor the tags that was added by other users....do we want them to see tags in the detail view? We probably want them see see it I'd think.

snopoke:
> Correct. They will see neither the multiselect widget, nor the tags that was added by other users....do we want them to see tags in the detail view? We probably want them see see it I'd think.

Yes, I think if the user has the 'view' permission they should be able to see the tags that have been assigned.

SmittieC:
> > Correct. They will see neither the multiselect widget, nor the tags that was added by other users....do we want them to see tags in the detail view? We probably want them see see it I'd think.
> 
> Yes, I think if the user has the 'view' permission they should be able to see the tags that have been assigned.

Ok cool. Forgot to add viewing ability here. I'll add it

PR #356, opened 2024-04-05
Show tags when user has view-only permissions

SmittieC:
Updated the chat detail view to show tags if the user has view-only permissions

Screenshot from 2024-04-05 08-27-47


snopoke (approved):
Great!

Issue #354, opened 2024-04-03
Add better UI for updating Event Action parameters 


proteusvacuum:
Currently this is a json blob. It should instead show widgets based on what parameters are available to each action.

PR #359, opened 2024-04-10
Add better UI for updating Event Action parameters

proteusvacuum:
This adds a better UI to the event creation page using separate forms for the different actions.

Fixes https://github.com/dimagi/open-chat-studio/issues/354

[Screencast from 2024-04-09 11:20:43 PM.webm](https://github.com/dimagi/open-chat-studio/assets/146896/01c206fe-1d3b-4986-9f0a-1df7d16fa8d5)

### DIFF
--- a/apps/events/forms.py
+++ b/apps/events/forms.py
@@ -1,4 +1,5 @@
 from django import forms
+from langchain.memory.prompt import SUMMARY_PROMPT
 
 from apps.generics.type_select_form import TypeSelectForm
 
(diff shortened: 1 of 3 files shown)


snopoke (approved):
Looks great. How did you find it working with TypeSelectForms?

proteusvacuum:
It was pretty good! The only headscratcher was adding the `secondary_key` to the context. 

I think down the line a good refactor of this code would be to somehow define the `params` alongside the function, and have the form be created automatically. But I figured that was some YAGNI for now as adding functions probably won't happen so often.

PR #363, opened 2024-04-11
Support participant details in prompt

SmittieC:
[Ticket](https://github.com/orgs/dimagi/projects/3?pane=issue&itemId=59131446)

Best to review this commit-wise

We now support including participant details in the prompt (if there are any). The `external_chat_id` and `user` was also moved from the `ExperimentSession` model to the `Participant` model.

This PR also includes a data migration to create participant records for all experiment session records where they are missing, and update those participant records exists with the necessary details. Note the following:
- Participant records were only created for webchannel chats (not for chats through external chats like telegram)
- If an OCS user specified an email in a public link, the experiment session's user gets populated with the OCS user and a participant gets created with identifier = specified email
- An OCS user could specify either their own email (so participating as themselves) or another email (impersonating an external user)
- External channels (i.e. Telegram) will not have a participant nor user on the experiment session

### Diagram explaining the logic that the data migration follows
image


apps/experiments/views/experiment.py
```diff
@@ -528,19 +528,32 @@ def update_delete_channel(request, team_slug: str, experiment_id: int, channel_i
...
+    with transaction.atomic():
+        session = ExperimentSession.objects.create(
+            team=experiment.team,
+            experiment=experiment,
+            llm=experiment.llm,
+            experiment_channel=experiment_channel,
+            status=SessionStatus.ACTIVE,
+        )
+
+        if not participant_identifier and not participant_user:
```
snopoke, line 543: do this check before creating the session

apps/experiments/views/experiment.py
```diff
@@ -528,19 +528,32 @@ def update_delete_channel(request, team_slug: str, experiment_id: int, channel_i
...
+            raise Exception("Either participant_identifier or participant_user must be specified!")
```
snopoke, line 544: Maybe use `ValueError` here

apps/experiments/views/experiment.py
```diff
@@ -528,19 +528,32 @@ def update_delete_channel(request, team_slug: str, experiment_id: int, channel_i
...
-        team=experiment.team,
-        user=user,
-        participant=participant,
-        experiment=experiment,
-        llm=experiment.llm,
-        external_chat_id=external_chat_id,
-        experiment_channel=experiment_channel,
-    )
+    with transaction.atomic():
+        session = ExperimentSession.objects.create(
```
snopoke, line 535: if you move this to after the participant is created then you don't need to do an additional save.

apps/experiments/views/experiment.py
```diff
@@ -528,19 +528,32 @@ def update_delete_channel(request, team_slug: str, experiment_id: int, channel_i
 def _start_experiment_session(
     experiment: Experiment,
     experiment_channel: ExperimentChannel,
-    user: CustomUser | None = None,
-    participant: Participant | None = None,
-    external_chat_id: str | None = None,
+    participant_user: CustomUser | None = None,
+    participant_identifier: str = "",
```
snopoke, line 532: Isn't this always required?

apps/channels/tests/test_base_channel_behavior.py
```diff
@@ -115,8 +116,10 @@ def test_different_sessions_created_for_different_users(_get_llm_response, teleg
     experiment_sessions_count = ExperimentSession.objects.count()
     assert experiment_sessions_count == 2
 
-    assert ExperimentSession.objects.filter(external_chat_id=user_1_chat_id).exists()
-    assert ExperimentSession.objects.filter(external_chat_id=user_2_chat_id).exists()
+    assert ExperimentSession.objects.filter(participant__external_chat_id=user_1_chat_id).exists()
+    assert Participant.objects.get(external_chat_id=user_1_chat_id)
```
snopoke, line 120: This 2nd assert is necessary since the one above would fail if the participant didn't exist

apps/channels/tests/test_base_channel_behavior.py
```diff
@@ -42,7 +42,7 @@ def test_incoming_message_adds_channel_info(_get_llm_response, _send_text_to_use
     _simulate_user_message(telegram_channel, message)
 
     experiment_session = ExperimentSession.objects.filter(
-        experiment=telegram_channel.experiment, external_chat_id=chat_id
+        experiment=telegram_channel.experiment, participant__external_chat_id=chat_id
```
snopoke, line 45: what do you think about making a method on the manager for this filter since it's quite common and long:

`ExperimentSession.objects.for_external_id(chat_id, experiment)`

apps/chat/channels.py
```diff
@@ -429,13 +429,20 @@ def _reset_session(self):
         self._create_new_experiment_session()
 
     def _create_new_experiment_session(self):
+        """Creates a new experiment session. If one already exists, the participant will be transfered to the new
+        session
+        """
+        if not self.experiment_session:
+            participant = Participant.objects.create(
```
snopoke, line 436: Shouldn't this use `get_or_create`?

apps/experiments/migrations/0076_create_participants.py
```diff
@@ -0,0 +1,85 @@
...
+    Participant.objects.filter(experimentsession__isnull=True).all().delete()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('experiments', '0075_alter_participant_unique_together_and_more'),
+    ]
+
+    operations = [migrations.RunPython(_create_participants, migrations.RunPython.noop),]
```
snopoke, line 88: nit: data migrations should be marked as `elidable`
```suggestion
    operations = [migrations.RunPython(_create_participants, migrations.RunPython.noop, elidable=True),]
```

apps/experiments/models.py
```diff
@@ -398,13 +402,21 @@ def get_llm_service(self):
...
+            return None
+
     def get_absolute_url(self):
         return reverse("experiments:single_experiment_home", args=[self.team.slug, self.id])
 
 
 class Participant(BaseTeamModel):
     identifier = models.CharField(max_length=320, blank=True)  # max email length
     public_id = models.UUIDField(default=uuid.uuid4, unique=True)
+    external_chat_id = models.CharField(null=False)
```
snopoke, line 418: How is this different from `identifier`? It seems like they are always assigned the same value. Maybe we can drop one?

apps/experiments/models.py
```diff
@@ -464,11 +493,8 @@ class Meta:
 
     def save(self, *args, **kwargs):
         if not hasattr(self, "chat"):
-            self.chat = Chat.objects.create(team=self.team, user=self.user, name=self.experiment.name)
```
snopoke, line 467: does this mean we can drop the `user` field from the Chat model?

apps/experiments/views/experiment.py
```diff
@@ -724,26 +739,32 @@ def experiment_invitations(request, team_slug: str, experiment_id: str):
...
+                    channel = _ensure_experiment_channel_exists(experiment, platform="web", name=f"{experiment.id}-web")
+                    # TODO: Use _start_experiment_session and pass in specific kwargs
+                    session = ExperimentSession.objects.create(
+                        team=request.team,
+                        experiment=experiment,
+                        llm=experiment.llm,
+                        status="setup",
+                        experiment_channel=channel,
+                    )
+                    participant, _ = Participant.objects.get_or_create(
```
snopoke, line 761: move this above the session create so that you don't have to re-save the session after

snopoke (commented):
It would be great to merge the migrations

apps/experiments/models.py
SmittieC: Correct. It was ultimately just a duplicate of the experiment session's user, which moved to participant and will always be there. I'll do this in a followup PR

apps/experiments/models.py
SmittieC: You are correct, they are the same. I was thinking of dropping one of these in a followup PR, but maybe only after the data migration was run and things have "settled". It feels like I don't want to lose this data just yet. Hmm, on second thought, if we make a DB dump before the migration it should be fine. (to be clear, I have confidence that the data migration will work, I just don't want to make too many changes at once)

apps/chat/channels.py
SmittieC: Good catch. I'll add a test for this case and update the code

apps/channels/tests/test_base_channel_behavior.py
SmittieC: Not a bad idea :+1:

apps/experiments/views/experiment.py
SmittieC: Correct. I'll update

apps/experiments/views/experiment.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/363/commits/e9032d56272b5862ba32d30d73a0adb459ed138d

SmittieC: The test: https://github.com/dimagi/open-chat-studio/pull/363/commits/0e77e1442bc31d120eaa5559527538e2084eb40e

apps/experiments/views/experiment.py
```diff
@@ -751,20 +749,19 @@ def experiment_invitations(request, team_slug: str, experiment_id: str):
                 with transaction.atomic():
                     channel = _ensure_experiment_channel_exists(experiment, platform="web", name=f"{experiment.id}-web")
                     # TODO: Use _start_experiment_session and pass in specific kwargs
+                    participant, _ = Participant.objects.get_or_create(
+                        team=request.team,
+                        external_chat_id=post_form.cleaned_data["email"],
+                        identifier=post_form.cleaned_data["email"],
```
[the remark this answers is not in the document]
SmittieC: The [email is required](https://github.com/dimagi/open-chat-studio/blob/0e77e1442bc31d120eaa5559527538e2084eb40e/apps/experiments/forms.py#L60) for this form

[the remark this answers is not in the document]
SmittieC: Although this made me wonder what I did for the `start_session_public` method. It's not necessarily required there. I'll make an update. Thanks for raising

apps/experiments/tests/test_views.py
```diff
@@ -252,3 +253,39 @@ def test_new_participant_created_for_different_teams(_trigger_mock, is_user):
     # There should be two participants with external_chat_id = identifier accross all teams
     assert Participant.objects.filter(external_chat_id=identifier).count() == 2
     assert session.participant.external_chat_id == identifier
+
+
+@pytest.mark.django_db()
+@mock.patch("apps.experiments.views.experiment.enqueue_static_triggers")
+def test_participant_gets_user_when_they_signed_up(_trigger_mock, client):
```
SmittieC, line 260: I made an oopsy with my commits. [This test](https://github.com/dimagi/open-chat-studio/pull/363/commits/2a26b181931560770905221b5161b8639cae427c) should have been included in this commit. This test you see here should be in [this commit](https://github.com/dimagi/open-chat-studio/pull/363/commits/23e5611f97ee507b613fecbd7a3f5c5258b65b62)

snopoke (approved):

apps/experiments/views/experiment.py
```diff
@@ -217,11 +218,11 @@ def save(self, commit=True):
 
 def _validate_prompt_variables(form_data):
     required_variables = set(PromptTemplate.from_template(form_data.get("prompt_text")).input_variables)
-    available_variables = set()
+    available_variables = set(["participant_details"])
```
snopoke, line 221: @bderenzi since we're about to introduce this as a variable that can be included in a prompt, do you want to suggest any alternative names? (we can add an alias later if necessary but easier to change it now).

During our call you mentioned that the term you'd been using was `user profile`. Should we go with that?

apps/experiments/migrations/0075_alter_participant_unique_together_and_more.py
```diff
@@ -0,0 +1,35 @@
...
+        migrations.AddField(
+            model_name='participant',
+            name='user',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to=settings.AUTH_USER_MODEL),
+        ),
+        migrations.AlterUniqueTogether(
+            name='participant',
+            unique_together={('team', 'identifier'), ('team', 'external_chat_id')},
+        ),
+    ]
```
snopoke, line 35: can merge these operations into the previous migration (or re-generate the previous migration). I was going to suggest using https://docs.djangoproject.com/en/5.0/ref/django-admin/#cmdoption-makemigrations-update but it only works with the latest migration.

apps/experiments/migrations/0077_alter_participant_external_chat_id.py
```diff
@@ -0,0 +1,18 @@
...
+        ('experiments', '0076_create_participants'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='participant',
+            name='external_chat_id',
+            field=models.CharField(),
+        ),
+    ]
```
snopoke, line 18: this can be merged into 0075

apps/experiments/migrations/0078_remove_experimentsession_external_chat_id_and_more.py
```diff
@@ -0,0 +1,21 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name='experimentsession',
+            name='external_chat_id',
+        ),
+        migrations.RemoveField(
+            model_name='experimentsession',
+            name='user',
+        ),
```
snopoke, line 20: also merge these into 0075

apps/experiments/models.py
```diff
@@ -450,7 +485,7 @@ class ExperimentSession(BaseTeamModel):
         max_length=40, blank=True, default="", help_text="System ID of the seed message task, if present."
     )
     no_activity_ping_count = models.IntegerField(default=0, null=False, blank=False)
-    external_chat_id = models.CharField(null=False)
+    # external_chat_id = models.CharField(null=False)
```
snopoke, line 488: nit: delete

apps/experiments/views/experiment.py
```diff
@@ -724,26 +753,31 @@ def experiment_invitations(request, team_slug: str, experiment_id: str):
...
-                    team=request.team,
-                    experiment=experiment,
-                    llm=experiment.llm,
-                    status="setup",
-                    participant=participant,
-                    experiment_channel=channel,
-                )
+                with transaction.atomic():
+                    channel = _ensure_experiment_channel_exists(experiment, platform="web", name=f"{experiment.id}-web")
+                    # TODO: Use _start_experiment_session and pass in specific kwargs
```
snopoke, line 767: is this a TODO for this PR?

snopoke (changes requested):

apps/experiments/migrations/0077_alter_participant_external_chat_id.py
SmittieC: Ah, yes.

apps/experiments/views/experiment.py
SmittieC: No. I don't want to change too many things in this PR. Thanks for pointing it out, I'll create an issue for this (should be a quick win)

apps/experiments/migrations/0077_alter_participant_external_chat_id.py
SmittieC: When I put the schema + data changes together, I get
```
django.db.utils.OperationalError: cannot ALTER TABLE "experiments_participant" because it has pending trigger events
```

when it tries to alter the `external_chat_id` field on `Participant` to be non-nullable. According to a few places I saw (including [this one](https://stackoverflow.com/questions/12838111/django-db-migrations-cannot-alter-table-because-it-has-pending-trigger-events)). I suspect this is because everything is run in a transaction, so somehow it's not detecting the changes to the `external_chat_id` field, so it flags this as being null when we try to set it to not-null.

Keeping the schema and data migration out of each other's hair works just fine, so I'll do that.

apps/experiments/migrations/0075_alter_participant_unique_together_and_more.py
snopoke: @SmittieC I think this can still be done which would reduce the migrations from 4 to 3. A workflow that I've used in the past is to undo the migrations and then have django re-make them:

```
# roll back to 73
./manage.py migrate experiments 0073

# remove the 2 migrations that need to be squashed
rm apps/experiments/migrations/0074*.py apps/experiments/migrations/0075*.py

# backup the later ones
mv apps/experiments/migrations/0076*.py apps/experiments/migrations/0077*.py apps/experiments/

# run django migrations to recreate initial model migrations
./manage.py makemigrations experiments

# move others back and rename
mv apps/experiments/0076_create_participants.py apps/experiments/migrations/0075_create_participants.py
mv apps/experiments/0077_create_participants.py apps/experiments/migrations/0076_create_participants.py

# edit the migrations to update their dependencies
```

SmittieC: Ah yes, there is the 0074 migration as well, thanks. I just copied the operations and the migration's dependencies over to the previous one. I missed [this](https://github.com/dimagi/open-chat-studio/pull/363/commits/37f819dbab90964e6a0bd7e815007b546920cc51) one in [this](https://github.com/dimagi/open-chat-studio/pull/363/commits/cfbbc0ca2e81714bb2b51070ed67acc5a8f1faf0) commit.

SmittieC: Hmm, actually I don't feel right about this move. Let me just do it the long way

apps/experiments/views/experiment.py
bderenzi: No strong feelings. Both `{participant_details}` and `{user_profile}` are 4 tokens each in the OpenAI world. In my brain, `user_profile` makes a little more sense. Or maybe even `participant_profile`. I think the world is pretty used to the idea of a profile being properties about an individual. 

I suspect in a lot of cases, we'll want to pull out specific details to use them in the prompt.

For fun, I asked ChatGPT a handful of times. It seemed to gravitate to `user_info`, `user_profile`, `user_context`, and `user_data`  as being simple, concise, and versatile options. Take that as you will--the robots voted strongly for starting with `user`. Of these options, `user_info` and `user_data` stand out to me as being pretty clear. Happy to defer to you all (though can pick if you want a tie breaker)

SmittieC: Thanks for going through all the trouble @bderenzi :smile:  `user_data` sounds good. I'll make an update

bderenzi: What's the confusion? You mean that `user` might be the prompt author instead of the chat participant? I don't think people would be confused. I also think we'll have to manually specify the properties that we want to track in `user_data` somewhere as well. 

Also, I'm happy with `participant_data` if that seems more clear.

bderenzi: Is now a reasonable time to add the alias and use both `user_data` and `participant_data`?

SmittieC: I think let's keep the alias for a separate PR. I want to make an update some time to show the user the available variables with a mini description. I'll ship the alias changes with that if we still want one. Questions on the alias: Should this be user defined, or do we want to have this statically defined?

bderenzi: I was thinking statically defined.

SmittieC: Oh ok then it should be a quick one

apps/chat/tasks.py
```diff
@@ -22,6 +22,30 @@
...
+
+    return (
+        ExperimentSession.objects.filter(id__in=Subquery(latest_participant_session_ids))
+        .prefetch_related("experiment")
+        .all()
+    )
+
+
 @shared_task
 def send_bot_message_to_users(message: str, chat_ids: list[str], is_bot_instruction: bool, experiment_public_id: UUID):
```
snopoke, line 50: probably for a separate PR but why does this need to support multiple chat IDs? As far as I can tell we only lever pass in one

snopoke (dismissed):

apps/chat/tasks.py
SmittieC: So this is task was originally built to support both tools and be used for manual scheduled messages. There are one or two cases where multiple chat IDs were used. This will be made obsolete by the scheduled message changes

snopoke (approved):

PR #364, opened 2024-04-11
Show user details in session detail view

SmittieC:
### Data collapsed
image

### Data shown
Screenshot from 2024-04-11 14-51-57

### Future improvements
it will be nice if we don't just dump the data, but format it a bit to look a bit nicer

### DIFF
--- a/templates/experiments/experiment_session_view.html
+++ b/templates/experiments/experiment_session_view.html
@@ -41,15 +41,13 @@
       {% include "experiments/components/experiment_details.html" %}
     </div>
     {% if experiment_session.get_participant_data %}
-      <div x-data="{showParticipantData: false}">
-        <h3 class="text-base font-semibold leading-7 mt-8 pl-4">
-          <div class="hover:cursor-pointer" @click="showParticipantData = !showParticipantData;">
-            <i x-show="!showParticipantData" class="fa fa-arrow-right" aria-hidden="true"></i>
-            <i x-cloak x-show="showParticipantData" class="fa fa-arrow-down" aria-hidden="true"></i>
-            Participant Data
-          </div>
-        </h3>
-        <div x-cloak x-show="showParticipantData" class="max-w-5xl mt-4 rounded-2xl">
+
+      <div class="app-card collapse collapse-arrow max-w-5xl">
+        <input type="checkbox" />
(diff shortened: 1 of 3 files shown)


snopoke (approved):

templates/experiments/experiment_session_view.html
```diff
@@ -40,6 +40,20 @@ <h3 class="text-base font-semibold leading-7 mt-8 pl-4">Experiment details</h3>
...
+      <div x-data="{showParticipantData: false}">
+        <h3 class="text-base font-semibold leading-7 mt-8 pl-4">
+          <div class="hover:cursor-pointer" @click="showParticipantData = !showParticipantData;">
+            <i x-show="!showParticipantData" class="fa fa-arrow-right" aria-hidden="true"></i>
+            <i x-cloak x-show="showParticipantData" class="fa fa-arrow-down" aria-hidden="true"></i>
+            Participant Data
+          </div>
+        </h3>
+        <div x-cloak x-show="showParticipantData" class="max-w-5xl mt-4 rounded-2xl">
+          {{ experiment_session.get_participant_data }}
```
snopoke, line 53: Suggestion:

Make this a tab UI with 2 tabs:

1. Key value pair list view
  This is the 'nice' view that shows each top level key in the data with it's associated value in a tabular style.
2. Raw view
  Show the formatted JSON data in a `<pre>` section.

SmittieC: Formatted:

formatted

Raw:

image

snopoke: nice!

templates/experiments/experiment_session_view.html
```diff
@@ -40,6 +40,20 @@ <h3 class="text-base font-semibold leading-7 mt-8 pl-4">Experiment details</h3>
     <div class="max-w-5xl mt-4 border rounded-2xl">
       {% include "experiments/components/experiment_details.html" %}
     </div>
+    {% if experiment_session.get_participant_data %}
+      <div x-data="{showParticipantData: false}">
```
snopoke, line 44: just wondering if an Accordion widget would work instead of using alpine: https://daisyui.com/components/accordion/

snopoke (dismissed):

snopoke (approved):

Issue #362, opened 2024-04-10
Upgrade langchain


snopoke:
Why?
* Stay on a recent version
* New Anthropic messaging API
* Anthropic tool support

and more

SmittieC:
Langchain is already updated and [here's](https://github.com/dimagi/open-chat-studio/pull/370) the anthropic tool support changes

bderenzi:
Great! Do we need to make a separate issue for the anthropic messaging API? I think that's still not working (so no claude-3 on OCS at the moment)

SmittieC:
The messaging API issue is fixed by us using the `ChatAnthoripic` model from langchain_anthropic now instead of the one from langchain_community. This reminds me, only the claude3 models supports tool usage

PR #368, opened 2024-04-15
Update langchain + langchain-openai packages

SmittieC:
For #362. For anthropic tool support, I think we'll have to install `langchain-anthropic`

### DIFF
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -26,6 +26,7 @@ tenacity
 whitenoise[brotli]
 langchain
 langchain-openai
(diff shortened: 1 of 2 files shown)


snopoke:
Have you been able to test this at all>

SmittieC:
> Have you been able to test this at all>

Yeah I tested it again. FYI: I see two of the anthorpic models doesn't work (on prod also). We can maybe tackle that with the anthropic tool change

snopoke (approved):

PR #373, opened 2024-04-22
Scheduled Messages

SmittieC:
These changes lays the infrastructure for schedules messages. In a followup PR I'll stick a UI on this.

[Spec](https://docs.google.com/document/d/1elyKzxkTAgH_yS_Nt5Vo6vUqBo_SZtkCh49l91HUa6k/edit)

Added two models
- ScheduledMessageConfig
- ScheduledMessage

Note: These changes only include event based schedule message creation, meaning that the experiment creator will have to specify on which event a scheduled message should be created. They will not be able to create messages manually (for the foreseeable future)

Still TODO, but in a followup PR:
- Create scheduled message when the `PARTICIPANT_JOINED_EXPERIMENT` event is fired.
- Add this event to all experiments

Note: This branch's changes starts at ([Add ScheduledMessageConfig and ScheduledMessage models and migration](https://github.com/dimagi/open-chat-studio/pull/373/commits/b0d0149f35ce464d2550f19c9f1173cf79011adf)). The commits before this looks like commits that's on main, but I changed the base to the `cs/user_details` branch. Anycase, once `cs/user_details` is merged, I suspect this should be fixed, otherwise I'll rebase/merge again


apps/chat/models.py
```diff
@@ -104,3 +113,107 @@ def _get_langchain_dict(self, content, message_type):
...
+                    When(total_triggers__gte=self.repetitions, then=True),
+                    output_field=models.BooleanField(),
+                ),
+                next_trigger_date=Case(
+                    When(last_triggered_at__isnull=True, then=F("created_at") + F("new_delta")),
+                    When(last_triggered_at__isnull=False, then=F("last_triggered_at") + F("new_delta")),
+                    output_field=DateTimeField(),
+                ),
+            )
+        )
```
snopoke, line 180: neat query!

apps/chat/models.py
```diff
@@ -104,3 +113,107 @@ def _get_langchain_dict(self, content, message_type):
...
+
+class ScheduledMessage(BaseTeamModel):
+    schedule = models.ForeignKey(ScheduledMessageConfig, on_delete=models.CASCADE, related_name="scheduled_messages")
+    participant = models.ForeignKey(
+        "experiments.Participant", on_delete=models.CASCADE, related_name="schduled_messages"
+    )
+    next_trigger_date = models.DateTimeField(null=True)
+    last_triggered_at = models.DateTimeField(null=True)
+    total_triggers = models.IntegerField(default=0)
+    resolved = models.BooleanField(default=False)
```
snopoke, line 191: wondering about renaming this to 'complete'

apps/chat/tests/test_scheduled_messages.py
```diff
@@ -0,0 +1,320 @@
...
+            expected_next_trigger_date=current_time - relativedelta(seconds=seconds_offset),
+            expected_last_triggered_at=current_time,
+            expected_total_triggers=2,
+            expected_resolved=True,
+            expect_next_trigger_date_changed=False,
+        )
+
+        # We are done now, but let's make 110% sure that another message will not be triggered
+        current_time = step_time(frozen_time, db_time, step_delta)
+        scheduled_message._trigger = Mock()
```
snopoke, line 180: I don't think this will work since the mock is on a different instance of the object than the one which will be returned from the DB in a query.

Maybe check that `_get_messages_to_fire` doesn't return any?

apps/chat/tests/test_scheduled_messages.py
```diff
@@ -0,0 +1,320 @@
...
+        utc_now = datetime.now().astimezone(pytz.timezone("UTC"))
+        db_time.return_value = utc_now + relativedelta(days=1.1)
+
+        pending_messages = _get_messages_to_fire()
+        assert len(pending_messages) == 1
+
+        poll_scheduled_messages()
+        assert len(caplog.records) == 1
+        for record in caplog.records:
+            record
```
snopoke, line 217: this doesn't seem to be doing anything

apps/chat/tests/test_scheduled_messages.py
SmittieC: Yeah you're right, not sure what I was thinking here tbh. Checking `_get_messages_to_fire` sounds better

apps/chat/models.py
SmittieC: Hmm..agreed, but maybe `is_complete` is a bit clearer than `complete`

apps/chat/tests/test_scheduled_messages.py
SmittieC: :man_facepalming: right, I was busy checking logs and probably got distracted. Thankss :sweat_smile:

apps/chat/tests/test_scheduled_messages.py
SmittieC: Changes are made [here](https://github.com/dimagi/open-chat-studio/pull/373/commits/6996e196dc707b432776eb8dc8b6312948cd14a4)

apps/chat/tests/test_scheduled_messages.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/373/commits/6996e196dc707b432776eb8dc8b6312948cd14a4

apps/chat/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/373/commits/6996e196dc707b432776eb8dc8b6312948cd14a4

apps/chat/models.py
```diff
@@ -104,3 +113,107 @@ def _get_langchain_dict(self, content, message_type):
...
+
+    def save(self, *args, **kwargs):
+        if not self.next_trigger_date:
+            delta = relativedelta(**{self.schedule.time_period: self.schedule.frequency})
+            utc_now = datetime.now().astimezone(pytz.timezone("UTC"))
```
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/373/commits/4b9f2715c0c56c8bf357347ae8b31c49aebc8ef1

proteusvacuum:
Probably a bit late for this question, but did you consider adding this to the `events` app? An alternate way of implementing this could have been to add a `ScheduleTrigger`, that would be created when the events mentioned in the doc are fired. You would then also have "for free" the UI, logging, and access to the other actions when the schedule is reached.

SmittieC:
> Probably a bit late for this question, but did you consider adding this to the `events` app? An alternate way of implementing this could have been to add a `ScheduleTrigger`, that would be created when the events mentioned in the doc are fired. You would then also have "for free" the UI, logging, and access to the other actions when the schedule is reached.

@proteusvacuum Is my understanding of what you're suggesting correct? -> The user goes to the events page, they choose some event, then for the "Then..." field, they choose `ScheduleTrigger` which then allows them to create a schedule?

proteusvacuum:
@SmittieC Yes, that's what I was imagining! There might be constraints that you have in your spec that doesn't allow you to do that, but I was just curious.

SmittieC:
> @SmittieC Yes, that's what I was imagining! There might be constraints that you have in your spec that doesn't allow you to do that, but I was just curious.

I must admit that I have not thought of this approach. Let me entertain this a bit...

apps/events/models.py
```diff
@@ -78,13 +103,19 @@ def trigger_type(self):
...
             self.event_logs.create(session=session, status=EventLogStatusChoices.SUCCESS, log=result)
             return result
         except Exception as e:
             logging.error(e)
             self.event_logs.create(session=session, status=EventLogStatusChoices.FAILURE, log=str(e))
 
+    @transaction.atomic()
+    def delete(self, *args, **kwargs):
+        result = super().delete(*args, **kwargs)
+        self.action.delete(*args, **kwargs)
```
snopoke, line 116: is this necessary for the `event_action_deleted` to be called?

apps/utils/django_db.py
```diff
@@ -11,7 +11,7 @@ class MakeInterval(models.Func):
     arg_joiner = " => "
 
     def __init__(self, unit, value):
-        assert unit in ("years", "months", "days", "hours", "mins", "secs")
+        assert unit in ("years", "months", "weeks", "days", "hours", "mins", "secs")
```
snopoke, line 14: 👍🏻

apps/events/actions.py
```diff
@@ -1,47 +1,118 @@
...
+                is_complete=Case(
+                    When(total_triggers__lt=action.params["repetitions"], then=False),
+                    When(total_triggers__gte=action.params["repetitions"], then=True),
+                    output_field=BooleanField(),
+                ),
+                next_trigger_date=Case(
+                    When(last_triggered_at__isnull=True, then=F("created_at") + F("new_delta")),
+                    When(last_triggered_at__isnull=False, then=F("last_triggered_at") + F("new_delta")),
+                    output_field=DateTimeField(),
+                ),
```
snopoke, line 95: I'm still a bit torn on this. I think it could create weird effects if we updated schedules that are already finished. e.g. schedules that have been completed for a "long time".

snopoke (commented):
One main comment but otherwise looks good

apps/events/models.py
SmittieC: Are you asking if it's needed to delete the `action` in order to call `event_action_deleted`? If so, not really no, but actions are not deleted when its `StaticTrigger` is deleted, which doesn't seem correct, since it has a 1-1 relationship with it. I hope this answers your question?

snopoke: Ah I see. One option would be to move the field so that it's on the action and not the trigger. You could still then use `trigger.action` but deleting a trigger would also delete the action.

Alternatively a `post_delete` signal could be used which also get's fired for queryset deleted.

I'd actually be in favour of changing the relationship since actions can't exist without triggers (but fine to be a separate PR).

apps/events/actions.py
SmittieC: You think we should rather just updated those that are not completed yet? We can always later on allow the user to choose the bahaviour if that's required

apps/events/models.py
SmittieC: I just realized that since scheduled messages has an action FK, they will be removed whenever we remove the action. I think this cleanup function was leftover code from the previous iteration. Anycase, see [this](https://github.com/dimagi/open-chat-studio/pull/373/commits/7443a18165d018dd662bad3c7e82e0b814e020d7)

snopoke: OK, but what about the action? Will the action get deleted when deleting the trigger?

SmittieC: Yes, since we still keep the delete method on the static trigger that deletes the action

apps/events/actions.py
SmittieC: Cool, let's do that then. I'll update

SmittieC: https://github.com/dimagi/open-chat-studio/pull/373/commits/344dec64f1362d18fe50305c590c5c478ab10976

snopoke (approved):

PR #376, opened 2024-04-25
Add another voice provider

SmittieC:
I first added a relation to `File` on the `VoiceProvider` model, but later switched to adding a single FK to it on the `SyntheticVoice` model. File uploads are not being done using a MultipleFileUpload widget, but rather by including the`file_formset` that we use in other places. We control whether or not to show/include this by the "allow_file_upload" attribute.


apps/service_providers/models.py
```diff
@@ -164,6 +180,79 @@ def get_speech_service(self) -> speech_service.SpeechService:
...
+                string = ", ".join(invalid_extentions)
+                raise UserServiceProviderConfigError(f"File extentions not supported: {string}")
+
+    @transaction.atomic()
+    def add_files(self, files):
+        self.validate_uploaded_file_names(files)
+        if self.type == VoiceProviderType.openai_voice_engine:
+            for file in files:
+                try:
+                    # TODO: Split file extention
```
SmittieC, line 202: Not for this PR

apps/service_providers/models.py
```diff
@@ -164,6 +180,79 @@ def get_speech_service(self) -> speech_service.SpeechService:
...
+        if self.type == VoiceProviderType.openai_voice_engine:
+            accepted_file_types = ["mp4", "mp3"]
+            invalid_extentions = set()
+            for file in uploaded_files:
+                file_extention = file.name.split(".")[1]
+                if file_extention not in accepted_file_types:
+                    invalid_extentions.add(f".{file_extention}")
+            if invalid_extentions:
+                string = ", ".join(invalid_extentions)
+                raise UserServiceProviderConfigError(f"File extentions not supported: {string}")
```
snopoke, line 194: I think you could use a standard Django `ValidationError` here.

Also, what about doing this validation during form validation instead of on save? You could extend the `BaseFileFormSet` and update the `get_file_formset` to take in a form class.

apps/experiments/views/experiment.py
```diff
@@ -332,14 +338,22 @@ def get_initial(self):
...
         "voice_providers_types": dict(request.team.voiceprovider_set.values_list("id", "type")),
         "synthetic_voice_options": sorted(
             [
-                {"value": voice.id, "text": str(voice), "type": voice.service.lower()}
-                for voice in SyntheticVoice.objects.all()
+                {
+                    "value": voice.id,
+                    "text": str(voice),
+                    "type": voice.service.lower(),
+                    "provider_id": voice.voice_provider.id if voice.voice_provider else None,
```
snopoke, line 354: nit: use _id field instead of doing related query:
```suggestion
                    "provider_id": voice.voice_provider_id,
```

Explanation: every foreign key field on a model exposes a `{field_name}_id` property which is the ID of the related object (or None if it's not set).

apps/service_providers/models.py
```diff
@@ -164,6 +180,79 @@ def get_speech_service(self) -> speech_service.SpeechService:
...
+
+    def remove_file(self, file_id: int):
+        synthetic_voice = self.syntheticvoice_set.get(file_id=file_id)
+        synthetic_voice.file.delete()
+        synthetic_voice.delete()
+
+    def get_files(self):
+        """Return the files found on the synthetic voices that points to this instance"""
+        files = []
+        for voice in self.syntheticvoice_set.all():
```
snopoke, line 225: what about adding a filter for file:
```suggestion
        for voice in self.syntheticvoice_set.filter(file__isnull=False).all():
```

apps/service_providers/models.py
```diff
@@ -164,6 +180,79 @@ def get_speech_service(self) -> speech_service.SpeechService:
...
+                "pk": self.id,
+            },
+        )
+
+    @transaction.atomic()
+    def delete(self):
+        if self.type == VoiceProviderType.openai_voice_engine:
+            files_to_delete = self.get_files()
+            [f.delete() for f in files_to_delete]
+            self.syntheticvoice_set.all().delete(audit_action=AuditAction.AUDIT)
```
snopoke, line 253: does this not happen automatically when the provider is deleted (via the cascade delete)?

snopoke (approved):
Generally looks good. Some comments but nothing blocking

apps/experiments/models.py
```diff
@@ -210,15 +214,19 @@ class SyntheticVoice(BaseModel):
     )
 
     gender = models.CharField(
-        null=False, blank=False, choices=GENDERS, max_length=14, help_text="The gender of this voice"
+        null=False, blank=True, choices=GENDERS, max_length=14, help_text="The gender of this voice"
     )
     service = models.CharField(
-        null=False, blank=False, choices=SERVICES, max_length=6, help_text="The service this voice is from"
+        null=False, blank=False, choices=SERVICES, max_length=17, help_text="The service this voice is from"
```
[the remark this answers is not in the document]
SmittieC: Not really no

apps/service_providers/models.py
SmittieC: Yeah you're right, thanks for pointing this out

apps/service_providers/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/376/commits/a1562a726747ab7fcbe5940b61719dabd03618f4

apps/experiments/models.py
```diff
@@ -252,7 +252,7 @@ def get_for_team(team: Team, exclude_services=None) -> list["SyntheticVoice"]:
         """Returns a queryset for this team comprising of all general synthetic voice records and those exclusive
         to this team. Any services specified by `exclude_services` will be excluded from the final result"""
         exclude_services = exclude_services or []
-        general_services = ~Q(service__in=SyntheticVoice.TEAM_SCOPED_SERVICES)
+        general_services = ~Q(service__in=SyntheticVoice.TEAM_SCOPED_SERVICES, voice_provider__isnull=False)
```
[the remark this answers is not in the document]
SmittieC: Goodness, you're right. The voice_provider__is_null=True must be constant, not changing along with the not. Well spotted :+1:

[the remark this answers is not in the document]
SmittieC: Extended the tests to cover this case (although unlikely to happen): https://github.com/dimagi/open-chat-studio/pull/376/commits/8c7b2beb1b6a3878ce4bb20d133dce2f024a6356

snopoke (approved):

PR #379, opened 2024-04-30
Use the last chat message's created_at date as the "last message" date

SmittieC:


snopoke (approved):
The only issue here is that we're doing one query per item in the table. A more performant way to do it would be to modify the queryset by annotating it with the last message date (See https://github.com/dimagi/open-chat-studio/blob/92bd1b58c97e32bfe9c6ba68dfd09bdb7f5226ab/apps/events/models.py#L114C9-L159C10 for an example)

SmittieC:
> The only issue here is that we're doing one query per item in the table. A more performant way to do it would be to modify the queryset by annotating it with the last message date (See https://github.com/dimagi/open-chat-studio/blob/92bd1b58c97e32bfe9c6ba68dfd09bdb7f5226ab/apps/events/models.py#L114C9-L159C10 for an example)

Good idea. See the latest commit

apps/experiments/views/experiment.py
```diff
@@ -392,7 +399,16 @@ class DeleteFileFromExperiment(BaseDeleteFileView):
...
+    last_message_created_at = (
+        ChatMessage.objects.filter(
+            chat__experiment_session=OuterRef("pk"),
+        )
+        .order_by("-created_at")
+        .values("created_at")[:1]
+    )
+    user_sessions = ExperimentSession.objects.annotate(
+        last_message_created_at=Subquery(last_message_created_at)
+    ).filter(
```
snopoke, line 411: nit: maybe worth making this a manager method to avoid the copy paste

snopoke (approved):

apps/experiments/views/experiment.py
SmittieC: Cool. Will throw this in with a followup PR. There are some other housekeeping things I want to do

PR #384, opened 2024-05-08
API based conversations + participant details updates

SmittieC:
Added two api endpoints. One can use these endpoints by creating an api key for you user + team. The key is scoped to a specific team, so a user should not be able to use one key across teams.

Endpoints:
1. List experiments
3. Upsert participant data. This will be used when we integrate CommCare.

Further, I added an `ApiChannel` to allow users with api keys to chat to a bot using the API (see the API README on how to use this). **Please note** This is a synchronous/blocking call, so the client must wait for the response. Response times aren't typically very long, so it should be fine for our purposes (for now). Also, the endpoint to send new messages sits in the `channels` app with all the other channels' endpoints.
Q: Why is this endpoint authed?
A: We don't want users of the endpoint to be able to guess participant identifiers and thereby impersonate other people, potentially accessing their conversation history, so we essentially authorize the user if they have an api key. If we restrict participant identifiers to being uuid's for api based conversations, then it would make it harder for people to guess other participant's ids but since this feature is (for now) only for internal use, I think we should keep it as-is. Only platform users should be able to use the API.

Q: Why does the `update_participant_data` endpoint require a UserApi key? I'm open to suggestions here. I was thinking that we can add an additional `TeamApiKey` model which external systems like CommCare can use, but bt using a `UserApiKey` we are able to link a specific user for any audit trails, which is nice.


apps/api/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from . import views
+
+app_name = "api"
+
+urlpatterns = [
+    path("experiments/", views.get_experiments, name="list-experiments"),
+    path("participants/experiments/data", views.update_participant_data, name="update-participant-data"),
```
snopoke, line 9: what about making this more RESTy?
```
POST /participants/<id>/<experiment-id>/
    body: {"name": "John"}

OR

POST /participants/<id>/
    body: {"<experiment-id>": {"name": "John"}}
    OR 
    body: {"data": [{"experiment_id": "<experiment-id>", "name": "John"}]}

```

apps/api/views.py
```diff
@@ -0,0 +1,45 @@
...
+    experiment_id = serializers.UUIDField(source="public_id")
+
+
+@api_view(["GET"])
+@permission_classes([HasUserAPIKey])
+def get_experiments(request):
+    data = []
+    for experiment in Experiment.objects.filter(team__slug=request.team.slug).all():
+        data.append(ExperimentSerializer(experiment).data)
+    return Response(data=data)
```
snopoke, line 22: I think we should generally use pagination even in cases where it may not seem necessary so that we get consistent responses: https://www.django-rest-framework.org/api-guide/pagination/#pagination

If you still want to use function based views you can do this: https://stackoverflow.com/a/34043578

I would also vote for using cursor based pagination though I could be persuaded: https://www.django-rest-framework.org/api-guide/pagination/#cursorpagination

apps/api/views.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@api_view(["POST"])
+@permission_classes([HasUserAPIKey])
+def update_participant_data(request):
```
snopoke, line 27: It would be nice to have a read API for this as well so that you could do partial updated (read + modify + write)

apps/chat/channels.py
```diff
@@ -622,3 +624,25 @@ def send_voice_to_user(self, synthetic_voice: SynthesizedAudio):
...
+    @property
+    def message_content_type(self):
+        return MESSAGE_TYPES.TEXT
+
+    @property
+    def message_text(self):
+        return self.message.message_text
+
+    def new_bot_message(self, bot_message: str):
+        # Simply adding a new AI message to the chat history will cause it to be sent to the UI
```
snopoke, line 647: nit: update comment

apps/experiments/views/experiment.py
```diff
@@ -116,12 +116,16 @@ def get_queryset(self):
 
 
 class ExperimentForm(forms.ModelForm):
+    PROMPT_HELP_TEXT = """
+        Use {source_material} to place source material in the prompt. Use {participant_data} to place participant
+        data in the prompt.
+    """
```
snopoke, line 122: nit: this could be put with the other help texts in `Meta.help_texts` (unless that doesn't work for some reason)

apps/api/views.py
```diff
@@ -0,0 +1,45 @@
...
+from apps.experiments.models import Experiment, Participant, ParticipantData
+
+
+class ExperimentSerializer(serializers.Serializer):
+    name = serializers.CharField()
+    experiment_id = serializers.UUIDField(source="public_id")
+
+
+@api_view(["GET"])
+@permission_classes([HasUserAPIKey])
```
snopoke, line 17: we should also check the django model permissions e.g. `experiments.view_experiment`

apps/api/urls.py
SmittieC: I like the second one. That way we can bulk update participant data for various experiments

apps/experiments/views/experiment.py
SmittieC: oh right, that's a thing. Thanks

SmittieC: Hmm, it does seem to not pick it up from Meta if we declare the field as an attribute

apps/api/urls.py
```diff
@@ -6,5 +6,5 @@
 
 urlpatterns = [
     path("experiments/", views.get_experiments, name="list-experiments"),
-    path("participants/experiments/data", views.update_participant_data, name="update-participant-data"),
+    path("participants/<uuid:participant_id>", views.update_participant_data, name="update-participant-data"),
```
snopoke, line 9: this won't always be a UUID will it?

apps/api/views.py
```diff
@@ -24,22 +26,30 @@ def get_experiments(request):
...
-
-    try:
-        participant_data = experiment.participant_data.get(participant=participant)
-        participant_data.data = new_data
-        participant_data.save(update_fields=["data"])
-    except ParticipantData.DoesNotExist:
-        ParticipantData.objects.create(
-            participant=participant, data=new_data, content_object=experiment, team=experiment.team
-        )
+    experiment_data = request.data["data"]
```
snopoke, line 36: why do we need to wrap this in `data`?

apps/api/views.py
```diff
@@ -24,22 +26,30 @@ def get_experiments(request):
...
+        experiment = experiment_map[experiment_id]
+
+        try:
+            participant_data = experiment.participant_data.get(participant=participant)
+            participant_data.data = new_data
+            participant_data.save(update_fields=["data"])
+        except ParticipantData.DoesNotExist:
+            ParticipantData.objects.create(
+                participant=participant, data=new_data, content_object=experiment, team=experiment.team
+            )
```
snopoke, line 57: I think you could replace this with `ParticipantData.objects.update_or_create`

apps/api/views.py
SmittieC: Preference I guess. I can change it if that will make things easier

apps/api/urls.py
SmittieC: Oh sorry forgot to update this name. This is now the participant's `public_id`, which is always a UUID, buuuuuut I'm now remembering that CommCare wouldn't know the participant's public id if we consider the workflow for SUDCare

apps/api/views.py
snopoke: I think if the input is already a dict then there's no need to wrap it. If we were using a list then I think wrapping is better.

apps/api/urls.py
snopoke: CommCare 'could' know the UUID if it's returned when CC creates the participant (assuming that's what happens). Otherwise we could accept either or require users to do a 'search' using the identifier and then use the public ID for writes.

SmittieC: Yeah depends on the final workflow, but if I remember correctly we'll require the participant to make first contact, so we wouldn't need to "register" a participant in the traditional sense (that is, CC makes a request to OCS which creates the participant in the DB). My understanding of this registration is more to do with adding data to their record (for now)

snopoke: Either way, I think we can leave this as the public ID for now. What do you think of adding the the participant public ID and experiment session public ID to the message response?

Otherwise we'd need to provide an API to access the participant public ID for a user (thinking of Brian doing testing).

SmittieC: I think if we're going to include those details, let's maybe instead have a GET "participants/<identifier>/info" endpoint. It feels too random to include it in the response of the update data call. We can then use this to get various info + stats (eventually, maybe) of the participant. So CC would be able to use this to get the `public_id` and in turn use that for the update_data call

apps/api/views.py
SmittieC: In normal circumstances that would have worked, but since the `data` field is encrypted in the DB, searching on that field doesn't play nice

apps/api/views.py
```diff
@@ -58,4 +56,7 @@ def update_participant_data(request, participant_id: UUID):
             ParticipantData.objects.create(
                 participant=participant, data=new_data, content_object=experiment, team=experiment.team
             )
-    return HttpResponse()
+    response_body = ""
+    if experiments_not_updated:
+        response_body = {"unsuccessful_updates": experiments_not_updated}
+    return Response(data=response_body)
```
SmittieC, line 62: I'm open so suggestions here

apps/api/urls.py
SmittieC: Ok so I sort of went the other way and just allowed the user to specify the participant identifier in the url. No need to do an additional request to get the public id. We _can_ still add an endnpoint to get participant data (latest session id, public_id etc), but since we don't really have a usecase for that right now, I think let's maybe leave it until we need it? Sorry for the back-and-forth on the public_id/identifier.

apps/api/views.py
snopoke: I don't quite follow. We don't need to search on data do we?:
```python
ParticipantData.objects.update_or_create(
    participant=participant, 
    content_object=experiment,
    defaults={"data": new_data},
    create_defaults={"team": experiment.team},
)```

Comment (SmittieC) [reply apps/api/views.py]: Sorry I meant the update query, but looks like your solution worked, since we don't have to actually search the data as well (except the usage of `create_defaults`, which was only introduced in django 5 I think. Django 4.2 doesn't have that, but it's fine)

File: apps/api/views.py (hunk 10 above)
Comment (SmittieC) [reply apps/api/views.py:61]: Yeah I was also looking into various others, but none of them "felt right". If we're returning a 422, then we must not update anything in my opinion, otherwise this will be a very interesting side effect. I like your suggestion, it simulates the [207 multi-status approach](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/207)

Comment (snopoke) [reply apps/api/views.py:61]: I think fine to exit with an error and not update anything if there are experiment IDs that are incorrect. 

I think it's easier to reason about.

Comment (SmittieC) [reply apps/api/views.py:61]: Cool cool cool

File: apps/api/views.py [hunk 11]
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Upsert participant data for all specified experiments in the payload
+    """
+    experiment_data = request.data
+    experiment_ids = experiment_data.keys()
+    experiments = Experiment.objects.filter(public_id__in=experiment_ids, team=request.team)
+    experiment_map = {str(experiment.public_id): experiment for experiment in experiments}
+    participant = get_object_or_404(Participant, identifier=participant_id, team=request.team)
+
+    missing_ids = [exp_id for exp_id in experiment_data if exp_id not in experiment_map]
```
snopoke, line 42: OR:

```suggestion
    missing_ids = set(experiment_ids) - set(experiment_map)
```

snopoke (approved):

snopoke (approved):

snopoke:
@SmittieC can you write up some quick docs on these APIs please

PR #385, opened 2024-05-08
only perform table row action if click on table element

snopoke:
This allows buttons in the row to work


SmittieC (approved):
Where are we not using a table?

snopoke:
> Where are we not using a table?

We are, but this event handler also receives events when a button in the table row is clicked.

PR #386, opened 2024-05-08
Sk/assistants v2

### LINKED ISSUE #375: Upgrade OCS to Support Assistants API v2 for Vector Store Retrieval
**Description**
Our current OCS platform does not support the new vector store retrieval syntax introduced in Assistants API v2, which is already being utilized in OpenAI's Playground. This limitation prevents the effective use of enhanced retrieval features and may affect the consistency and performance of assistants imported from the Playground to OCS. Although newly created Assistants can be switched back to a legacy v1 mode, once this is done they are no longer editable (e.g. no changes to system message, file uploads, or changes to any other features).

**Problem**

- Assistants that utilize vector store retrieval in the Playground are not fully functional when imported into OCS, as the platform does not currently support the v2 API syntax. Specifically, it appears the name of the tool changing from 'retrieval' to 'file_search'.

- This leads to errors and reduced functionality, impacting user experience in OCS.

**Expected Behavior**

- OCS support of current features of the Assistants API v2, including vector store retrieval.

- Assistants using these features in the Playground should maintain functionality when imported to OCS.

**Suggested Actions**

- API Upgrade: Implement full support for the Assistants API v2 in OCS, focusing on compatibility with the vector store retrieval feature.

snopoke:
resolves #375 

This ended up being quite a lot of effort.

(Applogies for the poor commit history - there was a fair amount of churn)


apps/assistants/tests/test_migration_v2.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+
+from apps.assistants.migrations.utils import migrate_assistant_to_v2
+from apps.utils.factories.assistants import OpenAiAssistantFactory
+from apps.utils.factories.files import FileFactory
+
+
+@pytest.mark.django_db()
+def test_migrate_assistant():
```
SmittieC, line 9: nice

SmittieC (dismissed):
Looks good to me :+1:

SmittieC (approved):

PR #390, opened 2024-05-10
Sureadhere messaging provider integrated

cws1121:


bderenzi:
@cws1121 - can you run `ruff` on the code to sort the code-style at some point? (fine to wait for further review comments)

apps/channels/models.py
```diff
@@ -159,6 +167,8 @@ def webhook_url(self) -> str:
             uri = reverse("channels:new_twilio_message")
         elif provider_type == MessagingProviderType.turnio:
             uri = reverse("channels:new_turn_message", kwargs={"experiment_id": self.experiment.public_id})
+        elif provider_type == MessagingProviderType.sureadhere:
+            uri = reverse("channels:new_sureadhere_message", kwargs={"client_id": self.extra_data.get("client_id", "")})
```
snopoke, line 171: Why use the `client_id` here instead of the `channel.external_id`?

apps/channels/tasks.py
```diff
@@ -56,6 +57,21 @@ def handle_twilio_message(self, message_data: str):
     message_handler.new_user_message(message)
 
 
+@shared_task(bind=True, base=TaskbadgerTask)
+def handle_sureadhere_message(self, client_id: str, message_data: dict):
+    message = SureAdhereMessage.parse(message_data)
+    experiment_channel = ExperimentChannel.objects.filter(
+        extra_data__client_id=client_id,
+        platform=ChannelPlatform.IN_APP,
+        messaging_provider__type=MessagingProviderType.sureadhere,
```
snopoke, line 66: Surely it's not possible to create a channel with this platform and a different messaging provider type?

apps/channels/tasks.py
```diff
@@ -56,6 +57,21 @@ def handle_twilio_message(self, message_data: str):
...
+    ).first()
+    if not experiment_channel:
+        return
```
snopoke, line 69: I think we should at least log this at the `info` level.

apps/channels/tests/message_examples/sureadhere_messages.py
```diff
@@ -0,0 +1,6 @@
+def inbound_message():
+    return {"message_text": "Hi", "patient_id": 6225}
+
+
+def outbound_message():
+    return {"patient_Id": 6225, "message_Body": "Hello, I am a chatbot, how can I help?"}
```
snopoke, line 6: Just noting that it's a bit odd that the inbound / outbound fields are different:
```diff
- message_text
+ message_Body
- patient_id
+ patient_Id
```

apps/chat/channels.py
```diff
@@ -587,6 +589,34 @@ def send_voice_to_user(self, synthetic_voice: SynthesizedAudio):
...
+    def send_text_to_user(self, text: str):
+        to_patient = self.chat_id
+        self.messaging_service.send_text_message(text, to=to_patient, platform=ChannelPlatform.IN_APP)
+
+    def get_chat_id_from_message(self, message):
+        return message.chat_id
+
+    @property
+    def voice_replies_supported(self) -> bool:
+        return bool(settings.AWS_ACCESS_KEY_ID) and self.messaging_service.voice_replies_supported
```
snopoke, line 605: It seems like voice isn't supported so you could leave this property out since the default is `False`

apps/service_providers/messaging_service.py
```diff
@@ -132,3 +133,34 @@ def get_message_audio(self, message: TurnWhatsappMessage) -> BytesIO:
...
+            "client_id": self.client_id,
+            "client_secret": self.client_secret,
+            "scope": "https://sureadherelabs.onmicrosoft.com/auth_demo_api1/.default",
+        }
+        response = requests.post(auth_url, data=auth_data)
+        response.raise_for_status()
+        return response.json()["access_token"]
+
+    def send_text_message(self, message: str, to: str, platform: ChannelPlatform, from_: str = None):
+        access_token = self.get_access_token()
```
snopoke, line 161: Is there a way we could reuse these? How long do they last for?

apps/service_providers/messaging_service.py
```diff
@@ -132,3 +133,34 @@ def get_message_audio(self, message: TurnWhatsappMessage) -> BytesIO:
...
+
+class SureAdhereService(MessagingService):
+    _type: ClassVar[str] = "sureadhere"
+    supported_platforms: ClassVar[list] = [ChannelPlatform.IN_APP]
+    voice_replies_supported: ClassVar[bool] = False
+    supported_message_types = [MESSAGE_TYPES.TEXT]
+
+    client_id: str
+    client_secret: str
+    base_url: str
```
snopoke, line 146: Is this different for different integrations? Where would one get this from?

apps/channels/forms.py
```diff
@@ -71,6 +71,33 @@ def get_success_message(self, channel: ExperimentChannel):
         return f"Use the following URL when setting up the webhook: {channel.webhook_url}"
 
 
+class InAppChannelForm(ExtraFormBase):
+    client_id = forms.CharField(label="Client ID", max_length=100)
```
snopoke, line 75: Is this `client_id` different from the one on the `MessagingProvider`. If they are the same the it doesn't seem useful to have it here as well.

apps/channels/views.py
```diff
@@ -32,6 +32,13 @@ def new_twilio_message(request):
     return HttpResponse()
 
 
+@csrf_exempt
```
snopoke, line 35: limit to POST requests:

```suggestion
@csrf_exempt
@require_POST
```

apps/channels/forms.py
snopoke: Can you confirm this understanding:

* `MessagingProvider.client_id`: this is an Oauth client ID used to authenticate with SureAdhere
* `Channel.client_id`: this is a workspace ID in SureAdhere (is a workspace SureAdhere's 'tenent'?)

Is `Client ID` the right name for it then?

It's also not clear to me what this is used for since it isn't included in the requests to send messages so it appears to only be used for the webhook.

I think that leads to the question of how the webhooks in SA are configured. Is there documentation on that somewhere? How does SA know which webhook to call for a specific message?

apps/channels/datamodels.py
```diff
@@ -93,6 +93,31 @@ def parse(message_data: dict) -> "TwilioMessage":
...
+    """
+
+    chat_id: int
+    message_text: str | None
+    content_type: MESSAGE_TYPES | None = Field(default=MESSAGE_TYPES.TEXT)
+    content_type_unparsed: str | None = Field(default=None)
+
+    @field_validator("content_type", mode="before")
+    @classmethod
+    def determine_content_type(cls, value):
```
snopoke, line 108: Since `content_type` is always `TEXT` I think this can be removed along with `content_type_unparsed`.

apps/channels/tests/message_examples/sureadhere_messages.py
snopoke: No need to change it. I was just making sure this wasn't a typo or something.

apps/channels/views.py
snopoke: You still need to add this (you added it to the twilio endpoint but not to this one)

apps/service_providers/forms.py
```diff
@@ -152,6 +152,14 @@ class TurnIOMessagingConfigForm(ObfuscatingMixin, ProviderTypeConfigForm):
     auth_token = forms.CharField(label=_("Auth Token"))
 
 
+class SureAdhereMessagingConfigForm(ObfuscatingMixin, ProviderTypeConfigForm):
+    obfuscate_fields = ["client_secret"]
+
+    client_id = forms.CharField(label=_("Client ID"))
+    client_secret = forms.CharField(label=_("Client Secret"))
+    base_url = forms.CharField(label=_("Base URL"))
```
snopoke, line 160: This should use a URL field:
```suggestion
    base_url = forms.URLField(label=_("Base URL"), validators=[URLValidator(schemes=["https"]))
```

apps/channels/forms.py
```diff
@@ -71,6 +71,33 @@ def get_success_message(self, channel: ExperimentChannel):
...
+            kwargs["initial"] = initial
+
+        super().__init__(*args, **kwargs)
+        if not channel:
+            # We only show the webhook URL field when there is something to show
+            self.fields["webook_url"].widget = forms.HiddenInput()
+
+    def get_success_message(self, channel: ExperimentChannel):
+        """The message to be displayed when the channel is successfully linked"""
+        return f"Use the following URL when setting up the webhook: {channel.webhook_url}"
```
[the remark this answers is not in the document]
snopoke: > The get_success_message method, remains in the respective classes. Let me know if you think we should move it to superclass as well.

I think it should also move since at the moment all the messages are identical. If a class wants to change the message it can override that method.

apps/channels/models.py
```diff
@@ -22,22 +22,20 @@
...
 FACEBOOK = "facebook"
+IN_APP = "in_app"
 
 
 class ChannelPlatform(models.TextChoices):
     TELEGRAM = "telegram", "Telegram"
     WEB = "web", "Web"
     WHATSAPP = "whatsapp", "WhatsApp"
     FACEBOOK = "facebook", "Facebook"
+    IN_APP = "in_app", "In App"
```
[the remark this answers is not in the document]
snopoke: > While SureAdhere is one of the in-app messaging platforms we currently support, what if we decide to add another provider in the future that also offers in-app messaging capabilities?

I'm not sure I quite follow this. If SA adds support for another provider e.g. telegram then OCS will still want to use the SA channel and SA will decide how to deliver that message to the user based on what provider that user is using to communicate with. But OCS won't know about the different downstream providers will it?

apps/channels/models.py
snopoke: > However, I agree that using experiment_id

I actually suggested `channel.external_id` since that is the most direct lookup. Is there an advantage to using the experiment ID?

apps/channels/models.py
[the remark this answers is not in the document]
snopoke: I feel like we're missing eachother somewhere and maybe we should go to voice but I'll take one more go at explaining what I think.

When you say "in app", do you mean "communication through a mobile app"? Is that not what WhatsApp and Telegram are already?

In terms of this data model, a channel type should be 1-1 with how the user access it e.g. whatsapp vs telegram vs sureadhere

The `MessagingProvider` ties the channels to a way of sending and receiving messages and is 1-many since the same provider could facilitate communications on multiple channels and multiple providers can support the same channel.

If we take WhatsApp as an example, both Turn and Twilio provide WhatsApp messaging capabilities but which one the user chooses depends on which provider they are using.

If you still have doubts, let's schedule a call.

apps/channels/forms.py
snopoke: Typically webhooks are configured in the sending system so that it know where to send messages. There are 2 options:

1. A generic webhook where the request body contains some information that identifies where the message is coming from e.g. client_id
2. A webhook that contains some identifying info in it e.g. experiment_id or client_id

Either of these approaches are fine but both require some way to configure them in SA.

For now we can revert to using `client_id` since that's what allows you to test but I'd like to hear what the product team has to say.

apps/channels/models.py
```diff
@@ -84,7 +86,13 @@ def get_unfiltered_queryset(self):
...
     objects = ExperimentChannelObjectManager()
     RESET_COMMAND = "/reset"
-    PLATFORM = ((TELEGRAM, "Telegram"), (WEB, "Web"), (WHATSAPP, "WhatsApp"), (FACEBOOK, "Facebook"))
+    PLATFORM = (
+        (TELEGRAM, "Telegram"),
+        (WEB, "Web"),
+        (WHATSAPP, "WhatsApp"),
+        (FACEBOOK, "Facebook"),
+        (SUREADHERE, "SureAdhere"),
+    )
```
snopoke, line 95: this doesn't appear to be used anywhere so you could delete it

apps/channels/forms.py
```diff
@@ -71,28 +66,17 @@ def get_success_message(self, channel: ExperimentChannel):
...
-            self.fields["webook_url"].widget = forms.HiddenInput()
+class WhatsappChannelForm(WebhookUrlFormBase):
+    number = forms.CharField(label="Number", max_length=100)
 
-    def get_success_message(self, channel: ExperimentChannel):
-        """The message to be displayed when the channel is successfully linked"""
-        return f"Use the following URL when setting up the webhook: {channel.webhook_url}"
+
+class SureAdhereChannelForm(WebhookUrlFormBase):
+    client_id = forms.CharField(label="Client ID", max_length=100)
```
snopoke, line 78: Please add some help text here assist the user in understanding what this field is.

snopoke (commented):
Thanks for all the updates. This is very close now. I think there are only 2 outstanding things I'd like addressed before approval:

* Revert the webhook to use `client_id` so that you can continue testing
* Add help text to `SureAdhereChannelForm.client_id` indicating what it is

apps/channels/forms.py
SmittieC: What's the difference between this `client_id` field and the `client_id` field from the `SureAdhereService`? This client id distinguishes a SureAdhere "project" so to speak whereas the other client id is an Azure id, yes?

apps/channels/models.py
```diff
@@ -168,7 +161,10 @@ def webhook_url(self) -> str:
         elif provider_type == MessagingProviderType.turnio:
             uri = reverse("channels:new_turn_message", kwargs={"experiment_id": self.experiment.public_id})
         elif provider_type == MessagingProviderType.sureadhere:
-            uri = reverse("channels:new_sureadhere_message", kwargs={"channel_external_id": self.external_id})
+            uri = reverse(
+                "channels:new_sureadhere_message",
+                kwargs={"sureadhere_tenant_id": self.extra_data.get("sureadhere_tenant_id", "")},
+            )
```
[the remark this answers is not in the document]
snopoke: Offlined on this and the webhook calling code in SureAdhere will use a template URL something like `https://ocs-domain.org/channels/sureadhere/{tenant_id}/incoming_message` where the tenant ID will be determined the by the user sending the message.

snopoke (approved):

PR #393, opened 2024-05-15
experiment search improvements

snopoke:
* search by username
* sort by rank
* limit searches to > 2 characters


SmittieC (approved):
Nice

PR #396, opened 2024-05-15
Fix assistant migration

snopoke:
This fixes this sentry error: https://dimagi.sentry.io/share/issue/bfda7256a2f54b0a98a119158ba88c58/

And also another error that occurs when files (the model) have no file attachement.

I do not know why the sentry error happens but I assume it's something related to how Django loads models during migrations.

Edit: found the Django docs: https://docs.djangoproject.com/en/4.2/topics/migrations/#historical-models


bderenzi (approved):
looks good!

Issue #410, opened 2024-05-21
Update `OpenAIAssistantRunnable` to use assistant v2 style requests


snopoke:
The langchain runnable does not support the v2 assistant API: https://dimagi.sentry.io/issues/5381380186/?alert_rule_id=14063560&alert_type=issue&notification_uuid=d949a2b7-e43e-4f70-932d-ec98c8f9104f&project=4505001320316928&referrer=issue_alert-slack

That specific error is related to this code in langchain which compiles the `thread` data for the API call and is still using the old structure: https://github.com/langchain-ai/langchain/blob/54adcd9e828e24bb24b2055f410137aca6a12834/libs/langchain/langchain/agents/openai_assistant/base.py#L302-L310

(See https://platform.openai.com/docs/api-reference/runs/createThreadAndRun)

Ideally we would create a PR in langchain to fix the issue but due to the urgency we may also have to copy the code to OCS temporarily so that we can release the fix quickly.

(Side note: I've considered copying the code to OCS previously in order to add support for file uploads / downloads)

SmittieC:
Fixed

PR #412, opened 2024-05-21
use v2 attachments for assistants

snopoke:
https://dimagi.sentry.io/share/issue/2a1ec1a629db46e18b2650b8e159658c/

@SmittieC this is the same issue as #410 but in a different place in the code.


bderenzi (approved):
Thanks, Simon!

PR #413, opened 2024-05-21
update request auditor to record hijacked users

snopoke:
I haven't thought through if we want to keep [hijack](https://django-hijack.readthedocs.io/) in the long term but for now let's at least preserve the correct audit trail.

### DIFF
--- a/apps/audit/auditors.py
+++ b/apps/audit/auditors.py
@@ -1,0 +1,45 @@
+import logging
+
+from django.core.exceptions import ValidationError
+from field_audit.auditors import BaseAuditor
+from field_audit.models import USER_TYPE_REQUEST
+
+from apps.users.models import CustomUser
+
+log = logging.getLogger("audit")
+
+
+class RequestAuditor(BaseAuditor):
+    """Auditor class for getting users from authenticated requests."""
+
+    def change_context(self, request):
+        if request is None:
+            return None
+        if request.user.is_authenticated:
+            username = get_request_username(request)
+            context = {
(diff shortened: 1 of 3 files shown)


apps/audit/auditors.py
```diff
@@ -0,0 +1,45 @@
...
+    if hijack_history:
+        if username := _get_hijack_username(hijack_history):
+            return username
+
+    return request.user.username
+
+
+def _get_hijack_username(hijack_history):
+    try:
+        pk = CustomUser._meta.pk.to_python(hijack_history[-1])
```
SmittieC, line 42: Why this line?

SmittieC (approved):

apps/audit/auditors.py
snopoke: it's a 'safe' way to convert from the serialised value (string) to the python datatype (int)

SmittieC: Maybe I'm missing an obvious scenario here, but when would it be unsafe? If users tamper with the session?

snopoke: It shouldn't be unsafe but we have to convert str to int somehow and this uses the model fields conversion instead of just going `int(...)`. Either way we still have to handle the error case/

SmittieC: So is it to achieve the following?

image

snopoke: No not really since that already works, it's just a validation step to make sure the `pk` is what we expect. We could probably remove it and Django would throw another error if the value wasn't an integer.

PR #414, opened 2024-05-21
OpenAIAssistantRunnable fix

SmittieC:
The invoke code was copied from [here](https://github.com/langchain-ai/langchain/blob/54adcd9e828e24bb24b2055f410137aca6a12834/libs/langchain/langchain/agents/openai_assistant/base.py#L256) and instead of passing `file_ids`, we now pass in `attachments`

### DIFF
--- a/apps/service_providers/llm_service/main.py
+++ b/apps/service_providers/llm_service/main.py
@@ -2,16 +2,81 @@ from io import BytesIO
 from typing import ClassVar
 
 import pydantic
-from langchain.agents.openai_assistant import OpenAIAssistantRunnable
+from langchain.agents.openai_assistant import OpenAIAssistantRunnable as BrokenOpenAIAssistantRunnable
 from langchain.chat_models.base import BaseChatModel
 from langchain_anthropic import ChatAnthropic
-from langchain_core.callbacks import BaseCallbackHandler
+from langchain_core.callbacks import BaseCallbackHandler, CallbackManager
 from langchain_core.language_models import BaseLanguageModel
+from langchain_core.load import dumpd
+from langchain_core.runnables import RunnableConfig, ensure_config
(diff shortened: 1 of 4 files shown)


apps/service_providers/llm_service/main.py
```diff
@@ -2,18 +2,83 @@
...
+                tool_outputs = self._parse_intermediate_steps(input["intermediate_steps"])
+                run = self.client.beta.threads.runs.submit_tool_outputs(**tool_outputs)
+            # Starting a new thread and a new run.
+            elif "thread_id" not in input:
+                thread = {
+                    "messages": [
+                        {
+                            "role": "user",
+                            "content": input["content"],
+                            "attachments": input.get("attachments", {}),
```
SmittieC, line 47: previously this was `file_ids`

apps/service_providers/llm_service/runnables.py
```diff
@@ -280,6 +280,24 @@ def invoke(self, input: str, config: RunnableConfig | None = None) -> ChainOutpu
...
+                    file_resources[file.external_id].append(resource)
+                else:
+                    file_resources[file.external_id] = [resource.tool_type]
+
+        for file_id, resources in file_resources.items():
+            tools = [{"type": resource} for resource in resources]
+            attachment = {"file_id": file_id, "tools": tools}
+            attachments.append(attachment)
+
+        input_dict["attachments"] = attachments
```
snopoke, line 300: This isn't necessary. The attachments here are per-message attachments which we don't support.

snopoke (changes requested):
One comment but otherwise looks good.

apps/service_providers/llm_service/runnables.py
SmittieC: Of course, we don't have to pass it along with each message. I wasn't thinking straight. Thanks!

snopoke (approved):

PR #416, opened 2024-05-22
Support participant preferences in scheduled messages

SmittieC:
- Participants should now be able to change the weekday + time of their any scheduled message they have (this does not include PeriodicTasks. We should port those to use scheduled messages)
- The bot will "see" all scheduled messages for this participant in the current team, regardless of experiment, since we don't have any ties to the experiment from the ScheduledMessage. The only downside I can think of now is conflicting schedule names.
- Timezone support is not yet implemented. I think we'll get this info from the participant data (separate PR)
- I've found that the other tools sometimes interferes with this new one. For instance, sometimes it calls one of the other tools after this new one, and then things break. We need a way to allow tool groups or let the user select which tools they want to have enabled (PR [here](https://github.com/dimagi/open-chat-studio/pull/417))


apps/chat/tests/test_tools.py
```diff
@@ -0,0 +1,66 @@
...
+        ("2024-05-29 08:00:00", 4, 8, 0, "2024-05-31 08:00:00"),  # Wednesday to Friday
+        ("2024-06-01 08:00:00", 0, 8, 0, "2024-05-27 08:00:00"),  # Saturday to Monday
+        ("2024-06-02 08:00:00", 0, 8, 0, "2024-05-27 08:00:00"),  # Sunday to Monday
+        ("2024-06-02 08:00:00", 1, 8, 0, "2024-05-28 08:00:00"),  # Sunday to Tuesday
+    ],
+)
+def test_move_datetime_to_new_weekday_and_time(
+    initial_datetime_str, new_weekday, new_hour, new_minute, expected_new_datetime_str
+):
+    """Test weekday and time changes. A weekday change will cause the datetime to jump to a different week"""
```
SmittieC: I'm assuming behaviour here: Suppose it's Tuesday the second today and a scheduled messages is scheduled to go off on Friday but the user wants a monday schedule, then by changing it to mondays, the trigger date will shift to the Monday the 1st, effectively triggering the scheduled message. So basically: a preference update **should not delay a schedule** more than a week. If we allowed this, then a user can update their preferences in such a way that they'll never get the scheduled message.

apps/chat/agent/tools.py
```diff
@@ -81,6 +83,42 @@ def action(
...
+    def action(
+        self,
+        name: str,
+        weekday: schemas.WeekdaysEnum,
+        hour: int,
+        minute: int,
+        user_specified_custom_date: bool,
+    ):
+        if user_specified_custom_date:
+            return "The user cannot do that. Only weekdays and time of day can be changed"
```
snopoke, line 101: Why was this added? Is this to prevent users from trying to change it to some future or past date?

apps/utils/time.py
```diff
@@ -23,3 +23,8 @@ def seconds_to_human(value):
 def timedelta_to_relative_delta(timedelta: timedelta):
     """Converts a `timedelta` instance to a `relativedelta` instance"""
     return relativedelta(seconds=timedelta.total_seconds())
+
+
+def pretty_date(date: datetime) -> str:
+    """Returns the date like this: 'Monday, 1 January 2024 08:00:00 UTC'"""
+    return date.strftime("%A, %d %B %Y %H:%M:%S %Z")
```
snopoke, line 30: 🌷

apps/experiments/models.py
```diff
@@ -639,8 +640,32 @@ def send_bot_message(self, instruction_prompt: str, fail_silently=True):
         bot_message = bot_prompt_for_user(self, prompt_instruction=instruction_prompt)
         try_send_message(self, message=bot_message, fail_silently=fail_silently)
 
+    def get_participant_scheduled_messages(self):
+        from apps.events.models import ScheduledMessage
+
+        messages = ScheduledMessage.objects.filter(participant=self.participant, team=self.team)
```
snopoke, line 646: What about adding a link to the experiment to this model?

apps/experiments/models.py
```diff
@@ -639,8 +640,32 @@ def send_bot_message(self, instruction_prompt: str, fail_silently=True):
...
+            frequency = schedule_info["frequency"]
+            time_period = schedule_info["time_period"]
+            repetitions = schedule_info["repetitions"]
+            schedule = f"{name}: Every {frequency} {time_period}, {repetitions} times"
+            if time_period not in ["hour", "day"]:
+                weekday = message.next_trigger_date.strftime("%A")
+                schedule = f"{name}: Every {frequency} {time_period} on {weekday} for {repetitions} times"
+            next_trigger = pretty_date(message.next_trigger_date)
+            schedule = f"{schedule} (next trigger is {next_trigger})"
+            scheduled_messages_str = schedule if not scheduled_messages_str else f"{scheduled_messages_str},{schedule}"
```
snopoke, line 660: what about making this the `__str__` method of the class?

apps/experiments/models.py
snopoke: You should also prefetch the action:
```suggestion
        messages = ScheduledMessage.objects.filter(participant=self.participant, team=self.team).select_related("action")
```

apps/experiments/models.py
```diff
@@ -639,8 +640,32 @@ def send_bot_message(self, instruction_prompt: str, fail_silently=True):
...
     def get_participant_data(self):
-        return self.experiment.get_participant_data(self.participant)
+        scheduled_messages_str = self.get_participant_scheduled_messages()
+        participant_data = self.experiment.get_participant_data(self.participant)
+        return f"{participant_data}. Scheduled messages:\n{scheduled_messages_str}"
 
     def get_participant_data_json(self):
-        return json.dumps(self.experiment.get_participant_data(self.participant), indent=2)
+        data = self.experiment.get_participant_data(self.participant) or {}
+        data["schedules"] = self.get_participant_scheduled_messages()
```
snopoke, line 670: shouldn't this be a more JSONy format instead of the string format?

apps/events/models.py
```diff
@@ -232,6 +232,7 @@ class ScheduledMessage(BaseTeamModel):
     last_triggered_at = models.DateTimeField(null=True)
     total_triggers = models.IntegerField(default=0)
     is_complete = models.BooleanField(default=False)
+    extra_data = models.JSONField(blank=True, default=dict)
```
snopoke, line 235: I was going to suggest making this a cron expression for future flexibility but that seems overkill.

I do think it should be renamed to `custom_schedule_params` or something to give it more meaning.

apps/chat/agent/tools.py
SmittieC: Correct. I chose to do this so that users cannot change the date to some date far in the future if they find check-ins annoying

apps/events/models.py
SmittieC: This confirms it then. I was between having a dedicated field vs an `extra_data` field that we just happen to use for the preferred schedule. I'll update :+1:

apps/experiments/models.py
SmittieC: > What about adding a link to the experiment to this model?

Good idea. For the multi-bot stuff this will be necessary :+1:

apps/experiments/models.py
SmittieC: yeah I realized there was some work missing in this area. I made a small update here: https://github.com/dimagi/open-chat-studio/pull/416/commits/eef15a38223fb73a2ebe9a06783025bd06d34f49

SmittieC:
I'll do the timezone support in a separate PR

apps/chat/agent/tools.py
snopoke: 👍🏻 Could you add a comment to the code please.

apps/experiments/models.py
snopoke: I think a JSON object would be better because otherwise if anyone wants to do anything with the schedule information they have to parse it out of the strings.

apps/experiments/models.py
```diff
@@ -639,8 +639,27 @@ def send_bot_message(self, instruction_prompt: str, fail_silently=True):
...
+        """Returns all scheduled messages for the associated participant for this session's experiment as well as
+        any child experiments in the case where the experiment is a parent"""
+        from apps.events.models import ScheduledMessage
+
+        child_experiments = ExperimentRoute.objects.filter(team=self.team, parent=self.experiment).values("child")
+        messages = ScheduledMessage.objects.filter(
+            Q(experiment=self.experiment) | Q(experiment__in=models.Subquery(child_experiments)),
+            participant=self.participant,
+            team=self.team,
+        )
```
snopoke, line 652:
```suggestion
        ).select_related("action")
```

snopoke (commented):
Almost there!

apps/experiments/models.py
SmittieC: Lol, I find it funny to see this since I had this in! I was debugging a query I tried to build and removed this part since it cluttered the sql....I guess my mind went too far and forgot to clean up after itself :rofl:. Thanks for spotting this

apps/experiments/models.py
SmittieC: Cool cool

apps/chat/agent/tools.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/416/commits/9cd8f7a853f7bd23a94614ae5481432d2e8cf2cb

apps/experiments/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/416/commits/3c3b08c47bde95a61768738d1336833b3bd65c43

apps/experiments/models.py
```diff
@@ -639,8 +640,48 @@ def send_bot_message(self, instruction_prompt: str, fail_silently=True):
...
+        scheduled_messages = []
+        for message in messages:
+            if as_dict:
+                scheduled_messages.append(
+                    {
+                        "name": message.name,
+                        "frequency": message.frequency,
+                        "time_period": message.time_period,
+                        "repetitions": message.repetitions,
+                        "next_trigger_date": pretty_date(message.next_trigger_date),
```
snopoke, line 669: nit: this should be isoformat
```suggestion
                        "next_trigger_date": message.next_trigger_date.isoformat(),
```

SmittieC: I was gonna ask why, but then realized that in order to reasily work with it, it should be. Good catch

snopoke (approved):

PR #417, opened 2024-05-23
Ability to select which tool to use with your llm

SmittieC:
Resolves https://github.com/dimagi/open-chat-studio/issues/426
Added:
- A `AgentToolResource` table (I tried to follow the naming convention of assistant resources, but in this case specific to agents). For now, I think it's fine to not have them scoped to a specific team, since 1. tools are hardcoded and 2. we don't have a usecase for them warrant being team-scoped.
- Updated the styling for the multiselect dropdown to actually show an arrow (and one or two other small things)
- Data migration to create tool resources for the two pre-existing tools (Unless we want to throw them away, in which case we can just remove the data migration)
- Removed the `tools_enabled` column and added a property on the `Experiment` model instead. Tools will now be enabled by default if they are specified.

image


apps/chat/agent/tools.py
```diff
@@ -2,17 +2,24 @@
...
+from apps.experiments.models import AgentTools, ExperimentSession
 from apps.utils.time import pretty_date
 
 BOT_MESSAGE_FOR_USER_TASK = "apps.chat.tasks.send_bot_message_to_users"
 
+TOOL_CLASS_MAP = {
+    AgentTools.SCHEDULE_UPDATE: "apps.chat.agent.tools.UpdateScheduledMessageTool",
+    AgentTools.ONE_OFF_REMINDER: "apps.chat.agent.tools.OneOffReminderTool",
+    AgentTools.RECURRING_REMINDER: "apps.chat.agent.tools.RecurringReminderTool",
+}
```
snopoke, line 21: if you put this below the class definitions you could reference the classes directly instead of as strings

apps/experiments/models.py
```diff
@@ -288,6 +288,20 @@ class VoiceResponseBehaviours(models.TextChoices):
     NEVER = "never", gettext("Never")
 
 
+class AgentTools(models.TextChoices):
+    RECURRING_REMINDER = "recurring-reminder", gettext("Recurring Reminder")
+    ONE_OFF_REMINDER = "one-off-reminder", gettext("One-off Reminder")
+    SCHEDULE_UPDATE = "schedule-update", gettext("Schedule Update")
+
+
+class AgentToolResource(BaseModel):
```
snopoke, line 297: Is it necessary to have a new model? You could make the field on the experiment an ArrayField (e.g. https://github.com/dimagi/open-chat-studio/blob/f17abadda0a2b957065069aa674efc8cf2782e30/apps/assistants/models.py#L33)

assets/styles/site-tailwind.css
```diff
@@ -52,3 +52,40 @@
...
+  @apply rounded-b-lg;
+}
+
+.ts-dropdown .create:hover, .ts-dropdown .option:hover{
+  @apply bg-sky-50 text-sky-900;
+}
+
+.ts-dropdown .active {
+  @apply bg-gray-100 text-gray-900;
+}
```
snopoke, line 91: ❤️

apps/chat/agent/tools.py
SmittieC: I did have it there at first, but it just felt like a weird place to put it. I can't really express the reason I'm against putting it there. Maybe I should just get over it. Ok, I'll update

snopoke: I know the feeling

SmittieC: https://github.com/dimagi/open-chat-studio/pull/417/commits/75b76722b668f89955574e0008f59fe2be9458be

apps/experiments/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/417/commits/ab100e7fdfd262484b99080495928bfb2e5ad4c1

apps/experiments/views/experiment.py
```diff
@@ -329,6 +336,8 @@ def form_valid(self, form, file_formset):
         if file_formset:
             files = file_formset.save(self.request)
             self.object.files.set(files)
+
+        self.object.set_tools(tool_names=self.request.POST.getlist("tools"))
```
snopoke, line 340: you could just make the form do this for you: https://github.com/dimagi/open-chat-studio/blob/6d68e88421dd60235e88b46b9c0d87b48129712c/apps/assistants/forms.py#L9

Even if you stick with the select box you could still have a field in the form and it would persist it to the model (I think)

templates/experiments/experiment_form.html
```diff
@@ -74,7 +74,11 @@
 
       <input type="radio" name="main_tabs" role="tab" class="tab" aria-label="Tools">
       <div role="tabpanel" class="tab-content">
-        {% include "experiments/components/tools.html" %}
+        <div class="py-4 form-text">
+          <h3 class="text-md font-bold">{% translate "Please note" %} <i class="fas fa-exclamation-triangle"></i></h3>
+          <p class="text-sm">{% translate "Make sure that your selected model supports tool usage" %}</p>
+        </div>
+        {% render_form_fields form "tools" %}
```
SmittieC, line 81: I like the "vanilla" multiselect more in this case

image

snopoke (approved):

SmittieC:
I'll merge this when the base branch is merged

stephherbers (approved):

PR #420, opened 2024-05-23
Pipelines

proteusvacuum:
Spec: https://docs.google.com/document/d/1jJVmkj1AON-puttwQ12hZXEx3HcZuvBaDJxcaQ2fVwY/edit

[Screencast from 2024-05-30 23-29-54.webm](https://github.com/dimagi/open-chat-studio/assets/146896/d8907f51-291a-41bb-89cd-2aa5190d75e1)

image


apps/pipelines/models.py
```diff
@@ -9,3 +16,49 @@ class Pipeline(BaseTeamModel):
...
+
+class PipelineRunStatus(models.TextChoices):
+    RUNNING = "running", "Running"
+    SUCCESS = "success", "Success"
+    ERROR = "error", "Error"
+
+
+class PipelineRun(BaseModel):
+    pipeline = models.ForeignKey(Pipeline, on_delete=models.CASCADE, related_name="runs")
+    status = models.CharField()
```
SmittieC, line 47: Is this `status` the same as the `status` on line 50?

apps/pipelines/graph.py
```diff
@@ -0,0 +1,62 @@
...
+
+        node_ids = {n.id for n in self.nodes}
+        incoming = {e.source for e in self.edges}
+        outgoing = {e.target for e in self.edges}
+        start = list(node_ids - outgoing)
+        end = list(node_ids - incoming)
+        if len(start) != 1:
+            raise PipelineBuildError(f"Expected 1 start node, got {len(start)}")
+        if len(end) != 1:
+            raise PipelineBuildError(f"Expected 1 end node, got {len(start)}")
```
snopoke, line 57:
```suggestion
            raise PipelineBuildError(f"Expected 1 end node, got {len(end)}")
```

apps/pipelines/nodes/nodes.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class CreateReport(PipelineNode):
+    prompt: str = (
+        "Make a summary of the following text: {input}. "
+        "Output it as JSON with a single key called 'summary' with the summary."
+    )
+
+    def get_runnable(self, node: Node) -> Runnable:
+        return PromptTemplate.from_template(template=self.prompt) | LLMResponse.build(node)
```
snopoke, line 50: Since we're passing `node` to `LLMResponse.build` doesn't that imply that `node` should also contain the params for the `LLMResponse` node?

Just wondering if the `LLMResponse` attributes shouldn't be copied over or if `CreateReport` shouldn't extend `LLMResponse`.

apps/pipelines/views.py
```diff
@@ -12,6 +15,39 @@ def pipeline_builder(request, team_slug: str):
     return render(request, "pipelines/pipeline_builder.html")
 
 
+def pipeline_node_input_types(request, team_slug):
+    """Returns all the input types for the various nodes
+
+    Example:
+        {
+          "CreateReport": {
+            "prompt": "<class 'str'>"
```
snopoke, line 24: as mentioned above, I think this should also have the inputs of `LLMResponse`

apps/pipelines/nodes/base.py
```diff
@@ -0,0 +1,93 @@
...
+            else:
+                input = state["messages"][-1]
+            result = built_node.invoke(input)
+            if isinstance(result, BaseMessage):
+                output = result.content
+            elif isinstance(result, StringPromptValue):
+                output = result.text
+            else:
+                output = result
+            return PipelineState(messages=[output])
```
SmittieC, line 81: Should we not append the output to the existing messages instead?

apps/pipelines/nodes/nodes.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def get_runnable(self, node: Node) -> Runnable:
+        from apps.service_providers.models import LlmProvider
+
+        try:
+            provider = LlmProvider.objects.get(id=self.llm_provider_id)
+        except LlmProvider.DoesNotExist:
+            raise PipelineNodeBuildError(f"LLM provider with id {self.llm_provider_id} does not exist")
+
+        service = provider.get_llm_service()
```
SmittieC, line 71: It might be good to catch `ServiceProviderConfigError` here as well

apps/pipelines/models.py
proteusvacuum: Good catch! I wanted the second one...

apps/pipelines/nodes/nodes.py
proteusvacuum: You are totally right. Before seeing this, I'd actually just removed the call to LLMResponse and added that in separately in the pipeline, but making it a subclass works well too - so I added it back in

apps/pipelines/views.py
proteusvacuum: Yes! I've now removed this docstring since the way I'm passing this to the frontend is different.

apps/pipelines/nodes/base.py
proteusvacuum: Yes, this is very confusing, but langgraph does what you think it should here. 

The outputs are collated based on the type annotation on `PipelineState` here:
https://github.com/dimagi/open-chat-studio/blob/a73095350768bf40caf70f6afb7a01d4097c273b/apps/pipelines/nodes/base.py#L19-L21

I think because of the way langgraph can create branches that then merge together, the collating of results happens outside of this function.  

https://python.langchain.com/v0.1/docs/langgraph/#define-the-agent-state

proteusvacuum:
I think this is in a good enough / useable state to be merged now, as it is getting quite big! 

Still to do: 
- Make a UI to "manage" pipelines: Create, Edit, View Runs, Delete, (Trigger?)
- Make the Flow UI respect the required input types
- Allow different inputs into the pipeline through the events system 

Known Issues:
 - This currently always loads the pipeline with id 1 (https://github.com/dimagi/open-chat-studio/blob/c3e766e0992cc5cf7a680a342a452eace6578cf0/templates/pipelines/pipeline_builder.html#L18)
 - The PipelineRun model is not saved after every node is completed, due to the way that langgraph handles threads. I'm still confused by this...

apps/pipelines/nodes/base.py
SmittieC: Ah right, thanks for the link to the docs. I saw the `add_messages` method and thought that that was used (surely), but couldn't find where, so then I assumed it wasn't. Cool cool!

apps/events/actions.py
```diff
@@ -101,3 +102,17 @@ def invoke(self, session: ExperimentSession, action) -> str:
...
+    def invoke(self, session: ExperimentSession, action) -> str:
+        from apps.pipelines.models import Pipeline
+
+        try:
+            pipeline: Pipeline = Pipeline.objects.get(id=action.params["pipeline_id"])
+        except KeyError:
+            raise ValueError("The action is missing the pipeline id")
+        except Pipeline.DoesNotExist:
+            raise ValueError("The selected pipeline does not exist, maybe it was deleted?")
+        last_message = session.chat.messages.last().content
```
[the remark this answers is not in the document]
proteusvacuum: What do you think the best way of passing multiple messages into the pipeline is? Something like this?

https://github.com/dimagi/open-chat-studio/blob/f0e4bec1162fa8b2bc64024ca13130c003c0276b/apps/events/actions.py#L116-L125

( I added that in a different branch for discussion in our next sync)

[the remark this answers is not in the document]
snopoke: that seems OK to me

snopoke (approved):

SmittieC (commented):
Super cool change. There's merge conflicts, but otherwise looks awesome

snopoke (approved):

Issue #418, opened 2024-05-23
Use `additional_instructions` to pass participant data to assistant backed bots


snopoke:
See https://platform.openai.com/docs/api-reference/runs/createRun#runs-createrun-additional_instructions

SmittieC:
https://github.com/dimagi/open-chat-studio/pull/422

PR #422, opened 2024-05-27
Assistants to support participant data

SmittieC:
#418 . It [doesn't look like](https://github.com/langchain-ai/langchain/blob/cccc8fbe2fe59bde0846875f67aa046aeb1105a3/libs/langchain/langchain/agents/openai_assistant/base.py#L491) LangChain supports the `additional_instructions` parameter that we could have used to pass in additional instructions. I went for overriding the instructions on each call rather.

Also I moved the `participant_data` to the base class of the class it was currently in to make that property available in the `AssistantExperimentRunnable` class

### DIFF
--- a/apps/service_providers/tests/test_assistant_runnable.py
+++ b/apps/service_providers/tests/test_assistant_runnable.py
@@ -33,6 +33,7 @@ def session():
     return session
 
 
+@pytest.mark.django_db()
 @patch("openai.resources.beta.threads.messages.Messages.list")
 @patch("openai.resources.beta.threads.runs.Runs.retrieve")
 @patch("openai.resources.beta.Threads.create_and_run")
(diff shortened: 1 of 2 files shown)


snopoke (approved):

apps/service_providers/tests/test_assistant_runnable.py
```diff
@@ -33,6 +33,7 @@ def session():
     return session
 
 
+@pytest.mark.django_db()
```
snopoke, line 36: Pity to add these since it slows down the tests quite a bit. What about patching the method?

SmittieC: Sounds reasonable :+1:

snopoke (approved):

Issue #352, opened 2024-04-03
Add tabs to experiment session view


proteusvacuum:
https://github.com/dimagi/open-chat-studio/pull/348#pullrequestreview-1976071296

    My Sessions
    All Sessions
    Events

SmittieC:
Closing this, since its done

PR #423, opened 2024-05-27
Add tab view for Experiments

stephherbers:
https://github.com/dimagi/open-chat-studio/issues/352 

Untitled design (1)

Adds tabs to page rather than having all the content on one page.

Also added custom CSS to override the tailwind defaults so that "All Sessions" is on on line when selected rather than changing to have the line going through the text. I opted for adding a new css file rather than directly modifying the default (but open to feedback)

All of the original content within the cards remains the same.

### DIFF
--- a/static/css/custom.css
+++ b/static/css/custom.css
@@ -1,0 +1,3 @@
+.tab:is(input[type="radio"]) {
+    width: max-content;
+}
--- a/templates/experiments/single_experiment_home.html
+++ b/templates/experiments/single_experiment_home.html
@@ -119,85 +119,97 @@
       {% endfor %}
     {% endif %}
   </div>
-  <div class="app-card">
 
-    <h2 class="pg-subtitle">My Previous Chat Sessions</h2>
-    <div class="overflow-x-auto">
-      <table class="pg-table">
-        <thead>
-          <tr>
-            <th>Started</th>
-            <th>Last Message</th>
-            <th>Actions</th>
-          </tr>
-        </thead>
-        <tbody>
-          {% for session in user_sessions %}
-            <tr>
-              <td>{{ session.created_at }}</td>
-              <td>{{ session.last_message_created_at }}</td>
-              <td>
-                {% if session.is_complete %}
-                  <a class="btn btn-sm btn-outline btn-primary"
-                     href="{% url 'experiments:experiment_session_view' team.slug experiment.public_id session.public_id %}"
-                     class="link">Review Chat</a>
-                {% else %}
-                  <a class="btn btn-sm btn-outline btn-primary"
-                     href="{% url 'experiments:experiment_chat_session' team.slug experiment.id session.id %}" class="link">
-                    Continue Chat
-                  </a>
-                {% endif %}
-              </td>
-            </tr>
-          {% endfor %}
-        </tbody>
-      </table>
-    </div>
-  </div>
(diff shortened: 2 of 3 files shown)


static/css/custom.css
```diff
@@ -0,0 +1,3 @@
+.tab:is(input[type="radio"]) {
```
SmittieC, line 1: Nice. We use the [site-tailwind.css](https://github.com/dimagi/open-chat-studio/blob/main/assets/styles/site-tailwind.css) file for custom styles (there are a few other files as well, but this seems to be the correct place to put this style override). Once it's in there, you can run `npm run dev`/`npm run build` for it to take affect. (I recently started running `npm run dev-watch` while developing. This way new styles are always being "pulled in")

templates/experiments/single_experiment_home.html
```diff
@@ -119,85 +119,97 @@ <h3 class="font-bold text-lg">Link with {{ platform.label }}</h3>
...
-              </td>
-            </tr>
-          {% endfor %}
-        </tbody>
-      </table>
+  <div role="tablist" class="tabs tabs-bordered">
+    <input type="radio" name="tab_group" role="tab" class="tab" aria-label="My Sessions" checked />
+    <div role="tabpanel" class="tab-content">
+      <div class="app-card">
+        <h2 class="pg-subtitle">My Previous Chat Sessions</h2>
```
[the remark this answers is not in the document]
snopoke: +1

static/css/custom.css
stephherbers: ahh good info. I went with the site-tailwind.css change [here](https://github.com/dimagi/open-chat-studio/pull/423#:~:text=width%20for%20tabs-,1275a62,-remove%20redundant%20subtitles)

templates/experiments/single_experiment_home.html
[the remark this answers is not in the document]
stephherbers: https://github.com/dimagi/open-chat-studio/pull/423/commits/d32159993b59ad5f69821bb4d572b8cb1bb33008

static/css/custom.css
SmittieC: Looks like you manually added the site-tailwind.css file. So we don't have to do that. There is [a file](https://github.com/dimagi/open-chat-studio/blob/main/assets/styles/site-tailwind.css) called site-tailwind.css in the assets/styles directory in which you can put the new style and run any one of these commands to load/apply the style: `npm run build`/`npm run dev`/`npm run dev-watch`. This will pull all JS and css/scss files from the assets directory, package them and put them in the static/ directory where the server can find them (see the webpack.config.js file. Not sure how well versed you are with webpack, but [here's](https://webpack.js.org/concepts/configuration/) their explanation of what's going on in there).

The static/js and static/css directories are being built each time we deploy, so we don't have to worry about putting anything in there manually. I hope I made sense :sweat_smile:

stephherbers: @SmittieC ah gotcha thanks for the explanation (that makes a lot of more sense)! I added change here and ran the command locally and got the expected results but lmk if I'm still not understanding it fully: https://github.com/dimagi/open-chat-studio/pull/423/commits/509fe0500685ae8ba0b5ae66a4317f6c47599eea

SmittieC (approved):
Nice!

PR #430, opened 2024-05-31
Make participant data editable

SmittieC:
This should probably be editable


snopoke (dismissed):
What about making this an inline for the participant?

SmittieC:
> What about making this an inline for the participant?

What do you mean exactly?

snopoke:
> > What about making this an inline for the participant?
> 
> Are you suggesting we expose a small UI where experiment admins can edit the participant data in the session view?

No, I mean in the django admin using https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.TabularInline

snopoke (approved):

PR #434, opened 2024-06-05
Use participant timezone in chats

SmittieC:
Included in this PR
- A fix: Web chats had this issue where a participant's latest session was being used regardless which session the participant (or user) used. So if had two sessions on an experiment and used the first (the earlier) session to chat, the history and all of the second session would be used. This is just a web chat issue, and is fixed here
- If we have a participant's timezone (the `timezone` key in the participant data), this will be used when the participant data - specifically the scheduled messages - is included in the chat

### DIFF
--- a/apps/chat/channels.py
+++ b/apps/chat/channels.py
@@ -208,7 +208,7 @@ class ChannelBase:
         if not self.is_message_type_supported():
             return self._handle_unsupported_message()
 
-        if self.experiment_channel.platform != "web":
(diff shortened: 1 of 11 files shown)


snopoke (approved):

SmittieC:
Sorry @snopoke, I made 1 more change

snopoke (approved):

snopoke (approved):

PR #435, opened 2024-06-05
WIP: Participant view

SmittieC:
- Added a participants app
- Added a participants view where one can see details of all participants in a team

### What it looks like thus far

image

#### TODO
- [ ] Move `Participant` and `ParticipantData` to the participants app
- [ ] Show the experiments that the participant participated in
- [ ] Show experiment sessions for this participant
- [ ] Show participant data (with a nice UI where we can add/update it)


apps/participants/views.py
```diff
@@ -100,7 +100,7 @@ def get_context_data(self, *args, **kwargs):
             sessions = participant.experimentsession_set.filter(experiment=experiment).all()
             experiment_data[experiment] = {
                 "sessions": sessions,
-                "participant_data": json.dumps(sessions.first().participant_data_from_experiment),
+                "participant_data": json.dumps(sessions.first().participant_data_from_experiment, indent=4),
```
SmittieC, line 103: image

apps/participants/tests/test_views.py
```diff
@@ -0,0 +1,37 @@
...
+    url = reverse(
+        "participants:edit-participant-data",
+        kwargs={
+            "team_slug": participant.team.slug,
+            "participant_id": participant.id,
+            "experiment_id": session.experiment.id,
+        },
+    )
+
+    data = QueryDict("", mutable=True)
```
snopoke, line 32: this shadows the `data` var on line 17 which makes like 34 confusing: ` data.update({"data": json.dumps(data)})`

apps/participants/views.py
snopoke: How about using the same view as in the session details page: https://github.com/dimagi/open-chat-studio/blob/bc07355c9d567020e4cfdee52c21d80be89e477a/templates/experiments/experiment_session_view.html#L45-L71

apps/participants/views.py
```diff
@@ -0,0 +1,123 @@
...
+            }
+        context["experiment_data"] = experiment_data
+        return context
+
+
+class EditParticipantData(LoginAndTeamRequiredMixin, TemplateView, PermissionRequiredMixin):
+    def post(self, request, team_slug, participant_id, experiment_id):
+        experiment = Experiment.objects.get(team__slug=team_slug, id=experiment_id)
+        participant = Participant.objects.get(id=participant_id)
```
snopoke, line 112: nit: use `get_object_or_404`

apps/participants/views.py
```diff
@@ -0,0 +1,123 @@
...
+
+class DeleteParticipant(LoginAndTeamRequiredMixin, View, PermissionRequiredMixin):
+    permission_required = "experiments.delete_participant"
+
+    def delete(self, request, team_slug: str, pk: int):
+        messages.error(request, "Cannot delete a Participant")
+        return HttpResponse()
+
+
+class ParticipantTableView(SingleTableView):
```
snopoke, line 77: this should also use the `view_participant` permission

apps/participants/views.py
```diff
@@ -0,0 +1,123 @@
...
+            sessions = participant.experimentsession_set.filter(experiment=experiment).all()
+            experiment_data[experiment] = {
+                "sessions": sessions,
+                "participant_data": json.dumps(sessions.first().participant_data_from_experiment, indent=4),
+            }
+        context["experiment_data"] = experiment_data
+        return context
+
+
+class EditParticipantData(LoginAndTeamRequiredMixin, TemplateView, PermissionRequiredMixin):
```
snopoke, line 109: missing `permission_required` field

apps/participants/views.py
```diff
@@ -0,0 +1,123 @@
...
+        ParticipantData.objects.update_or_create(
+            participant=participant,
+            content_type__model="experiment",
+            object_id=experiment_id,
+            team=request.team,
+            defaults={"team": experiment.team, "data": new_data, "content_object": experiment},
+        )
+        return HttpResponseRedirect(
+            reverse("participants:single-participant-home", args=[self.request.team.slug, participant_id])
+        )
```
snopoke, line 123: nit: use django shortcut:
```suggestion
        return redirect("participants:single-participant-home", self.request.team.slug, participant_id)
```

gpt_playground/urls.py
```diff
@@ -40,6 +40,7 @@
     path("pipelines/", include("apps.pipelines.urls")),
     path("files/", include("apps.files.urls")),
     path("annotations/", include("apps.annotations.urls", namespace="annotations")),
+    path("participants/", include("apps.participants.urls", namespace="participants")),
```
snopoke, line 43: `namespace` isn't needed here since you've included the `app_name` var in the `urls` module

templates/participants/partials/participant_experiments.html
```diff
@@ -0,0 +1,66 @@
...
+      <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold">Joined on</th>
+      <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold">Last message received at</th>
+    </tr>
+  </thead>
+  <tbody class="divide-y divide-gray-200" x-data="{experimentsRow: null}">
+    {% for experiment, info in experiment_data.items %}
+      {% with sessions=info|dict_lookup:"sessions" participant_data=info|dict_lookup:"participant_data" %}
+        <tr
+          class="cursor-pointer"
+          x-bind:class="experimentsRow === {{ forloop.counter }} ? 'bg-gray-300' : ''"
```
snopoke, line 16: you can do a striped table by adding the `table-zebra` class to the `table` element (https://v3.daisyui.com/components/table/#zebra)

templates/participants/partials/participant_experiments.html
```diff
@@ -0,0 +1,66 @@
...
+          x-on:click="experimentsRow = (experimentsRow === {{ forloop.counter }} ? null : {{ forloop.counter }})"
+        >
+          <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">{{ experiment.name }}</td>
+          <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">{{ experiment.joined_on }}</td>
+          <td class="px-3 py-4 text-sm text-gray-500">{{ experiment.last_message }}</td>
+        </tr>
+        <tr x-show="experimentsRow === {{ forloop.counter }}" x-cloak>
```
snopoke, line 23: maybe make this an accordion rather than a table?

https://v3.daisyui.com/components/accordion/#using-accordion-and-join-together

templates/participants/partials/participant_experiments.html
```diff
@@ -0,0 +1,66 @@
...
+  function isValidJSON(data) {
+    try {
+      JSON.parse(data);
+      return true;
+    } catch (err) {
+      return false;
+    }
+  }
+
+  document.getElementsByName("data").forEach(element => {
```
snopoke, line 55: what is this for?

templates/participants/single_participant_home.html
```diff
@@ -0,0 +1,29 @@
+{% extends 'web/app/app_base.html' %}
+{% load waffle_tags %}
+
+{% block breadcrumbs %}
+  {% load waffle_tags %}
```
snopoke, line 5: duplicate `load` tag (`waffle_tags` already loaded above). Also, I don't think it needs to be loaded at all.

apps/participants/views.py
SmittieC: We _can_, but that view isn't editable. Just making sure we're aligned that it should be editable. We can make the json view editable and the non-json view non-editable and let it serve as a preview?

snopoke: right, fine to leave it as is for now.

templates/participants/partials/participant_experiments.html
SmittieC: This is to enable tab completion in the textarea where we edit participant data. I'll add a comment and also rename "data" to something more specific

templates/participants/single_participant_home.html
SmittieC: Ah yes, you are correct

templates/participants/partials/participant_experiments.html
SmittieC: We have to either go with a table or an accordian, since we can't really mix the two.

Here's what an accordian might look like
image

PR #438, opened 2024-06-07
swap out sweetalert2 for alertifyjs

snopoke:
Resolves this security alert: https://github.com/dimagi/open-chat-studio/security/dependabot/1


SmittieC (approved):
coolzies

Issue #357, opened 2024-04-07
Flag chat messages that violated the safety layer in the transcript view


SmittieC:
Currently we send an email when a chat message violates a Safety Layer. The reviewer might find it helpful to know this happened and what message triggered it when going through the transcripts.

Thought: We update the event that fires when this happens in the following ways
- Add context to the event. This will allow us to "mark" the chat message in the handling of this event
- Move the sending of the email into the event handling as well

snopoke:
This should use the tags to apply a tag to messages that fail.

PR #440, opened 2024-06-10
Create system tags & record system tag message when using to multi-prompt bot

stephherbers:
Creates a new the of tag called a "System Tag" that is when a tag is created and added automatically and without a user

What's unique:
- No user is needed during creation
- They cannot be edited or deleted by a user
- Current use is to add which bot is used during a chat session, but can be applied to future uses

Example session with a multi-prompt bot:
Screenshot 2024-06-10 at 10 22 28 AM
Resulting tags added:
Screenshot 2024-06-10 at 10 23 11 AM
Example of the "Select a Tag..." multi-select where the system tags are shown, but cannot be added by user
Screenshot 2024-06-10 at 11 11 03 AM


templates/generic/tag_multiselect.html
```diff
@@ -1,8 +1,11 @@
 <div>
-  <select data-info="{{ object.object_info }}" class="tag-multiselect" name="state[]" multiple placeholder="Select a tag..." autocomplete="off">
-    {% for tag in available_tags %}
-      <option value="{{ tag }}" {% if tag in object.get_linked_tags %}selected{% endif %}>{{ tag }}</option>
-    {% endfor %}
-  </select>
+    <select data-info="{{ object.object_info }}" class="tag-multiselect" name="state[]" multiple placeholder="Select a tag..." autocomplete="off">
+        {% for tag in available_tags %}
+            {% if tag not in system_tags %}
```
proteusvacuum, line 4: You could instead just do {% if not tag.is_system_tag %}, and pass the actual tag objects into the template. 

Otherwise, I think if you have both a system tag and a regular tag with the same name, you won't be able to edit the regular tag.

apps/service_providers/llm_service/runnables.py
```diff
@@ -202,11 +203,14 @@ def _populate_memory(self):
         self.memory.chat_memory.messages = messages
 
     def _save_message_to_history(self, message: str, type_: ChatMessageType):
-        ChatMessage.objects.create(
+        chat_message = ChatMessage.objects.create(
             chat=self.session.chat,
             message_type=type_.value,
             content=message,
         )
+        if not Tag.objects.filter(name=self.experiment.name, team=self.session.team).exists():
```
SmittieC, line 211: In a multi-bot setup it would make sense to tag these messages with the experiment, but I don't think it makes that much sense in a single bot setup, since that bot is the only one that could have responded.

apps/annotations/views/tag_views.py
```diff
@@ -61,17 +61,26 @@ class EditTag(UpdateView, PermissionRequiredMixin):
...
+        return Tag.objects.filter(team=self.request.team, is_system_tag=False)
 
     def get_success_url(self):
         return reverse("annotations:tag_home", args=[self.request.team.slug])
 
+    def get(self, request, *args, **kwargs):
+        pk = kwargs.get("pk")
+        queryset = self.get_queryset()
+        if not queryset.filter(pk=pk).exists():
+            return HttpResponseRedirect(reverse("annotations:tag_home", args=[self.request.team.slug]))
```
snopoke, line 73: FYI: Django has a shortcut for doing redirects: https://docs.djangoproject.com/en/5.0/topics/http/shortcuts/#redirect

```suggestion
            return redirect("annotations:tag_home", team_slug=self.request.team.slug)
```

apps/annotations/views/tag_views.py
```diff
@@ -61,17 +61,26 @@ class EditTag(UpdateView, PermissionRequiredMixin):
...
+        return super().get(request, *args, **kwargs)
+
 
 class DeleteTag(LoginAndTeamRequiredMixin, View, PermissionRequiredMixin):
     permission_required = "annotations.delete_tag"
 
     def delete(self, request, team_slug: str, pk: int):
         tag = get_object_or_404(Tag, id=pk, team=request.team)
+        if tag.is_system_tag:
+            return HttpResponseForbidden("System tags cannot be deleted.")
```
snopoke, line 83: How is this show in the UI?

apps/generics/actions.py
```diff
@@ -13,8 +19,7 @@ class Action:
     icon_class: str = None
     extra_context: dict = None
     required_permissions: list = dataclasses.field(default_factory=list)
-
-    display_condition: callable = None
+    display_condition: Callable[[HttpRequest, Context], bool] = display_condition
```
snopoke, line 22: these models and functions are used in other tables as well do I don't think you can default them to functions that assume the record is a `Tag`. Instead you must find the the actions are being used and pass in the display condition function there. For tags I think it is the `apps.annotations.tables.TagTable` class

templates/generic/tag_multiselect.html
snopoke: What is the utility of showing system tags in the dropdown?

templates/experiments/components/experiment_chat.html
```diff
@@ -13,6 +13,9 @@
             {% if perms.annotations.add_customtaggeditem %}
               <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold">Tags</th>
             {% endif %}
+            {% if perms.annotations.add_customtaggeditem %}
```
snopoke, line 16: This should only require a view permission since the user can't add / remove them:
```suggestion
            {% if perms.annotations.view_customtaggeditem %}
```

templates/experiments/components/experiment_chat.html
```diff
@@ -47,6 +50,9 @@
                   {% include 'experiments/components/experiment_sessions_list_tags.html' with tags=message.tags.all %}
                 </td>
               {% endif %}
+              <td scope="col" class="w-1/4 overflow-y-visible">
+                {% include 'experiments/components/experiment_sessions_list_tags.html' with tags=message.get_system_tags %}
+              </td>
```
snopoke, line 55: This should be put inside the permission check block: `{% elif perms.annotations.view_customtaggeditem %}`

templates/experiments/components/experiment_chat.html
```diff
@@ -47,6 +50,9 @@
                   {% include 'experiments/components/experiment_sessions_list_tags.html' with tags=message.tags.all %}
```
snopoke, line 50: I think this will show user tags as well as system tags but it should probably only show user tags right?

apps/service_providers/llm_service/runnables.py
snopoke: You can pass through a flag from here to indicate if the message should be tagged: https://github.com/dimagi/open-chat-studio/blob/2a4ac2153d48b8ef404580c0c5397ef71d7d5225/apps/chat/bots.py#L84

Maybe something like `tag_message` or `add_experiment_tag`?

snopoke (commented):
Great progress!

apps/generics/actions.py
SmittieC, line 22: Since these are generic actions, I would suggest leaving these as-is and moving the `display_condition` method above to [this annotations file](https://github.com/dimagi/open-chat-studio/blob/main/apps/annotations/tables.py) and then passing it as the `display_condition` param like this:
```python
actions.edit_action(url_name="annotations:tag_edit", display_condition=display_condition),
```
and 
```python
                actions.delete_action(
                    url_name="annotations:tag_delete",
                    confirm_message="Continuing with this action will remove this tag from any tagged entity",
                    display_condition=display_condition
                ),
```

Nit: Maybe also rename the `display_condition` to something more specific?

SmittieC:
I like these changes! We can now easily use these changes to tackle [this ticket](https://github.com/dimagi/open-chat-studio/issues/357). General thoughts:
- I think we should keep a list of "reserved" tags that users are not allowed to duplicate. So if users goes to the tag management page and try and create a tag with the same name as a system tag, we should tell them that this is reserved.
- This makes me wonder about the following scenario: Since experiment names are now potential system tags, should we disallow users from the get-go to create tags with an experiment name?
- What about future experiments with the same name as existing tags? If I have a tag named "X" and I create an experiment named "X" later on, then this tag will be used when we tag a message. Should the user be allowed to delete this tag then?

I'm starting to wonder if we shouldn't make system tags even more distinguised. Maybe by prefixing system tags with "system:". This way, we should only disallow users to create tags prefixed with "system:". Thoughts?

templates/experiments/components/experiment_chat.html
stephherbers: [3945761](https://github.com/dimagi/open-chat-studio/pull/440/commits/39457616235039747b54db29804b02d7062afc05)

templates/experiments/components/experiment_chat.html
stephherbers: [3945761](https://github.com/dimagi/open-chat-studio/pull/440/commits/39457616235039747b54db29804b02d7062afc05)

apps/service_providers/llm_service/runnables.py
stephherbers: went with `add_experiment_tag` [e4df433](https://github.com/dimagi/open-chat-studio/pull/440/commits/e4df433afac92faa3fbcf0ba9162c239626e7a42)

templates/generic/tag_multiselect.html
stephherbers: ha I guess not much utility-- I removed all together from drop downs but it will stay in the Tabs table (but still cant be edited) [6412ba3](https://github.com/dimagi/open-chat-studio/pull/440/commits/6412ba37b10b0637bd6fd219aec27d9f3d2e486d)

apps/annotations/views/tag_views.py
stephherbers: with the deletion option taken away this ideally would never show in the UI but rather serve only as a backup, but can remove

apps/annotations/views/tag_views.py
stephherbers: [85486cd](https://github.com/dimagi/open-chat-studio/pull/440/commits/85486cdf2fd21a52de2afcf8bb20f0b6b2fce8da)

apps/generics/actions.py
stephherbers: [9b6c01f](https://github.com/dimagi/open-chat-studio/pull/440/commits/9b6c01f8dfe8bc7fb726f86447dd75f4ca53cb2b)

stephherbers: [9b6c01f](https://github.com/dimagi/open-chat-studio/pull/440/commits/9b6c01f8dfe8bc7fb726f86447dd75f4ca53cb2b)

stephherbers:
thanks for the feedback everyone. Main design changes implemented today:
- added Category to Tag model
- added category and is_system_tag to unique_together in Tag model
- changes system tag name to the ExpRoute keyword
- removed system tags from dropdowns
- only adds system tags when using a multi-prompt bot

apps/service_providers/llm_service/runnables.py
```diff
@@ -205,16 +209,17 @@ def _populate_memory(self):
...
             chat=self.session.chat,
             message_type=type_.value,
             content=message,
         )
-        experiment_route = ExperimentRoute.objects.filter(team=self.session.team, child=self.experiment.id).first()
-        if not Tag.objects.filter(name=experiment_route.keyword, team=self.session.team).exists() and experiment_route:
-            chat_message.tags.create(team=self.session.team, name=experiment_route.keyword, is_system_tag=True)
-        chat_message.add_tags([experiment_route.keyword], team=self.session.team, added_by=None)
+        if add_experiment_tag:
+            exp_route = ExperimentRoute.objects.filter(team=self.session.team, child=self.experiment.id).first()
```
SmittieC, line 219: Ooh, careful here. The first `ExperimentRoute` here might not be the correct one, since there can be multiple experiment routes in this team with this experiment as the child. We need to pass in the "parent" experiment as well to inform this query

apps/annotations/models.py
```diff
@@ -24,11 +29,13 @@ class Tag(TagBase, BaseTeamModel):
     name = models.CharField(verbose_name=pgettext_lazy("A tag name", "name"), max_length=100)
     created_by = models.ForeignKey("users.CustomUser", on_delete=models.DO_NOTHING, null=True, default=None)
     is_system_tag = models.BooleanField(default=False)
+    category_choices = [(category.value, category.name) for category in TagCategories]
```
SmittieC, line 32: Nit: If you use a TextChoices rather than an Enum, then you can do this:
```python
category = models.CharField(choices=TagCategories.choices)
```

See for instance the [VoiceResponseBehaviours](https://github.com/dimagi/open-chat-studio/blob/11c1ce89b8533a36247ccf28b6ed631c7e79627a/apps/experiments/models.py#L289) enum and [its usage](https://github.com/dimagi/open-chat-studio/blob/11c1ce89b8533a36247ccf28b6ed631c7e79627a/apps/experiments/models.py#L402).

apps/service_providers/llm_service/runnables.py
snopoke: We could pass in the tag value directly since it's available in the caller

apps/annotations/views/tag_views.py
snopoke: Fine to leave it as a safeguard

apps/annotations/models.py
stephherbers: [65479a4](https://github.com/dimagi/open-chat-studio/pull/440/commits/65479a45c0908473c634ea401f9b1dd0f0e6d049)

apps/service_providers/llm_service/runnables.py
stephherbers: [2fa9de3](https://github.com/dimagi/open-chat-studio/pull/440/commits/2fa9de34d6530c4357be1c2124bb0810ce16eded)

SmittieC (approved):
LGTM :guitar:

apps/annotations/models.py
```diff
@@ -22,12 +26,14 @@
 )
 class Tag(TagBase, BaseTeamModel):
     name = models.CharField(verbose_name=pgettext_lazy("A tag name", "name"), max_length=100)
-    created_by = models.ForeignKey("users.CustomUser", on_delete=models.DO_NOTHING)
+    created_by = models.ForeignKey("users.CustomUser", on_delete=models.DO_NOTHING, null=True, default=None)
+    is_system_tag = models.BooleanField(default=False)
+    category = models.CharField(choices=TagCategories.choices, default="bot_response")
```
snopoke, line 31: I think the default here should be blank otherwise all user generated tags will have this default value.
```suggestion
    category = models.CharField(choices=TagCategories.choices, blank=True, default="")
```

apps/experiments/views/experiment.py
```diff
@@ -1006,7 +1006,7 @@ def experiment_review(request, team_slug: str, experiment_id: str, session_id: s
             "survey_link": survey_link,
             "survey_text": survey_text,
             "form": form,
-            "available_tags": [t.name for t in Tag.objects.filter(team__slug=team_slug).all()],
+            "available_tags": [t.name for t in Tag.objects.filter(team__slug=team_slug, is_system_tag=False).all()],
```
snopoke, line 1009: Not blocking but you could make this a model manager method since it's a common use case

apps/chat/bots.py
```diff
@@ -84,13 +84,16 @@ def _initialize(self):
     def _call_predict(self, input_str, save_input_to_history=True):
         if self.child_chains:
             chain = self._get_child_chain(input_str)
+            add_experiment_tag = True
```
snopoke, line 87: Not blocking but if we made `_get_child_chain` also return the keyword then we could pass that in directly instead of the flag and we wouldn't need to look up the route in the runnable.

snopoke (commented):
One blocking command and a few smaller ones.

stephherbers:
leaving the remain unblocking comments to a separate PR since this one is already very big 👍

SmittieC (approved):

PR #444, opened 2024-06-11
A little less ugly?

proteusvacuum:
### Before:
image

### After:
image


SmittieC (approved):
A little less ugly indeed :sweat_smile:

PR #445, opened 2024-06-11
Move Participant models from the experiments app to the participants app

SmittieC:
There are [three ways](https://realpython.com/move-django-model/) one can approach moving models to a new app. I opted for the django way where one create the migrations, but use the [SeparateDatabaseAndState](https://docs.djangoproject.com/en/5.0/ref/migration-operations/#django.db.migrations.operations.SeparateDatabaseAndState) operation to update django's state by running the original migrations **only on the state**. The DB operations to delete the tables from the `experiments` app and create them in the `participants` app should be overwritten to rather just rename the table.

Initial testing showed that nothing breaks, but this doesn't mean it's working as expected. I still have to ensure I know exactly what happens to the indexes and unique constraints that would have been removed. My code just leaves them alone, which I suspect is wrong. Maybe we should rename them as well. For now, I'm parking this change until we decide to tackle it.

### DIFF
--- a/apps/api/tests/test_api.py
+++ b/apps/api/tests/test_api.py
@@ -3,7 +3,7 @@ import json
 import pytest
 from django.urls import reverse
 
-from apps.experiments.models import Participant
+from apps.participants.models import Participant
 from apps.utils.factories.experiment import ExperimentFactory
 from apps.utils.factories.team import TeamWithUsersFactory
 from apps.utils.tests.clients import ApiTestClient
--- a/apps/api/views.py
+++ b/apps/api/views.py
@@ -7,7 +7,8 @@ from rest_framework.decorators import api_view, permission_classes
 from rest_framework.generics import ListAPIView
 
 from apps.api.permissions import HasUserAPIKey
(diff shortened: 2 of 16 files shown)


SmittieC:
@snopoke for your review. Happy to create a ticket and link these changes for future implementation

apps/participants/models.py
```diff
@@ -1 +1,103 @@
...
+        return (
+            Experiment.objects.annotate(
+                joined_on=Subquery(joined_on),
+                last_message=Subquery(last_message),
+            )
+            .filter(sessions__participant=self)
+            .distinct()
+        )
+
+    class Meta:
```
snopoke, line 80: you'll also need to specify the [db_table](https://docs.djangoproject.com/en/5.0/ref/models/options/#db-table) here since the table is called `experiments_participant`. 

Alternately you could include the rename in the migrations.

SmittieC:
Closing to minimize PR clutter. I created an [issue](https://github.com/dimagi/open-chat-studio/issues/455) to tackle this and linked this PR to it

Issue #446, opened 2024-06-11
Slack integration


snopoke:
* Bot associated with a channel
* Bot responds in a thread to mentions

PR #451, opened 2024-06-12
Slack messaging provider and channels configuration

snopoke:
Resolves: https://github.com/dimagi/open-chat-studio/issues/446

This allows Slack to be used as a messaging provider.

See README.md for more details.

Peek 2024-06-13 13-27


apps/slack/slack_app.py
```diff
@@ -0,0 +1,154 @@
...
+
+from . import const
+from .const import INSTALLATION_CONFIG
+from .slack_datastores import DjangoInstallationStore, DjangoOAuthStateStore
+
+bolt_logger = get_bolt_logger(App)
+
+
+def get_slack_app():
+    client_id, client_secret, signing_secret, scopes = (
```
SmittieC, line 28: Should these not come from the slack messaging provider instance?

snopoke: Interesting question but I don't think so. This app represents the slack app which you can then add to different slack workspaces. Adding it to a workspace is the 'messaging provider' thing but each OCS instance should only have one slack app (I think).

But I'll give it some more thought.

snopoke: I did some more thinking and I'm sure this is the correct way to do it. If we make it a slack app per provider then users will need to create lots of different slack apps and then add them to their workspaces which is unnecessary. The message provider is connecting the Slack app to a slack workspace and to an OCS team which allows it to interact with the workspace.

I think it would be the same for any oauth based app. E.g. with CommCare we'd create one oauth app in CommCare for the Dimagi OCS instance but users would need to use oath to connect their domain.

apps/experiments/views/experiment.py
```diff
@@ -567,74 +568,14 @@ def update_delete_channel(request, team_slug: str, experiment_id: int, channel_i
...
-
 @require_POST
 @login_and_team_required
 def start_authed_web_session(request, team_slug: str, experiment_id: int):
     experiment = get_object_or_404(Experiment, id=experiment_id, team=request.team)
     experiment_channel = _ensure_experiment_channel_exists(
         experiment=experiment, platform="web", name=f"{experiment.id}-web"
     )
-    session = _start_experiment_session(
+    session = WebChannel.start_new_session(
```
SmittieC, line 578: I like this change. We can move the `experiment_channel = _ensure_experiment_channel_exists(...)` into the `start_new_session` method as well, since we know that the instance should always be a webchannel instance. Might as well create it in there. But maybe for a different PR. I'll take note of this

apps/slack/slack_listeners.py
```diff
@@ -0,0 +1,118 @@
...
+            experiment_channel.experiment, experiment_channel, slack_user, external_id=external_id
+        )
+
+    # strip out the mention
+    message_text = re.sub(rf"<@?{context.bot_user_id}>", "", event["text"])
+    message = SlackMessage(channel_id=channel_id, thread_ts=thread_ts, message_text=message_text)
+
+    ocs_channel = SlackChannel(experiment_channel, session, send_response_to_user=False)
+    response = ocs_channel.new_user_message(message)
+    context.say(response, thread_ts=thread_ts)
```
SmittieC, line 72: In the `send_text_to_user` method in `SlackChannel`, we call `self.messaging_service.send_text_message(...)`. Would `context.say(response, thread_ts=thread_ts)` not cause a duplicate message to be send? (I think it wouldn't, otherwise you would have picked it up, but this line confuses me)

snopoke: I chose to do it this way rather than using the messaging service because there is already an authed client available here.

The key is in the `send_response_to_user` param of the `SlackChannel` class which I set to `False` here to prevent it sending the message. When the channel is used for adhoc bot messages then `send_response_to_user` would be `True` and the message would be sent via the messaging service.

apps/slack/README.md
```diff
@@ -0,0 +1,14 @@
...
+The app supports to connecting OCS with Slack workspaces, handles OAuth authentication, and responds to Slack events.
+
+## How it works
+
+* Slack messaging provider configured using oauth
+* Experiment linked to a Slack channel via the channel ID
+* Slack events are handled by the app and respond directly to messages (not using Celery)
+
+### ALL channels listener
+In addition to linking an experiment to a single channel, there can be one experiment what will respond to mentions
```
SmittieC, line 13:
```suggestion
In addition to linking an experiment to a single channel, there can be one experiment that will respond to mentions
```

apps/slack/slack_listeners.py
SmittieC: Right! I missed the `send_response_to_user=False` :man_facepalming:. Maybe add a comment explaining the decision?

SmittieC: I'm still to read the slack docs on how this integration works exactly, so maybe I'll find the answer in there, but there's two ways we "send a message". 1 is here by doing `context.say(..)`, the other is using the `WebClient`. Why is this necessary?

SmittieC (commented):
Overall it looks great!

One question: Why should the `SlackChannel` override the `_ensure_sessions_exists` and basically handle session management outside of the channel like web channels do? If we include the slack_user (L58 in slack_listeners.py) in the SlackMessage and implement `get_chat_id_from_message` on the `SlackChannel`, then we can follow the "normal flow" by the looks of it. I'm not 100% against the current implementation, but just super wary

apps/slack/slack_listeners.py
snopoke: `context.say` is just a shortcut so that you don't have to create a new web client and authenticate it. It's only available when processing an incoming event and it uses `WebClient` under the hood.

If you want to send messages outside of event handling you have to construct a new WebClient

snopoke:
> Overall it looks great!
> 
> One question: Why should the `SlackChannel` override the `_ensure_sessions_exists` and basically do session management outside of the channel like web channels do? If we include the slack_user (L58 in slack_listeners.py) in the SlackMessage and implement `get_chat_id_from_message` on the `SlackChannel`, then we can follow the "normal flow" by the looks of it. I'm not 100% against the current implementation, but just super wary

The main thing is that we can have multiple active sessions per 'user'. Each slack thread is a session and a user can start new sessions at any time. This is very similar to how 'Web' sessions work. I've updated the readme with more details.

SmittieC:
> > Overall it looks great!
> > One question: Why should the `SlackChannel` override the `_ensure_sessions_exists` and basically do session management outside of the channel like web channels do? If we include the slack_user (L58 in slack_listeners.py) in the SlackMessage and implement `get_chat_id_from_message` on the `SlackChannel`, then we can follow the "normal flow" by the looks of it. I'm not 100% against the current implementation, but just super wary
> 
> The main thing is that we can have multiple active sessions per 'user'. Each slack thread is a session and a user can start new sessions at any time. This is very similar to how 'Web' sessions work. I've updated the readme with more details.

Thanks, that makes sense. This then reveals two possible paradigms for channels i.e. single threaded + multi threaded. Well, it was there all along, we just supported single threaded external channels up to this point. Although this distinction is revealed in the code, I feel like it's a bit sloppy atm and I would ideally like to see it more emphasized. I'm not 100% sure how yet, but my gut tells me we should do something like that. Ideally within the `ChannelBase` class to keep things neat. I'll apply my mind a bit

SmittieC (commented):
Should the backends.py file not be updated as well?

SmittieC (approved):
This is cool :rocket:

snopoke:
> Should the backends.py file not be updated as well?

no, I've updated the tests to ignore the slack app

snopoke:
@SmittieC small change to disable slack messaging provider and channels if slack isn't configured

SmittieC (approved):

PR #456, opened 2024-06-13
Fix index issue

SmittieC:


apps/experiments/models.py
```diff
@@ -487,7 +487,8 @@ def last_seen(self) -> datetime:
         return (
             ChatMessage.objects.filter(chat__experiment_session=models.Subquery(latest_session), message_type="human")
             .order_by("-created_at")[:1]
```
snopoke, line 489: Is this slice necessary if we're calling `.first`?

snopoke (approved):

PR #460, opened 2024-06-14
update_memory method for participants

SmittieC:
This new method does the following:
- For each experiment in the participant's team that they have participated in, it updates the data for that participant by merging the incoming data (updating or creating keys). For any of those experiments which does not have data for this participant yet, it creates a new record with the incoming data.
- Allows you to update the `ParticipantData` record for a specific experiment. This will come in useful when we extract participant data for a specfici experiment

Note: I added a comment, but will emphasize it here as well: Since the `data` field is encrypted at the application level, we are not able to use a DB query to insert/update the new data. We have to pull all records, access the data and put it back. Any suggestions on how I can make this process faster is super welcome!

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -12,7 +12,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres.fields import ArrayField
 from django.core.validators import MaxValueValidator, MinValueValidator, validate_email
 from django.db import models, transaction
-from django.db.models import Count, OuterRef, Q, Subquery
+from django.db.models import Count, OuterRef, Prefetch, Q, Subquery
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext
@@ -528,13 +528,13 @@ class Participant(BaseTeamModel):
         if experiment:
             experiments = [experiment]
         else:
-            experiments = Experiment.objects.filter(team=self.team).prefetch_related("participant_data").all()
-            if experiment:
-                experiments = experiments.filter(id=experiment.id)
+            experiments = Experiment.objects.filter(team=self.team).prefetch_related(
+                Prefetch("participant_data", queryset=ParticipantData.objects.filter(participant=self))
+            )
 
         records_to_update = []
         for experiment in experiments:
-            participant_data = experiment.participant_data.filter(participant=self).first()
+            participant_data = experiment.participant_data.first()
             # We cannot update the participant data using a single query, since the `data` field is encrypted at
             # the application level
             if participant_data:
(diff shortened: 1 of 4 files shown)


apps/experiments/models.py
```diff
@@ -512,6 +512,36 @@ def get_experiments_for_display(self):
...
+
+        Paramters
+        data:
+            A dictionary containing the new data
+        experiment:
+            If specified, only the data for this experiment will be updated
+        """
+        experiments = Experiment.objects.filter(team=self.team).prefetch_related("participant_data").all()
+        if experiment:
+            experiments = experiments.filter(id=experiment.id)
```
snopoke, line 530: Could be:
```suggestion
        if experiment:
            experiments = [experiment]
        else:
            experiments = Experiment.objects.filter(team=self.team).prefetch_related("participant_data").all()
            if experiment:
                experiments = experiments.filter(id=experiment.id)
```

apps/experiments/models.py
```diff
@@ -512,6 +512,36 @@ def get_experiments_for_display(self):
...
+
+        records_to_update = []
+        for experiment in experiments:
+            participant_data = experiment.participant_data.filter(participant=self).first()
```
snopoke, line 534: Doing a filter on `participant_data` runs a new query and won't use the prefetch cache. What you can do is filter the prefetch to only fetch data for this participant:

```python
.prefetch_related(Prefetch("participant_data", queryset=ParticipantData.objects.filter(participant=self)))
```

That should result in all objects in `experiment.participant_data` being for the current participant

SmittieC: beautiful, thanks Simon!

apps/experiments/models.py
```diff
@@ -512,6 +512,39 @@ def get_experiments_for_display(self):
...
+        if experiment:
+            experiments = [experiment]
+        else:
+            experiments = Experiment.objects.filter(team=self.team).prefetch_related(
+                Prefetch("participant_data", queryset=ParticipantData.objects.filter(participant=self))
+            )
+
+        records_to_update = []
+        for experiment in experiments:
+            participant_data = experiment.participant_data.first()
```
snopoke, line 537: I just realised that in the case of a single experiment being passed in, this will be data for all participants. I don't quite know how to handle that case except perhaps to have a getter function which is different for each case. Another option would be to separate this into two functions, one for global updates and one for experiment specific.

stephherbers: +1 I was thinking splitting into 2 functions, but also the getter sounds good too as long as the naming is clear enough

SmittieC: Ah, snappity snap you're right. Let me make an update

snopoke (approved):

PR #461, opened 2024-06-14
some refactoring of the channels and channel messages APIs

snopoke:
Review by commit

### DIFF
--- a/apps/chat/channels.py
+++ b/apps/chat/channels.py
@@ -1,5 +1,5 @@
 import logging
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 from enum import Enum
 from functools import cached_property
 from io import BytesIO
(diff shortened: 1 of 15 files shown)


stephherbers (commented):
looks good to me, other than failing tests with the `_set_telegram_webhook` changes

apps/chat/channels.py
```diff
@@ -441,24 +419,18 @@ def _create_new_experiment_session(self):
...
-            participant = self.experiment_session.participant
-        self.experiment_session = ExperimentSession.objects.create(
-            team=self.experiment.team,
-            participant=participant,
+        self.experiment_session = self.start_new_session(
             experiment=self.experiment,
-            llm=self.experiment.llm,
             experiment_channel=self.experiment_channel,
+            participant_identifier=self.participant_identifier,
+            session_status=SessionStatus.SETUP,
```
SmittieC, line 426: I think it would make more sense to leave the `session_status` so it can default to `SessionStatus.ACTIVE`, since at this point in the conversation, we know that the participant sent a message, making the session active

SmittieC (commented):
Awesome refactor. I especially like that the participant/session management for all channels converged to the `_start_experiment_session` method

apps/chat/channels.py
snopoke: Changing this to `ACTIVE` breaks this test `apps.channels.tests.test_base_channel_behavior.test_pre_conversation_flow`. I'm going to leave it for now to preserve existing functionality.

Issue #391, opened 2024-05-13
make team slugs in URLs case insensitive


snopoke:
When you create a team you get to specify the 'team ID' (ideally this would be auto-generated). This team ID appears to be case sensitive (accessing the URL with different case results in a 404)

PR #465, opened 2024-06-17
Enforce team slugs to lowercase

stephherbers:
https://github.com/dimagi/open-chat-studio/issues/391

Next step is to auto-generate slugs, but adding this as incremental step


SmittieC (commented):
Nice. I think this warrents a small update to maybe [this test](https://github.com/dimagi/open-chat-studio/blob/0adee2f2837aa24cdf25d1568c74c6cb438e31c0/apps/teams/tests/test_creation.py#L27)?

SmittieC:
I see we're using [get_next_unique_team_slug](https://github.com/dimagi/open-chat-studio/blob/0adee2f2837aa24cdf25d1568c74c6cb438e31c0/apps/teams/helpers.py#L15) to generate one when the user [signs up](https://github.com/dimagi/open-chat-studio/blob/0adee2f2837aa24cdf25d1568c74c6cb438e31c0/apps/teams/helpers.py#L57). Wondering if we shouldn't go with that

SmittieC (dismissed):

apps/teams/tests/test_creation.py
```diff
@@ -24,12 +24,26 @@ def test_create_for_user(self):
         assert [SUPER_ADMIN_GROUP] == [group.name for group in membership.groups.all()]
 
 
-def test_create_team_view(db, client):
+def test_slug_generation_on_team_creation(db, client):
+    """Test to verify tghat the slug is autogenerated when creating a team"""
```
SmittieC, line 28:
```suggestion
    """Test to verify that the slug is autogenerated when creating a team"""
```

apps/teams/tests/test_creation.py
```diff
@@ -24,12 +24,26 @@ def test_create_for_user(self):
         assert [SUPER_ADMIN_GROUP] == [group.name for group in membership.groups.all()]
 
 
-def test_create_team_view(db, client):
+def test_slug_generation_on_team_creation(db, client):
```
SmittieC, line 27: I can't seem to find the docs now, but the other day Simon linked me to pytest docs where they recommend that one reserve the `db` fixture for custom fixtures and rather use the `pytest.mark.django_db()` fixture for normal tests

apps/teams/tests/test_creation.py
```diff
@@ -24,12 +25,28 @@ def test_create_for_user(self):
         assert [SUPER_ADMIN_GROUP] == [group.name for group in membership.groups.all()]
 
 
-def test_create_team_view(db, client):
+@pytest.mark.django_db()
+def test_slug_generation_on_team_creation(db, client):
```
SmittieC, line 29: Since we're now using `@pytest.mark.django_db()`, you can remove the `db` parameter

SmittieC (approved):

apps/teams/forms.py
```diff
@@ -96,16 +96,21 @@ def save(self, request):
...
-            "slug": _("Team ID"),
         }
         help_texts = {
             "name": _("Your team name."),
-            "slug": _("A unique ID for your team. No spaces are allowed!"),
         }
 
+    def save(self, commit=True):
+        instance = super().save(commit=False)
+        instance.slug = get_next_unique_team_slug(instance.name)
```
snopoke, line 109: I think this should only happen if the instance is new. I don't think the slug should ever change once it's been set.

snopoke: https://github.com/dimagi/open-chat-studio/pull/487

Issue #360, opened 2024-04-10
Migrate existing No Activity Pings to Events [chore]


proteusvacuum:
Just to track the chore of writing the migration, running it, then deleting the old no activity pings code.
We probably need to do this in conjunction with removing the feature flag from events

PR #468, opened 2024-06-19
Migrate no activity pings to events

proteusvacuum:
Part 1 of: https://github.com/dimagi/open-chat-studio/issues/360

Once the data migration is run, we can delete the model, etc. (I'll open a separate PR to be merged after this has been in prod for a while).

Before deploying this, I think it would be worthwhile to get a dump of all the no activity pings along with the experiments they are associated with, as we'd lose that data if this migration goes wrong.


apps/events/migrations/0011_no_activity_pings_to_events_20240619_1436.py
```diff
@@ -0,0 +1,38 @@
...
+
+from apps.events.models import EventActionType
+
+
+def convert_no_activity_pings_to_event(apps, schema_editor):
+    NoActivityMessageConfig = apps.get_model("experiments", "NoActivityMessageConfig")
+    TimeoutTrigger = apps.get_model("events", "TimeoutTrigger")
+    EventAction = apps.get_model("events", "EventAction")
+
+    for no_activity_ping in NoActivityMessageConfig.objects.all():
```
SmittieC, line 13: Nit: use `prefetch_related`. I see there are only 18 records in prod, so not really necessary :man_shrugging: . Up to you

apps/events/migrations/0011_no_activity_pings_to_events_20240619_1436.py
```diff
@@ -0,0 +1,38 @@
...
+        for experiment in no_activity_ping.experiment_set.all():
+            action = EventAction.objects.create(
+                action_type=EventActionType.SEND_MESSAGE_TO_BOT,
+                params={"message_to_bot": no_activity_ping.message_for_bot}
+            )
+            TimeoutTrigger.objects.create(
```
SmittieC, line 19: This [will cause](https://github.com/dimagi/open-chat-studio/blob/33ba3ec0d6b53d5b301647317fc19a9e2557e6bb/apps/events/models.py#L212) all newly created triggers to go off since there are no event logs for these triggers?

SmittieC (commented):
Just one blocking comment

apps/events/migrations/0011_no_activity_pings_to_events_20240619_1436.py
proteusvacuum: Yikes, great point! https://github.com/dimagi/open-chat-studio/pull/468/commits/b7ab9ae8d32244b4a09f4cbce4261c6325af8001

apps/events/migrations/0011_no_activity_pings_to_events_20240619_1436.py
```diff
@@ -0,0 +1,52 @@
...
+                action_type=EventActionType.SEND_MESSAGE_TO_BOT,
+                params={"message_to_bot": no_activity_ping.message_for_bot}
+            )
+            trigger = TimeoutTrigger.objects.create(
+                action=action,
+                experiment=experiment,
+                delay=no_activity_ping.ping_after * 60,
+                total_num_triggers=no_activity_ping.max_pings,
+            )
+            for session in experiment.sessions.all():
```
SmittieC, line 29: I don't know where my previous comment went. Anycase, how about filter for sessions where `no_activity_ping_count` > 0? :see_no_evil:

proteusvacuum: OK! I have no idea how many people use OCS... so I wasn't sure how worried about this I should be 😆

apps/events/migrations/0011_no_activity_pings_to_events_20240619_1436.py
proteusvacuum: I'm actually not sure my solution is totally correct actually. I think I need to attach the EventLog to the last human message so that it still won't trigger. I'll look again tomorrow.

proteusvacuum: OK, added that here: https://github.com/dimagi/open-chat-studio/pull/468/commits/36ab331edf9b45d966dc8dc7f40cceb6fa46b689

apps/events/migrations/0011_no_activity_pings_to_events_20240619_1436.py
```diff
@@ -0,0 +1,63 @@
...
+                    EventLog.objects.create(
+                        object_id=trigger.id,
+                        content_type=timeout_trigger_content_type,
+                        chat_message_id=last_human_message_id,
+                        session=session,
+                        status=EventLogStatusChoices.SUCCESS,
+                        log="Automatically generated from no activity ping migration"
+
+                    )
+            experiment.no_activity_config = None
```
snopoke, line 50: You could leave this unchanged which would make rollback easier and remove the need to backup the data prior to the deploy. Since the celery task trigger is removed there shouldn't be any duplicate messages.

snopoke (commented):
One comment but otherwise looks good

apps/events/migrations/0011_no_activity_pings_to_events_20240619_1436.py
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/468/commits/26f85481149dfadc2d4f652d991875f8dabe0dea

snopoke (approved):

apps/events/tables.py
```diff
@@ -43,10 +44,16 @@ def render(self, value, record):
         )
 
 
+class ParamsColumn(tables.Column):
+    def render(self, value, record):
+        items = [f"<li><strong>{k}</strong>: {v}</li>" for k, v in value.items()]
+        return format_html(f"<ul>{''.join(items)}</ul>")
```
snopoke, line 50: `items` is the one that needs format_html. Also you can use [format_html_join](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.html.format_html_join):

```suggestion
        items = format_html_join("", "<li><strong>{}</strong>: {}</li>", value.items())
        return f"<ul>{items}</ul>"
```

snopoke (changes requested):

apps/events/tables.py
proteusvacuum: I still need to use `format_html` on the `<ul>` here, otherwise it looks like this: 
image
Thanks for the tip about `format_html_join` - I didn't know about that one! 

https://github.com/dimagi/open-chat-studio/pull/468/commits/a5f56ec112ee19bffce7addfad2cbeb6cba65e48

snopoke: > I still need to use format_html (to mark the string as safe) on the <ul> here

Oh right, you could just use `mark_safe` for that since it's not actually doing any formatting.

snopoke (approved):

PR #470, opened 2024-06-19
Use external_id instead of public_id

proteusvacuum:
Small bug I found


SmittieC (approved):
I had a sneaking suspicion there were one of these hiding somewhere :smile:

Issue #437, opened 2024-06-06
Move custom schedule tools to new schedule framework


snopoke:
* RecurringReminderTool
* OneOffReminderTool

SmittieC:
Context: the mentioned tools creates the equivalent of a "scheduled message" using the `PeriodicTask` object (see [here](https://github.com/dimagi/open-chat-studio/blob/0adee2f2837aa24cdf25d1568c74c6cb438e31c0/apps/chat/agent/tools.py#L128)). Once created, celery will pick it up when the configured time comes and fire whatever method we specified in its "task" attribute. The kwargs of the `PeriodicTask` contains details about the scheduled message i.e. which chats should be pinged, what message to use etc.

The new [ScheduledMessage model](https://github.com/dimagi/open-chat-studio/blob/0adee2f2837aa24cdf25d1568c74c6cb438e31c0/apps/events/models.py#L230) should be used instead to create these messages. Scheduled Messages are currently created through events where the event.params attribute houses the details of the message. You can see what details will be stored there by looking at the [scheduled message config form](https://github.com/dimagi/open-chat-studio/blob/0adee2f2837aa24cdf25d1568c74c6cb438e31c0/apps/events/forms.py#L65). 

We need to basically do the following:
- Migrate existing `PeriodicTask` instances that are used for scheduled messages and create `ScheduledMessage` records for each of these. We'll have to create event records to link to the `ScheduledMessage`. The event's params can be populated using the periodic task's `kwargs` and `interval` / `clocked` attributes.
- Update the above mentioned tools to create `ScheduledMessage` + event records instead of `PeriodicTask` records.

SmittieC:
This is done

PR #471, opened 2024-06-19
Move custom schedule tools to new schedule framework

stephherbers:
Work in Progress
https://github.com/dimagi/open-chat-studio/issues/437


apps/events/migrations/0011_modify_schedulemessage_class.py
```diff
@@ -0,0 +1,43 @@
+# Generated by Django 4.2.11 on 2024-06-19 17:15
+from django_celery_beat.models import PeriodicTask
+from django.db import migrations
+
+from apps.events.forms import EventActionForm
+from apps.events.models import EventActionType, ScheduledMessage
```
proteusvacuum, line 6: You might change this as I see it is a WIP, but when migrating models you need to do use `apps.get_model` instead of accessing the actual model: https://docs.djangoproject.com/en/5.0/topics/migrations/#data-migrations

SmittieC:
Just adding here for visibility: I offlined with @stephherbers on this PR. Additional changes coming soon

stephherbers:
input on tests welcome

apps/events/migrations/0011_modify_schedulemessage_class.py
```diff
@@ -0,0 +1,63 @@
...
+	last_time = task.start_time
+	while last_time <= datetime.now():
+		last_time += delta
+	return last_time - delta
+
+
+class Migration(migrations.Migration):
+    def migrate_periodic_tasks(apps, schema_editor):
+        PeriodicTask = apps.get_model('django_celery_beat', 'PeriodicTask')
+        for periodic_task in PeriodicTask.objects.filter(name__startswith="reminder-"):
```
SmittieC, line 33: Realizing now we should also filter for those that are not enabled (task.enabled=False). Since they are expired, I don't think it's necessary to create effectively completed scheduled messages for them.

apps/events/migrations/0011_modify_schedulemessage_class.py
```diff
@@ -0,0 +1,63 @@
...
+def calculate_next_run(task, current_time, is_complete):
+	if is_complete:
+		return None
+	delta = timedelta(seconds=task.interval.every.total_seconds())
+	next_time = task.start_time
+	while next_time <= current_time:
+		next_time += delta
+	return next_time
+
+def calculate_last_run(task, current_time, is_complete):
```
SmittieC, line 20: We have the last run time from the periodic task, so no need to calculate that :guitar:

apps/events/migrations/0011_modify_schedulemessage_class.py
```diff
@@ -0,0 +1,63 @@
...
+        PeriodicTask = apps.get_model('django_celery_beat', 'PeriodicTask')
+        for periodic_task in PeriodicTask.objects.filter(name__startswith="reminder-"):
+            parts = periodic_task.name.split("-")
+            if len(parts) >= 3:
+                experiment_session_id = parts[1]
+                experiment_session = ExperimentSession.objects.get(id=experiment_session_id)
+                params = {
+                    "time_period": periodic_task.kwargs.get("time_period", "day"),
+                    "frequency": periodic_task.kwargs.get("frequency", 1),
+                    "repetitions": periodic_task.kwargs.get("repetitions", 1),
```
SmittieC, line 41: looks like `periodic_task.kwargs` returns a json. I think the following should help

```python
from apps.events.models import EventActionType
import json

def calc_schedule_components(periodic_task):
    """Periodic task can be a one-off or interval task.
    
    one-off uses the clocked schedule
    interval uses the interval schedule
    """
    if periodic_task.one_off:
        schedule = periodic_task.clocked
        frequency = 1
        # PeriodicTask doesn't have a created at type of field, making it impossible to determine the time_period e.g. hours, days, weeks etc..let's keep it null. This is a doozy. Either we keep it None or assume some time period. I'm leaning to having it None to make it fail if the schedule is updated, which I doubt will happen tbh
        time_period = None
        repetitions = 1
        next_trigger_date = schedule.clocked_time
        total_triggers = 0
    else:
        schedule = periodic_task.interval
        frequency = schedule.every
        time_period = schedule.period
        repetitions = ... # needs to be calculated using expires
        next_trigger_date = ... # needs to be calculated using calculate_next_run
        total_triggers = ... # needs to be calculated

    return frequency, time_period, repetitions, total_triggers, next_trigger_date 

def create_scheduled_messages(apps, schema_editor):
    ScheduledMessage = apps.get_model('events', 'ScheduledMessage')
    EventAction = apps.get_model('events', 'EventAction')
    Experiment = apps.get_model('experiments', 'Experiment')
    Participant = apps.get_model('experiments', 'Participant')

    task = ... # filter for active ones
    # filter for tasks that are disabled
    task_kwargs = json.loads(task.kwargs)
    # There may be many participants, so we have to create a scheduled message + event for each
    participant_identifiers = task_kwargs["chat_ids"]
    message = task_kwargs["message"]
    experiment = Experiment.objects.filter(public_id=task_kwargs["experiment_public_id"])

    frequency, time_period, repetitions, total_triggers, next_trigger_date = calc_schedule_components(task)

    event_action_params = {
        "name": "", # we can generate some name
        "prompt_text": message,
        "frequency": frequency,
        "time_period": time_period,
        "repetitions": repetitions,
        "experiment_id": experiment.id
    }

    for identifier in participant_identifiers:
        # There may be many participants with the same identifier in other teams, so we need to filter by team as well
        participant = Participant.objects.get(team=experiment.team, identifier=identifier)

        event_action = EventAction.objects.create(
            action_type=EventActionType.SCHEDULETRIGGER,
            params=event_action_params
        )
        ScheduledMessage.objects.create(
            experiment=experiment,
            participant=participant,
            team=experiment.team,
            action=event_action,
            last_triggered_at=task.last_run_at,
            next_trigger_date=next_trigger_date,
            total_triggers=total_triggers
        )
```

apps/chat/agent/tools.py
```diff
@@ -124,24 +124,21 @@ def _move_datetime_to_new_weekday_and_time(date: datetime, new_weekday: int, new
...
+    from apps.utils.factories.events import EventActionFactory
+    additional_params = {
+        "name": f"reminder-{experiment_session.id}-{uuid.uuid4()}",
+        'prompt_text': message,
+    }
+    params = {**kwargs, **additional_params}
+    ScheduledMessage.objects.create(
+        participant=experiment_session.participant,
+        team=experiment_session.team,
+        action=EventActionFactory(params=params, action_type=EventActionType.SCHEDULETRIGGER),
```
SmittieC, line 137: I don't think we should be using factories outside of test code? Maybe one can in principle, but I would be very careful, since there may be auto generated fields in there.

SmittieC:
> input on tests welcome

I think if we can generate `one-off` and `interval` type `PeriodicTasks` in a test and run the function that creates scheduled messages from it, it would suffice.

My paranoia often leads me to pull prod's database, load it locally and run the migration against that as well, just to be extra safe. We could do that as well

apps/chat/agent/tools.py
```diff
@@ -124,24 +124,21 @@ def _move_datetime_to_new_weekday_and_time(date: datetime, new_weekday: int, new
...
-        task=BOT_MESSAGE_FOR_USER_TASK,
-        kwargs=task_kwargs,
-        **kwargs,
+def create_schedule_message(experiment_session: ExperimentSession, message: str, trigger_date: datetime, **kwargs):
+    from apps.utils.factories.events import EventActionFactory
+    additional_params = {
+        "name": f"reminder-{experiment_session.id}-{uuid.uuid4()}",
+        'prompt_text': message,
+    }
+    params = {**kwargs, **additional_params}
```
SmittieC, line 133: let's rather be explicit in which keys we put in the params for the Event, since we expect those that are in [this form](https://github.com/dimagi/open-chat-studio/blob/bfc1c1b0c5732286400789c8f447526c283f6a15/apps/events/forms.py#L65) to be in there. I quite liked what you did previously in the migration where you used the `ScheduledMessageConfigForm` to check the params.

What if, we treat the bot as just another type of user that needs to fill out the the `ScheduledMessageConfigForm` form? :thinking: So when we have all the necessary params, we then check the form's validity and if it's invalid, we tell the bot that it's not valid (we can do this by simply returning some string instead of the default "Success"). Hope this makes sense? Thoughts?

SmittieC: A way to verify if we calculated the parameters for the form correctly is to do [this calculation](https://github.com/dimagi/open-chat-studio/blob/bfc1c1b0c5732286400789c8f447526c283f6a15/apps/events/models.py#L261) when we have the params. If the `next_trigger_date` matches the `datetime_due`, then we know we did it right

SmittieC: Apologies, I'm using this as a brain dump :see_no_evil: 

Stepping back a bit makes me wonder: There are now two possible ways to create a scheduled message.
1. Experiment admins basically decide when/how it should be created. This is for a specific purpose i.e. the bot checking in on the user for some specific reason or to achive some goal.
2. The user asking the bot to check in on them or remind them about something. This is differnet though, because the user is the author and should have control over this schedule.

The the first case, it's an experiment admin that created the scheduled message indirectly through an event and the participant has some control over the schedule for the sake of preference i.e. they can ask for another day/time, but not move it out by say a week/month/year. Why? Because it's there for a reason.

Now in the second case, the participant is the one who authored the scheduled message, so it would make sense for them to be able to CRUD this schedule any way they see fit. When these changes are made, they will be able to create a schedule, but their ability to update/delete it will be limited. What if they asked the bot to create a reminder, but want to move it say a month later? We'll need to address this in a separate PR though. cc @snopoke

snopoke: Adding some context from the standup: I think we should combine the tools so that the same set are used for custom vs 'managed' schedules. In future we could provide options for experiment admins to restrict the changes allowed for managed schedules.

code notes:

`ScheduledMessage.action` can be made optional. Its purpose is to provide a reference to the params from the trigger.
For schedule messages that are created (or customised) by the user we can store params in `ScheduledMessage.custom_schedule_params`

When the ‘message’ is triggered we can use any values in `custom_schedule_params` to override the ones in the action (if there is an action)

stephherbers:
closing out to incorporate feedback into a separate PR

PR #472, opened 2024-06-20
Fix: Add team to UserAPIKey and excldue API sessions from sessions list

SmittieC:


stephherbers (approved):

snopoke (commented):
Why do we exclude API sessions?

SmittieC:
> Why do we exclude API sessions?

This was a request from Brian / Amy. Since they're using the API for evaluation / testing, they don't want it to clog up the "main" sessions view, since these are just for testing. The session can still be seen in the participants view, so they are not lost.

Another approach we can take if we want is to not exclude these sessions from the "main" sessions view, but rather add tags to them an filter them out by default. This is more effort though.

PR #474, opened 2024-06-21
Preparation work for data extraction

SmittieC:
[Here's the doc](https://docs.google.com/document/d/1awX0Lt7R1kGhGiM87p1JcpfXtbW385kWmoS2tsGE0m4/edit#heading=h.kg4tb12svbcz) where I'm jotting down thoughts on what needs to be done etc. This is very much a live doc, but do feel free to add comment/concerns.

- Updated the pipeline framework to pass the state to the runnables as well. This allows each step access to the session
- Added two new pipeline components to do **simple** data extraction:

Components:
### Extract data
The user defined schema is converted into a JSON schema and used with langchain's `.with_structured_output` method which will use supported/appropriate endpoints to achieve structured output, depending on the model. This is more powerful than asking the model nicely to give us structured output

data

### Data update
A default merge strategy is used.


apps/experiments/tests/test_models.py
```diff
@@ -22,24 +24,38 @@ def experiment_session():
...
+    data_dir = settings.BASE_DIR / "apps/experiments/migrations/preload_data"
+    voice_files = ["aws_voices.json", "azure_voices.json", "openai_voices.json"]
+    for voice_file in voice_files:
+        file_path = data_dir / voice_file
+        migration_utils.create_synthetic_voices_from_file(SyntheticVoice, file_path=file_path)
+
+
 class TestSyntheticVoice:
     @pytest.mark.django_db()
+    @pytest.mark.usefixtures("_load_voice_data")
```
snopoke, line 39: Did the [django_db_with_data](https://github.com/dimagi/open-chat-studio/blob/c6709572b4a46fe973db881713a13efc7ff95c41/apps/utils/pytest.py#L1) not work?

apps/pipelines/nodes/nodes.py
```diff
@@ -94,9 +94,92 @@ def fn(input: Input):
...
+            except ParticipantData.DoesNotExist:
+                ParticipantData.objects.create(
+                    participant=session.participant,
+                    content_type__model="experiment",
+                    object_id=session.experiment.id,
+                    team=session.team,
+                    data=extracted_data,
+                )
+
+        return RunnableLambda(partial(fn, state=state), name=self.__class__.__name__)
```
snopoke, line 185: I don't think you need to use `partial` here. Isn't the `input` param to `fn` the state already?

snopoke: Hmm, on digging further, I see it isn't. But I think it should be. I don't think `apps.pipelines.nodes.base.PipelineNode.get_callable` should be manipulating the runnable input based on the type of runnable. It should just pass in the state and the runnable must do the right thing. Same with the output.

apps/experiments/tests/test_models.py
SmittieC: It does, thanks

apps/pipelines/nodes/nodes.py
```diff
@@ -94,9 +94,92 @@ def fn(input: Input):
...
+class ExtractStructuredDataBasic(LLMResponse):
+    __human_name__ = "Extract structured data (Basic)"
+    data_schema: str
+
+    def get_runnable(self, node: Node, state: PipelineState) -> RunnableLambda:
+        json_schema = ExtractStructuredDataBasic.to_json_schema(json.loads(self.data_schema))
+        prompt = PromptTemplate.from_template(template="{input}.\nCurrent user data: {participant_data}")
+        return (
+            {"input": RunnablePassthrough()}
+            | RunnablePassthrough.assign(participant_data=RunnableLambda(lambda x: self.get_participant_data(state)))
```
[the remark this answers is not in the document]
SmittieC: Cool. I'll make the updates as we've discussed yesterday and then we can take it from there :+1:

SmittieC:
I'm stopping here for this PR. In a separate one I'll introduce some chunking of the data

apps/pipelines/nodes/base.py
```diff
@@ -23,71 +20,65 @@ def add_messages(left: list, right: list):
...
-    def logger(self, config):
-        for handler in config["callbacks"].handlers:
+    @cached_property
+    def logger(self):
+        for handler in self.config["callbacks"].handlers:
             if isinstance(handler, PipelineLoggingCallbackHandler):
                 return handler.logger
-        raise AttributeError("No logger found")
+
+    def experiment_session(self, state: PipelineState) -> ExperimentSession:
```
snopoke, line 83: nit: this could go on the state object itself

apps/pipelines/nodes/nodes.py
```diff
@@ -72,31 +70,118 @@ class CreateReport(LLMResponse):
...
+            )
 
-        return RunnableLambda(fn, name=self.__class__.__name__)
+        session = self.experiment_session(state)
+        try:
+            participant_data = ParticipantData.objects.for_experiment(session.experiment).get(
+                participant=session.participant
+            )
+            if self.key_name:
+                participant_data.data[self.key_name] = participant_data.data.get(self.key_name, {}) | extracted_data
```
snopoke, line 176: What if the value at `key_name` is a string? Might need a slightly more sophisticated merge method.

apps/pipelines/nodes/nodes.py
```diff
@@ -72,31 +70,118 @@ class CreateReport(LLMResponse):
...
+    the input data will be merged with `key_name` as the key name.
+    """
+
+    __human_name__ = "Update participant memory"
+    key_name: str | None = None
+
+    def _process(self, state: PipelineState) -> PipelineState:
+        extracted_data = state["messages"][-1]
+
+        if not isinstance(extracted_data, dict):
```
snopoke, line 162: If `key_name` is specified then it could be OK to get a string here and just set that value in the data.

apps/pipelines/tests/test_runnable_builder.py
```diff
@@ -16,76 +19,83 @@ def provider():
...
+    runnable.invoke(state)
     assert len(mail.outbox) == 1
     assert mail.outbox[0].subject == "This is an interesting email"
     assert mail.outbox[0].to == ["test@example.com"]
 
 
 @override_settings(CELERY_TASK_ALWAYS_EAGER=True)
-def test_send_email():
+@mock.patch("apps.pipelines.nodes.base.PipelineNode.logger")
+def test_send_email(logger):
```
snopoke, line 98: tip: If you don't need access to the mock object you can prevent having to add the argument by adding `new=Mock()` to the patch (kwarg not technically required since it's the first one):

```
@mock.patch("apps.pipelines.nodes.base.PipelineNode.logger", mock.Mock())
def test_send_email():
```

snopoke (commented):
This is looking great. Nice work splitting the data extraction and participant data update nodes.

apps/pipelines/nodes/base.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/474/commits/e5b09b58d8265b8aa51c146ff260cc27be54525a

apps/pipelines/tests/test_runnable_builder.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/474/commits/65986a3509e201feba514bac702e4d468f5e1d66

apps/pipelines/nodes/nodes.py
```diff
@@ -72,31 +70,119 @@ class CreateReport(LLMResponse):
...
+    """
+
+    __human_name__ = "Update participant memory"
+    key_name: str | None = None
+
+    def _process(self, state: PipelineState) -> PipelineState:
+        """Strategy:
+        - If the input is a string or list, a `key_name` must be present to be valid JSON.
+        - If `key_name` is present, we create a new dictionary with `input` as the value to `key_name`.
+        - Merge this new dictionary with the current participant data, essentially overriding existing data
```
SmittieC, line 163: Note: This will override the existing value at `key_value`
- A string can be replaced by an object or list of objects
- A list can be replaced by a new list
- An object will be replaced by the new object

Q: New data will override the old data such that we lose data?
A: Well, yes, but not if we have the extraction step before this one. When we extract data with the extraction step, we give the existing data to the LLM as well to consider. Test runs have shown so far that if there were entries in say an existing list, they are also included in the final output, so the LLM appended it and performed deduplication.

apps/pipelines/models.py
```diff
@@ -30,15 +30,19 @@ def invoke(self, input: PipelineState, session: ExperimentSession | None = None)
         from apps.pipelines.graph import PipelineGraph
 
         runnable = PipelineGraph.build_runnable_from_json(self.data)
-
+        # Django doesn't auto-serialize objects for JSON fields, so we need to copy the input and save the ID of
+        # the session instead of the session object.
+        input_copy = input.copy()
+        input_copy["experiment_session"] = session.id
```
snopoke, line 36: what about a `to_json` method on the state class?

apps/pipelines/nodes/nodes.py
```diff
@@ -72,31 +70,119 @@ class CreateReport(LLMResponse):
...
+                new_data = {self.key_name: new_data}
+
+            participant_data.data = participant_data.data | new_data
+            participant_data.save()
+        except ParticipantData.DoesNotExist:
+            ParticipantData.objects.create(
+                participant=session.participant,
+                content_object=session.experiment,
+                team=session.team,
+                data={self.key_name: new_data} if self.key_name else new_data,
```
snopoke, line 187: new_data has already been made into a dict by this stage so this should just be:
```suggestion
                data=new_data,
```

apps/pipelines/nodes/nodes.py
snopoke: In future we could consider adding a 'best effort merge' option

apps/pipelines/models.py
```diff
@@ -30,15 +30,19 @@ def invoke(self, input: PipelineState, session: ExperimentSession | None = None)
...
         pipeline_run = PipelineRun.objects.create(
-            pipeline=self, input=input, status=PipelineRunStatus.RUNNING, log={"entries": []}, session=session
+            pipeline=self, input=input_copy, status=PipelineRunStatus.RUNNING, log={"entries": []}, session=session
         )
 
         logging_callback = PipelineLoggingCallbackHandler(pipeline_run)
         logging_callback.logger.info("Starting pipeline run")
         try:
             output = runnable.invoke(input, config=RunnableConfig(callbacks=[logging_callback]))
+            output["experiment_session"] = session.id
```
proteusvacuum, line 45: You should change the signature of this method to not default to `None` for `session`, or handle that case here.

apps/pipelines/nodes/base.py
```diff
@@ -23,71 +20,62 @@ def add_messages(left: list, right: list):
...
+
     """
 
+    _config: RunnableConfig | None = None
+
     class Config:
         arbitrary_types_allowed = True
 
     @classmethod
     def build(cls, node: Node) -> Callable[[dict], dict]:
```
proteusvacuum, line 55: I don't think this is used anywhere now?

apps/pipelines/graph.py
```diff
@@ -42,7 +42,8 @@ def build_runnable(self) -> RunnableSequence:
         state_graph = StateGraph(PipelineState)
         for node in self.nodes:
             node_class = getattr(nodes, node.type)
-            state_graph.add_node(node.id, node_class.get_callable(node))
+            node_instance = node_class(**node.params)
```
proteusvacuum, line 45: Should we catch build errors here, to show somewhere as we used to in `.build`?

apps/pipelines/graph.py
```diff
@@ -42,7 +42,8 @@ def build_runnable(self) -> RunnableSequence:
...
+            state_graph.add_node(node.id, node_instance.process)
```
proteusvacuum, line 46: This is great!

apps/pipelines/nodes/base.py
```diff
@@ -23,71 +20,62 @@ def add_messages(left: list, right: list):
...
-                output = result.text
-            else:
-                output = result
-            return PipelineState(messages=[output])
-
-        return fn
-
-    def get_runnable(self, node: Node) -> Runnable:
-        """Get a predefined runnable to be used in the pipeline"""
+        self.logger.info(f"{cls_name} finished with output: {output}")
```
proteusvacuum, line 68: You probably saw in my other PR, but I started passing inputs and outputs as kwargs, otherwise the log page was getting too hard to read.

proteusvacuum (commented):
I think this is a great change, I had clearly overcomplicated this whole thing, which I think is why I got so bogged down in langchain intricacies.

I think with this change we can remove the `PipelineLoggingCallbackHandler` altogether and just pass in the logger when the graph is created. That will remove a lot of the complexity that was added in there in figuring out which nodes to log or not. It will also simplify calling the logger,

apps/pipelines/nodes/base.py
SmittieC: correct, I'll remove

apps/pipelines/graph.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/474/commits/aa06189b084668fa07aefd0053475a4fef2718fe

apps/pipelines/nodes/nodes.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/474/commits/bc731e6e2d699a35edb6832d8e761cd97647b5ed

apps/pipelines/models.py
SmittieC: I made [this change](https://github.com/dimagi/open-chat-studio/pull/474/commits/6ad8e2b8a4350c5b54224617bd78eeebb15e7356), which will essentially do what this line does, but only when the state has the session in it

SmittieC:
> I think with this change we can remove the PipelineLoggingCallbackHandler altogether and just pass in the logger when the graph is created. That will remove a lot of the complexity that was added in there in figuring out which nodes to log or not. It will also simplify calling the logger,

Cool, will leave this for a next PR

apps/pipelines/nodes/base.py
```diff
@@ -18,10 +18,17 @@ def add_messages(left: list, right: list):
     return left + right
 
 
-class PipelineState(TypedDict):
+class PipelineState(dict):
     messages: Annotated[Sequence[Any], add_messages]
     experiment_session: ExperimentSession
```
snopoke: Putting these on a normal dict doesn't work. I think you should convert this to a pydantic model: https://langchain-ai.github.io/langgraph/how-tos/state-model/#define-the-agent-state

SmittieC: > Putting these on a normal dict doesn't work.

Not sure I follow? It does actually work. I am able to run pipelines with this

snopoke: Interesting, I wonder if that will change when we upgrade langgraph.

What I mean by 'does't work' is that TypedDict is a special class that just uses those fields for type validation. I'm surprised that putting them on a normal dict class has any effect.

snopoke (approved):

apps/pipelines/nodes/base.py
SmittieC: To be honest, I was also surprised

SmittieC: I can change this as part of my followup PR if it makes you nervous. I must admit, now that you mentioned this it sort of makes me nervous

PR #476, opened 2024-06-24
Do not tag human messages with bot tags

SmittieC:
I think it makes sense to not tag human messags with the bot that's going to respond. After this is merged, we can deploy


stephherbers (approved):
good catch

Issue #463, opened 2024-06-17
Expand token based compression to consider the whole input to the LLM


SmittieC:
There's a few of these errors
```
Error code: 400 - {'error': {'message': "This model's maximum context length is 8192 tokens. However, your messages resulted in 8202 tokens. Please reduce the length of the messages.", 'type': 'invalid_request_error', 'param': 'messages', 'code': 'context_length_exceeded'}}
```

By default, we set the experiment's max token limit (under edit experiment -> safety) to 8192, which in the case of the above error is also the model's maximum context length. Since we only consider the chat history + summary when compressing, this results in the corner case when the history + summary comes close to the model's token length since it is just a component of the final input to the LLM. The other components of the input (system + user input + all other data we might give) sometimes drives the total context length to over the model's maximum again.

Two solutions
1. a non-robust, but intrim one: We set the experiment's token limit to be << model's maximum to allow the rest of input components some space to wiggle.
2. The more robust one: We consider all inputs to the LLM when checking the token count (but we only ever compress the history, which we currently do)

PR #477, opened 2024-06-24
Consider system message and user input when compressing history

SmittieC:
https://github.com/dimagi/open-chat-studio/issues/463


stephherbers (approved):
general logic looks good, but flagging I do't have too much background in this code area

apps/service_providers/llm_service/runnables.py
```diff
@@ -197,6 +197,9 @@ def _chat_is_cancelled(self):
     def source_material(self):
         return self.experiment.source_material.material if self.experiment.source_material else ""
 
+    def _input_chain(self) -> Runnable[dict[str, Any], Any]:
```
snopoke, line 200:
```suggestion
    def _input_chain(self) -> Runnable[dict[str, Any], Any]:
        """Return a langchain runnable that when invoked, will return
        the fully populated LLM input. This will be used during history compression.
        """
```

snopoke (dismissed):

snopoke (approved):

PR #478, opened 2024-06-25
Ensure all files are taken into account

jerome-white:
The vector store list protocol seems to "page" responses; by default returning only 20 items (see [here](https://github.com/openai/openai-python/blob/f3e6e634a86d5789ab1274ae27f43adc842f4ba8/src/openai/resources/beta/vector_stores/files.py#L162-L163)). This PR updates the sync code to take this paging into account, going through all files in a store rather than the default number.

### DIFF
--- a/apps/assistants/sync.py
+++ b/apps/assistants/sync.py
@@ -176,7 +176,7 @@ def _sync_vector_store_files_to_openai(client, vector_store_id, files_ids: list[
 
     while True:
         vector_store_files = client.beta.vector_stores.files.list(
-            order='asc',
+            order="asc",
             vector_store_id=vector_store_id,
             **kwargs,
         )
@@ -188,7 +188,7 @@ def _sync_vector_store_files_to_openai(client, vector_store_id, files_ids: list[
 
         if not vector_store_files.has_more:
             break
-        kwargs['after'] = vector_store_files.last_id
+        kwargs["after"] = vector_store_files.last_id
 
     for file_id in to_delete_remote:
         client.beta.vector_stores.files.delete(vector_store_id=vector_store_id, file_id=file_id)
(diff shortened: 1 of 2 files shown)


SmittieC (dismissed):
Thank you for this! Looks good to me :rocket:

SmittieC:
@jerome-white You can [use pre-commits](https://github.com/dimagi/open-chat-studio?tab=readme-ov-file#installing-git-commit-hooks) to fix the minor styling issue. Also just noting the test failure :pray:

jerome-white:
> Also just noting the failing test

The test that's failed has me stumped: I think there's a mismatch in the type that the test expects, and the type that OpenAI projects. The line in question:

https://github.com/dimagi/open-chat-studio/blob/524cd6fde7edcc269dd6c6cfcbd581e8b5a02db8/apps/assistants/sync.py#L178

The test feels `vector_store_files` should be a list; however, based on the OpenAI documentation, it should be a "SyncCursorPage" (see [here](https://github.com/openai/openai-python/blob/e9724398d2bdd87ba41f199c3577303f1b80f2c7/src/openai/resources/beta/vector_stores/files.py#L135)). It's based on what I'm seeing from OpenAI that I drafted this PR. I did some testing on my own OpenAI account with various vector stores and SyncCursorPage's were always returned, even when the vector store was empty.

It looks like the test case [may do some mocking](https://github.com/jerome-white/open-chat-studio/blob/72e6b4483eacc13a392d853cc3d78469a65d824a/apps/assistants/tests/test_sync.py#L74) (?). Irrespective, is handling a list -- along with SyncCursorPage's -- the intended behaviour?

jerome-white:
Altering the test case with the OpenAI return type allows the test to pass. See [this PR](https://github.com/jerome-white/open-chat-studio/pull/1) for details.

If altering the test in this way is acceptable, I'll merge those changes into this PR

SmittieC:
> Altering the test case with the OpenAI return type allows the test to pass. See [this PR](https://github.com/jerome-white/open-chat-studio/pull/1) for details.
> 
> If altering the test in this way is acceptable, I'll merge those changes into this PR

Apologies for only getting back now. Yeah that PR's changes seems fine :+1:

snopoke (approved):

PR #489, opened 2024-07-02
Add UI for experiment routes

proteusvacuum:
Adds a simple UI for experiment routes. 

[Screencast from 2024-07-02 17-03-06.webm](https://github.com/dimagi/open-chat-studio/assets/146896/3a3fe912-a7ba-483f-92a9-07d57e14999c)


templates/experiments/experiment_route_actions.html
```diff
@@ -0,0 +1,9 @@
+{% if record %}
+    {% url url_name request.team.slug record.parent_id record.id as action_url %}
```
proteusvacuum, line 2: This feels hacky, but these templates are here because the URL scheme is different to the "generic" urls (it wants the parent experiment too).

snopoke: The only alternative I can think of is to use a function to get the URL instead of doing it in the template. That way you could pass a custom function as a parameter instead of overriding the template:

```
def get_action_url(request, record):
   ....
```

PR #490, opened 2024-07-02
Add links to tabs on experiment page

proteusvacuum:
I'm not sure if there is a better way of doing this in tailwind.

This allows returning to the correct tab after editing or creating an item. Unfortunately Django doesn't allow using `#` in urls, otherwise I would have made this the "home" url for those items.

Based on `fr/routes`, but can rebase it once that has been merged.


snopoke (commented):
No sure if it's better but here's an alternative using AlpineJS: https://codebase-frontend-library.github.io/codebase-4/docs/8-alpinejs-components/alpinejs-tabs/

snopoke (approved):

snopoke (dismissed):

PR #492, opened 2024-07-03
separate state from experiment runnables + a few refactors & cleanups

snopoke:
The main goal of separating state from the runnables is to make it easier to use the runnables independently of experiments. We can create new state implementations that are backed by other data e.g. from an API or an experiment version.


apps/service_providers/llm_service/runnables.py
```diff
@@ -190,80 +142,54 @@ def _chat_is_cancelled(self):
...
 
     def _populate_memory(self, input: str):
-        # TODO: convert to use BaseChatMessageHistory object
-        model = self.llm_service.get_chat_model(self.experiment.llm, self.experiment.temperature)
-        input_messages = self._input_messages(input)
-        messages = compress_chat_history(
-            self.session.chat, llm=model, max_token_limit=self.experiment.max_token_limit, input_messages=input_messages
-        )
-        self.memory.chat_memory.messages = messages
+        input_messages = self.get_input_messages(input)
```
SmittieC, line 162: `input_messages` and consequently the chat memory will include the prompt (with source material / participant data etc) if we do this. I'm guessing we don't want this?

SmittieC (commented):
Nifty. Just one comment

apps/service_providers/llm_service/runnables.py
SmittieC: Apologies, you are correct. I misunderstood what I saw

SmittieC (dismissed):

PR #499, opened 2024-07-05
OpenAI compatible chat completions API

snopoke:
FYI @bderenzi 

This adds an API endpoint that is compatible (roughly) with the [OpenAI Chat Completions](https://platform.openai.com/docs/api-reference/chat/create) API.

This can be used to create a session, populate it with messages and get a response from the experiment in one request.

This should be considered in `alpha`.

This PR also changes the JSON format for the session creation API so that the message structure matches the OpenAI message format:

```json
"messages": [
      {"role": "assistant", "content": "hi"},
      {"role": "user", "content": "hello"},
]
```

This was done for simplicity but also consistency since it's easier for users if they don't have to use different formats in different places.

### DIFF
--- a/apps/api/urls.py
+++ b/apps/api/urls.py
@@ -11,6 +11,6 @@ router.register(r"sessions", views.ExperimentSessionViewSet, basename="session")
 
 urlpatterns = [
     path("participants/<str:participant_id>/", views.update_participant_data, name="update-participant-data"),
-    path("e/<str:experiment_id>/chat/completions", openai.chat_completions, name="openai-chat-completions"),
+    path("openai/<str:experiment_id>/chat/completions", openai.chat_completions, name="openai-chat-completions"),
     path("", include(router.urls)),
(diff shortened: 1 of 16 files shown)


apps/api/urls.py
```diff
@@ -11,5 +11,6 @@
 
 urlpatterns = [
     path("participants/<str:participant_id>/", views.update_participant_data, name="update-participant-data"),
+    path("openai/<str:experiment_id>/chat/completions", openai.chat_completions, name="openai-chat-completions"),
```
SmittieC, line 14: Is the reason for the `openai/` part to indicate that this follows the openai API?

snopoke: It's not really necessary at all, I just thought it was nice to have it in the URL as a clear indication that it's meant to be an OpenAI compatible API.

SmittieC (approved):
This is pretty cool

PR #502, opened 2024-07-10
make action ScheduledMessage optional

stephherbers:
Prep work for changes Schedule tools: [pr here](https://github.com/dimagi/open-chat-studio/pull/471)

Modifies such that action is not required for ScheduleMessage. Then, if there is not action it will use the `custom_schedule_params` for scheduled messages created by users.


apps/events/tasks.py
```diff
@@ -53,9 +53,9 @@ def fire_trigger(trigger_id, session_id):
 
 
 def _get_messages_to_fire():
-    return ScheduledMessage.objects.filter(is_complete=False, next_trigger_date__lte=functions.Now()).select_related(
-        "action"
-    )
+    return ScheduledMessage.objects.filter(
+        is_complete=False, action__isnull=False, next_trigger_date__lte=functions.Now()
```
snopoke, line 57: will you remove this filter once the rest is implemented?

apps/experiments/models.py
```diff
@@ -778,6 +778,7 @@ def get_participant_scheduled_messages(self, as_dict=False, as_timezone: str | N
             Q(experiment=self.experiment) | Q(experiment__in=models.Subquery(child_experiments)),
             participant=self.participant,
             team=self.team,
+            action__isnull=False,
```
snopoke, line 781: Same question as above about removing the filter

snopoke (approved):

apps/events/tasks.py
stephherbers: yes that was the idea but I was hoping to merge this prior to the other being merged just for ease

PR #506, opened 2024-07-10
auditing updates

snopoke:
* Add 'team' to audit context when available
* Add 'transaction_id' to audit context to allow grouping events e.g. based on a single request
* Add utility for deleting objects with auditing for related model changes

Best reviewed by commit


apps/utils/deletion.py
```diff
@@ -1,8 +1,35 @@
...
+
+def delete_object_with_auditing_of_related_objects(obj):
+    """Deletes the given object and its related objects, auditing the deletion of each object.
+
+    Args:
+        obj: The object to delete.
+    """
+    from field_audit.models import AuditAction
+
+    collector = NestedObjects(using="default")
```
SmittieC, line 20: Today I learned: This! Pretty cool

apps/utils/deletion.py
```diff
@@ -1,8 +1,81 @@
...
+            if model in models_to_delete or model._meta.auto_created or model not in get_audited_models():
+                continue
+            updates_not_part_of_delete[(field, value)] = instances_list
+
+    counter = Counter()
+    with transaction.atomic():
+        _perform_updates_for_delete(updates_not_part_of_delete)
+
+        for model, instances in reversed(collector.data.items()):
+            if model._meta.auto_created:
```
SmittieC, line 38: Why not delete these as well?

apps/audit/middleware.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        transaction_id = get_audit_transaction_id(request)
+        with audit_transaction(transaction_id):
+            return self.get_response(request)
+
+
+def get_audit_transaction_id(request):
```
SmittieC, line 18: Just to make sure I understand correctly, we don't want to have this transaction id to live longer than a single request, since we want to be able to tell which models were deleted together and by which request?

apps/utils/tests/conftest.py
```diff
@@ -12,3 +12,40 @@
...
+        -> b1
+            -> t1 -> p1, p2
+            -> t2
+        -> b2 -> t1 -> p1, p2
+        -> t1 -> p1, p2
+        -> t2
+    c2
+        -> t2
+        -> b3
+    t3
```
SmittieC, line 32: `t3` should be `-> t3`?

SmittieC (approved):
Pretty sweet test you got there to test the auditing events

apps/utils/deletion.py
snopoke: Django will handle their deletion itself when the models they refer to are deleted.

apps/audit/middleware.py
snopoke: If the headers exist then they may be present in other logs e.g. load balancers / WAF and so it's useful to be able to tie them together.

apps/utils/tests/conftest.py
snopoke: no because t3 isn't linked to any collection or bot

SmittieC: My bad, cool cool

PR #507, opened 2024-07-10
Only log the start and end of each defined node

proteusvacuum:
There was an issue with the the pipeline logger that I was having trouble testing before. Now that the tests are running, I was able to make this much better. 
It now automatically logs the entry and exit of each node, and the nodes can add logs as they please.
I also updated the tests.

There are now collapsible entries with the Input and Output of each log entry:
image

And debug logs work:
image

### DIFF
--- a/apps/pipelines/logging.py
+++ b/apps/pipelines/logging.py
@@ -15,42 +15,43 @@ class PipelineLoggingCallbackHandler(BaseCallbackHandler):
     def __init__(self, pipeline_run, verbose=False) -> None:
         self.pipeline_run = pipeline_run
         self.logger = get_logger(uuid.uuid4().hex, self.pipeline_run)
-        self.depth = 0
         self.verbose = verbose
         self._run_id_names = {}
 
     def _should_log(self, kwargs):
+        """We only log steps that are defined nodes inside of the pipeline.
+
+        The nodes themselves should handle logging further if more information is required.
+        """
+
         if "langsmith:hidden" in kwargs.get("tags", []):
             return False
+
         if self.verbose:
             return True
+
         if kwargs.get("run_id") in self._run_id_names:
             return True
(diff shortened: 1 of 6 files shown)


apps/pipelines/logging.py
```diff
@@ -15,42 +15,43 @@ class PipelineLoggingCallbackHandler(BaseCallbackHandler):
...
         return False
 
     def on_chain_start(self, serialized, inputs, *args, **kwargs):
-        self.depth += 1
-        from apps.pipelines.models import PipelineRunStatus
-
-        self.pipeline_run.status = PipelineRunStatus.RUNNING
         if self._should_log(kwargs):
-            self.logger.info(f"{kwargs.get('name', serialized.get('name'))} starting")
+            self.logger.info(
```
SmittieC, line 44: I have been making some changes to this code in [this PR](https://github.com/dimagi/open-chat-studio/pull/474). Logging of the node start + end happens within the nodes itself now, since they are not langchain runnables anymore, but pure functions. See [this](https://github.com/dimagi/open-chat-studio/pull/474/commits/91ef3d203f5ec4255425524a4b0b9e17e595e515) and [this](https://github.com/dimagi/open-chat-studio/pull/474/commits/2de42df142973be461d11ce90b28830f2db8fc22) change (oh and this [one](https://github.com/dimagi/open-chat-studio/pull/474/commits/259d3ed3fad52521abd25a94f6d9aeaf47d842c0) as well). I found it easier to work with if the nodes are not runnables.

SmittieC (approved):

snopoke (approved):
Looks great

apps/pipelines/logging.py
proteusvacuum: Commented on that PR directly - it looks great!

Issue #367, opened 2024-04-12
Validate channel inputs


SmittieC:
When linking a channel, we should validate the channel identifier inputs. For instance, whatsapp channels expect numbers, so we should normalize and validate it. This way, we can prevent users from linking two channels with the same number but different spacing.

snopoke:
@SmittieC is there any more to do on this or can we close it?

SmittieC:
If we want to validate facebook / slack inputs as well, then yes. Maybe not needed?

snopoke:
> If we want to validate facebook / slack inputs as well, then yes. Maybe not needed?

We are already validating slack but fine to keep open for facebook.

SmittieC:
I don't find a reliable way to verify a facebook page ID. There's [this handwavy method](https://stackoverflow.com/questions/17504486/how-to-check-if-a-facebook-id-is-valid), but I'm not sure how much I trust that in the longer term. I think until we have more people using this and starting to get issues, we can revisit. Closing this out for now

PR #510, opened 2024-07-12
Validation for Whatsapp and Telegram channels

SmittieC:
Part of [this issue](https://github.com/dimagi/open-chat-studio/issues/367)

- Telegram bot token is checked whenever a telegram channel is added / updated
- I used the [phonenumbers](https://pypi.org/project/phonenumbers/) library to parse the phone number for whatsapp channels to E.164 format, which is globally unique (https://pypi.org/project/phonenumbers/#:~:text=but%20you%20also%20need%20to%20specify%20the%20country%20that%20the%20phone%20number%20is%20being%20dialled%20from%20(unless%20the%20number%20is%20in%20E.164%20format%2C%20which%20is%20globally%20unique). This is also the format the Turn.io and Twilio parses their numbers. The library also throws an error when the phone number is invalid.

For instance, we consider a number without a country code invalid. Phone numbers can include hashes and spaces which will be automatically removed. See the test case for some valid and invalid examples. I also fixed the extra form errors that didn't show due to a double quote issue in JS (hence we replace all double quotes with single quotes in the message)

Twilio does not support setting whatsapp webhooks programatically, only the [SmsUrl](https://www.twilio.com/docs/phone-numbers/api/incomingphonenumber-resource#update-an-incomingphonenumber-resource), which is not the same as the whatsapp webhook URL. [This thread](https://stackoverflow.com/questions/57384707/set-webhook-for-incoming-messages-on-twilio-whatsapp-phone-number) seems to support this claim. Also, I inspected all "incoming numbers" we have and all of their sms_urls are the same (https://demo.twilio.com/welcome/sms/reply). This does not look like the correct whatsapp webhook url, so this is deffs something else.


requirements/requirements.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with Python 3.11
 # by the following command:
 #
-#    inv requirements
+#    pip-compile requirements/requirements.in
```
snopoke, line 5: out of curiosity, did you use `inv requirements` or `pip-compile requirements/requirements.in`?

snopoke (changes requested):
I think we should use the `phonenumberslite` library since it's might lighter on memory and we don't need the additional metadata ([docs](https://github.com/daviddrysdale/python-phonenumbers?tab=readme-ov-file#memory-usage))

requirements/requirements.txt
SmittieC: no, just raw `pip-compile`. For some reason `inv requirements` didn't want to generate a requirements.txt file for me, so I resorted to the snail. I forgot about this actually. Let me try again

SmittieC:
> I think we should use the `phonenumberslite` library since it's might lighter on memory and we don't need the additional metadata ([docs](https://github.com/daviddrysdale/python-phonenumbers?tab=readme-ov-file#memory-usage))

Good call, I'll update

snopoke (approved):

PR #511, opened 2024-07-12
Usage tracking

snopoke:
I want to go over this again but putting up for review now (it also needs more tests and better docs).

This adds a `Usage` model and utilities for tracking various kinds of usage. Usage is tracked against a source object e.g. experiment session, a service object e.g. an `LlmProvider` instance, and a type.

Currently I'm only tracking LLM token usage but in future we can track usage of other services e.g. TTS etc.

Initially I had the code to record usage at the point of use (or near) but that introduces a lot of boilerplate code that needs to be managed. Instead I decided to track the usage at the service level so that any usage of the service will result in usage being tracked:

```python
session = ...
llm_provider = ...
service = llm_provider.get_llm_service()
chat_model = service.get_chat_model("gpt-4")

with service.record_usage(session):
    chat_model.invoke("hi")
```

### Token counters

Prior to this PR we used langchains classes for token counting. This moves away from that to our own implementation (still does the same thing at the end of the day). The reason being that the langchain counters are cumulative where we needed finer grained control. Also, having the counters under our control gives us more flexibility e.g. using Anthropic values returned from the API instead of doing the counting manually.


apps/experiments/tasks.py
```diff
@@ -54,7 +55,12 @@ def get_prompt_builder_response_task(team_id: int, user_id, data_dict: dict) ->
         last_user_message = input_formatter.format(input=last_user_message)
 
     # Get the response from the bot using the last message from the user and return it
-    answer, input_tokens, output_tokens = conversation.predict(last_user_message)
+    history = PromptBuilderHistory.objects.create(team_id=team_id, owner=user, history=[])
```
stephherbers, line 63: would this give accurate values with compressed histories?

apps/chat/tests/test_routing.py
```diff
@@ -25,13 +23,11 @@
 def test_experiment_routing(with_default, routing_response, expected_tag):
     experiment = _make_experiment_with_routing(with_default)
     session = ExperimentSessionFactory(experiment=experiment)
-    fake_llm = FakeLlm(responses=[routing_response, "How can I help today?"], token_counts=[0])
-    fake_service = FakeLlmService(llm=fake_llm)
-    with patch("apps.experiments.models.Experiment.get_llm_service", new=lambda x: fake_service):
+    with mock_experiment_llm_service(["safe", "How can I help today?"], token_counts=[1]) as service:
```
stephherbers, line 26: thought for separate pr but I've seen mock conversations like this required a lot in tests, curiuos if we can extract to a different file and import in some way?

stephherbers (commented):
looking good so far, curious if there is a doc that goes along with this?

apps/service_providers/llm_service/callbacks.py
```diff
@@ -19,7 +20,8 @@ class TokenCountingCallbackHandler(BaseCallbackHandler):
     def __init__(self, model: BaseLanguageModel):
         super().__init__()
         self._lock = threading.Lock()
-        self.model = model
+        # Use weakref to avoid circular reference between callback handler and model
+        self.model = weakref.proxy(model)
```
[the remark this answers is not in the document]
snopoke: this code got removed later on so it's not an issue

apps/experiments/tasks.py
snopoke: The history compression happens before the LLM api is called so the token counts should be correct in terms of what get's sent to the LLM

apps/chat/tests/test_routing.py
snopoke: I'm curious how that would look. The `mock_experiment_llm_service` is already doing the heavy lifting and it's often necessary to customize the messages and token counts (I could have left the `token_counts` arg out since the value I'm providing is the same as the default).

But I'm all for reducing code duplication so keen to hear your ideas.

stephherbers: true it would only be helpful if those were consistent so I'll look through previous tests for more concrete examples and if it would actually be useful 👍

snopoke:
I'm going to put this on hold for now. It needs some more design thought and is taking longer than I expected. I'm going to shift focus to getting some quick and dirty features for our current billing needs.

PR #514, opened 2024-07-16
Recursive data extraction

SmittieC:
Building on [this change](https://github.com/dimagi/open-chat-studio/pull/474).

I'm using LangChain's `RecursiveCharacterSplitter` to split the input into chunks based on number of tokens. The chunk size in tokens is calculated like this:
the LLM's token limit - the prompt_token_count. The token limit is assumed to be 8192 for now. The `chunk_overlap` number is chosen to be 20%. This was chosen arbitrarily. I'm open to suggestions. Maybe we let the user choose and default to a sensible number?


SmittieC:
I looked into using a pydantic `BaseModel` for the `PipelineState`, but couldn't get it working. According to [this](https://langchain-ai.github.io/langgraph/how-tos/state-model/#:~:text=The%20state_schema%20can%20be%20any%20type%2C%20though%20we%20typically%20use%20a%20python%2Dnative%20TypedDict%20in%20our%20examples%20(or%20in%20the%20case%20of%20MessageGraph%2C%20a%20list).), the state can be any type. I tried the example in that link where they loaded the pydantic base model through langchain, but alas. [This function](https://github.com/langchain-ai/langgraph/blob/449b42717cb7540d37b35619a03fab6cb42269e5/libs/langgraph/langgraph/graph/state.py#L346) throws an error when the state is a pydantic model, since the `input` parameter is not a dictionary.

I think, since 1. the state being a `dict` works, 2. the docs saying it should be able to work with any type (weird?) and 3. that function is expecting a dict, not a pydantic model, I am going to keep the pipeline state a `dict`. Maybe there's other things I need to do to make it work, but we have to ask outselves if the lemon is worth the squeeze. In my humble opinion, no.

snopoke (commented):
Do you think recursion is the right approach here? I think we could simplify this if we used some knowledge of the schema. For example, we could merge the chunks into a multi-value dict based on the top level key and then we'd only need to go back to the llm for keys with more than one value.

SmittieC:
> Do you think recursion is the right approach here? I think we could simplify this if we used some knowledge of the schema. For example, we could merge the chunks into a multi-value dict based on the top level key and then we'd only need to go back to the llm for keys with more than one value.

As per our discussion, I'll loop through the chunks, merge it with the participant data and run inference with the next chunk again

Somethink like this
```python
prev_chunk = chunks[0]
for chunk in chunks[1:]:
  update_participant_data(prev_chunk)
  output = chain.invoke(chunk) # this will include the updated participant data
  prev_chunk = output
  
```

SmittieC:
An update on this: If we're going to be updating the participant data as we iterate through the chunks, then that renders the node to update the data sort of redundant. I'd like to keep that node in. Another option is to take the first chunk's data, and update that with the second chunk. Take this new chunk, update that with the 3rd one and so on. All chunks will have the same schema, so we can do this.

SmittieC:
@snopoke I went with [this approach](https://github.com/dimagi/open-chat-studio/pull/514/commits/ae7cc4a006a19a714bdbd7c85b73111a9c896d4a). Basically instead of updating the participant data as we iterate through the chunks, we'll just "merge" the chunks using an LLM. My reasoning for not updating / including the participant data in these steps are as follows:
1. If we want to keep the "update participant data" step separate, we cannot update it in here. I don't mind merging that step with this one, but I did find a way that seems to work quite well without having to update participant data in this step.
2. We already the participant data to extract each chunk, so the value we got from it is spent. Now it's just a matter of merging/deduplicating.

snopoke:
> @snopoke I went with [this approach](https://github.com/dimagi/open-chat-studio/pull/514/commits/ae7cc4a006a19a714bdbd7c85b73111a9c896d4a). Basically instead of updating the participant data as we iterate through the chunks, we'll just merge+deduplicate the chunks using an LLM. My reasoning for not updating / including the participant data in these steps are as follows:
> 
> 1. If we want to keep the "update participant data" step separate, we cannot update it in here. I don't mind merging that step with this one, but I did find a way that seems to work quite well without having to update participant data in this step.

I this applies equally to normal structured data extraction so I don't think there's any necessity to merge the steps.

> 2. We already the participant data to extract each chunk, so the value we got from it is spent. Now it's just a matter of merging/deduplicating.

As long as there is no overlap in the context between chunks this is true but we can't guarantee that at all so we may have extracted data in one chunk but there could be data in the next chunk that is also relevant and could result in updating previous data. As a simple example:

```
- Chunk 1:

There are 7 children

- Chunk 2:

Actually wait, let me count again. There are 9 children.
```

It seems to me that what you're doing in the 'reduce' step you could be doing in the main step directly.

SmittieC:
@snopoke I still want to do some more testing, but I think this resembles what we talked about

apps/pipelines/nodes/nodes.py
```diff
@@ -96,25 +98,88 @@ def _process(self, state: PipelineState) -> PipelineState:
...
+            new_reference_data = self.update_reference_data(output, reference_data)
+
+        self.post_extraction_hook(new_reference_data, state)
+        return new_reference_data
+
+    def post_extraction_hook(self, output, state):
+        pass
+
+    def get_reference_data(self, state):
+        return {}
```
snopoke, line 141: I think this should default to an empty string rather than a dict. Making it a dict may sway the LLM into producing a dict when it may have otherwise made a list

apps/pipelines/nodes/nodes.py
```diff
@@ -148,40 +213,54 @@ def _create_model_from_data(value_data, model_name: str):
...
-        """
-        new_data = state["messages"][-1]
+    def get_json_schema(self):
+        return ExtractParticipantData.to_json_schema(json.loads(self.data_schema))
+
+    def get_reference_data(self, state) -> dict:
         session = state["experiment_session"]
+        data = session.get_participant_data()
+        if self.key_name:
+            return data.get(self.key_name, {})
```
snopoke, line 233: similar to above re empty string vs dict
```suggestion
            return data.get(self.key_name, "")
```

apps/pipelines/nodes/nodes.py
```diff
@@ -148,40 +213,54 @@ def _create_model_from_data(value_data, model_name: str):
...
+    def update_reference_data(self, new_data: dict, reference_data: dict) -> dict:
         if isinstance(new_data, str) or isinstance(new_data, list):
             if not self.key_name:
                 raise KeyError("A key is expected for a string or list value.")
 
+        if isinstance(new_data, str):
+            return new_data
+
+        if isinstance(new_data, list):
+            return new_data
```
snopoke, line 245: it will be interesting to see if it only returns NEW items or if it includes existing items as well

apps/pipelines/nodes/nodes.py
```diff
@@ -148,40 +213,54 @@ def _create_model_from_data(value_data, model_name: str):
...
+            participant_data.data = participant_data.data | output
             participant_data.save()
         except ParticipantData.DoesNotExist:
             ParticipantData.objects.create(
                 participant=session.participant,
                 content_object=session.experiment,
                 team=session.team,
-                data=new_data,
+                data=output,
             )
```
snopoke, line 266: this try except could be replaced with by using `ParticipantData.objects.update_or_create` e.g. https://github.com/dimagi/open-chat-studio/blob/2ed3328579dff3f9685183e5e89b1a5ef2ef97d2/apps/api/views.py#L131-L137

apps/pipelines/nodes/nodes.py
```diff
@@ -96,25 +98,88 @@ def _process(self, state: PipelineState) -> PipelineState:
...
-    __human_name__ = "Extract structured data (Basic)"
-    data_schema: str
-
-    def _process(self, state: PipelineState) -> RunnableLambda:
-        json_schema = ExtractStructuredDataBasic.to_json_schema(json.loads(self.data_schema))
-        prompt = PromptTemplate.from_template(template="{input}.\nCurrent user data: {participant_data}")
-        chain = (
+class ExtractStructuredDataNodeMixin:
+    def _prompt_chain(self, reference_data):
+        prompt = PromptTemplate.from_template(template="Current user data: {reference_data}\nConversations: {input}")
```
snopoke, line 103: Is this the full prompt? Should it be expanded a bit to include more instruction?

apps/pipelines/nodes/nodes.py
SmittieC: Although I don't think the `reference_data` will influence how the output will be generated, since we give the LLM a `json_schema` to dictate the output structure, but it wouldn't hurt to just use an empty string :+1:

apps/pipelines/nodes/nodes.py
SmittieC: Sure. It seems to be doing great as-is, but that may just be due to it using OpenAI's json data endpoint, which other providers would not necessarily have

apps/pipelines/nodes/nodes.py
SmittieC: I don't think this will work here, because when the `key_name` is specified, the `output` will be a subset of the participant data (more specificially, the subset corresponding to `key_name`), in which case we need to fetch -> merge -> save.

apps/pipelines/nodes/nodes.py
SmittieC: I just realized this will never happen actually. Long story short, `new_data` can never be a pure list. It must always be a dict. A key inside the dict can be a list, but only a list of objects. That's currently a limitation. This means that if the existing data looks like
```python
{"some_key": ["val1", "val2", "val3"]}
```

but we want to extract
```python
{"some_key": [{"key1": "val"}]}
```

this will override the original list. I guess if the LLM cannot figure out what the items in the original list means, it will probably discard it. I'll have to think a bit more around this limitation.

SmittieC: easiest solution is to not allow updating keys with mismatched types

apps/pipelines/nodes/nodes.py
SmittieC: Wait, I think I'm wrong

apps/pipelines/nodes/nodes.py
```diff
@@ -96,25 +98,88 @@ def _process(self, state: PipelineState) -> PipelineState:
...
+
+    def get_json_schema(self):
+        return ExtractStructuredData.to_json_schema(json.loads(self.data_schema))
+
+    def _process(self, state: PipelineState) -> RunnableLambda:
+        json_schema = self.get_json_schema()
+        input: str = state["messages"][-1]
+        reference_data = self.get_reference_data(state)
+        prompt_token_count = self._get_prompt_token_count(reference_data)
+        message_chunks = self.chunk_messages(input, prompt_token_count=prompt_token_count)
```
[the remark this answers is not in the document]
SmittieC: I updated it to always include the schema. OpenAI + Azure (I presume the others as well) injects the schema in a special place in the system prompt under the hood, thus counting towards token usage. Thus we need to take the schema into consideration as well

api-schema.yml
```diff
@@ -127,7 +126,6 @@ paths:
                 value:
                   identifier: part1
                   platform: api
-                  experiment: exp1
```
snopoke, line 130: Thanks

snopoke (approved):

PR #516, opened 2024-07-16
API Key UI updates

snopoke:
* Don't use a disappearing message to show the key after creation
* Allow specifying team and name when creating API key
* Show team and name in table

Peek 2024-07-16 15-18

@bderenzi FYI

### DIFF
--- a/apps/users/forms.py
+++ b/apps/users/forms.py
@@ -3,6 +3,8 @@ from django.conf import settings
 from django.contrib.auth.forms import UserChangeForm
 from django.utils.translation import gettext
 
+from ..api.models import UserAPIKey
+from ..teams.models import Team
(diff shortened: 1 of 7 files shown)


SmittieC (dismissed):

bderenzi:
How tough would it be to add click-to-copy? and/or a little copy button next to the API key in that pop up?

snopoke:
> How tough would it be to add click-to-copy? and/or a little copy button next to the API key in that pop up?

I did consider it - I'll add it quickly.

snopoke:
> How tough would it be to add click-to-copy? and/or a little copy button next to the API key in that pop up?

Done

SmittieC (approved):

PR #517, opened 2024-07-16
Support api-key + bearer token auth for all APIs

snopoke:
This cleans up and standardized the api auth:

* Use authenticator for api key auth instead of 'permissions'
* Support both api-key and bearer auth on all APIs
  * Initially we only supported bearer auth for the openai api but there's no down side to supporting it for them all. Also, I needed to add api key auth to the openai api for Athina.

### DIFF
--- a/api-schema.yml
+++ b/api-schema.yml
@@ -18,7 +18,8 @@ paths:
       tags:
       - Experiments
       security:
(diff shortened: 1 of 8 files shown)


SmittieC (approved):
Cool. I'll need to sharpen up on the drf spectacular stuff

PR #518, opened 2024-07-17
global admin dashboard

snopoke:
This serves as a place to put global admin tools. Currently:

* basic charts
* Options to download usage data and whatsapp number list

In future this can be expanded with more features as needed.

Access to these pages is permitted to `staff` users.

Note: this also moves the Django Admin to `/django-admin/`

image

FYI @bderenzi

### DIFF
--- a/assets/javascript/admin-dashboard.js
+++ b/assets/javascript/admin-dashboard.js
@@ -1,0 +1,120 @@
+'use strict'
+import Chart from 'chart.js/auto'
+
+function listToDict (list) {
+  // gpt
+  return list.reduce((acc, item) => {
+    acc[item.date] = item.count
+    return acc
+  }, {})
+}
+
+function toDateString (dateObj) {
+  return dateObj.toISOString().split('T')[0]
+}
+
+function getTimeSeriesData (start, end, data) {
+  const dataDict = listToDict(data)
+  const chartData = []
+  const current = new Date(start)
+  while (current <= end) { // eslint-disable-line no-unmodified-loop-condition
+    const curString = toDateString(current)
+    chartData.push({
+      x: curString,
+      y: dataDict[curString] || 0
+    })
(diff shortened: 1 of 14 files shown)


stephherbers (commented):
changes look good, but curious why we chose staff vs admin to have easy access to downloading phone numbers? In my eyes, that is more sensitive information than just the usage data so maybe that can be an additional permission or restricted the downloading of data to admins? I know this isn't a big deal now with the number of people that are staff vs admin on projects, but just thinking about in the future if it were to scale

snopoke:
> changes look good, but curious why we chose staff vs admin to have easy access to downloading phone numbers? In my eyes, that is more sensitive information than just the usage data so maybe that can be an additional permission or restricted the downloading of data to admins? I know this isn't a big deal now with the number of people that are staff vs admin on projects, but just thinking about in the future if it were to scale

Sorry for the confusion. 

`staff` refers to the Django `is_staff` attribute on users which is what is used to gate access to the Django Admin panel. This would only be set for specific global admin users.

Also this dashboard is not team specific, it is a global admin dashboard.

stephherbers (approved):

assets/javascript/admin-dashboard.js
```diff
@@ -0,0 +1,120 @@
+'use strict'
+import Chart from 'chart.js/auto'
```
SmittieC, line 2: The other day I was thinking about trying out plotly / dash, [but for django](https://django-plotly-dash.readthedocs.io/en/latest/). I have worked with it before outside of django and it is pretty easy to build beautiful and powerful dashboards with graphs etc. Maybe also something to consider if we ever need to go bigger

SmittieC: And the nice thing about it is that the code is mostly python

snopoke: Looks interesting

PR #521, opened 2024-07-17
celery tweaks

snopoke:
* ignore results: this should reduce some load on celery
* global settings to improve reliable execution
* ignore some tasks in taskbadger


SmittieC (approved):
Nice

Issue #522, opened 2024-07-17
Add channel to participant


snopoke:
A participant is unique to a channel platform and the identifier is channel specific so we should include the channel platform on the participant object and use it in the uniqueness constraint.

PR #523, opened 2024-07-17
add channel platform to participant

snopoke:
Related to https://github.com/dimagi/open-chat-studio/issues/522

This makes it easier to identify participants.

I've also used the platform when looking for existing participants which removes the risk of identifier clashes across platforms e.g. whatsapp number / SMS number


apps/experiments/migrations/0084_participant_platform.py
```diff
@@ -0,0 +1,39 @@
...
+import django.db.models.deletion
+from django.db.models import OuterRef, Subquery, Value
+from django.db.models.functions import Coalesce
+
+
+def set_participant_platform(apps, schema_editor):
+    Participant = apps.get_model("experiments", "Participant")
+    ExperimentSession = apps.get_model("experiments", "ExperimentSession")
+    Participant.objects.update(
+        platform=Coalesce(
```
stephherbers, line 13: read up on the docs for Coalesce and Subquery since I haven't seen that syntax before, what was the give away that this needed to be used rather than a basis query?

stephherbers (dismissed):
non-blocking nit: for all the tests using the string rather than importing ChannelPlatform could be meh if we ever change the string for the value of ChannelPlatform

apps/experiments/migrations/0084_participant_platform.py
SmittieC: Also a question from my side out of curiosity: Did you encounter a case where `experiment_channel__platform` is `None`, or is the usage of `Coalesce` just a safety layer for incase?

apps/experiments/models.py
```diff
@@ -451,6 +451,7 @@ class Participant(BaseTeamModel):
     identifier = models.CharField(max_length=320, blank=True)  # max email length
     public_id = models.UUIDField(default=uuid.uuid4, unique=True)
     user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, null=True, blank=True)
+    platform = models.CharField(max_length=32)
```
SmittieC, line 454: What about adding `choices` here, similar to [what we do](https://github.com/dimagi/open-chat-studio/blob/7d21824d0336f392228e5ff9d4900640fbabdba7/apps/channels/models.py#L25) for the `ExperimentChannel`?

apps/participants/tables.py
```diff
@@ -18,7 +18,7 @@ class ParticipantTable(tables.Table):
 
     class Meta:
         model = Participant
-        fields = ("identifier", "created_at")
+        fields = ("get_platform_display", "identifier", "created_at")
```
SmittieC, line 21: This will show `get_platform_display` in the table header. I think you need something like

```python
platform = columns.Column(accessor="get_platform_display")
```

apps/experiments/models.py
SmittieC: I see from the [issue](https://github.com/dimagi/open-chat-studio/issues/522) that you intended to add a uniqueness constraint as well? An index might also be a good idea

apps/experiments/migrations/0084_participant_platform.py
snopoke: @stephherbers did you mean `basic` query (I'm not sure what a `basis` query is).

They subquery is required since I'm updating the column with a value from a different table so I need to do a join. In PostgreSQL you can do that with `update .. from` syntax but Django only supports doing it with a subquery.

The `Coalesce` was needed because `session.experiment_channel` is an optional field (if you delete a channel it becomes `null`).

SmittieC: Oh I see, since we exclude sessions where the channel is None, it could be that there are none left..thanks!

apps/experiments/models.py
snopoke: > What about adding `choices` here, similar to [what we do](https://github.com/dimagi/open-chat-studio/blob/7d21824d0336f392228e5ff9d4900640fbabdba7/apps/channels/models.py#L25) for the `ExperimentChannel`?

I decided not to do that since I don't think it adds much value and results in more DB migrations every time we add a platform. Perhaps it's a non-issue though. Let me know if you think it's worth it.

> I see from the [issue](https://github.com/dimagi/open-chat-studio/issues/522) that you intended to add a uniqueness constraint as well? An index might also be a good idea

Thanks for the reminder. I wanted to verify that there weren't any duplicates first which I have now done so I'll add the constraint.

SmittieC: > I decided not to do that since I don't think it adds much value and results in more DB migrations every time we add a platform. Perhaps it's a non-issue though. Let me know if you think it's worth it.

For me this adds value more for the devs. When you look at the field definition, you immediately see which values are allowed/used. I'm not against it being removed, but am slightly biased toward keeping it. Say a 6/10 (10=keep it, 0=leave it). If you feel stronger than this in the opposite direction, feel free to leave it out :smile:

apps/experiments/models.py
```diff
@@ -453,6 +453,10 @@ class Participant(BaseTeamModel):
     user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, null=True, blank=True)
     platform = models.CharField(max_length=32)
 
+    class Meta:
+        ordering = ["platform", "identifier"]
+        unique_together = [("team", "platform", "identifier")]
```
SmittieC, line 458: Just confirming we don't also want to index participant on the platform field?

snopoke: Do you mean a separate index for `platform`? I don't think that's necessary.

apps/experiments/models.py
snopoke: I'm going to leave it off for now since referencing it will require moving a bunch of code around to avoid circular dependencies. I also don't want to copy it otherwise they could get out of sync.

SmittieC (approved):
LGTM

apps/experiments/migrations/0084_participant_platform.py
stephherbers: haha no basis was a typo, thanks for the explaination

Issue #515, opened 2024-07-16
Support file upload for 'assistant' backed bots


snopoke:
* option on 'assistant' experiments to allow attachments
* add 'attach' button to chat UI
* attach file to OpenAI thread

PR #525, opened 2024-07-17
Support file uploads for assistants

SmittieC:
Related [issue](https://github.com/dimagi/open-chat-studio/issues/515)

Keeping in Draft for now, since I still need to round it off a bit.

I added a new `ThreadToolResources` model with a FK to the `Chat` model. Each chat message with just keep track of the `external_id` of the uploaded file in its newly added `metadata` field, but the tool resources will sit on the chat object. Having a tool resource on each chat message object feels weird, I can't articulate exactly why though (probably because it's denormalized?)

Anycase, I'm open to suggestions. Also, you'll see when the user uploads the files, we create file records there and then in the view and pass them as attachments through to the chain, where we then finally upload them to OpenAI and create/update the `ToolResource`. I'm on the fence between this approach and doing all of this in the view when the files are created. Maybe we'll want to support this feature on other channels, in which case this approach makes more sense? I'm open to suggestions here. It's not a bit lift to change it.

### Things I still need to do
- Add more tests
- Fix existing tests
- Show files were uploaded in the chat history view


apps/channels/datamodels.py
```diff
@@ -1,14 +1,28 @@
...
 from apps.chat.channels import MESSAGE_TYPES
+from apps.files.models import File
+
+
+class Attachment(TypedDict):
+    file: File
+    type: str
+
+    class Config:
+        arbitrary_types_allowed = True
```
snopoke, line 15: Is this needed on a `TypedDict`? I thought that was a pydantic specific thing

apps/chat/views.py
```diff
@@ -0,0 +1,17 @@
...
+from django.http import FileResponse, Http404
+from django.shortcuts import get_object_or_404
+
+from apps.files.models import File
+from apps.teams.decorators import login_and_team_required
+
+
+@login_and_team_required
+@permission_required("assistants.view_threadtoolresources")
+def download_file(request, team_slug: str, pk: int):
```
snopoke, line 11: There is already a view for this: https://github.com/dimagi/open-chat-studio/blob/bb9bdf14065245e475fefbc2858c3a02ab78ec8f/apps/files/urls.py#L8

If you want an additional permission you could extend that view and add the decorator

apps/experiments/tests/test_views.py
```diff
@@ -339,3 +346,24 @@ def test_timezone_saved_in_participant_data(_trigger_mock):
...
+
+    client.force_login(experiment.owner)
+    file_search_file = BytesIO(b"some content")
+    file_search_file.name = "fs.text"
+    code_interpreter_file = BytesIO(b"some content")
+    code_interpreter_file.name = "ci.text"
+    data = {"message": "Hi there", "file_search": [file_search_file], "code_interpreter": [code_interpreter_file]}
+    client.post(url, data=data)
+    File.objects.get(name="fs.text")
+    File.objects.get(name="ci.text")
```
snopoke, line 369: nit: use asserts:
```suggestion
    assert File.objects.filter(name="fs.text").exists()
    assert File.objects.filter(name="ci.text").exists()
```

apps/assistants/models.py
```diff
@@ -85,3 +85,17 @@ def label(self):
 
     def __str__(self):
         return f"Tool Resources for {self.assistant.name}: {self.tool_type}"
+
+
+class ThreadToolResources(BaseModel):
```
snopoke, line 90: I know that this is specific to assistants currently, but we'll likely expand it's usage later so I think this should go in the `chat` app.

apps/experiments/views/experiment.py
```diff
@@ -641,11 +642,22 @@ def experiment_chat_session(request, team_slug: str, experiment_id: int, session
...
 def experiment_session_message(request, team_slug: str, experiment_id: int, session_id: int):
     message_text = request.POST["message"]
+    uploaded_files = request.FILES
+    attachments = []
+    created_files = []
+    for file_type in ["code_interpreter", "file_search"]:
+        if file_type not in uploaded_files:
+            continue
+        for uploaded_file in uploaded_files.getlist(file_type):
+            new_file = File.objects.create(name=uploaded_file.name, file=uploaded_file, team=request.team)
```
snopoke, line 652: Why don't you create the `ThreadToolResources` object here too? There's some risk that the celery task fails and then these files will be orphaned. We're already fetching the session below so there's no additional query to get the chat ID (`session.chat_id`). It also saves you have to create them in the runnable.

apps/chat/views.py
snopoke: Another option here is to link to the `ThreadToolResources` object which would allow us to validate the user's access as well. I think we'll need to apply the same checks as in `apps.experiments.views.experiment.experiment_review` using the `@verify_session_access_cookie` decorator (or we just don't allow unauthenticated users to download the files).

I think it also makes sense to move this to the `experiments` views file so that it can be alongside the other chat views.

templates/experiments/chat/input_bar.html
```diff
@@ -13,12 +13,106 @@
...
+{% if session.experiment.assistant %}
+  <script>
+    document.addEventListener('alpine:init', () => {
+      Alpine.data('fileUploads', () => ({
+        code_interpreter_files: [],
+        file_search_files: [],
+        handleFileChange(event, targetArray) {
+          const files = Array.from(event.target.files);
+          this[targetArray].push(...files);
+          const dataTransfer = new DataTransfer();
```
snopoke, line 100: TIL: Manipulate input files using `DataTransfer`

apps/experiments/tasks.py
```diff
@@ -5,22 +5,32 @@
...
+def get_response_for_webchat_task(self, experiment_session_id: int, message_text: str, attachments: list) -> str:
     experiment_session = ExperimentSession.objects.get(id=experiment_session_id)
     web_channel = WebChannel(
         experiment_channel=experiment_session.experiment_channel, experiment_session=experiment_session
     )
-    message = BaseMessage(participant_id=experiment_session.participant.identifier, message_text=message_text)
+    message_attachments = []
+    for file_entry in attachments:
+        type, file_id = file_entry.values()
+        message_attachments.append(Attachment(file=File.objects.get(id=file_id), type=type))
```
snopoke, line 27: Instead of fetching these here I think it would be better to just pass the file ID and to the fetch just in time e.g. `attachment.get_file()`

snopoke (commented):
This is pretty great. Nice job on the UI!

snopoke:
> The assistant sometimes seems like it doesn't know about the uploaded file, even though I have confirmed that it is uploaded and attached to the thread. I have seen other people also complaining about this. I have confirmed that it happens mostly when uploading file_search files. I presume it's because it gets embedded whereas code_interpreter files gets opened instead (I think). Anycase, someone in some thread recommended that for a "knowledge base", rather upload file_search files at the assistant level. I have found this to be more reliable. cc @bderenzi

We can't upload it to the assistant since that would make it available to other threads. The issue may be that the embedding hasn't completed before we create the run. I don't see an easy way to check that the processing has been completed. I think you'd have to:

* Fetch the thread object and get the vector store ID
* Fetch the vectorstore and wait for `file_counts.file_counts` to be `0`

This would need to be done in the `OpenAIAssistantRunnable` class.

templates/experiments/chat/input_bar.html
SmittieC: TIL?

apps/experiments/views/experiment.py
SmittieC: I was between creating it here and inside the assistants runnable. Happy to move it to the view :+1:

templates/experiments/chat/input_bar.html
snopoke: Today I Learned

apps/channels/datamodels.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/525/commits/dea2fd86a8737af55d61342c8a15d31b7573231a

apps/experiments/views/experiment.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/525/commits/ba7e6c81462bdebfb60f6b64d37166a8b17a6a45

apps/assistants/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/525/commits/12f85cf7797dc7455ba67d91c4ee40719fedf264

apps/experiments/tasks.py
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/525/commits/ba7e6c81462bdebfb60f6b64d37166a8b17a6a45#diff-6f059913282d6a349d38608d1bc0164c17808f89eb02efb6e0a945e72d2e595f. What I did instead is keep the `file_id` in the attachment and do a bulk fetch in the runnable

apps/chat/models.py
```diff
@@ -127,3 +127,17 @@ def get_attached_files(self):
         if file_ids:
             return File.objects.filter(team=self.chat.team, external_id__in=file_ids)
         return []
+
+
+class ThreadToolResources(BaseModel):
```
snopoke, line 132: Sorry, I should have made this suggestion earlier. What about renaming this to `ChatAttachments`?

Also just to mention that the next ticket is to support downloading attachments generated by the AI and I think we can use the same model. Just want to make sure we get it right now so we don't have to immediately change it.

apps/chat/views.py
SmittieC: let me know what you think about [this approach](https://github.com/dimagi/open-chat-studio/pull/525/commits/865f469133b090f8dd5fe91ef544ff2eb46b8638)

snopoke (commented):
Looks good. I'm going to go through it once more tomorrow. Also fine to leave the 'vector store checking' to a future PR.

snopoke (commented):
Could you add a test for using the assistant runnable with attachments (see tests in `apps.service_providers.tests.test_assistant_runnable`

SmittieC:
> Could you add a test for using the assistant runnable with attachments (see tests in `apps.service_providers.tests.test_assistant_runnable`

Sure! https://github.com/dimagi/open-chat-studio/pull/525/commits/279a1ef7f508ebf9b4f46fb0147b38cc40387493

snopoke (approved):

PR #527, opened 2024-07-18
Move custom schedule tools away from using PeriodicTasks in favor of the new schedule framework

stephherbers:
https://github.com/dimagi/open-chat-studio/pull/471#:~:text=Work%20in%20Progress-,%23437,-migrate%20tools%20off

Incorporated feedback from previous [draft PR ](https://github.com/dimagi/open-chat-studio/pull/471) such that custom_schedule_params takes precedence over the action which is possible after the action was made option in a previous PR


apps/chat/agent/tools.py
```diff
@@ -125,21 +123,28 @@ def _move_datetime_to_new_weekday_and_time(date: datetime, new_weekday: int, new
...
-    )
-    PeriodicTask.objects.create(
-        name=f"reminder-{experiment_session.id}-{uuid.uuid4()}",
-        task=BOT_MESSAGE_FOR_USER_TASK,
-        kwargs=task_kwargs,
-        **kwargs,
-    )
+def create_schedule_message(experiment_session: ExperimentSession, message: str, **kwargs):
+    form = ScheduledMessageConfigForm(data=kwargs, experiment_id=experiment_session.experiment.public_id)
+    if form.is_valid():
```
snopoke, line 128: should this function return errors if it's not valid

apps/chat/agent/tools.py
```diff
@@ -125,21 +123,28 @@ def _move_datetime_to_new_weekday_and_time(date: datetime, new_weekday: int, new
...
+        cleaned_data = form.cleaned_data
+        try:
+            with transaction.atomic():
+                experiment = Experiment.objects.get(id=cleaned_data["experiment_id"])
```
snopoke, line 132: nit: you can use the experiment from the session `experiment_session.experiment`

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
```diff
@@ -0,0 +1,120 @@
...
+        Experiment = apps.get_model('experiments', 'Experiment')
+        Participant = apps.get_model('experiments', 'Participant')
+        PeriodicTask = apps.get_model('django_celery_beat', 'PeriodicTask')
+        current_time = datetime.now()
+
+        for task in PeriodicTask.objects.filter(name__startswith="reminder-", enabled=False):
+            task_kwargs = json.loads(task.kwargs)
+            participant_identifiers = task_kwargs["chat_ids"]
+            message = task_kwargs["message"]
+            experiment = Experiment.objects.filter(public_id=task_kwargs["experiment_public_id"])
```
snopoke, line 84: should check if the experiment exists and ignore tasks if not

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
```diff
@@ -0,0 +1,120 @@
...
+def calculate_total_triggers(task, current_time):
+    interval_duration = get_interval_duration(task.interval)
+    start_time = task.date_changed
+    if not start_time or not interval_duration:
+        return 0
+    elapsed_time = current_time - start_time
+    return int(elapsed_time / interval_duration)
+
+def calc_schedule_components(periodic_task, current_time):
+    task_complete = current_time > periodic_task.expires
```
snopoke, line 52: should we bother copying expired schedules?

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
```diff
@@ -0,0 +1,120 @@
...
+        frequency = 1
+        time_period = None
+        repetitions = 1
+        next_trigger_date = schedule.clocked_time
+        total_triggers = 0
+    else:
+        schedule = periodic_task.interval
+        frequency = schedule.every
+        time_period = schedule.period
+        repetitions = calculate_remaining_repetitions(periodic_task, current_time) if not task_complete else 0,
```
snopoke, line 64: also wondering if we need to copy over schedules that have no more repetitions

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
```diff
@@ -0,0 +1,120 @@
...
+    elif interval.period == 'days':
+        return timedelta(days=interval.every)
+    elif interval.period == 'weeks':
+        return timedelta(weeks=interval.every)
+    else:
+        return None
+
+def calculate_remaining_repetitions(task, current_time):
+    interval_duration = get_interval_duration(task.interval)
+    remaining_time = task.expires - current_time if task.expires else None
```
snopoke, line 37: If the task never expires doesn't that imply an infinite number of repetitions? I don't think we support that but maybe we should?

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
```diff
@@ -0,0 +1,120 @@
...
+                "experiment_id": experiment.id
+            }
+
+            for identifier in participant_identifiers:
+                participant = Participant.objects.get(team=experiment.team, identifier=identifier)
+
+                event_action = EventAction.objects.create(
+                    action_type=EventActionType.SCHEDULETRIGGER,
+                    params=event_action_params
+                )
```
snopoke, line 103: Why do we use an action here instead of `custom_schedule_params`? I think all of these are use initiated schedules.

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
```diff
@@ -0,0 +1,120 @@
...
+class Migration(migrations.Migration):
+    def migrate_periodic_tasks(apps, schema_editor):
+        ScheduledMessage = apps.get_model('events', 'ScheduledMessage')
+        EventAction = apps.get_model('events', 'EventAction')
+        Experiment = apps.get_model('experiments', 'Experiment')
+        Participant = apps.get_model('experiments', 'Participant')
+        PeriodicTask = apps.get_model('django_celery_beat', 'PeriodicTask')
+        current_time = datetime.now()
+
+        for task in PeriodicTask.objects.filter(name__startswith="reminder-", enabled=False):
```
snopoke, line 80: Why do we filter by `enabled=False`? Shouldn't it be `enabled=True`?

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
```diff
@@ -0,0 +1,120 @@
...
+                    next_trigger_date=next_trigger_date,
+                    total_triggers=total_triggers
+                )
+
+    dependencies = [
+        ('events', '0013_alter_scheduledmessage_action'),
+    ]
+
+    operations = [
+    migrations.RunPython(migrate_periodic_tasks),
```
snopoke, line 119: nit: make it elidable
```suggestion
    migrations.RunPython(migrate_periodic_tasks, elidable=True),
```

snopoke (commented):
Generally looks good. Mostly questions about the migration.

apps/events/models.py
```diff
@@ -305,21 +306,24 @@ def _trigger(self):
 
         self.save()
 
+    def get_params(self):
```
SmittieC, line 309: nit: make this a `@cached_property` as well

apps/chat/agent/tools.py
SmittieC: Since the return value of this will be fed back to the agent, I think we should catch any error and return something like "Could not do that". The LLM will use this to tell the user that it couldn't create a scheduled message

apps/chat/agent/tools.py
```diff
@@ -125,21 +123,28 @@ def _move_datetime_to_new_weekday_and_time(date: datetime, new_weekday: int, new
...
-            "experiment_public_id": str(experiment_session.experiment.public_id),
-        }
-    )
-    PeriodicTask.objects.create(
-        name=f"reminder-{experiment_session.id}-{uuid.uuid4()}",
-        task=BOT_MESSAGE_FOR_USER_TASK,
-        kwargs=task_kwargs,
-        **kwargs,
-    )
+def create_schedule_message(experiment_session: ExperimentSession, message: str, **kwargs):
```
SmittieC, line 126: Looks like `cleaned_data["name"]` will fail the form validation, since it would be missing? i don't think the name here is too important, so we could probably generate it. Also would be a good idea to add a test for this function.

stephherbers: [7ec855d](https://github.com/dimagi/open-chat-studio/pull/527/commits/7ec855d043b90502bb536237c2e585b95a38fdd7) modified so it gernerateds a name with a unique uuid

apps/chat/agent/tools.py
stephherbers: is this better? [e3cc6b2](https://github.com/dimagi/open-chat-studio/pull/527/commits/e3cc6b2779fcef5bf56c86ee140ecc359b3516c5)

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
stephherbers: ah I didn't think about that but that makes sense. In that case how should remaining time be represented?

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
```diff
@@ -0,0 +1,121 @@
+# Generated by Django 4.2.11 on 2024-07-12 14:50
+from datetime import datetime, timedelta
+from django.db import migrations
+import json
+
+from apps.events.models import EventAction, EventActionType, ScheduledMessage
```
SmittieC, line 6: these model imports can be removed

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class Migration(migrations.Migration):
+    def migrate_periodic_tasks(apps, schema_editor):
+        ScheduledMessage = apps.get_model('events', 'ScheduledMessage')
+        EventAction = apps.get_model('events', 'EventAction')
+        Experiment = apps.get_model('experiments', 'Experiment')
+        Participant = apps.get_model('experiments', 'Participant')
+        PeriodicTask = apps.get_model('django_celery_beat', 'PeriodicTask')
+        current_time = datetime.now()
```
SmittieC, line 77: Ooh, careful here. `datetime.now()` returns a naive date, which is incompatable with timzone aware dates, so you will not be able to do for instance `task.expires - current_time`.

SmittieC (commented):
Just two comments, otherwise looks great! This has been a bit more complicated than I initially thought it would have been.

Have you tested the migration run yet?

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
snopoke: You can use `timezone.now()` (`from django.utils import timezone`)

snopoke (commented):
Just a general comment for another PR; with better tool descriptions I think we should be able to combine these two tools into one e.g. by including examples.

apps/events/migrations/0014_move_schedule_messages_off_periodic_tasks.py
stephherbers: [b2d92d1](https://github.com/dimagi/open-chat-studio/pull/527/commits/b2d92d17f732edd2bb9a9e62929eef7614ea1d16)

apps/events/models.py
```diff
@@ -305,21 +306,25 @@ def _trigger(self):
 
         self.save()
 
+    @cached_property
+    def params(self):
```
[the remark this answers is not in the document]
stephherbers: ah yes thank you

stephherbers:
@SmittieC I did test the migration locally, but if you have suggestions for testing on prod data lmk

SmittieC:
> @SmittieC I did test the migration locally, but if you have suggestions for testing on prod data lmk

I DM'd you on Slack on the process that I normally follow. Let me know if you have any questions

snopoke:
@stephherbers other than the migration testing I think this is OK to merge and my remaining comments can be addressed in a follow up PR if you'd prefer to do it that way.

stephherbers:
@snopoke @SmittieC migration tested and I agree this can be merged to unblock Simon and address the comments separately after approval

SmittieC:
> @snopoke @SmittieC migration tested and I agree this can be merged to unblock Simon and address the comments separately after approval

Thanks Steph, this is great. I took the liberty to fix the conflicting migration files, hope you don't mind.

SmittieC (approved):

PR #530, opened 2024-07-19
avoid race condition with `get_or_create`

snopoke:
https://dimagi.sentry.io/share/issue/bfb2e86fb1cc486695e5d3d749b0f5e2/


SmittieC (approved):
Nice

PR #532, opened 2024-07-19
restrict session detail view to users in the team

snopoke:
I noticed that I could view session details in teams I am not a member of


SmittieC (approved):
Good catch

PR #534, opened 2024-07-19
settings cleanup and changes in prep for AWS

snopoke:
This shouldn't change anything for current deployment.

* Make AWS env vars optional (this allows us to use IAM policies instead of access keys
* Cleanup `.env.example` and add more comments
* Add alternative settings to `.env.example`
* Support Redis TSL when not using `REDIS_URL`


SmittieC (approved):
LGTM

PR #544, opened 2024-07-23
create participant if not exists + specify platform

snopoke:
Updates to the participant data API to allow creating participants. The data structure is now as follows:

```json
{
  "identifier": "abc",
  "platform": "api",
  "data": [{"experiment": "123", "data": {"name": "joe"}}],
}
```

`platform` is now required and the URL no longer contains the identifier

I don't think anyone is using this API so I think it's safe to update. I will post on the user group though.

### DIFF
--- a/apps/api/tests/test_api.py
+++ b/apps/api/tests/test_api.py
@@ -75,18 +75,17 @@ def test_only_experiments_from_the_scoped_team_is_returned():
 
 
 @pytest.mark.django_db()
-def test_update_participant_data():
+def test_create_and_update_participant_data():
     identifier = "part1"
     experiment = ExperimentFactory(team=TeamWithUsersFactory())
     experiment2 = ExperimentFactory(team=experiment.team)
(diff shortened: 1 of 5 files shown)


apps/api/tests/test_api.py
```diff
@@ -84,11 +84,15 @@ def test_update_participant_data():
     client = ApiTestClient(user, experiment.team)
 
     # This call should create ParticipantData
-    data = [
-        {"experiment": str(experiment.public_id), "data": {"name": "John"}},
-        {"experiment": str(experiment2.public_id), "data": {"name": "Doe"}},
-    ]
-    url = reverse("api:update-participant-data", kwargs={"participant_id": participant.identifier})
+    data = {
```
SmittieC, line 87: if we remove line 82, it will test both the create and update logic, whereas now it only tests the update logic

SmittieC (approved):
LGTM

SmittieC (approved):

PR #545, opened 2024-07-23
Assistant file downloads

SmittieC:
[Issue](https://github.com/dimagi/open-chat-studio/issues/256)

I'm putting this in a draft so long, but still need to give it a final polish.

There are two types of attachments:
1. `file_citation` (for already uploaded files).
2. `file_path`: This is a generated file

Currently we store the openai file's `external_id` in the chat message's metadata in this format:
```python:
{"tool_type1": ["file_id1", "file_id2"], "tool_type2": ["file_id1", "file_id2"]}
```

I want to remove the mapping to `tool_type` and keep this a flat list of file ids. Since these files sits on the attachment on the `chat` and the attachment has a `tool_type`, we can simply look it up if we ever need it. The new structure will look like
```python:
{"openai_file_ids": ["file_id1", "file_id2"]}
```

### What references looks like
Screenshot from 2024-07-23 17-06-54


gpt_playground/settings.py
```diff
@@ -555,3 +555,5 @@
 ]
 SLACK_BOT_NAME = env("SLACK_BOT_NAME", default="@ocs")
 SLACK_ENABLED = SLACK_CLIENT_ID and SLACK_CLIENT_SECRET and SLACK_SIGNING_SECRET
+
```
SmittieC, line 558: The commit message is a bit misleading. I forgot to remove `CRYPTOGRAPHY_KEY` from it when I removed it from this file. The reason I'm putting this back is because I get invalid token issues from the cryptography library. I think since we don't load this variable in the dev environment, it causes issues. After loading it, it seems to be fixed

SmittieC:
Regarding the file references in the chat message: 

Screenshot from 2024-07-23 17-06-54

Notice that the message has a reference to it and we attach it at the bottom of the message. I'm thinking that we can try some trickery to instead of building the references to look like `[{idx}]: {file_name}`, we do this instead: `[{idx}]: ref:{file_id}`. This way we can simply find and replace this with a piece of html that renders the download tag, given the `file_id` of the file. In the raw conversation history this would look...interesting, but I'd wager it would not confuse anyone. Thoughts?

snopoke:
> Regarding the file references in the chat message:
> 
> Screenshot from 2024-07-23 [)
> 
> Notice that the message has a reference to it and we attach it at the bottom of the message. I'm thinking that we can try some trickery to instead of building the references to look like `[{idx}]: {file_name}`, we do this instead: `[{idx}]: ref:{file_id}`. This way we can simply find and replace this with a piece of html that renders the download tag, given the `file_id` of the file. In the raw conversation history this would look...interesting, but I'd wager it would not confuse anyone. This is a nice-to-have though. Thoughts?

When I did this for analysis pipelines I added a [markdown extension](https://github.com/dimagi/open-chat-studio/blob/0aa4005ee3fe64ced830376b430e8db1cbbc2a79/apps/utils/markdown.py#L26) to render the links correctly. The raw markdown looks like this:

```markdown
[link text][0]
[0]: resource:dimagi:610
```

Where `resource:dimagi:610` is `<type>:<team>:<id>`. The markdown plugin recognises the type and parses the rest to produce an actual link.

We could take the same approach here.

bderenzi:
Two quick thoughts:

1. Just double checking that we're still supporting the case where the assistant produces an output data file? 
2. One thing we played around with in another prompt was having the assistant return the page number where it found the information in the citation. I think what I'm getting at is that the prompting can change the citation response, so I'm curious if we think this will work across various citation responses

apps/service_providers/llm_service/runnables.py
```diff
@@ -267,25 +270,95 @@ def invoke(
...
-        self.state.save_message_to_history(output, ChatMessageType.AI)
+        resource_file_ids = {}
+        generated_files = []
+        citations = []
+        for idx, annotation in enumerate(annotations):
+            file_id = None
+            # The returned message might include a reference to the annotation which we should replace with
+            # something that looks better. In this case, something like "[1]".
+            file_ref_text = annotation.text
+            output = output.replace(file_ref_text, f" [{idx}]")
```
snopoke, line 322: This should replace the parentheses as well: 
```suggestion
            output = output.replace(f"({file_ref_text})", f"[{idx}]")
```

See https://www.markdownguide.org/basic-syntax/#reference-style-links

apps/service_providers/llm_service/runnables.py
```diff
@@ -267,25 +270,95 @@ def invoke(
...
+            if annotation.type == "file_citation":
+                file_citation = annotation.file_citation
+                file_id = file_citation.file_id
+                cited_file = client.files.retrieve(file_id)
+
+                if file_citation.quote:
+                    # We need to escape the brackets, since its being omitted when rendering the message as
+                    # markdown
+                    citations.append(f"\[{idx}\] {file_citation.quote}: {cited_file.filename}")
```
snopoke, line 331: Use reference style link format:
```suggestion
                    citations.append(f"\[{idx}\]: {file_citation.quote} from {cited_file.filename}")
```

apps/service_providers/llm_service/runnables.py
```diff
@@ -267,25 +270,95 @@ def invoke(
...
+                else:
+                    citations.append(f"\[{idx}\]: {cited_file.filename}")
```
snopoke, line 333: We'll need to modify this so that we can generate the correct link during markdown processing

apps/service_providers/llm_service/runnables.py
```diff
@@ -267,25 +270,95 @@ def invoke(
...
+
+        client = self.state.get_openai_assistant().client
+        page: SyncCursorPage = client.beta.threads.messages.list(response.thread_id, run_id=response.run_id)
+        # We use a run to get a response, so it's safe to assume a run will have a single message
+        message: Message = page.data[0]
+        message_content = message.content[0]
+        annotations = message_content.text.annotations
         output = response.return_values["output"]
-        self.state.save_message_to_history(output, ChatMessageType.AI)
+        resource_file_ids = {}
```
snopoke, line 314: nit: using a defaultdict would mean you don't have to check if the key is present
```suggestion
        resource_file_ids = defaultdict(list)
```

```python
 if annotation.type not in resource_file_ids:
    resource_file_ids[annotation.type] = [file_id]
else:
    resource_file_ids[annotation.type].append(file_id)
```

becomes
```python
resource_file_ids[annotation.type].append(file_id)
```

apps/service_providers/tests/test_assistant_runnable.py
```diff
@@ -211,10 +234,112 @@ def test_assistant_uploads_new_file(create_and_run, retrieve_run, list_messages,
...
+        FilePathAnnotation(
+            end_index=174,
+            file_path=FilePath(file_id="openai-file-1"),
+            start_index=134,
+            text="sandbox:/mnt/data/random_string_file.txt",
+            type="file_path",
+        ),
+        FileCitationAnnotation(
+            end_index=174,
+            file_citation=FileCitation(file_id=existing_file.external_id, quote=""),
```
snopoke, line 297: add another annotation with a quote to test that part

apps/service_providers/tests/test_assistant_runnable.py
```diff
@@ -211,10 +234,112 @@ def test_assistant_uploads_new_file(create_and_run, retrieve_run, list_messages,
...
+        ),
+    ]
+    list_messages.return_value = SyncCursorPage(
+        data=[
+            ThreadMessage(
+                id="test",
+                assistant_id=ASSISTANT_ID,
+                metadata={},
+                created_at=0,
+                content=[TextContentBlock(text=Text(annotations=annotations, value="Hi there human"), type="text")],
```
snopoke, line 310: the message text here should contain the citation references to test that they get replaced correctly

SmittieC:
Ok, took longer than I hoped, but I decided to instead use normal markdown links instead of reference links. The messages will look something like this:

image

I don't know how to get the assistant to populate the "quote" field in the `FileCitationAnnotation` object. I don't think this is needed though if we go with this approach?

apps/service_providers/llm_service/runnables.py
SmittieC: This is cool, thanks!

SmittieC:
> Two quick thoughts:
> 
> 1. Just double checking that we're still supporting the case where the assistant produces an output data file?
> 2. One thing we played around with in another prompt was having the assistant return the page number where it found the information in the citation. I think what I'm getting at is that the prompting can change the citation response, so I'm curious if we think this will work across various citation responses

1. Correct
2. Do you have an example I can look at? Generally all we get from the assistant is a file id, but we can't download the content of the file once uploaded. So if it says it got something on page 3, then we cannot download page 3. We only have a reference to the whole document. If the assistant has a link to the doc in the text, we'll replace that with a link to the doc on OCS. Does this help?

apps/service_providers/llm_service/runnables.py
```diff
@@ -267,25 +272,99 @@ def invoke(
...
+        - Those of type `file_citation` cannot be read, but since we already have those files, we should be fine
+        by only storing the reference to the file i.e. its external_id
+        - Those of type `file_path` are generated and can be downloaded. A file is created in OCS for each of these
+
+        """
+        from apps.assistants.sync import get_and_store_openai_file
+
+        client = self.state.get_openai_assistant().client
+        page: SyncCursorPage = client.beta.threads.messages.list(response.thread_id, run_id=response.run_id)
+        # We use a run to get a response, so it's safe to assume a run will have a single message
```
snopoke, line 311: Is this really safe to assume? Can a run not produce more than one message?

apps/service_providers/llm_service/runnables.py
```diff
@@ -267,25 +272,99 @@ def invoke(
...
+            file_ref_text = annotation.text
+            if annotation.type == "file_citation":
+                file_citation = annotation.file_citation
+                file_id = file_citation.file_id
+
+                file_link = ""
+                if file_id not in assistant_files_ids:
+                    # Assistant level files should have an empty link
+                    team_slug = self.state.session.team.slug
+                    file = File.objects.get(external_id=file_id, team__slug=team_slug)
```
snopoke, line 339: This could error if the file wasn't uploaded via OCS

apps/service_providers/tests/test_assistant_runnable.py
```diff
@@ -211,10 +234,100 @@ def test_assistant_uploads_new_file(create_and_run, retrieve_run, list_messages,
...
+            file_path=FilePath(file_id="openai-file-1"),
+            start_index=134,
+            text="sandbox:/mnt/data/file.txt",
+            type="file_path",
+        ),
+        FileCitationAnnotation(
+            end_index=174,
+            file_citation=FileCitation(file_id=cited_file.external_id, quote=""),
+            start_index=134,
+            text="【6:0†source】",
```
snopoke, line 275: why the special unicode `【】`? Is that what the API returns?

snopoke (commented):
Few comments. I want to go over it again tomorrow

apps/service_providers/tests/test_assistant_runnable.py
SmittieC: Correct, this is what is being returned

apps/service_providers/llm_service/runnables.py
SmittieC: Fair, I'll handle that case, thanks

apps/service_providers/llm_service/runnables.py
SmittieC: Your question made me look again deeper, but I don't find anyting saying that it can or cannot. The closest thing I can find to an answer is what LangChain is doing. Seems like they expect multiple messages in there.

```python
    def _get_response(self, run: Any) -> Any:
        # TODO: Pagination

        if run.status == "completed":
            import openai

            major_version = int(openai.version.VERSION.split(".")[0])
            minor_version = int(openai.version.VERSION.split(".")[1])
            version_gte_1_14 = (major_version > 1) or (
                major_version == 1 and minor_version >= 14
            )

            messages = self.client.beta.threads.messages.list(
                run.thread_id, order="asc"
            )
            new_messages = [msg for msg in messages if msg.run_id == run.id]
            if not self.as_agent:
                return new_messages
            print(f"New message count:{len(new_messages)}")
            answer: Any = [
                msg_content for msg in new_messages for msg_content in msg.content
            ]
            if all(
                (
                    isinstance(content, openai.types.beta.threads.TextContentBlock)
                    if version_gte_1_14
                    else isinstance(
                        content, openai.types.beta.threads.MessageContentText
                    )
                )
                for content in answer
            ):
                answer = "\n".join(content.text.value for content in answer)
```

Maybe this is a wrong assumption. I'll update the code to accommodate the possibility of multiple messages.

SmittieC: https://github.com/dimagi/open-chat-studio/pull/545/commits/e2c651632a2fd191f7e1592bdccd56ee7e5d579e

apps/service_providers/llm_service/runnables.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/545/commits/f045b62a90311d185df691e3c55e616cf21972c5 and https://github.com/dimagi/open-chat-studio/pull/545/commits/45355729981087f14e55c34614da3b3f35a000ed

apps/service_providers/llm_service/runnables.py
```diff
@@ -267,25 +272,129 @@ def invoke(
...
+        output, ai_message_resource_file_ids = self._save_response_annotations(response)
+
         if not thread_id:
             self.state.set_metadata(Chat.MetadataKeys.OPENAI_THREAD_ID, response.thread_id)
 
-        output = response.return_values["output"]
-        self.state.save_message_to_history(output, ChatMessageType.AI)
+        ai_message = self.state.save_message_to_history(output, ChatMessageType.AI)
+        if ai_message_resource_file_ids:
+            self._add_resource_files_to_message_metadata(ai_message, ai_message_resource_file_ids)
```
snopoke, line 293: I think this should be part of `save_message_to_history`:
```python
self.state.save_message_to_history(output, ChatMessageType.AI, file_ids)
```

apps/service_providers/llm_service/runnables.py
```diff
@@ -267,25 +272,129 @@ def invoke(
...
+
+        return ChainOutput(output=output, prompt_tokens=0, completion_tokens=0)
+
+    @transaction.atomic()
+    def _save_response_annotations(self, response: OpenAIAssistantFinish) -> tuple[str, dict]:
```
snopoke, line 298: I think this can just return a list of `file_ids` now since we aren't using the other info outside of this function.

apps/service_providers/llm_service/runnables.py
SmittieC: Good call

apps/service_providers/llm_service/runnables.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/545/commits/defd877b67e088c5d20b528702e478d473ea88b2

apps/service_providers/llm_service/runnables.py
```diff
@@ -307,64 +307,77 @@ def _save_response_annotations(self, response: OpenAIAssistantFinish) -> tuple[s
...
-        return output, resource_file_ids
+        return output_message, resource_file_ids
+
+    def _fetch_thread_run_messages(self, thread_id: str, run_id: str, client) -> list[Message]:
+        messages = []
+        page_kwargs = {}
+        while True:
+            page: SyncCursorPage = client.beta.threads.messages.list(thread_id, run_id=run_id, **page_kwargs)
+            messages.extend(page.data)
+            if not page.has_more:
```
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/545/commits/91bdc78496b0f5326a9572ec313d458fc0d94a1c

apps/service_providers/llm_service/runnables.py
```diff
@@ -370,11 +365,11 @@ def _get_file_name_and_link_for_citation(self, file_id: str, forbidden_file_ids:
 
         if file_id not in forbidden_file_ids:
             # Don't allow downloading assistant level files
-            team_slug = self.state.session.team.slug
+            team = self.state.session.team
             session_id = self.state.session.id
             try:
-                file = File.objects.get(external_id=file_id, team__slug=team_slug)
-                file_link = f"file:{team_slug}:{session_id}:{file.id}"
+                file = File.objects.get(external_id=file_id, team_id=team.id)
```
snopoke, line 371: just FYI: `team_id=team.id` in a query filter is equivalent to `team=team` (Django does the ID part behind the scenes)

snopoke (approved):

apps/service_providers/llm_service/runnables.py
SmittieC: Got it, thanks

PR #552, opened 2024-07-25
Searchbar for source material + fixes

SmittieC:
[Issue](https://github.com/dimagi/open-chat-studio/issues/551)


SmittieC:
When we support pgvector, I'm keen to improve our search functionality a bit if it's not a heavy lift

snopoke (approved):

PR #555, opened 2024-07-26
add 'external_id' field to schedule messages

snopoke:
This is needed to make it possible to update the messages using the ID.

The ID must be unique per (experiment, participant) combination. A short ID is assigned by default if one is not supplied.


apps/utils/slug.py
```diff
@@ -1,23 +1,80 @@
...
 
 
-def get_next_unique_slug(model_class, display_name, slug_field_name):
+def get_next_unique_slug(model_class: type[Model], input_value: str, field_name: str, model_instance=None) -> str:
     """
     Gets the next unique slug based on the name. Appends -1, -2, etc. until it finds
     a unique value.
+
+    Args:
+        model_class: The Django emodel class to check for uniqueness.
```
SmittieC, line 14:
```suggestion
        model_class: The Django model class to check for uniqueness.
```

SmittieC (dismissed):

SmittieC (approved):

PR #557, opened 2024-07-26
Update participant data UI

snopoke:
* Show all participant data even if not linked to a session e.g. created by API
* Update permission checks and show read-only version if the user does not have write permission
* Expand the first row to make the participant data more discoverable

This PR also removes the new, edit and delete views (delete was already disabled). The edit view only allowed changing the 'user' field which shouldn't be changed and I don't think it is very useful to be able to create new participants individually. In future we could add an import feature for platforms where the identifier is known e.g. whatsapp.

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -508,9 +508,15 @@ class Participant(BaseTeamModel):
                 joined_on=Subquery(joined_on),
                 last_message=Subquery(last_message),
             )
-            .filter(sessions__participant=self)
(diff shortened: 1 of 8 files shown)


SmittieC (approved):
> I'm thinking we should remove the edit view altogether, what do you think?

Maybe that would be better, especially now that we do have a way to extract specific data. I think let's do it :fire:

SmittieC (approved):

PR #561, opened 2024-07-29
Experiment Versioning Proof of Concept

stephherbers:
docs to reference are [here](https://docs.google.com/document/d/1kz_QoODtuAgQvklgMGJUKRSmt2YRApB79Kr2AQh0Wl4/edit#heading=h.z4wfl4cr8hlu) and [here](https://docs.google.com/document/d/1M8xbKdjJ61tGT5diqReoM2lbYrxL_VEuA1Dctaeshhc/edit#heading=h.ckoqzlcrkl4o)


snopoke (commented):
@stephherbers quite a few small comments. I think for the most part we want to copy in data from referenced models to the experiment version JSON. The exceptions are:

* Service models e.g. LLMProvider because any changes to them should be applied to all versions e.g. change API key. They contain sensitive data that we don't want to copy in.
* File because we can't copy them into the JSON

For these models we should have FK's or ManyToMany fields on the version model.

apps/experiments/models.py
```diff
@@ -447,6 +435,28 @@ class Meta:
         )
 
 
+class ExperimentVersion(models.Model):
+    experiment = models.ForeignKey(Experiment, on_delete=models.CASCADE, related_name="versions")
+    version_data = JSONField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    is_default = models.BooleanField(default=False)
```
SmittieC, line 442: Wondering about indexing on the `is_default` field. I can imagine us using this a lot

apps/experiments/services.py
```diff
@@ -0,0 +1,33 @@
+from django.db import transaction
+from experiments.models import Experiment, ExperimentVersion
+from experiments.serializers import populate_experiment_version_data
+
+
+def save_experiment_version(experiment: Experiment, is_default: bool = False) -> str:
+    version_data = populate_experiment_version_data(experiment)
+    if version_data:
+        last_version = ExperimentVersion.objects.filter(experiment=experiment).order_by("-version_number").first()
+        next_version_number = last_version.version_number + 1 if last_version else 1
```
SmittieC, line 10: Just something to chew on, feel free to resolve once read: Since we have a unique constraint on `experiment` and `version_number` this could potentiall cause an `IntegrityError` to be raised when someone else also tries to make a new version at the same time. Granted, it's unlikely, but not impossible.

I am not sure what the best way to handle this would be. Do we let it die and give the user a nice message, or do we somehow figure out that the `version_number` was the one that caused this, so we simply retry by incrementing the number that we wanted to add originally.

apps/experiments/serializers.py
```diff
@@ -1,28 +1,63 @@
...
+
+
+class SyntheticVoiceData(BaseModel):
+    name: str
+    neural: bool
+    language: str
+    language_code: str
+    gender: str | None
+    service: str
+    voice_provider: str | None
```
snopoke, line 37: I don't think we need this reference here. We already have the one on the experiment.

apps/experiments/serializers.py
```diff
@@ -1,28 +1,63 @@
...
+    file: str | None
```
snopoke, line 38: I'm not sure what to do with this but I do think we should keep a FK to the file to make sure we can track references.

One option is to have a field `ExperimentVersion.voice_file`.

Another is to group all the referenced files together in `ExperimentVersion.files` and keep the ID reference in the JSON data as well.

apps/experiments/serializers.py
```diff
@@ -1,28 +1,63 @@
...
+    confirmation_text: str
+
+
+class ConsentFormData(BaseModel):
+    name: str
+    consent_text: str
+    capture_identifier: bool
+    identifier_label: str
+    identifier_type: str
+    is_default: bool
```
snopoke, line 26: not that important but I don't think we need this here

apps/experiments/models.py
snopoke: It would need to be a partial index. In Django you can do that by creating an Index with a condition: https://docs.djangoproject.com/en/5.0/ref/models/indexes/#condition

SmittieC:
We can close this PR, yes?

snopoke:
I think so yes

PR #563, opened 2024-07-30
Filter URLs and emojis for TTS

SmittieC:
https://github.com/dimagi/open-chat-studio/issues/554

The following happens to the text that will be used to synthesize the bot's response:
- Emojis are filtered out
- URLs are extracted, deduplicated and sent as a separate text message

Note that this does not change the chat history messages, only the one being used for TTS

### DIFF
--- a/apps/chat/channels.py
+++ b/apps/chat/channels.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from abc import ABC, abstractmethod
 from enum import Enum
 from functools import cached_property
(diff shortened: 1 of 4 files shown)


SmittieC:
> One thing I'm still to test is to replace the URLs with a reference (say, a number), and then in the text message include the references. This might sound a bit weird, but let's see

Using numbers to replace the urls with as references doesn't work well in my opinion. It will just confuse people. To make it sound more natural, we could try to use link references with the words like "see link <link number>" and number the links in the text message, but this means extra LLM calls. Q: How many extra? A: The amount of links in the text.

Suggestions welcome

apps/chat/channels.py
```diff
@@ -35,6 +37,21 @@
...
+def strip_urls_and_emojis(text: str) -> tuple[str, list[str]]:
+    """Strips any URLs in `text` and appends them to the end of the text. Emoji's are filtered out"""
+    text = emoji.replace_emoji(text, replace="")
+
+    url_pattern = re.compile(URL_REGEX)
+    urls = url_pattern.findall(text)
+    for url in urls:
+        text = text.replace(url, "")
+
+    return text, urls
```
snopoke, line 53: you should deduplicate the urls: `list(set(urls))`

SmittieC: I orignally did that, but thought "what if the bot references a URL now, then another two and then this one again"? Keeping the referenced URLs in order might help users a bit to know which URL the bot is referring to, but maybe we don't want to give users this potential cognative load where they have to keep track of where a url was referenced, then cross check that with the given list. Doesn't quite make sense. Back to deduplication!

SmittieC: https://github.com/dimagi/open-chat-studio/pull/563/commits/9a13579bad6754d2269c6311be6fa3a182fc46f4

snopoke: I see your point. Either way it's probably unlikely that there will be duplicates.

snopoke (approved):

PR #570, opened 2024-08-01
Confirm experiment deletion

SmittieC:
*advert music playing*

Are you afraid to edit an experiment for fear that you might accidentally click the `trash` icon next to it? Fear no more! We'll ask you to confirm deletion!

*crowd cheers* :partying_face:

### DIFF
--- a/apps/experiments/views/experiment.py
+++ b/apps/experiments/views/experiment.py
@@ -415,8 +415,7 @@ def _get_voice_provider_alpine_context(request):
 def delete_experiment(request, team_slug: str, pk: int):
     safety_layer = get_object_or_404(Experiment, id=pk, team=request.team)
     safety_layer.delete()
-    messages.success(request, "Experiment Deleted")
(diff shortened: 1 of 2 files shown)


apps/experiments/views/experiment.py
```diff
@@ -415,8 +415,7 @@ def _get_voice_provider_alpine_context(request):
 def delete_experiment(request, team_slug: str, pk: int):
     safety_layer = get_object_or_404(Experiment, id=pk, team=request.team)
     safety_layer.delete()
-    messages.success(request, "Experiment Deleted")
```
SmittieC, line 418: This message shows up very briefly when we use the `HX-Redirect` approach. I _think_ it might be because htmx gets the response and redirects the page again to the url in the HX-Redirect header. This would explain why the toast shows up and disappears again.

snopoke (commented):
Great!

snopoke (approved):

PR #572, opened 2024-08-01
Do not show a channel if the provider isn't configured

SmittieC:
Stems from [this issue](https://dimagi.sentry.io/issues/5664583590/?referrer=issue_alert-slack&notification_uuid=91e5f819-bb2c-42b8-aac9-7360616dbae6&alert_rule_id=14063560&alert_type=issue). We now only make channels available if a provider for it is configured. Doesn't really make sense to allow users to try and add them if they cannot do it in anycase. So users will not see for instance "Whatsapp" if turn.io or twilio is not configured

image

### DIFF
--- a/apps/channels/models.py
+++ b/apps/channels/models.py
@@ -1,4 +1,5 @@
 import uuid
+from collections import OrderedDict
 
 from django.conf import settings
 from django.db import models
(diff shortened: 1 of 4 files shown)


apps/channels/models.py
```diff
@@ -33,21 +34,26 @@ class ChannelPlatform(models.TextChoices):
...
-        """
+    def for_dropdown(cls, used_platforms, team) -> OrderedDict:
+        """Returns a dictionary of available platforms for this team. Available platforms will have a `True` value"""
         from apps.service_providers.models import MessagingProvider
 
-        available_platforms = [cls.TELEGRAM]
-        if settings.SLACK_ENABLED:
-            available_platforms.append(cls.SLACK)
+        all_platforms = cls.as_list(exclude=[cls.API, cls.WEB])
+        platform_availability = OrderedDict.fromkeys(all_platforms, value=False)
```
snopoke: no need to use `OrderedDict`, you can use `dict.fromkeys`. Dictionaries are always ordered.

apps/experiments/views/experiment.py
```diff
@@ -448,10 +448,10 @@ def single_experiment_home(request, team_slug: str, experiment_id: int):
     )
     channels = experiment.experimentchannel_set.exclude(platform__in=[ChannelPlatform.WEB, ChannelPlatform.API]).all()
     used_platforms = {channel.platform_enum for channel in channels}
-    available_platforms = set(ChannelPlatform.for_dropdown(experiment.team)) - used_platforms
+    available_platforms = ChannelPlatform.for_dropdown(used_platforms, experiment.team)
     platform_forms = {}
     form_kwargs = {"team": request.team}
-    for platform in available_platforms:
+    for platform in available_platforms.keys():
```
snopoke, line 454: no need to use `.keys()`, iterating over a dict will iterate over the keys already

snopoke (dismissed):

apps/channels/models.py
snopoke: I see you changed this in a later commit

apps/experiments/views/experiment.py
SmittieC: [changed](https://github.com/dimagi/open-chat-studio/pull/572/commits/cb0498ef7d72b9e20e252eed6e14d805ad1dd0a0)

bderenzi:
this looks like a really nice improvement from the bot-author perspective!

snopoke (approved):

PR #575, opened 2024-08-02
Save pipeline nodes as a model

proteusvacuum:
This saves the nodes coming from the frontend as a separate model to allow us to migrate these if ever anything changes in the required parameters, the definitions of nodes, etc. 

I still keep the node data that comes from the frontend as json in `Pipeline.data`, although it isn't used anywhere any more.


apps/pipelines/migrations/0005_auto_20240802_0039.py
```diff
@@ -0,0 +1,26 @@
...
+            node_object.label = node.data.label
+            node_object.save()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("pipelines", "0004_node"),
+    ]
+
+    operations = [migrations.RunPython(_migrate_nodes, reverse_code=migrations.RunPython.noop)]
```
SmittieC, line 26: nit: can't we just delete all `Node` objects in the down migration?

SmittieC (dismissed):
LGTM

snopoke (approved):

apps/pipelines/migrations/0005_auto_20240802_0039.py
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/575/commits/707b5c7391b471d9a8b92c48dbf1be649107b3c3

apps/pipelines/models.py
```diff
@@ -27,14 +36,57 @@ def __str__(self):
...
+        to_delete = current_ids - new_ids
+        Node.objects.filter(pipeline=self, flow_id__in=to_delete).delete()
+
+        # Set new nodes or update existing ones
+        for node in nodes:
+            node_object, _ = Node.objects.get_or_create(pipeline=self, flow_id=node.id)
+            node_object.type = node.data.type
+            node_object.params = node.data.params
+            node_object.label = node.data.label
+            node_object.save()
```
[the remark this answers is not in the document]
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/575/commits/965bf414f2c30461a4ee061da1a126567161db38

apps/pipelines/graph.py
```diff
@@ -4,6 +4,7 @@
 from pydantic_core import ValidationError
 
 from apps.pipelines.exceptions import PipelineBuildError, PipelineNodeBuildError
+from apps.pipelines.models import Pipeline
 
 
 class Node(pydantic.BaseModel):
```
snopoke, line 10: Is this class still necessary?

Issue #559, opened 2024-07-26
Load experiment data in participant view asynchronously


snopoke:
Loading the participant sessions and data is bit slow. It would be better to load the data using HTMX after the page has loaded. We could also only load the data for each experiment when the user expands the UI section. Which would mean we only load one set of data at a time.

PR #576, opened 2024-08-02
Lazy load participant data per experiment

SmittieC:
https://github.com/dimagi/open-chat-studio/issues/559

### A mini demo (look at the network tab as well)

mini_demo


snopoke (approved):

SmittieC:
This has halved the time it takes to load for instance my participant data in the dimagi team. From 6 down to 3.

Issue #579, opened 2024-08-05
File download from OpenAI is returning corrupt files


bderenzi:
When an OpenAI Assistant creates a file, we offer a download link to the user. Unfortunately, the contents of the file are corrupted. Viewing the same thread on the OpenAI site and downloading from there works correctly

See this slack thread for more details:
https://dimagi.slack.com/archives/C04VBQJ1DL3/p1722854203149329

SmittieC:
https://github.com/dimagi/open-chat-studio/pull/582

PR #582, opened 2024-08-05
Fix corrupt files

SmittieC:
https://github.com/dimagi/open-chat-studio/issues/579


bderenzi (approved):
Thanks for the fast work!

PR #584, opened 2024-08-05
only create new participant data for the current experiment

snopoke:
I noticed that we are creating participant data for every experiment in the team via the `update_memory` method which I don't think is correct. Rather we should updated existing participant data and only create new participant data for the current experiment.

This should avoid creating participant data where it's not needed.

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -12,7 +12,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres.fields import ArrayField
 from django.core.validators import MaxValueValidator, MinValueValidator, validate_email
 from django.db import models, transaction
-from django.db.models import Count, OuterRef, Prefetch, Q, Subquery
+from django.db.models import Count, OuterRef, Q, Subquery
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext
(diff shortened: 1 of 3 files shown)


apps/experiments/models.py
```diff
@@ -530,36 +530,31 @@ def get_data_for_experiment(self, experiment) -> dict:
             return {}
 
     @transaction.atomic()
-    def update_memory(self, data: dict, experiment: Experiment | None = None):
+    def update_memory(self, data: dict, experiment: Experiment):
         """
         Updates this participant's data records by merging `data` with the existing data. By default, data for all
-        experiments the this participant participated in will be updated. If there are no records for a specific
+        experiments that this participant participated in will be updated. If there are no records for a specific
```
SmittieC, line 536: The description is not 100% true anymore :smiley_cat:

apps/experiments/models.py
```diff
@@ -530,36 +530,31 @@ def get_data_for_experiment(self, experiment) -> dict:
...
-                records_to_update.append(participant_data)
-            else:
-                ParticipantData.objects.create(team=self.team, content_object=experiment, data=data, participant=self)
-
-        ParticipantData.objects.bulk_update(records_to_update, fields=["data"])
+        # Update all existing records
+        participant_data = ParticipantData.objects.filter(participant=self).select_for_update()
+        experiments = set()
+        with transaction.atomic():
+            experiment_content_type = ContentType.objects.get_for_model(Experiment)
```
SmittieC, line 549: super nit (feel free to ignore, since this will not have a performance impact at this stage): we can update the pariticipant data query like this to avoid fetching unwanted results:

```python
experiment_content_type = ContentType.objects.get_for_model(Experiment)
ParticipantData.objects.filter(participant=self, content_type=experiment_content_type).select_for_update()
```

SmittieC (approved):

PR #585, opened 2024-08-05
Fix: Do not save ouput of processor bot and do not save input of terminal bot

SmittieC:
It would help if I refresh my chats a bit more locally. We should not be saving the input to the terminal bot and the output of the "processor" bots as messages when there is a terminal bot.


apps/chat/bots.py
```diff
@@ -103,19 +104,29 @@ def _call_predict(self, input_str, save_input_to_history=True, attachments: list
...
+                    "save_output_to_history": self.terminal_chain is None,
                     "experiment_tag": tag,
                 }
             },
             attachments=attachments,
         )
 
         if self.terminal_chain:
-            result = self.terminal_chain.invoke(result.output)
+            result = self.terminal_chain.invoke(
```
SmittieC, line 121: It feels wrong to not save the output/input, but monitoring tools will pick the inputs/outputs up when we have it

apps/chat/bots.py
```diff
@@ -103,19 +104,29 @@ def _call_predict(self, input_str, save_input_to_history=True, attachments: list
...
+                result.output,
+                config={
+                    "configurable": {
+                        "save_input_to_history": False,
+                        "experiment_tag": self.terminal_chain_tag,
```
snopoke, line 126: Should we still not tag it with the 'processor' chain tag since that's the one that actually responded right? If we end up with multiple terminal bots we could pass a list of tags here, one for the processor and one for the terminal.

SmittieC: Ah yes you are correct

SmittieC:
This is ready for review again cc @snopoke

snopoke (approved):

PR #589, opened 2024-08-07
Allow public users to download their files and generated files

SmittieC:
## Description

This update will allow public users to download the files they uploaded during a conversation as well as those generated by the assistant.

Although this change works, I am a bit uneasy about the approach I took. I think it's because I'm "breaking" tradition here by allowing session id's to be integers (i.e. db primary keys) :shrug:.

Q: Why not just use the session's external id and pass in the experiment's public key to the `download_file` endpoint?
A: Doing that will break existing links to files, since they are embedded in chat messages and use integer ids. When a user wants to continue the chat and download the file, they wouldn't be able to anymore.

If we want to keep all public facing endpoints using external ids / public ids (so revert the changes I made to the decorator), then a possible solution that will require a migration is to find all chat messages that contain links and replace the links with the new format. Finding these messages should be relatively easy, since we can just find all chats with a "file_path" attachment, then in that chat find all chat messages having a file external id in its metadata that matches any file in this attachment. From here we can simply find and replace.

Question is: Are we fine with this approach here?

## User Impact

Public users will be able to download their own uploaded files as well as generated ones.

### Demo

N/A

### Docs

N/A

### DIFF
--- a/apps/experiments/decorators.py
+++ b/apps/experiments/decorators.py
@@ -2,11 +2,12 @@ from functools import wraps
 
 from django.contrib import messages
 from django.core import signing
+from django.db.models import Q
 from django.http import Http404, HttpResponseRedirect
 from django.shortcuts import get_object_or_404
 from django.urls import reverse
 
-from apps.experiments.models import Experiment, ExperimentSession, SessionStatus
+from apps.experiments.models import ExperimentSession, SessionStatus
 
 MAX_AGE = 180 * 24 * 60 * 60  # 6 months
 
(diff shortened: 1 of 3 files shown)


snopoke (commented):
I don't think this is terrible but it would be better to use the external IDs

SmittieC:
> I don't think this is terrible but it would be better to use the external IDs

Are you then advocating for rather using external ids (which would mean a migration as well)?

SmittieC:
> > > I don't think this is terrible but it would be better to use the external IDs
> > 
> > 
> > Are you then advocating for rather using external ids (which would mean a migration as well)?
> 
> I am but I'm not mandating it.

Ok. Unless we really want this in, I'm holding off on this for now until we know more. See [this thread](https://dimagi.slack.com/archives/C05PK63Q89H/p1723024763533049?thread_ts=1722949203.739729&cid=C05PK63Q89H)

stephherbers (approved):
overall I agree with Simon, but fine with revising if possible. 
nit: this can have doc updates, no?

SmittieC:
> overall I agree with Simon, but fine with revising if possible. nit: this can have doc updates, no?

I think so yeah.

SmittieC:
Closing this until we know we want to go this route. We can easily find it again.

PR #599, opened 2024-08-12
Some updates to agent tools

SmittieC:
## Description

- The tool to update scheduled messages should use the `message_id` param to find the message. (this is a fix)
- Users should be able to update their own scheduled messages.
- Updated what info is included in the scheduled message string representation
- Added missing tests for agent tools
- Lots of participant data will now cause scrolling in the session view where users can see the participant data

## User Impact

- Schedule message updates are fixed.

### Demo

Nothing really demo-worthy

### Docs

- Will add and link to docs once this is merged


apps/chat/agent/tools.py
```diff
@@ -97,15 +97,15 @@ def action(
         minute: int,
         user_specified_custom_date: bool,
     ):
-        message = ScheduledMessage.objects.get(participant=self.experiment_session.participant, external_id=message_id)
-        if message.was_created_by_system and user_specified_custom_date:
```
snopoke: why revert this?

SmittieC: I wanted the user to not be able to specify a new date, but went against that again. https://github.com/dimagi/open-chat-studio/pull/599/commits/0e6447c08f989c55d80630da664407a35209629f

apps/chat/agent/tools.py
```diff
@@ -79,9 +79,9 @@ def action(
         )
 
 
-class UpdateScheduledMessageTool(CustomBaseTool):
-    name = AgentTools.SCHEDULE_UPDATE
-    description = "useful to update the schedule of a scheduled message. Use only to update existing schedules"
+class MoveScheduledMessageDateTool(CustomBaseTool):
```
SmittieC, line 82: I think this reflects what this tool does more accurately

snopoke (approved):

PR #600, opened 2024-08-12
Warn the user if the whatsapp number as not found at the provider

SmittieC:
## Description

Warn the user if the whatsapp number as not found at the provider

## User Impact

Users will not be blocked if we cannot verify the number at the provider (in this case Twilio), but they will be warned and encouraged to double check

### Demo

demo

### Docs

Will update on merge


snopoke (dismissed):

SmittieC:
Sorry, I removed the "form.is_valid()" in the test which is needed to actually call `self.clean`. Fixed now

PR #603, opened 2024-08-13
Assistants should also save router tags

SmittieC:
## Description

1. Assistants didn't save router tags. That's fixed now.

## User Impact

1. Users will be able to see which assistant child bot generated an answer

### Demo

No demo needed, but here's a screenshot

image

### Docs

Release docs will be updated on release

### DIFF
--- a/templates/experiments/single_experiment_home.html
+++ b/templates/experiments/single_experiment_home.html
@@ -260,9 +260,7 @@
     <input type="radio" name="tab_group" role="tab" class="tab" aria-label="Routes" id="tab-routes"/>
     <div role="tabpanel" class="tab-content" id="content-routes">
       <div class="app-card">
-        {% if experiment.assistant %}
-          Assistants cannot be used as a router bot
-        {% elif can_make_child_routes %}
+        {% if can_make_child_routes %}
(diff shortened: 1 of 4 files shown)


templates/experiments/single_experiment_home.html
```diff
@@ -260,9 +260,7 @@ <h3 class="font-bold text-lg">Parent Experiments:
     <input type="radio" name="tab_group" role="tab" class="tab" aria-label="Routes" id="tab-routes"/>
     <div role="tabpanel" class="tab-content" id="content-routes">
       <div class="app-card">
-        {% if experiment.assistant %}
-          Assistants cannot be used as a router bot
-        {% elif can_make_child_routes %}
+        {% if can_make_child_routes %}
```
SmittieC, line 263: reverted this for another PR. There might be assistant bots that are already routing bots. We first need people to fix that

bderenzi: Is it easy to get a list of experiments & team spaces where we have assistants as routers? I would guess it's just the dimagi team and we could make the change pretty quickly

snopoke (approved):

PR #604, opened 2024-08-13
Experiment Versioning Proof of Concept (2)

stephherbers:
## Description

This version (ha) copies the Experiment model to make an immutable snapshot deemed with a "Released" or "Deprecated" statu with the original (I'm calling "core" to keep the parent/child language just to routing) will always remain a "Draft". See [bottom of doc ](https://docs.google.com/document/d/1M8xbKdjJ61tGT5diqReoM2lbYrxL_VEuA1Dctaeshhc/)for more thoughts.

### DIFF
--- a/apps/events/tasks.py
+++ b/apps/events/tasks.py
@@ -14,7 +14,7 @@ def enqueue_static_triggers(session_id, trigger_type):
     session = ExperimentSession.objects.get(id=session_id)
     experiment_id = session.experiment_id
     if session.experiment.version_number:
-        experiment_id = session.experiment.core_id
+        experiment_id = session.experiment.working_experiment
 
     trigger_ids = StaticTrigger.objects.filter(experiment_id=experiment_id, type=trigger_type).values_list(
         "id", flat=True
--- a/apps/events/tasks.py
+++ b/apps/events/tasks.py
@@ -12,8 +12,11 @@ logger = logging.getLogger(__name__)
 @shared_task(ignore_result=True)
 def enqueue_static_triggers(session_id, trigger_type):
     session = ExperimentSession.objects.get(id=session_id)
+    experiment_id = session.experiment_id
+    if session.experiment.version_number:
+        experiment_id = session.experiment.core_id
 
-    trigger_ids = StaticTrigger.objects.filter(experiment_id=session.experiment_id, type=trigger_type).values_list(
+    trigger_ids = StaticTrigger.objects.filter(experiment_id=experiment_id, type=trigger_type).values_list(
         "id", flat=True
     )
     for trigger_id in trigger_ids:
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -294,7 +294,7 @@ class Experiment(BaseTeamModel):
         choices=ExperimentStatus.choices,
         default=ExperimentStatus.DRAFT,
     )
-    parent_id = models.ForeignKey(
+    core_id = models.ForeignKey(
         "self",
         on_delete=models.CASCADE,
         null=True,
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -455,12 +455,18 @@ class ExperimentRoute(BaseTeamModel):
     type = models.CharField(choices=ExperimentRouteType.choices, max_length=64, default=ExperimentRouteType.PROCESSOR)
     condition = models.CharField(max_length=64, blank=True)
 
+    parent_version = models.PositiveIntegerField(null=True, blank=True)
+    child_version = models.PositiveIntegerField(null=True, blank=True)
+
     @classmethod
     def eligible_children(cls, team: Team, parent: Experiment | None = None):
-        """Returns a list of experiments: that are not parents, and are not children of the current experiment"""
+        """Returns a list of experiments: that are not parents, and are not children of the current experiment
+        If the parent is a released version, the child must also be a released version"""
+
         parent_ids = cls.objects.filter(team=team).values_list("parent_id", flat=True).distinct()
 
         if parent:
+            parent_is_released = parent.status == "Released"
             child_ids = cls.objects.filter(parent=parent).values_list("child_id", flat=True)
             eligible_experiments = (
                 Experiment.objects.filter(team=team)
@@ -468,6 +474,9 @@ class ExperimentRoute(BaseTeamModel):
                 .exclude(id__in=parent_ids)
                 .exclude(id=parent.id)
             )
+            if parent_is_released:
+                # Ensure child experiments are also 'Released' if parent is
+                eligible_experiments = eligible_experiments.filter(status="Released")
         else:
             eligible_experiments = Experiment.objects.filter(team=team).exclude(id__in=parent_ids)
 
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -275,6 +275,12 @@ class AgentTools(models.TextChoices):
     SCHEDULE_UPDATE = "schedule-update", gettext("Schedule Update")
 
 
+class ExperimentStatus(models.TextChoices):
+    DRAFT = "Draft", "Draft"
+    RELEASED = "Released", "Released"
+    DEPRECATED = "Deprecated", "Deprecated"
+
+
 @audit_fields(*model_audit_fields.EXPERIMENT_FIELDS, audit_special_queryset_writes=True)
 class Experiment(BaseTeamModel):
     """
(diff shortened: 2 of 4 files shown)


apps/experiments/models.py
```diff
@@ -433,19 +455,28 @@ class ExperimentRoute(BaseTeamModel):
     type = models.CharField(choices=ExperimentRouteType.choices, max_length=64, default=ExperimentRouteType.PROCESSOR)
     condition = models.CharField(max_length=64, blank=True)
 
+    parent_version = models.PositiveIntegerField(null=True, blank=True)
```
SmittieC, line 458: Are we not also copying these experiment routes? (I can't remember if we said we're going to do it or not)

apps/experiments/views/experiment_routes.py
```diff
@@ -42,6 +42,12 @@ def get_success_url(self):
     def form_valid(self, form):
         form.instance.team = self.request.team
         self.object = form.save(commit=False)
+
+        # Set parent_version and child_version
+        parent_experiment = get_object_or_404(Experiment, id=self.kwargs["experiment_id"])
+        self.object.parent_version = parent_experiment.version_number
+        self.object.child_version = Experiment.objects.get(id=self.object.child_id).version_number
```
snopoke, line 49: This is something to talk about. Should the 'core' route config always point to the 'core' version of the linked bots?

When you create a new version of the route bot you could have options to use the 'live' version of the linked bots or to select a specific version for each?

stephherbers: to clarify, is this the options you are thinking for what the routes are referencing for the experiment (I'll use "working" to describe the draft exp state):
1. parent: working -> child: working
2. parent: working -> child: versioned
3. parent: versioned -> child: versioned

right now are just options 1 and 3, but option 2 also makes sense to me to add

apps/experiments/models.py
```diff
@@ -283,6 +289,22 @@ class Experiment(BaseTeamModel):
     """
 
     objects = ExperimentObjectManager()
+    status = models.CharField(
+        max_length=10,
+        choices=ExperimentStatus.choices,
+        default=ExperimentStatus.DRAFT,
+    )
+    core_id = models.ForeignKey(
```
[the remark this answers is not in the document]
stephherbers: cool I'll switch it to one of those 👍

apps/experiments/views/experiment_routes.py
snopoke: 1 & 3 is great

stephherbers: ah gotcha yes that was an oversight: https://github.com/dimagi/open-chat-studio/pull/604#:~:text=%E2%80%A6-,3b3c977,-when%20creating%20a

mods will still need to be changed here

apps/experiments/models.py
```diff
@@ -283,6 +289,22 @@ class Experiment(BaseTeamModel):
     """
 
     objects = ExperimentObjectManager()
+    status = models.CharField(
```
SmittieC, line 292: How are we keeping track of the "active" version? By default an experiment is in "Draft" and when we make a version it becomes "released", but we can end up with multiple released versions yes?

apps/experiments/views/experiment_routes.py
```diff
@@ -43,10 +43,11 @@ def form_valid(self, form):
...
 
-        # Set parent_version and child_version
-        parent_experiment = get_object_or_404(Experiment, id=self.kwargs["experiment_id"])
-        self.object.parent_version = parent_experiment.version_number
-        self.object.child_version = Experiment.objects.get(id=self.object.child_id).version_number
+        # Set parent_version and child_version if experiment is versioned
+        experiment = Experiment.objects.get(id=self.kwargs["experiment_id"])
+        if experiment.status == "Released":
+            self.object.parent_version = experiment.version_number
+            self.object.child_version = Experiment.objects.get(id=self.object.child_id).version_number
```
snopoke, line 50: This code seems fine but I don't think it makes sense here in the 'route create view' since you'll never be creating new routes for released experiments right?

apps/experiments/utils.py
```diff
@@ -30,3 +34,15 @@ def calculate_version_number(parent_experiment_id):
...
+
+
+def copy_event_actions(experiment_id):
+    event_actions = EventAction.objects.filter(experiment_id=experiment_id)
+    for event_action in event_actions:
+        event_action.pk = None
+        if "pipeline_id" in event_action.params:
+            old_pipeline_id = event_action.params["pipeline_id"]
+            old_pipeline = Pipeline.objects.get(id=old_pipeline_id)
+            new_pipeline = Pipeline.objects.create(name=f"{old_pipeline.name} (copy)", data=old_pipeline.data)
```
snopoke, line 46: We'll need to copy and update the `Node` model instances too.

We'll definitely want to 'tag' these as copies and prevent them from showing in the UI by default (and make them readonly). In the long run we should apply a similar version scheme to these pipelines and then we can apply a similar mechanism as routes where the user can choose what to do (create a new version or use an existing one).

apps/events/tasks.py
```diff
@@ -12,8 +12,11 @@
 @shared_task(ignore_result=True)
 def enqueue_static_triggers(session_id, trigger_type):
     session = ExperimentSession.objects.get(id=session_id)
+    experiment_id = session.experiment_id
+    if session.experiment.version_number:
+        experiment_id = session.experiment.working_experiment
```
snopoke, line 17: If we're copying the triggers shouldn't this load the ones related to the versioned experiment?

apps/experiments/models.py
```diff
@@ -275,6 +275,12 @@ class AgentTools(models.TextChoices):
     SCHEDULE_UPDATE = "schedule-update", gettext("Schedule Update")
 
 
+class ExperimentStatus(models.TextChoices):
+    DRAFT = "Draft", "Draft"
+    RELEASED = "Released", "Released"
+    DEPRECATED = "Deprecated", "Deprecated"
```
snopoke, line 281: nit: use lowercase for slugs

```suggestion
    DRAFT = "draft", "Draft"
    RELEASED = "released", "Released"
    DEPRECATED = "deprecated", "Deprecated"
```

apps/experiments/models.py
```diff
@@ -283,6 +289,22 @@ class Experiment(BaseTeamModel):
...
+        "self",
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+        related_name="experiment_versions",
+    )
+    version_number = models.PositiveIntegerField(
+        null=True,
+        blank=True,
+    )
```
snopoke, line 307: note for future: it will be useful to be able to tell if edits have been made since the last 'version'. I see two options:

* A boolean: `has_changed` which get's set to `True` on every change
* An int: `patch_version` which get's incremented on every save

apps/experiments/models.py
```diff
@@ -433,19 +455,28 @@ class ExperimentRoute(BaseTeamModel):
     type = models.CharField(choices=ExperimentRouteType.choices, max_length=64, default=ExperimentRouteType.PROCESSOR)
     condition = models.CharField(max_length=64, blank=True)
 
+    parent_version = models.PositiveIntegerField(null=True, blank=True)
+    child_version = models.PositiveIntegerField(null=True, blank=True)
```
snopoke, line 459: Are these necessary? If we make a copy of the route and update the `parent` and `child` reference to point to the versioned experiments then I don't think we need these.

apps/experiments/models.py
```diff
@@ -433,19 +455,28 @@ class ExperimentRoute(BaseTeamModel):
...
     @classmethod
     def eligible_children(cls, team: Team, parent: Experiment | None = None):
-        """Returns a list of experiments: that are not parents, and are not children of the current experiment"""
+        """Returns a list of experiments: that are not parents, and are not children of the current experiment
+        If the parent is a released version, the child must also be a released version"""
+
         parent_ids = cls.objects.filter(team=team).values_list("parent_id", flat=True).distinct()
 
         if parent:
+            parent_is_released = parent.status == "Released"
```
snopoke, line 469: nit: make an `is_released` property on the experiment model

apps/experiments/views/experiment_routes.py
stephherbers: ah yes correct

SmittieC:
Closing this PR just to clean up a bit. The new one can be found [here](https://github.com/dimagi/open-chat-studio/pull/671)

PR #608, opened 2024-08-15
Chat to a pipeline

proteusvacuum:
## Description

This allows connecting an Experiment with a Pipeline. I went with creating the Pipeline runnable, as it seemed like the simplest option, although that required me to make the changes in 41e59850f27205029620ab89236126afc60557ae, which I'd love feedback on (the TopicBot requires an LLM provider to count tokens, etc.). I think the other option, making a `PipelineBot` might allow us to bypass that.

## User Impact

Chat to a pipeline! Once we have optional edges, you could then make router bots, safety bots, and any other complex-type bot :) 

I renamed the `CreateReport` node - I'm not sure if anyone has created pipelines in production with a node like that - if so I'll write a migration to rename those nodes. 

### Demo

[Screencast from 2024-08-14 22-54-14.webm](https://github.com/user-attachments/assets/6c04797b-4a0d-454b-90e1-e88cb5abdc9a)


bderenzi:
(very cool!)

proteusvacuum:
I'm working on getting branching to work in the pipelines, but will do that in a separate PR.

apps/pipelines/nodes/nodes.py
```diff
@@ -51,32 +60,68 @@ def _process(self, state: PipelineState) -> PipelineState:
...
     def _process(self, state: PipelineState) -> PipelineState:
-        chain = PromptTemplate.from_template(template=self.prompt) | super().get_chat_model()
-        output = chain.invoke(state["messages"][-1], config=self._config)
+        prompt = PromptTemplate.from_template(template=self.prompt)
+        context = self._get_context(state, prompt)
+        chain = prompt | super().get_chat_model()
+        output = chain.invoke(context, config=self._config)
         return output.content
 
+    def _get_context(self, state: PipelineState, prompt: PromptTemplate):
```
SmittieC, line 94: I'm wondering if this context shouldn't come from the [state](https://github.com/dimagi/open-chat-studio/blob/b88c5797dad31fb945d34a7661287b6e92331366/apps/service_providers/llm_service/state.py#L83-L98) that we use for the normal runnables in order to try and minimize duplication when trying to support experiment and pipeline based bots. It would have to be changed a bit, since the state expects the experiment as input as well.

apps/pipelines/models.py
```diff
@@ -116,6 +124,26 @@ def invoke(self, input: PipelineState, session: ExperimentSession | None = None)
...
+            session=session,
+        )
+
+    def _save_message_to_history(self, session: ExperimentSession, message: str, type_: ChatMessageType):
+        ChatMessage.objects.create(
+            chat=session.chat,
+            message_type=type_.value,
+            content=message,
+        )
+        # TODO: Add tags here?
```
SmittieC, line 145: We add tags for the router child bot (the "conditional" bot). I think it's fine to leave this as-is and only add tags here when we have support for branching

SmittieC:
My thoughts:

In my mind, the`TopicBot` class is a hardcoded pipeline. It's running each "bot"/experiment (child bot, terminal bot, main bot) as a separate runnable. I like the new `PipelineBot` class, I think this makes that distinction clear. I am thinking though that the `LLMResponseWithPrompt` should also invoke the runnables like we do in the `TopicBot` class, but ofcourse not in the same "hardcoded" way.

What I am suggesting is that we do the same as [what we're doing here](https://github.com/dimagi/open-chat-studio/blob/ca65ca4b2a55bd4c66ab3886ea794a3fb6c0baab/apps/chat/bots.py#L94) by creating a runnable and invoking that, but only now we should be creating and running the runnable inside the `_process` method of the `LLMResponseWithPrompt`.

We can also add a [new state ](https://github.com/dimagi/open-chat-studio/blob/b88c5797dad31fb945d34a7661287b6e92331366/apps/service_providers/llm_service/state.py#L16) for the pipeline runnable. This way we can get things like the chat model, the prompt etc from the pipeline instead of the experiment.

I hope this makes sense? I'm not sure if this is the correct way to go, but I have a feeling that it is.

Thoughts? cc @snopoke

apps/pipelines/models.py
snopoke: I'm not sure that tagging makes sense for pipelines at all (unless we make it a config option for the nodes).

An alternate approach would be to link the chat messages to the pipeline run which is a bit like a trace.

apps/pipelines/models.py
```diff
@@ -95,26 +96,25 @@ def get_llm_service(self):
...
-            status=PipelineRunStatus.RUNNING,
-            log={"entries": []},
-            session=session,
-        )
-
+        pipeline_run = self._create_pipeline_run(input, session)
         logging_callback = PipelineLoggingCallbackHandler(pipeline_run)
+
+        if save_run_to_history and session is not None:
+            self._save_message_to_history(session, input["messages"][-1], ChatMessageType.HUMAN)
```
snopoke, line 109: do you think this belongs here or in the bot class?

apps/pipelines/nodes/nodes.py
snopoke: I think it's fine for now, if it was exact duplicate I'd say yes but it would be a bit different since the source material is linked to the node. But I guess you could extract a utility function that's used in both places:

```python
def get_prompt_context(session, prompt, source_material_id):
```

snopoke:
> My thoughts:
> 
> In my mind, the`TopicBot` class is a hardcoded pipeline. It's running each "bot"/experiment (child bot, terminal bot, main bot) as a separate runnable. I like the new `PipelineBot` class, I think this makes that distinction clear. I am thinking though that the `LLMResponseWithPrompt` should also invoke the runnables like we do in the `TopicBot` class, but ofcourse not in the same "hardcoded" way.
> 
> What I am suggesting is that we do the same as [what we're doing here](https://github.com/dimagi/open-chat-studio/blob/ca65ca4b2a55bd4c66ab3886ea794a3fb6c0baab/apps/chat/bots.py#L94) by creating a runnable and invoking that, but only now we should be creating and running the runnable inside the `_process` method of the `LLMResponseWithPrompt`.
> 
> We can also add a [new state ](https://github.com/dimagi/open-chat-studio/blob/b88c5797dad31fb945d34a7661287b6e92331366/apps/service_providers/llm_service/state.py#L16) for the pipeline runnable. This way we can get things like the chat model, the prompt etc from the pipeline instead of the experiment.
> 
> I hope this makes sense? I'm not sure if this is the correct way to go, but I have a feeling that it is.
> 
> Thoughts? cc @snopoke

There is some scope for refactoring but I think it should be done separately.

SmittieC:
> > My thoughts:
> > In my mind, the`TopicBot` class is a hardcoded pipeline. It's running each "bot"/experiment (child bot, terminal bot, main bot) as a separate runnable. I like the new `PipelineBot` class, I think this makes that distinction clear. I am thinking though that the `LLMResponseWithPrompt` should also invoke the runnables like we do in the `TopicBot` class, but ofcourse not in the same "hardcoded" way.
> > What I am suggesting is that we do the same as [what we're doing here](https://github.com/dimagi/open-chat-studio/blob/ca65ca4b2a55bd4c66ab3886ea794a3fb6c0baab/apps/chat/bots.py#L94) by creating a runnable and invoking that, but only now we should be creating and running the runnable inside the `_process` method of the `LLMResponseWithPrompt`.
> > We can also add a [new state ](https://github.com/dimagi/open-chat-studio/blob/b88c5797dad31fb945d34a7661287b6e92331366/apps/service_providers/llm_service/state.py#L16) for the pipeline runnable. This way we can get things like the chat model, the prompt etc from the pipeline instead of the experiment.
> > I hope this makes sense? I'm not sure if this is the correct way to go, but I have a feeling that it is.
> > Thoughts? cc @snopoke
> 
> There is some scope for refactoring but I think it should be done separately.

Do you mean the refactoring should be done separately or the pipeline implementation should be done separately? (so we shouldn't have runnables inside the pipeline as I proposed)

snopoke:
> Do you mean the refactoring should be done separately or the pipeline implementation should be done separately? (so we shouldn't have runnables inside the pipeline as I proposed)

I meant the refactoring. I'm not sure exactly what approach makes sense yet. There are also some other changes to the runnables I'm interesting in looking at:

* move history management out of runnable (saving messages, history compression etc)
* prompt formatting

apps/pipelines/nodes/nodes.py
proteusvacuum: Yeah, I wasn't sure how best to do this - that state is pretty well coupled to how the TopicBot uses it, and all the inheritance in those state classes was becoming a bit confusing to me.  We can discuss quickly the best way of refactoring / implementing this tomorrow.

apps/experiments/views/experiment.py
```diff
@@ -239,15 +245,19 @@ def clean(self):
...
                 errors["llm_provider"] = "LLM Provider is required unless you select an OpenAI Assistant"
             if not cleaned_data.get("llm"):
                 errors["llm"] = "LLM is required unless you select an OpenAI Assistant"
-        else:
+        elif bot_type == "assistant":
             if not cleaned_data.get("assistant"):
                 errors["assistant"] = "Assistant is required when creating an assistant experiment"
+        elif bot_type == "pipeline":
+            if not cleaned_data.get("pipeline"):
+                errors["assistant"] = "Pipeline is required when creating a pipeline experiment"
```
proteusvacuum, line 260: boo

apps/experiments/views/experiment.py
```diff
@@ -239,15 +245,19 @@ def clean(self):
...
                 errors["prompt_text"] = "Prompt text is required unless you select an OpenAI Assistant"
             if not cleaned_data.get("llm_provider"):
                 errors["llm_provider"] = "LLM Provider is required unless you select an OpenAI Assistant"
             if not cleaned_data.get("llm"):
                 errors["llm"] = "LLM is required unless you select an OpenAI Assistant"
-        else:
+        elif bot_type == "assistant":
             if not cleaned_data.get("assistant"):
                 errors["assistant"] = "Assistant is required when creating an assistant experiment"
+        elif bot_type == "pipeline":
```
snopoke, line 258: Should this set `assistant` to None in `cleaned_data`? (also above set `pipeline` to None if it's an assistant)?

apps/experiments/models.py
```diff
@@ -415,6 +422,8 @@ def get_llm_service(self):
             return self.llm_provider.get_llm_service()
         elif self.assistant:
             return self.assistant.llm_provider.get_llm_service()
+        elif self.pipeline:
+            return self.pipeline.get_llm_service()
```
snopoke, line 427: is this changes still necessary?

apps/pipelines/models.py
```diff
@@ -87,26 +89,32 @@ def flow_data(self) -> dict:
     def node_ids(self):
         return self.node_set.values_list("flow_id", flat=True).all()
 
-    def invoke(self, input: PipelineState, session: ExperimentSession | None = None) -> PipelineState:
+    def get_llm_service(self):
```
snopoke, line 92: I think this can be removed

apps/pipelines/models.py
```diff
@@ -128,6 +156,12 @@ class Node(BaseModel):
     def __str__(self):
         return self.flow_id
 
+    def get_pipeline_node(self):
```
snopoke, line 159: can remove this if we also remove `get_llm_service`

apps/pipelines/nodes/nodes.py
```diff
@@ -51,32 +60,68 @@ def _process(self, state: PipelineState) -> PipelineState:
...
 
-class CreateReport(LLMResponse):
-    __human_name__ = "Create a report"
+class LLMResponseWithPrompt(LLMResponse):
+    __human_name__ = "LLM response with prompt"
 
+    source_material_id: SourceMaterialId | None = None
     prompt: str = (
         "Make a summary of the following text: {input}. "
         "Output it as JSON with a single key called 'summary' with the summary."
```
snopoke, line 84: Can we change this to:
```
You are a helpful assistant. Answer the user's query as best you can:
{input}
```

apps/pipelines/nodes/nodes.py
```diff
@@ -51,32 +60,68 @@ def _process(self, state: PipelineState) -> PipelineState:
...
+        chain = prompt | super().get_chat_model()
+        output = chain.invoke(context, config=self._config)
         return output.content
 
+    def _get_context(self, state: PipelineState, prompt: PromptTemplate):
+        session = state["experiment_session"]
+        context = {"input": state["messages"][-1]}
+
+        if "source_material" in prompt.input_variables and self.source_material_id is None:
+            raise PipelineNodeBuildError("No source material set, but the prompt expects it")
```
snopoke, line 99: can we validate this when saving the node as well?

apps/pipelines/tests/test_runnable_builder.py
```diff
@@ -160,6 +170,43 @@ def test_llm_response(get_llm_service, provider, pipeline):
...
+            {
+                "data": {
+                    "id": "llm-GUk0C",
+                    "label": "Get the robot to respond",
+                    "type": "LLMResponseWithPrompt",
+                    "params": {
+                        "llm_provider_id": provider.id,
+                        "llm_model": "fake-model",
+                        "source_material_id": source_material.id,
+                        "prompt": "Use this {source_material} to answer questions about {participant_data}."
```
snopoke, line 191: shouldn't this also include the `{input}` var?

apps/pipelines/nodes/nodes.py
proteusvacuum: We can, but we don't currently have a way of returning validations to the frontend. I can add that as a separate story.

proteusvacuum: https://github.com/dimagi/open-chat-studio/issues/627

apps/pipelines/tests/test_runnable_builder.py
proteusvacuum: I updated the test to have multiple nodes, which also includes that variable. https://github.com/dimagi/open-chat-studio/pull/608/commits/80b35e59a1e4ec4b8145151402bf2571f9a96d2a

snopoke (approved):

PR #611, opened 2024-08-15
tracing with langfuse

snopoke:
## Description
Adds support for using langfuse for tracing.

There is no UI for configuration yet so it has to be done via Django admin (adding a tracing provider).

## User Impact
User's can select a tracing provider when creating / editing an experiment.

## Docs
Will be done when the UI is done.


stephherbers (dismissed):

apps/service_providers/models.py
```diff
@@ -333,3 +333,34 @@ def type_enum(self):
...
+
+
+class TraceProviderType(models.TextChoices):
+    langfuse = "langfuse", _("Langfuse")
+
+    def get_service(self, config: dict) -> tracing.TraceService:
+        match self:
+            case TraceProviderType.langfuse:
+                return tracing.LangFuseTraceService(config)
+        raise Exception(f"No messaging service configured for {self}")
```
SmittieC, line 345:
```suggestion
        raise Exception(f"No tracing service configured for {self}")
```

SmittieC (approved):

SmittieC (approved):

PR #613, opened 2024-08-15
Cs/routing voice support

### LINKED ISSUE #450: Voice support for multi prompt
For one project they are interested in having different voice responses for the multi-prompt endpoints. Right now voice response is only configured by the router prompt (I believe). The project wanted to use different endpoints for different characters, so they should have a unique voice.

SmittieC:
resolves #450 

## Description

Added a `use_processor_bot_voice` field on the experiment. This way users can choose to always use the router bot's voice, or if the child bot generates the response, use their voice.

## User Impact

Users can choose whether or not they want the child bot that generated the response to also synthesize the voice message.

### Demo

image


apps/chat/bots.py
```diff
@@ -104,6 +106,7 @@ def _call_predict(self, input_str, save_input_to_history=True, attachments: list
         else:
             tag, chain = None, self.chain
 
+        self.processor_experiment = chain.experiment
```
snopoke, line 109: This could use a comment

apps/chat/channels.py
```diff
@@ -361,9 +363,14 @@ def _reply_voice_message(self, text: str):
         text, extracted_urls = strip_urls_and_emojis(text)
 
         voice_provider = self.experiment.voice_provider
+        synthetic_voice = self.experiment.synthetic_voice
+        if self.experiment.use_processor_bot_voice and self.bot.processor_experiment.voice_provider:
+            voice_provider = self.bot.processor_experiment.voice_provider
```
snopoke, line 368: could `bot` be None here?

apps/chat/channels.py
```diff
@@ -393,9 +400,8 @@ def _transcribe_audio(self, audio: BytesIO) -> str:
             if speech_service.supports_transcription:
                 return speech_service.transcribe_audio(audio)
 
-    def _get_experiment_response(self, message: str) -> str:
-        experiment_bot = TopicBot(self.experiment_session)
-        answer = experiment_bot.process_input(message, attachments=self.message.attachments)
+    def _get_bot_response(self, message: str) -> str:
+        answer = self.bot.process_input(message, attachments=self.message.attachments)
```
snopoke, line 404: bot might be None here if a new session was created right?

apps/service_providers/llm_service/runnables.py
```diff
@@ -452,3 +456,7 @@ def _cancel_run(self, assistant, thread_id, run_id):
             cancelling = run.status == "cancelling"
             if cancelling:
                 sleep(assistant.check_every_ms / 1000)
+
+    @property
+    def experiment(self) -> Experiment:
```
snopoke, line 461: where is this called?

apps/chat/channels.py
SmittieC: Bot cannot, but `bot.processor_experiment` can actually be since we call `send_message_to_user` from another place in the code where the bot hasn't generated a response, but rather we give it what it should say, so the TopicBot's `processor_experiment` will be `None`. I'll update the tests to cover this. Actually, there is a test, but the tests are not running against a multi bot setup, only a single bot setup, which is why everything is hunky dory.

apps/chat/channels.py
SmittieC: It wouldn't no, since we instantiate `self.bot` just after we called `self._ensure_sessions_exists()`. I think though it wouldn't hurt to just add a check here to make sure that the bot is always instantiated

apps/service_providers/llm_service/runnables.py
SmittieC: [Here](https://github.com/dimagi/open-chat-studio/pull/613/files#r1718584146)

SmittieC:
> General question, was this a user request? I'm a bit unsure about why you would want to have different voices for the same bot.

Yeah it's for [a project](https://github.com/dimagi/open-chat-studio/issues/450)

snopoke (approved):

PR #616, opened 2024-08-16
use a team global ExperimentChannel for API and WEB platforms

snopoke:
## Description
Instead of creating a new `ExperimentChannel` for each experiment, we reuse the 'global' (per team) channels for API and WEB platforms.

This requires a bit of refactoring:

1. Add team to `ExperimentChannel`
2. Make callers of `ChannelBase` and subclasses pass in the experiment always. This also reduces the amount of ambiguity in the code.

I did wonder about removing the need for these channel objects entirely but it is useful to have them attached to the session so that we know what platform the session is linked to.

In a followup PR I will create a migration to deal with existing data but to avoid issues during deploy, existing sessions will continue to use the channel from the session.

The impetus for this work was to resolve the creating of duplicate channels via the API: https://dimagi.sentry.io/share/issue/1ca122819bfc46e8a6fe0f6933121418/. I assume this is happening because of parallel requests to the API. Using a transaction should prevent duplicate objects being created but may not prevent initial errors when the channel does not exists.

## User Impact
There shouldn't be any user impact

### Docs
(one day we'll have tech docs)


apps/chat/channels.py
```diff
@@ -618,9 +618,6 @@ def send_voice_to_user(self, synthetic_voice: SynthesizedAudio):
 
 
 class SureAdhereChannel(ChannelBase):
-    def initialize(self):
-        self.messaging_service = self.experiment_channel.messaging_provider.get_messaging_service()
```
snopoke, line 622: There is already a `messaging_service` property with the same definition on the parent class

SmittieC:
@snopoke From what I can see this looks fine. If it's ok with you, I'd like to go over this once more on Monday morning first thing. My head is normally clearer early in the morning.

apps/channels/models.py
```diff
@@ -155,6 +162,7 @@ def check_usage_by_another_experiment(platform: ChannelPlatform, identifier: str
         filter_params = {f"extra_data__{platform.channel_identifier_key}": identifier}
         channel = ExperimentChannel.objects.filter(**filter_params).first()
         if channel and channel.experiment != new_experiment:
+            # TODO: check if it's in a different team and if the user has access to that team
```
SmittieC, line 165: Would we hide this link if the user doesn't have access, or what was your thoughts around this?

SmittieC (approved):

PR #618, opened 2024-08-19
Optionally append file content type to assistant instructions

### LINKED ISSUE #601: Detect filetype for uploads and let assistants know
Relevant threads on Slack:
https://dimagi.slack.com/archives/C05PK63Q89H/p1722930702167019
https://dimagi.slack.com/archives/C05PK63Q89H/p1723539681446419

I am proposing we implement two changes:

1. For all uploaded files we use the `magic` library (or similar) to confirm the filetype on the server side.
2. We append a user message with instructions stating the mimetype of the file.

For # 1, we would ideally send a message back to the user if this was an unsupported filetype and ask them to try again. If that's more complicated, I'd be happy to move that to a separate issue instead of blocking on it.

For # 2, here's a first pass on that instruction:
```txt
I have confirmed file is of type `{mimetype}`, so proceed to process it that way without any additional checks.
```

SmittieC:
resolves #601 

## Description

Added a checkbox so users can choose to append file content types to the assistant instruction. I noticed that PDFs are still tricky for assistant to read sometimes. I tried using the magic library as well, but doesn't really make a difference.

## User Impact

Assistants can be configured to append information about uploaded file content types. This should help nudge the assistant to use the correct tool to read files instead of guessing it a few times.

### Demo

N/A

### Docs

- Change: Assistants can be configured to append information about uploaded file content types. This should help nudge the assistant to use the correct tool to read files instead of guessing it a few times.


stephherbers (approved):
I'm in favor of the default for the checkbox to be True since it only adds useful information, but I'll defer to you

apps/channels/tests/test_base_channel_behavior.py
```diff
@@ -29,7 +29,7 @@
 def telegram_channel(db):
     experiment = ExperimentFactory(conversational_consent_enabled=False)
     channel = ExperimentChannelFactory(experiment=experiment)
-    channel = TelegramChannel(experiment, channel)
+    channel = TelegramChannel(experiment=experiment, experiment_channel=channel)
```
snopoke, line 32: just curious by why the change back to kwargs?

apps/channels/tests/test_base_channel_behavior.py
```diff
@@ -735,7 +735,7 @@ def test_send_message_to_user_with_multibot(
     ExperimentChannelFactory(experiment=router_exp)
     ExperimentRoute.objects.create(team=team, parent=router_exp, child=child_exp, keyword="keyword1", is_default=True)
 
-    channel = TelegramChannel(experiment_session=session)
+    channel = TelegramChannel.from_experiment_session(experiment_session=session)
```
snopoke, line 738: 👍🏻

apps/service_providers/llm_service/state.py
```diff
@@ -184,9 +184,25 @@ def get_assistant_instructions(self):
...
         )
+        return instructions
+
+    def get_file_type_info(self):
+        if not self.experiment.assistant.include_file_info:
+            return ""
+        attachments = self.session.chat.attachments.filter(tool_type__in=["code_interpreter"])
+        file_type_info = []
+        for att in attachments:
+            file_type_info.extend([{file.external_id: file.content_type} for file in att.files.all()])
```
snopoke, line 204: Is the assistant able to match the file ID to the file?

snopoke (approved):
+1 for Steph's suggestion.

apps/channels/tests/test_base_channel_behavior.py
SmittieC: It's just preference. I get somewhat "twitchy" whenever there's multiple args being passed to a function that relies on position.

SmittieC:
> I'm in favor of the default for the checkbox to be True since it only adds useful information, but I'll defer to you

Cool, I'll update

apps/service_providers/llm_service/state.py
SmittieC: From what I could tell, yes. Example:

### In this chat we append file types
image

### In this chat we don't append file types
Notice where it says "The presence of the word directory suggests that this file might be a Word document (.docx)". It only figures out it's a word doc later on

image

snopoke (approved):

PR #619, opened 2024-08-20
Combine the events and pipelines v2 FF

### LINKED ISSUE #591: events and pipeline v2 FF to use the same flag
To fully use piplines v2, one needs to also have the `events` FF enabled. It would make more sense to put both of these under the same FF then

SmittieC:
resolves #591 

## Description

Merged the events and pipelines v2 feature flags

## User Impact

Given the data migration, there should be no user impact


snopoke (approved):

proteusvacuum:
FYI, this will no longer be a requirement when https://github.com/dimagi/open-chat-studio/pull/608 is merged

SmittieC:
Ah ok no need to merge this then :+1:

PR #623, opened 2024-08-21
Revert "Revert "add db constraint to enforce uniqueness of global channels"" + migrate existing channels

snopoke:
Reverts dimagi/open-chat-studio#622 and adds a migration to consolidate existing web and api channels


SmittieC:
The revert is already deployed and this adds it back?

SmittieC (approved):
Looks fine to me. I suggest you hold off on merging this until 2 PM this afternoon. There's demos and stuff happening until then.

PR #625, opened 2024-08-21
Whitelisting Participants

### LINKED ISSUE #484: Ability to whitelist or blacklist participants for experiments
For the research studies, it's important that only participants who have been recruited and consented into the study have access to the chatbot (and the research team). We need the ability to specify a set of participants that should be allowed to interact with the chatbot.

For failures, we should either fail silently or allow an optional static reply. (I could imagine a future where we want to use an LLM to generate a reply, but let's wait until there's a clear use case before doing that.)

SmittieC:
resolves #484 

## Description

Added ability to whitelist participants

## User Impact

Bots with a whitelist configured are automatically considered

### Demo

image

### Docs

- New: Experiment admins can now whitelist participants by going to the experiment edit page -> Whitelist tab and adding the participant identifiers in the input box. Experiments with a whitelist configured are automatically considered private. All team members of this experiment are also automatically whitelisted.

## Things I would like feedback on specifically:
- The UI. Having the textbox like you can see in the screenshot is sub-optimal. I'm wondering if we shouldn't show a table rather and allow the user to add new entries.
- We currently send a fixed english message to the user if they are not allowed to chat to the bot. My guess is that this is fine?

cc @bderenzi


apps/experiments/models.py
```diff
@@ -422,6 +423,20 @@ def get_api_url(self):
...
+        Whether or not a bot should be limited to whitelisted participants is inferred by the presence of a whitelist
+        """
+        return len(self.participant_whitelist) > 0
+
+    def get_whitelisted_participant_identifiers(self):
+        whitelist = []
+        team_user_emails = self.team.members.all().values_list("email", flat=True)
+        whitelist.extend(list(team_user_emails))
+        whitelist.extend(self.participant_whitelist)
+        return whitelist
```
snopoke, line 438: Rather than doing a query every time you could first check the whitelist and then only query the DB if necessary. You also don't need to load all team members, just do an 'exists' query with filtering.

snopoke (commented):
A few thoughts:

### UI
The text box definitely doesn't see like a very user friendly option but might be sufficient for now. I can imagine users wanting and CSV import / export option from the main experiment view.

### Nits
* Rename 'whitelist' to 'allowlist'

apps/experiments/models.py
```diff
@@ -422,6 +423,20 @@ def get_api_url(self):
     def get_absolute_url(self):
         return reverse("experiments:single_experiment_home", args=[self.team.slug, self.id])
 
+    @property
+    def only_whitelisted_participants_allowed(self):
```
stephherbers, line 427: nit naming:`is_public` and switch logic since at first read it's more intuitive

apps/experiments/views/experiment.py
```diff
@@ -232,6 +235,12 @@ def __init__(self, request, *args, **kwargs):
         self.fields["synthetic_voice"].widget.template_name = "django/forms/widgets/select_dynamic.html"
         self.fields["llm"].widget.template_name = "django/forms/widgets/select_dynamic.html"
 
+    def clean_participant_whitelist(self):
+        cleaned_identifiers = []
+        for identifier in self.cleaned_data["participant_whitelist"]:
+            cleaned_identifiers.append(identifier.replace(" ", ""))
```
stephherbers, line 241: add a validation error if not in E164 format: (can use `re.compile(r'^\+[1-9]\d{1,14}$')`) (based on https://www.twilio.com/docs/glossary/what-e164)

stephherbers (changes requested):

apps/experiments/views/experiment.py
snopoke: Unfortunately we can't validate it like that since it depends on the experiment channels. But perhaps we could use the configured channels to do the validation?

apps/experiments/models.py
SmittieC: Good point thanks

apps/experiments/models.py
SmittieC: Good suggestion, I'll update

apps/experiments/views/experiment.py
SmittieC: Is the suggestion that we take the input list and run it though all configured experiment channel validators (that I need ot introduce). If it fails all of them, we know it's an invalid identifier. If it passes at least one of them, we know it is valid. Is this what ya'll is suggesting?

apps/experiments/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/625/commits/af8473ae0cea605473ed4794f1d5dc8848ed9754

apps/experiments/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/625/commits/af8473ae0cea605473ed4794f1d5dc8848ed9754

apps/experiments/views/experiment.py
snopoke: yes, but I'm not convinced it's worth it yet

templates/experiments/experiment_form.html
```diff
@@ -81,6 +81,18 @@ <h3 class="text-md font-bold">{% translate "Please note" %} <i class="fas fa-exc
         {% render_form_fields form "tools" %}
       </div>
 
+      <input type="radio" name="main_tabs" role="tab" class="tab" aria-label="Allowlist">
+      <div role="tabpanel" class="tab-content">
+        <h3 class="text-md font-bold">{% translate "Please note" %}</h3>
+        <p class="py-4 text-sm">
+          {% blocktranslate %}
+            To limit the set of particiapants that can chat with this bot, simply add their identifiers in this list.
+            When the list is empty, the bot will be public.
```
snopoke, line 90: I don't think users will know what being public means. Maybe reword to this:
```suggestion
            When the list is empty, the bot will accessible via its public link.
```

snopoke (commented):
LGTM. I think we should also add some indication on the main experiment page e.g. is public vs count of allowed participants.

templates/experiments/experiment_form.html
```diff
@@ -81,6 +81,18 @@ <h3 class="text-md font-bold">{% translate "Please note" %} <i class="fas fa-exc
...
+          {% endblocktranslate %}
+          {% render_form_fields form "participant_allowlist" %}
```
snopoke, line 92: Would it make sense to use the TomSelect widget here?

SmittieC: I was also thinking about that but I'd prefer a table of sorts to that list. But this list would be better than the current input element. It's not a big lift, so I can add this quickly and maybe down the line add better UI + support for CSV uploads

SmittieC: I'm torn between two: 1. do something quick and easy and 2. do it properly from the get-go

SmittieC: ah good point. Didn't think of that. I'll look into that as well

SmittieC: https://github.com/dimagi/open-chat-studio/pull/625/commits/b52e2e05d8e21b93587bc5f6176a9cad0d3db71d

Tom select supports pasting, but not copying, so I added a copy button.

image

snopoke (commented):
Looks good. Let me know when I should review again.

SmittieC:
> Looks good. Let me know when I should review again.

I don't plan to make any more changes, so I think it's ready

templates/experiments/experiment_form.html
```diff
@@ -81,6 +82,28 @@ <h3 class="text-md font-bold">{% translate "Please note" %} <i class="fas fa-exc
...
 
+      <input type="radio" name="main_tabs" role="tab" class="tab" aria-label="Allowlist">
+      <div role="tabpanel" class="tab-content">
+        <h3 class="text-md font-bold">{% translate "Please note" %}</h3>
+        <p class="py-4 text-sm">
+          {% blocktranslate %}
+            To limit the set of particiapants that can chat with this bot, simply add their identifiers in this list.
+            When the list is empty, the bot will accessible via its public link.
+          {% endblocktranslate %}
+          <div class="w-96 my-2" x-data="{allowlist: {{ form.initial.participant_allowlist }}}">
```
snopoke, line 93: Why does this need the alpine attrs?

apps/experiments/views/experiment.py
```diff
@@ -186,17 +186,17 @@ class Meta:
             "echo_transcript",
             "use_processor_bot_voice",
             "trace_provider",
-            "participant_whitelist",
+            "participant_allowlist",
         ]
-        labels = {"source_material": "Inline Source Material", "participant_whitelist": "Participant Whitelist"}
+        labels = {"source_material": "Inline Source Material", "participant_allowlist": "Participant Whitelist"}
```
bderenzi: I think this is the part where we want to change Whitelist to Allowlist yeah?

```suggestion
        labels = {"source_material": "Inline Source Material", "participant_allowlist": "Participant Allowlist"}
```

snopoke:
@SmittieC we noticed that this seems to have some overlap with the `is_active` field on experiments (https://github.com/dimagi/open-chat-studio/blob/main/apps/experiments/models.py#L318). I think that field should be removed or somehow incorporated / renamed and used in conjunction with this since it's essentially determining if the experiment is publicly accessible.

SmittieC:
> @SmittieC we noticed that this seems to have some overlap with the `is_active` field on experiments (https://github.com/dimagi/open-chat-studio/blob/main/apps/experiments/models.py#L318). I think that field should be removed or somehow incorporated / renamed and used in conjunction with this since it's essentially determining if the experiment is publicly accessible.


Sneaky field. Yes I agree, thanks for pointing it out. I'll address this

templates/experiments/experiment_form.html
SmittieC: It's mainly for the the copy feature. the copyToClipboard method wants to copy the contents of the div that we point it to, but if we point it to the select div (assuming we're not using the hidden input and populating its contents with alpineJS), then it copies the `<option>` tags as well. It basically copies the HTML. Having a hidden input there solves this issue, but also serves as a convenient way to store the data to be submitted. Hope this makes sense?

SmittieC:
> > @SmittieC we noticed that this seems to have some overlap with the `is_active` field on experiments (https://github.com/dimagi/open-chat-studio/blob/main/apps/experiments/models.py#L318). I think that field should be removed or somehow incorporated / renamed and used in conjunction with this since it's essentially determining if the experiment is publicly accessible.
> 
> Sneaky field. Yes I agree, thanks for pointing it out. I'll address this

It's not used and I think that it's not necessary to keep track in the DB if an experiment is public or not. Can easily be calculated and we can use a new field if new requirements surface

SmittieC:
> > > > @SmittieC we noticed that this seems to have some overlap with the `is_active` field on experiments (https://github.com/dimagi/open-chat-studio/blob/main/apps/experiments/models.py#L318). I think that field should be removed or somehow incorporated / renamed and used in conjunction with this since it's essentially determining if the experiment is publicly accessible.
> > > 
> > > 
> > > Sneaky field. Yes I agree, thanks for pointing it out. I'll address this
> > 
> > 
> > It's not used and I think that it's not necessary to keep track in the DB if an experiment is public or not. Can easily be calculated and we can use a new field if new requirements surface. I removed it [here](https://github.com/dimagi/open-chat-studio/pull/625/commits/81277ea8dcda56122e8614494976824c17c9bd4e)
> 
> What do you mean 'it's not being used'? There are 16 experiments in our prod instance which have `is_active=False`. It is being used to prevent access to their public chat link.

omw yes you are correct, my apologies :man_facepalming: I removed its usage in the template! Don't know what happened in my brain tbh. Ok nevermind I'll revert

SmittieC:
> > > > > @SmittieC we noticed that this seems to have some overlap with the `is_active` field on experiments (https://github.com/dimagi/open-chat-studio/blob/main/apps/experiments/models.py#L318). I think that field should be removed or somehow incorporated / renamed and used in conjunction with this since it's essentially determining if the experiment is publicly accessible.
> > > > 
> > > > 
> > > > Sneaky field. Yes I agree, thanks for pointing it out. I'll address this
> > > 
> > > 
> > > It's not used and I think that it's not necessary to keep track in the DB if an experiment is public or not. Can easily be calculated and we can use a new field if new requirements surface. I removed it [here](https://github.com/dimagi/open-chat-studio/pull/625/commits/81277ea8dcda56122e8614494976824c17c9bd4e)
> > 
> > 
> > What do you mean 'it's not being used'? There are 16 experiments in our prod instance which have `is_active=False`. It is being used to prevent access to their public chat link.
> 
> omw yes you are correct, my apologies 🤦‍♂️ I removed its usage in the template! Don't know what happened in my brain tbh. Ok nevermind I'll revert

I am wondering about the utility of this field. We don't have a way for users to change this field's value / to "archive" their experiments. Since people haven't asked for this feature again, I am wondering about its utility. Don't people just delete experiments instead?

We have two options as far as I can see.
1. Support this again (simply by including it in the experiment form)
2. Remove it. When we have the experiment versioning implemented, users will be able to publish experiments or not, essentially rendering this field redundant again.

I am leaning towards option 2

snopoke:
> I am wondering about the utility of this field. We don't have a way for users to change this field's value / to "archive" their experiments. Since people haven't asked for this feature again, I am wondering about its utility. Don't people just delete experiments instead?
> 
> We have three options as far as I can see.
> 
> 1. Support this again (simply by including it in the experiment form)
> 2. Remove it. When we have the experiment versioning implemented, users will be able to publish experiments or not, essentially rendering this field redundant again.
> 3. Merge this with the allowlist feature in the following way: the presence of identifiers in the allowlist does not dictate if the experiment is public or not. Only when this field is unchecked will the allowlist be enforced. I don't like this though
> 
> I am leaning towards option 2

I agree that it doesn't seem particularly useful but there is the case where you may want to disable public access without limiting access through other channels. If we only have the 'allowlist' then you have to specify participant IDs which limits availability on ALL channels.

However, I agree that it hasn't been asked for so I'm fine to remove it for now. We can add it back if necessary (with a better name).

SmittieC:
> > I am wondering about the utility of this field. We don't have a way for users to change this field's value / to "archive" their experiments. Since people haven't asked for this feature again, I am wondering about its utility. Don't people just delete experiments instead?
> > We have three options as far as I can see.
> > 
> > 1. Support this again (simply by including it in the experiment form)
> > 2. Remove it. When we have the experiment versioning implemented, users will be able to publish experiments or not, essentially rendering this field redundant again.
> > 3. Merge this with the allowlist feature in the following way: the presence of identifiers in the allowlist does not dictate if the experiment is public or not. Only when this field is unchecked will the allowlist be enforced. I don't like this though
> > 
> > I am leaning towards option 2
> 
> I agree that it doesn't seem particularly useful but there is the case where you may want to disable public access without limiting access through other channels. If we only have the 'allowlist' then you have to specify participant IDs which limits availability on ALL channels.
> 
> However, I agree that it hasn't been asked for so I'm fine to remove it for now. We can add it back if necessary (with a better name).

Sounds like the potential usecase is to limit the bot to external channels only (obviously without having an allowlist), since we currently force the bot to always be available on the web channel. Maybe the better approach would be to "link a web channel" in the same way we link other channels :thinking: 

Anycase, I'm keen to remove this, but do think I should just first inform people in the users channel about this situation.

snopoke (approved):

apps/experiments/views/experiment.py
SmittieC: @stephherbers are you fine with this not going in right now?

stephherbers (approved):

PR #632, opened 2024-08-23
allow builtin tools but not custom tools

snopoke:
## Description
Passing an empty list for tools disables all tools including builtin tools. We only want to disable custom tools in this instance since we don't want them to be called when we're generating automated responses eg. reminders.

## User Impact
Fix builtin tool calling for Assistants


SmittieC:
Thanks for this!

PR #642, opened 2024-08-26
Experiment Versioning UI: Add Versioning Tab

stephherbers:
## Description

This is just for the base version tab and not for the detail view. This change includes:
- adding a versions tab to Experiments
- Displaying the versions table and create version button to that tab

## User Impact

No user changes, everything is behind the pre-release FF

### Demo

N/A

### Docs

will be updated with future PRs


apps/experiments/views/experiment.py
```diff
@@ -1136,3 +1136,14 @@ def download_file(request, team_slug: str, session_id: int, pk: int):
         return FileResponse(file, as_attachment=True, filename=resource.file.name)
     except FileNotFoundError:
         raise Http404()
+
+
+@login_and_team_required
```
snopoke, line 1141: add `@require_POST`

snopoke: and `@transaction.atomic`

apps/experiments/tables.py
```diff
@@ -157,6 +157,21 @@ class Meta:
         empty_text = "No sessions yet!"
 
 
+class ExperimentVersionsTable(tables.Table):
+    version_number = columns.Column(verbose_name="Version Number", accessor="version_number")
```
snopoke, line 161: a date would also be good

snopoke (commented):
🚀

apps/experiments/views/experiment.py
```diff
@@ -1091,3 +1136,14 @@ def download_file(request, team_slug: str, session_id: int, pk: int):
...
+def set_default_experiment(request, team_slug: str, experiment_id: str, pk: int):
+    experiment = get_object_or_404(Experiment, pk=pk)
+
+    Experiment.objects.exclude(pk=experiment.pk).update(is_default=False)
```
SmittieC, line 1145: this should be `.update(is_default=True)`, no?

apps/experiments/tables.py
SmittieC: and show if it's the default one or not?

stephherbers:
draft until the experiment model changes complete

apps/experiments/urls.py
```diff
@@ -43,7 +43,7 @@
     path("table/", views.ExperimentTableView.as_view(), name="table"),
     path("e/<int:experiment_id>/", views.single_experiment_home, name="single_experiment_home"),
     path("e/<int:experiment_id>/sessions-table/", views.ExperimentSessionsTableView.as_view(), name="sessions-list"),
-    path("e/<int:pk>/set-default/", views.set_default_experiment, name="set_default_experiment"),
+    path("e/<int:pk>/", views.set_default_experiment, name="set-default-experiment"),
```
snopoke: just looking at the urls below there are a few that follow the patter `e/<int:pk>/{action}/` which makes me thing this should follow the same pattern. Also, a POST request to `/a/team/e/56/` would typically be an update request so this would break that pattern which makes it a possible source of confusion.

apps/experiments/urls.py
```diff
@@ -43,6 +43,8 @@
     path("table/", views.ExperimentTableView.as_view(), name="table"),
     path("e/<int:experiment_id>/", views.single_experiment_home, name="single_experiment_home"),
     path("e/<int:experiment_id>/sessions-table/", views.ExperimentSessionsTableView.as_view(), name="sessions-list"),
+    path("e/<int:experiment_id>/versions/", views.ExperimentVersionsTableView.as_view(), name="versions-list"),
+    path("e/<int:experiment_id>/create-version/", views.CreateExperimentVersion.as_view(), name="create_version"),
```
SmittieC, line 47: Nit:

```suggestion
    path("e/<int:experiment_id>/versions/create", views.CreateExperimentVersion.as_view(), name="create_version"),
```

apps/experiments/views/experiment.py
```diff
@@ -457,6 +469,43 @@ class DeleteFileFromExperiment(BaseDeleteFileView):
...
 
+# TODO: complete form
+class ExperimentVersionForm(forms.ModelForm):
+    class Meta:
+        model = Experiment
+        fields = ["description"]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["description"].widget.attrs.update({"class": "input input-bordered", "style": "width: 50%"})
```
SmittieC, line 480: We don't have a version description field yet. This will update the experiment's own description, which I'm guessing we don't want to do?

apps/experiments/views/experiment.py
```diff
@@ -457,6 +469,43 @@ class DeleteFileFromExperiment(BaseDeleteFileView):
...
+    model = Experiment
+    form_class = ExperimentVersionForm
+    template_name = "experiments/create_version_form.html"
+    title = "Create Experiment Version"
+    button_title = "Create"
+    permission_required = "experiments.add_experiment"
+
+    # TODO: add logic
+    def form_valid(self, form):
+        form.instance.team = self.request.team
```
SmittieC, line 493: Two things:
1. This view should also be a subclass of the `LoginRequiredMixin`
2. If you merge master into this branch (to bring the model changes) then we should be able to do the following:

```python
working_version = self.object
new_version = working_version.create_new_version()
# If we have a version specific description field, then we can pass in the description to the `create_new_version` method to create the version with that description
return HttpResponseRedirect(self.get_success_url())
```

SmittieC (commented):
The awesomeness. just two comments

apps/experiments/views/experiment.py
SmittieC: Merged https://github.com/dimagi/open-chat-studio/pull/642/commits/652155558a7c58a2e4e57fa786e55eb5cc2edfca

apps/experiments/views/experiment.py
stephherbers: hmm yes I was thinking that it would update the copied description filed that is then unique to that version-- but I'll remove for now and plan to add in later

stephherbers: https://github.com/dimagi/open-chat-studio/commit/307d0fe2afa335313a4ca7bebea09f8ca19fd67f (apologies for the force push just amended a file I forgot to add to the last commit no other changes

SmittieC (approved):
Looking good so far

PR #649, opened 2024-08-30
more robust encoding fallback

snopoke:
## Description
A more robust mechanism for selecting an encoding to use for token counting when the mapping is not present in langchain / tiktoken.

https://dimagi.sentry.io/share/issue/d71c6aceb9f14f2e8707a5cdaf821792/

### DIFF
--- a/apps/service_providers/llm_service/main.py
+++ b/apps/service_providers/llm_service/main.py
@@ -121,9 +121,19 @@ class OpenAILlmService(LlmService):
             openai_api_base=self.openai_api_base,
             openai_organization=self.openai_organization,
         )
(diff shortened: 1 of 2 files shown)


SmittieC (approved):
Will fine tuned model names always include its base model name? (So `ft:gpt-4o-mini-2024-07-18:abc::xyz` shows that `gpt-4o-mini-2024-07-18` was used, but will it always be the case?)

bderenzi:
With OpenAI's fine tuning, yes, their naming scheme appears to be:

```python
f"ft:{openai_model_name}:{organization_name}:{optional_custom_name}:{hash}"
```

When that `optional_custom_name` is omitted, we get the double `::`.

Issue #637, opened 2024-08-23
update session API to return full transcript


snopoke:
```json
{
  "role": "assistant|user|system",
  "content": "text",
  "tags": ["tag1", "tag2"],
  "attachments": [
    {"content-type": "", "name": "", "size": 123, "url": "..."}
  ]
}
```

snopoke:
What about summaries?

PR #650, opened 2024-08-30
session api messages

snopoke:
Resolves https://github.com/dimagi/open-chat-studio/issues/637

## Description
Update the `session_retrieve` API endpoint to include the session messages. The serialized message format is as follows:

```
{
  "created_at": "a date",
  "role": "assistant|user|system",
  "content": "text",
  "tags": ["tag1", "tag2"],
  "attachments": [
    {"content-type": "", "name": "", "size": 123, "url": "..."}
  ],
  "metadata": {}
}
```

### Docs
Docs included as part of the PR (API schema changes).

### DIFF
--- a/apps/chat/models.py
+++ b/apps/chat/models.py
@@ -39,13 +39,18 @@ class Chat(BaseTeamModel, TaggedModelMixin, UserCommentsMixin):
 
     def get_langchain_messages_until_summary(self) -> list[BaseMessage]:
         messages = []
-        for message in self.messages.order_by("-created_at").iterator(100):
+        for message in self.message_iterator():
             messages.append(message.to_langchain_dict())
-            if message.summary:
-                messages.append(message.summary_to_langchain_dict())
+            if message.is_summary:
                 break
 
         return messages_from_dict(list(reversed(messages)))
+
+    def message_iterator(self):
+        for message in self.messages.order_by("-created_at").iterator(100):
+            yield message
+            if message.summary:
+                yield message.get_summary_message()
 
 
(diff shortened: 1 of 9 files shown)


stephherbers:
looking at some of the existing test failures-- there are instances that the summary just repeats the human message, which is not what we would want, correct?

apps/chat/models.py
```diff
@@ -95,6 +100,22 @@ class ChatMessage(BaseModel, TaggedModelMixin, UserCommentsMixin):
...
+        return ChatMessage(message_type=ChatMessageType.SYSTEM, content=content, metadata={"summary": True})
+
+    def save(self, *args, **kwargs):
+        if self.is_summary:
+            raise ValueError("Cannot save a summary message")
+        super().save(*args, **kwargs)
+
+    def get_summary_message(self):
+        if not self.summary:
+            raise ValueError("Message does not have a summary")
```
SmittieC, line 131: Out of curiosity, what is the reason for raising a `ValueError` rather than just returning an empty string?

snopoke: Because I think it's better to make sure this isn't called when it shouldn't be than to to check for `None` when calling this or else expecting an `AttributeError`

snopoke:
> looking at some of the existing test failures-- there are instances that the summary just repeats the human message, which is not what we would want, correct?

This was a bug in a refactor I did towards the end. Fixed now: 34df8e29db9615f3c9496a1ce836199a2ed4dc5c

SmittieC (approved):
LGTM

PR #651, opened 2024-08-30
Preliminary Boolean Pipeline Nodes

proteusvacuum:
## Description

This adds the ability to have a conditional node in pipelines, with the goal to be able to build router bots and add other branching logic. 

There should be only one valid path through the pipeline (i.e. data must always flow down only a single path), otherwise the output is indeterminate - state is currently shared between every path in the graph, so things get confusing. I've left it as is for now, rather than validate, as validating would be very hard, and I'm hoping to fix that behaviour in coming PRs.

In separate PRs I'd like to:
- Add a "router" node with the ability to choose the number of outputs. You can fake this with multiple Boolean nodes, but it is a bit confusing. 
- Add langgraph "[subgraphs](https://langchain-ai.github.io/langgraph/how-tos/subgraph/)" so that having multiple pathways through the graph does what the user will think it should.
- Now that we have a dictionary of outputs of every node, it could be great to update how the logging works / how the pipeline runs are displayed. 

## User Impact

Users can add Boolean Nodes to Pipelines

### Demo

[Screencast from 2024-08-30 10-32-08.webm](https://github.com/user-attachments/assets/c11213e0-3216-435d-af9b-f7151a84aa7b)


snopoke (commented):
Is the plan to merge this and then build on it?

proteusvacuum:
@snopoke , that was my hope - so that the diff can be more incremental.

apps/pipelines/graph.py
```diff
@@ -57,15 +97,28 @@ def build_runnable(self) -> RunnableSequence:
         state_graph.set_finish_point(end[0])
 
         compiled_graph = state_graph.compile()
+        # compiled_graph.get_graph().print_ascii()
```
snopoke, line 100: nit: print

assets/javascript/apps/pipeline/PipelineNode.tsx
```diff
@@ -109,15 +111,59 @@ export function PipelineNode({ id, data, selected }: NodeProps<NodeData>) {
               inputParam={inputParam}
               value={params[inputParam.name]}
               onChange={updateParamValue}
-              provider={params.llm_provider_id}
+              provider={
+                Array.isArray(params.llm_provider_id)
+                  ? params.llm_provider_id.join("")
+                  : params.llm_provider_id
```
snopoke, line 117: when would this be an array?

assets/javascript/apps/pipeline/PipelineNode.tsx
```diff
@@ -109,15 +111,59 @@ export function PipelineNode({ id, data, selected }: NodeProps<NodeData>) {
...
+              }
             />
           </>
         );
+      case "NumOutputs":
+        return (
+          <>
+            <div className="m-1 font-medium text-center">Number of Outputs</div>
+            <input
```
snopoke, line 126:
```suggestion
            <input
              className="input input-bordered w-full"
```

snopoke (approved):
no blocking comments

PR #659, opened 2024-09-03
display schedules in participant data view

snopoke:
## Description
Add the list of schedules to the participant data view.

## User Impact
Shows the list of schedules in the participant data tab when viewing a participant.

### Demo
image

@SmittieC @stephherbers do you think this should be here or in a new tab. One issue is that they should not be editable so perhaps a new tab is better?


SmittieC:
I think a new tab with a pretty display will be nice yeah

SmittieC (dismissed):
Looks nice

SmittieC (approved):

PR #661, opened 2024-09-04
sync messages to OpenAI threads

### LINKED ISSUE #658: Sync OpenAI thread with OCS history in multi-bot setup
If there is a mix of note types in multi-prompt bots the OpenAI thread can get out of sync (missed messages).

When calling an assistant we need to make sure we add all the messages that have happened since the last thread call.

## Implementation thought
* add OpenAI message ID to the chat message metadata
* sync messages up to the last one with and openai message ID

snopoke:
Resolves #658

## Description
Make sure that the OpenAI thread has the same messages on OCS. This is important in a multi-bot setup where some but not all the bots are Assistants. When control passes to a non-assistant bot the OpenAI thread will not get updated so we need to add missing messages when control passes back to the assistant bot.

## User Impact
Improved performance when using Assistants in multi-bot configurations.

### Docs
NA

### DIFF
--- a/apps/chat/models.py
+++ b/apps/chat/models.py
@@ -47,10 +47,10 @@ class Chat(BaseTeamModel, TaggedModelMixin, UserCommentsMixin):
 
         return messages_from_dict(list(reversed(messages)))
 
-    def message_iterator(self):
+    def message_iterator(self, with_summaries=True):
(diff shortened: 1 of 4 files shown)


snopoke:
@bderenzi FYI

PR #662, opened 2024-09-05
Router Nodes

proteusvacuum:
## Description
Adds a RouterNode node type to pipelines which allows calling an LLM, then routing the original input through a different output.

### Demo

[Screencast from 2024-09-04 22-06-09.webm](https://github.com/user-attachments/assets/4624f30c-183b-4c19-b973-24b025134ee4)


apps/pipelines/nodes/nodes.py
```diff
@@ -150,12 +159,41 @@ def process_conditional(self, state: PipelineState) -> Literal["true", "false"]:
...
+    __human_name__ = "Router"
+    llm_provider_id: LlmProviderId
+    llm_model: LlmModel
+    prompt: str = "You are an extremely helpful router {input}"
+    num_outputs: NumOutputs = 2
+
+    def process_conditional(self, state: PipelineState):
+        prompt = PromptTemplate.from_template(template=self.prompt)
+        chain = prompt | self.get_chat_model()
+        result = chain.invoke(state["messages"][-1], config=self._config)
```
snopoke, line 180: could be useful to use langchains structured output feature: https://python.langchain.com/v0.1/docs/modules/model_io/chat/structured_output/

snopoke (commented):
Looking good!

SmittieC (commented):
Very nice

PR #664, opened 2024-09-05
properly call `get_or_create` to avoid errors

snopoke:
## Description
Make sure we call `get_or_create` correctly to avoid DB errors:

https://dimagi.sentry.io/share/issue/ccf2bfa2d5754e6fab5bdbe2fe956b58/

Fixes DIMAGI-BOTS-E8


SmittieC (approved):
Thanks

PR #670, opened 2024-09-06
handle edge case where existing summary is too large

snopoke:
Fixes [DIMAGI-BOTS-E9](https://dimagi.sentry.io/issues/5801611231/)

## Description
If the existing summary is too large we're unable to compute a new summary. In this case ignore the existing summary and compute the new summary from the remaining history.

## User Impact
This will result in some loss of information in the chat history but it won't error.

### DIFF
--- a/apps/chat/conversation.py
+++ b/apps/chat/conversation.py
@@ -201,7 +201,8 @@ def _get_new_summary(llm, pruned_memory, summary, max_token_limit):
 
     if not context["new_lines"]:
         log.error(SUMMARY_TOO_LARGE_ERROR_MESSAGE)
-        return summary
+        # If the summary is too large, discard it and compute a new summary from the pruned memory
+        return _get_new_summary(llm, next_batch, None, max_token_limit)
 
     chain = LLMChain(llm=llm, prompt=SUMMARY_PROMPT, name="compress_chat_history")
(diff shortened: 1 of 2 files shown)


apps/chat/conversation.py
```diff
@@ -193,10 +195,15 @@ def _get_new_summary(llm, pruned_memory, summary, max_token_limit):
     recursively call this function with the remaining memory."""
     tokens, context = _get_summary_tokens_with_context(llm, summary, pruned_memory)
     next_batch = []
-    while tokens > max_token_limit:
+    while pruned_memory and tokens > max_token_limit:
         next_batch.insert(0, pruned_memory.pop())
         tokens, context = _get_summary_tokens_with_context(llm, summary, pruned_memory)
 
+    if not context["new_lines"]:
```
SmittieC, line 202: Can we assume that `pruned_memory` will always be populated if we call this function? If not, then the possibility exists where if `pruned_memory` is empty, then the call on line 196 will return a context where `new_lines` will be empty and we will skip the while-loop, so the check on 202 will evaluate to True and the `summary-too-large` error will be logged. I'm just wondering if this will always be accurate?

SmittieC (approved):
Just one question, but I don't think it should be blocking

PR #671, opened 2024-09-06
Experiment Versioning | Code updates

SmittieC:
This PR is the code updates part of [the orignal PR](https://github.com/dimagi/open-chat-studio/pull/655). That PR is too big and confusing, so this is the new one that should be reviewed. I'm keeping it in draft until we have some UI changes to go along with it.

I will update this PR once the [UI changes](https://github.com/dimagi/open-chat-studio/pull/642) is in.

## Description

This is the code changes to use either the working or versioned experiments. This branch should be tested rigorously.

## User Impact

There should be no impact if there are no experiment versions.

### Demo

Stay tuned...

### Docs

Stay tuned...

### TODO
- More tests!
- Experiment deletion
- Related model deletion (should notify the user and deprecate versions that uses that)
- API to specify which version to use
- Web user to specify which version to use
- Events
- Triggers
- Scheduled Messages


apps/experiments/views/source_material.py
```diff
@@ -33,7 +33,7 @@ class SourceMaterialTableView(SingleTableView):
     template_name = "table/single_table.html"
 
     def get_queryset(self):
-        query_set = SourceMaterial.objects.filter(team=self.request.team)
+        query_set = SourceMaterial.objects.filter(team=self.request.team).exclude(working_version__isnull=False)
```
stephherbers, line 36: so this only show the source materials associated with a released version? I feel like maybe we should exclude material associated with an archived experiment, but why the working?

apps/experiments/views/safety.py
```diff
@@ -30,7 +30,7 @@ class SafetyLayerTableView(SingleTableView):
     template_name = "table/single_table.html"
 
     def get_queryset(self):
-        return SafetyLayer.objects.filter(team=self.request.team)
+        return SafetyLayer.objects.filter(team=self.request.team).exclude(working_version__isnull=False)
```
stephherbers, line 33: some comment here: maybe I'm missing the point of this logic rather than if the exp is archived

stephherbers:
good start-- noting too that when filtering for versions we should ensure that `is_archived` is not false for either the working or the version being queried-- maybe useful to modify the helper functions

BUT since this FFed and we actually can't archive experiments yet, we can hold off to a separate PR when we add in that functionality but I'll leave it up to you

apps/experiments/decorators.py
```diff
@@ -25,11 +25,14 @@ def decorator(view_func):
         def decorated_view(request, team_slug: str, experiment_id: str, session_id: str):
             request.experiment = get_object_or_404(Experiment, public_id=experiment_id, team=request.team)
             request.experiment_session = get_object_or_404(
-                ExperimentSession, experiment=request.experiment, external_id=session_id, team=request.team
+                ExperimentSession,
+                experiment_id=request.experiment.get_working_version_id(),
```
snopoke, line 29: fine to leave this call here but just confirming that this should already be the working version right?

apps/experiments/models.py
```diff
@@ -616,7 +685,13 @@ class ExperimentRoute(BaseTeamModel, VersionsMixin):
 
     @classmethod
     def eligible_children(cls, team: Team, parent: Experiment | None = None):
-        """Returns a list of experiments: that are not parents, and are not children of the current experiment"""
+        """
+        Returns a list of experiments that fit the following criteria:
+        - They are not the same as the parent
+        - they are not parents
+        - they are ot not children of the current experiment
```
snopoke, line 692:
```suggestion
        - they are not children of the current experiment
```

apps/experiments/models.py
```diff
@@ -626,6 +701,7 @@ def eligible_children(cls, team: Team, parent: Experiment | None = None):
                 .exclude(id__in=child_ids)
                 .exclude(id__in=parent_ids)
                 .exclude(id=parent.id)
+                .exclude(id__in=parent.versions.all())
```
snopoke, line 704: I think this would be more efficient as `working_version_id=parent.id`

apps/experiments/views/experiment.py
```diff
@@ -873,17 +885,18 @@ def start_session_public(request, team_slug: str, experiment_id: str):
 @permission_required("experiments.invite_participants", raise_exception=True)
 def experiment_invitations(request, team_slug: str, experiment_id: str):
     experiment = get_object_or_404(Experiment, id=experiment_id, team=request.team)
+    working_version = experiment.get_working_version()
```
snopoke, line 888: won't this always be the working version?

apps/experiments/views/experiment.py
```diff
@@ -943,30 +956,33 @@ def send_invitation(request, team_slug: str, experiment_id: str, session_id: str
...
+            redirect_url_name,
+            args=[team_slug, experiment_session.experiment_version.public_id, experiment_session.external_id],
         )
     )
     return set_session_access_cookie(response, experiment_session)
 
 
 @experiment_session_view(allowed_states=[SessionStatus.SETUP, SessionStatus.PENDING])
 def start_session_from_invite(request, team_slug: str, experiment_id: str, session_id: str):
+    # A session from invite will (for now?) always use the default experiment version
```
snopoke, line 977: I think it will always be the default. The only time you can use a non-default is if you're an experiment admin in which case you wouldn't use an invite.

apps/experiments/views/experiment.py
```diff
@@ -977,7 +993,6 @@ def start_session_from_invite(request, team_slug: str, experiment_id: str, sessi
     if request.method == "POST":
         form = ConsentForm(consent, request.POST, initial=initial)
         if form.is_valid():
-            WebChannel.check_and_process_seed_message(experiment_session)
```
snopoke, line 980: Is this handled in the channel?

snopoke (commented):
I can see it's going to be confusing to know when we're dealing with a version vs the default. I wonder if there's a clearer way to differentiate in code where we expect the working version vs a version vs either.

apps/experiments/decorators.py
SmittieC: Not necessarily in the future, but for now yes. If we have a button that allows the user to chat to a specific version, then this might be the versioned experiment

apps/experiments/models.py
SmittieC: Ah yes, good catch thanks

apps/experiments/views/experiment.py
SmittieC: Correct. When we start the session (this happens when we invite the user), the seed message is being generated, so doing it here essentially double generated the seed message

apps/experiments/views/experiment.py
SmittieC: Correct. I updated the views to always assume the `experiment_id` is the working experiment (https://github.com/dimagi/open-chat-studio/pull/671/commits/ebaac498dec2fe45ee72c80f7782421a3c55b2a1)

apps/experiments/views/experiment.py
```diff
@@ -963,10 +989,11 @@ def _record_consent_and_redirect(request, team_slug: str, experiment_session: Ex
 def start_session_from_invite(request, team_slug: str, experiment_id: str, session_id: str):
     experiment = get_object_or_404(Experiment, public_id=experiment_id, team=request.team)
     experiment_session = get_object_or_404(ExperimentSession, experiment=experiment, external_id=session_id)
-    consent = experiment.consent_form
+    default_version = experiment.default_version
+    consent = default_version.consent_form
 
     initial = {
-        "experiment_id": experiment.id,
+        "experiment_id": default_version.id,
```
snopoke, line 996: This doesn't seem to be used by the form. It's unclear to me if it's needed at all and if so whether it should be the working or default version

apps/slack/slack_listeners.py
```diff
@@ -61,10 +61,11 @@ def respond_to_message(event, context: BoltContext, session=None):
 
     slack_user = event.get("user")
 
+    experiment_version = experiment_channel.experiment.default_version
     if not session:
         external_id = make_session_external_id(channel_id, thread_ts)
         session = SlackChannel.start_new_session(
-            experiment_channel.experiment, experiment_channel, slack_user, session_external_id=external_id
+            experiment_version, experiment_channel, slack_user, session_external_id=external_id
```
snopoke, line 68: I think this should be the working version

apps/channels/views.py
```diff
@@ -96,9 +96,12 @@ def new_api_message(request, experiment_id: uuid):
     session = None
     if session_id := message_data.get("session"):
         try:
+            # TODO: Support ability to select a specific version
+            experiment = Experiment.objects.get(public_id=experiment_id)
+            working_version_id = experiment.get_working_version_id()
```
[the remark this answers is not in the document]
SmittieC: Correct. Yes forgot to update this piece. I am still to update the API endpoints to support multiple versions as well

apps/experiments/views/experiment.py
```diff
@@ -787,11 +803,14 @@ def get_message_response(request, team_slug: str, experiment_id: int, session_id
 
 
 def poll_messages(request, team_slug: str, experiment_id: int, session_id: int):
+    # experiment_id can be a version's ID
```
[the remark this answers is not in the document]
SmittieC: Nope, not true. Another one I didn't update. Thanks for your hawk eyes

apps/experiments/models.py
```diff
@@ -596,6 +691,21 @@ def is_public(self) -> bool:
     def is_participant_allowed(self, identifier: str):
         return identifier in self.participant_allowlist or self.team.members.filter(email=identifier).exists()
 
+    @transaction.atomic()
+    def delete(self, *args, **kwargs) -> tuple:
```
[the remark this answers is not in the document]
SmittieC: I think I'm rather going to take this out, since [this PR](https://github.com/dimagi/open-chat-studio/pull/676) is in the pipeline. Archiving a single version is supported by that PR, and I suppose it wouldn't make much sense to have an archive button for the working experiment. If we delete it, everything related goes, which is fine.

apps/experiments/models.py
```diff
@@ -596,6 +691,21 @@ def is_public(self) -> bool:
...
+        """Deletion strategy:
+        - If this experiment is a version, archive it
+        - If this experiment is the working version and has versions, archive all versions and this one
+        - If this experiment is the working version and does not have versions, delete it
+        """
+        self.experimentchannel_set.update(deleted=True, audit_action=AuditAction.AUDIT)
+        if self.is_working_version and not self.has_versions:
+            return super().delete(*args, **kwargs)
```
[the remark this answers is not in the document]
SmittieC: This is redundant now. See [this comment](https://github.com/dimagi/open-chat-studio/pull/671#discussion_r1754815245)

apps/experiments/models.py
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/671/commits/a46bb96e3050ca97ae118beaa030654f0d883a96

apps/experiments/views/experiment.py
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/671/commits/a46bb96e3050ca97ae118beaa030654f0d883a96

apps/experiments/views/experiment.py
SmittieC: You're right, it's not being used. Ripping it out...

apps/slack/slack_listeners.py
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/671#discussion_r1754850320

apps/experiments/views/experiment.py
```diff
@@ -691,10 +694,11 @@ def update_delete_channel(request, team_slug: str, experiment_id: int, channel_i
 @require_POST
 @login_and_team_required
 def start_authed_web_session(request, team_slug: str, experiment_id: int):
+    """Start an authed web session with the chosen experiment, be it a specific version or not"""
     experiment = get_object_or_404(Experiment, id=experiment_id, team=request.team)
 
     session = WebChannel.start_new_session(
-        experiment,
+        experiment_version=experiment.default_version,
```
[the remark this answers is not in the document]
snopoke: I agree with that except that it's not here where you make that call, it's where you pass the experiment to the task (which is a few lines below). The session should always be linked to the working version.

apps/slack/slack_listeners.py
snopoke: Same as reply to other comment

stephherbers:
still going through the review but wanted to add that the test plan sounds good and the scope of the PR

apps/experiments/views/experiment.py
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/671/commits/40f779b2f3819d6fa42a7426ab0177d2530e9611

SmittieC:
I built a bot on an unversioned experiment and went through basically all things affected by these changes. Then I made a version and did the same. No breakages (or at least those that did break were fixed). I think this is safe to merge as-is

apps/slack/slack_listeners.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/671/commits/40f779b2f3819d6fa42a7426ab0177d2530e9611

apps/experiments/models.py
```diff
@@ -1026,6 +1128,16 @@ def participant_data_from_experiment(self) -> dict:
         except ParticipantData.DoesNotExist:
             return {}
 
+    @cached_property
+    def experiment_version(self) -> Experiment:
+        """Returns the default experiment, or if there is none, the working experiment"""
+        return Experiment.objects.get_default_or_working(self.experiment)
```
snopoke, line 1134: nit:
```suggestion
        return self.experiment.default_version
```

snopoke (dismissed):
Nothing blocking

apps/chat/channels.py
```diff
@@ -121,6 +126,11 @@ def start_new_session(
         timezone: str | None = None,
         session_external_id: str | None = None,
     ):
+        if working_experiment.is_versioned:
+            raise VersionedExperimentSessionsNotAllowedException(
+                message="A session cannot be linked to an experiment version. "
+            )
```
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/671/commits/0a38ec682a07396da7c21727fd5e13b6d0f968ef

PR #673, opened 2024-09-09
Fix: Use the scheduled message's external id in its default name

SmittieC:
Fixes https://dimagi.sentry.io/issues/5814327413/?alert_rule_id=14063560&alert_type=issue&notification_uuid=ea9daed7-7fc3-4afb-9b90-0401964651c5&project=4505001320316928&referrer=issue_alert-slack.

## Description

The LLM seems to get confused with the id of the scheduled message and the external id that we append to the default name (e.g. `scheduled_message_213957b4-8e52-4371-8374-6efc03483133`). Since it uses the wrong ID, we cannot find the scheduled message. This PR does:
1. Catch the `ScheduledMessage.DoesNotExist` error and return a string that says that the message with the provided ID cannot be found.
2. Make it clearer in the string representation what the message ID is.

## User Impact

Less erros

### Demo

N/A

### Docs

I'll add a bugfix to the release notes


snopoke (approved):
One other thought is to let the LLM set the name of the schedule itself to avoid the auto-generated names, or to exclude the name if it matches the auto-generated pattern.

SmittieC:
> One other thought is to let the LLM set the name of the schedule itself to avoid the auto-generated names, or to exclude the name if it matches the auto-generated pattern.

Maybe not a bad idea to let the LLM generate a suitable name. I'll make an update

snopoke (approved):

PR #675, opened 2024-09-09
Experiment versioning UI: version details view

stephherbers:
## Description

Adds detail button view to the versions table to display the details of the experiment version and set that version as the defualt

## User Impact

None at this time, this feature is behind a pre-release FF for testing purposes

### Demo

Peek 2024-09-19 10-47


apps/experiments/urls.py
```diff
@@ -44,6 +44,7 @@
     path("e/<int:experiment_id>/", views.single_experiment_home, name="single_experiment_home"),
     path("e/<int:experiment_id>/sessions-table/", views.ExperimentSessionsTableView.as_view(), name="sessions-list"),
     path("e/<int:experiment_id>/versions/", views.ExperimentVersionsTableView.as_view(), name="versions-list"),
+    path("e/<int:pk>/versions/set_default", views.set_default_experiment, name="set-default-experiment"),
```
stephherbers, line 47: not working at the moment-- tried what seems like all possibilities

templates/experiments/components/experiment_version_details_button.html
```diff
@@ -0,0 +1,52 @@
+<button class="btn btn-sm btn-outline btn-primary"
+        onclick="document.getElementById('modal-{{ record.id }}').showModal()">
+    Details
+</button>
+
+<dialog id="modal-{{ record.id }}" class="modal">
```
snopoke, line 6: What about using HTMX to fetch the modal content rather than including it all in the main page (https://htmx.org/examples/modal-bootstrap/)

The benefit is a smaller page load for the main page which should be quicker to load and render. We also don't then need to prefetch all the related objects for each version on the main page.

apps/experiments/tables.py
```diff
@@ -167,6 +167,9 @@ class ExperimentVersionsTable(tables.Table):
         {% endif %}""",
         verbose_name="Default Version",
     )
+    details = columns.TemplateColumn(
```
SmittieC, line 170: Super nit : Thoughts on calling it something like "view" rather? I can't articulate exactly why, but if I were to click on a button that says "details" and see some actions where I can for instance archive the version of the experiment, my brain stop and asks "wait, what just happened". If this is not the case with you, fine to leave it as-is :smile:

apps/experiments/views/experiment.py
```diff
@@ -1158,3 +1159,23 @@ def download_file(request, team_slug: str, session_id: int, pk: int):
...
         raise Http404()
+
+
+@require_POST
+@transaction.atomic
+@login_and_team_required
+def set_default_experiment(request, team_slug: str, experiment_id: int, pk: int):
+    experiment = get_object_or_404(Experiment, working_version_id=experiment_id, version_number=pk, team=request.team)
+    Experiment.objects.exclude(version_number=pk).filter(team__slug=team_slug, working_version_id=experiment_id).update(
+        is_default_version=False, audit_action=AuditAction.AUDIT
```
stephherbers, line 1170: @snopoke got the auditing error without adding the "audit_action" but that feels bizarre for this instance to me-- is there a work around?

stephherbers:
@SmittieC to-do with this branch:
fix routing after setting as default to land on the version table (url error causing templating issue)
demo (duh!!)

apps/experiments/views/experiment.py
SmittieC: This [is expected](https://github.com/dimagi/django-field-audit?tab=readme-ov-file#auditing-special-queryset-writes). Since the .update method [doesn't call .save](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#django.db.models.query.QuerySet.update:~:text=Finally%2C%20realize%20that,%2C%20like%20this%3A), we need to manually mark this as being audited

.zed/settings.json
```diff
@@ -0,0 +1,14 @@
...
+{
+  "autosave": "on_focus_change",
+  "git_diff": true,
+  "hard_tabs": true,
+  "tabs": {
+    "close_position": "right",
+    "git_status": true
+  },
+  "format_on_save": "off"
+}
```
snopoke, line 14: nit: remote this file

apps/experiments/models.py
```diff
@@ -691,6 +691,42 @@ def is_public(self) -> bool:
     def is_participant_allowed(self, identifier: str):
         return identifier in self.participant_allowlist or self.team.members.filter(email=identifier).exists()
 
+    def version_details_for_display(self) -> list[dict]:
```
snopoke, line 694: nit: this is a good place to use a TypedDict or a dataclass

templates/experiments/components/experiment_version_details_button.html
```diff
@@ -0,0 +1,19 @@
+<button class="btn btn-sm btn-outline btn-primary"
+        hx-get="{% url 'experiments:experiment-version-details' team_slug=team.slug experiment_id=record.working_version_id version_number=record.version_number %}"
+        hx-target="#details_modal_content"
+        hx-trigger="click"
+        onclick="document.querySelector('#details_modal').showModal();">
+  View Details
+</button>
+
+<dialog id="details_modal" class="modal">
```
snopoke, line 9: can this be removed form here and we use the one in `experiment_version_table.html`?

snopoke (approved):
No blocking comments.

apps/experiments/models.py
SmittieC: Good idea: https://github.com/dimagi/open-chat-studio/pull/675/commits/a585408e1d250ad84136c5dc35e88629a76c6166

templates/experiments/components/experiment_version_details_button.html
SmittieC: https://github.com/dimagi/open-chat-studio/pull/675/commits/11abeb5d3c12d9c1b925e5eb4c0943ce31da1d2c

snopoke (approved):

PR #676, opened 2024-09-09
Experiment Versioning UI: archive button

stephherbers:
## Description

Adds an archive button and functionality to the details modal on an experiment version

## User Impact

None currently-- This is behind a pre-release FF

### Demo

https://github.com/user-attachments/assets/1d23189f-26d8-46c8-a83b-11645abbb443


apps/experiments/views/experiment.py
```diff
@@ -1170,3 +1170,15 @@ def set_default_experiment(request, team_slug: str, pk: int):
...
 
     return redirect("experiments:versions-list", team_slug=request.team.slug)
+
+
+@require_POST
+@transaction.atomic
+@login_and_team_required
+def archive_experiment(request, team_slug: str, pk: int):
+    experiment = get_object_or_404(Experiment, id=pk)
+    experiment.is_archived = True
```
snopoke, line 1180: what happens if it's also the default?

stephherbers: modify such that the user much change the default prior to archiving. I thought about making a different default for them, but feels risky not letting the user decide so opted for this approach

snopoke (approved):

PR #677, opened 2024-09-10
Participant Name

snopoke:
(feature request from users)

## Description
Add a 'name' field to the Participant model which can be displayed in the UI and used in the participant data.

This is a global field but when participant data is supplied to the bot, the experiment specific name will overwrite the global name if one is set.

## User Impact
Participants can have names besides the platform specific identifier.

### Demo
<div>
    <a href="https://www.loom.com/share/b0ffcbc0f63b402c8b70ec0ab9a34066">
      <p>Dimagi Chatbots | Experiments with AI, GPT and LLMs - 10 September 2024 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/b0ffcbc0f63b402c8b70ec0ab9a34066">
      
    </a>
  </div>

### Docs
API docs updates via schema.

There are no participant docs yet and I'm not going to create them as part of this change.


apps/participants/views.py
```diff
@@ -108,3 +114,20 @@ def get_context_data(self, *args, **kwargs):
         context["participant_data"] = json.dumps(data, indent=4)
         context["participant_schedules"] = participant.get_schedules_for_experiment(experiment, as_dict=True)
         return context
+
+
+@login_and_team_required
```
SmittieC, line 119: Is there a specific reason why you didn't go for a class based view here? That way these two views could be targeted by a single URI and be packaged together.

SmittieC (dismissed):

apps/participants/views.py
snopoke: I generally prefer function based views as they are simpler to with less hidden functionality. But I get your point about combining them.

SmittieC: Up to you

snopoke: [c7cb2d2](https://github.com/dimagi/open-chat-studio/pull/677/commits/c7cb2d25520d03230ea21b86187c61de88a2aff9)

SmittieC (approved):

PR #681, opened 2024-09-11
remove old task

snopoke:
This was used by the previous reminder tool.

I noticed in Task Badger that this was still being called. Upon further investigation, the periodic tasks are expired however they are still scheduled by the scheduler and then immediately revoked. This resulted in stale tasks in TB. I have removed all the expired periodic tasks so this should not get called anymore.


stephherbers (approved):
☠️

PR #683, opened 2024-09-12
Tag UI changes

snopoke:
I got a bit carried away implementing tag search in the session details view but I think the end result is good.

## Description
Updates to how we display and edit tags to improve the way they look and also so that they take up less space.

This also makes it consistent in all places where tags are shown.

You can also now search messages by tag in the session details view.

## User Impact
Improved UI around tags.

### Demo
<div>
    <a href="https://www.loom.com/share/76e8433256534261805ce9a5bc0548b7">
      <p>Dimagi Chatbots | Experiments with AI, GPT and LLMs - 12 September 2024 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/76e8433256534261805ce9a5bc0548b7">
      
    </a>
  </div>

### Docs
Hopefully this isn't something that needs much docs as it should be self explanatory.


SmittieC (approved):
Looks way better, thanks

PR #685, opened 2024-09-13
Various fixes for scheduled messages

snopoke:
Fixes [DIMAGI-BOTS-EK](https://dimagi.sentry.io/issues/5841775442/)

## Description
This fixes a few issues with scheduled messages:

* Handling 'null' repetitions parameter
* Improved formatting for string output
* Filter out complete schedules from participant data sent to bots

## User Impact
None

### Demo
See tests.

### Docs
None

### DIFF
--- a/apps/events/models.py
+++ b/apps/events/models.py
@@ -364,23 +364,23 @@ class ScheduledMessage(BaseTeamModel):
     def params(self):
         return self.custom_schedule_params or (self.action.params if self.action else {})
 
-    @cached_property
+    @property
     def name(self) -> str:
         return self.params["name"]
 
-    @cached_property
+    @property
     def frequency(self) -> int:
         return self.params["frequency"]
 
-    @cached_property
+    @property
     def time_period(self) -> str:
         return self.params["time_period"]
 
-    @cached_property
+    @property
     def repetitions(self) -> int:
         return self.params["repetitions"] or 0
 
-    @cached_property
+    @property
     def prompt_text(self) -> str:
         return self.params["prompt_text"]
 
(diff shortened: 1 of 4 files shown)


apps/events/models.py
```diff
@@ -388,19 +391,28 @@ def was_created_by_system(self) -> bool:
 
     def as_string(self, as_timezone: str | None = None):
         header_str = f"{self.name} (Message id={self.external_id}, message={self.prompt_text})"
-        if self.repetitions == 0:
+        if not self.repetitions or self.repetitions <= 1:
```
SmittieC, line 394: Will repetitions ever be negative? I think we can just do the `if not self.repetitions` check?

apps/events/models.py
```diff
@@ -347,17 +346,21 @@ def _trigger(self):
...
+        if self._should_mark_complete():
             self.is_complete = True
         else:
             delta = relativedelta(**{self.params["time_period"]: self.params["frequency"]})
             self.next_trigger_date = utc_now + delta
 
         self.save()
 
+    def _should_mark_complete(self):
+        repetitions = self.params.get("repetitions", None) or 0
```
SmittieC, line 358: Wondering if we shouldn't update the `repetitions` property to be something like

```python
    @cached_property
    def repetitions(self) -> int:
        return self.params["repetitions"] or 0
```

apps/events/models.py
snopoke: I don't think it will be negative. The issue with just checking 'not repetitions' is that a repetition of `1` is still a once off message since it only fires once.

apps/events/models.py
snopoke: We could do that. I'll update.

apps/events/models.py
SmittieC: Right yes :+1:

SmittieC (approved):
Sweet

PR #688, opened 2024-09-18
work around openai limits for creating threads

snopoke:
[DIMAGI-BOTS-F1](https://dimagi.sentry.io/issues/5864411225/)

See also https://community.openai.com/t/assistant-threads-create-400-messages-array-too-long/754574/4

### DIFF
--- a/apps/service_providers/llm_service/runnables.py
+++ b/apps/service_providers/llm_service/runnables.py
@@ -305,20 +305,19 @@ class AssistantExperimentRunnable(RunnableSerializable[dict, ChainOutput]):
         This is necessary in multi-bot setups if some of the bots are assistants but not all.
         """
 
-        def __sync_messages_to_thread(thread_id, messages):
+        def _sync_messages_to_existing_thread(thread_id, messages):
             for message in messages_to_sync:
                 self.state.raw_client.beta.threads.messages.create(current_thread_id, **message)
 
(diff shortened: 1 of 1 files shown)


SmittieC (dismissed):

apps/service_providers/llm_service/runnables.py
```diff
@@ -304,13 +304,21 @@ def _sync_messages_to_thread(self, current_thread_id):
...
-                for message in messages_to_sync:
-                    self.state.raw_client.beta.threads.messages.create(current_thread_id, **message)
+                __sync_messages_to_thread(current_thread_id, messages_to_sync)
             else:
-                thread = self.state.raw_client.beta.threads.create(messages=messages_to_sync)
+                # There is a 32 message limit when creating a new thread
+                first, rest = messages_to_sync[:32], messages_to_sync[32:]
+                thread = self.state.raw_client.beta.threads.create(messages=first)
                 current_thread_id = thread.id
+                if rest:
```
SmittieC, line 320: Super nitty comment: You don't have to check if there is a `rest` since it will not iterate over anything in the `__sync_messages_to_thread` method.

Question: Why the double underscore and not a single one?

snopoke: > Question: Why the double underscore and not a single one?

The outer method is called the same thing (with a single underscore) and I was too lazy to think of another name.

SmittieC (approved):

PR #692, opened 2024-09-20
Versioning | Basic diff view

SmittieC:
## Description

1. Added a basic diff view that the user will see when they want to create a new version. If there are no diffs, then they cannot create a new version.
2. When creating new versions, we'll check if there is a difference between the related field (like source material) and the latest version's related field. If not, we attach that version to the new experiment version instead of creating a new version for the related field. Suggestions welcome on better ways to check diffs

## User Impact

User's can now see changes between the current and previous version

### Demo

Peek 2024-09-20 15-27

### Docs

Will add once this is merged.

## Release Notes

- **New Features**
  - Enhanced admin interface to display all instances of Source Materials and Consent Forms.
  - Improved versioning capabilities for experiments, including a new version comparison view and detailed field change displays.
  - Introduced dynamic rendering in templates to reflect changes in experiment versions, including conditional content based on version states.
  - Added collapsible sections for displaying field changes in version comparisons.

- **Bug Fixes**
  - Corrected display logic in the experiment version details template.

- **Tests**
  - Added comprehensive unit tests for versioning functionality and model comparisons.

- **Chores**
  - General code refactoring for improved maintainability and readability.


SmittieC:
I did [a refactor](https://github.com/dimagi/open-chat-studio/pull/692/commits/8e4969fed96317571340ce821b51925b0e6b5eea) in preparation for field grouping (so we can display related fields nicely) and general better handling of the code. I'm still to
- [ ] write tests
- [ ] Use a collapsable to show field groups
- [ ] Cleanup

stephherbers:
> Use a collapsable to show field groups

this sounds like a good breakpoint for a second PR

apps/experiments/models.py
```diff
@@ -669,9 +677,27 @@ def create_new_version(self, version_description: str | None = None, make_defaul
...
-        if instance := getattr(self, attr_name):
-            setattr(new_version, attr_name, instance.create_new_version())
+        attr_instance = getattr(self, attr_name)
+        if not attr_instance:
+            return
+
+        latest_attr_version = attr_instance.latest_version
+
+        if latest_attr_version and not differs(
+            attr_instance, latest_attr_version, exclude_model_fields=self.DEFAULT_EXCLUDED_KEYS
```
snopoke, line 696: This should use the exclude keys from the related class (in case it is different from this class)

apps/experiments/tests/test_models.py
```diff
@@ -724,24 +766,9 @@ def test_archived_experiments_are_filtered_out(self):
...
-        if field_name not in model_fields:
-            continue
-
-        if field_name in excluded_keys:
-            continue
-        if field_value != new_dict[field_name]:
-            changed_fields.add(field_name)
-
-    field_difference = changed_fields.difference(set(expected_changed_fields))
+    field_difference = compare_models(original, new, VersionsMixin.DEFAULT_EXCLUDED_KEYS).difference(
```
snopoke, line 769:
```suggestion
    field_difference = compare_models(original, new, original.DEFAULT_EXCLUDED_KEYS).difference(
```

templates/experiments/components/versions/compare.html
```diff
@@ -0,0 +1,16 @@
+<div class="my-2">
+    {% if version_details.fields_changed %}
+        {% for field in version_details.fields %}
+            <div class="grid grid-cols-2 gap-2 max-h-64 overflow-y-auto">
+                {% include 'experiments/components/versions/version_field_details.html' with field=field.previous_field_version field_changed=field.changed diff_color='bg-red-400' %}
+                {% include 'experiments/components/versions/version_field_details.html' with field=field field_changed=field.changed diff_color='bg-emerald-400' %}
```
snopoke, line 6: Do these colors work in light and dark mode? You might need to add `dark:xxx` variants

templates/experiments/components/versions/version_field_details.html
```diff
@@ -0,0 +1,8 @@
+{% if field_changed or field.display_value %}
+    <div class="my-1">
+        <div class="grid grid-cols-3">
+            <dt class="text-sm font-medium leading-6">{{ field.label }}</dt>
+            <dd class="text-sm col-span-2 {% if field_changed|default:False %}rounded text-slate-950 {{ diff_color }}{% endif %}">{{ field.display_value }}</dd>
```
snopoke, line 5: Same here regarding the text color in dark mode

templates/experiments/components/versions/compare.html
SmittieC: I think it works well in light and dark mode Thougths?:

image

image

SmittieC:
> > Use a collapsable to show field groups
> 
> this sounds like a good breakpoint for a second PR

I sneaky sneaked the groups changes in as well :eyes: .

apps/experiments/models.py
```diff
@@ -744,55 +744,81 @@ def format_tools(tools: set):
         return Version(
             instance=self,
             fields=[
-                VersionField(group_name="Group 1", name="description", raw_value=self.description),
-                VersionField(group_name="Group 1", name="prompt_text", raw_value=self.prompt_text),
-                VersionField(group_name="Group 1", name="llm_model", raw_value=self.llm),
-                VersionField(group_name="Group 1", name="llm_provider", raw_value=self.llm_provider),
+                VersionField(group_name="General", name="description", raw_value=self.description),
```
SmittieC, line 747: This is what the user will see now:

image

image

apps/experiments/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/692/commits/d6843fb0869f40ae7603b0f94091ca98ac457994

apps/experiments/versioning.py
```diff
@@ -78,8 +89,16 @@ def get_field(self, field_name: str) -> VersionField:
 
     @property
     def fields_grouped(self):
-        # TODO: Return fields with group info for display purposes
-        pass
+        groups = defaultdict(dict)
```
snopoke, line 92: It doesn't seem like this needs to be a default dict. Also they dict values are not `dict` so it also doesn't seem correct.

apps/experiments/versioning.py
```diff
@@ -78,8 +89,16 @@ def get_field(self, field_name: str) -> VersionField:
...
+        for field in self.fields:
+            group_name = field.group_name
+            group_info = groups.get(group_name, FieldGroup(name=group_name))
```
snopoke, line 95: you could use `setdefault` so that you don't have to insert it into the dict at the end:

```suggestion
            group_info = groups.setdefault(group_name, FieldGroup(name=group_name))
```

apps/experiments/versioning.py
```diff
@@ -78,8 +89,16 @@ def get_field(self, field_name: str) -> VersionField:
...
+            group_info.show = group_info.show or bool(field.raw_value) or bool(field.changed)
+            group_info.has_changed_fields = group_info.has_changed_fields or field.changed
+            # We need to always include the field
+            group_info.fields.append(field)
+            groups[group_name] = group_info
+        return groups.items()
```
snopoke, line 100: or just `list(groups.values())` since the name is already in the group object

Note: wrapping with list to avoid returning a dict view object.

templates/experiments/components/versions/compare.html
```diff
@@ -1,16 +1,42 @@
...
-                {% include 'experiments/components/versions/version_field_details.html' with field=field field_changed=field.changed diff_color='bg-emerald-400' %}
-            </div>
+        {% for group_name, group in version_details.fields_grouped %}
+            {% if group.show %}
+                <div x-data="{ isOpen: {{ group.has_changed_fields|yesno:'true,false' }} }" class="mb-2">
+                    <div :class="isOpen ? 'collapse collapse-open collapse-plus bg-base-200' : 'collapse collapse-close collapse-plus bg-base-200'">
+                        <div class="collapse-title text-xl font-medium cursor-pointer hover:text-white" @click="isOpen = !isOpen">
+                            {{ group_name }}
+                        </div>
+                        <div class="collapse-content" x-show="isOpen">
```
snopoke, line 10: Is the `x-show` necessary? I would have thought the `collapse` classes would handle this.

templates/experiments/components/versions/compare.html
```diff
@@ -1,16 +1,42 @@
...
     {% if version_details.fields_changed %}
-        {% for field in version_details.fields %}
-            <div class="grid grid-cols-2 gap-2 max-h-64 overflow-y-auto">
-                {% include 'experiments/components/versions/version_field_details.html' with field=field.previous_field_version field_changed=field.changed diff_color='bg-red-400' %}
-                {% include 'experiments/components/versions/version_field_details.html' with field=field field_changed=field.changed diff_color='bg-emerald-400' %}
-            </div>
+        {% for group_name, group in version_details.fields_grouped %}
+            {% if group.show %}
+                <div x-data="{ isOpen: {{ group.has_changed_fields|yesno:'true,false' }} }" class="mb-2">
+                    <div :class="isOpen ? 'collapse collapse-open collapse-plus bg-base-200' : 'collapse collapse-close collapse-plus bg-base-200'">
```
snopoke, line 6: nit: avoid the repetition of the class names
```suggestion
                    <div class="collapse collapse-plus bg-base-200" :class="isOpen ? 'collapse-open' : 'collapse-close'">
```

Alpine won't overwrite the other classes: https://alpinejs.dev/directives/bind#special-behavior

templates/experiments/components/versions/compare.html
```diff
@@ -1,16 +1,42 @@
 <div class="my-2">
     {% if version_details.fields_changed %}
-        {% for field in version_details.fields %}
-            <div class="grid grid-cols-2 gap-2 max-h-64 overflow-y-auto">
-                {% include 'experiments/components/versions/version_field_details.html' with field=field.previous_field_version field_changed=field.changed diff_color='bg-red-400' %}
-                {% include 'experiments/components/versions/version_field_details.html' with field=field field_changed=field.changed diff_color='bg-emerald-400' %}
-            </div>
+        {% for group_name, group in version_details.fields_grouped %}
+            {% if group.show %}
+                <div x-data="{ isOpen: {{ group.has_changed_fields|yesno:'true,false' }} }" class="mb-2">
```
snopoke, line 5: Looking at this some more I don't think alpine is necessary here at all. I think the same can be achieved using the 'checkbox' collapse and setting the checkbox to 'checked' if you want the contents to be initially visible: https://daisyui.com/components/collapse/#collapse-with-checkbox

templates/experiments/components/versions/compare.html
bderenzi: Can I propose that we use a slightly more readable dark red text on light red background and dark green on light green? Similar to the badges here:

https://flowbite.com/docs/components/badge/

I'm guessing it'll still work on both, but I think will be a little easier on the eyes

SmittieC: Ooh thanks, this is a little easier on the eyes

## Dark mode
image

## Light mode
image

apps/experiments/versioning.py
SmittieC: Thanks. I was first going to use one, but later pivoted and didn't update this :+1:

templates/experiments/components/versions/compare.html
SmittieC: I remember previously I did try this, but my previous approach to achieving default open/close behaviour was wrong, which is why I resorted to using AlpineJS. Using the `checked` property solves this. Thanks for the suggestion

snopoke: Nice, the updated version is much better!

snopoke:
@SmittieC let me know when you want me to review this again

SmittieC:
> @SmittieC let me know when you want me to review this again

I think this is ready to be reviewed again

snopoke (approved):

PR #693, opened 2024-09-20
Pipeline Chat History

proteusvacuum:
## Description
This adds three different Chat History types to the pipelines:
- Node
- Named
- Global

`Node` history stores and includes the history for only that particular node. A `Named` history can be shared between nodes. `Global` history is the history for the full pipeline run (i.e. the input and output of the pipeline).

`Global` history uses the current `ChatHistory` model, and similarly will compress this history, using whatever LLM is associated with the node that is requesting that history to be inserted. 
Note, if you have multiple LLMs with different allowed token lengths, I think you will need to set all of them to be the minimum of all the token lengths, otherwise you get errors. 

For now, `node` and `named` history do not compress their history.

In follow up PRs I will work on ways of storing multiple compressed summaries so that the same history can be shared between nodes with different context lengths. At that time, I will also implement compression of the named and node history types.

## User Impact

Allows actually chatting with a pipeline, rather than just sending a single message.

### Demo

[Screencast from 2024-09-20 15-31-19.webm](https://github.com/user-attachments/assets/06463889-ecff-4b7b-9518-53acd0aad37d)


bderenzi:
Very cool. I think we also want no history, or perhaps it should be implemented as a max number of messages to keep (which could be zero). I think in the translation case you likely don't want much, if any, history as it's just x-shot prompting that unnecessarily gets expensive. I think you also wouldn't want to spend tokens on compression / summerization activities either in that case.

SmittieC (commented):
There's some conflicts, but overall it looks good to me

apps/pipelines/nodes/nodes.py
```diff
@@ -120,6 +129,33 @@ def _get_context(self, input, state: PipelineState, prompt: PromptTemplate):
...
+    def _get_history(self, session: ExperimentSession, node_id: str):
+        if self.history_type == PipelineChatHistoryTypes.NAMED:
+            history_name = self.history_name
+        elif self.history_type == PipelineChatHistoryTypes.GLOBAL:
+            history_name = "global"
+        else:
+            history_name = node_id
+
+        try:
+            history: PipelineChatHistory = session.pipeline_chat_history.get(type=self.history_type, name=history_name)
```
snopoke, line 150: could this be pushed into the state class so that it can be cached?

apps/pipelines/nodes/nodes.py
```diff
@@ -127,9 +130,14 @@ def _get_context(self, input, state: PipelineState, prompt: ChatPromptTemplate,
...
 
-    def _get_history(self, session: ExperimentSession, node_id: str):
+    def _get_history(self, session: ExperimentSession, node_id: str, input):
         if self.history_type == PipelineChatHistoryTypes.GLOBAL:
-            return session.chat.get_langchain_messages_until_summary()
+            return compress_chat_history(
+                chat=session.chat,
+                llm=self.get_chat_model(),
+                max_token_limit=self.max_token_limit,
+                input_messages=[HumanMessage(content=input)],
```
snopoke, line 141: This should include the prompt

snopoke (commented):
Possibly for a future PR but for named histories (node & named) we'll need to limit their size either by compression or filtering.

apps/pipelines/nodes/nodes.py
proteusvacuum: Do you mean into the `PipelineState` or some other state class? 

Yes, we can totally do that - however, it will only benefit when you have more two or more nodes with the same history name in the pipeline (since the pipeline is invoked with a new `PipelineState` for every message).

This will probably best be done after the refactor that @SmittieC was suggesting (moving the `PipelineState` into the invoke method), as you can't add arbitrary items onto a langchain state class and have them persist through the pipeline without them also being added to the first call. i.e. when invoking the pipeline we'd need to call it like this everywhere:

```
runnable.invoke(PipelineState(cached_history={}, messages=[user_input], experiment_session=experiment_session))
```

If it's OK with you, I might punt this one down the line as it will make the diff quite a bit larger on this PR. 

Here is what it would look like: https://github.com/dimagi/open-chat-studio/commit/9d3197889a5c791c8f0c5a2b5cb9db7917db7238

apps/pipelines/nodes/nodes.py
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/693/commits/72445d4572b68bbd14168420912257744e846f16

proteusvacuum:
> Possibly for a future PR but for named histories (node & named) we'll need to limit their size either by compression or filtering.

Yes, I'm working on this next. Might be worth doing it in a separate PR as there is a whole extra model to go with it.

apps/pipelines/nodes/nodes.py
snopoke: Fine to punt. On further consideration I'm not actually sure it's even relevant for our current use cases since even in the case of shared history, only one node would execute at a time. It would only make a difference if there were multiple nodes places one after another which used the same history.

apps/pipelines/nodes/nodes.py
```diff
@@ -79,7 +88,7 @@ def get_chat_model(self):
 class LLMResponse(PipelineNode, LLMResponseMixin):
     __human_name__ = "LLM response"
 
-    def _process(self, input, state: PipelineState) -> PipelineState:
+    def _process(self, input, state: PipelineState, node_id: str) -> PipelineState:
         llm = self.get_chat_model()
         output = llm.invoke(input, config=self._config)
         return output.content
```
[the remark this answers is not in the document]
snopoke: Would it be better to use `**kwargs` for unused arguments which are necessary for polymorphic consistency?

apps/pipelines/tests/test_pipeline_history.py
```diff
@@ -0,0 +1,530 @@
...
+from apps.utils.factories.pipelines import PipelineFactory
+from apps.utils.factories.service_provider_factories import LlmProviderFactory
+from apps.utils.langchain import (
+    FakeLlmEcho,
+    build_fake_llm_service,
+)
+from apps.utils.pytest import django_db_with_data
+
+
+@pytest.fixture()
```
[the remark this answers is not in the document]
proteusvacuum: `ruff` just adds these back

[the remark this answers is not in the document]
snopoke: There's a ticket to update the ruff config: https://github.com/dimagi/open-chat-studio/issues/711

snopoke (approved):

apps/pipelines/nodes/nodes.py
proteusvacuum: Yes, it's only valid if you have two or more nodes with the same history in the same pipeline.

proteusvacuum:
@coderabbitai pause

snopoke (approved):

PR #696, opened 2024-09-29
chat transcript view styling

snopoke:
## Description
Updated layout and styling for the chat transcript view. One outstanding issue (which can be resolved separately) is the rounding of the corners on the first and last row. For some reason I wasn't able to get this working.

The only think I haven't done which was discussed is provide some indication of pairing between user message and AI response.

### Demo
image


SmittieC (approved):
Looks great, thanks Simon

PR #702, opened 2024-10-02
Show routing tag in web UI when "debug mode" is enabled

SmittieC:
## Description

- Added a "debug_mode_enabled" field on the Experiments model. You can toggle this under the advanced tab
- When enabled, we show the `BOT_RESPONSE` tag in the conversation. Currently this is only used for routing purposes, so I think it's safe to just show this tag
- Added a new tag category for "Safety layer response". Safety layer responses are tagged with this. Also, I noticed we don't save default AI responses, so I made the change to save them and tag them as well. We also don't save human messages that triggers safety layer responses. Not 100% sure why, but I think we should.
- The way I got this to work is to get hold of the generated message instance and read the tag from that. We cannot simply find the latest AI message or the latest AI message where the text equals the generated text. This is not robust enough. We have to get the _correct_ message. If you follow the commits, you'll see that I first included the `message_id` in the response we get from the runnable, but man-o-man, a lot of things broke. Instead, I pivoted to save the AI message on the runnable's state and expose a getter method on the `TopicBot` to fetch the message_id from the runnable's state directly. It still feels a bit odd doing this, not going to lie, but having this "avenue" allows us to fetch more details from the state in the future if we wanted to without too much trouble.

## User Impact

The user will be able to see to which bot the message was relayed to in a routing setup when debug mode is enabled.

### Demo

https://www.loom.com/share/f73010def0fe4e1593d4983ddd944e2e?sid=aff942b4-2040-4176-b7cb-d7111922be35

image

### Docs

Will update when merged and link it here


apps/chat/bots.py
```diff
@@ -208,6 +213,9 @@ def main_bot_chain(user_input):
             }
         return main_bot_chain.invoke(user_input, config=config)
 
+    def get_ai_message_id(self) -> int | None:
```
SmittieC, line 216: This method essentialy relays information from the runnable's state to the caller and can easily be extended in the future to include other information that we might want to relay. I chose this "on-demand" method rather than returning some object as part of the original response for two reasons:
1. This way is much less effort
2. We don't have to return information when it's not going to be used. Currently only the webchannel cares about the message_id, so it can just ask for it whenever it needs it. I realize this isn't a very strong case for doing it this way in itself, but I'm going with it.

apps/chat/bots.py
```diff
@@ -208,13 +216,26 @@ def main_bot_chain(user_input):
...
         if safety_layer.prompt_to_bot:
-            safety_response = self._call_predict(safety_layer.prompt_to_bot, save_input_to_history=False)
-            return safety_response
+            bot_response = self._call_predict(safety_layer.prompt_to_bot, save_input_to_history=False)
         else:
             no_answer = "Sorry, I can't answer that. Please try something else."
-            return safety_layer.default_response_to_user or no_answer
+            bot_response = safety_layer.default_response_to_user or no_answer
+            # This is a bit of a hack to store the bot's response, since it didn't really generate it, but we still
+            # need to save it
```
stephherbers, line 231: since we'll move to pipelines soon (🤞) i think this is fine for now

PR #703, opened 2024-10-03
Save image files

SmittieC:
## Description

Currently we
1. don't show file attachments for AI messages. This changes in this PR
2. Ignore image files coming from OpenAI's assistant. We now save those as well.

Q: Why save image files as well?
A: Sometimes when the AI message doesn't reference a generated file, we miss it since it's not included as a message attachment. It is however available in the message contents array. What I've done is to save all images inside the message contents array as well.

Q: What happens when the image is in the message contents array and inside the text attachments? Will this not save duplicate images?
A: Yes it will. Turns out that the `file_id` for the near identical images are **not the same**. I use the term "near identical" since they differ in resolution, but the contents are the same. Even the file names are different. I could not find a robust way to check for duplicates yet. This is sub-optimal I know, but let's rather make the files available and possibly duplicate it than not having it and let the user wonder where the generated files are.

## User Impact

1. A list of file attachments will be visible in the 

### Demo

You can see in the image that the generated file is not being referenced in the text, but we make it available for download.

image

### Docs

No docs needed I think. This is more of a fix

### DIFF
--- a/apps/service_providers/llm_service/runnables.py
+++ b/apps/service_providers/llm_service/runnables.py
@@ -334,7 +334,8 @@ class AssistantExperimentRunnable(RunnableSerializable[dict, ChainOutput]):
         from apps.assistants.sync import get_and_store_openai_file
 
         client = self.state.raw_client
-        generated_files = []
+        chat = self.state.session.chat
+        session_id = self.state.session.id
 
         if isinstance(output, str):
             output_message = output
@@ -350,11 +351,15 @@ class AssistantExperimentRunnable(RunnableSerializable[dict, ChainOutput]):
         ).values_list("external_id", flat=True)
 
         file_ids = set()
+        image_file_attachments = []
+        file_path_attachments = []
         for message in client.beta.threads.messages.list(thread_id, run_id=run_id):
             for message_content in message.content:
                 if message_content.type == "image_file":
-                    # Ignore these for now. Typically, they are also referenced in the text content
-                    pass
+                    if created_file := self._create_image_file_from_image_message(client, message_content.image_file):
+                        image_file_attachments.append(created_file)
+                        file_ids.add(created_file.external_id)
+
                 elif message_content.type == "text":
                     annotations = message_content.text.annotations
                     for idx, annotation in enumerate(annotations):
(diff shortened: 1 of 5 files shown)


stephherbers:
> I could not find a robust way to check for duplicates yet.
lets make a backlog ticket for this, but I agree more data is better than losing data sometimes

apps/service_providers/llm_service/runnables.py
```diff
@@ -382,22 +387,56 @@ def _save_response_annotations(self, output, thread_id, run_id) -> tuple[str, di
...
+            file_id = image_file_message.file_id
+            openai_file = client.files.retrieve(file_id=file_id)
+            created_file = get_and_store_openai_file(
+                client=client,
+                file_name=f"{openai_file.filename}",
+                file_id=file_id,
+                team_id=self.state.experiment.team_id,
+            )
+            mimetype = magic.from_buffer(created_file.file.open().read(), mime=True)
+            extention = guess_extension(mimetype)
```
stephherbers, line 432: create an allowlist for the file types we expect for photos, we don't want any malicious file types getting through

stephherbers (changes requested):

apps/service_providers/llm_service/runnables.py
snopoke: Since this is coming from the bot I think it's unlikely to be malicious.

snopoke (approved):
nothing blocking

apps/service_providers/llm_service/runnables.py
```diff
@@ -382,22 +387,56 @@ def _save_response_annotations(self, output, thread_id, run_id) -> tuple[str, di
...
+        import magic
+
+        from apps.assistants.sync import get_and_store_openai_file
+
+        try:
+            file_id = image_file_message.file_id
+            openai_file = client.files.retrieve(file_id=file_id)
+            created_file = get_and_store_openai_file(
+                client=client,
+                file_name=f"{openai_file.filename}",
```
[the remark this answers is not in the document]
SmittieC: No, it's just a UUID

apps/service_providers/llm_service/runnables.py
SmittieC: These files are generated by the assistant, so I doubt it will be malicious?

snopoke (approved):

apps/service_providers/llm_service/runnables.py
SmittieC: > Since this is coming from the bot I think it's unlikely to be malicious.

I only saw this after reloading the page :smile:

PR #706, opened 2024-10-04
render generated images inline with the message

snopoke:
## Description
This updates the message processing and rendering to insert generated images as inline image links into the output markdown.

The markdown rendered will then render them as images.

## User Impact
Images are displayed in the message. This only applies to web chat when using Assistants to generate images.

### Demo
image

### Docs
NA

### DIFF
--- a/apps/service_providers/llm_service/runnables.py
+++ b/apps/service_providers/llm_service/runnables.py
@@ -287,7 +287,7 @@ class AssistantExperimentRunnable(RunnableSerializable[dict, ChainOutput]):
             input_dict["thread_id"] = current_thread_id
         input_dict["instructions"] = self.state.get_assistant_instructions()
         thread_id, run_id = self._get_response_with_retries(merged_config, input_dict, current_thread_id)
-        output, annotation_file_ids = self._save_response_annotations(thread_id, run_id)
+        output, annotation_file_ids = self._get_output_with_annotations(thread_id, run_id)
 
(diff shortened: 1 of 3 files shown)


SmittieC (dismissed):
cool

apps/service_providers/llm_service/runnables.py
```diff
@@ -405,39 +408,27 @@ def _save_response_annotations(self, output, thread_id, run_id) -> tuple[str, di
             resource, _created = chat.attachments.get_or_create(tool_type="image_file")
             resource.files.add(*image_file_attachments)
 
-        return output_message, list(file_ids)
+        return output_message.strip(), list(file_ids)
```
SmittieC, line 411: Should this be `output_message.strip('\n')`, since we might be adding an extra newline character?

SmittieC: nevermind, I see strip removed newlines as well by default

SmittieC (dismissed):
This is quite cool, thanks. Do you think we should make the image a bit smaller? Not sure how easy that would be.

snopoke:
> This is quite cool, thanks. Do you think we should make the image a bit smaller? Not sure how easy that would be.

Perhaps but I'm considering this out of scope for now. We could potentially implement it using markdown without resizing the image itself: https://stackoverflow.com/a/21242579

SmittieC (approved):

PR #707, opened 2024-10-04
Make safety layer tags red

SmittieC:
## Description

Red safety layer violation tags

## User Impact

N/A

### Demo

this one

### Docs

N/A

### DIFF
--- a/apps/annotations/models.py
+++ b/apps/annotations/models.py
@@ -98,20 +98,29 @@ class TaggedModelMixin(models.Model, AnnotationMixin):
         self.tags.add(tag, through_defaults={"team": team, "user": added_by})
 
     def get_linked_tags(self):
-        return self.tags.all()
+        tags = []
+        for tag in self.tags.all():
+            bage_color = "badge-neutral"
+            if tag.is_system_tag:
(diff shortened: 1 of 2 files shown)


apps/annotations/models.py
```diff
@@ -98,20 +98,29 @@ def add_tag(self, tag: Tag, team: Team, added_by: CustomUser):
         self.tags.add(tag, through_defaults={"team": team, "user": added_by})
 
     def get_linked_tags(self):
-        return self.tags.all()
+        tags = []
+        for tag in self.tags.all():
+            bage_color = "badge-neutral"
+            if tag.is_system_tag:
+                bage_color = "badge-error" if tag.category == TagCategories.SAFETY_LAYER_RESPONSE else "badge-warning"
```
snopoke, line 105: It can be dangerous to use css classes in Python since the post processor will remove classes that aren't used and it doesn't currently check python files for usages.

What about doing this logic in the template and just using 'tags_json' instead of this method? I think we could get rid of this method entirely if we did that.

SmittieC: https://github.com/dimagi/open-chat-studio/pull/707/commits/41776c6d115b3d487621f811b232d70b966fb349

snopoke (approved):

PR #712, opened 2024-10-09
work around empty database when slack is setup

snopoke:
This is quite an edge case but something that came up during my AWS migration testing.


SmittieC (approved):
Seems fine to me.

PR #718, opened 2024-10-10
Versioning updates 4 | Better diff view

SmittieC:
Parking this here for now.

This is a first pass at more granular diffs for string values using python's [difflib](https://docs.python.org/3/library/difflib.html). 
Rendering it happens character wise currently, since the info regarding the characters' addition or removal is character-wise. Rendering it character-wise introduces some weird spacing issues that I'm yet to solve.

### What it looks like
Screenshot from 2024-10-17 11-40-32

FYI cc @snopoke @stephherbers


apps/experiments/versioning.py
```diff
@@ -124,6 +134,34 @@ def _compare_queryset(self, previous_queryset):
...
+
+        differ = Differ()
+        difflines = list(differ.compare(self.previous_field_version.raw_value, self.raw_value))
+        operations = {
+            "no_change": " ",
+            "removed": "-",
+            "added": "+",
+        }
+
+        print(difflines)
```
snopoke, line 148:
```suggestion
```

snopoke (commented):
Looks pretty good to me.

ChatGPT recommend this approach:
```
<div class="prose max-w-none">
    <span>Hello</span>
    <span class="text-green-600 bg-green-100">world</span>
    <span class="text-red-600 bg-red-100 line-through">!</span>
</div>
```

SmittieC:
> Looks pretty good to me.
> 
> ChatGPT recommend this approach:
> 
> ```
> <div class="prose max-w-none">
>     <span>Hello</span>
>     <span class="text-green-600 bg-green-100">world</span>
>     <span class="text-red-600 bg-red-100">!</span>
> </div>
> ```

This looks a bit worse tbh. We can maybe roll out the existing stuff and let people benefit from it so long? This is what it looks like now:
image

snopoke (approved):

PR #734, opened 2024-10-16
Add get_ai_message_id to PipelineBot

proteusvacuum:
## Description

The `PipelineBot` didn't have the `get_ai_message_id` parameter, and so chatting with one was failing. This was introduced here: https://github.com/dimagi/open-chat-studio/pull/702

I added a test that failed without this change. 

## User Impact

Users were unable to chat with Pipeline Bots. Now they can.

### DIFF
--- a/apps/chat/bots.py
+++ b/apps/chat/bots.py
@@ -8,7 +8,7 @@ from pydantic import ValidationError
 from apps.annotations.models import TagCategories
 from apps.chat.conversation import BasicConversation, Conversation
 from apps.chat.exceptions import ChatException
-from apps.chat.models import ChatMessageType
+from apps.chat.models import ChatMessage, ChatMessageType
 from apps.events.models import StaticTriggerType
 from apps.events.tasks import enqueue_static_triggers
 from apps.experiments.models import Experiment, ExperimentRoute, ExperimentSession, SafetyLayer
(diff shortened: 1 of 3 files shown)


apps/chat/bots.py
```diff
@@ -287,3 +287,13 @@ def process_input(self, user_input: str, save_input_to_history=True, attachments
             PipelineState(messages=[user_input], experiment_session=self.session), self.session
         )
         return output["messages"][-1]
+
+    def get_ai_message_id(self) -> int | None:
```
proteusvacuum, line 291: Not sure if this is the right way of going about this... Currently we always save history for a pipeline run, but that might not always be true. I guess in that case this would be None, and we move on as before 🤷🏽‍♂️ . However, this does feel prone to race conditions if this is fetched for some reason before the message was committed...

apps/experiments/views/experiment.py
```diff
@@ -837,7 +837,10 @@ def get_message_response(request, team_slug: str, experiment_id: int, session_id
     message_details = {"message": None, "error": False, "complete": progress["complete"]}
     if progress["complete"] and progress["success"]:
         result = progress["result"]
-        message_details["message"] = ChatMessage.objects.get(id=result["message_id"])
+        if result["message_id"]:
```
proteusvacuum, line 840: The return type for `get_ai_message_id` also includes `None`, so I added this check.

apps/chat/bots.py
SmittieC: A more robust way would be for the pipeline to return the AI generated `ChatMessage`. I am doing some work in this area at the moment for versioning, so I'll update the code accordingly. I think this approach is fine for now. It works

SmittieC (approved):

PR #740, opened 2024-10-17
Summarize Pipeline Chat History

proteusvacuum:
## Description
Adds the ability to summarize `node` and `named` histories. It stores the summary in a similar way to the regular chat messages models.

## User Impact

Users can have long-running chats with Pipelines where the nodes have individual / shared histories.


apps/pipelines/migrations/0008_pipelinechatmessages_node_id.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 5.1 on 2024-10-17 14:19
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
SmittieC, line 6: Nit: How about merging this and the previous migration?

apps/pipelines/models.py
```diff
@@ -227,6 +227,7 @@ def get_langchain_messages_until_summary(self):
 
 class PipelineChatMessages(BaseModel):
     chat_history = models.ForeignKey(PipelineChatHistory, on_delete=models.CASCADE, related_name="messages")
+    node_id = models.TextField()
```
SmittieC, line 230: So this node id's purpose is purely informational?

apps/chat/conversation.py
```diff
@@ -114,45 +115,75 @@ def compress_chat_history(
...
     except (NameError, ImportError, ValueError, NotImplementedError):
         # typically this is because a library required to count tokens isn't installed
         log.exception("Unable to compress history")
-        return chat.get_langchain_messages_until_summary()
+        return history_messages
 
 
-def _compress_chat_history(
-    chat: Chat, llm: BaseChatModel, max_token_limit: int, input_messages: list, keep_history_len: int = 10
+def compress_pipeline_chat_history(
```
SmittieC, line 137: The only difference between this method and the `compress_chat_history` above it seems to be some class types i.e. `Chat` <-> `PipelineChatHistory` and `ChatMessage` <-> `PipelineChatMessages`. Looks like this method generally expects something that implements `get_langchain_messages_until_summary` and it needs some way of storing the new summary. I'm seeing some small abstraction that we can do here, maybe a thin wrapper class of sorts?

apps/pipelines/models.py
```diff
@@ -1,10 +1,12 @@
+from collections.abc import Iterator
 from datetime import datetime
 from functools import cached_property
 
 import pydantic
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
 from django.urls import reverse
+from langchain_core.messages import AIMessage, BaseMessage, HumanMessage, SystemMessage
```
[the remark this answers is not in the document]
snopoke: Either of these appear to be correct

apps/pipelines/models.py
```diff
@@ -195,17 +197,66 @@
...
+        langchain_messages = [
+            AIMessage(content=self.ai_message, additional_kwargs={"id": self.id, "node_id": self.node_id}),
+            HumanMessage(content=self.human_message, additional_kwargs={"id": self.id, "node_id": self.node_id}),
+        ]
+        if self.summary:
+            langchain_messages.append(
+                SystemMessage(content=self.summary, additional_kwargs={"id": self.id, "node_id": self.node_id})
+            )
+
+        return langchain_messages
```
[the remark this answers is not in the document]
snopoke: This order is correct because the method is returning the messages in reverse order. I do think a docstring would be useful to explain this.

apps/pipelines/models.py
```diff
@@ -195,17 +197,66 @@ class PipelineChatHistory(BaseModel):
...
+            message_tuples.append((ChatMessageType.SYSTEM.value, self.summary))
+        message_tuples.extend(
+            [
+                (ChatMessageType.HUMAN.value, self.human_message),
+                (ChatMessageType.AI.value, self.ai_message),
+            ]
+        )
+        return message_tuples
+
+    def as_langchain_messages(self) -> list[BaseMessage]:
```
snopoke, line 252: @coderabbitai generate a docstring for this method

snopoke:
```suggestion
    def as_langchain_messages(self) -> list[BaseMessage]:
        """
        Converts this message instance into a list of Langchain `BaseMessage` objects.
        The message order is the reverse of the typical order because of where this
        method is called. The returned order is: [`AIMessage`, `HumanMessage`, `SystemMessage`].
        
        The `SystemMessage` represents the conversation summary and will only be
        included if it exists.
        """
```

apps/pipelines/models.py
proteusvacuum: Currently, yes.

apps/chat/conversation.py
proteusvacuum: Yes, you are correct, and it does feel like somewhat duplicated code, although I couldn't really think of another way of writing this that was as clear as this. I thought about adding the types as parameters to the original function, but it is less clear what is going on then.

Do you think adding a wrapper class would be clearer that what is happening here? Personally, I think that would just be a level of indirection more, but I'm happy to change it if you'd like me to.

apps/pipelines/migrations/0008_pipelinechatmessages_node_id.py
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/740/commits/d748ecf69ddf279c631019311c347ecf3cb1a04f

apps/chat/conversation.py
SmittieC: Nah if you've explored it and found that this implementation was clearer, then I'm happy. Thanks for clarifying :+1:

SmittieC (approved):
Looks fine to me

apps/chat/conversation.py
proteusvacuum: Thanks!

PR #769, opened 2024-10-21
Pipeline UX updates

### LINKED ISSUE #716: pipeline UX improvements
### General
* Nodes should have some [help text](https://daisyui.com/components/dropdown/#helper-dropdown) giving more info about what they do
* ~~Move node list to dropdown. See https://flowiseai.com/~~

### Router

* Clicking the up / down arrow of the number of outputs input makes the value increase/ decrease indefinitely
* Output nodes should be numbered or named to make it easier to know which  is which

SmittieC:
Resolved #716 

## Description

1. Notify the user whenever the pipeline is saved
2. Ability to save the pipeline using `ctrl+s`
3. Node descriptions in the side menu
4. Annotate node output handles to make it easier to know which is which. This will only be annotated when there are more than 1 output handle
5. Ability to delete a node connection using the `delete` key as well
6. Removed the spinner controls for all inputs with `type=number`, since they kept incrementing/decrementing when clicked. I only got this in Chrome, not Firebox (didn't test other browsers).

## User Impact

UI improvements

#### Node descriptions
image

#### Handle output annotations
image

### Docs

N/A

### DIFF
--- a/assets/javascript/apps/pipeline/PipelineNode.tsx
+++ b/assets/javascript/apps/pipeline/PipelineNode.tsx
@@ -238,14 +238,17 @@ export function PipelineNode({ id, data, selected }: NodeProps<NodeData>) {
       { length: numberOfOutputs },
       (_, index) => {
         const position = (index / (numberOfOutputs - 1)) * 100; // Distributes evenly between 0% to 100%
+        const handleText = `Output ${index + 1}`;
         return (
-          <Handle
-            key={`output_${index}`}
-            type="source"
-            position={Position.Right}
-            style={{ top: `${position}%` }}
-            id={numberOfOutputs > 1 ? `output_${index}` : "output"}
-          />
+            <Handle
+              key={`output_${index}`}
+              type="source"
+              position={Position.Right}
+              style={{ top: `${position}%` }}
+              id={numberOfOutputs > 1 ? `output_${index}` : "output"}
+            >
+              <div className="handle-text">{handleText}</div>
+            </Handle>
         );
       },
     );
--- a/assets/javascript/apps/pipeline/PipelineNode.tsx
+++ b/assets/javascript/apps/pipeline/PipelineNode.tsx
@@ -267,7 +267,7 @@ export function PipelineNode({ id, data, selected }: NodeProps<NodeData>) {
       <div
         className={classNames(
           selected ? "border border-primary" : "border",
-          "px-4 py-2 shadow-md rounded-md border-2 border-stone-400 bg-base-100",
+          "px-4 py-2 shadow-md rounded-xl border-2 border-stone-400 bg-base-100",
         )}
       >
         <Handle type="target" position={Position.Left} id="input" />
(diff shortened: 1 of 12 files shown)


assets/javascript/apps/pipeline/PipelineNode.tsx
```diff
@@ -238,7 +238,12 @@ export function PipelineNode({ id, data, selected }: NodeProps<NodeData>) {
       { length: numberOfOutputs },
       (_, index) => {
         const position = (index / (numberOfOutputs - 1)) * 100; // Distributes evenly between 0% to 100%
-        const handleText = `Output ${index + 1}`;
+        let handleAnnotation;
```
SmittieC, line 241: It would be nice if the widgets can optionally output where _it_ would like its output handle to be (if it is tied to one). This way each widget can optionally output a handle position, equal to its own Y position. We can then simply iterate over the list of output position requests and put the handles there. This way we can show the handles inline with the widget. I am not yet convinved that this is worth undertaking though, but just a thought

assets/javascript/apps/pipeline/types/pipelineManagerStore.ts
```diff
@@ -7,7 +7,7 @@ export type PipelineManagerStoreType = {
   loadPipeline: (pipelineId: number) => void;
   isLoading: boolean;
   setIsLoading: (isLoading: boolean) => void;
-  savePipeline: (pipelne: PipelineType, silent?: boolean) => Promise<void>;
+  savePipeline: (pipelne: PipelineType, isAutoSave?: boolean, silent?: boolean) => Promise<void>;
```
snopoke, line 10: `silent` doesn't appear to be used - you could remove it.

templates/pipelines/pipeline_builder.html
```diff
@@ -18,6 +18,14 @@
...
       transform: translate(0, -50%);
     }
+
+    .handle-text {
+      font-weight: bold;
+      position: relative;
+      width: 100px;
+      transform: translate(15px, -15px);
+      pointer-events: none;
+    }
```
snopoke, line 28: This could be replaced with the following tailwind classes:

```
font-bold relative w-24 translate-x-[15px] -translate-y-[15px] pointer-events-none
```

assets/javascript/apps/pipeline/stores/pipelineManagerStore.ts
```diff
@@ -50,6 +50,7 @@ const usePipelineManagerStore = create<PipelineManagerStoreType>((set, get) => (
           if (updatedFlow) {
             set({currentPipeline: pipeline});
             resolve();
+            // @ts-expect-error. This module is available
```
snopoke, line 53: I think you could update the `eslint` config to add `alertify` as a global.

snopoke (approved):
Mostly small comments

assets/javascript/apps/pipeline/stores/pipelineManagerStore.ts
SmittieC: Cool. I'll do so in a new PR. Thanks for the tips

snopoke: Turns out it was something different. I'll PR a change.

PR #774, opened 2024-10-23
Sync with openAI when editing assistant 

stephherbers:
## User Impact

Now the sync will happen automatically without having to press the sync button

### Demo

Screen Recording 2024-10-30 at 3 17 36 PM

(demo shows warning if files are out of sync as well)

### Docs

to be updated

### DIFF
--- a/apps/assistants/sync.py
+++ b/apps/assistants/sync.py
@@ -195,7 +195,29 @@ def are_files_in_sync_with_openai(assistant: OpenAiAssistant) -> bool:
         ocs_file_ids = [file.external_id for file in resource.files.all() if file.external_id]
         if set(ocs_file_ids) != set(openai_file_ids):
             return False
+
+    if not check_tool_enabled(
+        "code_interpreter", tool_resources, client, assistant.assistant_id
+    ) or not check_tool_enabled("file_search", tool_resources, client, assistant.assistant_id):
+        return False
+
     return True
+
+
+def check_tool_enabled(tool_type: str, tool_resources: list, client, assistant_id: str):
+    """Check if a specific tool type is enabled in OpenAI."""
+    if tool_type in tool_resources:
+        return True
+    try:
+        if tool_type == "code_interpreter":
+            tool_resources_data = client.beta.assistants.retrieve(assistant_id).tool_resources
+        elif tool_type == "file_search":
+            tool_resources_data = client.beta.vector_stores.files
+        else:
+            return False
+        return bool(tool_resources_data)
+    except Exception:
+        return False
 
 
 @wrap_openai_errors
--- a/apps/assistants/sync.py
+++ b/apps/assistants/sync.py
@@ -201,16 +201,17 @@ def are_files_in_sync_with_openai(assistant: OpenAiAssistant) -> bool:
 
 def check_tool_enabled(tool_type: str, tool_resources: list, client, assistant_data):
     """Check if a specific tool type is enabled in OpenAI."""
-    if tool_type in tool_resources:
+    if tool_type in [resource.tool_type for resource in tool_resources]:
         return True
     try:
+        assistant_tool_resources = assistant_data.tool_resources
         if tool_type == "code_interpreter":
-            tool_resources_data = assistant_data.tool_resources
+            tool_resources_data = assistant_tool_resources.code_interpreter
         elif tool_type == "file_search":
-            tool_resources_data = client.beta.vector_stores.files
+            tool_resources_data = assistant_tool_resources.file_search
         else:
             return False
-        return bool(tool_resources_data)
+        return tool_resources_data is not None
     except Exception:
         return False
 
--- a/apps/assistants/sync.py
+++ b/apps/assistants/sync.py
@@ -198,6 +198,17 @@ def are_files_in_sync_with_openai(assistant: OpenAiAssistant) -> bool:
     return True
 
 
+@wrap_openai_errors
+def is_synced_with_openai(assistant: OpenAiAssistant):
+    client = assistant.llm_provider.get_llm_service().get_raw_client()
+    openai_assistant = client.beta.assistants.retrieve(assistant.assistant_id)
+    for key, value in _openai_assistant_to_ocs_kwargs(openai_assistant).items():
+        current_value = getattr(assistant, key, None)
+        if current_value != value:
+            return False
+    return True
+
+
 def _fetch_file_from_openai(assistant: OpenAiAssistant, file_id: str) -> File:
     client = assistant.llm_provider.get_llm_service().get_raw_client()
     openai_file = client.files.retrieve(file_id)
--- a/apps/assistants/sync.py
+++ b/apps/assistants/sync.py
@@ -166,6 +166,38 @@ def delete_openai_assistant(assistant: OpenAiAssistant):
             delete_file_from_openai(client, file)
 
 
+@wrap_openai_errors
+def are_files_in_sync_with_openai(assistant: OpenAiAssistant) -> bool:
+    """Checks if the files for an assistant in OCS match the files in OpenAI."""
+    tool_resources = assistant.tool_resources.all()
+    client = assistant.llm_provider.get_llm_service().get_raw_client()
+
+    for resource in tool_resources:
+        openai_file_ids = []
+        if resource.tool_type == "code_interpreter":
+            try:
+                openai_file_ids = client.beta.assistants.retrieve(
+                    assistant.assistant_id
+                ).tool_resources.code_interpreter.file_ids
+            except AttributeError:
+                openai_file_ids = []
+        elif resource.tool_type == "file_search":
+            try:
+                openai_vector_store_id = resource.extra.get("vector_store_id")
+                if not openai_vector_store_id:
+                    continue
+                openai_file_ids = [
+                    file.id
+                    for file in client.beta.vector_stores.files.list(vector_store_id=openai_vector_store_id).data
+                ]
+            except AttributeError:
+                openai_file_ids = []
+        ocs_file_ids = [file.external_id for file in resource.files.all() if file.external_id]
+        if set(ocs_file_ids) != set(openai_file_ids):
+            return False
+    return True
+
+
 def _fetch_file_from_openai(assistant: OpenAiAssistant, file_id: str) -> File:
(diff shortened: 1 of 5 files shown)


snopoke (commented):
I think we want to check for other changes as well, not just files. 

Also, we shouldn't rely on the the data in OCS to know what to check  e.g. file_search may not even be enabled in OCS but could be enabled in OpenAI.

apps/assistants/views.py
```diff
@@ -140,6 +159,18 @@ def form_valid(self, form):
         return response
 
 
+class SyncEditingOpenAiAssistant(BaseOpenAiAssistantView, View):
+    permission_required = "assistants.change_openaiassistant"
+
+    def post(self, request, team_slug: str, pk: int):
+        assistant = get_object_or_404(OpenAiAssistant, team=request.team, pk=pk)
```
SmittieC, line 166: Nit: doing `request.team` will fire a DB query. Rather check the team using the `team_slug`

apps/assistants/views.py
```diff
@@ -124,6 +126,23 @@ class EditOpenAiAssistant(BaseOpenAiAssistantView, UpdateView):
     button_text = "Update"
     permission_required = "assistants.change_openaiassistant"
 
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        try:
+            is_synced = is_synced_with_openai(self.object)
```
SmittieC, line 132: What's your thoughts on doing this check asynchronously? Using htmx, this should be easy enough I'd imagine. In my mind we have a sync status indicator. We can play with a few icons to indicate "checking status", "out of sync", "in sync" and "check failed". The out of sync status can be a nice scary red that draws attention. Up to you

stephherbers:
@coderabbitai please review

apps/assistants/views.py
stephherbers: I forgot why I did this and made the change and remembered, team_slug is not in the correct format and causes and error when used instead. This is akso used in a few other views on the page so if you have and ideas lmk and I can fix them all

apps/assistants/views.py
stephherbers: I like that and made the fix

stephherbers:
@snopoke I would like your input on whether `are_files_in_sync_with_openai` properly checks the file status becuase this is what I'm unsure about

apps/assistants/views.py
```diff
@@ -124,6 +126,13 @@ class EditOpenAiAssistant(BaseOpenAiAssistantView, UpdateView):
     button_text = "Update"
     permission_required = "assistants.change_openaiassistant"
 
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        if are_files_in_sync_with_openai(self.object):
```
SmittieC, line 131: This should call `is_synced_with_openai` instead?

apps/assistants/views.py
snopoke: This should be fine since we'll almost always have to load the team for a request. But if you do want to use `team_slug` you'll need to do it like this `team__slug=team_slug`

templates/assistants/assistant_form.html
```diff
@@ -13,9 +13,30 @@
...
+    <div id="sync-status"
+         hx-get="{% url 'assistants:check_sync_status' request.team.slug object.pk %}"
+         hx-trigger="load"
+         hx-target="#sync-container"
+         class="flex w-full gap-2">
+      <span class="indicator" id="sync-indicator"></span>
+      <span id="sync-message">Checking sync status...</span>
+    </div>
+
+    <div id="sync-container" hx-on="htmx:afterSwap: this.previousElementSibling.remove();">
```
snopoke, line 34: Is the call to `previousElementSibling` to remove the `sync-status` div above? I think if you change the `hx-target` to `sync-wrapper` you won't need to do this (maybe also set `hx-swap="innerHTML"`

apps/assistants/views.py
snopoke: Can this call not be bundled with the `check_sync_status` view?

apps/assistants/views.py
```diff
@@ -140,6 +149,31 @@ def form_valid(self, form):
...
+class SyncEditingOpenAiAssistant(BaseOpenAiAssistantView, View):
+    permission_required = "assistants.change_openaiassistant"
+
+    def post(self, request, team_slug: str, pk: int):
+        assistant = get_object_or_404(OpenAiAssistant, team=request.team, pk=pk)
+        try:
+            sync_from_openai(assistant)
+        except OpenAiSyncError as e:
+            messages.error(request, f"Error syncing assistant: {e}")
+        return check_sync_status(request, team_slug, pk)
```
snopoke, line 174: Instead of re-doing the check here, this should probably do a page reload so that any changes that were pulled in are also reflected in the form. This will also then re-trigger the sync check on page load.

You can do that by setting the `HX-Refresh` header to true:

```suggestion
        return HttpResponse(headers={"HX-Refresh": "true"})
```

apps/assistants/views.py
stephherbers: bundled here: https://github.com/dimagi/open-chat-studio/pull/774/commits/17d212d6230f6e0d322a7881ff4a740479e591aa

apps/assistants/sync.py
```diff
@@ -166,6 +166,71 @@ def delete_openai_assistant(assistant: OpenAiAssistant):
...
+    client = assistant.llm_provider.get_llm_service().get_raw_client()
+
+    for resource in tool_resources:
+        openai_file_ids = []
+        if resource.tool_type == "code_interpreter":
+            try:
+                openai_file_ids = client.beta.assistants.retrieve(
+                    assistant.assistant_id
+                ).tool_resources.code_interpreter.file_ids
+            except AttributeError:
```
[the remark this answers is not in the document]
stephherbers: https://github.com/dimagi/open-chat-studio/pull/774/commits/cd86627d710ed131b8e01ad423b0739914f5b9d5

templates/assistants/assistant_form.html
stephherbers: thanks I was playing around with this and that solution is just what I was looking for https://github.com/dimagi/open-chat-studio/pull/774/commits/a93fd4d5b3916ef6ca8632eedeb0217785b86797

apps/assistants/sync.py
```diff
@@ -166,6 +166,71 @@ def delete_openai_assistant(assistant: OpenAiAssistant):
...
+    if tool_type in tool_resources:
+        return True
+    try:
+        if tool_type == "code_interpreter":
+            tool_resources_data = client.beta.assistants.retrieve(assistant_id).tool_resources
+        elif tool_type == "file_search":
+            tool_resources_data = client.beta.vector_stores.files
+        else:
+            return False
+        return bool(tool_resources_data)
```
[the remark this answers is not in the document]
stephherbers: thanks code rabbit here's the fix: https://github.com/dimagi/open-chat-studio/pull/774/commits/e761caa90a2588c9161a502679523044853a1258 🥕

apps/assistants/sync.py
```diff
@@ -166,6 +166,71 @@ def delete_openai_assistant(assistant: OpenAiAssistant):
...
+    ) or not check_tool_enabled("file_search", tool_resources, client, assistant.assistant_id):
+        return False
+
+    return True
+
+
+def check_tool_enabled(tool_type: str, tool_resources: list, client, assistant_id: str):
+    """Check if a specific tool type is enabled in OpenAI."""
+    if tool_type in tool_resources:
+        return True
```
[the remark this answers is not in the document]
stephherbers: thanks code rabbit here's the fix: https://github.com/dimagi/open-chat-studio/pull/774/commits/e761caa90a2588c9161a502679523044853a1258 🥕

stephherbers:
going back though the logic of `are_files_in_sync_with_openai` I realize that it wasn't quite checking what I wanted in terms of the case if a tool is being used in OpenAI but not found in OCS. I revised the logic to cover these cases:

1 | A tool is enabled in OpenAI but not locally | False
-- | -- | --

2 | A tool is locally configured but not enabled in OpenAI | True (allowed)
-- | -- | --

3 | Files associated with a tool type differ between OCS and OpenAI | False
-- | -- | --

4 | All tools and files are synchronized between OCS and OpenAI | False
-- | -- | --

stephherbers:
@coderabbitai please review again specifically the logic of are_files_in_sync_with_openai

templates/assistants/assistant_form.html
```diff
@@ -16,6 +16,21 @@
   {% if object.id %}
     <h2 class="text-l pg-text-muted">Assistant ID: {{ object.assistant_id }}</h2>
   {% endif %}
+  <div id="sync-wrapper" class="flex w-full gap-2"
+       hx-get="{% url 'assistants:check_sync_status' request.team.slug object.pk %}"
```
SmittieC, line 20: This would need to be done only when `object` exists. For new assistants, object will be `None`

apps/assistants/sync.py
```diff
@@ -166,6 +166,60 @@ def delete_openai_assistant(assistant: OpenAiAssistant):
             delete_file_from_openai(client, file)
 
 
+def is_tool_configured_remotely_but_missing_locally(assistant_data, local_tool_types, tool_name: str) -> bool:
+    """Checks if a tool is configured in OpenAI but missing in OCS."""
+    tool_configured_in_openai = hasattr(assistant_data.tool_resources, tool_name) and getattr(
```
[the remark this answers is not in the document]
SmittieC: Oh wait I see it's `None` when it's not configured. Deleting

templates/assistants/assistant_form.html
snopoke: you could wrap the whole 'sync-wrapper' div in an if statement `{% if object.pk %}`

snopoke (commented):
Looks good pending the one comment

templates/assistants/assistant_form.html
stephherbers: good call https://github.com/dimagi/open-chat-studio/pull/774/commits/28567c77c22303ff4725d5de803a462f33f72798

SmittieC (approved):
:rocket:

PR #781, opened 2024-10-28
Upgrade to Langchain 0.3

### LINKED ISSUE #415: Upgrade to Langchain v0.2
See [the guide](https://python.langchain.com/v0.2/docs/versions/v0_2/) on upgrading to v0.2.

SmittieC:
resolves #415 

## Description

Upgraded to Langchain 2 and then 3. I see this deprecation warning in the shell

> LangChainDeprecationWarning: Please see the migration guide at: https://python.langchain.com/docs/versions/migrating_memory/

Looks like they really want us to use langgraph's persistance, or at least `RunnableWithMessageHistory`. To be clear, everything still works, and this will only be deprecated in future versions, so we're fine with what we have for now, but I think this is a good time to pause and think about the future of the TopicBot. Thoughts welcome.

Two things to note from this upgrade:
1. Langchain version > 0.2 doesn't play along with the `freezegun` library super well, so I had to update some tests using freezegun to not use it anymore.
2. The pipeline logger seems to log 1 less message than before. In the `test_running_failed_pipeline_logs_error` test, we force the two `Passthrough` nodes (the only two in the test pipeline) to raise an exception when called. Previously, it generated 3 error logs in total for this exception, but now it generates 2 error logs, one for each node which seems correct. Since the library upgrades were the only things that happened, I'm assuming it has to do with internal callbacks. cc @proteusvacuum fyi

### Testing the upgrade
I tested by running the unittests and by using the platform normally, with tools, with a pipeline, an assistant, an assistant with tools, source material etc. All seems fine.

## User Impact

No impact

### Demo

N/A

### Docs

N/A


SmittieC:

apps/experiments/tasks.py
```diff
@@ -2,7 +2,7 @@
 from datetime import datetime
 
 from celery.app import shared_task
-from langchain.schema import AIMessage, HumanMessage
+from langchain_core.messages import AIMessage, HumanMessage
```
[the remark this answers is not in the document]
SmittieC: These suggestions are incorrect for these use cases.

apps/service_providers/tests/test_assistant_runnable.py
```diff
@@ -590,5 +590,6 @@ def _create_run(
         status=status,
         thread_id=thread_id,
         tools=[],
+        parallel_tool_calls=False,
```
snopoke, line 593: can you say more about this change

apps/service_providers/tests/test_runnables.py
```diff
@@ -52,7 +52,7 @@ def build(self, *args, **kwargs):
 
 
 runnables = {
-    "simple": RunnableFixture(SimpleExperimentRunnable),
+    # "simple": RunnableFixture(SimpleExperimentRunnable),
```
snopoke, line 55: need to uncomment this

requirements/dev-requirements.txt
```diff
@@ -2,118 +2,122 @@
...
     #   httpcore
     #   httpx
 termcolor==2.4.0
-    # via -r dev-requirements.in
+    # via -r requirements/dev-requirements.in
 virtualenv==20.26.3
     # via pre-commit
 watchfiles==0.22.0
-    # via -r dev-requirements.in
+    # via -r requirements/dev-requirements.in
```
snopoke, line 121: nit: it would be great to undo these comment changes since it makes it hard to see what's changed. 

I've updated the `requirements` task to prevent this in future: https://github.com/dimagi/open-chat-studio/commit/d1d8be2dec8a0b9dadb49bda06104c94cbb7ca5e

snopoke (commented):
These changes look OK to me

apps/service_providers/tests/test_assistant_runnable.py
SmittieC: I think [this link](https://platform.openai.com/docs/guides/function-calling/configuring-parallel-function-calling) explains it best. Basically do we want the model to output multiple function calls, or one at a time. This is a new required param, and since this is a test, the value of this param doesn't really matter

apps/service_providers/tests/test_runnables.py
SmittieC: ah thanks

requirements/dev-requirements.txt
SmittieC: https://github.com/dimagi/open-chat-studio/pull/781/commits/d951b0c1f9571eb6d3127820a5f8a955d1d5cb32

SmittieC:
Sorry, setting back to draft. There's a few coderabbit comments I wanna check out

apps/chat/agent/tools.py
```diff
@@ -3,7 +3,7 @@
 from typing import Any
 
 from django.db import transaction
-from langchain.tools.base import BaseTool
+from langchain_core.tools import BaseTool
```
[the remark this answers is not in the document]
SmittieC: So many (if not all) of these imports are doing an import from `langchain_core` under the hood.

requirements/requirements.txt
```diff
@@ -227,45 +230,53 @@ jsonschema-specifications==2023.7.1
...
     #   langchain-anthropic
     #   langchain-community
     #   langchain-openai
     #   langchain-text-splitters
     #   langgraph
-langchain-openai==0.1.3
-    # via -r requirements.in
-langchain-text-splitters==0.0.1
+    #   langgraph-checkpoint
+langchain-openai==0.2.3
```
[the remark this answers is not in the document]
SmittieC: The current imports are fine. They are not deprecated and some of them imports from `langchain_core` anyways

snopoke (approved):

PR #793, opened 2024-10-28
Verify participants in public chats

### LINKED ISSUE #732: Improve user management for public bot links
### The problem
When someone follows a public link to chat to a bot and enters an email address that belongs to a platform user, they will not be able to send the message, since they are not logged in. There's no feedback though, so this is confusing.

### Proposed solution
We send the chat link to the specified email address (I think Simon suggested this way back?). This serves as a way to verify the email address (better security) and if we see that this belongs to a platform user, the link can look something like "https://chatbots.com/accounts/login?next=<the chat link>", thus forcing the user to log in first.

Toughts welcome

SmittieC:
Resolves #732 

## Description

[Spec doc](https://docs.google.com/document/d/1aSFeRWy2nxsyMa03kcF6Bps3l-ApFqxpM30f11tA36w/edit?tab=t.0#heading=h.nso8knmuctmt).

Notable changes:
1. Participant emails will be verified by means of a link that is sent to the specified email address when
- they are not logged in
- they have a `chat_session_access` cookie, but the specified participant doesn't match the one in the cookie
- they don't have a `chat_session_access` cookie
2. Decorated `experiment_session_message` and `get_message_response` with the `verify_session_access_cookie` as well.

The link contains a JWT with the session id in it. This token expires after 5 minutes. When the user clicks on the link in the email, we'll verify the token, and if valid, update the `chat_session_access` cookie on the request with the new session's details and redirect the user to the chat. If the token is invalid, they are redirected to the consent form where they must start over.

Q: If someone finds a JWT lying around, can they use the `session_id` in there and chat to that session?
A: Only when the JWT is still valid (you have 5 mins buddy), otherwise no. The `experiment_chat` view is decorated with the `verify_session_access_cookie` decorator that will prevent the user from getting to the chat view, but even if they try to hit the `experiment_session_message` and `get_message_response` views directly, they will still be forbidden.

We now decorate the `experiment_session_message` and `get_message_response` views with the `experiment_session_view` decorator that loads the experiment + session on the request and also with the `verify_session_access_cookie` that checks the `chat_session_access` cookie to see if the user has the correct session. Since they will not, they will not be able to use the session_id. The only way you can chat to a session is 
1. Be logged in
2. If you have a proper `chat_session_access` cookie that you only get when your email is verified.
3. Identifier capture is **off**, in which case anonymous participant's will be assigned an identifier, effectively having anonymous and one-off sessions

## User Impact

Participants will have to verify their email address before being allowed to chat to a bot.

### Demo

(Sorry for the mp4 file, I have trouble with loom)

https://github.com/user-attachments/assets/26b75802-fb29-487a-993e-b88ad4d6278c

### Docs

I'll update once this is merged


SmittieC:
@coderabbitai please do a full review with attention to security

apps/experiments/views/experiment.py
```diff
@@ -767,12 +774,13 @@ def experiment_chat_session(request, team_slug: str, experiment_id: int, session
     )
 
 
+@experiment_session_view()
+@verify_session_access_cookie
 @require_POST
 def experiment_session_message(request, team_slug: str, experiment_id: int, session_id: int, version_number: int):
-    working_experiment = get_object_or_404(Experiment, id=experiment_id, team=request.team)
+    working_experiment = request.experiment
     # hack for anonymous user/teams
```
snopoke, line 782: remove old comment

apps/experiments/views/experiment.py
```diff
@@ -822,12 +830,12 @@ def experiment_session_message(request, team_slug: str, experiment_id: int, sess
...
 
 
-# @login_and_team_required
-def get_message_response(request, team_slug: str, experiment_id: int, session_id: int, task_id: str):
-    experiment = get_object_or_404(Experiment, id=experiment_id, team=request.team)
+@experiment_session_view()
+@verify_session_access_cookie
+def get_message_response(request, team_slug: str, experiment_id: str, session_id: str, task_id: str):
+    experiment = request.experiment
     # hack for anonymous user/teams
```
snopoke, line 837: remove old comment

snopoke (approved):
This looks great.

snopoke (approved):

PR #794, opened 2024-10-28
Add default to participant_allowlist

proteusvacuum:
## Description
Small thing I found - when making a new experiment I was getting errors in the javascript console as `{{ form.initial.participant_allowlist}}` resolves to nothing. 

image


SmittieC (approved):
Thanks

PR #808, opened 2024-10-31
Fix: Only try to send an email if the identifier type is email

SmittieC:
## Description

Issue: https://dimagi.sentry.io/share/issue/164b40cb13ce485fb96a1ca409a51e31/
When the identifier type is text, any identifier is accepted, but we'll try to send an email to this identifier. We now only send emails when the identifier type is email, so we know an email is entered.

## User Impact

Participants using non-email identifiers (in consent forms expecting text based identifiers) cannot start a chat


SmittieC:
merging, since this is a bugfix

PR #810, opened 2024-10-31
Track LLM Model Metadata

proteusvacuum:
## Description

This moves the definition of the LLM model into its own Django model, instead of a string defined on the experiment / pipeline / assistant / analysis pipeline.

It allows creating both "global" LLM models (by omitting the team), and "custom" LLM models by associating those with a team.

Currently, this only stores `max_token_limit`, but it opens the door to adding more metadata alongside these models (e.g. pricing, tool usage, etc.)

I've kept the old data around, and in a followup migration will remove the old columns once this has been deployed stably.

## User Impact

Users of "standard" LLM models (e.g. `gpt-4o-mini`) will now be able to select these from a dropdown, rather than have to know and enter this as a string in all the places that it is required (e.g. creating an experiment).

### Demo

https://github.com/user-attachments/assets/64829577-f808-4afb-b00c-14e64763b3f5


apps/service_providers/migrations/0019_auto_20241029_0901.py
```diff
@@ -0,0 +1,222 @@
...
+def _create_default_llm_provider_models(apps, schema_editor):
+    LlmProviderModel = apps.get_model("service_providers", "LlmProviderModel")
+    for default_provider_model in DEFAULT_LLM_PROVIDER_MODELS:
+        LlmProviderModel.objects.create(
+            team=None,
+            type=default_provider_model.type,
+            name=default_provider_model.name,
+            max_token_limit=default_provider_model.max_token_limit,
+        )
+    # TODO: Get list of current global models
```
snopoke, line 31: Maybe from here: https://raw.githubusercontent.com/FlowiseAI/Flowise/main/packages/components/models.json

apps/service_providers/models.py
```diff
@@ -120,14 +128,71 @@ def get_llm_service(self):
...
+    def __str__(self):
+        label = f"{LlmProviderTypes[self.type].label}: {self.name}"
+        if self.is_custom():
+            label = f"{label} (custom for {self.team.name})"
+        return label
+
+    def is_custom(self):
+        return self.team is not None
+
+    def delete(self, *args, **kwargs):
```
snopoke, line 175: Just noting that we said we'd do a similar check when editing to prevent changing the name of a model in use.

Also, we'd need to check more than just experiments right? For Django model references we could do this:
```python
for field in self._meta.get_fields():
    if field.one_to_many and field.auto_created:
        related_objects = getattr(instance, field.get_accessor_name(), None)
        if related_objects and related_objects.exists():
            raise DjangoValidationError(f"The model '{instance.name}' is still referenced by other objects.")
```

We'd have to manually check pipelines unless setup up FK fields.

apps/service_providers/tables.py
```diff
@@ -36,3 +37,25 @@ def url_factory(url_name, request, record, value):
         return reverse(url_name, args=[request.team.slug, provider_type, record.pk])
 
     return url_factory
+
+
+class LlmProviderModelTable(tables.Table):
+    def show_delete(request, record) -> bool:
+        return not record.experiment_set.exists()
```
snopoke, line 44: Also check other references

templates/generic/action_ajax.html
```diff
@@ -21,3 +21,17 @@
...
+        alertify.message('OK');
+      }).setting({
+        title: 'Error',
+        transition: 'fade',
+        onok: function() {},
+      }).show();
+    ;
+
+  });
+</script>
```
snopoke, line 37: maybe we should make this a global handler on the site?

apps/service_providers/migrations/0019_auto_20241029_0901.py
```diff
@@ -0,0 +1,222 @@
...
+    Node = apps.get_model("pipelines", "Node")
+
+    for analysis in Analysis.objects.select_related("llm_provider").all():
+        _handle_analysis(LlmProviderModel, analysis)
+
+    for node in Node.objects.filter(Q(type="LLMResponseWithPrompt") | Q(type="LLMResponse") | Q(type="RouterNode")).all():
+        _handle_pipeline_node(LlmProvider, LlmProviderModel, node)
+
+    for experiment in Experiment.objects.select_related("llm_provider").all():
+        if experiment.llm_provider is None and experiment.assistant:
```
SmittieC, line 198: I think the `experiment.llm_provider is None` check should be dropped. Reason being that we don't remove the llm provider from the experiment instance when we link an assistant, so this check will essentially exclude those assistants (there are 16 cases like this on prod). 0 cases for pipelines.

apps/service_providers/migrations/0019_auto_20241029_0901.py
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/810/commits/ed258a38cf4ca9aa6ffe7313bb3d6ac27b5d6d40

I removed the "vision" ones, but kept all the others.

apps/service_providers/models.py
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/810/commits/b82b3ecd55030ce39fe590b7b8f6c4f2e79faa6d

apps/service_providers/migrations/0019_auto_20241029_0901.py
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/810/commits/b82b3ecd55030ce39fe590b7b8f6c4f2e79faa6d

apps/service_providers/tables.py
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/810/commits/115e62f7d4a7f674e774bae08fe032c963ffe41a

templates/generic/action_ajax.html
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/810/commits/39c9364be6567c802ac736a59000a309fdf5dd32

apps/service_providers/models.py
```diff
@@ -173,12 +175,30 @@ def is_custom(self):
...
-            experiment_info = [f"{id}: {name}" for id, name in experiments]
+        related_object_strings = []
+        for field in self._meta.get_fields():
+            if field.one_to_many and field.auto_created:
+                related_objects = getattr(self, field.get_accessor_name(), None)
+                if related_objects and related_objects.exists():
+                    for obj in related_objects.all():
+                        related_model_name = obj._meta.verbose_name
+                        obj_name = obj.name
+                        related_object_strings.append(f"{related_model_name}: {obj_name}")
```
snopoke, line 199: This seems like a useful utility

apps/service_providers/tests/test_models.py
```diff
@@ -34,17 +73,30 @@ def test_provider_models_for_team_includes_global(self, llm_provider_model):
         assert len(global_models) == len(team_models) - 1
         assert all(not m.is_custom() for m in global_models)
 
+    @pytest.mark.parametrize("fixture_name", ["experiment", "assistant", "analysis"])
+    @django_db_with_data()
+    def test_cannot_delete_provider_models_with_associated_models(self, request, fixture_name):
+        associated_object = request.getfixturevalue(fixture_name)
```
snopoke, line 79: neat

apps/service_providers/migrations/0019_auto_20241029_0901.py
```diff
@@ -12,23 +12,64 @@ def _create_llm_provider_models(apps, schema_editor):
...
+        "gpt-4-0613",
+        "gpt-4-32k",
+        "gpt-4-32k-0613",
+        "gpt-3.5-turbo",
+        "gpt-3.5-turbo-0125",
+        "gpt-3.5-turbo-1106",
+        "gpt-3.5-turbo-0613",
+        "gpt-3.5-turbo-16k",
+        "gpt-3.5-turbo-16k-0613"
+    ]
```
snopoke, line 60: Nice, where did this list come from? Is there more metadata?

proteusvacuum: It's a modified version of this: https://raw.githubusercontent.com/FlowiseAI/Flowise/main/packages/components/models.json

I only included the ones supported in OCS (anthropic, openai, azure openai). The only extra metadata is a `label` field, which is mostly the same as the `name`, except for things like `gpt-4o`, which say `"label": "gpt-4o (latest)"`. 

We could probably add a label field to the model if we wanted, but this is getting pretty large now, so we could do it separately.

proteusvacuum:

apps/service_providers/forms.py
```diff
@@ -216,3 +218,17 @@ class LangsmithTraceProviderForm(ObfuscatingMixin, ProviderTypeConfigForm):
     api_key = forms.CharField(label=_("API Key"))
     api_url = forms.URLField(label=_("API URL"), initial="https://api.smith.langchain.com")
     project = forms.CharField(label=_("Project Name"))
+
+
+class LlmProviderModelForm(forms.ModelForm):
+    class Meta:
+        model = LlmProviderModel
+        fields = ("type", "name", "max_token_limit")
+
```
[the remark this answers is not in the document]
proteusvacuum: The help text is already on the model field

apps/service_providers/forms.py
```diff
@@ -216,3 +218,17 @@ class LangsmithTraceProviderForm(ObfuscatingMixin, ProviderTypeConfigForm):
...
+    def __init__(self, team, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        types = LlmProvider.objects.filter(team=team).values_list("type", flat=True).all()
+        self.fields["type"].choices = [choice for choice in LlmProviderTypes.choices if choice[0] in types]
+        if len(types) == 0:
+            url = reverse("service_providers:new", kwargs={"team_slug": team.slug, "provider_type": "llm"})
+            self.fields["type"].help_text = f"You must create an <a href={url}>LLM provider</a> first"
```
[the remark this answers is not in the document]
proteusvacuum: Please generate the test cases

apps/service_providers/llm_service/state.py
```diff
@@ -61,11 +61,13 @@ def get_llm_service(self):
...
     def get_chat_model(self):
-        return self.get_llm_service().get_chat_model(self.experiment.llm, self.experiment.temperature)
+        return self.get_llm_service().get_chat_model(
+            self.experiment.llm_provider_model.name, self.experiment.temperature
+        )
 
     @property
     def callback_handler(self):
-        return self.get_llm_service().get_callback_handler(self.experiment.llm)
+        return self.get_llm_service().get_callback_handler(self.experiment.llm_provider_model.name)
```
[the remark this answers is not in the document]
proteusvacuum: Correct, I haven't made that other migration yet as it will be handled in a separate PR which will get merged later.

apps/experiments/models.py
```diff
@@ -829,11 +855,6 @@ def format_route(route) -> str:
                     name="input_formatter",
                     raw_value=self.input_formatter,
                 ),
-                VersionField(
-                    group_name="Safety",
-                    name="max_token_limit",
-                    raw_value=self.max_token_limit,
-                ),
```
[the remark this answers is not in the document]
proteusvacuum: Yes, the `max_token_limit` is now part of the `llm_provider_model`.

apps/service_providers/migrations/0020_auto_20241111_1113.py
```diff
@@ -0,0 +1,264 @@
...
+        return
+    except LlmProviderModel.DoesNotExist:
+        pass
+
+    new_custom_provider_model = LlmProviderModel.objects.create(
+        team=experiment.team,
+        type=experiment.llm_provider.type,
+        name=experiment.llm,
+        max_token_limit=experiment.max_token_limit_old,
+    )
```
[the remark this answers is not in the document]
proteusvacuum: Yes, I was going to refactor these, but they are different enough (e.g. `experiment` has the `max_token_limit` check) that I thought it was OK to leave as-is, especially since this code will be run only once. 
What do you think?

apps/service_providers/tables.py
```diff
@@ -36,3 +37,22 @@ def url_factory(url_name, request, record, value):
...
+
+
+class LlmProviderModelTable(tables.Table):
+    actions = actions.ActionsColumn(
+        actions=[
+            actions.edit_action(url_name="service_providers:llm_provider_model_edit"),
+            actions.delete_action(
+                url_name="service_providers:llm_provider_model_delete",
+                confirm_message="Are you sure you want to delete this custom LLM Provider Model?",
+                display_condition=lambda request, record: not record.has_related_objects(),
```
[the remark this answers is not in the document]
snopoke: True, it will be a small list. Maybe fine for now.

PR #823, opened 2024-11-05
Small channel form updates

SmittieC:
An informal request - [link](https://dimagi.slack.com/archives/C05PK63Q89H/p1730803103698439)
For now I just went with defaulting the channel name to the experiment name if the user don't want to bother with it. I want to comb through the codebase some time to see if we're actually using the channel name for anything useful. I suspect not.

## Description

1. Be more explicit as to what number to input when adding a whatsapp number
2. Optional channel name. Default to experiment name

## User Impact

Same description as above

### Demo

image

### Docs

N/A

### DIFF
--- a/apps/channels/forms.py
+++ b/apps/channels/forms.py
@@ -136,7 +136,7 @@ class WhatsappChannelForm(WebhookUrlFormBase):
         label="Number",
         max_length=20,
         help_text=(
-            "This is the number you got from your provider and should be in any of the formats: "
+            "This is the WhatsApp Business Number you got from your provider and should be in any of the formats: "
             "+27812345678, +27-81-234-5678, +27 81 234 5678"
         ),
     )
--- a/apps/channels/forms.py
+++ b/apps/channels/forms.py
@@ -26,12 +26,14 @@ class ChannelForm(forms.ModelForm):
         widgets = {"platform": forms.HiddenInput()}
 
     def __init__(self, *args, **kwargs):
-        team = kwargs.pop("team", None)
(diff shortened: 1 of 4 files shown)


apps/channels/forms.py
```diff
@@ -18,6 +18,8 @@
 
 
 class ChannelForm(forms.ModelForm):
+    name = forms.CharField(required=False, help_text="If you leave this blank, it will default to the experiment name")
```
snopoke, line 21: Maybe add a placeholder: `widget=forms.TextInput(attrs={'placeholder': 'Channel Name'})`

(it would be nice if this was the experiment name)

apps/channels/forms.py
```diff
@@ -131,7 +133,12 @@ def clean_bot_token(self):
 
 class WhatsappChannelForm(WebhookUrlFormBase):
     number = forms.CharField(
-        label="Number", max_length=20, help_text="e.g. +27812345678, +27-81-234-5678, +27 81 234 5678"
+        label="Number",
+        max_length=20,
+        help_text=(
+            "This is the number you got from your provider and should be in any of the formats: "
```
snopoke, line 139:
```suggestion
            "This is the WhatsApp Business Number you got from your provider and should be in any of the formats: "
```

apps/channels/forms.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/823/commits/b62b2c6f0e28124412a10cbbfd4b799e5df1fb91

snopoke (approved):

PR #832, opened 2024-11-06
Delete 'analysis' app: Part I

snopoke:
## Description
This deletes most of the code and data models and adds a migration to delete the tables from the database. 

I have stored a DB backup in S3 so the data in the DB can be removed.


snopoke:
@coderabbitai configuration

PR #833, opened 2024-11-06
Experiment Versioning: Assistants

stephherbers:
## Description

Allow versioning of assistants

## User Impact

Now experiments with assistants can be versioned where a copy of the assistant if made if attached to that experiment

### Demo

not part of this PR

### Docs

Docs updates are coming be ready


apps/assistants/models.py
```diff
@@ -76,6 +87,32 @@ def supports_file_search(self):
...
+        assistant_version.tools = self.tools.copy()
+        for tool_resource in self.tool_resources.all():
+            files = tool_resource.files.all()
+            if files:
+                new_tool_resource = ToolResources.objects.create(
+                    assistant=assistant_version,
+                    tool_type=tool_resource.tool_type,
+                )
+                new_tool_resource.files.set(files)
+                new_tool_resource.extra = tool_resource.extra
```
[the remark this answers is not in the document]
stephherbers: fixed

apps/experiments/tests/test_models.py
```diff
@@ -794,6 +806,13 @@ def _assert_pipeline_is_duplicated(self, original_experiment, new_version):
         for node in original_experiment.pipeline.node_set.all():
             assert new_version.pipeline.node_set.filter(working_version_id=node.id).exists()
 
+    def _assert_assistant_is_duplicated(self, original_experiment, new_version):
+        assert new_version.assistant.working_version == original_experiment.assistant
+        assert new_version.assistant.version_number == 1
+        assert original_experiment.assistant.version_number == 2
+        assert new_version.assistant.tool_resources.filter(tool_type="file_search").first().files.count() == 1
+        assert new_version.assistant.tool_resources.filter(tool_type="code_interpreter").first().files.count() == 1
```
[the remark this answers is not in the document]
stephherbers: tests to resolving

apps/assistants/models.py
```diff
@@ -76,6 +87,33 @@ def supports_file_search(self):
...
+        from .sync import push_assistant_to_openai
+
+        version_number = self.version_number
+        self.version_number = version_number + 1
+        self.save(update_fields=["version_number"])
+        assistant_version = super().create_new_version(save=False, *args, **kwargs)
+        assistant_version.version_number = version_number
+        assistant_version.save()
+
+        assistant_version.files.set(self.files.all())
```
SmittieC, line 101: I think we'd want to duplicate the files rather? People won't be able to edit the file from the other assistant, but they would be able to delete it from there. There is a `duplicate` method on the File model which seems useful here

apps/assistants/models.py
```diff
@@ -76,6 +87,33 @@ def supports_file_search(self):
...
+            if files:
+                new_tool_resource = ToolResources.objects.create(
+                    assistant=assistant_version,
+                    tool_type=tool_resource.tool_type,
+                )
+                new_tool_resource.files.set(files)
+                new_tool_resource.extra = tool_resource.extra
+                new_tool_resource.save()
+
+        push_assistant_to_openai(assistant_version, internal_tools=get_assistant_tools(assistant_version))
```
SmittieC, line 114: The `assistant_id` doesn't seem to be cleared, in which case the versions will point to the working version at OpenAI. We'll need to clear the `assistant_id` just before calling `push_assistant_to_openai`. Reason being that the `assistant_id` field cannot be empty, so you wouldn't be able to clear it and save the assistant.

apps/assistants/views.py
```diff
@@ -65,7 +65,7 @@ class OpenAiAssistantTableView(SingleTableView, PermissionRequiredMixin):
     permission_required = "assistants.view_openaiassistant"
 
     def get_queryset(self):
-        return OpenAiAssistant.objects.filter(team=self.request.team).order_by("name")
+        return OpenAiAssistant.objects.filter(team=self.request.team, is_archived=False).order_by("name")
```
SmittieC, line 68: You can use the [VersionsObjectManagerMixin](https://github.com/dimagi/open-chat-studio/blob/2030037d5e0ced3aafa06f3afe96fec18dec6c49/apps/experiments/models.py#L35C7-L35C33) for the `OpenAiAssistantManager` instead of doing this. This will update the default queryset to automatically exclude `is_archived=True` versions

SmittieC: I apologize, a readme of sorts would probably have helped. It would be useful to know what we have for versioning models and what to keep in mind. I'll add one

SmittieC (commented):
Looking good. One last thing: We should probably also filter assistant versions from the assistants table, similar to what we do for instance for the consent forms [here](https://github.com/dimagi/open-chat-studio/blob/2030037d5e0ced3aafa06f3afe96fec18dec6c49/apps/experiments/views/consent.py#L35)

apps/assistants/views.py
stephherbers: https://github.com/dimagi/open-chat-studio/pull/833/commits/382cca744d40f19cefe2489929c4fa89b153ce5a

apps/assistants/tables.py
```diff
@@ -41,9 +43,35 @@ class OpenAiAssistantTable(tables.Table):
...
+        if self.name_linkify(record):
+            return mark_safe(f'<a href="{record.get_absolute_url()}">{record.name}</a>')
+        else:
+            return record.name
+
     class Meta:
         model = OpenAiAssistant
         fields = ("name", "assistant_id")
-        row_attrs = settings.DJANGO_TABLES2_ROW_ATTRS
+        row_attrs = {
```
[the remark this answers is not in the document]
stephherbers: https://github.com/dimagi/open-chat-studio/pull/833/commits/3b0cb7a6ab8ddd64849a03d6eb565ed731117543

apps/assistants/models.py
stephherbers: https://github.com/dimagi/open-chat-studio/pull/833/commits/a21c1fd2094048a45d5e0cab589aba481e2eb226

SmittieC:
> > We should probably also filter assistant versions from the assistants table, similar to what we do for instance for the consent forms [here](https://github.com/dimagi/open-chat-studio/blob/2030037d5e0ced3aafa06f3afe96fec18dec6c49/apps/experiments/views/consent.py#L35)
> > @SmittieC I opted to show the versioned assistants and modified the logic such that it does not link to the edit view & cannot sync. I opted for this just because it will also exist in OpenAI, so because of that felt wrong to hide from view in OCS (when for other things that lived only in OCS it feels more natural to hide things that are simply duplicates in time). But let me know if you disagree

I see. My only concern with this is that the assistants table will get cluttered, but let's see what the user feedback is. Maybe they like the fact that one can see the versions. Fine by me to leave as-is for now

SmittieC:
I fixed some issues that came from the merge with the main branch and also a small issue when we try to duplicate files that doesn't really exist. This happens when we create a file that was uploaded through OpenAI, but since we can't pull its contents, we don't have a file associated with the record. See [this](https://github.com/dimagi/open-chat-studio/pull/833/commits/6471e33f28635852514c4576efc7af47a2eb5955) commit. I hope you don't mind me making these changes.

All else looks good from my side, thanks @stephherbers. @snopoke, do you have any reservations on the last commit I made?

apps/assistants/tables.py
```diff
@@ -1,29 +1,39 @@
...
-        },
+        linkify=False,
         orderable=True,
     )
     actions = actions.ActionsColumn(
         actions=[
             actions.edit_action(
                 "assistants:edit",
                 required_permissions=["assistants.change_openaiassistant"],
+                display_condition=lambda request, record: record.working_version is None,
```
snopoke, line 29: avoid querying the working version:

```suggestion
                display_condition=lambda request, record: record.working_version_id is None,
```

apps/assistants/tables.py
```diff
@@ -1,29 +1,39 @@
...
             ),
             actions.AjaxAction(
                 "assistants:sync",
                 title="Update from OpenAI",
                 icon_class="fa-solid fa-rotate",
                 required_permissions=["assistants.change_openaiassistant"],
+                display_condition=lambda request, record: record.working_version is None,
```
snopoke, line 36:
```suggestion
                display_condition=lambda request, record: record.working_version_id is None,
```

apps/assistants/tables.py
```diff
@@ -1,29 +1,39 @@
...
+from django.utils.safestring import mark_safe
 from django_tables2 import columns, tables
 
 from apps.assistants.models import OpenAiAssistant
 from apps.generics import actions
 
 
+def get_assistant_row_attrs():
+    def _get_redirect_url(record):
+        return record.get_absolute_url() if record.working_version is None else ""
```
snopoke, line 11:
```suggestion
        return record.get_absolute_url() if record.working_version_id is None else ""
```

apps/assistants/tables.py
```diff
@@ -41,9 +51,15 @@ class OpenAiAssistantTable(tables.Table):
         ]
     )
 
+    def render_name(self, record):
+        if record.working_version is None:
```
snopoke, line 55:
```suggestion
        if record.working_version_id is None:
```

apps/assistants/views.py
```diff
@@ -176,6 +176,9 @@ class DeleteOpenAiAssistant(LoginAndTeamRequiredMixin, View, PermissionRequiredM
     @transaction.atomic()
     def delete(self, request, team_slug: str, pk: int):
         assistant = get_object_or_404(OpenAiAssistant, team=request.team, pk=pk)
+        if assistant.is_a_version and not assistant.is_archived:
```
snopoke, line 179:
```suggestion
        if assistant.working_version_id is None and not assistant.is_archived:
```

snopoke (commented):
I'm going to test this locally

SmittieC (approved):

PR #837, opened 2024-11-07
Versioning for custom action operations

snopoke:
## Description
Support versioning of custom actions (operations) linked to experiments. 

Quote from [README.md](https://github.com/dimagi/open-chat-studio/blob/99571a1dfcc867bff0fb35cb90b322c394d9e086/apps/custom_actions/README.md)

The `CustomAction` model is not versioned. However, the `CustomActionOperation` model is versioned since this is the
model that is linked to the Experiment.

**Working Version**

When a custom action or operation is first created, it is in a "working version" state. In this
state, the schema is not persisted to the database but is computed dynamically when needed.
This means that changes to the `CustomAction` schema will be reflected in the working version of the Experiment.

**Persisted Version**

When a new version of an experiment is created, the custom action operations are versioned. The
schema for each operation is then persisted to the database, ensuring that the exact state of the operation at the
time of versioning is preserved regardless of changes to the `CustomAction`. It also ensures that the operation is still
valid when the version is created.

### Demo
<div>
    <a href="https://www.loom.com/share/cbd7c586c8114a758e2a5921cf2b32b5">
      <p>Experiment Versioning Demo with Custom Actions 👩‍🔬 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/cbd7c586c8114a758e2a5921cf2b32b5">
      
    </a>
  </div>


snopoke:
@coderabbit review

snopoke:
@coderabbit review

snopoke:
@coderabbit resolve

SmittieC (dismissed):
Nice. Looks good to me. 1 non-blocking question. Is the only reason we're not saving the `_operation_schema` for working version `CustomActionOperation` records the fact that if the associated `CustomAction` changes, we'd have to update this?

apps/custom_actions/models.py
```diff
@@ -70,7 +74,18 @@ def get_operations_by_id(self):
         return {op.operation_id: op for op in self.operations}
 
 
-class CustomActionOperation(models.Model):
+class CustomActionOperationManager(VersionsObjectManagerMixin, models.Manager):
+    pass
+
+
+class CustomActionOperation(models.Model, VersionsMixin, VersioningMixin):
```
SmittieC, line 81: Maybe `VersioningMixin` should be collapsed onto `BaseModel`? There's only 1 method in there

snopoke: That wouldn't help here since this model doesn't extend `BaseModel` but maybe it should?

snopoke:
> Is the only reason we're not saving the `_operation_schema` for working version `CustomActionOperation` records the fact that if the associated `CustomAction` changes, we'd have to update this?

Yup, that's correct.

apps/custom_actions/models.py
SmittieC: I don't see why not? My philosohy is to rather have those created_at / updated_at timestamps from BaseModel than not.

snopoke:
putting this back in draft mode so that I can address the assistant versioning changes once those are merged

apps/utils/models.py
```diff
@@ -1,8 +1,10 @@
+from typing import Self
+
 from django.db import models
 
 
 class VersioningMixin:
-    def compare_with_model(self, new: models.Model, exclude_fields: list[str]) -> set:
+    def compare_with_model(self, new: Self, exclude_fields: list[str]) -> set[str]:
```
SmittieC, line 7: `Self` as a type. The more you know

SmittieC (approved):
Nice

PR #835, opened 2024-11-07
Expose open AI assistant and langfuse trace info

snopoke:
## Description
This PR includes a few changes:

1. Display OpenAI Assistant ID and Thread ID in the session view when available

Peek 2024-11-07 12-03

2. Store the Langfuse trace ID and URL in the chat message metadata
3. Display a link to the Langfuse trace in the session view

image

4. Display a link to the Langfuse trace in the chat view when debug mode is on

image
## User Impact

5. Fix a JS error I noticed in the chat UI when assistants aren't enabled

PR #843, opened 2024-11-08
revert revert trace details

snopoke:
## Description
Redo of https://github.com/dimagi/open-chat-studio/pull/835

The issue that caused [DIMAGI-BOTS-HA](https://dimagi.sentry.io/issues/6051321440/) is that `functools.cache` creates a global cache which resulted in the same 'trace service' being used in successive calls. This PR adds a new commit which changes the cached function to a cached property which is local to the class instance.

For more info on the topic see https://rednafi.com/python/lru_cache_on_methods/


snopoke:
@SmittieC this is now actually ready for review

PR #845, opened 2024-11-11
Fix: Loop when vector store has many files

SmittieC:
## Description

- We did not consider the `has_more=True` case in the vector store result when checking if the assistants are in sync
- The default pagination result size is 20. The maximum we can set is 100
- We now set it to 100 and loop when there are even more.

## User Impact

The message saying that there are files at OpenAI but not in OCS should now show correctly.

### Demo

N/A

### Docs

N/A

### DIFF
--- a/apps/assistants/sync.py
+++ b/apps/assistants/sync.py
@@ -190,24 +190,47 @@ def are_files_in_sync_with_openai(assistant: OpenAiAssistant) -> bool:
 
     # ensure files match
     for resource in tool_resources:
-        openai_file_ids = []
-        if resource.tool_type == "code_interpreter":
-            code_interpreter = getattr(assistant_data.tool_resources, "code_interpreter", None)
-            if code_interpreter is not None and code_interpreter.file_ids:
-                openai_file_ids = code_interpreter.file_ids
-            else:
-                openai_file_ids = []
-        elif resource.tool_type == "file_search":
-            openai_vector_store_id = resource.extra.get("vector_store_id")
-            if openai_vector_store_id:
-                vector_store_data = client.beta.vector_stores.files.list(vector_store_id=openai_vector_store_id)
-                openai_file_ids = [file.id for file in getattr(vector_store_data, "data", [])]
-            else:
-                openai_file_ids = []
+        openai_file_ids = _get_tool_file_ids_from_openai(client, assistant_data, resource)
         ocs_file_ids = [file.external_id for file in resource.files.all() if file.external_id]
         if set(ocs_file_ids) != set(openai_file_ids):
             return False
     return True
+
+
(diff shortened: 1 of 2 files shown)


SmittieC:
@coderabbitai Do do you see any issues with this change?

SmittieC:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> @SmittieC, I reviewed the changes and have a few observations:
> 
> * **Loop Termination:** Ensure that the loop handling the `has_more=True` case has a proper termination condition. This will prevent potential infinite loops if the vector store API behaves unexpectedly.
> * **Pagination Limit:** Setting the pagination size to the maximum of 100 is acceptable. However, consider the performance impact if there are significantly more files. It might be beneficial to implement asynchronous calls or batching mechanisms for better efficiency.
> * **Error Handling:** Make sure to handle any possible exceptions or errors from the vector store API, especially during pagination, to enhance the robustness of the application.
> * **Unit Tests:** It would be helpful to add unit tests covering scenarios with more than 100 files. This ensures that the new looping logic works correctly and helps prevent regressions in the future.
> 
> Overall, the changes seem to address the issue effectively. With these considerations, the solution will be more robust and maintainable.

- Loop termination is there
- The max I can set the limit is 100.
- I am simulating pagination in the test by returning 1 file on the first call, and another on the second. The `has_more=True` in the mocked response simulates this

SmittieC:
@coderabbitai resolve all

apps/assistants/sync.py
```diff
@@ -190,26 +190,49 @@ def are_files_in_sync_with_openai(assistant: OpenAiAssistant) -> bool:
...
+        if openai_vector_store_id:
+            kwargs = {}
+            while True:
+                vector_store_data = client.beta.vector_stores.files.list(
+                    vector_store_id=openai_vector_store_id, limit=100, **kwargs
+                )
+                openai_file_ids.extend([file.id for file in getattr(vector_store_data, "data", [])])
+                if not vector_store_data.has_more:
+                    break
+                kwargs["after"] = vector_store_data.last_id
```
snopoke, line 232: manual pagination shouldn't be necessary since the openai python library provides auto-paginating iterators: https://github.com/openai/openai-python?tab=readme-ov-file#pagination

I'm not sure why we are doing `getattr(vector_store_data, "data", [])` we should just be able to iterate over the response like we do here: https://github.com/dimagi/open-chat-studio/blob/860fc789a16484b03b2293c844c5a21ac6a742be/apps/assistants/sync.py#L275-L280

SmittieC: https://github.com/dimagi/open-chat-studio/pull/846

PR #848, opened 2024-11-12
Prompt Builder Fixes

proteusvacuum:
## Description

I missed a few places that the previous "model" was used:
https://github.com/dimagi/open-chat-studio/pull/810 

## User Impact

This was causing issues with accessing the prompt builder in a new team, and running a prompt in the prompt builder.

### DIFF
--- a/apps/experiments/tasks.py
+++ b/apps/experiments/tasks.py
@@ -9,7 +9,7 @@ from apps.channels.datamodels import Attachment, BaseMessage
 from apps.chat.bots import create_conversation
 from apps.chat.channels import WebChannel
 from apps.experiments.models import Experiment, ExperimentSession, PromptBuilderHistory, SourceMaterial
-from apps.service_providers.models import LlmProvider
+from apps.service_providers.models import LlmProvider, LlmProviderModel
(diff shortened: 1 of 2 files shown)


apps/experiments/tasks.py
```diff
@@ -49,6 +49,7 @@ def get_response_for_webchat_task(
 @shared_task
 def get_prompt_builder_response_task(team_id: int, user_id, data_dict: dict) -> dict[str, str | int]:
     llm_service = LlmProvider.objects.get(id=data_dict["provider"]).get_llm_service()
+    llm_provider_model = LlmProviderModel.objects.get(id=data_dict["model"])
```
bderenzi, line 52: Can do it in a future PR, but we should update the frontend to send `model_id` instead of `model` if this is what it really is.

bderenzi (approved):

SmittieC (approved):

apps/experiments/tasks.py
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/849

PR #849, opened 2024-11-12
Fix saving prompts and fetching history

proteusvacuum:
## Description

A few other issues I found in the prompt builder: saving prompts and restoring prompts from history.
I think restoring prompts from history was broken before, as the provider was not being included.

## User Impact

Users can save prompts and load history state again.


SmittieC (approved):
Thanks. Looks fine to me

snopoke (approved):

PR #850, opened 2024-11-12
use llm provider model from assistant

snopoke:
Fix for [DIMAGI-BOTS-HM](https://dimagi.sentry.io/issues/6059206363/)


proteusvacuum (approved):

SmittieC (approved):

proteusvacuum:
Ah! I added that method right before merging - I should have done better to go and change all the call sites after that. Thanks!

PR #854, opened 2024-11-12
Remove references to old llm_model, llm_models and max_token_limit

proteusvacuum:
These are no longer used after the migration to LlmProviderModel. 

I would probably wait a bit to merge / deploy this in case there is data we want in these columns.


proteusvacuum:
FYI: Base branch is https://github.com/dimagi/open-chat-studio/tree/fr/more-llm-provider-fixes as changes in that branch are needed for tests to pass.

SmittieC (approved):
Looks fine to me

PR #855, opened 2024-11-12
Resolve circular import error

stephherbers:
## Description

Erroring during tests and running locally of a circular import error with OpenAiAssistant
https://docs.python.org/3/library/typing.html#typing.Union

## User Impact

None

### Demo

N/A

### Docs

N/A

### DIFF
--- a/apps/chat/agent/tools.py
+++ b/apps/chat/agent/tools.py
@@ -1,18 +1,20 @@
 import logging
 from datetime import datetime, timedelta
-from typing import Any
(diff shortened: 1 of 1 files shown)


SmittieC:
I don't see this happening on main? I did see this happen on the assistants versioning branch. I fixed it on that branch [here](https://github.com/dimagi/open-chat-studio/pull/833/commits/0f2c730857a452285a907310740dac6690f529fe). Hope you don't mind.

PR #856, opened 2024-11-12
Make new version when creating experiment

stephherbers:
## Description

When creating a new experiment, create a v1

at the create view for a new experiment a version and working version is created

## User Impact

now an experiment will always have a version and a working version

### Demo

N/A

### Docs

docs coming soon


snopoke (commented):
I'm not a fan of putting this kind of logic in the save function since it get's called in unnecessary places (like tests etc).

I'd prefer just doing it from view.

SmittieC:
Thinking about this now, do we have to create a new version when we create the experiment. We want to be clear about what version is deployed and which one we're working on at any point in time if I remember the context for this correctly?

Since the working version already has its version set to 1, I'm thinking if we show the working version in the versions table and indicate at the top that the current deployed version is 1, which it is, then we don't have to do much else? The only quirk I can think of now would be that we'd either have to 1. show this version is in "draft" the whole time, or 2. don't show anything until version 2 since we wouldn't have a previous version to compare this one to. Any changes to this version would reflect in "production". I'm in favor of option 1 in this scenario.

Thoughts?

snopoke:
@SmittieC I think the issue is that it's inconsistent with the state once you have a release. As long as there are not releases all channels will be using the 'working version', but as soon as you create a release it becomes the default and now all channels are talking to that version.

I do think it makes sense to create the first release initially.

SmittieC:
> @SmittieC I think the issue is that it's inconsistent with the state once you have a release. As long as there are not releases all channels will be using the 'working version', but as soon as you create a release it becomes the default and now all channels are talking to that version.
> 
> I do think it makes sense to create the first release initially.

Allow me to push back one last time. To be clear, I'm not against us creating a version on create, I'm just exploring alternatives a bit. Last one:

How about we disable channels for the working version instead? So when you create an experiment, you can only chat to it through the web UI. Once you're ready to make it public, you can create a new version. Maybe we can have a button saying "deploy now" as well? I don't know. Anycase, when you created your first version, you'll see that version in the versions table with the "deployed" marker against it, indicating that the channels are live and pointing to that version.

snopoke:
> > @SmittieC I think the issue is that it's inconsistent with the state once you have a release. As long as there are not releases all channels will be using the 'working version', but as soon as you create a release it becomes the default and now all channels are talking to that version.
> > I do think it makes sense to create the first release initially.
> 
> Allow me to push back one last time. To be clear, I'm not against us creating a version on create, I'm just exploring alternatives a bit. Last one:
> 
> How about we disable channels for the working version instead? So when you create an experiment, you can only chat to it through the web UI. Once you're ready to make it public, you can create a new version. Maybe we can have a button saying "deploy now" as well? I don't know. Anycase, when you created your first version, you'll see that version in the versions table with the "deployed" marker against it, indicating that the channels are live and pointing to that version.

That would work but it adds more complexity to the process of getting started. @bderenzi what do you think?

SmittieC (approved):
No blockers from my side

snopoke (approved):

PR #857, opened 2024-11-13
Node input validation and help text

SmittieC:
## Description

1. Added help text to some fields. These are populated from the field descriptions.
2. Input validation for nodes. Validation happens on the BE and is propegated to the UI on save or load

## User Impact

Users will see when their input is wrong, but will not be prevented from saving the pipeline.

### Demo

demo

### Docs

N/A

### DIFF
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -104,7 +104,7 @@ class Pipeline(BaseTeamModel, VersionsMixin):
         for node in self.node_set.all():
             node_class = getattr(pipeline_nodes, node.type)
             try:
-                node_class(**node.params)
+                node_class.model_validate(node.params)
             except pydantic.ValidationError as e:
                 errors[node.flow_id] = {err["loc"][0]: err["msg"] for err in e.errors()}
         return errors
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -102,9 +102,16 @@ class Pipeline(BaseTeamModel, VersionsMixin):
         flow = Flow(**self.data)
         flow_nodes_by_id = {node.id: node for node in flow.nodes}
         nodes = []
+        errors = {}
+
         for node in self.node_set.all():
             node_class = getattr(pipeline_nodes, node.type)
             input_types = get_input_types_for_node(node_class)
+            try:
+                node_class(**node.params)
+            except pydantic.ValidationError as e:
+                errors[node.flow_id] = {err["loc"][0]: err["msg"] for err in e.errors()}
+
             nodes.append(
                 FlowNode(
                     id=node.flow_id,
@@ -119,6 +126,7 @@ class Pipeline(BaseTeamModel, VersionsMixin):
                 )
             )
         flow.nodes = nodes
+        flow.errors = errors
         return flow.model_dump()
 
     @cached_property
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -95,6 +95,20 @@ class Pipeline(BaseTeamModel, VersionsMixin):
                 },
             )
 
+    def validate(self) -> dict:
+        """Validate the pipeline nodes and return a dictionary of errors"""
+        from apps.pipelines.nodes import nodes as pipeline_nodes
+
+        errors = {}
+
+        for node in self.node_set.all():
+            node_class = getattr(pipeline_nodes, node.type)
+            try:
+                node_class(**node.params)
+            except pydantic.ValidationError as e:
+                errors[node.flow_id] = {err["loc"][0]: err["msg"] for err in e.errors()}
+        return errors
+
     @cached_property
     def flow_data(self) -> dict:
         from apps.pipelines.nodes import nodes as pipeline_nodes
@@ -102,15 +116,10 @@ class Pipeline(BaseTeamModel, VersionsMixin):
         flow = Flow(**self.data)
         flow_nodes_by_id = {node.id: node for node in flow.nodes}
         nodes = []
-        errors = {}
 
         for node in self.node_set.all():
             node_class = getattr(pipeline_nodes, node.type)
             input_types = get_input_types_for_node(node_class)
-            try:
-                node_class(**node.params)
-            except pydantic.ValidationError as e:
-                errors[node.flow_id] = {err["loc"][0]: err["msg"] for err in e.errors()}
 
             nodes.append(
                 FlowNode(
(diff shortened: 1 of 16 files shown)


apps/pipelines/models.py
```diff
@@ -95,16 +95,32 @@ def set_nodes(self, nodes: list[FlowNode]) -> None:
...
+    def validate(self) -> dict:
+        """Validate the pipeline nodes and return a dictionary of errors"""
+        from apps.pipelines.nodes import nodes as pipeline_nodes
+
+        errors = {}
+
+        for node in self.node_set.all():
+            node_class = getattr(pipeline_nodes, node.type)
+            try:
+                node_class(**node.params)
```
snopoke, line 107: nit: use 'model_validate' to make this clearer
```suggestion
                node_class.model_validate(node.params)
```

apps/pipelines/nodes/nodes.py
```diff
@@ -201,9 +207,18 @@ def _get_current_datetime(self, session):
...
-    recipient_list: str
+    recipient_list: str = Field(description="A comma-separated list of email addresses")
     subject: str
 
+    @field_validator("recipient_list", mode="before")
+    def recipient_list_has_valid_emails(cls, value):
+        email_pattern = re.compile(r"^[\w\.-]+@[\w\.-]+\.\w+$")
+        emails = [email.strip() for email in value.split(",")]
+        for email in emails:
+            if not email_pattern.match(email):
```
snopoke, line 218: Django already has an email validator: `django.core.validators.validate_email`

apps/pipelines/nodes/nodes.py
```diff
@@ -394,16 +411,36 @@ def _create_model_from_data(value_data, model_name: str):
...
+        return value
+
+
+class ExtractStructuredData(ExtractStructuredDataNodeMixin, LLMResponse, StructuredDataSchemaValidatorMixin):
     __human_name__ = "Extract Structured Data"
     __node_description__ = "Extract structured data from the input"
-    data_schema: ExpandableText
+    data_schema: ExpandableText = Field(
+        default='{"name": "the name of the user"}',
+        description="A key-value pair where the key is the name of the field and the value the description",
```
snopoke, line 433: This doesn't have to be just key-value pair:
```suggestion
        description="A JSON data structure with descriptions for each object key.",
```

apps/pipelines/nodes/nodes.py
```diff
@@ -394,16 +411,36 @@ def _create_model_from_data(value_data, model_name: str):
...
 
 
-class ExtractParticipantData(ExtractStructuredDataNodeMixin, LLMResponse):
+class ExtractParticipantData(ExtractStructuredDataNodeMixin, LLMResponse, StructuredDataSchemaValidatorMixin):
     __human_name__ = "Extract Participant Data"
     __node_description__ = "Extract structured data and saves it as participant data"
-    data_schema: ExpandableText
+    data_schema: ExpandableText = Field(
+        default='{"name": "the name of the user"}',
+        description="A key-value pair where the key is the name of the field and the value the description",
```
snopoke, line 442:
```suggestion
        description="A JSON data structure with descriptions for each object key.",
```

assets/javascript/apps/pipeline/stores/pipelineManagerStore.ts
```diff
@@ -12,11 +12,14 @@ const usePipelineManagerStore = create<PipelineManagerStoreType>((set, get) => (
   dirty: false,
   isSaving: false,
   isLoading: true,
+  errors: {},
   loadPipeline: async (pipelineId: number) => {
     set({isLoading: true});
     apiClient.getPipeline(pipelineId).then((pipeline) => {
       if (pipeline) {
         set({currentPipeline: pipeline, currentPipelineId: pipelineId});
+        // TODO: Remove this below and read erros from the ReactFlowJsonObject
```
snopoke, line 21: TODO?

assets/javascript/apps/pipeline/widgets.tsx
```diff
@@ -317,13 +323,17 @@ export function HistoryTypeWidget({
...
-export function InputField({label, children}: React.PropsWithChildren<{ label: string | ReactNode }>) {
+export function InputField({label, help_text, inputError, children}: React.PropsWithChildren<{ label: string | ReactNode, help_text: string, inputError?: string | undefined }>) {
   return (
     <>
       <div className="form-control w-full capitalize">
         <label className="label font-bold">{label}</label>
         {children}
       </div>
+      <div className="flex flex-col">
+        {<small className="text-red-500">{inputError}</small>}
```
snopoke, line 334: Why the extra `{}`?

snopoke (commented):
Small comments but otherwise looks good

apps/pipelines/nodes/nodes.py
SmittieC: Nice, here it is: https://github.com/dimagi/open-chat-studio/pull/857/commits/a8631e6f0c0c070ddfa018e2b0c58bb6e1ee660a

assets/javascript/apps/pipeline/stores/pipelineManagerStore.ts
SmittieC: Oh yeah that is not relevant anymore. I'm keeping the errors on this store level. I'll remove this

assets/javascript/apps/pipeline/widgets.tsx
SmittieC: Good question. Not sure. I'll remove

apps/pipelines/nodes/nodes.py
SmittieC: Technically it does have to be a dict-like object. We don't support lists at the top-level currently. I can't remember why exactly, but thinking about it, what would that actually mean and what schema do we give the LLM?

snopoke: That's fine, maybe we can say a `JSON object structure`?

snopoke (approved):

apps/pipelines/nodes/nodes.py
SmittieC: Sure, I'll update. I see you've approve, so Ima merge once I updated. Thanks!

SmittieC: https://github.com/dimagi/open-chat-studio/pull/857/commits/e828ec811ea4cded658edecabb4c1480417af5a9

PR #860, opened 2024-11-13
LLM Provider updates

snopoke:
## Description

* Show the list of models on the provider create / edit page
* Allow the user to add / remove custom models from that page

## User Impact
* Default models are displayed to the user on the provider page
* Users do not have to navigate to a separate page to view custom models

### Demo
<div>
    <a href="https://www.loom.com/share/a135a1a688b849d28aeafec80bc5397c">
      <p>OCS custom model updates - 13 November 2024 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/a135a1a688b849d28aeafec80bc5397c">
      
    </a>
  </div>


apps/service_providers/views.py
```diff
@@ -230,11 +147,15 @@ def _get_models_by_type(queryset):
...
 @permission_required("service_providers.add_llmprovidermodel")
 def create_llm_provider_model(request, team_slug: str):
-    form = LlmProviderModelForm2(request.POST)
+    form = LlmProviderModelForm(request.team, request.POST)
     if form.is_valid():
         model = form.save(commit=False)
         model.team = request.team
         model.save()
+    else:
+        if len(form.errors) == 1:
```
proteusvacuum, line 156: should this be `> 1`?

proteusvacuum (approved):
LGTM!
How often do users have more than one provider of the same type? 
I think I would find it confusing to add models for one provider but then have them show up elsewhere, although I do see why you moved this here. 
Also 👍🏽 to removing the edit view for these models.

apps/service_providers/views.py
snopoke: No, this is a dict of 'field' -> list[error]. But I do only want to do this if the single field is `__all__` otherwise you won't know which field the error relates to

snopoke:
> LGTM! How often do users have more than one provider of the same type? I think I would find it confusing to add models for one provider but then have them show up elsewhere, although I do see why you moved this here. Also 👍🏽 to removing the edit view for these models.

Not many have more than one. Currently there are 3 teams that have 2 openai providers (not counting grok or others)

PR #864, opened 2024-11-14
Refactor Assistant Runnable + state

SmittieC:
## Description

This is a prefactor for the assistant node. I'm planning on calling the `AssistantRunnable` from the node, but with a new  state that reads the config from the node/assistant instead of the experiment. There's still a lot we can refactor in the state/runnables files, but I'm going to only change that which is needed atm

## User Impact

None

### Demo

N/A

### Docs

N/A


snopoke (commented):
I didn't review this very closely but generally LGTM

PR #865, opened 2024-11-14
LLM Model updates

snopoke:
## Description
* Set the correct token limits for default models
* Convert some default models to custom models
  * I think default models should generally not reference specific model instances eg. `o1-preview-2024-09-12` since these models get replaced quite regularly.
* Remove some default models that aren't being used
* Add groq and perplexity

I have run the migration on a copy of the prod DB successfully.


apps/utils/deletion.py
```diff
@@ -133,3 +134,40 @@ def _get_m2m_related_models(model):
             through_model = field.through if hasattr(field, "through") else field.remote_field.through
             m2m_models[through_model] = field.related_model
     return m2m_models
+
+
+def get_related_objects(instance, pipeline_param_key: str = None) -> list:
```
SmittieC, line 139: Just wondering if this is the best file to put these utils?

SmittieC (approved):
Nice

apps/utils/deletion.py
snopoke: I'm open to moving them - got a suggestion?

SmittieC: Hmm, maybe the file should be renamed instead? `model_relations`?  (let's see if this works: @coderabbit can you suggest a suitable name?)

SmittieC: `relations.py` seems suitable to me. I realize I'm being nit picky here. @snopoke up to you

PR #869, opened 2024-11-15
Assistant Node

SmittieC:
## Description

- Added an assistant node
- File uploads works as before
- Added validation for the router node

Versioning assistant nodes PR: https://github.com/dimagi/open-chat-studio/pull/875

## User Impact

- Users will be able to run assistants in the pipleline
- File uploads should still work

### Demo

My video making tools decided to take an early weekend, so now I can't include a nice demo. Here's a pic though

image

### Docs

N/A for now

### Question
Configuring a trace provider: Should that still happen through the experiment UI? It would be nice to have separate inputs to the node. Potential inputs can include
- Trace provider
- Tools
- Voice provider


SmittieC:
@coderabbitai full review pretty please

apps/chat/bots.py
```diff
@@ -289,8 +289,10 @@ def __init__(self, session: ExperimentSession, experiment: Experiment):
         self.ai_message_id = None
 
     def process_input(self, user_input: str, save_input_to_history=True, attachments: list["Attachment"] | None = None):
+        serializable_attachments = [attachment.model_dump() for attachment in attachments]
```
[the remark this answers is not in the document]
SmittieC: Thanks I'll address

apps/experiments/models.py
```diff
@@ -974,6 +974,28 @@ def format_custom_action_operation(op) -> str:
...
+        from apps.assistants.models import OpenAiAssistant
+        from apps.pipelines.models import Node
+        from apps.pipelines.nodes.nodes import AssistantNode
+
+        if self.pipeline:
+            node_name = AssistantNode.model_json_schema()["title"]
+            # TODO: What about multiple assistant nodes?
+            assistant_node = Node.objects.filter(type=node_name, pipeline=self.pipeline).first()
+            if assistant_node and assistant_node.params["assistant_id"]:
+                return OpenAiAssistant.objects.get(id=assistant_node.params["assistant_id"])
```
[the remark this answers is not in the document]
SmittieC: I think fine to keep it as-is. The node is doing validation, so if it fails here, there's problems that we should know about (which we will through sentry)

apps/service_providers/llm_service/state.py
```diff
@@ -337,3 +313,68 @@ def citations_enabled(self):
...
+        return self.get_llm_service().get_chat_model(self._get_provider_model_name(), self.assistant.temperature)
+
+    @property
+    def callback_handler(self):
+        return self.get_llm_service().get_callback_handler(self._get_provider_model_name())
+
+    def format_input(self, input: str) -> str:
+        if self.input_formatter:
+            input = self.input_formatter.format(input=input)
+        return input
```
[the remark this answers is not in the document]
SmittieC: Thanks, I'll address this

apps/pipelines/nodes/nodes.py
```diff
@@ -233,7 +243,7 @@ class Passthrough(PipelineNode):
 
     def _process(self, input, state: PipelineState, node_id: str) -> str:
```
proteusvacuum, line 244: nit: you should change the output type to `PipelineState` here.

apps/experiments/models.py
```diff
@@ -974,6 +974,28 @@ def format_custom_action_operation(op) -> str:
...
+        """
+        from apps.assistants.models import OpenAiAssistant
+        from apps.pipelines.models import Node
+        from apps.pipelines.nodes.nodes import AssistantNode
+
+        if self.pipeline:
+            node_name = AssistantNode.model_json_schema()["title"]
+            # TODO: What about multiple assistant nodes?
+            assistant_node = Node.objects.filter(type=node_name, pipeline=self.pipeline).first()
+            if assistant_node and assistant_node.params.get("assistant_id"):
```
proteusvacuum, line 995: nit: I think you could add this in the filter:

```
Node.objects.filter(..., params__assistant_id__isnull=False).values_list("params__assistant_id", flat=True).first()
```

apps/experiments/models.py
```diff
@@ -974,6 +974,28 @@ def format_custom_action_operation(op) -> str:
...
+        - If no assistant node is found or if the pipeline is not set, it returns the default assistant associated with
+        the instance.
+        """
+        from apps.assistants.models import OpenAiAssistant
+        from apps.pipelines.models import Node
+        from apps.pipelines.nodes.nodes import AssistantNode
+
+        if self.pipeline:
+            node_name = AssistantNode.model_json_schema()["title"]
+            # TODO: What about multiple assistant nodes?
```
proteusvacuum, line 993: Is this method only used for versioning? For a pipeline experiment, could you instead show the version of the pipeline? I guess if you changed the assistants definitions outside of the pipeline, you'd want to show that. Alternatively, you could return a list here to show in the version.

proteusvacuum (approved):

apps/experiments/models.py
SmittieC: Good call, I'll update

SmittieC: https://github.com/dimagi/open-chat-studio/pull/869/commits/d166b39ff88e83eedd3aa7f38909bf98417781e6

apps/pipelines/nodes/nodes.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/869/commits/d166b39ff88e83eedd3aa7f38909bf98417781e6

apps/experiments/models.py
snopoke: I suggest we move this to an advanced experiment config. In future there may be other ways to use attachments other than with pipelines.

apps/pipelines/nodes/nodes.py
```diff
@@ -63,7 +71,8 @@ def all_variables(in_):
             # As a last resort, just set the all the variables in the template to the input
             content = all_variables(input)
         template = SandboxedEnvironment().from_string(self.template_string)
-        return template.render(content)
+        output = template.render(content)
+        return PipelineState(messages=[output], outputs={node_id: output})
```
snopoke, line 75: could be nice to make this a classmethod:

```python
def for_node_output(node_id, output)
   ...
```

apps/pipelines/nodes/nodes.py
```diff
@@ -488,3 +507,55 @@ def post_extraction_hook(self, output, state):
                 team=session.team,
                 data=output,
             )
+
+
+class AssistantNode(Passthrough):
```
snopoke, line 512: why does this inherit from `Passthrough`?

apps/pipelines/nodes/nodes.py
```diff
@@ -488,3 +507,55 @@ def post_extraction_hook(self, output, state):
...
+        attachments = [Attachment.model_validate(params) for params in state.get("attachments", [])]
+        chain_output: ChainOutput = runnable.invoke(input, config={}, attachments=attachments)
+        output = chain_output.output
+
+        return PipelineState(
+            messages=[output],
+            outputs={node_id: output},
+            message_metadata={
+                "input": runnable.state.get_message_metadata(ChatMessageType.HUMAN),
+                "output": runnable.state.get_message_metadata(ChatMessageType.AI),
```
snopoke, line 545: Seeing this now I think in the long run it would be better to have this added to the chain output. OK to leave it for now though.

apps/service_providers/llm_service/state.py
```diff
@@ -355,3 +331,68 @@ def citations_enabled(self):
...
+        return self.session.chat
+
+    @cache
+    def get_llm_service(self):
+        return self.assistant.llm_provider.get_llm_service()
+
+    def _get_provider_model_name(self) -> str:
+        return self.assistant.llm_provider_model.name
+
+    def get_chat_model(self):
```
snopoke, line 365: Where is this called from?

apps/pipelines/nodes/nodes.py
SmittieC: That is an excellent question and observation. Copy pasta with a pinch of forgetfullness. I'll fix, thanks for pointing it out

apps/service_providers/llm_service/state.py
SmittieC: Hmm, looks like another copy pasta leftover. I copied a bunch of methods in one go and didn't double check this method's usage. Should be fixed [here](https://github.com/dimagi/open-chat-studio/pull/869/commits/b64af838631ec1744469ff955cb828d13f28a114)

apps/pipelines/nodes/nodes.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/869/commits/05f72d601cbb53144d864d5dd7fbe26cc1da58c5

apps/pipelines/nodes/nodes.py
SmittieC: Cool, agreed. We can then also remove [this ugly](https://github.com/dimagi/open-chat-studio/blob/7c2a37cb157a6e71f683e954260f2a2f772c3f0e/apps/chat/bots.py#L227) code by returning a proper object + metadata.

apps/experiments/models.py
SmittieC: @snopoke do you mean we should add a setting to indicate if the user should be able to upload files, in which case it's a good idea, but a little bit separate from what this method is doing. Confirming this is what you meant?

apps/pipelines/nodes/nodes.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/869/commits/eb528cb4b649bfe8e5192d71a4982bc0a62d7fcf

apps/experiments/models.py
SmittieC: Ok shap. I'll add that in a separate PR

snopoke (approved):
small comments but otherwise looks good

snopoke (approved):

PR #871, opened 2024-11-15
Management command to bootstrap versions for a team

snopoke:
## Description

```
./manage.py enable_versioning_for_team --help

usage: manage.py enable_versioning_for_team [-h] [--force] [--continue-on-error] team

positional arguments:
  team                  The team slug to enable versioning for.

options:
  --force               Force enable versioning.
  --continue-on-error   Continue creating versions even if an error occurs.

```

This will create versions for all experiments in the team (that don't already have versions) and add the team to the feature flag.

It will stop if there are any errors unless the `continue-on-error` flag is set.

Use `--force` to run this for a team that already has the feature flag enabled.

# Bonus

When I ran this locally it gave some errors when syncing tools to OpenAI. This was caused by  a change I made recently to use 'strict' mode which some of our tools don't support. For now I've disabled strict mode if the tool doesn't support it.

### DIFF
--- a/apps/assistants/sync.py
+++ b/apps/assistants/sync.py
@@ -123,21 +123,16 @@ def _convert_to_openai_tool(tool):
     except KeyError:
         return function
 
-    # OpenAI function schemas don't support 'format' or 'default' and must always specify a type
+    # check if this function can use 'strict' mode
     properties = parameters.get("properties", {})
-    for prop, schema in properties.items():
-        if schema.pop("format", False):
-            if "type" not in schema:
-                schema["type"] = "string"
-        schema.pop("default", False)
+    # all fields are required
+    is_strict = set(parameters["required"]) == set(properties)
+    if is_strict:
+        for prop, schema in properties.items():
+            # format and default not supported + type must be present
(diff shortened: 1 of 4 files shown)


stephherbers (approved):

apps/files/models.py
```diff
@@ -48,7 +48,7 @@ def duplicate(self):
             schema=self.schema,
             team=self.team,
         )
-        if self.file:
+        if self.file and self.file.storage.exists(self.file.name):
```
SmittieC, line 51: Is / was there a case that necessitates this extra check, or is this precautionary?

apps/assistants/sync.py
```diff
@@ -115,6 +115,27 @@ def push_assistant_to_openai(assistant: OpenAiAssistant, internal_tools: list |
         assistant.save()
 
 
+def _convert_to_openai_tool(tool):
```
SmittieC, line 118: :+1:

apps/files/models.py
snopoke: There shouldn't be but it's possible the file get's removed. I did have this happen locally but that's because I cleared the files.

PR #875, opened 2024-11-18
Version assistant nodes

SmittieC:
Will mark this ready for review when the base branch is merged

## Description

- Version the linked assistant when the node is being versioned
- Allow only working version pipelines to be linked to an experiment

## User Impact

More stable assistant pipelines

### Demo

N/A

### Docs

N/A

### DIFF
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -246,7 +246,7 @@ class Node(BaseModel, VersionsMixin):
         from apps.assistants.models import OpenAiAssistant
         from apps.pipelines.nodes.nodes import AssistantNode
 
-        assistant_node_name = AssistantNode.model_json_schema()["title"]
+        assistant_node_name = AssistantNode.__name__
 
         new_version = super().create_new_version(save=False)
 
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -238,6 +238,28 @@ class Node(BaseModel, VersionsMixin):
     def __str__(self):
         return self.flow_id
 
+    def create_new_version(self, save: bool = True):
+        """
+        Create a new version of the node and if the node is an assistant node, create a new version of the assistant
+        and update the `assistant_id` in the node params to the new assistant version id.
+        """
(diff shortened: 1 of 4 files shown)


apps/pipelines/models.py
```diff
@@ -238,6 +238,28 @@ class Node(BaseModel, VersionsMixin):
...
 
+    def create_new_version(self, save: bool = True):
+        """
+        Create a new version of the node and if the node is an assistant node, create a new version of the assistant
+        and update the `assistant_id` in the node params to the new assistant version id.
+        """
+        from apps.assistants.models import OpenAiAssistant
+        from apps.pipelines.nodes.nodes import AssistantNode
+
+        assistant_node_name = AssistantNode.model_json_schema()["title"]
```
snopoke, line 249: Isn't this just `AssistantNode.__name__`

SmittieC: Ah yes it is. Forgot about this fella, thanks

apps/pipelines/tests/test_models.py
```diff
@@ -9,3 +12,37 @@ def test_archive_pipeline_archives_nodes_as_well():
...
+
+@pytest.mark.django_db()
+class TestNode:
+    @pytest.mark.parametrize("versioned_assistant_linked", [True, False])
+    @patch("apps.assistants.sync.push_assistant_to_openai", Mock())
+    def test_versioning_assistant_node(self, versioned_assistant_linked):
+        """
+        Versioning an assistant node should version the assistant as well, but only when the linked assistant is not
+        already a version
+        """
```
stephherbers, line 25: > only when the linked assistant is not already a version
when would this case happen? Wouldn't this be versioned at the same time as the experiment when the assistant versioning happens as well?

SmittieC: This can happen when I for instance have an assistant version that I'd like to call in the pipeline. Not 100% how often this would be the case, but it is a real possibility

snopoke (commented):
LGTM

PR #877, opened 2024-11-18
Case insensitive commands

SmittieC:
## Description

Make user input case insensive when checking commands

## User Impact

These commands will all reset the chat
- /reset
- /Reset
- /RESET

### Docs

I'll add a section in the docs once merged

### DIFF
--- a/apps/chat/channels.py
+++ b/apps/chat/channels.py
@@ -489,7 +489,7 @@ class ChannelBase(ABC):
         )
 
     def _is_reset_conversation_request(self):
-        return self.user_query.lower() == ExperimentChannel.RESET_COMMAND
+        return self.user_query.lower().strip() == ExperimentChannel.RESET_COMMAND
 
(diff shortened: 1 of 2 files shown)


apps/chat/channels.py
```diff
@@ -489,7 +489,7 @@ def _create_new_experiment_session(self):
         )
 
     def _is_reset_conversation_request(self):
-        return self.user_query == ExperimentChannel.RESET_COMMAND
+        return self.user_query.lower() == ExperimentChannel.RESET_COMMAND
```
snopoke, line 492: How about also adding `.strip()` to remove any whitespace?

SmittieC: https://github.com/dimagi/open-chat-studio/pull/877/commits/0cb7d48c827e6a88961d6da2932f103b7b3a1bee

stephherbers (approved):

Issue #873, opened 2024-11-15
Support archiving experiments


bderenzi:
Right now the only way to remove an experiment from the list is to delete it. That's rather permanent. Can we instead archive it and have some way to look at archived projects?

SmittieC:
Sounds good. Just to be sure I understand the context: I remember you saying that people generally don't delete their bots. Do you have a sense of why? I'm presuming one reason would be to reference it agan in the future, as you mentioned above. If that's the case, I can imagine a "favourites" list feature might come in handy :thinking: 

I'm going out on a limb here

SmittieC:
This is done

PR #880, opened 2024-11-18
Versioning UI Updates: edit experiment page updates

stephherbers:
## Description

Adds a "create version" and "delete" buttons to the edit screen
removed delete from the tooltip

## User Impact

the delete experiment will be moved, but this is not a frequestly used button but still worth flagging

### Demo

[loom video
](https://www.loom.com/share/9688d01739fc43b8aa6c98460b55f595?sid=5b6d05c8-f1bb-437f-b836-43bf82d3ff6f)
[loom video of making new version](https://www.loom.com/share/8d86324e90f44cfd9c55bdfdc98c5cc4?sid=e09eaae5-eb4d-488f-80fe-1aedcdd807a2)
Screenshot 2024-11-20 at 12 03 18 PM
Screenshot 2024-11-20 at 12 03 18 PM
disabled when no changes are made from last version

### Docs

n/a


templates/generic/object_form.html
```diff
@@ -23,6 +24,11 @@ <h1 class="pg-title">{{ title }}</h1>
         {% endblock form %}
         {% block form_actions %}
           <input type="submit" class="pg-button-primary mt-2" value="{{ button_text }}">
+          {% flag "experiment_versions" %}
+            {% if show_version_action_options %}
+              {% include 'experiments/experiment_version_edit_actions.html' %}
+            {% endif %}
+          {% endflag %}
```
snopoke, line 31: These changes should go in `experiment_form.html`. You can override the `form_actions` block.

That would also remove the need for the `show_version_action_options` var. You could instead check `object.id` to determine if the experiment is saved or not and use that as a display condition for the delete button

SmittieC: Also, we probably want to show the delete button regardless of the `experiment_versions` flag status. Looks like it will only show when the flag is active

SmittieC (commented):
General thought: I'm wondering if it would make sense if we move the buttons to the top of the form to under the "Update Experiment" title. Thoughts?

SmittieC:
May I ask that you also add an "archive" button to where the buttons are moved? The button should archive the working version. Looks like we can simply do it by calling this URL:
```python
{% url 'experiments:archive-experiment' team_slug=team.slug experiment_id=experiment.working_version_id version_number=experiment.version_number %}
```

stephherbers:
> General thought: I'm wondering if it would make sense if we move the buttons to the top of the form to under the "Update Experiment" title. Thoughts?

I like all the potential actions being in the same area of the page, but open to moving to the top

stephherbers:
> May I ask that you also add an "archive" button to where the buttons are moved (resolving https://github.com/dimagi/open-chat-studio/issues/873)? The button should archive the working version. Looks like we can simply do it by calling this URL:
> 
> {% url 'experiments:archive-experiment' team_slug=team.slug experiment_id=experiment.working_version_id version_number=experiment.version_number %}

@SmittieC I looked into adding this and this method is for archiving a version and not the working version, so I think Ill do it in a separate PR since it requires adding a different view if that sounds okay just to prevent this PR from getting too large

apps/experiments/views/experiment.py
```diff
@@ -452,6 +452,11 @@ class EditExperiment(BaseExperimentView, UpdateView):
     button_title = "Update"
     permission_required = "experiments.change_experiment"
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["show_version_action_options"] = True
```
snopoke, line 457: this can be removed now right?

templates/experiments/experiment_form.html
```diff
@@ -122,6 +122,17 @@ <h3 class="text-md font-bold">{% translate "Please note" %}</h3>
       </div>
     </div>
 {% endblock form %}
+{% block form_actions %}
+  <input type="submit" class="pg-button-primary mt-2" value="{{ button_text }}">
+  {% flag "experiment_versions" %}
+    {% if show_version_action_options %}
```
snopoke, line 128: I think this if statement isn't necessary now that this is in the experiment form template

templates/experiments/experiment_version_edit_actions.html
```diff
@@ -0,0 +1,3 @@
+<a href="{% url 'experiments:create_version' team.slug experiment.id %}" class="btn btn-secondary" style="float: right;">
+  Create Version
+</a>
```
snopoke, line 3: Seems like this is only referenced in one place - maybe just inline to content?

snopoke (commented):
A screenshot / demo would be great :)

stephherbers:
Screenshot 2024-11-20 at 10 20 34 AM
I don't know why the photo is not appearing in the description @snopoke @SmittieC

templates/experiments/experiment_form.html
```diff
@@ -122,6 +122,17 @@ <h3 class="text-md font-bold">{% translate "Please note" %}</h3>
       </div>
     </div>
 {% endblock form %}
+{% block form_actions %}
+  <input type="submit" class="pg-button-primary mt-2" value="{{ button_text }}">
```
snopoke, line 126: the `mt-2` on this is going to put the alignment out - maybe wrap them all in a div with `mt-2`

templates/experiments/experiment_form.html
```diff
@@ -122,6 +122,17 @@ <h3 class="text-md font-bold">{% translate "Please note" %}</h3>
...
+  {% flag "experiment_versions" %}
+    <a href="{% url 'experiments:create_version' team.slug experiment.id %}" class="btn btn-secondary" style="float: right;">
```
snopoke, line 128: I don't think this is going to save the experiment - you'll loose changes.

You can use a normal form submit button and add a name attribute so that you tell which button got pressed:

```
<button type="submit" name="action" value="save_and_version">Create Version</button>
```

On the server side check `request.POST.get("action") == "save_and_version"`

snopoke (commented):
I think this needs a bit more work:

* Prefer not to use custom styles, tailwind has classes for everything: https://tailwindcss.com/docs/float
* Also I don't think it looks quite right with the buttons on the right - they are a bit out of line with the tabs
* I think you need to re-generate your CSS (`inv npm`) - the 'secondary' color is not showing up in your screenshot

stephherbers:
@snopoke @SmittieC updated with latest loom video and screenshot in the description

bderenzi:
This looks great! Could we disable the create version button when there are no outstanding changes? Clicking it and going to a whole new screen to learn that there are no changes seems a bit disruptive

stephherbers:
> This looks great! Could we disable the create version button when there are no outstanding changes? Clicking it and going to a whole new screen to learn that there are no changes seems a bit disruptive
@bderenzi will do 👍

apps/experiments/views/experiment.py
```diff
@@ -363,6 +363,19 @@ def extra_context(self):
             team_participant_identifiers.extend(self.object.participant_allowlist)
             team_participant_identifiers = set(team_participant_identifiers)
 
+        excluded_fields = [
```
stephherbers, line 366: offlined, but outstanding questing if these can be added to the default excluded keys

apps/experiments/views/experiment.py
```diff
@@ -363,6 +363,19 @@ def extra_context(self):
...
+            "version_number",
+            "is_default_version",
+            "public_id",
+            "version_description",
+        ] + self.object.get_fields_to_exclude()
+        fields_changed = self.object.latest_version.compare_with_model(self.object, excluded_fields)
+        if "consent_form" in fields_changed:
```
stephherbers, line 373: also offlined, but here is the (temp) fix for the consent forms being matched by id (which is different version to version always)

stephherbers:
notes on the formatting-- I ran it locally on the file and it was a little wild so I didn't push it

apps/experiments/views/experiment.py
snopoke: Seems fine to add `version_number`. I think the others are experiment specific so probably best to override `get_fields_to_exclude` on the Experiment model to add them there.

apps/experiments/views/experiment.py
```diff
@@ -363,6 +363,19 @@ def extra_context(self):
             team_participant_identifiers.extend(self.object.participant_allowlist)
             team_participant_identifiers = set(team_participant_identifiers)
 
+        excluded_fields = [
+            "version_number",
+            "is_default_version",
+            "public_id",
+            "version_description",
+        ] + self.object.get_fields_to_exclude()
+        fields_changed = self.object.latest_version.compare_with_model(self.object, excluded_fields)
```
snopoke, line 372: I think here you want to follow the pattern in the create version view:
```python
working_experiment = self.object
version = working_experiment.version
if prev_version := working_experiment.latest_version:
    # Populate diffs
    version.compare(prev_version.version)
```

I think it would also be useful to make this into a method on the Experiment model so that you can call `experiment.compare_with_latest()`. Another useful method would be `experiment.has_changes()` which may be able to shortcircuit the process when it finds the first diff.

apps/experiments/views/experiment.py
```diff
@@ -452,6 +471,10 @@ class EditExperiment(BaseExperimentView, UpdateView):
     button_title = "Update"
     permission_required = "experiments.change_experiment"
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        return context
```
snopoke, line 476: nit: this can be removed

templates/experiments/single_experiment_home.html
```diff
@@ -61,25 +61,6 @@ <h1 class="pg-title">{{ experiment.name }}</h1>
               <i class="fa-solid fa-pencil"></i>
             </a>
           </div>
-          <div class="tooltip" data-tip="Delete">
-            <button
-              id="action-delete-{{ experiment.id }}"
-              class="btn btn-primary join-item rounded-r-full"
```
snopoke, line 67: I think yo should move the `rounded-r-full` class to the previous button (the one that's now on the end)

snopoke (approved):
Small comments but otherwise looks good. Not blocking

snopoke (approved):

templates/experiments/experiment_form.html
```diff
@@ -122,6 +122,19 @@ <h3 class="text-md font-bold">{% translate "Please note" %}</h3>
...
     </div>
 {% endblock form %}
+{% block form_actions %}
+  <div class="mt-2">
+    <input type="submit" class="pg-button-primary" value="{{ button_text }}">
+    {% if experiment.id %}
+      {% flag "experiment_versions" %}
+        <button type="submit" class="pg-button-secondary" name="action" {% if disable_version_button %}disabled="True"{% endif %} value="save_and_version">Create Version</button>
+      {% endflag %}
+      <a href="{% url 'experiments:delete' team.slug experiment.id %}" class="pg-button-danger" onclick="return confirm('Are you sure you want to delete this experiment? This action cannot be undone.')">
```
snopoke, line 132: @stephherbers this doesn't redirect after clicking it - it should go back to the experiment list page

PR #881, opened 2024-11-19
Prevent multiple files from being returned

stephherbers:
## Description

[sentry](https://dimagi.sentry.io/issues/5998995252/?project=4505001320316928&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=14d&stream_index=0)

feels like a bandaid-- why are there multiple files with the same external id in the team in the first place? but that should be resolved in another PR since this check doesn't hurt to add

## User Impact

prevents it from erroring out

### Demo

n/a

### Docs

n/a


snopoke:
> feels like a bandaid-- why are there multiple files with the same external id in the team

See comment on https://github.com/dimagi/open-chat-studio/pull/882

PR #882, opened 2024-11-19
File creation refactor

snopoke:
## Description
Refactor file creation from external sources to get more consistency and to check for existing files.

## Duplicate files thoughts
We are still going to have duplicate files because when we create a new version of an assistant we duplicate the attached files. One option would be to clear out the 'external_id' on the file when we duplicate it. This would force us to re-upload the file to OpenAI which would assign it a new file ID. Thoughts?


SmittieC (approved):
LGTM: Regarding your comment: Have a look at [this PR](https://github.com/dimagi/open-chat-studio/pull/883). It basically does what you suggested

PR #883, opened 2024-11-19
Clear vector_store_id and file external_ids when versioning assistants

SmittieC:
## Description

- This is a fix for the root cause of the multiple file ids being returned. We forgot to clear the `vector_store_id` from the tool resource, as well as the `external_id` from the files when being duplicated.
- I added a management command to fix the existing versions where there are duplicates. Basically we clear those, then we update the assistant version at OpenAI's side to clear the vector stores there, then we sync the version manually.

## User Impact

This fixes the multiple files returned issue

### Demo

Logs from a local run
```
Evaluating Broken (id=9)
Looking at version 1 (id=10)
Vector store ids match! (vector_store_id=vs_99sBwgoOvc93VxdMMZ)
...clearning vector_store_id
...clearing file external ids
...removing existing vector store id at OpenAI
...syncing version with OpenAI
...syncing complete!
Version's new vector_store_id is vs_xUm4g9God6Fb6T5VTFs
Diff between original and new file ids: {'file-buOcVHVygsV0dU5rYb'}


Evaluating Fixed one (id=7)
Looking at version 1 (id=8)
Vector store ids do not match. Continuing...


Evaluating Versioning (id=5)
Looking at version 1 (id=6)
Vector store ids do not match. Continuing...

```


stephherbers (approved):
great addition of the clear_external_id param

apps/experiments/management/commands/fix_assistant_file_duplication.py
```diff
@@ -0,0 +1,80 @@
...
+                                self.stdout.write("Vector store ids do not match. Continuing...")
+                                continue
+
+                            self.stdout.write(f"Vector store ids match! (vector_store_id={original_vector_store_id})")
+                            self.stdout.write("...clearning vector_store_id")
+                            tool_resource.extra["vector_store_id"] = None
+                            tool_resource.save()
+                            self.stdout.write("...clearing file external ids")
+                            for file in tool_resource.files.all():
+                                file.external_id = ""
```
snopoke, line 55: Also clear `external_source`

snopoke (commented):
This is great, thanks. A few clarifying comments.

apps/experiments/management/commands/fix_assistant_file_duplication.py
```diff
@@ -0,0 +1,80 @@
...
+                        self.stdout.write(f"\n\nEvaluating {assistant} (id={assistant.id})")
+                        original_resource = assistant.tool_resources.get(tool_type="file_search")
+                        original_vector_store_id = original_resource.extra.get("vector_store_id")
+                        original_file_ids = set(original_resource.files.values_list("external_id", flat=True))
+
+                        for version in assistant.versions.all():
+                            self.stdout.write(f"Looking at version {version.version_number} (id={version.id})")
+                            tool_resource = version.tool_resources.get(tool_type="file_search")
+                            if tool_resource.extra.get("vector_store_id") != original_vector_store_id:
+                                self.stdout.write("Vector store ids do not match. Continuing...")
```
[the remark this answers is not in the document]
SmittieC: Not in OCS. At least not in the way we do versioning. See [here](https://github.com/dimagi/open-chat-studio/blob/76aad5e76bd87120f08601dc9d97a8428bf18c29/apps/assistants/models.py#L122-L132). When the assistant is versioned, we create a new `ToolResource` and copy the working version's `extra` to the new one's extra, which is why we ended up with multiple `ToolResource` instances that had the same `vector_store_id` attached.

We also go through all the files linked to the working version's tool resource and duplicate them to the assistant version's tool resource.

Hope this clears it up a bit?

apps/experiments/management/commands/fix_assistant_file_duplication.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/883/commits/b87672afd7e1c8a9e19dc46395c1d5954d2558d3

SmittieC:
I added a test to instill a bit more confidence in the command

snopoke (approved):

PR #884, opened 2024-11-19
Refactor prompt context

snopoke:
## Description
Extract common functionality for getting prompt context into a separate class. This encapsulated the functionality, allows reuse and also caching.


SmittieC (dismissed):
:guitar:

PR #885, opened 2024-11-19
allow referencing participant data internals

snopoke:
## Description
This change makes it possible to selectively include parts of the participant data in prompts and input formats.

For example:
```
{participant_data.name}
{participant_data.schedules.0.is_complete}
```

If any attribute or item access fails, the result will be an empty string (see test cases).

The use cases for this are probably quite niche but likely to be useful in routing. It also makes it possible to split up the participant data in the prompt (since you can reference the variable multiple times)

### Docs
TODO


bderenzi (approved):
This looks great! 

The tests that you added feel like they could flow right into documentation about how to use this

PR #892, opened 2024-11-20
Refactor pipeline frontend to use json schemas

snopoke:
## Description
The main reason for this change to is to leverage more standard schemas rather than our own custom schema. I think this also makes the node classes more standard.

Prior to this the pipelines UI used a [custom data structure](https://github.com/dimagi/open-chat-studio/blob/main/apps/pipelines/nodes/utils.py) for building the node widgets.

This PR moves to using JsonSchema's generated by pydantic from the node models.

There are a few [custom pieces](https://github.com/dimagi/open-chat-studio/blob/sk/pipeline-json-schema/apps/pipelines/nodes/base.py#L107-L148) added to the schemas to support custom widgets, labels and option sources.

This PR also does some refactoring and fixes a few small bugs.

## User Impact
None

### Docs
Some here: https://github.com/dimagi/open-chat-studio/blob/sk/pipeline-json-schema/apps/pipelines/README.md


proteusvacuum (approved):
Very cool!

apps/pipelines/nodes/nodes.py
```diff
@@ -74,9 +65,11 @@ def all_variables(in_):
 
 
 class LLMResponseMixin(BaseModel):
-    llm_provider_id: LlmProviderId
-    llm_provider_model_id: LlmProviderModelId
-    llm_temperature: LlmTemperature = Field(default=1.0, gt=0.0, le=2.0)
+    llm_provider_id: int = Field(..., title="LLM Model", json_schema_extra=UiSchema(widget=Widgets.llm_provider_model))
+    llm_provider_model_id: int = Field(..., json_schema_extra=UiSchema(widget=Widgets.none))
```
SmittieC, line 69: Do we have to use the `Widgets.none` widget instead of `None` from now on?

SmittieC (approved):
How do we regenerate the schemata in the data folder if we change / add nodes?

snopoke:
> How do we regenerate the schemata in the data folder if we change / add nodes?

There's a test in the same file that's skipped - you unskip it and run the test: `test_update_schemas`. Maybe I should convert that into a management command. The main reason for having the schemas in the codebase is because it makes it easier to see what's going to change on the UI.

apps/pipelines/nodes/nodes.py
snopoke: Depends what you want. `Widgets.none` will not display the field at all. If you want to use the default widget then you don't have to specify one and it will use one based on the field type.

apps/pipelines/nodes/base.py
```diff
@@ -41,7 +43,7 @@ def json_safe(self):
         return copy
 
     @classmethod
-    def from_node_output(cls, node_id: str, output: any = None, **kwargs) -> "PipelineNode":
+    def from_node_output(cls, node_id: str, output: any = None, **kwargs) -> Self:
```
proteusvacuum, line 46: FYI, I was getting type hint errors from this change: 
```
Expression of type "PipelineState" cannot be assigned to return type "Self@PipelineState"
```
I'll update to return `cls(**kwargs)`

PR #893, opened 2024-11-20
Ignore the version suffix when comparing assistant names

### LINKED ISSUE #890: [Versioning] Comparing an assistant with a version of itself always reports a diff

SmittieC:
resolves #890 

## Description

Since we appended the version as a suffix to the assistant version name and it will always change, we must ignore it when computing the difference between two versions. I think it makes sense to literally add the version in the name instead of doing it in the `__str__` method, since then we don't have to manually add the version each time we care about query results. A good example is in pipelines when we list the available assistants. It doesn't use the `__str__` representation, but rather the DB name

## User Impact

The diff view will show the assistant changed when it actually did

### DIFF
--- a/apps/assistants/models.py
+++ b/apps/assistants/models.py
@@ -1,3 +1,5 @@
+from typing import Self
+
 from django.contrib.postgres.fields import ArrayField
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models, transaction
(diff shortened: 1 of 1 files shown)


snopoke (approved):
Do we care if the name changes at all?

SmittieC:
> Do we care if the name changes at all?

I can imagine if you change you assistant's behaviour significantly, you might want to rename is (super edge case), but other than that, probably not

PR #895, opened 2024-11-21
Send test messages to a pipeline

proteusvacuum:
## Description

Adds the ability to send test messages through a pipeline. This allows users to validate that the pipeline builds, and users can also visualize the outputs / inputs between each node.

## User Impact

Users will have much faster feedback on what is happening within a pipeline. 

### Demo

https://github.com/user-attachments/assets/d2899b91-9896-4119-b6f8-938f054b56d6

image


snopoke (commented):
This looks useful and a good starting point. I pushed a few commits to tweak the UI.

Future thoughts:
* use a mock session for participant data access
* show progress incrementally

bderenzi:
Cool that we can see the path a message takes! Agreed with Simon that this looks useful and like a good starting point.

One other future thought is that we'll want to chat to individual nodes and/or subsets of the pipeline. Choosing an arbitrary start and end node would be useful.

snopoke (approved):

PR #896, opened 2024-11-21
Sk/pipeline misc

snopoke:
## Description

* Some followup refactoring to widgets to reduce indirection and give a more standard interface
* Reset the viewport on load (so that all nodes are visible)
* Don't save the viewport to the DB (and don't trigger saves when the viewport changes)
* Update scroll mode to make navigation easier
* Fix error border color
* Fix editor panel when switching nodes

I've tried to keep the commits clean so best reviewed by commit.

### Demo
<div>
    <a href="https://www.loom.com/share/40faff8dd962471f865fd2232427eb73">
      <p>Pipline UI updates - 21 November 2024 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/40faff8dd962471f865fd2232427eb73">
      
    </a>
  </div>


assets/javascript/apps/pipeline/nodes/GetInputWidget.tsx
```diff
@@ -86,150 +81,7 @@ export const getInputWidget = (params: InputWidgetParams) => {
       schema={params.schema}
       nodeParams={params.params}
       required={params.required}
+      getFieldError={getFieldError}
```
SmittieC, line 84: Why not use the existing `inputError` prop?

apps/pipelines/nodes/nodes.py
```diff
@@ -256,7 +262,9 @@ class RouterNode(Passthrough, HistoryMixin):
     model_config = ConfigDict(json_schema_extra=NodeSchema(label="Router"))
 
     prompt: str = Field(
-        default="You are an extremely helpful router", json_schema_extra=UiSchema(widget=Widgets.expandable_text)
+        default="You are an extremely helpful router",
+        min_length=1,
```
SmittieC, line 266: Why is this length 1?

SmittieC (approved):
Looks good to me, thanks. Just two non major comments

apps/pipelines/nodes/nodes.py
snopoke: This was my attempt to make it required. This can definitely be improved.

assets/javascript/apps/pipeline/nodes/GetInputWidget.tsx
snopoke: Because the history widget represents 2 fields so it needs access to the error for both fields.

PR #897, opened 2024-11-21
Only verify email when participant data is being injected into the prompt

SmittieC:
See [this thread](https://dimagi.slack.com/archives/C05PK63Q89H/p1732192937178459)

## Description

I think it makes sense to only verify email addresses when participant data is being injected. We're not really sharing sensitive data if it's not being injected. Email verification is seen as a slight hinderance when we want to get mass participantion

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -1566,3 +1566,6 @@ class ExperimentSession(BaseTeamModel):
         experiment's version number
         """
         return self.chat.metadata.get(Chat.MetadataKeys.EXPERIMENT_VERSION, Experiment.DEFAULT_VERSION_NUMBER)
+
+    def requires_participant_data(self) -> bool:
+        return "{participant_data}" in self.experiment.prompt_text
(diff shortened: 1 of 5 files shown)


snopoke (approved):

apps/experiments/models.py
```diff
@@ -1566,3 +1566,6 @@ def get_experiment_version_number(self) -> int:
         experiment's version number
         """
         return self.chat.metadata.get(Chat.MetadataKeys.EXPERIMENT_VERSION, Experiment.DEFAULT_VERSION_NUMBER)
+
+    def requires_participant_data(self) -> bool:
+        return "{participant_data}" in self.experiment.prompt_text
```
snopoke, line 1571: what if it's a pipeline or assistant?

SmittieC: huh, yeah. Forgot about that, thanks

apps/pipelines/models.py
```diff
@@ -213,6 +213,9 @@ def archive(self):
         super().archive()
         self.node_set.update(is_archived=True)
 
+    def get_node_param_values(self, node_cls, param_name: str) -> list:
+        return list(self.node_set.filter(type=node_cls.__name__).values_list(f"params__{param_name}", flat=True))
```
snopoke, line 217: nice

snopoke (approved):

PR #900, opened 2024-11-21
update handle with node changes

snopoke:
## Description
I noticed that as we add / remove error messages to fields in the nodes, the output handles get out of alignment with the text:

image

This change re-renders the handles when the node size changes.


SmittieC (approved):
Huh, I thought react was built to make life easier..

PR #902, opened 2024-11-22
Create new experiment version async

### LINKED ISSUE #901: Create versions async

SmittieC:
Resolves #901 

## Description

Since creating versions of router experiments with assistant children and many files bombed out (gateway timeout), we're now creating versions async.

## User Impact

users should be able to create versions of large assistants. Should being the keyword here

### Demo

Will include one once the UI is done


SmittieC:
Ready

SmittieC:
I'm not convinced that we should be creating a bunch of tasks in the management command, but it will fail for some assistants if we do not. Maybe we should be smarter in the management command and figure out when to give it to a task? I propose:
1. We see how this does with the large assistants
2. If it goes well, we can create versions async for experiments with assistants that have > X files.

stephherbers (approved):

apps/experiments/tasks.py
```diff
@@ -15,6 +15,13 @@
 from apps.utils.taskbadger import update_taskbadger_data
 
 
+@shared_task(bind=True, base=TaskbadgerTask)
+def async_create_experiment_version(self, experiment_id: int, *arg, **kwargs) -> dict:
```
snopoke, line 19: nit: specify the kwargs explicitly

snopoke (changes requested):
Currently this won't give any feedback that something is happening behind the scenes. I think it's likely users will do the workflow repeatedly until they see something on the UI change.

My thought is that once the user clicks the button, the form is submitted using HTMX and the page content is replaced with a progress bar which polls for task completion.

apps/experiments/tasks.py
SmittieC: Addressed both of your comments (see the demo)

apps/experiments/models.py
```diff
@@ -588,6 +588,7 @@ class Experiment(BaseTeamModel, VersionsMixin):
     )
     debug_mode_enabled = models.BooleanField(default=False)
     citations_enabled = models.BooleanField(default=True)
+    create_version_task_id = models.CharField(max_length=128, blank=True)
```
SmittieC, line 591: This field's value is not really being used directly, but I think we should keep it like this. Might come in handy if we're wanting to debug something

snopoke (approved):

Issue #903, opened 2024-11-22
No drop-down if only one or the other is used for file attachment


bderenzi:
We've had a request to remove the drop-up menu containing "code interpreter" and "file search". Instead, allow users to upload files directly by clicking the clipper icon.

This would be relevant if an experiment has selection only either "code interpreter" or "file search". Possibly we should allow bot-builders to choose which upload is enabled and if there's only one then we allow upload with the attachment icon directly

PR #907, opened 2024-11-22
don't show chat attachment dropdown if only one kind of file input is allowed

snopoke:
## Description
Resolves https://github.com/dimagi/open-chat-studio/issues/903

#### Tech note:
This adds [django-template-partials](https://github.com/carltongibson/django-template-partials?tab=readme-ov-file#basic-usage) library which can be used instead of 'include' in certain places.

## User Impact
Users don't have to click twice to upload files when only one kind of file is allowed.

### Demo
<div>
    <a href="https://www.loom.com/share/9cc7157966834047a252aafa4946efc7">
      <p>Dimagi Chatbots | Chat attachment dropdown changes - 22 November 2024 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/9cc7157966834047a252aafa4946efc7">
      
    </a>
  </div>

### DIFF
--- a/templates/experiments/chat/input_bar.html
+++ b/templates/experiments/chat/input_bar.html
@@ -20,55 +20,59 @@
   <div class="flex flex-col" {% if assistant %}x-data="fileUploads"{% endif %}>
     <div class="flex items-center w-full">
       {% if assistant %}
-        <div class="flex flex-row gap-2">
-          <div class="dropdown dropdown-top">
-            <div tabindex="0" role="button" class="btn m-1"><i class="fa-solid fa-paperclip"></i></div>
-            <ul tabindex="0" class="dropdown-content menu bg-base-100 rounded-box z-[1] w-52 p-2 shadow">
-              <li>
-                <a
-                  class="btn btn-sm"
-                  {% if not assistant.supports_code_interpreter %}
-                    disabled
-                  {% else %}
-                    x-bind:disabled="disableCodeInterpreterUpload"
-                  {% endif %}
-                  @click="$refs.codeInterpreter.click()">
-                  Code Interpreter
-                </a>
-              </li>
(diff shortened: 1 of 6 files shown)


templates/experiments/chat/input_bar.html
```diff
@@ -20,55 +20,59 @@
...
+                class="file-input file-input-bordered file-input-sm w-full max-w-xs"
+              />
+            {% endpartialdef %}
+          </div>
+        {% elif assistant.supports_code_interpreter %}
+          <div role="button" class="btn m-1" x-bind:disabled="disableCodeInterpreterUpload"
+               @click="$refs.codeInterpreter.click()">
+            <i class="fa-solid fa-paperclip"></i>
+          </div>
+          {% partial code-interpreter-input %}
```
snopoke, line 70: This tag renders the template partial defined above

templates/experiments/chat/input_bar.html
```diff
@@ -20,55 +20,59 @@
...
+                <li>
+                  <a x-bind:disabled="disableCodeInterpreterUpload" @click="$refs.codeInterpreter.click()">Code
+                    Interpreter</a>
+                </li>
+                <li>
+                  <a @click="$refs.fileSearch.click()">File Search</a>
+                </li>
+              </ul>
+            </div>
+            {% partialdef code-interpreter-input inline=True %}
```
snopoke, line 38: I'm using 'inline' partials so that they render here and can also be referenced elsewhere

stephherbers: Nice so if without that flag the definition doesn't render it until called later on? Neat

stephherbers (approved):
I'm excited to use template-partials in other places 🌟

PR #910, opened 2024-11-22
Experiment UI: add archive button

stephherbers:
## Description

replace deletion with archiving
I decided to save the experiment prior to archiving since the motivation for archiving is to use it as a reference for the future so we don't want to discard those changes before archiving

## User Impact

users have the ability to archive working experiment with all of it's versions from the edit page

### Demo

[loom](https://www.loom.com/share/3ac1c1a25c8149f2ae92069b263b981e?sid=069019ad-34ba-4ddf-8701-7047511fb394) (the pop up for confirming if you want to archive the experiment does not appear in the demo sorry about that)


templates/experiments/experiment_form.html
```diff
@@ -129,9 +129,12 @@ <h3 class="text-md font-bold">{% translate "Please note" %}</h3>
       {% flag "experiment_versions" %}
         <button type="submit" class="pg-button-secondary" name="action" {% if disable_version_button %}disabled="True"{% endif %} value="save_and_version">Create Version</button>
       {% endflag %}
-      <a href="{% url 'experiments:delete' team.slug experiment.id %}" class="pg-button-danger" onclick="return confirm('Are you sure you want to delete this experiment? This action cannot be undone.')">
-        Delete
-      </a>
+      <form action="{url 'experiments:archive' team_slug=team.slug pk=experiment.id}" method="post" onsubmit="return confirm('Are you sure you want to archive this experiment?');">
```
stephherbers, line 132: keeping as a draft until this is functional-- url works but something is wrong with the from since adding a "dummy" form before this one make this one work

apps/experiments/models.py
```diff
@@ -747,6 +747,13 @@ def compare_with_latest(self):
             version.compare(prev_version.version)
         return version.fields_changed
 
+    def archive_working_experiment(self):
+        if self.is_working_version:
+            if self.has_versions:
+                for version in self.versions.all():
```
SmittieC, line 753: This can be done with a single query. Something like

```python
from field_audit.models import AuditAction

Experiment.objects.filter(working_version=self).update(is_archived=True, audit_action=AuditAction.AUDIT)
```

apps/experiments/models.py
```diff
@@ -747,6 +747,13 @@ def compare_with_latest(self):
             version.compare(prev_version.version)
         return version.fields_changed
 
+    def archive_working_experiment(self):
```
SmittieC, line 750: This looks like it can be done inside the existing `archive` method?

SmittieC (commented):
Overall looks great. One thing I think we should address before merging though:
Users will not be able to see archived experiments, so I'm thinking that we should release the channel resources when the user archives the experiment, otherwise they will have no way of getting for instance a whatsapp channel back to use for another experiment. Basically we should delete the experiment channel instances where this experiment is linked. We tell them that this is what we're going to do beforehand of course.

Or, we could still show the archived versions in the list of experiments so that users will be able to navigate to it, then we can add a tag/pill/"some sort of indicator" to clearly show that these experiments are archived. (and we should disable editing these in this case)

apps/experiments/tests/test_models.py
```diff
@@ -980,6 +981,29 @@ def test_get_version(self, experiment):
...
+        second_version.refresh_from_db()
+
+        assert experiment.is_archived is True
+        assert first_version.is_archived is True
+        assert second_version.is_archived is True
+
+    def test_archive_deletes_channels(self, experiment):
+        experiment_channel = ExperimentChannelFactory(experiment=experiment)
+        experiment_version = experiment.create_new_version()
+        experiment_version_channel = ExperimentChannelFactory(experiment=experiment_version)
```
SmittieC, line 999: We wouldn't really have version channels, since all channels will be linked to the working experiment

SmittieC (approved):
Looks good to me :rocket:

templates/experiments/experiment_form.html
```diff
@@ -129,9 +129,7 @@ <h3 class="text-md font-bold">{% translate "Please note" %}</h3>
       {% flag "experiment_versions" %}
         <button type="submit" class="pg-button-secondary" name="action" {% if disable_version_button %}disabled="True"{% endif %} value="save_and_version">Create Version</button>
       {% endflag %}
-      <a href="{% url 'experiments:delete' team.slug experiment.id %}" class="pg-button-danger" onclick="return confirm('Are you sure you want to delete this experiment? This action cannot be undone.')">
-        Delete
-      </a>
+      <button type="submit" class="pg-button-danger" name="action" value="save_and_archive" onclick="return confirm('Are you sure you want to archive this experiment?');">Archive</button>
```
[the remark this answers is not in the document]
snopoke: I think it's implied.

apps/experiments/tests/test_models.py
snopoke: +1 - can remove this part of the test

apps/experiments/models.py
snopoke: I think the logic here was that if versions could have channels then you'd want to call `.archive` on the versions to also archive the channels. 

But since you can't attach channels to a version I don't think it's necessary to do it that way.

PR #911, opened 2024-11-26
Async chat exports

SmittieC:
## Description

- Added expiry field to file model
- Celery task to clean up expired files
- Async chat exports

## User Impact

No more 504 errors

### Demo

demo


snopoke (commented):
I don't think it's necessary to create a new model. My ideal scenario would be to create a temporary storage bucket in S3 with a lifecycle policy that cleans up but for now it's probably easier to just add an 'expiry' field to the `File` model.

In the task you can create the file and return the `file_id` which you can access from the view in the task result using the `task_id`:

```python
info = Progress(AsyncResult(task_id)).get_info()
if info["complete"] and info["success"]:
   file_id = info["result"]["file_id"]
```

On a separate note - have you profiled the `experiment_to_message_export_rows` method? I wonder if it's not doing additional queries in the loop. Could be worth checking as well.

SmittieC:
> I don't think it's necessary to create a new model. My ideal scenario would be to create a temporary storage bucket in S3 with a lifecycle policy that cleans up but for now it's probably easier to just add an 'expiry' field to the `File` model.
> 
> In the task you can create the file and return the `file_id` which you can access from the view in the task result using the `task_id`:
> 
> ```python
> info = Progress(AsyncResult(task_id)).get_info()
> if info["complete"] and info["success"]:
>    file_id = info["result"]["file_id"]
> ```
> 
> On a separate note - have you profiled the `experiment_to_message_export_rows` method? I wonder if it's not doing additional queries in the loop. Could be worth checking as well.

Offlined with Simon. I'm going to add an `expiry` field to the File model and use  new celery task to do some cleanup

apps/experiments/tasks.py
```diff
@@ -1,20 +1,41 @@
...
 from apps.utils.taskbadger import update_taskbadger_data
 
 
+@shared_task(bind=True, base=TaskbadgerTask)
+def async_export_chat(self, experiment_id: int, tags: list[str] = None, participant: str = None) -> dict:
+    experiment = Experiment.objects.get(id=experiment_id)
+    csv_in_memory = experiment_to_csv(experiment, tags, participant)
+    bytes_buffer = io.BytesIO(csv_in_memory.getvalue().encode("utf-8"))
+    bytes_buffer.seek(0)
+    uploaded_file = SimpleUploadedFile(name="chat_export.csv", content=bytes_buffer.read(), content_type="text/csv")
```
snopoke, line 29: It's probably simpler to use a `ContentFile` (see `File.duplicate`)

apps/experiments/tasks.py
```diff
@@ -100,7 +121,7 @@ def get_prompt_builder_response_task(team_id: int, user_id, data_dict: dict) ->
             "id": f"s{int(time.time() * 1000)}",
         }
     )
-    history_event |= {"preview": answer, "time": datetime.now().time().strftime("%H:%M")}
+    history_event |= {"preview": answer, "time": timedelta.now().time().strftime("%H:%M")}
```
snopoke, line 124: should this not be `timezone`
```suggestion
    history_event |= {"preview": answer, "time": timezone.now().time().strftime("%H:%M")}
```

apps/experiments/views/experiment.py
```diff
@@ -1078,22 +1077,32 @@ def experiment_invitations(request, team_slug: str, experiment_id: str):
...
+        file_id = info["result"]["file_id"]
+        download_url = reverse("files:base", kwargs={"team_slug": team_slug, "pk": file_id})
+        context["export_download_url"] = download_url
+    else:
+        context["task_id"] = task_id
+    return TemplateResponse(request, "experiments/components/exports.html", context)
 
 
-@login_and_team_required
-@permission_required("experiments.invite_participants", raise_exception=True)
```
snopoke, line 1096: where these removed intentionally?

templates/experiments/components/exports.html
```diff
@@ -0,0 +1,21 @@
...
+            disabled
+            hx-get="{% url 'experiments:get_export_download_link' team.slug experiment.id task_id %}"
+            hx-trigger="every 2s"
+        {% else %}
+            hx-post="{% url 'experiments:generate_chat_export' team.slug experiment.id %}"
+        {% endif %}
+        hx-swap="outerHTML"
+        hx-target="#chat-exports"
+    >
+        Generate Chat Export
```
snopoke, line 14: nit: update button text to indicate loading
```suggestion
        {% if task_id %}
        <span class="loading loading-bars loading-xs"></span> Generating
        {% else %}
        Generate Chat Export
        {% endif %}
```

templates/experiments/single_experiment_home.html
```diff
@@ -173,13 +173,7 @@ <h3 class="font-bold text-lg">Parent Experiments:
           <div x-data="{tags: '', participant: null}">
             <div class="flex gap-x-2">
               {% if perms.experiments.download_chats %}
-                <form method="post" action="{% url 'experiments:download_experiment_chats' team.slug experiment.id %}"
-                      class="my-2">
-                  {% csrf_token %}
-                  <input name="participant" class="hidden" x-bind:value="participant"/>
-                  <input name="tags" class="hidden" x-bind:value="tags"/>
```
snopoke, line 180: I don't think these will get submitted anymore

snopoke (commented):
Changes look good. Some other comments need addressing - especially the filter params.

templates/experiments/single_experiment_home.html
SmittieC: Good catch, thanks.

apps/experiments/tasks.py
SmittieC: haha interesting. Nice catch

apps/experiments/views/experiment.py
SmittieC: Whoops, nope

SmittieC:
Ok I confirmed that the tags / participant filters are submitted as well

apps/experiments/tasks.py
```diff
@@ -1,20 +1,42 @@
...
 from apps.utils.taskbadger import update_taskbadger_data
 
 
+@shared_task(bind=True, base=TaskbadgerTask)
+def async_export_chat(self, experiment_id: int, tags: list[str] = None, participant: str = None) -> dict:
+    experiment = Experiment.objects.get(id=experiment_id)
+    csv_in_memory = experiment_to_csv(experiment, tags, participant)
+    bytes_buffer = io.BytesIO(csv_in_memory.getvalue().encode("utf-8"))
+    bytes_buffer.seek(0)
+    uploaded_file = ContentFile(content=bytes_buffer.read(), name="chat_export.csv")
```
snopoke, line 29: nit: BytesIO is not needed

```suggestion
    uploaded_file = ContentFile(content=csv_in_memory.getvalue().encode("utf-8"), name="chat_export.csv")
```

snopoke (approved):

PR #915, opened 2024-11-26
temporary superuser privilege access

snopoke:
## Description
This is both a security change and a usability change for superusers.

#### On the usability side

When you attempt to access a team resource for a team you are not part of (e.g. during a support request):

**Previously**
You would get a 404 page and then you would either need to join that team or else 'impersonate' another use in order to view that page

**Now**
You can re-authenticate using your password and gain temporary access to the team (30 minutes)

#### On the security side

Accessing other team resources this way gives a clearer audit trail and does not require logging in as another user.

Access to the Django Admin site is also gated in the same way. Superusers must re-authenticate before accessing the admin site and then again every 30 minutes.

Both of these also guard against open sessions being used for privileged access. 

### Demo
<div>
    <a href="https://www.loom.com/share/daa35583b0a741189d2225dd7c1aa4b7">
      <p>Dimagi Chatbots | Temporary privilege escalation - 26 November 2024 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/daa35583b0a741189d2225dd7c1aa4b7">
      
    </a>
  </div>


snopoke:
@coderabbit resolve

apps/web/superuser_utils.py
```diff
@@ -0,0 +1,63 @@
...
+def apply_temporary_superuser_access(request, grant):
+    if not isinstance(grant, str) or not grant.strip():
+        raise ValueError("Invalid grant")
+
+    remove_expired_temporary_superuser_access(request)
+    if has_temporary_superuser_access(request, grant):
+        return
+
+    elevated_privileges = request.session.get("elevated_privileges", [])
+    if len(elevated_privileges) >= MAX_CONCURRENT_PRIVILEGES:
```
SmittieC, line 22: When we have 5 active "elevated privileges" and need to (for some reason) get a 6th one, we'd have to wait until one of the 5 expires first, yes?

SmittieC (approved):
Looks great!

PR #918, opened 2024-11-27
State refactor

SmittieC:
## Description

Prefactored the `state.py` file to use "adapters" instead (the file was also renamed to "adapaters.py"). This is in preparation for pipelines also making use of the runnables. The idea is this:
- We can chat to either a base LLM, or an assistant
- We want to be able to chat to either of these "endpoints" either using a pipeline or a normal "experiment". These are just config sources, hence you'll see the name "adapter" appear often. I'm open to change

image

## User Impact

N/A

### Demo

N/A

### Docs

N/A


apps/service_providers/llm_service/state.py
```diff
@@ -32,69 +67,128 @@ def set_chat_metadata(self, key: Chat.MetadataKeys, value):
...
 
     @property
-    def chat(self):
-        return self.session.chat
+    def citations_enabled(self):
+        return self.experiment.citations_enabled
+
+    @cached_property
+    def team(self) -> Team:
+        return self.assistant.team
```
snopoke, line 154: shouldn't this be `experiment.team`? One thought is that if we use `session.team` then it will work for all the adapter types.

apps/service_providers/llm_service/state.py
```diff
@@ -32,69 +67,128 @@ def set_chat_metadata(self, key: Chat.MetadataKeys, value):
...
+
+    @property
+    def provider_model_name(self):
+        return self.experiment.get_llm_provider_model_name()
```
snopoke, line 158: should come from the assistant if there is one

apps/service_providers/llm_service/state.py
```diff
@@ -32,69 +67,128 @@ def set_chat_metadata(self, key: Chat.MetadataKeys, value):
...
+
+    @property
+    def input_formatter(self):
+        return self.experiment.input_formatter
 
     @cache
     def get_llm_service(self):
         return self.experiment.get_llm_service()
```
snopoke, line 166: should use the assistant if there is one

apps/service_providers/llm_service/state.py
```diff
@@ -32,69 +67,128 @@ def set_chat_metadata(self, key: Chat.MetadataKeys, value):
...
+
+    def post_run_hook(self, output, config, message_metadata):
+        experiment_tag = config.get("configurable", {}).get("experiment_tag")
+        self.save_message_to_history(
+            output, type_=ChatMessageType.AI, message_metadata=message_metadata, experiment_tag=experiment_tag
+        )
+
     def get_chat_model(self):
         return self.get_llm_service().get_chat_model(
             self.experiment.get_llm_provider_model_name(), self.experiment.temperature
```
snopoke, line 183: Should use `self.provider_model_name`

apps/service_providers/llm_service/state.py
```diff
@@ -282,52 +307,46 @@ def save_message_to_history(
...
+    def citations_enabled(self):
+        return self._citations_enabled
 
-    @cache
-    def get_llm_service(self):
-        return self.assistant.llm_provider.get_llm_service()
+    @cached_property
+    def team(self) -> Team:
+        # TODO: Change
+        return self.assistant.team
```
snopoke, line 338: see comment about using the session

apps/service_providers/llm_service/state.py
```diff
@@ -282,52 +307,46 @@ def save_message_to_history(
...
+        source_material_id: int | None = None,
     ):
-        self._assistant = assistant
+        self.assistant = assistant
         self.session = session
         self.trace_service = trace_service
+        self.template_context = PromptTemplateContext(session, source_material_id)
 
-        self.input_formatter = input_formatter
+        self._input_formatter = input_formatter
```
snopoke, line 326: why change this to a private var?

apps/service_providers/llm_service/state.py
```diff
@@ -338,16 +357,98 @@ def pre_run_hook(self, input, config, message_metadata):
...
-
     def get_message_metadata(self, message_type: ChatMessageType) -> dict:
         """
         Retrieve metadata for a given message type.
         """
         return self.input_message_metadata if message_type == ChatMessageType.HUMAN else self.output_message_metadata
+
+    def get_chat_model(self):
+        return self.get_llm_service().get_chat_model(
+            self.experiment.get_llm_provider_model_name(), self.experiment.temperature
```
snopoke, line 368: same as above, use `self.provider_model_name`. Also `PipelineAdapter` doesn't have an experiment so `self.experiment.temperature` will fail

snopoke (commented):
Made a few initial comments but also wondering if the abstraction is quite right.

As you have it we've got `ExperimentAdapter` and `PipelineAdapter`, each of which could be used in `AssistantChat` or `SimpleLLMChat` (or their agent versions).

It seems like we've still got a problem of having the adapters containing info they only need sometimes.

What if we split it like this:

* ChatAdapter - constructed with the info needed for 'LLMChat'. Can be constructed from an experiment or a Pipeline Node.
* AssistantAdapter - constructed with the info needed for 'AssistantChat'. Can be constructed from an experiment or a Pipeline Node.

I could see them being constructed manually or from classmethods like `ChatAdapter.from_experiment(session, experiment)`.

SmittieC:
@snopoke I'm quite keen to get going with the node tools changes, so I'll add a runnable history manager in a future refactor. The plan is to do that after the tool changes are made

SmittieC:
> Made a few initial comments but also wondering if the abstraction is quite right.
> 
> As you have it we've got `ExperimentAdapter` and `PipelineAdapter`, each of which could be used in `AssistantChat` or `SimpleLLMChat` (or their agent versions).
> 
> It seems like we've still got a problem of having the adapters containing info they only need sometimes.
> 
> What if we split it like this:
> 
> * ChatAdapter - constructed with the info needed for 'LLMChat'. Can be constructed from an experiment or a Pipeline Node.
> * AssistantAdapter - constructed with the info needed for 'AssistantChat'. Can be constructed from an experiment or a Pipeline Node.
> 
> I could see them being constructed manually or from classmethods like `ChatAdapter.from_experiment(session, experiment)`.

Seems reasonable. I think this will work better yeah

SmittieC:
@snopoke see https://github.com/dimagi/open-chat-studio/pull/918/commits/37297226b9d2708597ce9f6a83a6fea916a34ff6. This approach does make the `adapters.py` file much shorter, thanks for the suggestion. Now on to fixing tests...

SmittieC:
@codetabbitai do a full review

SmittieC:
@coderabbitai do a full review

apps/slack/slack_datastores.py
```diff
@@ -200,12 +200,12 @@ def issue(self, team, config: dict) -> str:
...
         expire_at = timezone.now() + timezone.timedelta(seconds=self.expiration_seconds)
         # save the team on the state so that we can look it up during the callback request
-        row = SlackOAuthState(team=team, state=state, expire_at=expire_at, config=config)
+        row = SlackOAuthState(team=team, adapter=state, expire_at=expire_at, config=config)
         row.save()
         return state
 
     def consume(self, state: str) -> bool:
-        rows = SlackOAuthState.objects.filter(state=state).filter(expire_at__gte=timezone.now())
+        rows = SlackOAuthState.objects.filter(adapter=state).filter(expire_at__gte=timezone.now())
```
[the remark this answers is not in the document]
SmittieC: Thanks for catching this

apps/service_providers/llm_service/runnables.py
```diff
@@ -55,24 +51,31 @@ def create_experiment_runnable(
...
+    chat_adapter = ChatAdapter.from_experiment(experiment, session, trace_service)
     if experiment.tools_enabled and not disable_tools:
-        return AgentExperimentRunnable(state=state)
+        runnable = AgentLLMChat(adapter=chat_adapter)
 
-    return SimpleExperimentRunnable(state=state)
+    runnable = SimpleLLMChat(adapter=chat_adapter)
+    # This is a temporary hack until we return an object with metadata about the run
+    runnable.experiment = experiment
+    return runnable
```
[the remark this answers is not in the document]
SmittieC: Addressed in this commit: 834659e36ac2ba0b3083bf92f3c008062c16cd62

apps/chat/bots.py
```diff
@@ -238,13 +238,13 @@ def _get_safe_response(self, safety_layer: SafetyLayer):
             self._save_message_to_history(bot_response, ChatMessageType.AI)
             self.generator_chain = self.chain
 
-        self.generator_chain.state.ai_message.add_system_tag(
+        self.generator_chain.adapter.ai_message.add_system_tag(
             safety_layer.name, tag_category=TagCategories.SAFETY_LAYER_RESPONSE
         )
```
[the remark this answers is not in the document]
SmittieC: Addressed in this commit: 834659e36ac2ba0b3083bf92f3c008062c16cd62

SmittieC:
@coderabbitai resolve

apps/service_providers/llm_service/adapters.py
```diff
@@ -119,17 +93,176 @@ def save_message_to_history(self, message: str, type_: ChatMessageType, experime
...
+        provider_model_name: str,
+        llm_service,
+        temperature: int,
+        prompt_text: str,
+        max_token_limit: int,
+        experiment_version_number: int,
+        experiment_is_a_version: bool,
+        tools: list = None,
+        input_formatter: str | None = None,
+        source_material_id: str | None = None,
```
snopoke, line 115:
```suggestion
        source_material_id: int | None = None,
```

apps/service_providers/llm_service/adapters.py
```diff
@@ -119,17 +93,176 @@ def save_message_to_history(self, message: str, type_: ChatMessageType, experime
...
+        self.tools = tools or []
+        self.input_formatter = input_formatter
+        self.source_material_id = source_material_id
+
+        self.trace_service = trace_service
+        self.template_context = PromptTemplateContext(session, source_material_id)
+
+    @staticmethod
+    def from_experiment(experiment: Experiment, session: ExperimentSession, trace_service=None) -> Self:
+        return ChatAdapter(
```
snopoke, line 136: This should be a classmethod and use `cls` as the constructor (https://typing.readthedocs.io/en/latest/spec/generics.html#use-in-classmethod-signatures):

```suggestion
    @classmethod
    def from_experiment(cls, experiment: Experiment, session: ExperimentSession, trace_service=None) -> Self:
        return cls(
```

Same for the other ones

apps/service_providers/llm_service/adapters.py
```diff
@@ -119,17 +93,176 @@ def save_message_to_history(self, message: str, type_: ChatMessageType, experime
...
+        self.citations_enabled = citations_enabled
+        self.experiment_version_number = experiment_version_number
+        self.experiment_is_a_version = experiment_is_a_version
+        self.tools = get_assistant_tools(assistant, experiment_session=session)
+        self.input_formatter = input_formatter
+        self.trace_service = trace_service
+        self.save_message_metadata_only = save_message_metadata_only
+        self.input_message_metadata = {}
+        self.output_message_metadata = {}
+        self.template_context = PromptTemplateContext(session, source_material_id=None)
```
snopoke, line 205: nit: (This is purely a stylistic thing but it does help with readability and cognitive overhead.)

Values coming from the method args should go above values that are computed. Also, where possible keep the ordering consistent between the args and the property assignment:

```python
def __init__(self, a, b, c):
    self.a = a
    self.b = b
    self.c = c
    self.service = a.get_service()
```
```

File: apps/service_providers/llm_service/adapters.py [hunk 14]
```diff
@@ -119,17 +93,176 @@ def save_message_to_history(self, message: str, type_: ChatMessageType, experime
...
+            assistant=experiment.assistant,
+            citations_enabled=experiment.citations_enabled,
+            experiment_version_number=experiment.version_number,
+            experiment_is_a_version=experiment.is_a_version,
+            input_formatter=experiment.input_formatter,
+            trace_service=trace_service,
+        )
+
+    @staticmethod
+    def from_pipeline(session: ExperimentSession, node: "AssistantNode") -> Self:
```
snopoke, line 220: Should this be called `from_pipeline_node`?

Alternatively, we could rename them `for_experiment` and `for_pipeline`

apps/service_providers/llm_service/adapters.py
```diff
@@ -160,11 +293,7 @@ def get_file_type_info(self, attachments: list) -> list:
         return file_type_info
 
     def get_attachments(self, attachment_type: str):
```
snopoke, line 295:
```suggestion
    def get_attachments(self, attachment_type: list[str]):
```

apps/service_providers/llm_service/adapters.py
```diff
@@ -194,158 +323,11 @@ def get_messages_to_sync_to_thread(self):
 
     def get_metadata(self, key: Chat.MetadataKeys):
         """Chat metadata"""
-        return self.chat.get_metadata(key)
+        return self.session.chat.get_metadata(key)
```
snopoke, line 326: I think this is only used for getting the thread ID - we could rename it to `get_thread_id` or make it a property `thread_id`:

```python
    @property
    def thread_id(self):
        return self.session.chat.get_metadata(Chat.MetadataKeys.OPENAI_THREAD_ID)

    @thread_id.setter
    def set_thread_id(self, value):
        key = Chat.MetadataKeys.OPENAI_THREAD_ID
        if self.trace_service:
            self.trace_service.update_trace({key: value})
        self.session.chat.set_metadata(key, value)
```

snopoke (dismissed):
Looks great. A few small comments but nothing major.

apps/chat/tests/test_topic_bot.py
```diff
@@ -25,17 +24,18 @@ def test_safety_response(is_safe_mock):
     experiment.safety_layers.add(layer)
 
     expected = "Sorry I can't help with that."
+    experiment.get_llm_service = lambda: build_fake_llm_service([expected], token_counts=[1])
+
     bot = TopicBot(session)
     with patch.object(TopicBot, "_get_safe_response", wraps=bot._get_safe_response) as mock_get_safe_response:
-        with mock_experiment_llm(experiment, responses=[expected]):
```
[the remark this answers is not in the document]
SmittieC: No. The patch doesn't patch anymore. I found this interesting, since this test and the one below it is pretty much the same. I am not 100% sure why yet, but didn't want to spend too much time on figuring it out atm

apps/service_providers/llm_service/adapters.py
SmittieC: I like `for_experiment`. Let's go with that

SmittieC: https://github.com/dimagi/open-chat-studio/pull/918/commits/930a80a13eb58f0919231e398cf1a51d0f39fcd0

apps/service_providers/llm_service/adapters.py
SmittieC: Good idea. I'll update

apps/service_providers/llm_service/runnables.py
```diff
@@ -122,7 +128,7 @@ def invoke(self, input: str, config: RunnableConfig | None = None, *args, **kwar
             self._populate_memory(input)
 
         if save_input_to_history:
-            self.state.save_message_to_history(input, ChatMessageType.HUMAN)
+            self.adapter.save_message_to_history(input, ChatMessageType.HUMAN)
```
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/commit/d44f1eab3dac35a0a5f5d2f283231aea7021e1c0

apps/service_providers/llm_service/adapters.py
snopoke: I think this should work:

```python
setter_mock = Mock()
mock_property = ChatAdapter.thread_id.setter(setter_mock)
with patch.object(ChatAdapter, 'thread_id', mock_property):
    adapter.thread_id = 1
    setter_mock.assert_called_with(adapter, 1)
```

apps/service_providers/llm_service/adapters.py
```diff
@@ -0,0 +1,339 @@
...
+    def assistant_tools_enabled(self):
+        return self.assistant.tools_enabled
+
+    @property
+    def assistant_builtin_tools(self) -> list:
+        return self.assistant.builtin_tools
+
+    @property
+    def thread_id(self):
+        return self.session.chat.thread_id
```
snopoke, line 265: I think this is causing the test failures:

`AttributeError: 'Chat' object has no attribute 'thread_id'`

apps/service_providers/llm_service/adapters.py
SmittieC: Hmm, that didn't work either. I might have done it wrong, I don't know. I went with [this approach](https://github.com/dimagi/open-chat-studio/pull/918/commits/64c8818aea6cba386689298005c14d48929be11f) rather to save time, which is suboptimal, since I'm writing code to accomodate a test, but at least we're testing that the `thread_id` was called again. Feel free to take a swing at it if you'd like.

snopoke (approved):

PR #922, opened 2024-11-27
Add Archived Experiment to Experiments Table

stephherbers:
## Description

Adds archived experiments to experiments table, that is not navigable and have no actions. Future work that already has a ticket is to add the ability to navigate to a read-only view of the experiment home page.
Screenshot 2024-12-04 at 10 04 06 AM

## User Impact

Allows user to view archived experiments

### Demo

[loom video](https://www.loom.com/share/dd768eba8f2043c1a56d8990743553d4?sid=a4a247b0-d41a-4b6d-aa39-95d139e30a6d)

### Docs

coming soon


snopoke (commented):
Do we want the archived experiments always shown in the table or should we add a toggle to display them?

templates/generic/object_home_content.html
```diff
@@ -35,9 +36,37 @@ <h1 class="{{ title_class|default:"pg-title" }}">{{ title }}</h1>
...
+      </div>
+      <div class="my-2" div>
+        <label for="show_archived" class="label cursor-pointer flex items-center space-x-2">
+          <span class="label-text mr-2">Show Archived</span>
+          <input
+            type="checkbox"
+            name="show_archived"
+            class="toggle"
+            {% if show_archived == "true" %}checked{% endif %}
+            hx-get="{{ table_url }}{% querystring %}"
```
snopoke, line 59: does this work with the search input?

stephherbers: https://github.com/dimagi/open-chat-studio/pull/922/commits/dec229f83ad303ceea27972aed6508c3190e38a7

stephherbers: even better https://github.com/dimagi/open-chat-studio/pull/922#:~:text=%E2%80%A6-,aeb8a1c,-Merge%20state and updated the loom video

apps/experiments/models.py
```diff
@@ -748,6 +748,23 @@ def compare_with_latest(self):
...
+
+    def delete_experiment_channels(self):
+        from apps.channels.models import ExperimentChannel
+
+        for channel in ExperimentChannel.objects.filter(experiment_id=self.id):
+            channel.soft_delete()
+
+    def archive_working_experiment(self):
+        if self.is_working_version:
+            if self.has_versions:
```
snopoke, line 763: Probably don't need this check since we're going to do the full query anyway.

apps/experiments/models.py
```diff
@@ -748,6 +748,23 @@ def compare_with_latest(self):
...
+                for version in self.versions.all():
+                    version.archive()
```
snopoke, line 765: I thought we decided we could do this with a query because we don't need to `delete_experiment_channels` for versions (since the channels are only attached to the working version.

stephherbers:
kept the test, it's a bit overkill, but I thought it good to keep in case something breaks that functionality in the future

snopoke (approved):

Issue #638, opened 2024-08-23
Add an entry point node to pipelines v2


bderenzi:
This is purely a UI/UX thing and doesn't need to be saved. We should have a node that is pre-populated on the canvas for the entry point. We should not be able to delete the entry point. Maybe it should auto-connect to the input side of the first thing we drag in? (seems convenient). 

I think it'll just help the user see where things start.

PR #929, opened 2024-11-28
Add required start and end nodes and allow chatting to subset of pipelines

proteusvacuum:
## Description
This adds the requirement that there is always a single `Start` and `End` node, and that there is an accessible path between them. 

When building up the pipeline graph, we can now traverse from the start node to the end node, so testing subgraphs of the pipeline in the pipeline builder is now possible by moving where the start and end points to. 

This also adds automatically incrementing "New Pipeline" names, and adds a migration to create start and end nodes for existing pipelines. 

I also did a refactor of the pipeline tests to make them less verbose and make things a bit easier to read.

## User Impact

Users will be required to have start and end nodes in their pipelines which clarifies the flow of data.

### Demo

https://github.com/user-attachments/assets/91bce7f3-0070-4356-9521-2d0229896db1

Fixes: https://github.com/dimagi/open-chat-studio/issues/638


bderenzi:
What happened at 1:04 / 1:05 in the video where it partially ran through the top path and partially through the bottom path?

proteusvacuum:
@bderenzi The change I made hadn't saved yet, so it was using the previous pipeline definition - I think we should "block" until the save has completed. I'll do that as part of a separate PR.

proteusvacuum:
In another separate followup, it might be worth also preventing cycles from being made on the frontend: https://reactflow.dev/examples/interaction/prevent-cycles

apps/pipelines/models.py
```diff
@@ -67,6 +68,37 @@ def version_display(self) -> str:
...
+    @classmethod
+    def create_default(cls, team):
+        from apps.pipelines.nodes.nodes import EndNode, StartNode
+
+        default_name = "New Pipeline"
+        existing_pipeline_count = cls.objects.filter(team=team, name__startswith=default_name).count()
+
+        start_node = {
+            "id": str(uuid4()),
+            "type": "pipelineNode",
```
snopoke, line 80: What about changing this type to something else and having a completely different rect component for the start & end nodes? We could do away with the 'ui:show_input" and "ui:show_output" schema customizations and we would have easier control over the style of those nodes e.g. we could make them round, not show the toolbar etc.

apps/pipelines/models.py
```diff
@@ -67,12 +68,43 @@ def version_display(self) -> str:
...
+            name=f"New Pipeline {existing_pipeline_count + 1}",
+        )
+        new_pipeline.set_nodes()
+        return new_pipeline
+
     def get_absolute_url(self):
         return reverse("pipelines:details", args=[self.team.slug, self.id])
 
-    def set_nodes(self, nodes: list[FlowNode]) -> None:
+    def set_nodes(self) -> None:
```
snopoke, line 105: nit: what about renaming this to "update_nodes_from_data"

snopoke (commented):
This is great. Thanks for the refactor and other changes. I think my main question is whether we should use a different node type for the start / end nodes (along with a different react component).

apps/pipelines/models.py
proteusvacuum: I updated this to use `startNode` and `endNode`, with different styling for each. 

image

apps/pipelines/models.py
proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/929/commits/de8400045bcc58a2ae821e28e8ee296a179cb991

snopoke (dismissed):

snopoke (approved):

PR #930, opened 2024-11-29
Add pytest-xdist

proteusvacuum:
Running the full test suite locally with `pytest -n auto` (with 16 workers) runs in 80 seconds. Wanted to see what the speedup was on GHA.

### DIFF
--- a/apps/events/tests/test_timeout_trigger.py
+++ b/apps/events/tests/test_timeout_trigger.py
@@ -275,7 +275,7 @@ def test_new_human_message_resets_count(session):
 @pytest.mark.django_db()
 @pytest.mark.parametrize(
     ("status", "matches"),
-    {
+    [
         (SessionStatus.SETUP, True),
         (SessionStatus.PENDING, True),
         (SessionStatus.PENDING_PRE_SURVEY, True),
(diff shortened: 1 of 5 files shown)


apps/events/tests/test_timeout_trigger.py
```diff
@@ -275,15 +275,15 @@ def test_new_human_message_resets_count(session):
 @pytest.mark.django_db()
 @pytest.mark.parametrize(
     ("status", "matches"),
-    {
```
proteusvacuum, line 278: This change was required since the ordering of the set can change between workers. xdist requires the exact same test suite to be generated between workers so it can choose which ones to pick.

proteusvacuum:
Validated that it seems to have run the same number of tests:
773 items on this PR
And [this run](https://github.com/dimagi/open-chat-studio/actions/runs/12070120831/job/33658897635), on the previous master commit also has 773 items

Time taken:
previous: 9m10s
with x-dist: 5m49
So, about 37% speedup on github actions. 

Depending on your home machine, this could have a much bigger decrease.

We could always add verbose output (`-v`) if we feel we're missing the filenames.

apps/teams/tests/test_permissions.py
```diff
@@ -65,6 +65,7 @@ def test_content_types():
     "template_partials",
     "tz_detect",
     "users",
+    "utils_tests",
```
proteusvacuum, line 68: I'm not 100% certain why this only started failing here, or if this is valid...

There is indeed an app with this label: 

https://github.com/dimagi/open-chat-studio/blob/45c09ef80c7dae4f3b5da7cdc567ddaa4552d6b5/apps/utils/tests/conftest.py#L5

snopoke (approved):

PR #931, opened 2024-11-29
Click to add a component to pipelines

proteusvacuum:
## Description

Adds the ability to click on a component to add it to the canvas.

### Demo

https://github.com/user-attachments/assets/6340ffe8-d3b0-46bd-9eca-90fd2705bbbf


bderenzi:
This looks great. Can we prevent it from placing a new node on top of an existing node? I think in the demo, the "conditional node" is given the same top,left coordinates as LLM Response and covers it.

snopoke (approved):

PR #948, opened 2024-12-02
Fix submit button

SmittieC:
## Description

The submit was disabled the whole time unless you used an assistant. I removed the assistant checks and only watch the tool elements when their refs exists


snopoke (commented):
Thanks

PR #949, opened 2024-12-02
copy files memory without blowing up memory

snopoke:
I haven't done any profiling but I think this will help

### DIFF
--- a/apps/assistants/models.py
+++ b/apps/assistants/models.py
@@ -112,7 +112,7 @@ class OpenAiAssistant(BaseTeamModel, VersionsMixin):
         assistant_version.name = f"{self.name} v{version_number}"
         assistant_version.assistant_id = ""
         assistant_version.save()
-        file_ids = duplicate_files(self.files.iterator())
+        file_ids = duplicate_files(self.files.iterator(chunk_size=50))
         if file_ids:
             OpenAiAssistant.files.through.objects.bulk_create(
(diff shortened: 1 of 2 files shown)


apps/assistants/models.py
```diff
@@ -111,22 +112,33 @@ def create_new_version(self, *args, **kwargs):
         assistant_version.name = f"{self.name} v{version_number}"
         assistant_version.assistant_id = ""
         assistant_version.save()
-        files = [file.duplicate() for file in self.files.all()]
-        assistant_version.files.set(files)
+        file_ids = duplicate_files(self.files.iterator())
```
SmittieC, line 115: [Django will default](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#django.db.models.query.QuerySet.iterator:~:text=So%20long%20as%20the%20QuerySet%20does%20not%20prefetch%20any%20related%20objects%2C%20providing%20no%20value%20for%20chunk_size%20will%20result%20in%20Django%20using%20an%20implicit%20default%20of%202000%2C%20a%20value%20derived%20from%20a%20calculation%20on%20the%20psycopg%20mailing%20list%3A) to a `chunk_size` of 2000 rows. Maybe we should make this smaller. Is 20 too small?

SmittieC (approved):
Looks good to me. One question on the chunk size

PR #953, opened 2024-12-03
allow tools to return artifacts

snopoke:
## Description
This modifies how agents are run when OpenAI Assistants are used. The reason for the change is to support tools that return artifacts (responses too large to send as tool results or not text).

The basic AgentLLMChat runnables ignore the artifacts but the AgentAssistantChat gives them special treatment. Instead of submitting the tool outputs to the thread run, a new run is created with a message containing the artefact as an attachment.

## User Impact
Hopefully this will allow users to get useful information from larger amounts of data.

### Demo
[Screencast from 2024-12-05 20-35-44.webm](https://github.com/user-attachments/assets/7d536a0d-8cd6-4c49-8228-f8d46d4bd4b4)

### Docs
Updated internal readme


snopoke:
@coderabbit review

snopoke:
@coderabbitai full review

stephherbers (approved):
no glaring concerns from my end
One question: in the demo the artifact is queried, but the goal is for the user to be able to download the file, no?

snopoke:
> no glaring concerns from my end One question: in the demo the artifact is queried, but the goal is for the user to be able to download the file, no?

I don't think the goal is for the user to download the file - I think the goal is for the Assistant to be able to run some analysis on the data.

I did originally save the file to OCS but didn't have a good way to link it to the message. If that use case comes up I'll revisit it.

@bderenzi FYI

PR #955, opened 2024-12-04
Fix: Chunk with chunksize=500 for batch API

SmittieC:
A followup on [this PR](https://github.com/dimagi/open-chat-studio/pull/954). Seems like the batch API has a 500 file_id limit


snopoke (commented):
should this be applied to the other uses of `file_batches`?

SmittieC:
Probably yes. Forgot about that, thanks

PR #961, opened 2024-12-05
Tagging improvements

### LINKED ISSUE #928: Tag UI improvements
* Show who added a tag via a tooltip
* Differentiate between tags I added vs tags added by others (e.g. by color?)
* Use arrow keys to navigate between tags in tag widget
  * https://tom-select.js.org/plugins/caret-position/
* ‘x’ on tag badge to remove it
  * https://tom-select.js.org/plugins/remove-button/

SmittieC:
@resolves #928 

The linked ticket pretty much describes it

Question: Do you find the green badges awkward?

### Demo
demo

### DIFF
--- a/apps/annotations/models.py
+++ b/apps/annotations/models.py
@@ -107,10 +107,6 @@ class TaggedModelMixin(models.Model, AnnotationMixin):
         return [tag["name"] for tag in self.tags_json]
 
     @cached_property
-    def content_type(self) -> ContentType:
-        return
-
-    @cached_property
     def tags_json(self):
         tagged_items = CustomTaggedItem.objects.filter(
             content_type__model=self._meta.model_name, content_type__app_label=self._meta.app_label, object_id=self.id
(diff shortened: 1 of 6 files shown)


apps/annotations/models.py
```diff
@@ -106,11 +106,24 @@ def system_tags_names(self):
     def all_tag_names(self):
         return [tag["name"] for tag in self.tags_json]
 
+    @cached_property
+    def content_type(self) -> ContentType:
```
snopoke, line 110: what's this for?

snopoke (commented):
I like the green but I having the others neutral makes them disappear.

What about using a box shadow to highlight 'mine'? (https://tailwindcss.com/docs/ring-width)

apps/annotations/models.py
SmittieC: I thought I removed this. I used it before but realized it's not necessary

snopoke (approved):

PR #979, opened 2024-12-12
Pipeline trace provider

### LINKED ISSUE #972: Tracing service configuration for pipelines

SmittieC:
Resolves #972 

## Description

- Ability to configure a tracing provider for a pipeline. I confirmed this works for both langsmith and langfuse. It tracks all node invocations as well.
- I noticed that we don't update the consent form's `is_default` field to False when creating a version of it, thus creating duplicate multiple consent forms in the team that could be the default. I added a migration to set all versioned consent forms to `is_default=False` and add a DB constraint to prevent future mishaps.

## User Impact

Tracing for pipelines (and its nodes)

### Demo

N/A

### Docs

Will update the release nodes when merged


apps/experiments/models.py
```diff
@@ -714,7 +728,7 @@ def create_new_version(self, version_description: str | None = None, make_defaul
         new_version.version_number = version_number
 
         self._copy_attr_to_new_version("source_material", new_version)
-        self._copy_attr_to_new_version("consent_form", new_version)
+        self._copy_attr_to_new_version("consent_form", new_version, update_fields={"is_default": False})
```
snopoke, line 731: Since this is only used in one place what about just overriding 'create_new_version' in the consent form model to set it to False?

snopoke (approved):
One comment but not blocking.

apps/experiments/models.py
SmittieC: Could do!

SmittieC: https://github.com/dimagi/open-chat-studio/pull/979/commits/67d90cb8860031cbc62c7a1c66f9d7ea60e681bc

PR #980, opened 2024-12-12
Assistant versioning updates

snopoke:
## Description
The primary change here is to stop copying files when creating new versions of an assistant. Instead we just copy references to the file. In OpenAI this means that the new vectorstore will get created using existing files instead of new files and we don't have to re-upload all the files.

I did test this locally with an OpenAI assistant as it does work.

The remainder of the changes are cleanups / updates to version comparison logic (review by commit).

## User Impact
This should greatly decrease the amount of time it takes to create a new version.


apps/assistants/models.py
```diff
@@ -64,8 +64,6 @@ class OpenAiAssistant(BaseTeamModel, VersionsMixin):
     is_archived = models.BooleanField(default=False)
     tools = ArrayField(models.CharField(max_length=128), default=list, blank=True)
 
-    files = models.ManyToManyField("files.File", blank=True)
```
snopoke, line 67: This is a legacy field which is no longer used. I checked on prod and there are no assistants with any files linked through this field.

apps/assistants/sync.py
```diff
@@ -105,7 +105,7 @@ def push_assistant_to_openai(assistant: OpenAiAssistant, internal_tools: list |
     data["tool_resources"] = _sync_tool_resources(assistant)
 
     if internal_tools:
-        data["tools"] = [convert_to_openai_tool(tool) for tool in internal_tools]
+        data["tools"].extend([convert_to_openai_tool(tool) for tool in internal_tools])
```
snopoke, line 108: This bug has been around for ages and likely contributes to some of the weird assistant behaviour we've seen when tools are used in conjunction with builtin tools.

SmittieC (dismissed):
Looks good to me, thanks!

PR #982, opened 2024-12-12
Fix "unique_global_channel_per_team" error in temporary session

proteusvacuum:
There is a constraint that there is a single `WEB` channel for every team. If you have a previously created session with a web channel, then you were not able to use `simple_invoke` as it attempted to make a new channel. 

I'm not sure how re-using a channel that already exists affects the rollback, but it "fixes" the bug.


proteusvacuum:
I see this was fixed here: https://github.com/dimagi/open-chat-studio/commit/5f558e8722aabf68f829ccf7a2596431d507d7b9

PR #992, opened 2024-12-13
Add CodeNode

proteusvacuum:
## Description

Runs arbitrary python within the pipeline.
It essentially is running a function:
```python
def main(input):
    {The Code Goes Here}
```
The return value of the function is used as the output.

The main reason I encapsulated it inside of a separate function is that you can then:
- define other functions e.g.
```python
def foo(bar):
    return bar + 1

return foo(3)
```

- clearly see what is being returned with a `return` statement

I still need to add: a code widget (is there a preference? I was thinking of using `ace` as I think that is what is used in HQ)

For now, this doesn't access "shared state" that we had also spoken about, but I wanted to get this out there for discussion first.

### Demo

https://github.com/user-attachments/assets/869d7afa-c49a-4cd6-92c9-b72660597c44

### DIFF
--- a/apps/pipelines/nodes/nodes.py
+++ b/apps/pipelines/nodes/nodes.py
@@ -633,7 +633,7 @@ class CodeNode(PipelineNode):
     model_config = ConfigDict(json_schema_extra=NodeSchema(label="Python Node"))
     code: str = Field(
         description="The code to run",
-        json_schema_extra=UiSchema(widget=Widgets.expandable_text),  # TODO: add a code widget
+        json_schema_extra=UiSchema(widget=Widgets.code),
     )
 
     @field_validator("code")
--- a/apps/pipelines/nodes/nodes.py
+++ b/apps/pipelines/nodes/nodes.py
@@ -1,4 +1,6 @@
+import datetime
 import json
+import time
 from typing import Literal
 
 import tiktoken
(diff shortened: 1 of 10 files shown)


snopoke (commented):
This is great. 

In terms of an editor I've been wanting to get going with https://codemirror.net/

proteusvacuum:
https://github.com/user-attachments/assets/8c4ca68d-3d10-4ebf-b2f7-c084bdb9e390

proteusvacuum:
Currently the python is _extremely_ restricted. As we overwrite `__builtins__` we need to redefine things like `_inplacevar_` in a safe way, like this: 

https://github.com/zopefoundation/AccessControl/blob/f9ae58816f0712eb6ea97459b4ccafbf4662d9db/src/AccessControl/ZopeGuards.py#L625

There are other potentially safe functions that we could add, e.g:
https://github.com/ndurner/amz_bedrock_chat/blob/830e496bcc0e6fa8e8b6e361c5419dd53d5d72e6/code_exec.py#L41-L99

apps/pipelines/tests/test_code_node.py
```diff
@@ -0,0 +1,88 @@
...
+
+@django_db_with_data(available_apps=("apps.service_providers",))
+@mock.patch("apps.pipelines.nodes.base.PipelineNode.logger", mock.Mock())
+@pytest.mark.parametrize(
+    ("code", "input", "output"),
+    [
+        ("return f'Hello, {input}!'", "World", "Hello, World!"),
+        ("", "foo", "foo"),  # No code just returns the input
+        (EXTRA_FUNCTION, "blah", "other blah"),  # Calling a separate function is possible
+        ("'foo'", "", "None"),  # No return value will return "None"
```
proteusvacuum, line 47: I'm wondering if we should instead raise an exception in this case, as it was probably not intentional to not return a string.

proteusvacuum:
Made a few updates after @bderenzi gave some feedback on slack. 

Now the `main` function is exposed to the user, and there is a bit more information provided for syntax errors, etc.


https://github.com/user-attachments/assets/8e028650-b8c4-48b3-8be2-bc579f941251

apps/pipelines/nodes/nodes.py
```diff
@@ -622,3 +626,118 @@ def _get_assistant_runnable(self, assistant: OpenAiAssistant, session: Experimen
...
+    def _get_custom_builtins(self):
+        allowed_modules = {
+            "json",
+            "re",
+            "datetime",
+            "time",
+        }
+        custom_builtins = safe_builtins.copy()
+        custom_builtins.update(
+            {
```
SmittieC, line 726: (not for this PR) It would be nice to have some way of telling/showing the user which builtin modules and imports are available. Maybe we can utilize codemirror's [autocomplete](https://codemirror.net/examples/autocompletion/)

apps/pipelines/tests/test_code_node.py
SmittieC: I think it makes sense to raise. How easy would it be to detect this when saving?

PR #995, opened 2024-12-16
Fix: Use correct versioned bot for scheduled messages

### LINKED ISSUE #987: Verify that events are using the correct experiment version
e.g. should this be getting the published version: https://github.com/dimagi/open-chat-studio/blob/f598270b25824597a3d4381aea4d31454456fd77/apps/events/models.py#L360

SmittieC:
Resolves #987 

## Description

Scheduled messages always point to a working experiment. When one triggers, we need to use the deployed version to generate the response. In a router setup, it can become a bit more complicated. If no experiment is specified to generate the response (see [this form](https://github.com/dimagi/open-chat-studio/blob/e00b60ba0171bb5dd1a830d95a36081a2a7b2ee9/apps/events/forms.py#L79-L81)), then we simply use the router itself. If a child bot is specified, it is the working version of said child bot.

Suppose the router bot is versioned and some version is deployed, then we can't simply use the working version of the child to generate the response, we have to use the correct version of the child bot for that router version. That's why we now filter the versioned router's child bots to find the one whose `working_version_id` is the `experiment_id` in the action params.

## User Impact

The correct bot will be used to generate responses

### Demo

The test should be enough

### Docs

N/A

### DIFF
--- a/apps/events/models.py
+++ b/apps/events/models.py
@@ -352,14 +352,13 @@ class ScheduledMessage(BaseTeamModel):
             logger.exception(f"An error occured while trying to send scheduled messsage {self.id}. Error: {e}")
 
     def _trigger(self):
-        experiment_id = self.params.get("experiment_id", self.experiment.id)
         experiment_session = self.participant.get_latest_session(experiment=self.experiment)
         if not experiment_session:
             # Schedules probably created by the API
             return
-        experiment_to_use = Experiment.objects.get(id=experiment_id)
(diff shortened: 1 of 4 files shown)


SmittieC:
I just realized that we don't update the `experiment_id` in the params of the action that is linked to existing scheduled messages whenever the generating experiment is changed between versions

stephherbers (approved):

PR #997, opened 2024-12-16
Fix: Use the `version` instances to compare router children

SmittieC:
## Description

Comparing router setups were not very comprehensive, because we limited the number of fields to compare on child experiments. This causes us to miss changes in child bot assistants. The change I made here is basically to use the `version` method on the child experiments to compare it to each other. This is very comprehensive and if we ever change the definition of an experiment version i.e. add or remove a field, then it would automatically apply when comparing router children as well. One slight concern I have with doing this is the memory footprint for large assistants, but let's see how it goes. I do think this is the right approach.

## User Impact

Fixes the issue described above

### Demo

N/A

### Docs

N/A


SmittieC:
Merging since this fix is needed

PR #999, opened 2024-12-17
Archive pipeline versions when archiving experiment versions

### LINKED ISSUE #988: Archive pipeline versions when archiving experiment versions

SmittieC:
resolves #988 

## Description

- Archiving the pipeline working version also archives its versions
- Prevent users from archiving a pipeline when there are still objects referencing it + show these to the user

## User Impact

Better UI/UX

### Demo

demo

### Docs

N/A

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -769,13 +769,22 @@ class Experiment(BaseTeamModel, VersionsMixin):
 
     @transaction.atomic()
     def archive(self):
+        """
+        Archive the experiment and all versions in the case where this is the working version. The linked assistant and
+        pipeline for the working version should not be archived.
+        """
         super().archive()
+        self.static_triggers.update(is_archived=True)
+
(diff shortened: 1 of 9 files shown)


apps/experiments/models.py
```diff
@@ -769,13 +769,22 @@ def compare_with_latest(self):
...
         if self.is_working_version:
             self.delete_experiment_channels()
             self.versions.update(is_archived=True, audit_action=AuditAction.AUDIT)
             self.scheduled_messages.all().delete()
-        elif self.assistant:
-            self.assistant.archive()
+        else:
+            if self.assistant:
+                self.assistant.archive()
+            elif self.pipeline:
```
stephherbers, line 786: is this an elif of an if?

SmittieC: Correct, the nested `if`

stephherbers (approved):

PR #1000, opened 2024-12-17
Check imported assistant instructions for illegal variables

SmittieC:
## Description

We now also check the assistant instructions for illegal variables for imported assistants

## User Impact

Users will not be able to use non-supported variables in the prompts. This is good, since their assistant will not break when they try to chat to it anymore.

### Demo

Example: this will show when you try and import an invalid one:

Screenshot from 2024-12-17 13-39-34


SmittieC:
@coderabbitai full review please

stephherbers (approved):

PR #1004, opened 2024-12-18
👍👎 for UI feedback

### LINKED ISSUE #927: Feedback in chat UI
Add message level 👍👎 and comment form in the web chat.

The 👍👎 will add tags to the message with those emojis.

SmittieC:
Resolves #927 

## User Impact

All participants can give 👍👎 feedback in the UI

### Demo

demo

### DIFF
--- a/apps/chat/models.py
+++ b/apps/chat/models.py
@@ -219,8 +219,8 @@ class ChatMessage(BaseModel, TaggedModelMixin, UserCommentsMixin):
         self.add_tag(tag, team=self.chat.team, added_by=None)
 
     def rating(self) -> str | None:
-        if rating := self.tags.filter(category=TagCategories.RESPONSE_RATING).first():
-            return rating.name
+        if rating := self.tags.filter(category=TagCategories.RESPONSE_RATING).values_list("name", flat=True).first():
+            return rating
 
     def get_processor_bot_tag_name(self) -> str | None:
         """Returns the tag of the bot that generated this message"""
--- a/apps/chat/models.py
+++ b/apps/chat/models.py
@@ -209,6 +209,19 @@ class ChatMessage(BaseModel, TaggedModelMixin, UserCommentsMixin):
             tag = f"{tag}-unreleased"
         self.add_system_tag(tag=tag, tag_category=TagCategories.EXPERIMENT_VERSION)
 
+    def add_rating(self, tag: str):
+        tag, _ = Tag.objects.get_or_create(
+            name=tag,
+            team=self.chat.team,
+            is_system_tag=False,
+            category=TagCategories.RESPONSE_RATING,
+        )
+        self.add_tag(tag, team=self.chat.team, added_by=None)
+
+    def rating(self) -> str | None:
+        if rating := self.tags.filter(category=TagCategories.RESPONSE_RATING).first():
+            return rating.name
+
     def get_processor_bot_tag_name(self) -> str | None:
         """Returns the tag of the bot that generated this message"""
         if self.message_type != ChatMessageType.AI:
--- a/apps/experiments/views/chat.py
+++ b/apps/experiments/views/chat.py
@@ -1,0 +1,22 @@
+from django.http import Http404
+from django.shortcuts import get_object_or_404, render
+
+from apps.chat.models import ChatMessage, ChatMessageType
+
+
+def rate_message(request, team_slug: str, message_id: int, rating: str):
+    if rating not in ["👍", "👎"]:
+        raise Http404()
+
+    message = get_object_or_404(ChatMessage, id=message_id, message_type=ChatMessageType.AI)
+    # TODO
+    # message.add_rating(rating)
+
+    return render(
+        request,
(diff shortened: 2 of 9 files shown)


templates/experiments/chat/components/message_rating.html
```diff
@@ -0,0 +1,20 @@
+<div class="flex flex-row gap-1" id="rating-{{message.id}}">
+    {% if message.rating %}
+        <div class="text-sm">{{ message.rating }}</div>
+    {% else %}
+        <button
+            hx-post="{% url 'experiments:rate_message' team_slug=team_slug message_id=message.id rating='👍' %}"
+            hx-swap="outerHTML"
+            hx-target="#rating-{{message.id}}"
+            class="text-sm p-1 rounded-md hover:bg-gray-50">
+            👍
```
stephherbers, line 10: add in a title to indicate what the icon means-- in chatgpt they have "good response" which I think it pretty clear

templates/experiments/chat/components/message_rating.html
```diff
@@ -0,0 +1,20 @@
...
+        </button>
+        <button
+            hx-post="{% url 'experiments:rate_message' team_slug=team_slug message_id=message.id rating='👎' %}"
+            hx-swap="outerHTML"
+            hx-target="#rating-{{message.id}}"
+            class="text-sm p-1 rounded-md hover:bg-gray-50">
+            👎
+        </button>
```
stephherbers, line 18: same here but for "bad response"

apps/experiments/views/chat.py
```diff
@@ -0,0 +1,21 @@
+from django.http import Http404
+from django.shortcuts import get_object_or_404, render
+
+from apps.chat.models import ChatMessage, ChatMessageType
+
+
+def rate_message(request, team_slug: str, message_id: int, rating: str):
+    if rating not in ["👍", "👎"]:
```
stephherbers, line 8: will we have any issues with these emojis being updated/modified?

apps/chat/models.py
```diff
@@ -209,6 +209,19 @@ def add_version_tag(self, version_number: int, is_a_version: bool):
...
         self.add_system_tag(tag=tag, tag_category=TagCategories.EXPERIMENT_VERSION)
 
+    def add_rating(self, tag: str):
+        tag, _ = Tag.objects.get_or_create(
+            name=tag,
+            team=self.chat.team,
+            is_system_tag=False,
+            category=TagCategories.RESPONSE_RATING,
+        )
+        self.add_tag(tag, team=self.chat.team, added_by=None)
```
stephherbers, line 219: I like having the participant being anonymous, but then having the added_by=None feels not intuitive along with some of the filtering-- I'm curious if rather than a category if you considered going the route of a `is_system_tag` that would also open the possibly of more participant guided tags in the future (which I _think_ was discussed vaguely but perhaps not)-- either approach works just bringing up the idea

stephherbers (changes requested):
looks great-- only blocking comment is the clarifying text for the icons

templates/experiments/chat/components/message_rating.html
SmittieC: Good idea, I'll add

templates/experiments/chat/components/message_rating.html
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1004/commits/85d5c5d498a3a2bf97bae7c489300d02d024b17b

apps/chat/models.py
SmittieC: Good question. I did consider the system tag route, but I think we should keep it reserved for case where the server created and added the tag, with no human intervention. I agree that we can and should improve the meaning of system tag vs user tag vs participant tag. I think let's keep things as-is for now until we have a clearer solution to this. A mini spec might do.

There are two fundamental questions to answer:

1. Who created this tag?
2. Who added this tag to this message?

For the most part question 2 is the most interesting one for users I'd imagine. As for determining this, we could do it either at the application layer or DB layer. In the latter case, we can have a flag or something on the `CustomTaggedItem` model to indicate which entity added this to the message i.e. user/system/participant.

SmittieC: @coderabbitai please create a ticket for this.

apps/experiments/views/chat.py
snopoke: wondering if we should use the html encodings or not `&#128077;` `&#128078;`

SmittieC: These emoji's are part of unicode, which should make it fine to include as-is? Or am I missing something?

stephherbers: The main concern if are we sure that all of our users are on environments that support Unicode emojis? I know there are some older software so I wonder if we should error on the safe side

stephherbers: we can circle back on this in a future convo

stephherbers (approved):

Issue #990, opened 2024-12-13
Clarify logic for deleting / archiving working copy of assistants


snopoke:
I think the logic and conditions for deleting / archiving assistants needs to be revised / reviewed. I think there are cases where we should be cleaning up on openai where we aren't and perhaps visa version.

### Initial thoughts:
 
* don't delete if there are other usages (other experiments / pipeline nodes)

#### Local delete (archive)
* archive self
* archive versions
* delete openai assistant for all versions

#### Hard delete:
* Same as above but also delete working version from openai

See code:
* https://github.com/dimagi/open-chat-studio/blob/32e4c79b32f66c4be7ed4dd309c9292ee72882fd/apps/assistants/views.py#L183-L210
* https://github.com/dimagi/open-chat-studio/pull/989

PR #1011, opened 2024-12-19
Assistant Archival Cleanup

stephherbers:
## Description

[990](https://github.com/dimagi/open-chat-studio/issues/990)
Followup logic changes for assistant versioning archiving.

https://github.com/user-attachments/assets/e62dd24c-f18f-4e3d-9b02-ef32b181752f

Now, when a assistant is archived that is a working version, all its versions is also archived and deleted from open AI.

Also from a UI standpoint. When it checks whether is can be archived, if its the working version it checks for all dependencies for all the versions not just the working version and displays everything that needs to be archived first. This way, the version won't be archived with the open AI deleted with something (exp or pipeline) still referencing it). Only the working and published references will be displayed as blocking to the user for the assistant to be archived

### Demo

Screenshot 2025-01-28 at 2 09 43 PM

[loom demo](https://www.loom.com/share/bdd5e8664a2b4baca4cffc76d58d02a1?sid=f4388283-208e-43fb-b3b3-ec8a4072f6d3)

When a user goes to archive an assistant, there will be 3 uses cases checked:

1. Are there any published or working experiments that are referencing the assistant?
2. Are there and working pipelines are thar referencing the assistant
3. Are there and published experiments that have pipelines that are referencing the assistant.
For cases 1 & 3, it links to the experiment version that is blocking, and for case 2 it links to the pipeline that is blocking

### Docs

being updated now


apps/assistants/views.py
```diff
@@ -212,13 +213,23 @@ def delete(self, request, team_slug: str, pk: int):
             messages.success(request, "Assistant Archived")
             return HttpResponse()
         else:
+            version_query = None
+            if assistant.is_working_version:
+                version_query = list(
+                    map(
```
SmittieC, line 219: I might be wrong, but I don't think we have to map the ids to strings?

snopoke (commented):
After reviewing this and testing it locally I think we need to take a step back and look more holistically at how we trace references to objects and make that visible in the UI.

I've started a doc to make it easier to collaborate on the thoughts and ideas: https://docs.google.com/document/d/1Z09GNpO17izoVcRPGSmppNthFruyRJ0033T-RYvapoE/edit?tab=t.0

stephherbers:
noting that we are moving forward with changes from the doc so those commits will be coming in later and I will re-requests reviews

apps/assistants/tests/test_delete.py
```diff
@@ -28,36 +29,92 @@ def code_resource(assistant):
...
+        assistant = OpenAiAssistantFactory()
+        assert assistant.is_archived is False
+        assistant.archive()
+        assert assistant.is_archived is True
+
+    @patch("apps.assistants.sync.push_assistant_to_openai", Mock())
+    def test_archive_assistant_succeeds_with_released_experiment_experiment(self):
+        experiment = ExperimentFactory()
+        exp_v2 = experiment.create_new_version()
+        experiment.is_default = True
```
SmittieC, line 79: Did you mean to make `exp_v2` the default here?

```python
exp_v2.is_default_version = True
```

apps/assistants/tests/test_delete.py
```diff
@@ -28,36 +29,92 @@ def code_resource(assistant):
...
+        experiment.save()
+        assistant = OpenAiAssistantFactory()
+        exp_v2 = experiment.create_new_version()
```
SmittieC, line 82: Nit: Since this will be version 2 (exp_v2), the previous `exp_v2` should be `exp_v1`

apps/assistants/models.py
```diff
@@ -171,29 +172,47 @@ def compare_with_model(self, new: Self, exclude_fields: list[str], early_abort=F
...
+            return Experiment.objects.filter(
+                Q(working_version_id=None) | Q(is_default_version=True),
+                assistant_id__in=assistant_ids,
+                is_archived=False,
+            )
+
+        return self.experiment_set.filter(Q(working_version_id=None) | Q(is_default_version=True), is_archived=False)
 
-    def get_related_pipeline_node_queryset(self):
+    def get_related_pipeline_node_queryset(self, assistant_ids: list = None):
```
SmittieC, line 204: These should check the experiments that the pipeline belongs to instead of the pipeline itself?

apps/assistants/tests/test_delete.py
```diff
@@ -28,36 +29,92 @@ def code_resource(assistant):
...
-    assert all(f.external_id for f in all_files)
-    assert all(f.external_source for f in all_files)
-    client = Mock()
-    delete_openai_files_for_resource(client, code_resource.assistant.team, code_resource)
-
-    assert client.files.delete.call_count == 2
-    all_files = list(code_resource.files.all())
-    assert not any(f.external_id for f in all_files)
-    assert not any(f.external_source for f in all_files)
+class TestAssistantArchival:
```
SmittieC, line 68: Nice tests. It would be great if we can add tests for pipelines using an assistant as well.

apps/assistants/models.py
```diff
@@ -171,32 +172,57 @@ def compare_with_model(self, new: Self, exclude_fields: list[str], early_abort=F
...
-            pipeline__is_archived=False,
-        )
+        if assistant_ids:
+            nodes = Node.objects.filter(type="AssistantNode").filter(
+                Q(pipeline__working_version_id=None),
+                params__assistant_id__in=assistant_ids,
+                pipeline__is_archived=False,
+            )
+        else:
+            nodes = Node.objects.filter(type="AssistantNode").filter(
```
stephherbers, line 214: @SmittieC since there is no "published" version of the pipeline-- is this the correct logic? For the demo you see that a pipeline with an experiment will only return the working experiment as the blocking issue for archiving the assistant and not the published experiment. 
Archiving the working experiment _will_ archive the published experiment, so this is really fine either way. But want to make sure we're on the same page

SmittieC: We care about experiments being working or published in anycase, so I think it's fine. I do think that the nodes query should be 
```python
        nodes = Node.objects.filter(type="AssistantNode").filter(
            params__assistant_id=str(self.id),
            pipeline__is_archived=False,
        )
```
since if there's _any_ pipeline (version or not) that uses this assistant, we'd want to know the state of the experiment (if any) that is using that pipeline, if that makes sense? So we don't care about the pipeline version status, but we do care about the experiment's status _using_ that pipeline.

One other thing regarding the queries: We'd want to use the same filters in the experiments query on line 221 as we do in `get_related_experiments_queryset` to get experiments that are either the working or published versions.

Where do we archive the version of the experiment using the assistant or pipeline that uses the assistant in question? Is that part of this PR?

Two small nitty refactoring suggestions:
- Since this method is now returning an `Experiment` queryset, I'm wondering if the name can do with an update to make it clear that we're looking for pipeline type experiments that uses the assistant(s) in question?
- You can replace lines 207-218 with this
```python
        assistant_ids = assistant_ids if assistant_ids else [self.id]
        nodes = Node.objects.filter(type="AssistantNode").filter(
            Q(pipeline__working_version_id=None),
            params__assistant_id=str(self.id),
            pipeline__is_archived=False,
        )
```

snopoke: I think that's mostly correct. But I would add that it's possible for a Pipeline to exist without being attached to an experiment (which would be the 'working version'). We would consider that blocking in terms of archiving the assistant.

In fact I think it's true that any 'working version' pipeline should block the archive (since we can assume that either it's not attached to an experiment or it's attached to the working version of the experiment).

That means that any pipeline working version should block archive OR any pipeline version that is attached to a 'published' experiment version.

stephherbers: added https://github.com/dimagi/open-chat-studio/pull/1011#:~:text=%E2%80%A6-,a1e85df,-update%20views%20to

apps/assistants/models.py
```diff
@@ -171,33 +173,64 @@ def compare_with_model(self, new: Self, exclude_fields: list[str], early_abort=F
...
+        assistant_ids = assistant_ids if assistant_ids else [str(self.id)]
+        nodes = Node.objects.filter(type="AssistantNode").filter(
+            pipeline__working_version_id__isnull=False,
+            params__assistant_id__in=assistant_ids,
+            pipeline__is_archived=False,
+        )
+        if nodes.exists():
+            pipeline_ids = nodes.values_list("pipeline_id", flat=True)
+            return Experiment.objects.filter(
+                is_default_version=True,
```
SmittieC, line 228: Do we care about working experiments as well?

apps/assistants/views.py
```diff
@@ -212,20 +213,38 @@ def delete(self, request, team_slug: str, pk: int):
...
+                        str,
+                        OpenAiAssistant.objects.filter(
+                            Q(id=assistant.id) | Q(working_version__id=assistant.id)
+                        ).values_list("id", flat=True),
+                    )
+                )
+            pipeline_nodes = assistant.get_related_pipeline_node_queryset()
```
SmittieC, line 226: Is this supposed to be here? It gets overwritten on line 231.

apps/assistants/models.py
```diff
@@ -171,33 +173,64 @@ def compare_with_model(self, new: Self, exclude_fields: list[str], early_abort=F
...
-        super().archive()
         if self.is_working_version:
-            # TODO: should this delete the assistant from OpenAI?
+            for (
+                version
+            ) in self.versions.all():  # first perform all checks so assistants are not archived prior to return False
+                if (
+                    version.get_related_experiments_queryset().exists()
+                    or version.get_related_pipeline_node_queryset().exists()
+                    or self.get_related_experiments_with_pipeline_queryset().exists()
```
snopoke, line 189:
```suggestion
                    or version.get_related_experiments_with_pipeline_queryset().exists()
```

apps/assistants/models.py
```diff
@@ -171,33 +173,64 @@ def compare_with_model(self, new: Self, exclude_fields: list[str], early_abort=F
...
+        if assistant_ids:
+            return Experiment.objects.filter(
+                Q(working_version_id=None) | Q(is_default_version=True),
+                assistant_id__in=assistant_ids,
+                is_archived=False,
+            )
 
-    def get_related_pipeline_node_queryset(self):
-        from apps.pipelines.models import Node
+        return self.experiment_set.filter(Q(working_version_id=None) | Q(is_default_version=True), is_archived=False)
```
snopoke, line 208: nit: to avoid the repetition you could write this as:

```python
assistant_ids = assistant_ids or [self.id]
return Experiment.objects.filter(
    assistant_id__in=assistant_ids,
    Q(working_version_id=None) | Q(is_default_version=True),
    is_archived=False
)
```

apps/assistants/models.py
snopoke: @stephherbers correct me if I'm wrong but I think we assume that a working version of an experiment must reference a working version of a pipeline which is covered in the query above.

Maybe a comment in code would help clarify the logic for posterity.

apps/assistants/models.py
```diff
@@ -171,33 +173,64 @@ def compare_with_model(self, new: Self, exclude_fields: list[str], early_abort=F
...
 
+    def get_related_experiments_with_pipeline_queryset(self, assistant_ids: list = None):
+        assistant_ids = assistant_ids if assistant_ids else [str(self.id)]
+        nodes = Node.objects.filter(type="AssistantNode").filter(
+            pipeline__working_version_id__isnull=False,
+            params__assistant_id__in=assistant_ids,
+            pipeline__is_archived=False,
+        )
+        if nodes.exists():
+            pipeline_ids = nodes.values_list("pipeline_id", flat=True)
```
snopoke, line 226: Since we're going to execute the query anyway it doesn't make sense to do the exists check:
```suggestion
        if pipeline_ids := nodes.values_list("pipeline_id", flat=True):
```

apps/assistants/models.py
```diff
@@ -171,33 +173,64 @@ def compare_with_model(self, new: Self, exclude_fields: list[str], early_abort=F
...
+            return Experiment.objects.filter(
+                is_default_version=True,
+                pipeline_id__in=pipeline_ids,
+                is_archived=False,
+            )
+        return nodes
```
snopoke, line 232: to avoid doing this query again and since we know this is an empty queryset by now, you can return `PipelineNode.objects.none()`.

Actually, to be consistent you should return `Experiment.objects.none()`

apps/assistants/tests/test_delete.py
```diff
@@ -28,36 +30,146 @@ def code_resource(assistant):
...
+    def test_archive_assistant(self):
+        assistant = OpenAiAssistantFactory()
+        assert assistant.is_archived is False
+        assistant.archive()
+        assert assistant.is_archived is True
+
+    @patch("apps.assistants.sync.push_assistant_to_openai", Mock())
+    def test_archive_assistant_succeeds_with_released_related_experiment(self):
```
snopoke, line 77: nit: maybe just me but I find this naming clearer
```suggestion
    def test_archive_assistant_succeeds_with_unpublished_related_experiment(self):
```

apps/assistants/tests/test_delete.py
```diff
@@ -28,36 +30,146 @@ def code_resource(assistant):
...
+        assert assistant.is_archived is True  # archiving succeeded
+
+    @patch("apps.assistants.sync.push_assistant_to_openai", Mock())
+    def test_asistant_archive_blocked_by_working_related_experiment(self):
+        assistant = OpenAiAssistantFactory()
+        experiment = ExperimentFactory(assistant=assistant)
+        experiment.save()
+
+        assert experiment.is_working_version is True
+        assistant.archive()
```
snopoke, line 99: nit: you can also assert the return value:
```suggestion
        assert not assistant.archive()
```

apps/assistants/tests/test_delete.py
```diff
@@ -28,36 +30,146 @@ def code_resource(assistant):
...
+        assistant.refresh_from_db()
+        assert assistant.is_archived is False  # archiving blocked
+
+    @patch("apps.assistants.sync.push_assistant_to_openai", Mock())
+    def test_asistant_archive_blocked_by_published_related_experiment(self):
```
snopoke, line 104: I don't thinks this test is testing what is says. There are no 'published' experiments in the test. It seems to be testing the same as `test_asistant_archive_blocked_by_working_related_experiment` except also testing that the assistant versions are archived.

apps/assistants/tests/test_delete.py
```diff
@@ -28,36 +30,146 @@ def code_resource(assistant):
...
+
+        experiment.archive()  # first archive related experiment through v2_assistant
+        assistant.archive()
+        v2_assistant.refresh_from_db()
+
+        assert assistant.is_archived is True  # archiving successful
+        assert v2_assistant.is_archived is True
+
+    @patch("apps.assistants.sync.push_assistant_to_openai", Mock())
+    def test_archive_assistant_succeeds_with_released_related_pipeline(self):
```
snopoke, line 122:
```suggestion
    def test_archive_assistant_succeeds_with_unpublished_related_pipeline(self):
```

apps/assistants/tests/test_delete.py
```diff
@@ -28,36 +30,146 @@ def code_resource(assistant):
...
+        pipeline = PipelineFactory()
+        exp_v1 = ExperimentFactory(pipeline=pipeline)
+        exp_v2 = exp_v1.create_new_version()
+        assistant = OpenAiAssistantFactory()
+        NodeFactory(pipeline=exp_v2.pipeline, type="AssistantNode", params={"assistant_id": assistant.id})
+        exp_v2.is_default_version = False
+        exp_v2.save()
```
snopoke, line 129: I don't think this is necessary since `create_new_version` won't make it the default version unless you pass `make_default=True`

apps/assistants/views.py
snopoke: we do because the pipeline params stores the values as strings

snopoke (commented):
This is very close now. I'm going to check it out and test it locally.

templates/assistants/partials/referenced_objects.html
```diff
@@ -18,6 +18,16 @@ <h2 class="font-semibold my-2">Pipelines</h2>
     </ul>
   {% endif %}
 
+  {% if pipeline_nodes %}
```
snopoke, line 21:
```suggestion
  {% if experiments_with_pipeline_nodes %}
```

apps/assistants/tests/test_delete.py
stephherbers: here it is because it's the first version (other than the working), so it will make it the default version unless you set it to false

stephherbers:
@snopoke @SmittieC updated the screenshot in the description-- and I'll try for rearrqanging the confirmation modals a bit more before calling it quits. I believe this covers everything we discussed, plus updated tests

snopoke (changes requested):
so close!

snopoke (approved):
Looks ready to merge once tests are passing:

```
=========================== short test summary info ============================
FAILED apps/assistants/tests/test_delete.py::TestAssistantArchival::test_archive_assistant_fails_with_working_related_versioned_pipeline_and_working_experiment - assert not True
 +  where True = <bound method OpenAiAssistant.archive of <OpenAiAssistant: >>()
 +    where <bound method OpenAiAssistant.archive of <OpenAiAssistant: >> = <OpenAiAssistant: >.archive
FAILED apps/assistants/tests/test_delete.py::TestAssistantArchival::test_archive_assistant_fails_with_working_related_pipeline - assert not True
 +  where True = <bound method OpenAiAssistant.archive of <OpenAiAssistant: >>()
 +    where <bound method OpenAiAssistant.archive of <OpenAiAssistant: >> = <OpenAiAssistant: >.archive
=========== 2 failed, 878 passed, 448 warnings in 323.87s (0:05:23) ============
```

SmittieC (approved):

PR #1012, opened 2024-12-30
Add get and set participant data functions to code node

proteusvacuum:
## Description

Allows access to participant data in the code node.
I went with the functional approach as it seemed to be the simplest, but happy to revisit this if there are strong opinions.
We can then have corresponding `get_state_value` and `set_state_value` with the global state object. 

### Demo

https://github.com/user-attachments/assets/e44a2615-9562-4213-9d39-951b03c43912


SmittieC (dismissed):
Nice!

snopoke (dismissed):
Is there a way we could cache the participant data object to avoid having to look it up each time? Is it possible to add methods to the PipelineState object?

PR #1019, opened 2025-01-02
Pipeline Shared State and State Key Router Node

proteusvacuum:
## Description

Adds a `shared_state` that can be accessed between all the nodes. I didn't yet add the ability to modify this in any node other than the pipeline node, as I wanted to make sure the general shape of this made sense to others first. However, since the state includes the output of every node, I think most usecases can be covered already. 

By default, this state includes:
```
outputs: { <node_name>: <output of the node> }
user_input: <input to the pipeline>
```
In order to have "friendly" names that could be accessed in the state, I also added a new required parameter `name` to all nodes, which defaults to the node ID, but can be changed by the user. 

This PR also includes a "State Key Router" which allows you to route based on any key in the `shared_state`. Currently, you'd have to set the value of this key to a string. 

## User Impact

Users can set and get keys in a shared state in the code node, then subsequently route based on the value of those keys. 

### Demo

https://github.com/user-attachments/assets/1b222cd9-2c78-43a7-95bf-61dce13da49c


proteusvacuum:
@snopoke currently, yes that's correct. If we wanted to route on participant data, should we make a separate node type, `ParticipantDataRouterNode`?

snopoke:
> @snopoke currently, yes that's correct. If we wanted to route on participant data, should we make a separate node type, `ParticipantDataRouterNode`?

OK - let's leave that for now then

snopoke (approved):
If a name has been set it would be nice to display it in the node header

apps/pipelines/nodes/base.py
```diff
@@ -91,6 +109,7 @@ def process(self, node_id: str, incoming_edges: list, state: PipelineState, conf
                 break
         else:  # This is the first node in the graph
             input = state["messages"][-1]
+            state["shared_state"]["user_input"] = input
```
snopoke, line 112: Why not do this where the state is initially created?

SmittieC (dismissed):
nice

apps/pipelines/nodes/base.py
proteusvacuum: I think we'd have to remember to do it everywhere we call `invoke` on the pipeline, and this seemed like something the caller shouldn't have to care about.

snopoke:
@coderabbit review

snopoke:
@coderabbit review

snopoke:
@coderabbit resolve

SmittieC (approved):
Overall looks good to me

snopoke:
🎈

PR #1023, opened 2025-01-06
filter out versioned consent forms

snopoke:
## Description
This fixes a bug where `ConsentForm.get_default` would error with a Multiple Results error because it wasn't filtering out form versions.

This PR also includes a small refactor to move the `working_versions_queryset` method to the `VersionsObjectManagerMixin`


SmittieC (dismissed):
Interesting that this is only surfacing now

snopoke:
> Interesting that this is only surfacing now

The only usage of the `get_default` method is in `apps.pipelines.nodes.helpers.temporary_session` so I guess nobody has tested a versioned pipeline yet.

PR #1027, opened 2025-01-08
Show badge when changes detected

SmittieC:
## Description

Check for changes async an show a badge to the user when changes are detected. A timeit test shows no major improvement between aborting early or not. Around 3ms. That said, my local bots are very simple

## User Impact

User will clearly see on the experiment home screen when new changes are detected

### Demo

demo

### Docs

N/A since this is a UI update only

### Comment
Disabling the "+ create version" button in the versions table when there are no diff - without having to do a synchronous second version comparison - is a bit more tricky (the first comparison is done by the "draft" badge async call). I'll need to play around with components chatting to each other, maybe through triggers. Not sure yet. Anycase, this is left for a followup PR

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -1134,7 +1134,7 @@ class ExperimentRoute(BaseTeamModel, VersionsMixin):
                 # Compare experimens using their `version` instances for a comprehensive comparison
                 child_version = self.child.version
                 latest_version = latest_child_version.version
-                child_version.compare(latest_version, early_abort=True)
+                child_version.compare(latest_version)
 
                 if not child_version.fields_changed:
                     new_child = latest_child_version
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -766,7 +766,7 @@ class Experiment(BaseTeamModel, VersionsMixin, CustomActionOperationMixin):
         """
         version = self.version
         if prev_version := self.latest_version:
-            version.compare(prev_version.version)
+            version.compare(prev_version.version, early_abort=True)
         return version.fields_changed
 
     @transaction.atomic()
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -1156,7 +1156,7 @@ class ExperimentRoute(BaseTeamModel, VersionsMixin):
             description = f"{description} since {changed_fields} changed."
         return description
 
-    def compare_with_model(self, route: "ExperimentRoute", exclude_fields):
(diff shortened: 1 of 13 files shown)


SmittieC:
@coderabbitai please review.

apps/experiments/models.py
```diff
@@ -1134,7 +1134,7 @@ def create_new_version(self, new_parent: Experiment) -> "ExperimentRoute":
                 # Compare experimens using their `version` instances for a comprehensive comparison
                 child_version = self.child.version
                 latest_version = latest_child_version.version
-                child_version.compare(latest_version)
+                child_version.compare(latest_version, early_abort=True)
```
snopoke, line 1137: Won't this affect the `_generate_version_description` function?

apps/experiments/versioning.py
```diff
@@ -125,14 +130,19 @@ def _compare_queryset(self, previous_queryset):
                 # TODO: When comparing static trigger versions and only the action changed, it is not being picked up.
                 previous_record_version_ids.append(previous_record.id)
                 prev_version_field = VersionField(raw_value=previous_record, to_display=self.to_display)
-                version_field.compare(prev_version_field)
+                version_field.compare(prev_version_field, early_abort=early_abort)
                 self.changed = self.changed or version_field.changed
             else:
                 version_field.changed = self.changed = True
 
+        if early_abort and self.changed:
```
snopoke, line 138: this could be in the for loop which would abort even earlier if a change was detected

snopoke (commented):
> A timeit test shows no major improvement between aborting early or not. Around 3ms. That said, my local bots are very simple

Regardless I still this it's better to do it async and it's a good practice to get into.

> Disabling the "+ create version" button in the versions table when there are no diff - without having to do a synchronous second version comparison - is a bit more tricky (the first comparison is done by the "draft" badge async call). I'll need to play around with components chatting to each other, maybe through triggers. Not sure yet. Anycase, this is left for a followup PR

This is where an out of bounds update might help (https://htmx.org/attributes/hx-swap-oob/). You could fire the stale check on the table button and use it to update both the button and the header indicator.

apps/experiments/models.py
SmittieC: You are correct. Nice catch

apps/experiments/versioning.py
SmittieC: Updated

snopoke (approved):

PR #1031, opened 2025-01-09
Set current team in the temporary_session context manager

SmittieC:
See https://dimagi.sentry.io/issues/6206061542/?alert_rule_id=14063560&alert_type=issue&notification_uuid=0be72297-5885-4a6c-8879-bde7414fb805&project=4505001320316928&referrer=issue_alert-slack


snopoke (dismissed):
This is one of those times I'd want to turn off auditing since we don't care about the new objects being created. Unfortunately there's no way to do that.

snopoke (approved):

PR #1032, opened 2025-01-09
Docs / help text updates

snopoke:
## Description
Change how help text is shown to be more consistent and to prevent clutter of the UI. Also add a utility for rendering a link to documentation at the end of help text.

This PR makes help text render using next to the form element label as an icon with the content shown on hover:

image

## User Impact
Improved help text formatting and links to docs.

### Demo
See screenshot


apps/assistants/forms.py
```diff
@@ -64,6 +58,9 @@ def __init__(self, request, *args, **kwargs):
         self.fields["builtin_tools"].widget.attrs = {
             "x-model.fill": "builtinTools",
         }
+        self.fields["instructions"].help_text = render_help_with_link(
+            INSTRUCTIONS_HELP_TEXT, "/concepts/prompt_variables/"
```
SmittieC, line 62: How about the link paths to the `DOCUMENTATION_LINKS` variable?

SmittieC (approved):
These are some sweet changes. One non-blocking comment

apps/assistants/forms.py
snopoke: Can you explain the rationale behind the `DOCUMENTATION_LINKS` setting? It is just so that all the docs links are centralised?

SmittieC: Pretty much. Since it's external links that may change, I feel it would make it just that little bit easier to find and update the previous one. I don't feel very strong about this preference, so welcome to ignore it

stephherbers (dismissed):
this is great-- for dev docs for how to add this maybe it would even be good to add in our docs repo?

SmittieC (approved):

PR #1036, opened 2025-01-13
Add seed message to assistant advanced settings

stephherbers:
## Description

This was a side effect of [this PR](https://github.com/dimagi/open-chat-studio/pull/979) flagged in slack that removed the ability to configure a seed message with an experiment using an assistant

## User Impact

brings back functionality that went away ~1 month ago

### Demo

Screenshot 2025-01-13 at 1 25 42 PM
Screenshot 2025-01-13 at 1 25 55 PM
Screenshot 2025-01-13 at 1 26 03 PM

### Docs

There is not a how-to for experiment set up yet to break down the settings but it should spell out which setting are available for each

### DIFF
--- a/templates/experiments/experiment_form.html
+++ b/templates/experiments/experiment_form.html
@@ -120,7 +120,7 @@
           {% render_form_fields form "seed_message" %}
         </span>
         <span x-show="type == 'assistant'">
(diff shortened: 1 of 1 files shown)


SmittieC (approved):
Thanks!

PR #1040, opened 2025-01-14
switch to UV wholesale

snopoke:
## Description
Update the infrastructure dependency management to move away from using PIP requirements files to managing in `pyproject.toml` ([PEP-0621](https://peps.python.org/pep-0621/)). This also moves away from `pip-compile` in favour of [uv](https://docs.astral.sh/uv/). This was already partially done before however the change in this PR completely remove `pip-tools` from the dependencies as well as the tooling for using them.

Along with this change the `Dockerfile` is updated to use UV as well as the github actions.

## User Impact
This should not impact users at all.

### Docs
README updated (we need dev docs)


snopoke:
Replaced by https://github.com/dimagi/open-chat-studio/pull/1530

PR #1041, opened 2025-01-14
Ability to message using the connect messaging channel | Phase 2

SmittieC:
I might break this PR up into two after code rabbit reviewed it

## Description

This completes all the work for the integration, except ability to link a connect messaging channel to your experiment through the UI.
- Bot is able to initiate a conversation based on the prompt given through the API
- Message encryption and decryption
- Tests

## Still todo
- Ability to link a connect messaging channel to an experiment
- Manual testing


apps/channels/models.py
```diff
@@ -95,7 +104,7 @@ def channel_identifier_key(self) -> str:
             case self.SLACK:
                 return "slack_channel_id"
             case self.COMMCARE_CONNECT:
-                return "connect_id"
+                return "bot_name"
```
SmittieC, line 107: The reasoning behind using the bot name is that this will be shown to the user, and we don't want other bots to use the same name, which can become misleading

apps/channels/clients/connect_client.py
```diff
@@ -29,3 +45,53 @@ def create_channel(self, connect_id: UUID, channel_source: str) -> UUID:
...
+        Raises:
+            ValueError if the message authenticity cannot be trusted
+        """
+        decrypted_messages = []
+        for message in messages:
+            nonce = base64.b64decode(message["nonce"])
+            tag = base64.b64decode(message["tag"])
+            ciphertext = base64.b64decode(message["ciphertext"])
+            cipher = AES.new(key, AES.MODE_GCM, nonce=nonce)
+            decrypted_messages.append(cipher.decrypt_and_verify(ciphertext, tag).decode("utf-8"))
```
snopoke, line 62: There is an `encrypt_message` method below. It is nice to have symmetrical methods for this kind of thing. Maybe consider extracting this to a `decrypt_message(key, ciphertext, tag, nonce)` method.

You could also move the b64 encoding there to make it completely symmetrical: `assert message == decrypt_message(key, *encrypt_message(key, message))`

apps/channels/forms.py
```diff
@@ -217,3 +217,20 @@ def post_save(self, channel: ExperimentChannel):
...
             except Exception as e:
                 raise ExperimentChannelException("Failed to join the channel") from e
+
+
+class CommCareConnectChannelForm(ExtraFormBase):
+    bot_name = forms.CharField(
+        label="Bot Name",
+        help_text=(
+            "This is the name of the bot that will be displayed to users on commcare connect"
+            "Once set, this name cannot be changed."
```
snopoke, line 227: Why shouldn't it be changed? I don't think it matters if they change it here since it will only get used for new users. If we don't allow them to change it they will remove the channel and then add a new one with a different name.

@calellowitz other than displaying this to the user for consent, is the `channel_source` used for anything?

apps/channels/models.py
snopoke: Interesting. I'm not sure about this, or whether the connect channel even needs this uniqueness check. But I do think that you're right in saying that it might be confusing or misleading for users if there are two bots with the same name. I guess let's keep this for now. Please add a comment to the code with some reasoning.

apps/channels/urls.py
```diff
@@ -16,4 +16,5 @@
     ),
     path("whatsapp/turn/<uuid:experiment_id>/incoming_message", views.new_turn_message, name="new_turn_message"),
     path("api/<uuid:experiment_id>/incoming_message", views.new_api_message, name="new_api_message"),
+    path("commcare_connect/new_message/", views.new_connect_message, name="new_connect_message"),
```
snopoke, line 19: use consistent url paths
```suggestion
    path("commcare_connect/incoming_message", views.new_connect_message, name="new_connect_message"),
```

apps/channels/views.py
```diff
@@ -124,3 +127,32 @@ def new_api_message(request, experiment_id: uuid):
...
+@csrf_exempt
+@verify_hmac
+def new_connect_message(request: HttpRequest):
+    serializer = CommCareConnectMessageSerializer(data=json.loads(request.body))
+    if not serializer.is_valid():
+        return JsonResponse(serializer.errors, status=400)
+
+    connect_channel_id = serializer.data["channel_id"]
+    try:
+        participant_data = ParticipantData.objects.prefetch_related("participant").get(
```
snopoke, line 142: unnecessary prefetch
```suggestion
        participant_data = ParticipantData.objects.get(
```

apps/channels/views.py
```diff
@@ -124,3 +127,32 @@ def new_api_message(request, experiment_id: uuid):
...
+            content_type=ContentType.objects.get_for_model(Experiment),
+            system_metadata__commcare_connect_channel_id=connect_channel_id,
+        )
+
+        channel = ExperimentChannel.objects.prefetch_related("experiment").get(
```
snopoke, line 147:
```suggestion
        channel = ExperimentChannel.objects.get(
```

apps/experiments/models.py
```diff
@@ -1389,6 +1390,10 @@ class ParticipantData(BaseTeamModel):
         models.CharField(max_length=255, blank=True, help_text="The base64 encoded encryption key")
     )
 
+    def get_encryption_key_bytes(self):
+        # TODO: What if we use a getter/setting to ensure what we have is always bytes?
```
snopoke, line 1394: probably not necessary

apps/teams/migrations/0007_create_commcare_connect_flag.py
snopoke, line 1: This isn't necessary - Flags get created on demand

apps/api/tests/test_api.py
```diff
@@ -465,3 +460,70 @@ def test_empty_request_body(self, client):
...
+
+def _setup_participant_data(experiment, connect_id, commcare_connect_channel_id, encryption_key=None):
+    participant = ParticipantFactory(
+        team=experiment.team, identifier=connect_id, platform=ChannelPlatform.COMMCARE_CONNECT
+    )
+    return ParticipantData.objects.create(
+        team=experiment.team,
+        participant=participant,
+        content_object=experiment,
+        system_metadata={"commcare_connect_channel_id": commcare_connect_channel_id},
```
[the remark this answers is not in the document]
SmittieC: Ah yes, thanks

apps/channels/clients/connect_client.py
```diff
@@ -29,3 +45,53 @@ def create_channel(self, connect_id: UUID, channel_source: str) -> UUID:
...
+                "tag": base64.b64encode(tag_bytes).decode(),
+                "nonce": base64.b64encode(nonce_bytes).decode(),
+            },
+            "message_id": str(uuid4()),
+        }
+
+        url = f"{self._base_url}/messaging/send_fcm/"
+        response = self.client.post(url, json=payload)
+        if response.status_code == 403:
+            logger.info("User did not give consent to receive messages")
```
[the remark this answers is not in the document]
SmittieC: Makes sense. I'm letting it cause an exception then as well: https://github.com/dimagi/open-chat-studio/pull/1041/commits/0be24a826c9850227d1c65712b62d48c2d55c2e6

apps/channels/clients/connect_client.py
SmittieC: I like this. Like this? https://github.com/dimagi/open-chat-studio/pull/1041/commits/42ad7f26f758ea118ca1c36b37d54e62fd59175f

apps/channels/models.py
SmittieC: I share the sentiment. Comment added [here](https://github.com/dimagi/open-chat-studio/pull/1041/commits/fffac63d6ae45f33a5a378daf61ca54681351a1a)

apps/teams/migrations/0007_create_commcare_connect_flag.py
SmittieC: I also thought so, but looking at the methods that I'm calling here, there's no logic that does this. It only adds it to the cache. The `is_active` method from `waffle` does it though. We can update our `is_active_for_team` code to do the same?

apps/channels/forms.py
calellowitz: It is used for the user visible name on the mobile, so in your list of channels.

calellowitz: Yeah, connectid can't tell those channels are related in any way. Source is set once at creation, but otherwise the channels are exclusively referenced by UUID, and don't share anything on the backend data model. Even if they had the same "source," they wouldn't be connected in any way.

SmittieC: Updated. https://github.com/dimagi/open-chat-studio/pull/1041/commits/ca2ac38a3876e20d36c23173a82b0d0eb6bdf31e

apps/channels/models.py
```diff
@@ -53,6 +53,13 @@ def for_dropdown(cls, used_platforms, team) -> dict:
         if not settings.SLACK_ENABLED:
             platform_availability.pop(cls.SLACK)
 
+        flag = Flag.get("commcare_connect")
+        commcare_connect_flag_enabled = flag.is_active_for_team(team)
```
snopoke, line 57: I thought I'd created a utility for this but I can't find it now:

```python
flag_is_active_for_team(team, flag_name: str):
    flag = get_waffle_flag_model().get(flag_name)
    return flag.is_active_for_team(team)
```

apps/channels/models.py
snopoke: Just wondering if we shouldn't make this name more specific since it will be a global check. Maybe something like `cc_bot_name` or `connect_bot_name` or `commcare_connect_bot_name`

apps/channels/serializers.py
```diff
@@ -0,0 +1,14 @@
...
+    timestamp = serializers.CharField()
+    message_id = serializers.UUIDField()
+    ciphertext = serializers.CharField()
+    tag = serializers.CharField()
+    nonce = serializers.CharField()
+
+
+class CommCareConnectMessageSerializer(serializers.Serializer):
+    channel_id = serializers.UUIDField()
+    messages = serializers.ListField(child=MessageSerializer())
```
snopoke, line 14: The recommended way to do this is with `many=True`:
```suggestion
    messages = MessageSerializer(many=True)
```

See https://www.django-rest-framework.org/api-guide/serializers/#dealing-with-nested-objects

apps/channels/views.py
```diff
@@ -124,3 +127,32 @@ def new_api_message(request, experiment_id: uuid):
...
+@csrf_exempt
+@verify_hmac
+def new_connect_message(request: HttpRequest):
+    serializer = CommCareConnectMessageSerializer(data=json.loads(request.body))
+    if not serializer.is_valid():
+        return JsonResponse(serializer.errors, status=400)
+
+    connect_channel_id = serializer.data["channel_id"]
+    try:
+        participant_data = ParticipantData.objects.get(
```
snopoke, line 142: nit:
```suggestion
        participant_data = ParticipantData.objects.defer("data").get(
```

snopoke (commented):
mostly small comments

apps/teams/migrations/0007_create_commcare_connect_flag.py
snopoke: Ah yes, I see. But there's no need to create it either right? We can as easily create it manually via the Django admin?

Either way is fine. If you want to add the 'create' logic to the `is_active_for_team` method that's also fine.

SmittieC: > If you want to add the 'create' logic to the is_active_for_team method that's also fine.

I'll do this in a separate PR

apps/api/views.py
```diff
@@ -382,3 +382,83 @@ def consent(request: Request):
...
+
+    try:
+        participant_data = ParticipantData.objects.defer("data").get(
+            participant__identifier=identifier,
+            participant__platform=platform,
+            object_id=experiment.id,
+            content_type=ContentType.objects.get_for_model(Experiment),
+        )
+
+        ExperimentChannel.objects.get(platform=platform, experiment=experiment)
```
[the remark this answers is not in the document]
SmittieC: Ah yes good point

apps/api/views.py
```diff
@@ -382,3 +382,83 @@ def consent(request: Request):
...
+    except ParticipantData.DoesNotExist:
+        return Response({"detail": "Participant not found"}, status=status.HTTP_404_NOT_FOUND)
+    except ExperimentChannel.DoesNotExist:
+        return Response(
+            {"detail": f"Experiment cannot send messages on the {platform} channel"}, status=status.HTTP_404_NOT_FOUND
+        )
+
+    if platform == ChannelPlatform.COMMCARE_CONNECT and participant_data.system_metadata.get("consent", False) is False:
```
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1041/commits/36132d73d3a5502be49b80ac2ad9a9e6b4e7a67f

apps/api/views.py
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/commit/36132d73d3a5502be49b80ac2ad9a9e6b4e7a67f

apps/channels/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1041/commits/b8d3c9509f249c4574e2880652c3c07cb1b7d02e

apps/api/views.py
```diff
@@ -373,7 +373,86 @@ def consent(request: Request):
...
+
+    try:
+        participant_data = ParticipantData.objects.defer("data").get(
+            participant__identifier=identifier,
+            participant__platform=platform,
+            object_id=experiment.id,
+            content_type=ContentType.objects.get_for_model(Experiment),
+        )
+
+        ExperimentChannel.objects.filter(platform=platform, experiment=experiment).exists()
```
snopoke, line 445: this won't raise an exception now so you'll need to check that the return value is `True`

SmittieC: https://github.com/dimagi/open-chat-studio/pull/1041/commits/0b87ceb0a93bedd5d97331a9f88a2ee4555b8a1a

apps/channels/serializers.py
SmittieC: Ah thanks: https://github.com/dimagi/open-chat-studio/pull/1041/commits/210136831ed9dad0af7614616ff616c166bc5314

snopoke (dismissed):

PR #1042, opened 2025-01-15
Node deprecation

snopoke:
## Description
Tools for deprecating pipeline nodes. 

This adds some more `ui` fields to the node schema:

| | |
| -- | -- |
| ui:can_add | Controls whether the node will appear in the component list |
| ui:can_delete | Controls whether a node can be deleted |
| ui:deprecated | Set's `ui:can_add` to False and displays a deprecation icon with a message (see below) |
| ui:deprecation_message | Custom message to append to the end of the standard message |

This also adds a `deprecated_node` decorator for clarity and simplicity.

## User Impact
User's won't be able to add the 'Do Nothing', 'Conditional' or 'LLM Response' nodes. In addition the 'Conditional' and 'LLM Response' nodes will display deprecation warnings.

The 'LLM response with prompt' node has been renamed to 'LLM'

### Demo
image

### Docs
Pipeline docs still TODO


SmittieC (approved):
LGTM

stephherbers (commented):
🌟

PR #1043, opened 2025-01-15
Check assistant exists before making requests

snopoke:
resolves [DIMAGI-BOTS-M7](https://dimagi.sentry.io/issues/6220897112/)

## Description
In some cases the assistant is saved in OCS but the push to OpenAI failed. This leaves it with a blank `assistant_id`.

This PR checks that the `assistant_id` is set before making requests.

### DIFF
--- a/apps/assistants/sync.py
+++ b/apps/assistants/sync.py
@@ -158,6 +158,9 @@ def delete_file_from_openai(client: OpenAI, file: File):
 @wrap_openai_errors
 def sync_from_openai(assistant: OpenAiAssistant):
     """Syncs the local assistant instance with the remote OpenAI assistant."""
+    if not assistant.assistant_id:
+        return
+
     client = assistant.llm_provider.get_llm_service().get_raw_client()
     openai_assistant = client.beta.assistants.retrieve(assistant.assistant_id)
     for key, value in _openai_assistant_to_ocs_kwargs(openai_assistant, team=assistant.team).items():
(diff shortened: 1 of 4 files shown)


apps/assistants/sync.py
```diff
@@ -158,6 +158,9 @@ def delete_file_from_openai(client: OpenAI, file: File):
 @wrap_openai_errors
 def sync_from_openai(assistant: OpenAiAssistant):
     """Syncs the local assistant instance with the remote OpenAI assistant."""
+    if not assistant.assistant_id:
```
SmittieC, line 161: Am I understanding the issue correctly when I say that in some cases where we failed to push the assistant, users tried to sync and this method failed? Is this what we're trying to fix?

SmittieC (approved):
Overall looks great. The complexities with syncing two systems just upgraded it seems.

I'm wondering about how we should handle assistant versions that failed to sync. So when we create a new experiment version, and the new assistant version fails to sync, how would we effectively make that known to the user?

We can easily detect this, so I'm thinking we show an exclamation mark either at the specific version and/or the assistant whose version didn't sync with a button that allows them to redo it?

apps/assistants/sync.py
snopoke: This only happens if the initial push failed. After at least one push has succeeded the `assistant_id` will be set. But if that push failed then we shouldn't try these other methods until we've successfully pushed at least once.

snopoke:
> I'm wondering about how we should handle assistant versions that failed to sync. So when we create a new experiment version, and the new assistant version fails to sync, how would we effectively make that known to the user?
> 
> We can easily detect this, so I'm thinking we show an exclamation mark either at the specific version and/or the assistant whose version didn't sync with a button that allows them to redo it?

Good questions. Food for extra thought.

PR #1044, opened 2025-01-15
monkeypatch executor for pipeline test runs

snopoke:
Resolves [DIMAGI-BOTS-M3](https://dimagi.sentry.io/issues/6220745979/)
Resolves [DIMAGI-BOTS-M4](https://dimagi.sentry.io/issues/6220774167/)

## Description
When running the 'test pipeline' task we create a temporary experiment and session and use Django transactions to roll that back after the pipeline has completed. If a pipeline has concurrent nodes langgraph will use a `ThreadPoolExecutor` to run the nodes. Because the nodes execute in a separate thread they will use a different DB connection and hence won't be able to see the temporary session because the transaction is still open.

This PR patches the langchain code to use a fake executor during the pipeline test. The fake executor runs the nodes in the main thread.

## User Impact
Users will be able to test pipelines that have concurrent nodes without receiving weird DB errors.

### Demo
image

### DIFF
--- a/apps/pipelines/executor.py
+++ b/apps/pipelines/executor.py
@@ -1,0 +1,34 @@
+from concurrent.futures import Executor, Future
+from contextlib import contextmanager
+
+
+class CurrentThreadExecutor(Executor):
+    """Fake thread pool executor that runs tasks in the current thread."""
+
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


SmittieC (approved):
Oh interesting!

PR #1055, opened 2025-01-17
Get participant data from the working version

SmittieC:
## Description

Get participant data from the working version. It's always saved on the working version.

### DIFF
--- a/apps/chat/channels.py
+++ b/apps/chat/channels.py
@@ -825,7 +825,10 @@ class CommCareConnectChannel(ChannelBase):
 
     @cached_property
     def participant_data(self) -> ParticipantData:
-        return self.experiment.participant_data.defer("data").get(participant__identifier=self.participant_identifier)
+        experiment = self.experiment
+        if self.experiment.is_a_version:
+            experiment = self.experiment.working_version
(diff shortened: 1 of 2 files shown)


apps/chat/channels.py
```diff
@@ -825,7 +825,10 @@ def send_text_to_user(self, text: str):
 
     @cached_property
     def participant_data(self) -> ParticipantData:
-        return self.experiment.participant_data.defer("data").get(participant__identifier=self.participant_identifier)
+        experiment = self.experiment
+        if self.experiment.is_a_version:
+            experiment = self.experiment.working_version
+        return experiment.participant_data.defer("data").get(participant__identifier=self.participant_identifier)
```
snopoke, line 831: You could reduce this from 2 queries to 1 by doing a straight query instead of an FK + MtoM query (since you don't need the working version for anything other than it's ID):

```python
ParticipantData.filter(content_type__model="experiment", object_id=e_id, team=team)...
```

snopoke: (You could create another manager method like `for_experiment` that takes and experiment ID and team)

PR #1056, opened 2025-01-17
Remove trailing slashes

SmittieC:
## Description

I decided to remove the `/` at the end of some endpoints to be consistent with others. I'll ping Cal to update the paths on the connectid server

## User Impact

Just need to update the API at connect

### Docs

API docs updates are included in this PR

### DIFF
--- a/apps/api/urls.py
+++ b/apps/api/urls.py
@@ -10,9 +10,9 @@ router.register(r"experiments", views.ExperimentViewSet, basename="experiment")
 router.register(r"sessions", views.ExperimentSessionViewSet, basename="session")
 
 connect_patterns = [
-    path("generate_key/", views.generate_key, name="generate_key"),
-    path("callback/", views.callback, name="callback"),
-    path("consent/", views.consent, name="consent"),
+    path("generate_key", views.generate_key, name="generate_key"),
+    path("callback", views.callback, name="callback"),
+    path("consent", views.consent, name="consent"),
 ]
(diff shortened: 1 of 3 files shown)


snopoke (commented):
Seems OK though I see the participant API has a trailing slash - maybe we should duplicate that without the slash and migrate usages before we get too many

SmittieC:
> Seems OK though I see the participant API has a trailing slash - maybe we should duplicate that without the slash and migrate usages before we get too many

Good idea: https://github.com/dimagi/open-chat-studio/pull/1056/commits/f1cfcb87923f3945616341dbc7153e17c1fbf15c

apps/api/urls.py
```diff
@@ -10,16 +10,18 @@
...
-    path("consent/", views.consent, name="consent"),
+    path("generate_key", views.generate_key, name="generate_key"),
+    path("callback", views.callback, name="callback"),
+    path("consent", views.consent, name="consent"),
 ]
 
 urlpatterns = [
     path("participants/", views.update_participant_data, name="update-participant-data"),
+    # Duplicate update-participant-data without a trailing "/" for backwards compatibility
+    path("participants", views.update_participant_data, name="update-participant-data"),
```
SmittieC, line 21: I confirmed that when we run `reverse("api:update-participant-data")` it returns the route without the slash

snopoke (approved):

PR #1066, opened 2025-01-20
Change ScheduleMessage `TimePeriod.MINUTES` value from `minutes` to `mins`

SmittieC:
Fixes [this issue](https://dimagi.sentry.io/issues/6230129236/?alert_rule_id=14063560&alert_type=issue&notification_uuid=7472caaf-6b53-4ea7-a268-8ed18385fbe5&project=4505001320316928&referrer=issue_alert-slack)

## Description

[This line](https://github.com/dimagi/open-chat-studio/blob/61b9e6b254a0f722d2f6f94cf545f2f84f986d9c/apps/events/actions.py#L79) fails because of the assertion in [this file](https://github.com/dimagi/open-chat-studio/blob/f6bfc6fd8af77da435d93f7543a8ef6c11d82d1c/apps/utils/django_db.py). The value is `minutes` (see [here](https://github.com/dimagi/open-chat-studio/blob/cbdc1d9cf4fd64e4e4547d7b0bcc69017463111b/apps/events/models.py#L303)), but should be `mins`.

The PR ensures that the minutes value is `mins` and updates all existing param values where `minutes` were used to now use `mins`

## User Impact

Users will now be able to update an event schedule to use the `minute` time period.

#### Previously
previous

#### Now
now

### Docs

N/A


SmittieC:
There's no one using "minutes" as the time period in `custom_schedule_params`, so I don't think a data migration is necessary for that

snopoke (approved):

PR #1071, opened 2025-01-21
Link to experiment in participant view

### LINKED ISSUE #1064: Make participant data linkable
Update the participant data UI so that you can link directly to the participant data for a specific experiment.

SmittieC:
Resolves #1064 

I'll mark this PR for review once the base branch has been merged

## User Impact

Users can navigate to participant data pertaining to the specific experiment from the sessions view(s)

### Demo

demo

### Docs

N/A

### DIFF
--- a/apps/api/tasks.py
+++ b/apps/api/tasks.py
@@ -43,8 +43,6 @@ def setup_connect_channels_for_bots(self, connect_id: UUID, experiment_data_map:
 
     connect_client = CommCareConnectClient()
 
-    # TODO: Refactor when experiment_id is directly on the ParticipantData table
-    # https://github.com/dimagi/open-chat-studio/issues/1046
     channels = ExperimentChannel.objects.filter(
         platform=ChannelPlatform.COMMCARE_CONNECT,
         experiment_id__in=[participant_data.experiment_id for participant_data in participant_data],
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -1270,7 +1270,8 @@ class Participant(BaseTeamModel):
         return self.get_link_to_experiment_data(experiment)
 
     def get_link_to_experiment_data(self, experiment: Experiment) -> str:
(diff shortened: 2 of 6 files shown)


apps/api/tasks.py
```diff
@@ -43,8 +43,6 @@ def setup_connect_channels_for_bots(self, connect_id: UUID, experiment_data_map:
 
     connect_client = CommCareConnectClient()
 
-    # TODO: Refactor when experiment_id is directly on the ParticipantData table
```
SmittieC, line 46: I tried prefetching the experiment channel as well and doing
```python
channel = experiment.experimentchannel_set.get(platform=ChannelPlatform.COMMCARE_CONNECT)
```
in the for loop, but through inspecting the number of DB queries that we then run I can see the "refactored" version runs more queries than this version, so I'm leaving this as-is for now.

snopoke (commented):
Just a question about the approach - could this not be done using url hashes e.g `/#exp-1234`? That is how we do it for the tabs on the experiment home page. Changing the tab also updates the URL hash so you can copy / paste the url which is nice.

SmittieC:
> Just a question about the approach - could this not be done using url hashes e.g `/#exp-1234`? That is how we do it for the tabs on the experiment home page. Changing the tab also updates the URL hash so you can copy / paste the url which is nice.

Interesting question. Let me try that

SmittieC:
@snopoke I took a little bit of a different approach. Let me know if you think we should rather do the hash approach.

snopoke:
So with this new approach each experiment has it's own unique URL and we reload the page each time. I guess this is OK and probably not worth optimising since most participants will only interact with one or 2 experiments. One issue I noticed is that when you click on a row the view is reset to the top of the page and you have to scroll back down to get to the experiment - could we add an anchor and a hash to force the scoll to the correct place?

SmittieC:
Scrolling to the selected experiment

demo

snopoke (approved):

PR #1072, opened 2025-01-21
Ensure commcare connect identifiers are upper case

SmittieC:
Solves [this issue](https://dimagi.sentry.io/issues/6223705650/?notification_uuid=a661e3ac-b2cd-4078-bac0-d0bec85b3c70&project=4505001320316928&referrer=regression_activity-slack)

## Description

CommCareConnect identifiers turns out to be case sensitive. We got a 404 using a lower case identifier. I manually retried that call with the upper case version of that identifier and it passed.

CommCare sends lower case identifiers, but I think the API will be more robust if we ensure that it is upper-case

## User Impact

Things should work now

### Demo

N/A

### Docs

N/A


snopoke (approved):

snopoke:
@calellowitz FYI

PR #1075, opened 2025-01-21
Iterate the queryset iso the list

SmittieC:


SmittieC:
The final fix can be found [here](https://github.com/dimagi/open-chat-studio/pull/1078)

PR #1078, opened 2025-01-22
Sk/fix deletion

snopoke:
## Description
Prevent objects from being updated which do not match the queryset filters.

```python
combined_updates.update = field.model.objects.update
```

This was an issue with method binding (`field.model.objects.update` is bound to a queryset with no filtering). This could have been solved by binding the `.update` method to the `combined_updates` queryset however that introduced another error because of the `super()` calls in `AuditQuerySet.update`

## User Impact
Allow safe deletion of teams.


SmittieC (approved):
:guitar:

PR #1069, opened 2025-01-20
Experiment FK on ParticipantData

SmittieC:
Resolves #1046 

## Description

- Added an `Experiment` relation to `ParticipantData`
- I left the `object_id` and `content_type fields there for now, but made them optional so we can still reference them if need be
- I updated `experiment.participant_data` to `experiment.participantdata_set`, since the `GenericFK` reserved `participant_data`. Should be fine
- There's some refactoring I will do in a next PR in which I'm also planning to do https://github.com/dimagi/open-chat-studio/issues/1064

## User Impact

N/A

### Demo

N/A

### Docs

N/A

PR #1084, opened 2025-01-24
fix PD to show experiments without sessions

snopoke:
This was broken by the recent data model refactor: https://github.com/dimagi/open-chat-studio/pull/1069

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -1272,20 +1272,20 @@ class Participant(BaseTeamModel):
             message_type="human",
             chat__experiment_session__experiment__id=OuterRef("id"),
         )
-        joined_on = exp_scoped_human_message.order_by("created_at")[:1].values("created_at")
         last_message = exp_scoped_human_message.order_by("-created_at")[:1].values("created_at")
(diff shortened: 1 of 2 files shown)


apps/experiments/models.py
```diff
@@ -1272,20 +1272,20 @@ def get_experiments_for_display(self):
...
             )
-            .filter(Q(sessions__participant=self) | Q(participant_data__participant=self))
+            .filter(Q(sessions__participant=self) | Q(id__in=Subquery(self.data_set.values("experiment"))))
             .distinct()
         )
 
     def get_data_for_experiment(self, experiment) -> dict:
         try:
-            return self.data_set.get(bots=experiment).data
+            return self.data_set.get(experiment=experiment).data
```
SmittieC, line 1288: Huh, `bots=`. Wonder what that's about

SmittieC (approved):
Thanks

apps/experiments/models.py
snopoke: It's the related query name from the `participant_data` field on the `Experiment` model

Issue #1048, opened 2025-01-16
Show nested changes in version diff view


snopoke:
So that you can see diff of changes to assistants / pipelines or other nested models.

SmittieC:
Done. [This ](https://github.com/dimagi/open-chat-studio/pull/1102) was the last PR

PR #1085, opened 2025-01-24
Nested diff view | part 1

SmittieC:
[Issue](https://github.com/dimagi/open-chat-studio/issues/1048)

## Description

- I did a mini refactor that I hope will make this code a easier to navigate.
- To achieve nested diffs, we need to keep track of each versioned model's changed fields, which is being done by implementing the `version` property for each model. I still need to look at `ExperimentRoute` and `Triggers`/`Events`.

Note that comparing two models can be done in two different ways:
1. Implement the version property and call `.compare()`
2. Call `model1.compare_with_model(model2)`. This is also called inside `.compare()` for versioned fields that does not have the `version` property.

The goal is to get rid of the `model1.compare_with_model(model2)` way of doing things...maybe.

I initially thought to - instead of implementing the `version` property for versioned models - why not construct it on the fly, since we'd have to maintain that property if we add new fields to the model.
On one hand, it would be nice to build that `version` property on the fly with the model's fields, excluding some internal / version related fields but then we need to maintain a list of attributes and put the `to_display` callables that we use for some fields somewhere.
On the other hand, I quite like the fact that we need to be explicit about what we consider part of a version and also the display callables can be gathered together inside that property definition.

Thougths welcome.

## User Impact

Why a version changed should be more obvious now.

### Demo

demo

### Docs

N/A

### To do in followup PRs
- Address the `ExpermientRoute`, `EventAction` and trigger models
- UI improvements


SmittieC:
@coderabbitai please review

SmittieC:
@coderabbitai do a full review again please

snopoke (commented):
I haven't been through this in detail but a few thoughts:

- the version code adds a lot to the models, it would be nice to separate it out into independent code e.g. `get_version_details(experiment)` instead of `experiment.version`
- nit: rename `Version` to `VersionDetails` since it's not really a version of the model (also the property `.version`)
- `VersionField` should be able to get the raw value from the instance directly (or have it set by the `Version` class instead of having to pass it in each time:

```python
# this could set the raw value from the model instance
Version(instance, fields=[VersionField(name="name")])

# even better (allow different types):
Version(instance, fields=["name", VersionField(name=..., queryset=...)])
```

apps/pipelines/models.py
```diff
@@ -116,8 +117,17 @@ def update_nodes_from_data(self) -> None:
...
+
+        pipeline_nodes = Node.objects.annotate(versions_count=models.Count("versions")).filter(
+            pipeline=self, flow_id__in=to_remove
+        )
+        nodes_to_archive = pipeline_nodes.filter(versions_count__gt=0)
+        pipeline_nodes.filter(versions_count=0).delete()
+
+        for node in nodes_to_archive:
+            # Preserve the node if it has versions, otherwise we tamper with previous versions
+            node.archive()
```
stephherbers, line 130: I think one thing we need to iron out still with versioning is when to delete-- I think originally the workflow was active -> archived -> deleted, but this would make it such that a node would not be able to be deleted once archived. I think its out of scope for this PR since it's already reaching was past just a diff view but working discussing

stephherbers:
+1 to Simon's thoughts of separating out from the models and the rename from Version

apps/pipelines/models.py
```diff
@@ -325,6 +335,24 @@ def get_static_trigger_experiment_ids(self) -> models.QuerySet:
...
+        reserved_types = ["StartNode", "EndNode"]
+
+        return Version(
+            instance=self,
+            fields=[
+                VersionField(name="name", raw_value=self.name),
+                VersionField(
+                    name="nodes",
+                    queryset=self.node_set.exclude(type__in=reserved_types),
+                    to_display=VersionFieldDisplayFormatters.format_nodes,
```
snopoke, line 351: Do you still need this formatter now that it's a queryset?

apps/pipelines/models.py
```diff
@@ -395,6 +423,28 @@ def archive(self):
                 assistant = OpenAiAssistant.objects.get(id=assistant_id)
                 assistant.archive()
 
+    @property
+    def version(self) -> Version:
+        from apps.experiments.models import VersionFieldDisplayFormatters
+
+        node_name = self.params.get("name", self.flow_id)
```
snopoke, line 430: The default name is the node ID which is quite ugly. A 'to_display' function for this would be nice to apply similar logic to the UI:
```python
node_type = node_cls.model_config['json_schema_extra'].label
if not name or name == node_id:
    return node_type
else:
    return f"{name} ({node_type})"
```

apps/utils/models.py
```diff
@@ -15,21 +15,28 @@ def compare_with_model(self, new: Self, exclude_fields: list[str], early_abort=F
                 # These are reverse relations, so let's ignore them
                 continue
 
-            current_values = getattr(self, field.attname)
-            new_values = getattr(new, field.attname)
             if field.many_to_many or field.one_to_many:
+                current_values = getattr(self, field.attname)
+                if new:
+                    new_values = getattr(new, field.attname) if new else None
```
snopoke, line 21: if statement above already checks for falsy `new`
```suggestion
                    new_values = getattr(new, field.attname)
```

snopoke (commented):
Some small comments but otherwise LGTM. 

I think further refactoring can wait till a future PR and we should have a design session first.

apps/experiments/models.py
```diff
@@ -213,6 +290,17 @@ def archive(self):
         super().archive()
         self.experiment_set.update(source_material=None, audit_action=AuditAction.AUDIT)
 
+    @property
+    def version(self) -> Version:
+        return Version(
+            instance=self,
+            fields=[
+                VersionField(name="topic", raw_value=self.topic),
+                VersionField(name="description", raw_value=self.description),
```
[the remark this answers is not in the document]
SmittieC: Hmm, fair question. I guess users can ultimately choose to version if they want to with only description changes

apps/pipelines/models.py
SmittieC: Correct, thanks

apps/pipelines/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1085/commits/cf01f5ff9b10048676c439f6735dee55b97335e1

snopoke (approved):

Issue #1039, opened 2025-01-14
Improved multimedia support for pipelines


snopoke:
Allow access to multimedia from users in the pipeline nodes.

>  the ability to process an uploaded file from a user and add the contents to the context window. This is specifically for the XX project. We found that uploading and handing it to the code assistant it would only read the first 2000 characters or so, no matter what we instructed it.

snopoke:
* Attachments are exposed via the `attachments` key in `shared_state`:

```python
get_state_key("attachments")
```

This will return a list of [attachment objects](https://github.com/dimagi/open-chat-studio/blob/ebf965e34e48ac0d846390c69f6be3eca7877ff4/apps/channels/datamodels.py#L12)
 
* Assistant node won't attach media to threads by default
* Users will need to use a code node to specify which attachments to upload

```python
for att in get_state_key("attachments"):
    att.upload_to_assistant = True
```

snopoke:
> Assistant node won't attach media to threads by default

@bderenzi should this apply to attachments for 'code_interpreter' and 'file_search'?

bderenzi:
Definitely for `code_interpreter`. I think probably yes (just to be consistent) for `file_search` too.

PR #1087, opened 2025-01-24
updates to attachment handling in pipelines

snopoke:
Resolves https://github.com/dimagi/open-chat-studio/issues/1039

## Description
From the ticket:

* Attachments are exposed via the `attachments` key in `shared_state`:

```python
get_state_key("attachments")
```

This will return a list of [attachment objects](https://github.com/dimagi/open-chat-studio/blob/ebf965e34e48ac0d846390c69f6be3eca7877ff4/apps/channels/datamodels.py#L12)
 
* Assistant node won't attach media to threads by default
* Users will need to use a code node to specify which attachments to upload

```python
for att in get_state_key("attachments"):
    att.upload_to_assistant = True
```

In addition, the attachment content should be accessible in the Python Node:

```python
def main(input, **kwargs):
    attachment = get_state_key("attachments")[0]
    file_content = attachment.read_string()
    return f'Here is the file content {file_content}'
```

## User Impact
Bot authors have full control over how attachments are handled and can access attachment data via the python node.

### DIFF
--- a/apps/channels/datamodels.py
+++ b/apps/channels/datamodels.py
@@ -17,7 +17,7 @@ class Attachment(BaseModel):
     file_id: int
     type: AttachmentType
     name: str
-    size: int
+    size: int = Field(..., ge=0)
     content_type: str = "application/octet-stream"
 
     upload_to_assistant: bool = False
--- a/apps/channels/datamodels.py
+++ b/apps/channels/datamodels.py
@@ -6,10 +6,25 @@ from pydantic import BaseModel, ConfigDict, Field, field_validator
 from apps.channels.models import ChannelPlatform
 from apps.chat.channels import MESSAGE_TYPES
 
+AttachmentType = Literal["code_interpreter", "file_search"]
+
 
 class Attachment(BaseModel):
     file_id: int
-    type: Literal["code_interpreter", "file_search"]
+    type: AttachmentType
+    name: str
+    size: int
+    content_type: str = Field(default="application/octet-stream")
+
+    @classmethod
+    def from_file(cls, file, type: AttachmentType):
+        return cls(
+            file_id=file.id,
+            type=type,
+            name=file.name,
+            size=file.content_size,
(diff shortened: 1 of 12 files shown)


snopoke:
@coderabbit review

apps/channels/datamodels.py
```diff
@@ -1,13 +1,63 @@
...
+
+    def read_bytes(self):
+        if not self._file:
+            return b""
+        return self._file.file.read()
+
+    def read_string(self):
+        # TODO: error handling for read errors or decode errors
+        # TODO: large file handling (maybe limit file size?)
+        content = self.read_bytes()
```
snopoke, line 59: This needs to support PDF files at a minimum.

SmittieC (approved):
What is the intended usecase for the code node to deliberately have to enable file uploads? Is it so users can upload attachments to one assistant in the pipeline, but not the other?

apps/channels/datamodels.py
SmittieC: > # TODO: large file handling (maybe limit file size?)

I'm wondering if the LLM should know about these kinds of errors to be able to give some feedback to the user.

snopoke:
> What is the intended usecase for the code node to deliberately have to enable file uploads? Is it so users can upload attachments to one assistant in the pipeline, but not the other?

The current use case is that the assistant doesn't 'read' the full file (regardless of what you tell it to do). If you use 'file_search' then it indexes the file and can answer questions about it but can't do a full review. If you use 'code_interpreter' it will read the first portion or last portion but not the whole file.

So if you want it to do a document review the best approach is to pass the file text contents as part of the message instead of as an attachment. This is accomplished using the Python Node to read the attachment's contents and append it to the input message.

The thinking behind forcing one to use a Python Node to 'enable' attachments is that it makes the bot builder specifically decide how files are handled since the failure mode for not reading the whole file is hard to detect.

Issue #1081, opened 2025-01-23
Rename 'shared_state' to 'temp_state'


snopoke:
In pipelines, rename the 'shared_state' variable and related public methods to 'temp_state' (or 'temp_data') to make it clearer that it is temporary.

* CodeNode functions
* StaticRouter references

PR #1090, opened 2025-01-28
Rename 'shared state' to 'temp state' for clarity on temporary nature

snopoke:
Resolves https://github.com/dimagi/open-chat-studio/issues/1081


SmittieC (approved):
an alternative could be `ephemeral_state`

snopoke:
> an alternative could be `ephemeral_state`

That would also work but 'temporary' is simpler and easier to understand than 'ephemeral'

PR #1100, opened 2025-01-29
Fix: Check ids only when object is present

SmittieC:
Fixes https://dimagi.sentry.io/issues/6255736023/?environment=production&project=4505001320316928&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=1h&stream_index=0


SmittieC:
This is to fix an issue on prod that might cause issues for certain bots. I added a test to replicate the issue (which is confirmed fixed). Ima merge this

PR #1101, opened 2025-01-29
add help links for pipelines + fix others

snopoke:


stephherbers (approved):
maybe we can add a node to check for these help links in the docs repo in case someone changes it there and they would then need to change it here

SmittieC (approved):

SmittieC:
> maybe we can add a node to check for these help links in the docs repo in case someone changes it there and they would then need to change it here

node or note? @stephherbers

stephherbers:
thanks typo *note @SmittieC @snopoke

PR #1107, opened 2025-01-30
Add LLM provider to required fields for OpenAIAssistant

stephherbers:
flagged from an ops channels error

## Description

Before this, an assistant form could be entered without an LLM provider or model, but then would fail immediately on `push_assistant_to_openai` causing a "shucks" page

## User Impact

now it will fail gracefully

### Demo

Screenshot 2025-01-30 at 11 43 44 AM

### Docs

n/a


SmittieC (approved):

snopoke (commented):
Should we also add checks in the sync code for safety?

PR #1108, opened 2025-01-30
Checks related models before deleting service provider

stephherbers:
## Description

tests in progress

right now only unreleased and published experiments and assistants will be displayed to the user

## User Impact

This prevents users from accidentally sabotaging other experiments when deleting a service provider with warning on which models will be affected

### Demo

to-do

Screenshot 2025-01-30 at 4 23 08 PM

### Docs

to do


SmittieC (commented):
nice

snopoke (approved):
nice

PR #1125, opened 2025-02-03
rename `Attachment.read_string` to `Attachment.read_text`

snopoke:
## Description
I think `read_text` is more self explanatory and easier for non-programmers to understand.

## User Impact
This will break any existing Python Nodes using the `read_string` function however there are no usages of them currently.

### Docs
Already updated in docs


SmittieC (approved):
Just need to update the schema

SmittieC (approved):

PR #1128, opened 2025-02-04
WIP: Some versioning refactor

SmittieC:
This PR is not meant to be merged, but rather to evoke discussion.

The current versioning code is too complex. Here's some initial code to convey a different approach. Do let me know what you think.

The `InstanceDiff` would typically be the main entry point, so maybe start there if you want to build a mental model

The main things to focus on is the code in the `versioning.py` file and the new `versioning_fields`, `get_versioned_field_value` and `get_versioned_fields_for_display` methods.

Basically, we need 1. a way to know which fields to get, 2. we need to get the values for those fields. Sometimes it's not just a simle `getattr`, but a more comlex query, or in the case of triggers where we want to fetch action params, we need to do that in this step. Lastly, method 3 is for display purposes.

I think here we can do the same as we did with `VersionField` where we return a single wrapper that knows how to display the current value, but I still need to do that. Anycase, leaving this as-is for now.

### DIFF
--- a/apps/assistants/models.py
+++ b/apps/assistants/models.py
@@ -276,6 +276,10 @@ class OpenAiAssistant(BaseTeamModel, VersionsMixin, CustomActionOperationMixin):
             ],
         )
 
+    @property
+    def versioned_fields(self):
+        return [field.name for field in self.version_details.fields]
+
 
 @audit_fields(
     "assistant_id",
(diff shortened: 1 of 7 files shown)


stephherbers:
+1 to the declarative style-- it would help with the complexity issue with clearly defined fields and how they should be handled. It cuts down on scattered logic, keeps complex queries neatly tucked into field definitions, and makes future updates easier than they have been previously. Plus as you mentioned, it would keep everything in one place

PR #1132, opened 2025-02-04
[bugfix] select correct node input when multiple are possible

snopoke:
## Description
In the following graph the `end` node needs to select between the `router` and `template` output when the router routes through the template:

```
start --> router -+-> template --> end
                  |                 ^
                  +-----------------+
```

The method we had for doing this before would select the first available input which is dependent on the order in which the edges are created in the graph.

This PR keeps that mechanism as a fallback but prefers to use the `langgraph_triggers` metadata to select the input.

## User Impact
Pipelines operate as expected

### Demo
See test

### Docs and Changelog
NA

### DIFF
--- a/apps/pipelines/nodes/base.py
+++ b/apps/pipelines/nodes/base.py
@@ -1,3 +1,4 @@
+import logging
 import operator
 from abc import ABC
 from collections.abc import Sequence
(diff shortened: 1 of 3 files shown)


SmittieC (approved):
LGTM. IMHO it would be great if we can have methods on the state in order to do stuff like

```python
state.get_node_output(node) # where node is node_id or something
```

snopoke:
> LGTM. IMHO it would be great if we can have methods on the state in order to do stuff like
> 
> ```python
> state.get_node_output(node) # where node is node_id or something
> ```

A challenge is that the state doesn't have enough information to conclusively know

PR #1142, opened 2025-02-05
External channels should not resue sessions that ended

SmittieC:
## Description

I realized that when we end a session, say with a timeout trigger, then the next time that the participant sends a message, that same session would be re-used. I added logic to exclude sessions with the ended status when we try and find existing sessions.

This change will also make [Jon's request](https://dimagi.slack.com/archives/C05PK63Q89H/p1738705116418499) possible

## User Impact

Users would be able to use timeout triggers to end participant sessions. What this mean is that the next time the participant chats to the bot, it would be as if it was the first time.

### Demo

N/A

### Docs and Changelog

I'll add a page on sessions. Stay tuned for the PR link


SmittieC:
This makes me wonder, should the bot send a "Hi, just letting you know that I'm ending this session" message just before ending it? This might prevent potential confustion from the participant's side

snopoke (approved):

Issue #1144, opened 2025-02-05
Allow sessions to be reset via the API


snopoke:
Update the [session](https://chatbots.dimagi.com/api/docs/#tag/Experiment-Sessions) API to allow ending the session.

Update the [ExperimentSessionViewSet](https://github.com/dimagi/open-chat-studio/blob/main/apps/api/views.py#L288) to add an [custom action](https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing) to end the session by calling `session.end()`. It is not necessary to create a new session as one will be created when needed.

---

### Already done

Add a `start_new_session` flag to the [trigger bot message](https://chatbots.dimagi.com/api/docs/#tag/Channels/operation/trigger_bot_message). When set to 'True' this will end the current session if there is one and start a new session for the user.

SmittieC:
@snopoke do we still want an endpoint to end sessions ie Option 2?

snopoke:
I think it would be useful but we can de-prioritize for now

vaibhavK1199:
.take

vaibhavK1199:
@snopoke  MR for review - https://github.com/dimagi/open-chat-studio/pull/1210

vaibhavK1199:
@snopoke New MR added for the review - https://github.com/dimagi/open-chat-studio/pull/1213

PR #1145, opened 2025-02-05
Support for resetting sessions in the trigger bot message endpoint

SmittieC:
## Description

resolves part 1 of #1144 

## User Impact

Users will be able to optionally use new sessions when hitting the trigger-bot-message API.

### Demo

N/A

### Docs and Changelog

The api schema is updated with an example

### DIFF
--- a/apps/api/views.py
+++ b/apps/api/views.py
@@ -417,6 +417,18 @@ def consent(request: Request):
             status_codes=[200],
         ),
         OpenApiExample(
+            name="GenerateBotMessageAndSendUsingANewSession",
+            summary="Generates a bot message and sends it to the user using a new session",
+            value={
+                "identifier": "part1",
+                "experiment": "exp1",
+                "platform": "connect_messaging",
+                "prompt_text": "Tell the user to do something",
+                "start_new_session": "true",
+            },
+            status_codes=[200],
+        ),
+        OpenApiExample(
             name="ParticipantNotFound",
             summary="Participant not found",
             value={"detail": "Participant not found"},
--- a/apps/api/views.py
+++ b/apps/api/views.py
@@ -467,14 +467,14 @@ def trigger_bot_message(request):
 
     experiment = get_object_or_404(Experiment, public_id=experiment_public_id, team=request.team)
 
-    try:
-        participant_data = ParticipantData.objects.defer("data").get(
-            participant__identifier=identifier,
-            participant__platform=platform,
-            experiment=experiment.id,
-        )
-
-    except ParticipantData.DoesNotExist:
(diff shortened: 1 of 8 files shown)


apps/api/views.py
```diff
@@ -416,6 +416,18 @@ def consent(request: Request):
...
+            summary="Generates a bot message and sends it to the user using a new session",
+            value={
+                "identifier": "part1",
+                "experiment": "exp1",
+                "platform": "connect_messaging",
+                "prompt_text": "Tell the user to do something",
+                "start_new_session": "true",
+            },
+            status_codes=[200],
+        ),
```
snopoke, line 430: Do you find these useful in the docs? I'm not sure it's worth having too many examples.

apps/api/views.py
```diff
@@ -455,14 +467,14 @@ def trigger_bot_message(request):
...
-
-    except ParticipantData.DoesNotExist:
+    participant_data = ParticipantData.objects.filter(
+        participant__identifier=identifier,
+        participant__platform=platform,
+        experiment=experiment.id,
+    ).first()
+    if platform == ChannelPlatform.COMMCARE_CONNECT and not participant_data:
+        return Response({"detail": "Participant not found"}, status=status.HTTP_404_NOT_FOUND)
+    elif not Participant.objects.filter(identifier=identifier, platform=platform).exists():
```
snopoke, line 477: Can you give more context on this change - it doesn't seem related to the rest of the PR.

snopoke (approved):

apps/api/views.py
SmittieC: This endpoint will not work for non-connect channels unless they have particpant data, which is typically the case for non-connect channels.

apps/api/views.py
SmittieC: Personally no, but I generally like examples. We can take it out if you feel it's cluttering it?

apps/api/views.py
snopoke: I think one or two is great but this one doesn't add much value over the basic one. It just shows the one additional field.

apps/api/views.py
snopoke: I think one or two is great but this one doesn't add much value over the basic one. It just shows the one additional field.

apps/api/views.py
snopoke: A comment in code would be useful - is it because connect is a 'push' channel? Maybe we can make that more explicit.

SmittieC: The main reason that connect requires participant data is because we save the consent, channel id and encryption key on there. I'll add a comment

PR #1151, opened 2025-02-05
Remove delete as an option from OpenAiAssistant table

stephherbers:
## Description

You can't delete without archiving first

## User Impact

archiving really achieves what they want 

### Demo

Screenshot 2025-02-05 at 9 29 52 AM

### Docs and Changelog

not needed

### DIFF
--- a/apps/assistants/tables.py
+++ b/apps/assistants/tables.py
@@ -31,14 +31,6 @@ class OpenAiAssistantTable(tables.Table):
                 confirm_message="This will delete the assistant from OpenAI.",
                 hx_method="delete",
             ),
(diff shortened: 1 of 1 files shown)


apps/assistants/tables.py
```diff
@@ -31,14 +31,6 @@ class OpenAiAssistantTable(tables.Table):
                 confirm_message="This will delete the assistant from OpenAI.",
                 hx_method="delete",
             ),
-            actions.AjaxAction(
-                "assistants:delete",
```
snopoke, line 35: can we remove the view too?

snopoke (approved):

Issue #1097, opened 2025-01-28
Clean up OpenAI Assistants when team is deleted


snopoke:
* Make team deletion async with email confirmation when done
* Update confirmation dialog to indicate that all referenced openai assistants will be deleted

PR #1157, opened 2025-02-06
Display which Assistants will be deleted with team deletion

stephherbers:
Second part of this ticket https://github.com/dimagi/open-chat-studio/issues/1097

## Description

(Just a warning not blocking like for archiving) This informing the the assistants will be deleted and which ones for the user and links to them if they so desire to look at them, or they can ignore them completely and still delete the team 

## User Impact

users are more informed when delete teams

### Demo

Screenshot 2025-02-06 at 2 22 32 PM

### DIFF
--- a/apps/teams/views/manage_team_views.py
+++ b/apps/teams/views/manage_team_views.py
@@ -21,7 +21,6 @@ from apps.web.forms import set_form_fields_disabled
 
 def get_related_assistants(team):
     related_assistants = OpenAiAssistant.objects.filter(team=team)
-    print(related_assistants)
     return [Chip(label=assistant.name, url=assistant.get_absolute_url()) for assistant in related_assistants]
 
 
--- a/apps/teams/views/manage_team_views.py
+++ b/apps/teams/views/manage_team_views.py
@@ -7,6 +7,8 @@ from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 from django.views.decorators.http import require_POST
 
+from apps.assistants.models import OpenAiAssistant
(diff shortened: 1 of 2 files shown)


apps/teams/views/manage_team_views.py
```diff
@@ -17,6 +19,12 @@
 from apps.web.forms import set_form_fields_disabled
 
 
+def get_related_assistants(team):
+    related_assistants = OpenAiAssistant.objects.filter(team=team)
+    print(related_assistants)
```
SmittieC, line 24: Print statement

templates/teams/components/delete_team_modal.html
```diff
@@ -9,6 +9,14 @@ <h3 class="font-bold text-lg">{% translate "Really delete team?" %}</h3>
         <strong>This action cannot be undone.</strong>
       {% endblocktranslate %}
     </p>
+    {% if related_assistants %}
+      <h4 class="font-semibold my-2">These OpenAI Assistants will also be deleted:</h4>
```
SmittieC, line 13: Wondering if we should tell users that we'll remove the assistants from OpenAI as well?

snopoke: +1 - explicit is better and this could be interpreted as we'll delete them from OCS only
```suggestion
      <h4 class="font-semibold my-2">These OpenAI Assistants will also be deleted from OpenAI along with the related files and vector stores:</h4>
```

snopoke (dismissed):
Looks good - one comment on wording

snopoke (approved):

PR #1169, opened 2025-02-07
Async team deletion with email

stephherbers:
## Description

first part of this ticket: https://github.com/dimagi/open-chat-studio/issues/1097
Screenshot 2025-02-18 at 10 19 24 AM

## User Impact

All teams admins will receive emails and it will be more clear to the delete when the team is actually finished deleted rather than seemingly instant

Users will also be able to select whether to send the email to just themeselves (default) to the team admins, or to all members of the domain via a dropdown. 

### Demo

Loom demo [here](https://www.loom.com/share/046313ff19644c5e8b9883124054706c?sid=bf18adf3-28a3-49c7-b5c1-9d552aabdc5d)

### Docs and Changelog

will add 👍


gpt_playground/settings.py
```diff
@@ -548,6 +548,7 @@
     "account_reset_password_from_key",
     "teams:signup_after_invite",
     "account_login",
+    "teams:delete_team",
```
stephherbers, line 551: doesn't fix the "No team found for audit context" errors since the request is `None` [here](https://github.com/dimagi/open-chat-studio/blob/main/apps/audit/auditors.py#L22) and then by [this check](https://github.com/dimagi/open-chat-studio/blob/main/apps/audit/auditors.py#L81) it does not return a view name

apps/utils/deletion.py
```diff
@@ -218,3 +222,27 @@ def get_related_pipelines_queryset(instance, pipeline_param_key: str = None):
...
+    from apps.teams.models import Membership
+
+    return list(
+        Membership.objects.filter(team__name=team.name, groups__permissions__codename="delete_team").values_list(
+            "user__email", flat=True
+        )
+    )
+
+
+def send_domain_deleted_notification(team_name, admin_emails):
```
snopoke, line 237: nit: `domain` -> `team`
```suggestion
def send_team_deleted_notification(team_name, admin_emails):
```

gpt_playground/settings.py
```diff
@@ -548,6 +548,7 @@
     "account_reset_password_from_key",
     "teams:signup_after_invite",
     "account_login",
+    "single_team:delete_team",
```
snopoke, line 551: Why was this necessary?

templates/teams/email/domain_deleted_notification.html
snopoke, line 1: nit: rename `domain` -> `team`

templates/teams/email/domain_deleted_notification.txt
snopoke, line 1: nit: rename `domain` -> `team`

apps/utils/deletion.py
```diff
@@ -218,3 +222,27 @@ def get_related_pipelines_queryset(instance, pipeline_param_key: str = None):
...
+    email_context = {
+        "team_name": team_name,
+    }
+    send_mail(
+        subject=_("Domain '{}' has been deleted").format(team_name),
```
snopoke, line 242:
```suggestion
        subject=_("Team '{}' has been deleted").format(team_name),
```

apps/utils/deletion.py
```diff
@@ -218,3 +222,27 @@ def get_related_pipelines_queryset(instance, pipeline_param_key: str = None):
         Q(**{f"params__{pipeline_param_key}": instance.id}) | Q(**{f"params__{pipeline_param_key}": str(instance.id)})
     )
     return pipelines
+
+
+def get_admin_emails(team):
```
snopoke, line 227: this is a good candidate for a test

snopoke (commented):
Few small comments but looks good

apps/teams/tasks.py
```diff
@@ -1,10 +1,26 @@
...
     send_invitation_accepted(invitation)
+
+
+@shared_task
+def delete_team_async(team_id):
+    team = Team.objects.get(id=team_id)
+
+    # get required info for emails prior to deleting team info
+    team_name = team.name
+    admin_emails = get_admin_emails_with_delete_permission(team)
```
SmittieC, line 24: Should all users maybe be notified instead of only those that can delete the team? It might be a bit surprising for someone to find their team is gone. Probably an edge case, but still

gpt_playground/settings.py
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/1169#discussion_r1947136972

SmittieC:
I realize you are still busy with this PR, but just flagging that we are not removing assistants from OpenAI yet? A test for `delete_team_async` might also be a good idea to check that 1. assistants are removed from OpenAI and 2. that any archived objects are also removed. I think number 2 is being done already, but might be good to test..up to you

apps/teams/tests/test_get_admin_emails.py
```diff
@@ -21,3 +22,18 @@ def test_get_admin_emails(team_with_users):
     }
     emails = get_admin_emails(team_with_users)
     assert set(emails) == expected
+
+
+@pytest.mark.django_db()
+def test_get_admin_emails_can_delete_team(team_with_users):
+    set_current_team(team_with_users)
```
snopoke, line 29: you could use the `current_team` context manager to ensure that it is unset

apps/teams/tasks.py
```diff
@@ -1,10 +1,26 @@
...
 
 @shared_task(ignore_result=True)
 def send_invitation_accepted_notification(invitation_id):
     invitation = Invitation.objects.get(id=invitation_id)
     send_invitation_accepted(invitation)
+
+
+@shared_task
+def delete_team_async(team_id):
+    team = Team.objects.get(id=team_id)
```
snopoke, line 20: use `with current_team(team)` to set the team for auditing

apps/teams/tasks.py
snopoke: FYI there is a limit of 50 recipients per email so this will need to be chunked. See https://github.com/dimagi/open-chat-studio/pull/1202

apps/teams/tasks.py
```diff
@@ -18,9 +20,11 @@ def send_invitation_accepted_notification(invitation_id):
...
-    # get required info for emails prior to deleting team info
-    team_name = team.name
     admin_emails = get_admin_emails_with_delete_permission(team)
+    team_name = team.name
+    chunk_size = 50
+    chunked_emails = chunk_list(admin_emails, chunk_size)
     delete_object_with_auditing_of_related_objects(team)
-    send_team_deleted_notification(team_name, admin_emails)
+    for chunk_emails in chunked_emails:
+        with current_team(team):
```
snopoke, line 29: I think this should be guarding the `delete_object_with_auditing_of_related_objects` call. I don't think it's needed for the email sending.

apps/teams/tests/test_get_admin_emails.py
```diff
@@ -0,0 +1,23 @@
...
+from apps.teams.backends import TEAM_ADMIN_GROUP
+from apps.teams.utils import set_current_team
+from apps.users.models import CustomUser
+from apps.utils.deletion import get_admin_emails_with_delete_permission
+from apps.utils.factories.team import MembershipFactory
+
+
+@pytest.mark.django_db()
+def test_get_admin_emails_can_delete_team(team_with_users):
+    set_current_team(team_with_users)
```
snopoke, line 13: I think this can be removed

snopoke (approved):
minor comments but nothing blocking

apps/teams/views/manage_team_views.py
```diff
@@ -85,8 +86,14 @@ def create_team(request):
...
 @permission_required("teams.delete_team", raise_exception=True)
 def delete_team(request, team_slug):
-    delete_object_with_auditing_of_related_objects(request.team)
-    messages.success(request, _('The "{team}" team was successfully deleted').format(team=request.team.name))
+    notify_recipients = request.POST.get("notification_recipients", "self")
+    delete_team_async.delay(request.team.id, request.user.email, notify_recipients)
+    messages.success(
+        request,
+        _(
+            'The "{team}" team deletion process has started. An email will be sent to admins once it is complete.'
```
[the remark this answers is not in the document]
stephherbers: good call 👍

snopoke (approved):

PR #1171, opened 2025-02-10
upgrade taskbadger to record celery task args

snopoke:
See https://docs.taskbadger.net/python-celery/#system-integration-options


SmittieC (approved):
This should be useful.

PR #1186, opened 2025-02-13
Tiny form update

SmittieC:
## Description

1. Moved `ExperimentForm` and `ExperimentVersionForm` into the `forms.py` file
2. When creating a new experiment version, the "is default version" checkbox will now be ticked by default. You'll have to untick it if you do not want it to be the default. 

## User Impact

Users will not have to remember to tick that box each time they create a new version. If they missed it, they'd have to find that version in the versions table and go make it the default version.

### Demo

N/A

### Docs and Changelog

Will update the changelog once this is merged

### DIFF
--- a/templates/experiments/create_version_form.html
+++ b/templates/experiments/create_version_form.html
@@ -21,7 +21,11 @@
       {% if not version_details.instance.has_versions or version_details.fields_changed %}
         <form method="post">
           {% csrf_token %}
-          {% render_form_fields form %}
+          {% render_form_fields form "version_description" %}
+          <div class="w-1/6">
+            {% render_form_fields form "is_default_version" %}
+          </div>
+
           <div class="form-actions">
             <p class="buttons my-3">
               <button type="submit" class="btn btn-primary">Create</button>
--- a/apps/experiments/forms.py
+++ b/apps/experiments/forms.py
@@ -1,8 +1,23 @@
 from django import forms
 from django.core import validators
-from django.utils.translation import gettext_lazy
+from django.utils.translation import gettext, gettext_lazy
+from waffle import flag_is_active
 
-from apps.experiments.models import ExperimentRoute, ExperimentRouteType, Survey
+from apps.custom_actions.form_utils import (
+    clean_custom_action_operations,
+    initialize_form_for_custom_actions,
+    set_custom_actions,
+)
+from apps.experiments.models import (
+    AgentTools,
+    Experiment,
+    ExperimentRoute,
+    ExperimentRouteType,
+    Survey,
+    SyntheticVoice,
+)
+from apps.generics.help import render_help_with_link
+from apps.utils.prompt import PromptVars, validate_prompt_variables
 
 
 class ConsentForm(forms.Form):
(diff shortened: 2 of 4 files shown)


snopoke (commented):
Code looks fine but wondering about the change to the default for `publish`. Where did this requirement come from? It seems like quite a big change and could result in users publishing versions unintentionally.

SmittieC:
> Code looks fine but wondering about the change to the default for `publish`. Where did this requirement come from? It seems like quite a big change and could result in users publishing versions unintentionally.

I've been chatting to a user who struggled to get surveys working for their bot. I did mention that they need to create a new version to "apply" these changes to a "production" bot. They did, but did not mark the bot as default, and flagged that the survey is not working.

This got me thinking about the fact that we leave that box unchecked. Does it make sense to create a new version if you're not going to publish it as well? Maybe this question _should_ be asked on the users channel.

snopoke:
> > Code looks fine but wondering about the change to the default for `publish`. Where did this requirement come from? It seems like quite a big change and could result in users publishing versions unintentionally.
> 
> I've been chatting to a user who struggled to get surveys working for their bot. I did mention that they need to create a new version to "apply" these changes to a "production" bot. They did, but did not mark the bot as published in that form, and flagged that the survey is not working.
> 
> This got me thinking about the fact that we leave that box unchecked. Does it make sense to create a new version if you're not going to publish it as well? Maybe this question _should_ be asked on the users channel.

I definitely think we want to check before making this change. CommCare works the same - you make a build and then separately mark it as published. Maybe the UI just needs some adjustment to make it clearer - for one thing the 'Set as Published Version' checkbox isn't very visible - adding some padding to that and bringing the checkbox closer to the label would likely help.

SmittieC:
> > > Code looks fine but wondering about the change to the default for `publish`. Where did this requirement come from? It seems like quite a big change and could result in users publishing versions unintentionally.
> > 
> > 
> > I've been chatting to a user who struggled to get surveys working for their bot. I did mention that they need to create a new version to "apply" these changes to a "production" bot. They did, but did not mark the bot as published in that form, and flagged that the survey is not working.
> > This got me thinking about the fact that we leave that box unchecked. Does it make sense to create a new version if you're not going to publish it as well? Maybe this question _should_ be asked on the users channel.
> 
> I definitely think we want to check before making this change. CommCare works the same - you make a build and then separately mark it as published. Maybe the UI just needs some adjustment to make it clearer - for one thing the 'Set as Published Version' checkbox isn't very visible - adding some padding to that and bringing the checkbox closer to the label would likely help.

> adding some padding to that and bringing the checkbox closer to the label would likely help.

Good idea, let's try this rather

SmittieC:
image

templates/experiments/create_version_form.html
```diff
@@ -21,7 +21,11 @@ <h2 class="pg-subtitle pb-5">A new version being created. Please wait until it i
       {% if not version_details.instance.has_versions or version_details.fields_changed %}
         <form method="post">
           {% csrf_token %}
-          {% render_form_fields form %}
+          {% render_form_fields form "version_description" %}
+          <div class="w-1/6">
```
snopoke, line 25: Does this work OK on smaller screens? Might need to make it bigger by default and then have md:w-1/6 (see https://tailwindcss.com/docs/hover-focus-and-other-states#responsive-breakpoints)

snopoke (approved):
One comment but otherwise 👍🏻

templates/experiments/create_version_form.html
snopoke: mm, not sure I agree - it squashes the text and makes it wrap when it doesn't need to. Try `w-full md:w-1/3 lg:w-1/6`

SmittieC: That does look nicer, thanks. https://github.com/dimagi/open-chat-studio/pull/1188

PR #1192, opened 2025-02-14
Use booleans for toggle nodes

SmittieC:
## Description

Followup on [this PR](https://github.com/dimagi/open-chat-studio/pull/1189/files).

The toggle value for the `citations_enabled` field in the node params should be a boolean, not a string.

The issue that was patched in the linked PR came to be because we default to a `True` value for that parameter. If the user doesn't change that value, the value that was saved in the params was `True`, not `"true"`. Since we cast the boolean to a string only when the value changes, the value's type only changes when the user actually updates that field.

Since we now use and expect boolean values, there is a data migration to cast all string values to boolean.

An issue with this that is now also fixed:

Suppose the user doesn't change the value for the "citations_enabled" toggle when adding an assistant node, they would be misled to think that it is disabled. Here's why: Since the initial value is `True`, the UI sees a boolean value, but because we do this check:
```JS
checked={props.paramValue === "true"}
```
the `checked` attribute would always be `false`, so toggle would be disabled where the server thinks that it was enabled.

Another fix:
We now also check for pipeline errors before creating a version.

## User Impact

- Users will not be misled anymore when adding an assistant node by thinking that citations are disabled. 

### Demo

N/A

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/42


snopoke (approved):

stephherbers:
also resolved [this ticket](https://dimagi.atlassian.net/browse/OCSBUGS-3)-- thanks @SmittieC!

PR #1201, opened 2025-02-18
Sort pipelines by name in ascending order

### LINKED ISSUE #1184: Sort pipelines by name or date
It feels like pipelines are sorted by some hash code we use behind the scenes. New pipelines are showing up randomly in the middle of the list.

SmittieC:
## Description

Resolves #1184 
The pipelines were sorted by `created_at` date in descending order by default. Not sure why this didn't apply in the pipelines table. They are not equally sorted by name in ascending order.

## User Impact

Might be easier to find a pipeline

#### Pipelines table
image

#### Experiment form
image

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/45

### DIFF
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -62,7 +62,7 @@ class Pipeline(BaseTeamModel, VersionsMixin):
     objects = PipelineManager()
 
     class Meta:
-        ordering = ["name"]
+        ordering = ["-created_at"]
 
     def __str__(self):
         if self.working_version is None:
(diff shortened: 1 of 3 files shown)


apps/pipelines/models.py
```diff
@@ -62,7 +62,7 @@ class Pipeline(BaseTeamModel, VersionsMixin):
     objects = PipelineManager()
 
     class Meta:
-        ordering = ["-created_at"]
+        ordering = ["name"]
```
snopoke, line 65: since you're overriding this in the view you could leave it as is to avoid having to create a migration:

snopoke (dismissed):

snopoke (approved):

PR #1203, opened 2025-02-18
open markdown in new tab

### LINKED ISSUE #1116: Render links in markdown so that they open in a new tab
Render links in Markdown with `_target="blank"`

## Details

Update the [render_markdown](https://github.com/dimagi/open-chat-studio/blob/main/apps/chat/templatetags/chat_tags.py#L18) template tag so that all links are rendered with `_target="blank"`

This may be incorporated into the existing link extensions in https://github.com/dimagi/open-chat-studio/blob/main/apps/utils/markdown.py or done separately.

barora05:
## Description

This PR overrides the handleMatch method to automatically add target="_blank" to all <a> tags.
## User Impact. Fixes #1116 

Markdown links would open in new tab


snopoke (commented):
Code changes look good. It would be great to have a test case as well. I don't think we have any test cases for our markdown rendering but it should be simple enough to create a new test file in the `chat` app to test the `apps.chat.templatetags.chat_tags.render_markdown` function

barora05:
Sure, will right down the test cases for this method.

apps/chat/tests/test_chat_tags.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+# Test markdown with special characters
+def test_render_markdown_special_characters(mock_reverse):
+    markdown_text = "Special characters: <, >, &"
+    result = render_markdown(markdown_text)
+    print(result)
+    assert result == "<p>Special characters: &lt;, &gt;, &amp;</p>"
+
+def test_render_image_markdown(mock_reverse):
```
snopoke, line 55: I think this is the only test that needs the `mock_reverse` fixture. It can be removed from the others to simplify them.

apps/chat/tests/test_chat_tags.py
```diff
@@ -0,0 +1,61 @@
...
+    markdown_text = "<div>Some HTML content</div>"
+    result = render_markdown(markdown_text)
+    assert result == "<div>Some HTML content</div>"
+
+
+# Test markdown with special characters
+def test_render_markdown_special_characters(mock_reverse):
+    markdown_text = "Special characters: <, >, &"
+    result = render_markdown(markdown_text)
+    print(result)
```
snopoke, line 52: nit: print

snopoke (dismissed):
small comments but otherwise looks good.

snopoke (approved):

PR #1204, opened 2025-02-18
add theme toggle

### LINKED ISSUE #1165: Add a theme controller toggle
Allow users to switch between dark / light mode.

https://daisyui.com/components/theme-controller/

The theme preference should persist in local strogage.

barora05:
## Description
Adds theme toggle checkbox in MyAccount section of navbar and saves the theme in localstorage. Fixes #1165 
## User Impact
Allows users to change between dark and light mode.

### Demo
[Theme toggle video link](https://drive.google.com/file/d/1zrQhSBGgUjtulUbBbrNkZGpzYoI4aOAt/view?usp=drive_link)


templates/web/base.html
```diff
@@ -70,5 +70,7 @@
     {% include 'web/components/ajax_errors-js.html' %}
     {% load tz_detect %}
     {% tz_detect %}
+    {% load static %}
+  <script src="{% static 'js/theme-toggle.js' %}"></script>
```
snopoke, line 74: Was this file meant to be added?

barora05: yes. theme-toggle.js file has the code for setting the theme and saving it to localstorage.

barora05:
> Could you add a demo to the PR description please. We typically use [loom](https://www.loom.com/) or the native screen recorder (https://www.microsoft.com/en-us/windows/learning-center/how-to-record-screen-windows-11)

I have added the demo link

templates/web/components/app_nav_menu_items.html
```diff
@@ -27,6 +27,11 @@
       {% translate "Sign out" %}
     </a>
   </li>
+  <div class="flex justify-end items-center gap-x-2 px-4 my-2">
+    <i class="fa-solid fa-circle-half-stroke h-4 w-4"></i>
+    <span>Toggle Theme</span>
+    <input type="checkbox" id="theme-toggle" class="toggle">
+  </div>
```
snopoke, line 34: Maybe just personal preference but I prefer this toggle:

image


https://daisyui.com/components/theme-controller/#theme-controller-using-a-toggle-with-icons

templates/web/base.html
barora05: I have added it now. It was ignored earlier as static folder files are not committed.

templates/web/components/app_nav_menu_items.html
barora05: I tried adding this. The issue that I am facing is that after hard refresh the checked state of this checkbox is not maintained. Let's suppose if you are on dark theme, the checkbox would be checked but if you hard refresh the page theme would still remain dark but the checkbox would not be checked. I have added the code of maintaing checked state of toggle in set-theme method but still it doesn't work. I observed the same behavior with show archived checkbox of experiment tab
`function setTheme(theme) {
  document.body.setAttribute("data-theme", theme); // Apply to body
  localStorage.setItem("theme", theme);
  const themeToggle = document.getElementById("theme-toggle-base");
  if (themeToggle) {
    themeToggle.checked = theme === "light"; // Sync checkbox state
  }
 }`

snopoke: OK, no worries. Let's leave it as it is.

static/js/theme-toggle.js
snopoke, line 1: Rather than adding this directly to `static/js` you should add it to `assets/javascript`.

You can then add a line to the `assets/javascript/site.js` file to import the module:

```
import './theme-toggle'
```

This will result in the JS code being bundled into the global `site-bundle.js` file which is loaded on all pages.

To rebuild the JS bundles you can run `inv npm` or `inv npm -w` for the watch version. (Or you can run the npm commands directly: `npm run dev` / `npm run dev-watch`)

static/js/theme-toggle.js
```diff
@@ -0,0 +1,37 @@
+document.addEventListener("DOMContentLoaded", function () {
+
+// Function to set the theme
+function setTheme(theme) {
+  document.body.setAttribute("data-theme", theme); // Apply to body
+  localStorage.setItem("theme", theme);
+  const themeToggle = document.getElementById("theme-toggle-base");
```
snopoke, line 7: I think this should be `theme-toggle`:
```suggestion
  const themeToggle = document.getElementById("theme-toggle");
```

static/js/theme-toggle.js
```diff
@@ -0,0 +1,37 @@
...
+// Function to initialize the theme
+function initializeTheme() {
+  const savedTheme = localStorage.getItem("theme");
+  const systemTheme = window.matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light";
+  const theme = savedTheme || systemTheme;
+  setTheme(theme); // Set the theme and sync the checkbox
+}
+
+// Event delegation for theme toggle
+document.body.addEventListener('change', function (event) {
```
snopoke, line 29: Does this have to listen to all change events on the page? Would it not be better to only listen to change events on the `theme-toggle` element?
```suggestion
const toggle = document.getElementById("theme-toggle");
if (toggle) {
  toggle.addEventListener('change', function (event) {
    toggleTheme();
  }
}
```

static/js/theme-toggle.js
snopoke, line 1: I think this might fail the linter - do you have the pre-commit hooks installed: `pre-commit install --install-hooks`?

Though maybe it's ignoring the file because of it's location.

snopoke (approved):

Issue #1092, opened 2025-01-28
Deepseek integration


SmittieC:
See the [API docs](https://api-docs.deepseek.com/)

This would get added as a new [LLM Provider ](https://github.com/dimagi/open-chat-studio/blob/main/apps/service_providers/models.py#L112)

See also https://github.com/dimagi/open-chat-studio/blob/main/apps/service_providers/README.md

Related to #726

SmittieC:
We might want to hold off on this for a while. https://www.wiz.io/blog/wiz-research-uncovers-exposed-deepseek-database-leak

bderenzi:
I think Teams can make their own decision about whether to use a model or not, but from a platform it seems easy/good to add integration.

barora05:
.take

barora05:
@snopoke can we close this?

PR #1205, opened 2025-02-18
deepseek integration

barora05:
## Description
Integrate deepseek. #1092 

## User Impact
Allow users to use deepseek as LLM model


apps/service_providers/migrations/0024_add_deepseek_models.py
snopoke, line 1: This code is fine though you can also use the `_update_llm_provider_models` utility e.g. https://github.com/dimagi/open-chat-studio/blob/00d0a0958c65013416f0b51bf1898cec05db2d35/apps/service_providers/migrations/0021_migrate_llm_models.py#L8-L11

apps/service_providers/llm_service/token_counters.py
```diff
@@ -67,3 +67,23 @@ def get_tokens_from_text(self, text) -> int:
         tokenizer = sync_get_tokenizer()
         encoded_text = tokenizer.encode(text)
         return len(encoded_text.ids)
+
+class DeepSeekTokenCounter(TokenCounter):
```
snopoke, line 71: This seems identical to the OpenAI one - any reason not to reuse that one?

apps/service_providers/llm_service/token_counters.py
```diff
@@ -67,3 +67,23 @@ def get_tokens_from_text(self, text) -> int:
...
+            return None
+
+        token_usage = response.llm_output["usage"]
+        output_tokens = token_usage.get("completion_tokens", 0)
+        input_tokens = token_usage.get("prompt_tokens", 0)
+
+        return input_tokens, output_tokens
+
+    def get_tokens_from_text(self, text) -> int:
+        tokenizer = sync_get_tokenizer()
```
snopoke, line 86: I don't think this should use the the anthropic tokenizer - I see they have their own tokenizer (https://api-docs.deepseek.com/quick_start/token_usage) but I'm not sure we want to have that in the repo.

We don't currently do much with these token counts so for now we could just use the OpenAI one and maybe add a note for future reference.

apps/service_providers/llm_service/default_models.py
```diff
@@ -73,6 +73,10 @@ def k(n: int) -> int:
         Model("llama-3.1-8b-instruct", 131072),
         Model("llama-3.1-70b-instruct", 131072),
     ],
+    "deepseek": [
+        Model("deepseek-v3", 128000),
+        Model("deepseek-r1", 128000)
```
snopoke, line 78: These should be `deepseek-chat` and `deepseek-reaoner`: https://api-docs.deepseek.com/quick_start/pricing

apps/service_providers/forms.py
```diff
@@ -131,6 +131,17 @@ class AnthropicConfigForm(ObfuscatingMixin, ProviderTypeConfigForm):
         initial="https://api.anthropic.com",
     )
 
+class DeepSeekConfigForm(ObfuscatingMixin, ProviderTypeConfigForm):
+    obfuscate_fields = ["deepseek_api_key"]
+
+    deepseek_api_key = forms.CharField(label=_("API Key"))
+    deepseek_api_base = forms.URLField(
```
snopoke, line 138: I don't think this field is necessary since it there is only one possible value. You can rather just provide the URL statically when instantiating the client.

apps/service_providers/migrations/0024_add_deepseek_models.py
barora05: done

apps/service_providers/llm_service/token_counters.py
barora05: Reused OpenAI token counter.

apps/service_providers/llm_service/token_counters.py
barora05: done

apps/service_providers/forms.py
barora05: removed

snopoke (approved):

snopoke:
Can be merged once the lint errors are fixed. This can be done by running `ruff` locally (either via `pre-commit run` or with `ruff check . --fix` and `ruff format .`

snopoke (approved):

PR #1207, opened 2025-02-18
allow unauthenticated users to download attachments

snopoke:
Resolves https://github.com/dimagi/open-chat-studio/issues/1206

## Description
Chat attachments should be downloadable by unauthenticated users.

### Security considerations
Downloading an attachment requires knowing the team slug, session ID and file ID which makes guessing non-trivial.

## User Impact
Public users can access chat attachments.

### Docs and Changelog
https://github.com/dimagi/open-chat-studio-docs/pull/47

### DIFF
--- a/apps/experiments/views/experiment.py
+++ b/apps/experiments/views/experiment.py
@@ -1356,8 +1356,6 @@ def experiment_session_pagination_view(request, team_slug: str, experiment_id: u
     return redirect("experiments:experiment_session_view", team_slug, experiment_id, next_session.external_id)
 
 
-@login_and_team_required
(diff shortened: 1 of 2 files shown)


SmittieC (commented):
Did we show the attachments to participants but returned a 403 when they try to download it?

snopoke:
> Did we show the attachments to participants but returned a 403 when they tries to download it?

If the user uploaded the attachment then we didn't make a link but for attachments generated by the bot we return a 403 if they click the link or when the image tries to load

SmittieC (approved):

PR #1211, opened 2025-02-19
render description as markdown on consent and pre-survey pages

snopoke:
## Description
The experiment description is shown on the consent and pre-survey page. Without markdown rendering all formatting in the text is not shown (newlines etc). One option would be to use a 'pre' block to display the description but that also shows it with monospace font etc.

Rendering it as markdown give much more control to chatbot builders.

#### Unrelated change
I also noticed that the invitation page does not have breadcrumbs on it so I added those too.

## User Impact
Descriptions can now include markdown and will be rendered correctly.

### Markdown rendered description
Screenshot from 2025-02-19 12-33-38

#### Breadcrumbs on the invitations page
Screenshot from 2025-02-19 12-30-01

## Docs and Changelog
https://github.com/dimagi/open-chat-studio-docs/pull/48

### DIFF
--- a/templates/experiments/experiment_invitations.html
+++ b/templates/experiments/experiment_invitations.html
@@ -1,5 +1,14 @@
 {% extends 'web/app/app_base.html' %}
 {% load form_tags %}
+{% block breadcrumbs %}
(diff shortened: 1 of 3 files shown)


SmittieC (approved):
Nice

Issue #1170, opened 2025-02-10
Re-evaluate (and block) events that could cause a loop


SmittieC:
We had an issue where the a triggering event was "when bot message received" and the action was to send another bot message. This caused a loop which killed our celery workers. We need to re-evaluate if we need the "when bot message received" and "when human message received" events. If we want to keep "when bot message received" around, we need to prevent actions that will cause the bot to send a message.

PR #1212, opened 2025-02-19
Prevent event loops with new received bot messages

stephherbers:
## Description

[ticket](https://github.com/dimagi/open-chat-studio/issues/1170)

A loop was caused by "when bot message received" event triggered the action was to send another bot message which overloaded our celery workers

Note:I chose these to limit because testing locally I was able to break my local set up with these options.

## User Impact

Limits their ability to break OCS

### Demo

Screenshot 2025-02-19 at 3 44 43 PM

### DIFF
--- a/apps/events/forms.py
+++ b/apps/events/forms.py
@@ -107,6 +107,20 @@ class EventActionForm(forms.ModelForm):
         fields = ["action_type"]
         labels = {"action_type": "Then..."}
 
+    def clean(self):
+        cleaned_data = super().clean()
+        action_type = cleaned_data.get("action_type")
+
+        if self.data and "type" in self.data:
+            trigger_type = self.data.get("type")
+            if trigger_type == "new_bot_message" and action_type in [
+                "send_message_to_bot",
+                "pipeline_start",
+                "summarize",
+            ]:
+                raise forms.ValidationError("This action is not allowed when 'A new bot message is received'")
+        return cleaned_data
+
     def save(self, commit=True, *args, **kwargs):
         experiment_id = kwargs.pop("experiment_id")
         instance = super().save(commit=False, *args, **kwargs)
--- a/templates/events/manage_event.html
+++ b/templates/events/manage_event.html
@@ -14,12 +14,50 @@
         <div>
             {% block pre_form %}
             {% endblock pre_form %}
-            <form method="post" class="my-2" {% include "generic/attrs.html" with attrs=form_attrs %} x-data="{ type: '{{ secondary_key }}' }">
+            <form method="post" class="my-2" {% include "generic/attrs.html" with attrs=form_attrs %}
+                  x-data="eventForm('{{ secondary_key }}', '{{ trigger_form.initial.type }}')"
+                  x-init="watchTriggerType()">
                 {% csrf_token %}
                 {% block form %}
                     <h1 class="pg-title">Event Details</h1>
                     {% render_form_fields trigger_form %}
                     {% render_form_fields action_form.primary %}
+                    <script>
+                        function eventForm(secondaryKey, initialTriggerType) {
+                            return {
+                                type: secondaryKey,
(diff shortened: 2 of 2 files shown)


apps/events/forms.py
```diff
@@ -107,6 +107,21 @@ class Meta:
...
+        action_type = cleaned_data.get("action_type")
+
+        if self.data and "type" in self.data:
+            trigger_type = self.data.get("type")
+            if trigger_type == "new_bot_message" and action_type in [
+                "send_message_to_bot",
+                "pipeline_start",
+                "summarize",
+                "schedule_trigger",
+            ]:
```
snopoke, line 121: Can you say more about why these ones were chosen? I don't think `summarize` can cause a loop, or `pipeline_start` - but I could be wrong. Not sure about `schedule_trigger` either.

SmittieC: Pipeline start will cause a loop if it sends a bot message and the trigger is `new_bot_message`. Same with `schedule_trigger`. `summarize` on the other hand should be fine

templates/events/manage_event.html
```diff
@@ -14,12 +14,50 @@ <h1 class="pg-title">{{ title }}</h1>
...
-            <form method="post" class="my-2" {% include "generic/attrs.html" with attrs=form_attrs %} x-data="{ type: '{{ secondary_key }}' }">
+            <form method="post" class="my-2" {% include "generic/attrs.html" with attrs=form_attrs %}
+                  x-data="eventForm('{{ secondary_key }}', '{{ trigger_form.initial.type }}')"
+                  x-init="watchTriggerType()">
                 {% csrf_token %}
                 {% block form %}
                     <h1 class="pg-title">Event Details</h1>
                     {% render_form_fields trigger_form %}
                     {% render_form_fields action_form.primary %}
+                    <script>
```
SmittieC, line 25: Nit: Can we put this script at the bottom? It will make the code look cleaner

apps/events/forms.py
snopoke: `pipeline_start` can't send messages. The output is just logged: https://github.com/dimagi/open-chat-studio/blob/46a93dfa1fb1b9413caf4497bb8da16318e950c5/apps/events/models.py#L123-L124

Fine to keep `schedule_trigger`.

**Side Note**: `log` and `summarize` can be removed from code since they don't actually do anything useful.

templates/events/manage_event.html
```diff
@@ -14,12 +14,50 @@ <h1 class="pg-title">{{ title }}</h1>
...
+                        function eventForm(secondaryKey, initialTriggerType) {
```
snopoke, line 26: This doesn't need to be a function, you can just hard code the values in:

```
{
  type: '{{ secondary_key }}',
  triggerType: '{{ trigger_form.initial.type }}',
  ...
}
```
```

File: templates/events/manage_event.html (hunk 2 above)
Comment (snopoke) [reply templates/events/manage_event.html:25]: You can use the [reusable data](https://alpinejs.dev/directives/data#re-usable-data) pattern as well

**Review by snopoke — APPROVED** [review]

File: templates/events/manage_event.html [hunk 4]
```diff
@@ -14,11 +14,15 @@ <h1 class="pg-title">{{ title }}</h1>
...
             {% endblock pre_form %}
-            <form method="post" class="my-2" {% include "generic/attrs.html" with attrs=form_attrs %} x-data="{ type: '{{ secondary_key }}' }">
+            <form method="post" class="my-2" {% include "generic/attrs.html" with attrs=form_attrs %}
+                  x-data="formData"
+                  x-init="watchTriggerType()">
                 {% csrf_token %}
                 {% block form %}
                     <h1 class="pg-title">Event Details</h1>
-                    {% render_form_fields trigger_form %}
+                    {% if trigger_form.type %}
```
SmittieC, line 23: @stephherbers what does this change do?

PR #1230, opened 2025-02-21
improve pipeline save and reload

snopoke:
## Description
This change makes pipelines only save when data that we care about is changed and also doesn't reset the flow after successful save.

Prior to this there were various UI artifacts that occurred due to re-rendering after the autosave. Also, almost every UI interaction would trigger a save e.g. changing which node was selected.

## User Impact
Improved user experience when editing pipelines.

### Docs and Changelog
NA

### DIFF
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -192,6 +192,7 @@ class Pipeline(BaseTeamModel, VersionsMixin):
                         id=node.flow_id,
                         type=node.type,
                         params=node.params,
+                        label=node.label,
                     ),
                 )
             )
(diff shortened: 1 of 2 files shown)


bderenzi:
This addresses the constant glitching where you'd drag a node and it'd get reset and/or start typing code and it'd go away etc? (has been on my mind to make an issue for that, but will skip if it's already addressed here)

snopoke:
> This addresses the constant glitching where you'd drag a node and it'd get reset and/or start typing code and it'd go away etc? (has been on my mind to make an issue for that, but will skip if it's already addressed here)

Yes, it addresses those. If you still see anything after this is deployed please do create an issue.

SmittieC (approved):

PR #1233, opened 2025-02-21
Google Gemini Support Integration

vaibhavK1199:
Integration for Google API Support


SmittieC (commented):
LGTM. You need to add the `langchain_google_genai` library to the requirements file

snopoke:
Here are the basic instructions for adding requirements: https://github.com/dimagi/open-chat-studio/?tab=readme-ov-file#adding-a-new-requirement

apps/service_providers/llm_service/default_models.py
```diff
@@ -74,6 +74,7 @@ def k(n: int) -> int:
         Model("llama-3.1-70b-instruct", 131072),
     ],
     "deepseek": [Model("deepseek-chat", 128000), Model("deepseek-reaoner", 128000)],
+    "google_gemini": [Model("gemini-1.5-pro", 128000), Model("gemini-1.5-flash", 128000)],
```
snopoke, line 77: Should probably include all of the models listed here: https://ai.google.dev/gemini-api/docs/models/gemini

(with the exception of the text embedding model)

apps/service_providers/llm_service/token_counters.py
```diff
@@ -67,3 +68,46 @@ def get_tokens_from_text(self, text) -> int:
...
+@dataclasses.dataclass
+class GeminiTokenCounter(TokenCounter):
+    model: str
+    google_api_key: str
+
+    def __post_init__(self):
+        if not self.google_api_key:
+            raise ValueError("KEY not found!")
+
+        genai.configure(api_key=self.google_api_key)
```
snopoke, line 82: This does global config setup which could result in using different keys from different projects (depending on race conditions).

The new API handles this differenly:

```python
client = genai.Client(api_key=...)
response = client.models.count_tokens(
    model='gemini-2.0-flash',
    contents='The quick brown fox jumps over the lazy dog.',
)
```

apps/service_providers/llm_service/default_models.py
vaibhavK1199: Have added other models as well

apps/service_providers/llm_service/token_counters.py
vaibhavK1199: replaced the configure logic with the client creation logic.

apps/service_providers/migrations/0029_alter_llmprovider_type_alter_llmprovidermodel_type.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 5.1.5 on 2025-02-24 14:21
```
SmittieC, line 1: Looks like there are two migrations that adds the same service provider types. I'd suggest you remove the last 3 migrations (numbers 27, 28 and 29) and regenerate it to get one migration file.

On a related point, I would suggest using a single migration file for the schema and data changes here. It's just a bit cleaner to have only one file and it makes management a tad bit easier. You can append the `operations` of your newly generated schema migration into the data migration's (number 26) `operations` list, but I would put it before the `migrations.RunPython(add_google_gemini_models)` operation.

vaibhavK1199: Yes I do agree, this is done.

apps/service_providers/migrations/0026_add_google_gemini_models.py
```diff
@@ -0,0 +1,21 @@
+import logging
```
SmittieC, line 1: Nit: You can merge this migration and the one after it into one by manually merging the `operations`:

```python
    operations = [
        migrations.AlterField(
            model_name='llmprovider',
            name='type',
            field=models.CharField(choices=[('openai', 'OpenAI'), ('azure', 'Azure OpenAI'), ('anthropic', 'Anthropic'), ('groq', 'Groq'), ('perplexity', 'Perplexity'), ('deepseek', 'DeepSeek'), ('google', 'Google Gemini')], max_length=255),
        ),
        migrations.AlterField(
            model_name='llmprovidermodel',
            name='type',
            field=models.CharField(choices=[('openai', 'OpenAI'), ('azure', 'Azure OpenAI'), ('anthropic', 'Anthropic'), ('groq', 'Groq'), ('perplexity', 'Perplexity'), ('deepseek', 'DeepSeek'), ('google', 'Google Gemini')], max_length=255),
        ),
        migrations.RunPython(add_google_gemini_models),
    ]
```

requirements/dev-requirements.txt
snopoke, line 1: I see some updates due to being on Windows and also using `pip compile` instead of `uv`. I'd prefer not to get these changes into `main` so I'm going to suggest we change the base branch for this PR to a new branch so that we can rebuild the requirements files before merging into `main`.

snopoke: I've changed the base branch to `gemini_temp`

apps/service_providers/llm_service/token_counters.py
```diff
@@ -67,3 +69,46 @@ def get_tokens_from_text(self, text) -> int:
...
+        if input_tokens is None or output_tokens is None:
+            return None
+
+        return input_tokens, output_tokens
+
+    def get_tokens_from_text(self, text: str) -> int:
+        if not text:
+            return 0
+
+        response = self.client.count_tokens(model=self.model, contents=text)
```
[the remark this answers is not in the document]
vaibhavK1199: Yes I have tested this and it worked.

snopoke (approved):

PR #1235, opened 2025-02-21
Summary events fixes

stephherbers:
## Description

When trying to display the summary events in the events table an "oh shucks" page appeared. 

The error occurs because the prompt string contains {summary} and {new_lines} placeholders, and when the format_html function tries to format the string, it's looking for these keys but they don't exist in the context.

Also fixes Validation error that prevents summary from happening: 
Screenshot 2025-02-21 at 3 30 37 PM

BONUS: shortens values in the param column to 100 characters: 
Screenshot 2025-02-25 at 11 11 52 AM

## User Impact

Experiments are usable again after configuring summary after an event

### Demo

n/a

### Docs and Changelog

n/a


apps/events/tables.py
```diff
@@ -46,8 +46,11 @@ def render(self, value, record):
 
 class ParamsColumn(tables.Column):
     def render(self, value, record):
-        items = format_html_join("", "<li><strong>{}</strong>: {}</li>", value.items())
-        return format_html(f"<ul>{items}</ul>")
+        try:
+            items = format_html_join("", "<li><strong>{}</strong>: {}</li>", value.items())
+            return format_html(f"<ul>{items}</ul>")
```
snopoke, line 51: this isn't actually doing a format since it's using a f-string. Changing this line to `format_html("<ul>{}</ul>", items)` should also fix the issue.

stephherbers: sweet thanks-- fyi this does make the row take up the whole screen for the summary each time unless we reformat, just flagging in case thats a deal breaker
Screenshot 2025-02-21 at 3 32 51 PM

snopoke: I think we should trim the value to a smaller size - maybe 100 chars

stephherbers: https://github.com/dimagi/open-chat-studio/pull/1235#:~:text=%E2%80%A6-,8e849fb,-Merge%20branch%20%27smh

apps/events/utils.py
```diff
@@ -0,0 +1,18 @@
+def truncate_dict_items(dictionary, max_limit=100):
+    """
+    Truncate dictionary items to fit within a char limit. Returns list of (key, value)
+    tuples that fit within the limit without breaking up the key value
+    """
+    formatted_items = []
+    chars_used = 0
+    for key, val in dictionary.items():
+        key_str, val_str = str(key), str(val)
+        if chars_used + len(key_str) > max_limit:
```
stephherbers, line 10: note: this won't cut off a key halfway through like it will a value. ex. if the key is `time_period` it will not do `time...`

snopoke: won't this prevent showing some keys? I think we just want to truncate the values, not the entire dict.

stephherbers: ah gotcha this will prevent showing some keys-- I'll fix

snopoke (approved):

PR #1236, opened 2025-02-24
Ba/add version param

### LINKED ISSUE #962: Add chatbot version support to APIs
**Note**: This is not about versioning the API itself but is about exposing the chatbot versions via the APIs.

## Task
Modify the [openai completions API](https://github.com/dimagi/open-chat-studio/blob/main/apps/api/openai.py#L90) and the [new API message API](https://github.com/dimagi/open-chat-studio/blob/main/apps/channels/views.py#L93) to accept an optional path element to set the version of the bot to use for that request.

### URL scheme
Default version: `/api/openai/{experiment_id}/chat/completions`
Version N: `/api/openai/{experiment_id}/vN/chat/completions`

Default version: `/channels/api/{experiment_id}/incoming_message`
Version N: `/channels/api/{experiment_id}/vN/incoming_message`

## Additional work

Add a button to the versions table actions column to copy the API URL for that bot version:

Image

barora05:
## Description
Add API button in every version of experiment and version param in url. Add version param in open_chat_api. Closes #962 

### Demo
[API button demo](https://drive.google.com/file/d/1MxbPwqF3McL-3NtVt6-D-Fuc4IcSTI5K/view?usp=drive_link)

### DIFF
--- a/apps/api/openai.py
+++ b/apps/api/openai.py
@@ -111,7 +111,7 @@ def chat_completions_schema(versioned: bool):
 @chat_completions_schema(versioned=False)
 @api_view(["POST"])
 def chat_completions(request, experiment_id: uuid.UUID):
-    return chat_completions_version(request._request, experiment_id)
+    return chat_completions_version(request, experiment_id)
 
 
 @chat_completions_schema(versioned=True)
--- a/apps/api/openai.py
+++ b/apps/api/openai.py
@@ -110,7 +110,7 @@ def chat_completions(request, experiment_id: uuid.UUID, version=None):
         return _make_error_response(400, str(e))
 
     session = serializer.save()
-    experiment_version = session.experiment_version if version is None else session.fetch_experiment(experiment_id, version)
+    experiment_version = session.experiment.get_version(version) if version is not None else session.experiment_version
     response_message = handle_api_message(
         request.user,
         experiment_version,
--- a/apps/api/openai.py
+++ b/apps/api/openai.py
@@ -114,13 +114,13 @@ def chat_completions_schema(versioned: bool):
 @chat_completions_schema(versioned=False)
 @api_view(["POST"])
 def chat_completions(request, experiment_id: uuid.UUID):
-    return chat_completions_version(request, experiment_id)
+    return _chat_completions(request, experiment_id)
 
 
 @chat_completions_schema(versioned=True)
 @api_view(["POST"])
 def chat_completions_version(request, experiment_id: uuid.UUID, version=None):
-    _chat_completions(request, experiment_id, version)
+    return _chat_completions(request, experiment_id, version)
 
 
 def _chat_completions(request, experiment_id: uuid.UUID, version=None):
--- a/apps/api/openai.py
+++ b/apps/api/openai.py
@@ -111,7 +111,7 @@ def chat_completions_schema(versioned: bool):
 @chat_completions_schema(versioned=False)
 @api_view(["POST"])
 def chat_completions(request, experiment_id: uuid.UUID):
-    return chat_completions_version(request, experiment_id)
+    return chat_completions_version(request._request, experiment_id)
 
 
 @chat_completions_schema(versioned=True)
--- a/apps/api/openai.py
+++ b/apps/api/openai.py
@@ -103,10 +103,12 @@ def chat_completions_schema(versioned: bool):
         parameters=parameters,
     )
 
+
 @chat_completions_schema(versioned=False)
 @api_view(["POST"])
 def chat_completions(request, experiment_id: uuid.UUID):
     return chat_completions_version(request._request, experiment_id)
+
 
 @chat_completions_schema(versioned=True)
 @api_view(["POST"])
--- a/apps/api/openai.py
+++ b/apps/api/openai.py
@@ -12,12 +12,65 @@ from rest_framework.response import Response
 from apps.api.serializers import ExperimentSessionCreateSerializer, MessageSerializer
 from apps.channels.tasks import handle_api_message
 
+create_chat_completion_request = inline_serializer(
+    "CreateChatCompletionRequest", {"messages": MessageSerializer(many=True)}
+)
+
+create_chat_completion_response = inline_serializer(
+    "CreateChatCompletionResponse",
+    {
+        "id": serializers.CharField(),
+        "choices": inline_serializer(
+            "ChatCompletionResponseChoices",
+            {
+                "finish_reason": serializers.CharField(),
+                "index": serializers.IntegerField(),
+                "message": inline_serializer(
+                    "ChatCompletionResponseMessage",
+                    {
+                        "role": serializers.ChoiceField(choices=["assistant"]),
+                        "content": serializers.CharField(),
+                    },
+                ),
+            },
+            many=True,
+        ),
+        "created": serializers.IntegerField(),
+        "model": serializers.CharField(),
+        "object": serializers.ChoiceField(choices=["chat.completion"]),
+    },
+)
+
 
 def chat_completions_schema(versioned: bool):
     operation_id = "openai_chat_completions"
     summary = "Chat Completions API for Experiments"
+    base_url = "https://chatbots.dimagi.com/api/openai/{experiment_id}"
+
+    parameters = [
+        OpenApiParameter(
+            name="experiment_id",
+            type=OpenApiTypes.STR,
+            location=OpenApiParameter.PATH,
+            description="Experiment ID",
+        ),
+    ]
+
+    if versioned:
+        operation_id = f"{operation_id}_versioned"
+        summary = "Versioned Chat Completions API for Experiments"
+        base_url = "https://chatbots.dimagi.com/api/openai/{experiment_id}/v{version}"
+        parameters.append(
+            OpenApiParameter(
+                name="version",
+                type=OpenApiTypes.STR,
+                location=OpenApiParameter.QUERY,
+                description="Version of experiment",
+            )
+        )
+
     description = textwrap.dedent(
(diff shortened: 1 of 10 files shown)


templates/experiments/components/experiment_version_actions.html
```diff
@@ -22,6 +22,10 @@
             onclick="document.querySelector('#versionDetailsModal').showModal();">
       View Details
     </button>
+    <input type="hidden" id="api-url-{{ record.version_number }}" value="{{record.api_url }}">
```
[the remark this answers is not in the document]
snopoke: Not a bad idea but that will end up being a long list unless we limit it to some number. But then we need a way to get the URL for ones that don't make it to the dropdown. I'm fine with leaving this as is for now.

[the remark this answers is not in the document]
SmittieC: Cool, fine to leave as-is. It might be useful to have a dropdown with a mini search bar for cases like these

apps/experiments/models.py
```diff
@@ -848,9 +848,15 @@ def trace_service(self):
         if self.trace_provider:
             return self.trace_provider.get_service()
 
-    def get_api_url(self):
+    def get_api_url(self, version_number=None):
+        if version_number:
+            return absolute_url(reverse("api:openai-chat-completions-versioned", args=[self.public_id, version_number]))
```
snopoke, line 853: I think this is related to what @SmittieC was saying above about using the public ID of the working version.

I'm not sure the best way to get that here without doing another DB hit. Maybe a separate method that allows passing in the working version's public ID?

apps/api/tests/test_openai_api.py
```diff
@@ -91,6 +91,52 @@ def test_chat_completion(mock_experiment_response, experiment, api_key, live_ser
...
+    Test OpenAI chat completion API with versioned endpoint (v2).
+    """
+    mock_experiment = MagicMock()
+    mock_experiment_get.return_value = mock_experiment  # Returning mock experiment instance
+
+    mock_experiment_response.return_value = "So, this ain't the end, I saw you again today"
+    mock_fetch_experiment.return_value = experiment
+    version = 2
+    base_url = f"{live_server.url}/api/openai/{experiment.public_id}/v{version}"
+    print(f"Testing versioned API: {base_url}")
```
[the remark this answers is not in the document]
barora05: done

apps/channels/views.py
```diff
@@ -109,17 +109,19 @@ def new_api_message(request, experiment_id: uuid):
             )
         except ExperimentSession.DoesNotExist:
             raise Http404
-
         participant_id = session.participant.identifier
         experiment_channel = session.experiment_channel
         experiment = session.experiment
+        experiment_version = (
+            session.fetch_experiment(public_id=experiment_id, version_number=version) if version else experiment
+        )
```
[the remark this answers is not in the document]
barora05: done

apps/api/openai.py
```diff
@@ -111,9 +110,10 @@ def chat_completions(request, experiment_id: uuid.UUID):
         return _make_error_response(400, str(e))
 
     session = serializer.save()
+    experiment_version = session.experiment_version if version is None else session.fetch_experiment(experiment_id, version)
```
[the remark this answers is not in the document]
barora05: done

apps/experiments/models.py
```diff
@@ -1681,6 +1687,9 @@ def try_send_message(self, message: str, fail_silently=True):
             if not fail_silently:
                 raise e
 
+    def fetch_experiment(self, experiment_id, version_number) -> Experiment:
+        return Experiment.objects.get(public_id=experiment_id, version_number=version_number)
```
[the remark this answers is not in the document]
barora05: removed

apps/experiments/models.py
barora05: done. used working version's public id

snopoke (commented):
I was just looking at this locally and realized that the schema docs are identical for the version & non-version API. To fix this I think we should split the views and create a helper function to get the schema:

```python
@chat_completions_schema(versioned=False)
@api_view(["POST"])
def chat_completions(request, experiment_id: uuid.UUID):
    return chat_completions_version(request, experiment_id)


@chat_completions_schema(versioned=True)
@api_view(["POST"])
def chat_completions_version(request, experiment_id: uuid.UUID, version=None):
    # current view implementation goes here
```

The difference between the schema will be the `operation_id`, the `summary` and the additional path parameter.

Ideally the description would get updated to include the versioned URL in the example code too.

apps/api/openai.py
```diff
@@ -88,6 +88,11 @@
 )
 @api_view(["POST"])
 def chat_completions(request, experiment_id: uuid.UUID):
+    return chat_completions_version(request._request, experiment_id)
+
+
+@api_view(["POST"])
+def chat_completions_version(request, experiment_id: uuid.UUID, version=None):
```
snopoke, line 95: @barora05 this still needs to have the `@extends_schema` decorator applied to it. In my previous comment I suggested making a utility function which would return the decorator to avoid duplication. Here's the basic pattern:

```python
def chat_completions_schema(versioned: bool):
    operation_id = "openai_chat_completions"
    summary = ...
    parameters = [...]
    if versioned:
        operation_id = f"{operation_id}_versioned"
        summary = ...
        parameters.append(...)

    return extend_schema(
        operation_id=operation_id,
        summary=summary,
        ...,
        parameters=parameters
    )

@chat_completions_schema(versioned=False)
@api_view(["POST"])
def chat_completions(request, experiment_id: uuid.UUID):
    return chat_completions_version(request, experiment_id)


@chat_completions_schema(versioned=True)
@api_view(["POST"])
def chat_completions_version(request, experiment_id: uuid.UUID, version=None):
    # current view implementation goes here
```

Here is a good introduction to decorators in Python: https://realpython.com/primer-on-python-decorators/

apps/channels/views.py
```diff
@@ -90,7 +90,7 @@ def new_turn_message(request, experiment_id: uuid):
     ],
 )
 @api_view(["POST"])
-def new_api_message(request, experiment_id: uuid):
+def new_api_message(request, experiment_id: uuid, version=None):
```
snopoke, line 93: This API needs the same treatment as the other one in terms of providing separate schema for the versioned and un-versioned API.

apps/api/openai.py
```diff
@@ -45,49 +45,74 @@
         reply = completion.choices[0].message
```
snopoke, line 45: the `base_url=f"https://chatbots.dimagi.com/api/openai/{experiment_id}",` section of the docs should be different in the versioned and un-versioned schema.

apps/api/openai.py
```diff
@@ -65,27 +67,29 @@ def chat_completions_schema(versioned: bool):
                 description="Version of experiment",
             )
         )
+        request_serializer_name = "CreateChatCompletionRequestVersioned"
+        response_serializer_name = "CreateChatCompletionResponseVersioned"
```
snopoke, line 71: I don't think these need to be different since they are identical in all other respects. It's fine to reuse the same body / response definitions.

apps/channels/views.py
```diff
@@ -61,36 +61,62 @@ def new_turn_message(request, experiment_id: uuid):
...
+        parameters.append(
+            OpenApiParameter(
+                name="version",
+                type=OpenApiTypes.STR,
+                location=OpenApiParameter.QUERY,
+                description="Version of experiment",
+            )
+        )
+        request_serializer_name = "NewAPIMessageVersioned"
+        response_serializer_name = "NewAPIMessageResponseVersioned"
```
snopoke, line 89: same comment as above - I think it's fine to keep these names the same for both APIs since both accept the same data

apps/api/openai.py
```diff
@@ -111,7 +111,7 @@ def chat_completions_schema(versioned: bool):
 @chat_completions_schema(versioned=False)
 @api_view(["POST"])
 def chat_completions(request, experiment_id: uuid.UUID):
-    return chat_completions_version(request._request, experiment_id)
```
snopoke: Can you explain this, I've never seen this being done before.

apps/api/openai.py
barora05: I added these as I was getting a warning related to this in the logs. Will remove these

apps/api/openai.py
barora05: Unit tests failed when I passed the request object directly. Read at few places that DRF wraps HttpRequest  into a request object so to pass the actual Http Request I used request._request.

apps/api/openai.py
snopoke: The warnings can be resolved by extracting the serializer to a variable that lives outside the function. e.g:

```python
create_chat_completion_request = inline_serializer(
    "CreateChatCompletionRequest", {"messages": MessageSerializer(many=True)}
)
```

apps/api/openai.py
snopoke: This is because the one view is being called by the other one and they are both decorated with `@api_view`. The solution to to create a 3rd function that both views call which isn't decorated:

```python
@chat_completions_schema(versioned=False)
@api_view(["POST"])
def chat_completions(request, experiment_id: uuid.UUID):
    return _chat_completions(request, experiment_id)


@chat_completions_schema(versioned=True)
@api_view(["POST"])
def chat_completions_version(request, experiment_id: uuid.UUID, version=None):
    return _chat_completions(request, experiment_id, version)


def _chat_completions(request, experiment_id: uuid.UUID, version=None):
    ...
```

snopoke (approved):
@barora05 I changed the UI a bit. I'll merge once the build passes.

PR #1238, opened 2025-02-24
Limit the number of messages that can be sent to the LLM

SmittieC:
## Description

- Limit the number of messages that can be sent to the LLM. This fixes [this error](https://dimagi.sentry.io/issues/5842110431/?project=4505001320316928&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=7d&stream_index=1#exception). This happens when the total token count of all messages are smaller than the model's limit, so then the API's limit is reached. I couldn't find this limit in the docs.

## User Impact

Users should be able to have super long conversations without this happening again

### Demo

N/A

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/53

### DIFF
--- a/apps/chat/conversation.py
+++ b/apps/chat/conversation.py
@@ -259,7 +259,7 @@ def _get_new_summary(llm, pruned_memory, summary, max_token_limit):
     recursively call this function with the remaining memory."""
     tokens, context = _get_summary_tokens_with_context(llm, summary, pruned_memory)
     next_batch = []
-    while pruned_memory and tokens > max_token_limit and len(pruned_memory) > MAX_UNCOMPRESSED_MESSAGES:
+    while pruned_memory and tokens > max_token_limit or len(pruned_memory) > MAX_UNCOMPRESSED_MESSAGES:
         next_batch.insert(0, pruned_memory.pop())
         tokens, context = _get_summary_tokens_with_context(llm, summary, pruned_memory)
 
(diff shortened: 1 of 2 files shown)


stephherbers (approved):

apps/chat/conversation.py
```diff
@@ -208,9 +209,18 @@ def compress_chat_history_from_messages(
...
+        or _tokens_exceeds_limit(
+            history, token_count=(history_tokens + summary_tokens + input_message_tokens), limit=max_token_limit
+        )
+    ):
         first_pass_done = True
-        while history and history_tokens + summary_tokens + input_message_tokens > max_token_limit:
+        while _tokens_exceeds_limit(
+            history, token_count=(history_tokens + summary_tokens + input_message_tokens), limit=max_token_limit
+        ) or _messages_exceeds_limit(history, input_messages):
             pruned_memory.append(history.pop(0))
```
snopoke, line 224: Instead of just popping off one message we could trim it to the correct size in one go which would reduce the number of loops.

snopoke (approved):
One comment but not blocking

apps/chat/conversation.py
SmittieC: Good idea, I'll open a new PR for this

SmittieC: https://github.com/dimagi/open-chat-studio/pull/1243

PR #1239, opened 2025-02-24
remove num_outputs

barora05:

### DIFF
--- a/assets/javascript/apps/pipeline/nodes/widgets.tsx
+++ b/assets/javascript/apps/pipeline/nodes/widgets.tsx
@@ -639,24 +639,10 @@ export function KeywordsWidget(props: WidgetParams) {
   const setEdges = usePipelineStore((state) => state.setEdges);
   const updateNodeInternals = useUpdateNodeInternals()
 
-  function getNewNodeData(old: Node, keywords: any[], numOutputs: number) {
-    return {
-      ...old,
-      data: {
-        ...old.data,
-        params: {
-          ...old.data.params,
-          ["keywords"]: keywords,
-          ["num_outputs"]: numOutputs,
-        },
-      },
-    };
-  }
-
   const addKeyword = () => {
     setNode(props.nodeId, (old) => {
       const updatedList = [...(old.data.params["keywords"] || []), ""];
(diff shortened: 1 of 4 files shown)


assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -639,24 +639,10 @@ export function KeywordsWidget(props: WidgetParams) {
   const setEdges = usePipelineStore((state) => state.setEdges);
   const updateNodeInternals = useUpdateNodeInternals()
 
-  function getNewNodeData(old: Node, keywords: any[], numOutputs: number) {
```
SmittieC, line 642: This is still called by the code. Only the `num_outputs` should be removed

barora05: There are 2 getNewNodeData functions in the code.
```
 function getNewNodeData(old: Node, updatedList: Array<string>) {
    return {
      ...old,
      data: {
        ...old.data,
        params: {
          ...old.data.params,
          [props.name]: updatedList,
        },
      },
    };
  }
```
I have kept this one and removed the one which had numOutputs

barora05: that is in a diff function. my bad. corrected it

snopoke:
@barora05 can you record a quick demo showing that the router nodes still work as expected.

snopoke (commented):
LGTM. Is any data migration necessary or are these changes backwards compatible with the old data that has the `num_outputs` field?

barora05:
> LGTM. Is any data migration necessary or are these changes backwards compatible with the old data that has the `num_outputs` field?

I don't think there would be a migration required.

barora05:
> @barora05 can you record a quick demo showing that the router nodes still work as expected.

@snopoke Attatched demo. Fixed the issue

snopoke:
> > @barora05 can you record a quick demo showing that the router nodes still work as expected.

you attached the wrong demo

barora05:
@snopoke Updated the demo link.

snopoke (approved):

PR #1240, opened 2025-02-25
Some small changes

SmittieC:
## Description

- Show uploaded file count. I decided to add this after someone asked me to run a query to count how many files they had uploaded to their assistant
- The bot will try and tell users (external channels only) when something goes wrong while generating a message.
- Fix: When the LLM responded with an empty string, the ai message was not created or saved on the history manager, causing [this error](https://dimagi.sentry.io/issues/6328678790/?alert_rule_id=14063560&alert_type=issue&notification_uuid=2f444287-2986-47f1-bd63-452511c14073&project=4505001320316928&referrer=issue_alert-slack). We now save every response.

## User Impact

User will see the number of uploaded files

#### File count
image

### Docs and Changelog

Will update and link it once this PR is approved

### DIFF
--- a/apps/chat/channels.py
+++ b/apps/chat/channels.py
@@ -47,6 +47,7 @@ UNSUPPORTED_MESSAGE_BOT_PROMPT = """
 Tell the user (in the language being spoken) that they sent an unsupported message.
 You only support {supported_types} messages types. Respond only with the message for the user
 """
+DEFAULT_ERROR_RESPONSE_TEXT = "Sorry, something went wrong while processing your message. Please try again later"
 
 # The regex from https://stackoverflow.com/a/6041965 is used, but tweaked to remove capturing groups
 URL_REGEX = r"(?:http|ftp|https):\/\/(?:[\w_-]+(?:(?:\.[\w_-]+)+))(?:[\w.,@?^=%&:\/~+#-]*[\w@?^=%&\/~+#-])"
@@ -556,6 +557,7 @@ class ChannelBase(ABC):
         """Simply tells the user that something went wrong to keep them in the loop. This method wil not raise an error
         if something went wrong during this operation.
         """
+
         try:
             bot_message = self._generate_response_for_user(
                 """
@@ -563,9 +565,16 @@ class ChannelBase(ABC):
                 try again later
                 """
             )
+        except Exception as e:
+            logger.exception(
+                f"Something went wrong while trying to generate an appropriate error message for the user\n\n{e}"
+            )
+            bot_message = DEFAULT_ERROR_RESPONSE_TEXT
+
+        try:
             self.send_message_to_user(bot_message)
         except Exception as e:
-            logger.exception(f"Something went wrong while trying to inform the user of an error.\{e}")
+            logger.exception(f"Something went wrong while trying to inform the user of an error\n\n{e}")
 
     def _generate_response_for_user(self, prompt: str) -> str:
         """Generates a response based on the `prompt`."""
--- a/apps/chat/channels.py
+++ b/apps/chat/channels.py
@@ -40,6 +40,8 @@ from apps.service_providers.speech_service import SynthesizedAudio
 from apps.slack.utils import parse_session_external_id
 from apps.users.models import CustomUser
 
+logger = logging.getLogger("ocs.channels")
+
 USER_CONSENT_TEXT = "1"
 UNSUPPORTED_MESSAGE_BOT_PROMPT = """
 Tell the user (in the language being spoken) that they sent an unsupported message.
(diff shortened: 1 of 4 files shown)


apps/chat/channels.py
```diff
@@ -262,29 +264,32 @@ def _participant_is_allowed(self):
...
-            return self.send_message_to_user("Sorry, you are not allowed to chat to this bot")
-
-        if not self.is_message_type_supported():
-            return self._handle_unsupported_message()
 
-        if self.experiment_channel.platform != ChannelPlatform.WEB:
-            if self._is_reset_conversation_request():
-                # Webchats' statuses are updated through an "external" flow
-                return ""
+            if not self.is_message_type_supported():
```
SmittieC, line 268: The diff view is confusing here. I moved the code under the `except ParticipantNotAllowedException` into the `try` block and added another `except  Exception` after that.

apps/chat/channels.py
```diff
@@ -361,11 +366,7 @@ def _user_gave_consent(self) -> bool:
 
     def _extract_user_query(self) -> str:
         if self.message.content_type == MESSAGE_TYPES.VOICE:
-            try:
```
SmittieC, line 364: This will be caught by the exception I added in the `_new_user_message` method and ultimately do there what it did here

templates/files/partials/file_list.html
```diff
@@ -26,3 +26,21 @@ <h3 class="text-lg font-medium">{{ files_title|default:"FILES" }}</h3>
   {% endfor %}
 </ul>
 </div>
+
+<script>
```
SmittieC, line 30: Suggestion for simpler ways to watch the `<ul>` for changes to its children elements welcome.

apps/service_providers/llm_service/history_managers.py
```diff
@@ -89,6 +89,7 @@ def __init__(
         self.max_token_limit = max_token_limit
         self.chat_model = chat_model
         self.trace_service = trace_service
+        self.ai_message = ""
```
SmittieC, line 92: We set this attribute [here](https://github.com/dimagi/open-chat-studio/blob/02fdb02c8c669426f197165fc7236357be595e28/apps/service_providers/llm_service/history_managers.py#L62), but it was never initialized to a default value. Now it is

apps/chat/channels.py
```diff
@@ -551,14 +552,19 @@ def _unsupported_message_type_response(self):
...
+        try:
+            bot_message = self._generate_response_for_user(
+                """
+                Tell the user that something went wrong while processing their message and that they should
+                try again later
+                """
+            )
+            self.send_message_to_user(bot_message)
+        except Exception as e:
+            logger.exception(f"Something went wrong while trying to inform the user of an error.\{e}")
```
snopoke, line 568: In this case should we just send a generic error message to the user. If we move the 'send_message_to_user' call to it's own try catch block then we can handle the errors independently.

SmittieC: https://github.com/dimagi/open-chat-studio/pull/1240/commits/23bc1d69f921844432f077cd0f43116d1fe1533e

apps/chat/channels.py
```diff
@@ -556,16 +557,24 @@ def _inform_user_of_error(self):
...
         try:
             bot_message = self._generate_response_for_user(
                 """
                 Tell the user that something went wrong while processing their message and that they should
                 try again later
                 """
             )
+        except Exception as e:
+            logger.exception(
+                f"Something went wrong while trying to generate an appropriate error message for the user\n\n{e}"
```
snopoke, line 570: nit: don't add the error to the message. The log record will include the exception details automatically and adding the error to the messages breaks exception grouping in tools like sentry

apps/chat/channels.py
```diff
@@ -556,16 +557,24 @@ def _inform_user_of_error(self):
...
+            )
+            bot_message = DEFAULT_ERROR_RESPONSE_TEXT
+
+        try:
             self.send_message_to_user(bot_message)
         except Exception as e:
-            logger.exception(f"Something went wrong while trying to inform the user of an error.\{e}")
+            logger.exception(f"Something went wrong while trying to inform the user of an error\n\n{e}")
```
snopoke, line 577: Same as above

snopoke (dismissed):

apps/chat/channels.py
SmittieC: Oh, I did not know that. https://github.com/dimagi/open-chat-studio/pull/1240/commits/76a11d6fba4004abc3074f1906116c76cb134867

snopoke (approved):

PR #1246, opened 2025-02-25
support 'system' theme and fix theme flicker

snopoke:
## Description
This updates the theme selector to allow selecting 'light', 'dark' or 'system'. It also adds a cookie which is used during server side rendering to set the HTML attributes which prevents a flicker if the set theme is different from the initial theme.

## User Impact
User's can select 'system' theme to revert theme control to their browser. 

### Demo
[Screencast from 2025-02-26 11-10-46.webm](https://github.com/user-attachments/assets/e55a4f5a-3b65-4348-b9e1-60e9b31ce226)

### Docs and Changelog
NA


SmittieC (approved):
:moon:

barora05 (approved):

Issue #1195, opened 2025-02-17
Tool based reminders sometimes causes the bot to reply to itself


SmittieC:
Related [Slack thread](https://dimagi.slack.com/archives/C05PK63Q89H/p1739571399405219).
The [Jira ticket](https://dimagi.atlassian.net/browse/OCSBUGS-4)

The bug seems to occur with the dynamic reminders (i.e. the one-off tool based reminder). It results in a weird “meta conversation” loop where the LLM sends a reminder and then responds to itself, thanking itself for the reminder and continuing the exchange unnecessarily.

snopoke:
Thoughts:

* use external llm from experiment i.e. not the pipeline
* use the system prompt from the node that created the message

PR #1247, opened 2025-02-26
Propagate `save_input_to_history` config for pipelines

SmittieC:
## Description

For pipelines, any scheduled message or messages originating from the [generate_response_for_user](https://github.com/dimagi/open-chat-studio/blob/76a11d6fba4004abc3074f1906116c76cb134867/apps/chat/channels.py#L577-L580) method are saved as a human message. Now, it's not, since we check the `save_input_to_history` flag.

I suspect this issue might be involved in the [bot replying to itself issue](https://github.com/dimagi/open-chat-studio/issues/1195), but I could not replicate this, so I am not 100% sure.

## User Impact

Users will not see the prompt for the bot that is used to generate a response in the chat history any more.

### Demo

N/A

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/55

### DIFF
--- a/apps/chat/bots.py
+++ b/apps/chat/bots.py
@@ -300,6 +300,7 @@ class PipelineBot:
                 pipeline_version=self.experiment.pipeline.version_number,
             ),
             self.session,
+            save_input_to_history=save_input_to_history,
         )
         self.ai_message_id = output["ai_message_id"]
         return output["messages"][-1]
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -215,7 +215,9 @@ class Pipeline(BaseTeamModel, VersionsMixin):
             output = PipelineState(**output).json_safe()
         return output
 
-    def invoke(self, input: PipelineState, session: ExperimentSession, save_run_to_history: bool = True) -> dict:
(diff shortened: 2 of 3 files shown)


SmittieC:
As a general thought, I am starting to wonder if user messages should not be saved as soon as it can be i.e. either inside the channel after the session is ensured or in the bot.py file. Not entirely sure what I think of this yet

snopoke (approved):

snopoke:
> As a general thought, I am starting to wonder if user messages should not be saved as soon as it can be i.e. either inside the channel after the session is ensured or in the bot.py file. Not entirely sure what I think of this yet

I'm pro this and have wanted to do more centralized history management external to pipelines. I think there are a few things we'd need:

1. Exclude user message from history sent to bot - not sure if this is just because of how we have the langchain prompts set up but currently we pass history + user message so if we save the user message first then we'd need to make sure it doesn't get included twice
2. Merge adjacent messages of the same type: human, human, bot, human, bot, bot -> human, bot, human, bot
    - This helps the bot have a consistent view of the world and it is a requirement for Anthropic

PR #1250, opened 2025-02-26
Catch exec error

SmittieC:
## Description

Fixes https://dimagi.sentry.io/issues/6333464466/?alert_rule_id=14063560&alert_type=issue&notification_uuid=6b9208ed-1176-478e-b1d0-0bc6c381023e&project=4505001320316928&referrer=issue_alert-slack

## User Impact

The site wouldn't break when the `exec` fails

### Demo

demo

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/56

### DIFF
--- a/apps/pipelines/nodes/nodes.py
+++ b/apps/pipelines/nodes/nodes.py
@@ -778,7 +778,10 @@ class CodeNode(PipelineNode):
                 mode="exec",
             )
             custom_locals = {}
-            exec(byte_code, {}, custom_locals)
+            try:
+                exec(byte_code, {}, custom_locals)
+            except Exception as exc:
+                raise PydanticCustomError("invalid_code", "{error}", {"error": exc.msg})
 
             try:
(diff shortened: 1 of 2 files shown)


snopoke (approved):
I think this probably also fixes https://github.com/dimagi/open-chat-studio/issues/1180

@vaibhavK1199 FYI

apps/pipelines/nodes/nodes.py
```diff
@@ -778,7 +778,10 @@ def validate_code(cls, value, info: FieldValidationInfo):
                 mode="exec",
             )
             custom_locals = {}
-            exec(byte_code, {}, custom_locals)
+            try:
+                exec(byte_code, {}, custom_locals)
+            except Exception as exc:
+                raise PydanticCustomError("invalid_code", "{error}", {"error": exc.msg})
```
snopoke, line 784: I think the safe way to do this is `str(exc)`

PR #1259, opened 2025-02-27
Multimedia | 1 | File management UI

### LINKED ISSUE #1216: [Multimedia support] Add data models

We need to add the [data models](https://docs.google.com/document/d/1rXIbpHf-coprjIo-qsCS0A8PjsVEvoEP7Y2F0K7P4kI/edit?tab=t.nnf8cwfrza7y#heading=h.ia7vi96woqqu) from **Approach 2**

### LINKED ISSUE #1217: [Multimedia support] Document management screen

**User story**
As a user, I need to be able to create a collection and upload files to it.

Depends on the [data models](https://github.com/dimagi/open-chat-studio/issues/1216).

- We need the document management screen described [here](https://docs.google.com/document/d/1rXIbpHf-coprjIo-qsCS0A8PjsVEvoEP7Y2F0K7P4kI/edit?tab=t.nnf8cwfrza7y#heading=h.359oparyirks).
- This should be built behind a feature flag.
- File tab
- Collections tab
- File upload modal

SmittieC:
## Description

Resolves #1216 
Resolves #1217

UI setup for file management.

I tried reusing the code we have for tables so I don't have to do pagination etc, but it became a mini nightmare quite soon with having to add styles and alpine/htmx attributes in the python code etc.

## User Impact

Currently users can only upload files to an existing collection.

### Demo

demo

### Docs and Changelog

N/A yet

### Things I think I'll leave for a next PR
- Error message when too many files are added to a collection either in terms of number of files or total size
- Ability to preselect a list item using a # in the URL. I have a solution, but its not very robust, but I don't want to spend too much time on that, so I'll open this PR in the meantime.
- Tests


apps/documents/urls.py
```diff
@@ -0,0 +1,21 @@
+from django.urls import path
+
+from apps.documents import views
+
+app_name = "documents"
+
+urlpatterns = [
```
SmittieC, line 7: I see a possible factory coming, especially when we add the document index tab.

templates/documents/shared/tab_content.html
```diff
@@ -0,0 +1,33 @@
...
+    <div class="col-span-3">
+        <div id="{{ tab_name }}_details"></div>
+    </div>
+</div>
+
+
+<dialog id="add_new" class="modal">
+    <div class="modal-box max-w-none w-1/2 max-h-2/5 overflow-y-auto">
+        <form method="dialog">
+            <button class="btn btn-sm btn-circle btn-ghost absolute right-2 top-2">✕</button>
```
stephherbers, line 26: add aria-label

templates/documents/upload_files_modal.html
```diff
@@ -0,0 +1,93 @@
...
+            <div class="grid grid-cols-2 gap-1 border-b border-base-300 py-3 items-center">
+                <span class="col-span-1" x-text="file.name"></span>
+                <div class="col-span-1 flex flex-row gap-1 items-center">
+                    <textarea
+                        x-model="file_summaries[file.name]"
+                        class="col-span-1 textarea textarea-bordered w-full placeholder:text-gray-500"
+                        maxlength="400"
+                        placeholder="Enter a summary of the file..."
+                    ></textarea>
+                    <button type="button" class="btn btn-xs" @click="removeFile(index)">
```
stephherbers, line 30: also aria-label

stephherbers:
you could make a combined template to replace both templates/documents/collection_details.html and templates/documents/file_details.html: I believe for each you would need to get the object_type, edit_url, and delete_url, and then just for file: current_collections & available_collections

apps/documents/models.py
```diff
@@ -0,0 +1,26 @@
...
+    type = models.CharField(choices=RepositoryType.choices, default=RepositoryType.COLLECTION)
+    files = models.ManyToManyField("files.File", blank=False)
+
+    class Meta:
+        constraints = [models.UniqueConstraint(fields=["team", "type", "name"], name="unique_repository_per_team")]
+
+    @property
+    def size(self) -> float:
+        """Returns the size of this repository in megabytes"""
+        return bytes_to_megabytes(sum([bytes for bytes in self.files.values_list("content_size", flat=True)]))
```
snopoke, line 23: Does this work if a file has a null value for `content_size`?

apps/documents/urls.py
```diff
@@ -0,0 +1,21 @@
...
+    path("<slug:tab_name>", views.RepositoryHome.as_view(), name="repositories"),
+    # Files
+    path("files/", views.FileListView.as_view(), name="files_list"),
+    path("files/<int:id>/details", views.FileDetails.as_view(), name="file_details"),
+    path("files/upload", views.upload_files, name="upload_files"),
```
snopoke, line 12: nit: move this up one so that the `files/<int:id>` urls are grouped. Same below for `new_collection`

apps/documents/urls.py
```diff
@@ -0,0 +1,21 @@
...
+    path("files/<int:id>/delete", views.delete_file, name="delete_file"),
+    path("files/<int:id>/edit", views.edit_file, name="edit_file"),
+    # Collections
+    path("collections/", views.CollectionListView.as_view(), name="collections_list"),
```
snopoke, line 16: Should we rename the model to `Collection`? I think it would be good to start off with having the same terminology.

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+    def get_context_data(self, team_slug: str, tab_name: str, **kwargs):
+        context = {
+            "active_tab": "manage_files",
+            "title": "Manage Files",
+            "tab_name": tab_name,
+            "files_list_url": reverse("documents:files_list", kwargs={"team_slug": team_slug}),
+            "upload_files_url": reverse("documents:upload_files", kwargs={"team_slug": team_slug}),
+            "collections_list_url": reverse("documents:collections_list", kwargs={"team_slug": team_slug}),
+            "new_collection_url": reverse("documents:new_collection", kwargs={"team_slug": team_slug}),
+            "files_count": File.objects.filter(team__slug=team_slug).count(),
```
snopoke, line 32: Should this filter files? Do we want to show assistant files, generated files etc?

Maybe we need to add a 'purpose' field? For now we could filter anything that has and 'external_id' since all assistant and generated files should have that.

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+    details_url_name: str
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context["details_url_name"] = self.details_url_name
+        context["tab_name"] = self.tab_name
+        return context
+
+
+class BaseDetailsView(TemplateView, PermissionRequiredMixin):
```
snopoke, line 55: Any reason not to extend Django's `DetailView`? Just curious.

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+        files.append(
+            File.objects.create(
+                team=request.team,
+                name=uploaded_file.name,
+                file=uploaded_file,
+                summary=file_summaries[uploaded_file.name],
+            )
+        )
+    repo, _ = Repository.objects.get_or_create(
+        type=RepositoryType.COLLECTION, team=request.team, name=request.POST.get("collection_name")
```
snopoke, line 124: not: query based on ID rather than name

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+    file = get_object_or_404(File, team__slug=team_slug, id=id)
+    file.delete()
+    return redirect(reverse("documents:repositories", kwargs={"team_slug": team_slug, "tab_name": "files"}))
+
+
+@login_and_team_required
+@require_POST
+@permission_required("files.change_file")
+def edit_file(request, team_slug: str, id: int):
+    file = get_object_or_404(File.objects.defer("file"), team__slug=team_slug, id=id)
```
snopoke, line 142: I don't think the 'defer' is necessary here. It won't actually read the file when you load it from the DB. It will just contain the path / reference to the file.

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+    file.summary = request.POST.get("summary")
+
+    existing_collections = set(
+        file.repository_set.filter(type=RepositoryType.COLLECTION).values_list("name", flat=True)
+    )
+    collection_set = set()
+    # Handle new collections
+    for collection_name in request.POST.getlist("collections[]"):
+        collection_set.add(collection_name)
+        repo = get_object_or_404(Repository, team__slug=team_slug, type=RepositoryType.COLLECTION, name=collection_name)
```
snopoke, line 153: Rather do this query in bulk, or, don't load them at all:
```python
Repository.files.through.objects.create(file=file, repository_id=X)
```

You could even do that in bulk.

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+    file = get_object_or_404(File.objects.defer("file"), team__slug=team_slug, id=id)
+    file.name = request.POST.get("name")
+    file.summary = request.POST.get("summary")
+
+    existing_collections = set(
+        file.repository_set.filter(type=RepositoryType.COLLECTION).values_list("name", flat=True)
+    )
+    collection_set = set()
+    # Handle new collections
+    for collection_name in request.POST.getlist("collections[]"):
```
snopoke, line 151: I think this is missing a check to see if the collection is new or existing.

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+        collection_set.add(collection_name)
+        repo = get_object_or_404(Repository, team__slug=team_slug, type=RepositoryType.COLLECTION, name=collection_name)
+        repo.files.add(file)
+    file.save(update_fields=["name", "summary"])
+
+    # Remove from collections
+    for collection_name in existing_collections - collection_set:
+        repo = get_object_or_404(Repository, team__slug=team_slug, type=RepositoryType.COLLECTION, name=collection_name)
+        repo.files.remove(file)
```
snopoke, line 160: Cal also do this in bulk with `Repository.files.through.objects.filter(...).delete()`

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        collection = self.get_object()
+        context["edit_url"] = reverse(
+            "documents:edit_collection", kwargs={"team_slug": self.kwargs["team_slug"], "id": collection.id}
+        )
+        context["delete_url"] = reverse(
+            "documents:delete_collection", kwargs={"team_slug": self.kwargs["team_slug"], "id": collection.id}
+        )
```
snopoke, line 203: It's a bit unusual to see all URLs being added to the view context like this - the usual pattern is to use the `url` template tag. Curious about your reasoning for doing it this way.

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+        return context
+
+
+@require_POST
+@transaction.atomic()
+def new_collection(request, team_slug: str):
+    """Create a new collection"""
+    Repository.objects.create(type=RepositoryType.COLLECTION, team=request.team, name=request.POST.get("name"))
```
snopoke, line 211: This needs to handle uniqueness failures with a useful error message.

apps/files/models.py
```diff
@@ -17,6 +18,7 @@ class File(BaseTeamModel):
     content_type = models.CharField(blank=True)
     schema = models.JSONField(default=dict, blank=True)
     expiry_date = models.DateTimeField(null=True)
+    summary = models.TextField(max_length=400, blank=True)  # This is roughly 1 short paragraph
```
snopoke, line 21: I can see us needing to increase this limit. How about we go with 1024?

templates/documents/shared/details_base.html
```diff
@@ -0,0 +1,28 @@
...
+                Edit <i class="fa-solid fa-pencil fa-xs"></i>
+            </button>
+            <button type="submit" class="btn btn-sm join-item" x-cloak x-show="editing">
+                Update <i class="fa-solid fa-check fa-xs"></i>
+            </button>
+            <a type="button" class="btn btn-sm join-item" x-show="!editing" href="{{ delete_url }}">
+                <i class="fa-solid fa-trash htmx-hide"></i>
+            </a>
+            <button type="button" @click="editing = false" class="btn btn-sm join-item" x-cloak x-show="editing">
+                Cancel
```
snopoke, line 21: Add icon for consistency
```suggestion
                Cancel <i class="fa-solid fa-ban fa-xs"></i>
```

I think all buttons should be the same so all have text & icon, or only text, or only icon. My preference is only icons for a button bar like this.

templates/documents/shared/details_base.html
```diff
@@ -0,0 +1,28 @@
...
+    <div class="py-3 col-span-1">
+        <div class="join">
+            <button type="button" class="btn btn-sm join-item pl-3" x-show="!editing" @click="editing = true">
+                Edit <i class="fa-solid fa-pencil fa-xs"></i>
+            </button>
+            <button type="submit" class="btn btn-sm join-item" x-cloak x-show="editing">
+                Update <i class="fa-solid fa-check fa-xs"></i>
+            </button>
+            <a type="button" class="btn btn-sm join-item" x-show="!editing" href="{{ delete_url }}">
+                <i class="fa-solid fa-trash htmx-hide"></i>
```
snopoke, line 18: If you're going to use `htmx-hide` then there should also be an element with `htmx-show`:
```suggestion
                <i class="fa-solid fa-trash fa-xs htmx-hide"></i>
                <span class="loading loading-spinner loading-xs htmx-show"></span>
```

snopoke: Addressed in https://github.com/dimagi/open-chat-studio/pull/1317

apps/documents/models.py
SmittieC: No. I'll update to default to 0

apps/documents/urls.py
SmittieC: Are you suggesting we then rename it again to `Repository` when we introduce RAG, or introduce the notion of collection types then?

apps/documents/views.py
SmittieC: I'm not convinced that we should show assistant files here?

> Maybe we need to add a 'purpose' field?

I like this idea. I'll make a ticket to do so

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+            "new_collection_url": reverse("documents:new_collection", kwargs={"team_slug": team_slug}),
+            "files_count": File.objects.filter(team__slug=team_slug).count(),
+            "collections_count": Repository.objects.filter(
+                team__slug=team_slug, type=RepositoryType.COLLECTION
+            ).count(),
+        }
+        if tab_name == "files":
+            context["collections"] = Repository.objects.filter(
+                team__slug=team_slug, type=RepositoryType.COLLECTION
+            ).all()
```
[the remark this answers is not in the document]
SmittieC: The list is being loaded async. See [here](https://github.com/dimagi/open-chat-studio/blob/abe82b84d21a6014261848badc5170c19978a123/templates/documents/shared/tab_content.html#L14-L15).

apps/documents/views.py
SmittieC: I forgot that that is a thing. I'll use that to keep thing Django-y

apps/documents/views.py
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/1259/commits/e0439855ef17b20794aa758f5a3f72c2c127bde2. This addresses the two comments below as well

apps/documents/views.py
SmittieC: The `file_details.html` and `collection_detals.html` templates extends the `details_base.html` one, which expects these urls.

apps/documents/views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1259/commits/7d8614b273f9666ce10f429e43ce97d2fbcbbe7c

apps/files/models.py
SmittieC: Sure https://github.com/dimagi/open-chat-studio/pull/1259/commits/b461bfa628e68fc30a4240c2c933275d7c14a0e2

apps/documents/views.py
snopoke: Exactly - I don't think we should but currently this isn't doing any filtering so we are right?

apps/documents/urls.py
snopoke: I was assuming we'd use the same name. It's not exactly clear to me how we'll incorporate the RAG collections in the UI i.e. whether they should be entirely separate or kept together and differentiated with filtering and UI hints e.g. icons

templates/documents/upload_files_modal.html
```diff
@@ -24,7 +24,7 @@
                     <textarea
                         x-model="file_summaries[file.name]"
                         class="col-span-1 textarea textarea-bordered w-full placeholder:text-gray-500"
-                        maxlength="400"
+                        maxlength="max_summary_length"
```
snopoke, line 27:
```suggestion
                        maxlength="{{ max_summary_length }}"
```

apps/documents/models.py
```diff
@@ -0,0 +1,26 @@
...
+    type = models.CharField(choices=RepositoryType.choices, default=RepositoryType.COLLECTION)
+    files = models.ManyToManyField("files.File", blank=False)
+
+    class Meta:
+        constraints = [models.UniqueConstraint(fields=["team", "type", "name"], name="unique_repository_per_team")]
+
+    @property
+    def size(self) -> float:
+        """Returns the size of this repository in megabytes"""
+        return bytes_to_megabytes(sum([bytes or 0 for bytes in self.files.values_list("content_size", flat=True)]))
```
snopoke, line 23: Any reason not to do this sum in the database? There's event a [pg_size_pretty](https://www.postgresql.org/docs/11/functions-admin.html#FUNCTIONS-ADMIN-DBOBJECT) function which you could use if you wanted to.

apps/documents/views.py
```diff
@@ -0,0 +1,248 @@
...
+                queryset.annotate(similarity=name_similarity).filter(Q(similarity__gt=0.2)).order_by("-similarity")
+            )
+        return queryset
+
+
+class BaseDetailsView(DetailView, PermissionRequiredMixin):
+    pass
+
+
+class FileDetails(LoginAndTeamRequiredMixin, BaseDetailsView):
```
snopoke, line 83: nit: could add `LoginAndTeamRequiredMixin` to the `BaseDetailsView` class

apps/files/migrations/0005_alter_file_summary.py
snopoke, line 1: merge with previous migration

apps/documents/views.py
SmittieC: Yeah you are correct. It didn't do any filtering, but we are filtering now based on `external_id`. I do like your suggestion of having a purpose field

templates/documents/repositories.html
```diff
@@ -0,0 +1,47 @@
...
+        </div>
+
+        <div role="tablist" class="tabs tabs-bordered">
+            <input
+                type="radio"
+                name="repository"
+                role="tab"
+                class="tab"
+                aria-label="Files ({{ files_count }})"
+                @click="window.location.href = '{% url 'documents:repositories' request.team.slug 'files' %}'"
```
snopoke, line 23: Since you're loading a new page when changing tabs I think it makes more sense to make this styled links instead of an actual tab list.

templates/documents/upload_files_modal.html
```diff
@@ -0,0 +1,93 @@
...
+{% block modal_body %}
+    <div class="mt-2 flex flex-col gap-5">
+        <input
+            x-ref="fileUpload"
+            name="files"
+            @change="handleFileChange($event)"
+            type="file"
+            multiple
+            class="file-input file-input-bordered w-full max-w-xs"
+            accept=".txt,.pdf,.jpg,.png"/>
```
snopoke, line 12: why only these?

templates/documents/upload_files_modal.html
```diff
@@ -0,0 +1,93 @@
...
+    <script>
+        document.addEventListener('alpine:init', () => {
+            const megabyte_in_bytes = 1048576;
+            const byte_limit = megabyte_in_bytes * 512;
+            Alpine.data('modalData', () => ({
+                message: "",
+                uploaded_files: [],
+                file_summaries: {},
+                selected_collection: null,
+                total_upload_size_bytes: 0,
```
snopoke, line 54: Shouldn't this start at the current collection size? Or is this just to limit to 512MB per batch?

templates/documents/upload_files_modal.html
```diff
@@ -11,7 +11,7 @@
             class="file-input file-input-bordered w-full max-w-xs"
             accept=".txt,.pdf,.jpg,.png"/>
 
-        <select class="select select-bordered w-full" name="collection_name">
+        <select id="upload-collections-select" name="collection_names[]" multiple placeholder="Select collections" autocomplete="off">
```
snopoke, line 14: nit: we don't use this field naming convention anywhere else
```suggestion
        <select id="upload-collections-select" name="collection_names" multiple placeholder="Select collections" autocomplete="off">
```

apps/documents/views.py
SmittieC: Actually, I think we should only allow a single collection selection like before

apps/documents/models.py
SmittieC: No reason, I'll make an update, thanks for the suggestion

templates/documents/upload_files_modal.html
SmittieC: Yeah just is just a batch upload limit. I copied this code from the `input_bar.html`. In that context, it's the max limit that assistants can handle. In this context, it doesn't really make much sense to limit the total file size come to think of it, since the limits are on:
1. Individual file size
2. The number of files per collection
3. The sum of all files per team

I haven't really done any of those checks yet, so I'm going to remove this code and create a ticket for the above size checks, otherwise the PR just gets too large.

templates/documents/repositories.html
SmittieC: I'm curious what your reasoning behind this is?

templates/documents/upload_files_modal.html
SmittieC: Oh this was just to get started with something. I [added more](https://github.com/dimagi/open-chat-studio/pull/1259/commits/fb74a2fedf02be12af4b904c7fc63ad4e4f445b2)

apps/documents/views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1259/commits/bccd4ccc1889c988b6f6c4775b61d5a7854be9d6

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+            .filter(type=RepositoryType.COLLECTION, team__slug=self.kwargs["team_slug"])
+            .order_by("-created_at")
+        )
+
+        search = self.request.GET.get("search")
+        if search:
+            name_similarity = TrigramSimilarity("name", search)
+            queryset = (
+                queryset.annotate(similarity=name_similarity).filter(Q(similarity__gt=0.2)).order_by("-similarity")
+            )
```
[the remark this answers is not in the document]
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/1259/commits/5faea2cdd4a4b206a9da25f84e3563b6d3798e91

apps/documents/urls.py
snopoke: Might be good to give Brian a demo and get some input

templates/documents/collection_details.html
```diff
@@ -0,0 +1,35 @@
+{% extends 'documents/shared/details_base.html' %}
+
+{% block detail_content %}
+    <div class="col-span-3">
+        <input x-cloak x-show="editing" name="name" class="input input-bordered" value="{{ object.name }}"></input>
+        <h1 x-show="!editing" class="pg-title">{{ object.name }}</h1>
+
+        <dl class="grid grid-cols-2 gap-1">
+            <div class="py-3 col-span-1">
```
snopoke, line 9: `col-span-1` isn't required for these since it's the default

templates/documents/collection_details.html
```diff
@@ -0,0 +1,35 @@
...
+                <dt class="text-sm font-medium leading-6">Uploaded at</dt>
+                <dd class="text-sm col-span-2">{{ object.created_at }}</dd>
```
snopoke, line 11: what does the col span do here?

templates/documents/collection_details.html
```diff
@@ -0,0 +1,35 @@
+{% extends 'documents/shared/details_base.html' %}
+
+{% block detail_content %}
+    <div class="col-span-3">
+        <input x-cloak x-show="editing" name="name" class="input input-bordered" value="{{ object.name }}"></input>
+        <h1 x-show="!editing" class="pg-title">{{ object.name }}</h1>
+
+        <dl class="grid grid-cols-2 gap-1">
+            <div class="py-3 col-span-1">
+                <dt class="text-sm font-medium leading-6">Uploaded at</dt>
```
snopoke, line 10: `leading` changes the line height which is usually used for prose. For vertical spacing you should use margin or padding. If you need alignment then `align-x`

templates/documents/file_details.html
snopoke, line 1: similar css comments as above

templates/documents/new_collection_modal.html
```diff
@@ -0,0 +1,26 @@
+{% extends 'documents/shared/modal_body.html' %}
+
+{% block modal_body %}
+    <div class="mt-2 flex flex-col gap-5 form-control w-full">
+        <label class="label font-bold" for="id_name">
+            <div>Name</div>
```
snopoke, line 6: unnecessary `<div>`

templates/documents/repositories.html
```diff
@@ -0,0 +1,47 @@
+{# Generic template to render a list of objects with a 'new' button #}
+{% extends 'web/app/app_base.html' %}
+{% block app %}
+    {% load i18n %}
+    <div class="app-card" x-data>
+        <div class="grid grid-cols-6">
+            <div class="col-span-5">
```
snopoke, line 7: a better option here might be to use `flex` with `justify-between` which deals better with different screen sizes

templates/documents/shared/details_base.html
```diff
@@ -0,0 +1,28 @@
+<form
+    id="detail-form"
+    action="{{ edit_url }}"
+    method="POST"
+    class="m-2 grid grid-cols-4" x-data="{editing: false}">
```
snopoke, line 5: why `gid-cols-4`?

templates/documents/shared/details_base.html
```diff
@@ -0,0 +1,28 @@
...
+    {% endblock %}
+    <div class="py-3 col-span-1">
+        <div class="join">
+            <button type="button" class="btn btn-sm join-item pl-3" x-show="!editing" @click="editing = true">
+                <i class="fa-solid fa-pencil fa-xs"></i>
+            </button>
+            <button type="submit" class="btn btn-sm join-item" x-cloak x-show="editing">
+                <i class="fa-solid fa-check fa-xs"></i>
+            </button>
+            <a type="button" class="btn btn-sm join-item" x-show="!editing" href="{{ delete_url }}">
```
snopoke, line 17: This should be a form that does a POST or DELETE request. You can just wrap a button in a form and use the button to submit the form. Here's an example: https://github.com/dimagi/open-chat-studio/blob/2b45c04f0a8fbe163bbc37dfb22ab562a452016f/templates/account/components/api_keys.html#L44-L50

templates/documents/shared/tab_content.html
```diff
@@ -0,0 +1,33 @@
+<div class="my-2 border-t border-base-300 grid grid-cols-5">
+    <div class="col-span-2 border-r border-base-300">
+        <div class="flex flex-row my-2 justify-evenly items-center">
+            <input class="input input-bordered join-item" type="search"
```
snopoke, line 4: this has the `join-item` class but it's not nested in a `join` element

templates/documents/repositories.html
SmittieC: That makes sense. See https://github.com/dimagi/open-chat-studio/pull/1259/commits/39a343c4ab1629ad66674a4ce75ead4114ca823e

Screenshot
image

templates/documents/upload_files_modal.html
```diff
@@ -0,0 +1,74 @@
...
+                        maxlength="{{ max_summary_length }}"
+                        placeholder="Enter a summary of the file..."
+                    ></textarea>
+                    <button aria-label="Remove File" title="Remove File" type="button" class="btn btn-xs" @click="removeFile(index)">
+                        <i class="fa-solid fa-trash htmx-hide"></i>
+                    </button>
+                </div>
+            </div>
+        </template>
+        <input type="hidden" name="file_summaries" :value="JSON.stringify(file_summaries)" />
```
snopoke, line 36: just curious why you decided to use JSON here instead of standard form fields which already being displayed.

apps/documents/views.py
```diff
@@ -180,30 +180,14 @@ def _update_collection_membership(file: File, collection_names: list[str]):
...
     paginate_by = 10
     details_url_name = "documents:collection_details"
     tab_name = "collections"
     permission_required = "documents.view_repository"
 
-    def get_queryset(self):
-        queryset = (
-            super()
-            .get_queryset()
-            .filter(type=RepositoryType.COLLECTION, team__slug=self.kwargs["team_slug"])
```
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1259/commits/8b2aa00b2ff9fed87926eff6c4af922bc0ee7fd5

apps/documents/models.py
```diff
@@ -0,0 +1,26 @@
...
+
+    class Meta:
+        constraints = [models.UniqueConstraint(fields=["team", "type", "name"], name="unique_repository_per_team")]
+
+    @property
+    def size(self) -> float:
+        """Returns the size of this repository in megabytes"""
+        return bytes_to_megabytes(self.files.aggregate(bytes=models.Sum("content_size"))["bytes"])
+
+    def file_names(self) -> list[str]:
```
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1259/commits/ac3366bf7f0aeea5da6eaa91ca38000c82286314

templates/documents/upload_files_modal.html
```diff
@@ -9,7 +9,7 @@
             type="file"
             multiple
             class="file-input file-input-bordered w-full max-w-xs"
-            accept=".txt,.pdf,.jpg,.png"/>
+            accept=".txt,.pdf,.doc,.docx,.xls,.xlsx,.csv,.jpg,.jpeg,.png,.gif,.bmp,.webp,.svg,.mp4,.mov,.avi,.mp3,.wav"/>
```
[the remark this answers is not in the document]
SmittieC: Sure: https://github.com/dimagi/open-chat-studio/pull/1259/commits/8cb1766341ebf88e4f5f96898fd63e90937be4bc

apps/documents/views.py
```diff
@@ -0,0 +1,231 @@
...
+    try:
+        Repository.objects.create(type=RepositoryType.COLLECTION, team=request.team, name=request.POST.get("name"))
+    except IntegrityError:
+        messages.error(request, "A collection with that name already exists.")
+    return redirect(reverse("documents:repositories", kwargs={"team_slug": team_slug, "tab_name": "collections"}))
+
+
+@login_and_team_required
+@permission_required("documents.delete_repository")
+def delete_collection(request, team_slug: str, pk: int):
```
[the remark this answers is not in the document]
SmittieC: I'll address this in [this PR](https://github.com/dimagi/open-chat-studio/pull/1317/files), since I'm swapping the anchor tag for the `action_ajax` template

templates/documents/upload_files_modal.html
SmittieC: I simplified it [here](https://github.com/dimagi/open-chat-studio/pull/1259/commits/d2f5a2af24fd0df17c3b3c3315d00c5998fffc31). I did not think of this way back

templates/documents/shared/details_base.html
SmittieC: See [this comment](https://github.com/dimagi/open-chat-studio/pull/1259#discussion_r2003210314)

SmittieC:
@snopoke I chatted with Brian who gave some UI feedback. I'll create a new PR to this one

SmittieC:
> I was assuming we'd use the same name. It's not exactly clear to me how we'll incorporate the RAG collections in the UI i.e. whether they should be entirely separate or kept together and differentiated with filtering and UI hints e.g. icons

@snopoke The idea was to have two types of repositories. One of type collection and another of type index (or document_index) for the RAG stuff.

We could rename this model to `Collection` and have a flag or something on it to indicate if it's a RAG collection.

Do you have a preference?

SmittieC:
> > I was assuming we'd use the same name. It's not exactly clear to me how we'll incorporate the RAG collections in the UI i.e. whether they should be entirely separate or kept together and differentiated with filtering and UI hints e.g. icons
> 
> @snopoke The idea was to have two types of repositories. One of type collection and another of type index (or document_index) for the RAG stuff.
> 
> We could rename this model to `Collection` and have a flag or something on it to indicate if it's a RAG collection.
> 
> Do you have a preference?

I chatted with Simon and we agreed that `Collection` made more sense. [Here's](https://github.com/dimagi/open-chat-studio/pull/1259/commits/8786e899f56819a6c3983e47fc57c251769d4732) the change

snopoke (approved):

apps/documents/views.py
```diff
@@ -0,0 +1,222 @@
...
+@permission_required("files.change_file")
+@transaction.atomic()
+def edit_file(request, team_slug: str, pk: int):
+    file = get_object_or_404(File, team__slug=team_slug, id=pk)
+    file.name = request.POST.get("name")
+    file.summary = request.POST.get("summary")
+    file.save(update_fields=["name", "summary"])
+    _update_collection_membership(file=file, collection_names=request.POST.getlist("collections"))
+
+    return redirect(reverse("documents:collections", kwargs={"team_slug": team_slug, "tab_name": "files"}))
```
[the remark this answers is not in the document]
SmittieC: Thanks. I'm making another tiny update, so will throw this in there. I think it's good to keep things consistent

templates/documents/shared/list.html
```diff
@@ -0,0 +1,43 @@
...
+
+<ul x-data="{selected_item: null}">
+    {% for object in page_obj %}
+        <li
+            class="text-left px-3 py-3 border-b border-base-300 hover:cursor-pointer"
+            hx-get="{% url details_url_name request.team.slug object.id %}"
+            hx-trigger="click"
+            hx-target="#{{ tab_name }}_details"
+            @click="selected_item = {{ object.id }}"
+            :class="selected_item === {{ object.id }} ? 'bg-gray-500': 'hover:bg-base-200'"
```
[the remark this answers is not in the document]
SmittieC: Yes, but no. I think it might be better to stick to the `hover:bg-base-200`, even if the row is selected. Now it looks like this:

image

Screenshot from 2025-03-25 10-23-40

Issue #1161, opened 2025-02-07
Improve the table UI for filtering


snopoke:
The current table filtering is very limited and table specific. Investigate options for building a more comprehensive and flexible table filter component.

A product that does this quite well is https://langfuse.com/

* Maybe consider https://django-filter.readthedocs.io/en/stable/

stephherbers:
update: PR out for review 2/27

PR #1260, opened 2025-02-27
Experiment Session Filter

stephherbers:
## Description

A new and much improved filter [ticket](https://github.com/dimagi/open-chat-studio/issues/1161)
Design doc [here](https://docs.google.com/document/d/1TYQcQ2AwcU4vcKwVlZDkkAyd7rTMzXPVfSl08VFYT2Y/edit?tab=t.0)

PRE-QA
Screenshot 2025-02-27 at 2 54 40 PM

POST-QA
image

Dark mode:
image

Using django-filter turned out to be a bit more complicated than I thought- I still think it's worth looking into, but not worth blocking the rollout of this feature.

I still would like to work on getting the column types in the html templated so this can be used elsewhere in the codebase, but that can be a followup ticket as well

## User Impact

It will make querying the data much easier and granular and hopefully they will not have to export the data just to filter it.

Also the filters are in the url so users can copy and send the link to others with the filters applied which I bet will come in handy

### Demo

[Pre-QA Loom](https://www.loom.com/share/8d0f1a07d16e4fe2af13b5d903d44849?sid=7b0a0261-b306-4dde-bfed-61cc304914a5)
[Post-QA Loom](https://www.loom.com/share/8600bd1dbff2428cb8a619d1f2a5439a?sid=8fcdc37c-eddc-43ec-af69-55e1aa70b728)

### Docs and Changelog

changelog yes indeed


templates/experiments/single_experiment_home.html
```diff
@@ -212,54 +212,215 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+                          >
+                            <template x-for="operator in filter.availableOperators" :key="operator">
+                              <option :value="operator" x-text="operator"></option>
+                            </template>
+                          </select>
+                        </div>
+                        <template x-if="filter.operator">
+                          <div class="relative">
+                          <!-- Multi-select dropdown-->
+                            <template x-if="['tags', 'versions'].includes(filter.column) && ['any of', 'all of'].includes(filter.operator)">
```
snopoke, line 294: can this use the column type or is it particularly specific to these two fields?
```suggestion
                            <template x-if="filterData.columns[filter.column]?.type === 'choice' && ['any of', 'all of'].includes(filter.operator)">
```

templates/experiments/single_experiment_home.html
```diff
@@ -212,54 +212,215 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+                          aria-label="Remove filter"
+                        >
+                          <span class="text-lg">×</span>
+                        </button>
+                      </div>
+                    </template>
+                  <!-- Add Filter Button -->
+                    <button
+                      @click="addFilter()"
+                      class="flex items-center gap-2 px-4 py-2 mt-3 text-gray-700 border border-gray-200 rounded-lg hover:bg-gray-50 text-sm"
```
snopoke, line 400: For consistency with the rest of the UI it might be better to use the DaisyUI button classes: https://v4.daisyui.com/components/button/

Possibly similar comment for other buttons. Also not a blocker

templates/experiments/single_experiment_home.html
```diff
@@ -335,47 +490,7 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
-                      <span class="h4">
-                        ${ escape(item.name || 'No Name') }
-                      </span>
-                      <span class="text-muted"> (${ escape(item.identifier) })</span>
-                    </div>`;
-          }
-        },
-      });
-
+      showTabContent('allsessions');
```
snopoke, line 493: Why did this need to be added? Also it won't work if the user doesn't have 'view_chat' permissions

templates/experiments/single_experiment_home.html
```diff
@@ -389,4 +504,139 @@ <h3 class="font-bold text-lg">Parent Experiments:
 
     window.addEventListener('hashchange', selectTabFromHash);
   </script>
+  <script>
+    const tags = {{ available_tags|safe|default:"[]" }};
+    const versionsList = {{ experiment_versions|safe|default:"[]" }};
```
snopoke, line 509: It would be safer to pass these using the [json_script](https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#std-templatefilter-json_script) filter

templates/experiments/single_experiment_home.html
```diff
@@ -389,4 +504,139 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+          filters: [],
+          columns: {
+            'participant': {type: 'string', operators: ['equals', 'contains', 'starts_with', 'ends_with']},
+            'last_message': {type: 'timestamp', operators: ['on', 'before', 'after']},
+            'tags': {type: 'choice', operators: ['any of', 'all of'], options: tags },
+            'versions': {type: 'choice', operators: ['any of', 'all of'], options: versionsList}
+          }
+        },
+        init() {
+          htmx.process(this.$el);
```
snopoke, line 523: can you say more about what this line is for

templates/experiments/single_experiment_home.html
```diff
@@ -389,4 +504,139 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+        clearAllFilters() {
+          this.filterData.filters = [];
+          this.addFilter();
+          this.filterData.showFilters = false;
+          this.triggerFilterChange();
+        },
+        triggerFilterChange() {
+          let filterParams = {};
+          this.filterData.filters.forEach((filter, i) => {
+            if (filter.column && filter.operator && (filter.value || filter.selectedValues.length > 0)) {
```
snopoke, line 605: I notice above that you're checking `Array.isArray(filter.selectedValues)`. Should that also be done here?

templates/experiments/single_experiment_home.html
```diff
@@ -389,4 +504,139 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+              filterParams[`filter_${i}_column`] = filter.column;
+              filterParams[`filter_${i}_operator`] = filter.operator;
+              if (['any of', 'all of'].includes(filter.operator)) {
+                filterParams[`filter_${i}_value`] = JSON.stringify(filter.selectedValues);
```
snopoke, line 609: Won't `filter.value` already be the json formatted value based on the code in `updateFilterValue`

templates/experiments/single_experiment_home.html
```diff
@@ -389,4 +504,139 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+          const sessionsTableUrl = document.getElementById('sessions-table').getAttribute('hx-get');
+          const filterUrl = new URL(sessionsTableUrl, window.location.origin);
+          Object.entries(filterParams).forEach(([key, value]) => {
+            filterUrl.searchParams.set(key, value);
+          });
+          const showAllCheckbox = document.querySelector('input[name="show-all"]');
+          if (showAllCheckbox && showAllCheckbox.checked) {
+            filterUrl.searchParams.set('show-all', 'on');
+          }
+          htmx.ajax('GET', filterUrl.toString(), {
```
snopoke, line 634: Rather than generating the URL it might be easier to pass the data in the `values` field of the context object:

`htmx.ajax('GET', sessionsTableUrl, {..., values: filterParams})`

templates/experiments/components/archived_experiment_details.html
snopoke, line 1: Is this file (and the other two in the `assistants` directory) used anywhere?

snopoke (commented):
This is looking great @stephherbers. Some comments but nothing major.

templates/experiments/components/archived_experiment_details.html
stephherbers: mistakingly added from PR of long ago ☠️

templates/experiments/single_experiment_home.html
stephherbers: I initially included it believing it was necessary to ensure HTMX would recognize/process attributes within the Alpine-rendered components. But I just tested with commenting it out and it works fine. 
I think since the filter functionality triggers HTMX actions via `htmx.ajax()` calls rather than relying on HTMX attributes being processed after Alpine renders elements. Since there's no race condition between Alpine rendering and HTMX processing here, the line is unnecessary

templates/experiments/single_experiment_home.html
stephherbers: changed for all but one that was giving me a particularly hard time

templates/experiments/single_experiment_home.html
stephherbers: much cleaner([a71e63e](https://github.com/dimagi/open-chat-studio/pull/1260/commits/a71e63e057a425126c006ceef0e7e9d0d752ca3d))

apps/experiments/filters.py
```diff
@@ -42,10 +43,17 @@ def build_tags_filter(operator, value):
...
+                chat_id__in=ExperimentSession.objects.filter(chat__tags__name__in=selected_tags)
+                .values("id")
+                .annotate(
+                    matching_tag_count=Count(
+                        "chat__tags__name", filter=Q(chat__tags__name__in=selected_tags), distinct=True
+                    )
+                )
+                .filter(matching_tag_count=len(selected_tags))
+                .values_list("id", flat=True)
+            )
```
snopoke, line 56: I think a simpler way of doing this is to use an exists query:

```python
content_type = ContentType.objects.get_for_model(Chat)
conditions = Q()
for tag in selected_tags:
  conditions &= Exists(
    CustomTaggedItem.objects.filter(
        object_id=OuterRef("id"),
        content_type_id=content_type,
        tag__name=name,
    )
  )
return conditions
```

snopoke (commented):
One comment on the filter query for tags and then I did some testing locally and put together a QA doc: https://docs.google.com/document/d/1u5BHH4bKjggG3XZMk_-Y7oke6ft6nlxZb-J3cuE46Uk/edit?tab=t.0

It's mostly small stuff but I think it'd be good to get the 'MUST' section done before merging.

apps/experiments/views/experiment.py
```diff
@@ -161,19 +167,46 @@ def get_queryset(self):
...
+        return query_set
+
+    def apply_dynamic_filters(self, query_set):
+        filter_conditions = Q()
+        filter_applied = False
 
-        if tags_query := self.request.GET.get("tags"):
-            tags = tags_query.split("&")
-            query_set = query_set.filter(chat__tags__name__in=tags).distinct()
+        for i in range(30):  # arbitrary number higher than any # of filters we'd expect
```
SmittieC, line 179: we could also do a while loop here and increment the index variable. Something like
```python
idx = 0
while True:
   ...
   idx += 1
```

templates/experiments/single_experiment_home.html
SmittieC: Hmm, I think somewhere the tab is being hidden unexpectedly rather. It should be showing up by default, since the `allsessions` tab is checked by default

apps/experiments/views/experiment.py
snopoke: It was originally a while loop. I suggested changing it to a for loop to avoid any chance of runaway

SmittieC: what about doing
```python
while filter_column and filter_operator and filter_value:
   ...
```

stephherbers:
@snopoke @SmittieC updated the description after the qa-doc feedback, for the rest not impleneted in the doc, I'll make tickets for

templates/experiments/single_experiment_home.html
```diff
@@ -212,54 +212,244 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+                          <div class="dropdown dropdown-bottom">
+                            <label
+                              tabindex="0"
+                              class="block w-40 px-3 border border-gray-300 rounded-lg bg-white text-sm cursor-pointer flex justify-between items-center"
+                              style="height: 2rem;"
+                            >
+                              <span x-text="filter.column ? filterData.columns[filter.column].label : 'Column'"></span>
+                              <i class="fa-solid fa-caret-down fa-sm"></i>
+                            </label>
+                            <ul tabindex="0" class="dropdown-content z-[1] menu p-2 bg-base-100 rounded-box border border-gray-300 w-40">
```
snopoke, line 283: styling:
```suggestion
                            <ul tabindex="0" class="dropdown-content z-[1] mt-1 menu p-2 bg-base-100 rounded-lg border border-gray-300 w-40">
```

(apply the same to the operator selector)

templates/experiments/single_experiment_home.html
```diff
@@ -389,4 +532,238 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+            params.set('show-all', 'on');
+          } else {
+            params.delete('show-all');
+          }
+
+          const newUrl = window.location.pathname + (params.toString() ? '?' + params.toString() : '') + window.location.hash;
+          window.history.replaceState({}, '', newUrl);
+        },
+
+        triggerFilterChange(updateUrl = true) {
```
snopoke, line 729: When adding a new filter and selecting the column etc, this gets called and makes a request even though the filter isn't complete yet. I'm not sure the best way to fix that but one option might be to pass the index of the filter being edited and check if it's valid before continuing.

templates/experiments/single_experiment_home.html
```diff
@@ -212,54 +212,244 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+                            <label
+                              tabindex="0"
+                              class="block w-40 px-3 border border-gray-300 rounded-lg bg-white text-sm cursor-pointer flex justify-between items-center"
+                              style="height: 2rem;"
+                            >
+                              <span x-text="filter.column ? filterData.columns[filter.column].label : 'Column'"></span>
+                              <i class="fa-solid fa-caret-down fa-sm"></i>
+                            </label>
+                            <ul tabindex="0" class="dropdown-content z-[1] menu p-2 bg-base-100 rounded-box border border-gray-300 w-40">
+                              <li><a @click="filter.column = ''; updateOperatorsByColumn(index)">Column</a></li>
```
snopoke, line 284: I think this line should be removed. It is always displayed and clicking on it doesn't do anything.

templates/experiments/single_experiment_home.html
```diff
@@ -212,54 +212,244 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+                    x
+                  </button>
+                </div>
+
+              <!-- Filters Panel -->
+                <div
+                  x-show="filterData.showFilters"
+                  class="absolute left-0 mt-2 p-4 bg-white border border-gray-200 rounded-lg shadow-lg z-10 w-fit"
+                >
+                  <div class="space-y-3">
```
snopoke, line 265: nit: make it more compact
```suggestion
                  <div class="space-y-2">
```

templates/experiments/single_experiment_home.html
```diff
@@ -212,54 +212,244 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+                    <template x-for="(filter, index) in filterData.filters" :key="index">
+                      <div class="flex items-center gap-3">
+                      <!-- Where/AND Label -->
+                        <div class="text-gray-500 font-medium w-16">
+                          <span x-text="index === 0 ? 'Where' : 'AND'"></span>
+                        </div>
+                      <!-- Column Selector -->
+                        <div class="relative py-2">
```
snopoke, line 273: spacing is handled by `gap-2` and `space-y-2`
```suggestion
                        <div class="relative">
```

templates/experiments/single_experiment_home.html
```diff
@@ -212,54 +212,244 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+                </div>
+
+              <!-- Filters Panel -->
+                <div
+                  x-show="filterData.showFilters"
+                  class="absolute left-0 mt-2 p-4 bg-white border border-gray-200 rounded-lg shadow-lg z-10 w-fit"
+                >
+                  <div class="space-y-3">
+                    <template x-for="(filter, index) in filterData.filters" :key="index">
+                      <div class="flex items-center gap-3">
```
snopoke, line 267: nit: more compact (also matches y gap)
```suggestion
                      <div class="flex items-center gap-2">
```

templates/experiments/single_experiment_home.html
```diff
@@ -212,54 +212,244 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+                            <ul tabindex="0" class="dropdown-content z-[1] menu p-2 bg-base-100 rounded-box border border-gray-300 w-40">
+                              <li><a @click="filter.column = ''; updateOperatorsByColumn(index)">Column</a></li>
+                              <template x-for="(config, column) in filterData.columns" :key="column">
+                                <li><a @click="filter.column = column; updateOperatorsByColumn(index)" x-text="config.label"></a></li>
+                              </template>
+                            </ul>
+                          </div>
+                        </div>
+                      <!-- Operator Selector -->
+                        <div x-show="filter.column" class="relative py-1 px-1">
```
snopoke, line 292: same as above re spacing
```suggestion
                        <div x-show="filter.column" class="relative">
```

templates/experiments/single_experiment_home.html
```diff
@@ -212,54 +212,244 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
+                            </label>
+                            <ul tabindex="0" class="dropdown-content z-[1] menu p-2 bg-base-100 rounded-box border border-gray-300 w-40">
+                              <template x-for="operator in filter.availableOperators" :key="operator">
+                                <li><a @click="filter.operator = operator; triggerFilterChange()" x-text="operator"></a></li>
+                              </template>
+                            </ul>
+                          </div>
+                        </div>
+                        <template x-if="filter.operator">
+                          <div class="relative py-1 pl-1">
```
snopoke, line 310: same as above re spacing
```suggestion
                          <div class="relative">
```

snopoke (commented):
Also some more QA feedback in the doc

templates/experiments/single_experiment_home.html
```diff
@@ -375,19 +374,21 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
-                              <div class="relative">
+                              <div class="relative" x-init="if (!('isValidDate' in filter)) filter.isValidDate = true">
                                 <input
                                   type="date"
                                   x-model="filter.value"
                                   @input="filter.isValidDate = $event.target.validity.valid"
                                   @change="filter.isValidDate && triggerFilterChange()"
-                                  class="input input-bordered input-sm w-40"
+                                  class="w-40 block px-3 border border-base-200 rounded-lg bg-base-100 text-sm"
+                                  style="height: 2rem;"
```
snopoke, line 391: This is equivalent to using the `h-8` class

apps/experiments/models.py
```diff
@@ -268,6 +268,10 @@ def get_version_name(self):
             return "unreleased"
         return f"v{self.version_number}"
 
+    def get_version_name_list(self):
+        """Returns list of version names in form of v + version number including working version."""
+        return [v.get_version_name() for v in self.versions.all()] + [f"v{self.version_number}"]
```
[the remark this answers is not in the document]
SmittieC: This should do the trick

```python
from django.db.models.functions import Concat, Cast
from django.db.models import CharField

self.versions.annotate(friendly_name=Concat(Value("v"), Cast(F("version_number"), output_field=CharField()))).values_list("friendly_name", flat=True)
```

templates/experiments/single_experiment_home.html
stephherbers: I tried add the `h-8` class, but it doesn't apply unless I explicitly put it in the style, so I'm not quite sure where it's getting overridden

snopoke: Did you run `npm run dev` and reload the page without cache?

Not a big issue - fine to leave it if you can't make it work.

apps/experiments/models.py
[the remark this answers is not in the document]
stephherbers: https://github.com/dimagi/open-chat-studio/pull/1260/commits/d88989f5d6f5f40338e36a5e0e4311f3f81b4ac2 the 'unreleased' is not added it's sill just 'v1', 'v2'... including the working version

apps/experiments/filters.py
```diff
@@ -0,0 +1,96 @@
...
+    """Build filter condition for timestamp"""
+    try:
+        date_value = datetime.strptime(value, "%Y-%m-%d").date()
+        if operator == "on":
+            return Q(last_message_created_at__date=date_value)
+        elif operator == "before":
+            return Q(last_message_created_at__date__lt=date_value)
+        elif operator == "after":
+            return Q(last_message_created_at__date__gt=date_value)
+    except (ValueError, TypeError):
```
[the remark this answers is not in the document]
stephherbers: there is front end validation as well for a better indicator the user, this is so it doesn't error out in a worse way

stephherbers:
@snopoke the export is not compatible- turns out it's a decent sized change so I started it in a separate PR

templates/experiments/single_experiment_home.html
stephherbers: yep tried both of those and no luck

snopoke (approved):
@stephherbers I pushed a few small changes from my final testing. I think this is GTG now. I've updated the [QA doc](https://docs.google.com/document/d/1u5BHH4bKjggG3XZMk_-Y7oke6ft6nlxZb-J3cuE46Uk/edit?tab=t.cnmptpvxy4wy) with some outstanding  items but I don't think they need to block releasing this.

PR #1261, opened 2025-02-28
Updating Render Template Content

vaibhavK1199:
Updated render template content using ParticipantDataProxy .


apps/pipelines/nodes/nodes.py
```diff
@@ -62,32 +61,32 @@ class RenderTemplate(PipelineNode):
...
-            # As a last resort, just set the all the variables in the template to the input
-            content = all_variables(input)
-        template = SandboxedEnvironment().from_string(self.template_string)
-        output = template.render(content)
+            participant_data_proxy = ParticipantDataProxy.from_state(state)
+            content = {
+                "input": input,
+                "participant_details": {
+                    "identifier": participant_data_proxy.get("identifier"),
+                    "platform": participant_data_proxy.get("platform"),
```
snopoke, line 77: These actually come from the `Participant` object: https://github.com/dimagi/open-chat-studio/blob/main/apps/experiments/models.py#L1268-L1273

Probably the best place to load that from would be the session: `state.experiment_session.participant`

vaibhavK1199: `content = {
                "input": input,
                "participant_details": {
                    "identifier": state.experiment_session.participant.identifier,
                    "platform": state.experiment_session.participant.platform,
                },
                "participant_data": json.dumps(state.experiment_session.participant_data_from_experiment, indent=4),
                "participant_schedules": state.experiment_session.participant.get_schedules_for_experiment(state.experiment_session.experiment, as_dict=True, include_complete=True),
                "temp_state": state.get("temp_state", {}),
            }
`

can we implement it like this as the ParticipantDataProxy.get() is expecting a single argument but is getting called with two arguments causing multiple test failures.

snopoke: Yes, you can do it that way.

snopoke (approved):
Please also prepare a docs PR to update the template node documentation: https://github.com/dimagi/open-chat-studio-docs/blob/main/docs/concepts/pipelines/nodes.md#template

apps/pipelines/nodes/nodes.py
```diff
@@ -62,32 +61,52 @@ class RenderTemplate(PipelineNode):
...
+                "messages": state.get("messages", []),
+                "temp_state": state.get("temp_state", {}),
+            }
+
+            if state.get("messages"):
+                for message in state["messages"]:
+                    if isinstance(message, dict):
+                        content.update(message)
+                    else:
+                        content["input"] = message
```
snopoke, line 83: I'm not following what this is for. Can you explain.

snopoke: OK, but why are we adding all the messages directly into the content dict? It's going to continually overwrite the previous values. I also don't think it's useful to have `state.messages` in the template context.

apps/pipelines/nodes/nodes.py
```diff
@@ -62,32 +61,52 @@ class RenderTemplate(PipelineNode):
...
-                    # e.g. it was just a string or an int
-                    content = all_variables(input)
-        except json.JSONDecodeError:
-            # As a last resort, just set the all the variables in the template to the input
-            content = all_variables(input)
-        template = SandboxedEnvironment().from_string(self.template_string)
-        output = template.render(content)
+            content = {
+                "input": input,
+                "messages": state.get("messages", []),
```
snopoke, line 74: This can be removed. I don't think it's useful to have this in the context.

apps/pipelines/nodes/nodes.py
```diff
@@ -62,32 +61,52 @@ class RenderTemplate(PipelineNode):
...
+
+            if hasattr(state, "experiment_session") and state.experiment_session:
+                participant = state.experiment_session.participant
+                content.update(
+                    {
+                        "participant_details": {
+                            "identifier": participant.identifier,
+                            "platform": participant.platform,
+                        },
+                        "participant_data": json.dumps(
```
snopoke, line 93: this should be a dict, not a json string so that the template can access it

snopoke (commented):
In addition to the comments, please add a test which uses some of the new context keys. An example template might be:

```
input: {{input}}
temp_state.my_key: {{temp_state.my_key}}
participant_id: {{participant_details.identifier}}
```
```

**Review by snopoke — APPROVED** [review]
Great, thanks for the updates.

**Comment by snopoke** [comment]
There is still one test failure:

```
FAILED apps/pipelines/tests/test_code_node.py::test_render_template_with_context_keys - apps.pipelines.exceptions.PipelineNodeRunError: Error rendering template: 'participant_details' is undefined
```

File: apps/pipelines/tests/test_code_node.py [hunk 5]
```diff
@@ -308,3 +308,24 @@ def main(input, **kwargs):
...
+
+
+@django_db_with_data(available_apps=("apps.service_providers",))
+@mock.patch("apps.pipelines.nodes.base.PipelineNode.logger", mock.Mock())
+def test_render_template_with_context_keys(pipeline, experiment_session):
+    nodes = [
+        start_node(),
+        render_template_node(
+            "input: {{input}}, temp_state.my_key: {{temp_state.my_key}}, "
+            "participant_id: {{participant_details.identifier}}"
```
snopoke, line 320: This should fix the test:
```suggestion
            "participant_id: {{participant_data.identifier}}"
```

vaibhavK1199: @snopoke  I think it worked , let me know if there are any other suggestions on the update

vaibhavK1199: there were conflicting changes in the main tried rebasing it but it got messier so will create a new branch if the conflicts persist.

Issue #1181, opened 2025-02-12
Expand participant data available to the Python Node


snopoke:
The [Code Node](https://github.com/dimagi/open-chat-studio/blob/e705ca6aef5b8a1c8872853d1407ed08e0be17f4/apps/pipelines/nodes/nodes.py#L760) gives access to participant data but it is slightly different from the participant data available via the `{participant_data}` prompt variable:

* does not include the [global data](https://github.com/dimagi/open-chat-studio/blob/f075dd4e29661b077e6441e4e35e4eb0abdd3986/apps/experiments/models.py#L1296)
* does not include [schedules](https://github.com/dimagi/open-chat-studio/blob/f075dd4e29661b077e6441e4e35e4eb0abdd3986/apps/experiments/models.py#L1724-L1726)

### Include global data

Update the `get_participant_data` function to include 'global data'. When updating participant data also update the 'global data' if necessary.

### Expose schedules

Add a new function `get_participant_schedules` which returns the schedule data. Support for creating / updating schedules is considered out of scope.

stephherbers:
@snopoke created [draft pr](https://github.com/dimagi/open-chat-studio/pull/1271) for the include global data. Curious how you think is the best way going about testing this?

stephherbers:
Prioritzed the filter widget work and ad hoc bug fixes over this ticket, but final part 3 pr is in review: https://github.com/dimagi/open-chat-studio/pull/1329

PR #1271, opened 2025-03-03
Add participant global data to getter/setter for python node

stephherbers:
## Description

part one of ticket [here](https://github.com/dimagi/open-chat-studio/issues/1181)

## User Impact

includes global_data like the`{participant_data}` prompt variable

### Demo

n/a

### Docs and Changelog

change log yes (link to be added)

### DIFF
--- a/apps/pipelines/nodes/helpers.py
+++ b/apps/pipelines/nodes/helpers.py
@@ -55,14 +55,9 @@ class ParticipantDataProxy:
             )
         return self._participant_data
 
-    def get(self, include_global=True):
+    def get(self):
         data = self._get_db_object().data
-        if include_global and hasattr(self.session, "participant") and hasattr(self.session.participant, "global_data"):
-            # Only add global data keys that don't already exist in participant data
-            for key, value in self.session.participant.global_data.items():
-                if key not in data:
-                    data[key] = value
-        return data
+        return self.session.participant.global_data | data
 
     def set(self, data):
         participant_data = self._get_db_object()
--- a/apps/pipelines/nodes/helpers.py
+++ b/apps/pipelines/nodes/helpers.py
@@ -64,4 +64,4 @@ class ParticipantDataProxy:
         participant_data.data = data
         participant_data.save(update_fields=["data"])
 
-        self.session.participant.update_from_data(data)
+        self.session.participant.update_name_from_data(data)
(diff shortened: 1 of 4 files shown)


apps/pipelines/nodes/helpers.py
```diff
@@ -55,10 +55,26 @@ def _get_db_object(self):
             )
         return self._participant_data
 
-    def get(self):
-        return self._get_db_object().data
+    def get(self, include_global=True):
```
snopoke, line 58: This doesn't need to be optional - we can always include it

apps/pipelines/nodes/helpers.py
```diff
@@ -55,10 +55,26 @@ def _get_db_object(self):
...
+        data = self._get_db_object().data
+        if include_global and hasattr(self.session, "participant") and hasattr(self.session.participant, "global_data"):
```
snopoke, line 60: I don't think these checks are necessary, a session will always have a participant and a participant will always have the 'global_data' property

apps/pipelines/nodes/helpers.py
```diff
@@ -55,10 +55,26 @@ def _get_db_object(self):
...
+            # Only add global data keys that don't already exist in participant data
+            for key, value in self.session.participant.global_data.items():
+                if key not in data:
+                    data[key] = value
```
snopoke, line 64: You can do the same as what we do in `ExperimentSession.get_participant_data`:

```python
return participant.global_data | participant_data
```

apps/pipelines/nodes/helpers.py
```diff
@@ -55,10 +55,26 @@ def _get_db_object(self):
...
+        if hasattr(self.session, "participant") and hasattr(self.session.participant, "global_data"):
+            participant = self.session.participant
+            global_data_keys = list(participant.global_data.keys())
+            updated_fields = []
+            for key in global_data_keys:
+                if key in data and hasattr(participant, key):
+                    setattr(participant, key, data[key])
+                    updated_fields.append(key)
+            if updated_fields:
+                participant.save(update_fields=updated_fields)
```
snopoke, line 80: A nice option would be to put this on the participant object directly. Something like:

```python
participant.update_from_data(data)
```

That would encapsulate the 'global data' access and update within the participant itself. You could also use that method here: https://github.com/dimagi/open-chat-studio/blob/c3920a1639704d741fcf79f17ed14d4d57ea7ce4/apps/api/views.py#L154-L157

snopoke (commented):
This can be tested using standard unit tests. I don't think anything fancy is required

stephherbers:
@snopoke took you comments and also just decided to modify to handle the current "name" in global data rather than the general case since I figure not there are no plans of that to change? It's easy enough to switch in the furture and feels clearer to me, but happy to switch back

snopoke (approved):

PR #1272, opened 2025-03-04
history compression modes

### LINKED ISSUE #1059: More options for history compression in pipelines
Give pipeline authors options for how history is managed. In the UI this should be presented as a dropdown + additional fields under the advanced editor. On the node itself, the configured mode should be displayed in small text below the history dropdown.

### Summarize
If the token count is > X tokens, summarize the history but keep the last N messages. 

#### Config options:
- X: Token Limit
  - help text: When the token count of the history + prompt + user input exceeds this number, the history will be compressed by summarizing the oldest messages.

### Truncate Tokens
If the token count is > X tokens, truncate the history until it is < X tokens. This works by removing older messages until the token count meets the requirement.

#### Config options:
- X: Token Limit
  - help text: When the token count of the history + prompt + user input exceeds this number, the history will be truncated by removing the oldest messages until the token count is below the limit.

### Max history length
Always truncate the history to the last N messages.

#### Config options:
- N: Max History Length
  - help text: The chat history that is sent to the LLM will be limited to this number of messages. Older messages will not be sent.

barora05:
## Description
Add modes for chat history compression. Closes #1059 

### Demo
[Demo](https://drive.google.com/file/d/1l-C_PLcmsXr5k5fVDwXLCiHDDz_OViXf/view?usp=drive_link)

Updated demo image after review changes
image


assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -845,9 +849,40 @@ export function HistoryTypeWidget(props: WidgetParams) {
...
+          name="history_mode"
+          onChange={(e) => {
+            setHistoryMode(e.target.value);
+            props.updateParamValue(e);
+          }}
+          value={historyMode}
+        >
+          <option value="Summarize">Summarize</option>
+          <option value="Truncate Tokens">Truncate Tokens</option>
+          <option value="Max History Length">Max History Length</option>
```
snopoke, line 866: options should come from the schema:
```
const options = getSelectOptions(props.nodeSchema.properties['history_mode']);
...
{options.map((option) => (
  <option key={option.value} value={option.value}>
    {option.label}
  </option>
))}
```

assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -845,9 +849,40 @@ export function HistoryTypeWidget(props: WidgetParams) {
...
+        </select>
+      </InputField>
+    </div>
+
+    {(historyMode === "Summarize" || historyMode === "Truncate Tokens") && (
+      <div className="flex join">
+        <InputField label="Max Token Size" help_text="">
```
snopoke, line 873: nit: update label
```suggestion
        <InputField label="Token Limit" help_text="">
```

We should also find a way to add some help text for this and the mode select:

> When this limit is reached, the history will be compressed using the method selected.

apps/pipelines/models.py
```diff
@@ -504,6 +504,12 @@ class PipelineChatHistoryTypes(models.TextChoices):
     NONE = "none", "No History"
 
 
+class PipelineChatHistoryModes(models.TextChoices):
+    SUMMARIZE = "summarize", "Summarize"
+    TRUNCATE_TOKENS = "truncate tokens", "Truncate Tokens"
+    MAX_HISTORY_LENGTH = "max history length", "Max History Length"
```
snopoke, line 510: nit: values should be slugs
```suggestion
    SUMMARIZE = "summarize", "Summarize"
    TRUNCATE_TOKENS = "truncate_tokens", "Truncate Tokens"
    MAX_HISTORY_LENGTH = "max_history_length", "Max History Length"
```

apps/chat/conversation.py
```diff
@@ -192,69 +201,71 @@ def _compress_chat_history(
...
 
-def compress_chat_history_from_messages(
-    llm, history, keep_history_len: int, max_token_limit: int, input_messages: list
-):
-    summary = history.pop(0).content if history[0].type == ChatMessageType.SYSTEM else None
-    history, pruned_memory = history[-keep_history_len:], history[:-keep_history_len]
-    # The latest message
-    latest_message = history[-1]
 
+def summarize_history(llm, history, max_token_limit, input_message_tokens, summary):
```
snopoke, line 219: seems like this implementation is missing some of the features of the existing implementation around the max messages length (_messages_exceeds_limit)

apps/chat/conversation.py
```diff
@@ -192,69 +201,71 @@ def _compress_chat_history(
...
+        history_tokens = llm.get_num_tokens_from_messages(history)
+    summary = _get_new_summary(llm, pruned_memory, summary, max_token_limit)
+    return history, pruned_memory, summary
 
-    while (
-        not first_pass_done
-        # Check `_tokens_exceeds_limit` here, since the generated summary can cause a token limit breach
-        or _tokens_exceeds_limit(
-            history, token_count=(history_tokens + summary_tokens + input_message_tokens), limit=max_token_limit
+def compress_chat_history_from_messages(
```
snopoke, line 233: this should be a noop if no compression is required

apps/chat/conversation.py
```diff
@@ -192,69 +201,71 @@ def _compress_chat_history(
...
+    Handles chat history compression based on selected mode:
+    - "Summarize": Summarizes older messages when exceeding token limit.
+    - "Truncate Tokens": Deletes old messages when exceeding token limit.
+    - "Max History Length": Always keeps the last `keep_history_len` messages.
+    """
+    summary = history.pop(0).content if history and history[0].type == ChatMessageType.SYSTEM else None
+    history, pruned_memory = history[-keep_history_len:], history[:-keep_history_len]
+    latest_message = history[-1] if history else None
+    input_message_tokens = llm.get_num_tokens_from_messages(input_messages)
+    if history_mode == "Max History Length":
```
snopoke, line 246: With some of the other suggestions you should be able to use the `PipelineChatHistoryModes` enum:

```suggestion
    if history_mode == PipelineChatHistoryModes.MAX_HISTORY_LENGTH:
```

snopoke (commented):
This is a good start.

assets/javascript/apps/pipeline/nodes/GetInputWidget.tsx
```diff
@@ -106,7 +106,7 @@ export const getNodeInputWidget = (param: InputWidgetParams) => {
   }
 
   const allowedInNode = nodeTypeToInputParamsMap[param.nodeType];
-  if (param.name == "name" || (allowedInNode && !allowedInNode.includes(param.name))) {
+  if (param.name == "name" || param.name == "history_mode" || (allowedInNode && !allowedInNode.includes(param.name))) {
```
snopoke, line 109: I don't think this change is necessary since `history_mode` doesn't appear in the `allowedInNode` mapping

assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -846,8 +849,56 @@ export function HistoryTypeWidget(props: WidgetParams) {
...
+
+export function HistoryModeWidget(props: WidgetParams) {
+  const options = getSelectOptions(props.schema);
+  const userMaxTokenLimit = concatenate(props.nodeParams["user_max_token_limit"]);
+  const initialHistoryMode = concatenate(props.nodeParams["history_mode"]);
+  const [historyMode, setHistoryMode] = useState(initialHistoryMode || "Summarize");
+   const historyModeHelpTexts: Record<string, string> = {
+    "Summarize": "If the token count exceeds the limit, older messages will be summarized while keeping the last few messages intact.",
+    "Truncate Tokens": "If the token count exceeds the limit, older messages will be removed until the token count is below the limit.",
+    "Max History Length": "The chat history will always be truncated to the last 10 messages. Older messages will not be sent."
```
snopoke, line 863: "last 10 messages": this number should be configurable. The ticket mentions a "Max History Length" parameter when using this mode which should be an integer.

apps/chat/conversation.py
```diff
@@ -224,37 +234,59 @@ def compress_chat_history_from_messages(
...
+    - "Summarize": Summarizes older messages when exceeding token limit.
+    - "Truncate Tokens": Deletes old messages when exceeding token limit.
+    - "Max History Length": Always keeps the last `keep_history_len` messages.
+    """
+    summary = history.pop(0).content if history and history[0].type == ChatMessageType.SYSTEM else None
+    history, pruned_memory = history[-keep_history_len:], history[:-keep_history_len]
+    latest_message = history[-1] if history else None
+    input_message_tokens = llm.get_num_tokens_from_messages(input_messages)
+    if history_mode == PipelineChatHistoryModes.MAX_HISTORY_LENGTH:
+        return history, latest_message, summary
```
snopoke, line 268: I think this will never happen because of the condition on line 193 (in `_compress_chat_history`): 

```python
if current_token_count <= max_token_limit and len(total_messages) <= MAX_UNCOMPRESSED_MESSAGES:
    log.info("Skipping chat history compression: %s <= %s", current_token_count, max_token_limit)
    return history, None, None
```

snopoke (commented):
This is looking good.

snopoke:
Still needs some more tests

apps/chat/conversation.py
barora05: Should I check the current_token_count <= max_token_limit for summarize and truncate_tokens mode and len(total_messages) <= max_history_length for max_history_length mode?

assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -846,8 +849,80 @@ export function HistoryTypeWidget(props: WidgetParams) {
...
+          >
+            {options.map((option) => (
+              <option key={option.value} value={option.value}>
+                {option.label}
+              </option>
+            ))}
+          </select>
+          <br />
+          <br />
+          <div>{historyModeHelpTexts[historyMode]}</div>
```
snopoke, line 889: this spacing should be done with a margin:
```suggestion
          <div className="mt-4">{historyModeHelpTexts[historyMode]}</div>
```

assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -846,8 +849,80 @@ export function HistoryTypeWidget(props: WidgetParams) {
...
+        </InputField>
+      </div>
+
+      {(historyMode === "summarize" || historyMode === "truncate_tokens") && (
+        <div className="flex join mb-4">
```
snopoke, line 894: I'm curious what this looks like. I see you're using the `join` class but there there is only on`join-item` in the div.

If the help text is being a problem we can find another way to display it, perhaps using the help dropdown

assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -846,8 +849,76 @@ export function HistoryTypeWidget(props: WidgetParams) {
...
-    ;
+  );
+}
+
+export function HistoryModeWidget(props: WidgetParams) {
+  const options = getSelectOptions(props.schema);
+  const userMaxTokenLimit = concatenate(props.nodeParams["user_max_token_limit"]);
+  const maxHistoryLength = concatenate(props.nodeParams["max_history_length"]);
+  const initialHistoryMode = concatenate(props.nodeParams["history_mode"]);
+  const [historyMode, setHistoryMode] = useState(initialHistoryMode || "Summarize");
```
snopoke, line 860: This should be the value, not the label
```suggestion
  const [historyMode, setHistoryMode] = useState(initialHistoryMode || "summarize");
```

assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -846,8 +849,76 @@ export function HistoryTypeWidget(props: WidgetParams) {
...
+            }}
+            value={historyMode}
+          >
+            {options.map((option) => (
+              <option key={option.value} value={option.value}>
+                {option.label}
+              </option>
+            ))}
+          </select>
+          <div>{historyModeHelpTexts[historyMode]}</div>
```
snopoke, line 887: Help text should use `<small className="text-muted">...</small>`

(same for help text below)

snopoke (commented):
Almost there

apps/chat/conversation.py
```diff
@@ -181,80 +190,104 @@ def _compress_chat_history(
...
+        if current_token_count <= max_token_limit and len(total_messages) <= MAX_UNCOMPRESSED_MESSAGES:
+            log.info("Skipping chat history compression: %s <= %s", current_token_count, max_token_limit)
+            return history, None, None
+    elif history_mode == PipelineChatHistoryModes.MAX_HISTORY_LENGTH:
+        if keep_history_len is not None and len(total_messages) <= keep_history_len:
+            log.info("Skipping chat history compression: %d <= %d", len(total_messages), keep_history_len)
+            return history, None, None
 
     log.debug(
         "Compressing chat history: %s/%s(max) tokens and %d/%d(max) messages",
```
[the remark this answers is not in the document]
barora05: done

assets/javascript/apps/pipeline/nodes/widgets.tsx
barora05: done

barora05:
@snopoke I have added the updated demo image after review changes

apps/pipelines/nodes/nodes.py
```diff
@@ -143,6 +143,22 @@ class HistoryMixin(LLMResponseMixin):
             widget=Widgets.none,
         ),
     )
+    history_mode: PipelineChatHistoryModes = Field(
```
SmittieC, line 146: Might be good to add a `title`. Something like "History management strategy".

apps/service_providers/llm_service/history_managers.py
```diff
@@ -153,6 +156,7 @@ def __init__(
         history_name: str | None = None,
         max_token_limit: int | None = None,
         chat_model: BaseChatModel | None = None,
+        history_mode: str | None = None,
```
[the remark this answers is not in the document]
snopoke: for existing data they can be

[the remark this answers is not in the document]
SmittieC: I guess my question is: Should it ever be `None`? It feels like the answer should be no, in which case we'll need to migrate existing nodes to have a default value, which would be the summarize one

apps/pipelines/nodes/nodes.py
snopoke: Thanks @SmittieC, good point about a default. There is a default set via the UI but it would make sense to have a default set here too.

apps/pipelines/nodes/nodes.py
```diff
@@ -143,6 +143,22 @@ class HistoryMixin(LLMResponseMixin):
...
+        None,
```
snopoke, line 147:
```suggestion
        PipelineChatHistoryModes.SUMMARIZE,
```

apps/service_providers/llm_service/history_managers.py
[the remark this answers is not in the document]
snopoke: The code handles the case where it is None so I'm ok with leaving the migration for a future PR.

snopoke (approved):

snopoke:
I've made a ticket with some followup tasks here: https://github.com/dimagi/open-chat-studio/issues/1337

PR #1279, opened 2025-03-04
UI tweak: add subtext "(upload files)"

stephherbers:
## Description

Requested in slack after noting that sometimes easy to forget to add files for both code interpreter and file search
Screenshot 2025-03-04 at 8 12 48 AM

## User Impact

adds a bit more clarity

### Demo

N/a

### Docs and Changelog

n/a


snopoke (approved):
could be worth showing the screen shot to the user to confirm this is what they meant. Otherwise LGTM

PR #1281, opened 2025-03-05
Format Specifier Fix For Prompts

vaibhavK1199:
Fixes - #1274


snopoke (commented):
I don't think this fixes the issue. Please write a test that demonstrates the bug and then you can fix it.

You can use a prompt like `{source_material:xyz}`

apps/utils/prompt.py
```diff
@@ -27,8 +28,16 @@ def validate_prompt_variables(form_data, prompt_key: str, known_vars: set):
...
+    if not prompt_text:
+        return set()
+
+    prompt_variables = set()
     try:
-        prompt_variables = {get_root_var(var) for var in PromptTemplate.from_template(prompt_text).input_variables}
+        for literal, field_name, format_spec, conversion in Formatter().parse(prompt_text):
+            if field_name is not None:
+                if format_spec:
+                    raise ValidationError({prompt_key: "Invalid format in prompt: format specifiers are not allowed."})
```
snopoke, line 39: Updated error message:
```suggestion
                    conversion = f"!{conversion}" if conversion else ""
                    format_spec = f":{format_spec}" if format_spec else ""
                    variable = f"{{{field_name}{conversion}{format_spec}}}"
                    bad_part = f"{conversion}{format_spec}"
                    raise ValidationError({prompt_key: f"Invalid prompt variable '{variable}'. Remove the '{bad_part}'."})
```

apps/utils/prompt.py
```diff
@@ -27,8 +28,16 @@ def validate_prompt_variables(form_data, prompt_key: str, known_vars: set):
...
     prompt_text = form_data.get(prompt_key, "")
+    if not prompt_text:
+        return set()
+
+    prompt_variables = set()
     try:
-        prompt_variables = {get_root_var(var) for var in PromptTemplate.from_template(prompt_text).input_variables}
+        for literal, field_name, format_spec, conversion in Formatter().parse(prompt_text):
+            if field_name is not None:
+                if format_spec:
```
snopoke, line 38: Just thinking that we should also disallow conversions:
```suggestion
                if format_spec or conversion:
```

Here's a test case for conversions: `"{var!s}"`, and one for both: `"{var!r:xyz}"`

Issue #1183, opened 2025-02-12
Update 'RenderTemplate' node context


snopoke:
The [RenderTemplate](https://github.com/dimagi/open-chat-studio/blob/e705ca6aef5b8a1c8872853d1407ed08e0be17f4/apps/pipelines/nodes/nodes.py#L56) node can be used to generate a formatting text which can be used by other nodes such as the 'SendEmail' node.

Currently the context provided to the Jinja template engine in the template node is only the node input. 

To make the template more flexible it should be possible to configure the input context to allow access to:

* participant details (identifier, platform)
* participant data
* participant schedules
* pipeline temp state

This could be done using a Code Node before the template node and having the code node return the context, however, it would be simpler if the template node could access this data without an additional node.

Providing this context could be done by reusing the `ParticipantDataProxy` object which is used by the Code Node.

See also #1181

### Full template context

| Key | Description | Type |
| -- | -- | -- |
| input | The input to the node | string |
| participant_data | Participant data for this session | dict |
| participant_details | Dict containing `identifier` and `platform` keys |
| participant_schedules | Participant schedule data | dict |
| temp_state | Pipeline temporary state | dict |

PR #1283, opened 2025-03-06
Updating Render Template

vaibhavK1199:
Updating Render Template Context 
Fixes - #1183

### DIFF
--- a/apps/pipelines/tests/test_code_node.py
+++ b/apps/pipelines/tests/test_code_node.py
@@ -381,6 +381,9 @@ def main(input, **kwargs):
 @django_db_with_data(available_apps=("apps.service_providers",))
 @mock.patch("apps.pipelines.nodes.base.PipelineNode.logger", mock.Mock())
 def test_render_template_with_context_keys(pipeline, experiment_session):
+    participant_data = {
+        "identifier": "participant_123",
+    }
     nodes = [
         start_node(),
         render_template_node(
@@ -393,7 +396,7 @@ def test_render_template_with_context_keys(pipeline, experiment_session):
         experiment_session=experiment_session,
         messages=["Cycling"],
         temp_state={"my_key": "example_key"},
-        participant_details={"identifier": "participant_123"},
+        participant_data=participant_data,
     )
     result = create_runnable(pipeline, nodes).invoke(state)
     assert result["messages"][-1] == ("input: Cycling, temp_state.my_key: example_key, participant_id: participant_123")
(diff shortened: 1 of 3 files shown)


apps/pipelines/tests/test_code_node.py
```diff
@@ -376,3 +376,27 @@ def main(input, **kwargs):
...
+            "input: {{input}}, temp_state.my_key: {{temp_state.my_key}}, "
+            "participant_id: {{participant_data.identifier}}"
+        ),
+        end_node(),
+    ]
+    state = PipelineState(
+        experiment_session=experiment_session,
+        messages=["Cycling"],
+        temp_state={"my_key": "example_key"},
+        context={"participant_data": participant_data},
```
snopoke, line 399: This won't do anything. `PipelineState` has no `context` key.

To set participant data you'll need to update it in the DB unless you mock it on the session:
```python
pd = sessionexperiment.participantdata_set.get(participant=session.participant)
pd.data = {...}

# OR

session.participant_data_from_experiment = lambda: *args, **kwargs: {...}
```

apps/pipelines/tests/test_code_node.py
```diff
@@ -376,3 +376,27 @@ def main(input, **kwargs):
...
+@mock.patch("apps.pipelines.nodes.base.PipelineNode.logger", mock.Mock())
+def test_render_template_with_context_keys(pipeline, experiment_session):
+    participant_data = {
+        "identifier": "participant_123",
+    }
+    nodes = [
+        start_node(),
+        render_template_node(
+            "input: {{input}}, temp_state.my_key: {{temp_state.my_key}}, "
+            "participant_id: {{participant_data.identifier}}"
```
snopoke, line 391: I also think I misled you on the previous comment, I see this should actually be `participant_details`. Apologies for that.

snopoke (commented):
Are you able to run these tests locally?

Try this:
```
pytest apps/pipelines/tests/test_code_node.py::test_render_template_with_context_keys
```

apps/pipelines/tests/test_code_node.py
vaibhavK1199: I am actually not able to run the tests locally thats why i am pushing commits to test them, its not importing factory properly so it gives me "TypeError: 'class Meta' for <class 'apps.utils.factories.user.UserFactory'> got unknown attribute(s) skip_postgeneration_save" .

snopoke: That sounds like a library version issue - I suggest you run `uv pip install -r dev-requirements.txt` or else re-create your virtual env

snopoke (approved):

PR #1287, opened 2025-03-06
Allow users to cancel reminders from the participant data page

snopoke:
## Description
Allow cancelling scheduled messages via the UI.

This also changes the tools and API to cancel scheduled messages instead of deleting them.

## User Impact
Cancelled schedules can be seen on the participant data page and users can cancel scheduled messages for participants.

### Demo
[Screencast from 2025-03-06 18-48-46.webm](https://github.com/user-attachments/assets/cb4e6937-974f-4860-bbd4-9db3afb53a2e)

### Docs and Changelog
TODO


apps/events/models.py
```diff
@@ -460,7 +484,8 @@ def expected_trigger_count(self):
 
     @property
     def remaining_triggers(self):
-        return self.expected_trigger_count - self.total_triggers
+        remaining = self.expected_trigger_count - self.total_triggers
+        return max(remaining, 0)
```
stephherbers, line 488: nice

stephherbers (approved):

PR #1294, opened 2025-03-07
connect channel consent tweaks

snopoke:
Sentry: OPEN-CHAT-STUDIO-NY

## Description
Some small tweaks to where consent is checked. The main goal is to check consent earlier in the process and to give feedback in the API views if consent has not been given.

## User Impact
Better API feedback. Also using less resources for participants that haven't consented.

### Demo
NA

### Docs and Changelog
NA

### DIFF
--- a/apps/chat/channels.py
+++ b/apps/chat/channels.py
@@ -196,7 +196,7 @@ class ChannelBase(ABC):
         pass
 
     @staticmethod
-    def get_channel_class_for_platform(platform: ChannelPlatform) -> "ChannelBase":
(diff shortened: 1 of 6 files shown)


apps/chat/channels.py
```diff
@@ -266,7 +266,8 @@ def _new_user_message(self, message) -> str:
         try:
             self._add_message(message)
         except ParticipantNotAllowedException:
-            return self.send_message_to_user("Sorry, you are not allowed to chat to this bot")
+            self.send_message_to_user("Sorry, you are not allowed to chat to this bot")
+            return ""
```
SmittieC, line 270: :+1:

SmittieC (approved):

PR #1300, opened 2025-03-10
Add getter function for schedules to python node

stephherbers:
## Description

Part 2 of [this ticket](https://github.com/dimagi/open-chat-studio/issues/1181) to add access to participant's schedules to the python nodes-- only active. 

## User Impact

Can be accessed in the python node

### Demo

N/A

### Docs and Changelog

yes to both

### DIFF
--- a/apps/pipelines/nodes/helpers.py
+++ b/apps/pipelines/nodes/helpers.py
@@ -72,13 +72,13 @@ class ParticipantDataProxy:
         """
         from apps.events.models import ScheduledMessage
 
-        experiment = self.session.experiment
-        participant = self.session.participant
+        experiment = self.session.experiment_id
+        participant = self.session.participant_id
         team = self.session.experiment.team
         messages = (
             ScheduledMessage.objects.filter(
-                experiment=experiment,
-                participant=participant,
+                experiment_id=experiment,
+                participant_id=participant,
                 team=team,
                 is_complete=False,
                 cancelled_at=None,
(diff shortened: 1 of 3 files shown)


apps/pipelines/nodes/helpers.py
```diff
@@ -65,3 +65,28 @@ def set(self, data):
...
+                team=team,
+                is_complete=False,
+                cancelled_at=None,
+            )
+            .select_related("action")
+            .order_by("created_at")
+        )
+        scheduled_messages = []
+        for message in messages:
+            scheduled_messages.append(message.as_string())
```
snopoke, line 91: I think this should use `as_dict` so that the schedule can be interrogated by the python code.

apps/pipelines/nodes/helpers.py
```diff
@@ -65,3 +65,28 @@ def set(self, data):
...
+        Returns all active scheduled messages for the participant in the current experiment session.
+        """
+        from apps.events.models import ScheduledMessage
+
+        experiment = self.session.experiment
+        participant = self.session.participant
+        team = self.session.experiment.team
+        messages = (
+            ScheduledMessage.objects.filter(
+                experiment=experiment,
```
SmittieC, line 80: Nit: To save a few milliseconds here, you can use `self.session.experiment_id` and `self.session.participant_id` instead. That way we don't have to fire two queries to load the instances

snopoke (approved):
This needs a docs update too.

PR #1301, opened 2025-03-11
Bump langfuse version

SmittieC:
## Description

Fix to this issue that seemed to kill the celery worker, since it ran in what looks like an infinite loop:
image

I also got this issue locally, but after bumping the version, it disappeared and the bot was able to respond.

## User Impact

Users should see bots using the langfuse tracing provider responding again.

### Demo

N/A

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/64

### DIFF
--- a/requirements/dev-requirements.txt
+++ b/requirements/dev-requirements.txt
@@ -66,6 +66,8 @@ packaging==23.2
     #   -c requirements.txt
     #   build
     #   pytest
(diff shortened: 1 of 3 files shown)


SmittieC:
Merging, since this a fix

snopoke (commented):
:+1:

PR #1310, opened 2025-03-11
Send bot message when schedule is triggered

stephherbers:
## Description

Flagged in slack

Theory: In ScheduledMessage._trigger(), it was using experiment_session.ad_hoc_bot_message() to send the reminder, but that, takes the message as an instruction_prompt and passes it to _bot_prompt_for_user(). So the bot is treating the reminder text (e.g., "Take your medication") as an instruction to generate a reminder-related response, which results in sending a confirmation message about scheduling a reminder.

Fix: directly call try_send_message() which sends the message straight to the user

## User Impact

fixes schedules

### Docs and Changelog

yes to change log

### DIFF
--- a/apps/events/models.py
+++ b/apps/events/models.py
@@ -415,11 +415,7 @@ class ScheduledMessage(BaseTeamModel):
             return
 
         with current_team(experiment_session.team):
-            experiment_session.ad_hoc_bot_message(
-                self.params["prompt_text"],
-                fail_silently=False,
-                use_experiment=self._get_experiment_to_generate_response(),
-            )
+            experiment_session.try_send_message(self.params["prompt_text"], fail_silently=False)
 
         utc_now = timezone.now()
         self.last_triggered_at = utc_now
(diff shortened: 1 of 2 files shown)


stephherbers:
the remaining tests that are failing are failing on the main branch as well

apps/events/models.py
```diff
@@ -415,11 +415,7 @@ def _trigger(self):
             return
 
         with current_team(experiment_session.team):
-            experiment_session.ad_hoc_bot_message(
-                self.params["prompt_text"],
-                fail_silently=False,
-                use_experiment=self._get_experiment_to_generate_response(),
-            )
+            experiment_session.try_send_message(self.params["prompt_text"], fail_silently=False)
```
SmittieC, line 418: Hmm, I don't think this will quite work. Since `self.params["prompt_text"]` is a prompt, we should be calling `ad_hoc_bot_message` to treat that as an instruction, otherwise it would be awkward for the user to receive a message saying "tell the user to make take a break" for instance.

Also, when we call `_bot_prompt_for_user`, the generated message is being saved whereas `try_send_message` only tries to send the message.

snopoke:
I don't think this is the correct solution. I've documented my thoughts here: https://github.com/dimagi/open-chat-studio/issues/1195#issuecomment-2674673226

Issue #1296, opened 2025-03-07
Allow chat users to download assistant files


snopoke:
Some bots that are using OpenAI assistants need to allow the participants to download the assistants files when the assistant references them via a citation.

### Data model and UI
Add a field to the `apps.assistants.models.ToolResources` class to store this preference and expose it on the UI under the [tool resource section](https://github.com/dimagi/open-chat-studio/blob/2b45c04f0a8fbe163bbc37dfb22ab562a452016f/templates/assistants/assistant_form.html#L63-L69)

### Render citations as links

Update the assistant runnable to render links if the configuration option is enabled. See https://github.com/dimagi/open-chat-studio/blob/422ee2ab9a748eca32d0c975a550c47e2ae9179b/apps/service_providers/llm_service/runnables.py#L374-L376

PR #1314, opened 2025-03-12
Allowing Chat Users to Download Files

vaibhavK1199:
Adding new feature to allow chat users to download assitant files via citation.
Fixes https://github.com/dimagi/open-chat-studio/issues/1296


SmittieC (commented):
Hmm, I think having the config on the `OpenAiAssistant` model instead would make more sense and be easier to manage. You can do the same here as what we did with the `allow_file_search_attachments` and `allow_code_interpreter_attachments` fields to allow file uploads (see [this PR](https://github.com/dimagi/open-chat-studio/pull/947)), but in this case it would just be for downloading files/attachments.

This should simplify the form code quite a bit as well. See [here](https://github.com/dimagi/open-chat-studio/blob/a27025d4da7e027e4f67a4e0094dabd8fa936993/apps/assistants/forms.py#L38-L44). cc @snopoke let me know if you agree/disagree

apps/service_providers/llm_service/runnables.py
```diff
@@ -378,8 +382,10 @@ def _get_output_with_annotations(self, thread_id, run_id) -> tuple[str, list[str
                             # Original citation text example:【6:0†source】
                             if self.adapter.citations_enabled:
                                 message_content_value = message_content_value.replace(file_ref_text, f" [{idx}]")
-                                if file_link:
-                                    message_content_value += f"\n[{idx}]: {file_link}"
+                                if file_link and allow_downloads:
+                                    message_content_value += f"\n[{idx}]: <a href='{file_link}'>{file_name}</a>"
```
SmittieC, line 386: A valid download link should be generated without this change. Is that not happening?

apps/service_providers/llm_service/runnables.py
```diff
@@ -378,8 +382,10 @@ def _get_output_with_annotations(self, thread_id, run_id) -> tuple[str, list[str
...
+                                elif file_link:
+                                    message_content_value += f"\n[{idx}]: {file_name} (download not permitted)"
```
SmittieC, line 388: From a UX perspective, I'm thinking we don't have to tell users that they cannot download the file. Simply not having a download link should be enough indication? Happy to be wrong though :smile:

apps/service_providers/llm_service/runnables.py
```diff
@@ -365,7 +366,7 @@ def _get_output_with_annotations(self, thread_id, run_id) -> tuple[str, list[str
                     message_content_value = message_content.text.value
 
                     annotations = message_content.text.annotations
-                    for idx, annotation in enumerate(annotations):
+                    for idx, annotation in enumerate(annotations, 1):
```
snopoke, line 369: why this change?

vaibhavK1199: Yes i agree this was unecessary,have removed this change

apps/service_providers/llm_service/runnables.py
vaibhavK1199: with the original implementation it didnt show valid url that would seems like a download link to the user, tried that you can see it in the screenshot attached below 
image

apps/service_providers/llm_service/runnables.py
vaibhavK1199: Yep i agree, have removed the message.

apps/service_providers/llm_service/runnables.py
```diff
@@ -441,35 +442,35 @@ def _create_image_file_from_image_message(self, client, image_file_message) -> F
...
-        `forbidden_file_ids`, the link will be empty to prevent unauthorized access.
-        """
-        file_link = ""
-
+        """Returns a file name and a download URL for `file_id`."""
         team = self.adapter.session.team
-        session_id = self.adapter.session.id
         try:
             file = File.objects.get(external_id=file_id, team_id=team.id)
-            file_link = f"file:{team.slug}:{session_id}:{file.id}"
```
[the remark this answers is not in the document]
vaibhavK1199: Actually to avoid the need to integrate with FileLinkInlineProcessor i have used this implementation to directly return the download url, let me know if you still want to revert this usage.

apps/service_providers/llm_service/runnables.py
```diff
@@ -441,35 +442,35 @@ def _create_image_file_from_image_message(self, client, image_file_message) -> F
...
                 openai_file = client.files.retrieve(file_id=file_id)
                 file_name = openai_file.filename
+                return file_name, None
             except Exception as e:
                 logger.error(f"Failed to retrieve file {file_id} from OpenAI: {e}")
                 file_name = "Unknown File"
-
-        if file_id in forbidden_file_ids:
-            # Don't allow downloading assistant level files
-            return file_name, None
```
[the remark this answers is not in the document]
vaibhavK1199: Agreed! adding back the forbiden files check

PR #1321, opened 2025-03-13
Exp Sessions Filter Widget: Export sessions compatibility 

### LINKED ISSUE #1320: Generate Export on experiment sessions compatibility with filter widget
Ensure the filters applied via the new widget are reflected in the export file

stephherbers:
## Description

Resolves #1320
Branches off of https://github.com/dimagi/open-chat-studio/pull/1260

## User Impact

when filters are applied, it's also reflected in the data export

### Docs and Changelog

Yes to both


templates/experiments/components/exports.html
```diff
@@ -1,30 +1,39 @@
...
-            hx-target="#chat-exports"
-        >
+                hx-include="[name='show-all']"
+                hx-vals='js:{
+                         {% for param, value in request.GET.items %}
+                             {% if param|matches:"^filter_" %}
+                                 "{{ param }}": "{{ value|escapejs }}",
+                             {% endif %}
+                         {% endfor %}
+                         }'
```
snopoke, line 14: This will get generated at load time right? And since the button doesn't get reloaded when the filters update I don't think this is going to work correctly. If you add a filter and then export the data without refreshing the page does the filter get applied to the export?

If I'm right, maybe you could use the `HX-Current-URL` header to get the current filter set? (I haven't tested that)

stephherbers: I tried `HX-Current-URL` and it gave me just the POST, but I tried this https://github.com/dimagi/open-chat-studio/pull/1321/commits/70c1ebad9a0562e9c058a4223e0a65ce8ed9e9ee

stephherbers: I take that back: https://github.com/dimagi/open-chat-studio/pull/1321/commits/da757fbdc406f99bd66912d3e8a5a4d1427d1dab

apps/experiments/tasks.py
```diff
@@ -24,18 +23,21 @@
...
-    uploaded_file = ContentFile(content=csv_in_memory.getvalue().encode("utf-8"), name="chat_export.csv")
-    file = File.objects.create(
-        name=uploaded_file.name,
-        content_type="text/csv",
-        file=uploaded_file,
+    filter_params = kwargs.get("filter_params", {})
+    show_all = kwargs.get("show_all", False)
+    filtered_sessions = get_filtered_sessions(experiment, filter_params, show_all)
+    session_ids = list(filtered_sessions.values_list("id", flat=True))
+    csv_in_memory = filtered_export_to_csv(experiment, session_ids)
```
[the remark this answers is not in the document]
stephherbers: I didn't want them to feel lonely (resolved [bb82ecc](https://github.com/dimagi/open-chat-studio/pull/1321/commits/bb82ecc2fd94f65a2f750f78b078d117f353a6ba))

templates/experiments/components/exports.html
```diff
@@ -1,30 +1,29 @@
...
+            <span class="loading loading-bars loading-xs"></span> Generating
+        {% else %}
+            Generate Chat Export
+        {% endif %}
+    </button>
+
+    {% if task_id %}
+        <div
+            hx-get="{% url 'experiments:get_export_download_link' team.slug experiment.id task_id %}"
+            hx-headers='{"HX-Current-URL": true}'
```
snopoke, line 18: nit: unnecessary
```suggestion
```

PR #1322, opened 2025-03-13
langfuse fix

snopoke:
## Description
This is a quick fix for the langfuse tracing issue. I have not yet determined why this started happening and I also have not been able to consistently reproduce the issue however I'm alsmost certain that the error described on https://github.com/dimagi/open-chat-studio/pull/1301 is the culprit. The upgrade did not fix that error, I think it just masked it.

With the changes in this PR the celery process memory does not grow continuously. 

I think the issue is that a span / trace gets into the langfuse queue to be sent but then fails during serialization which triggers an infinite retry. The serialization fails when trying to serialize Django model objects.

The fix in this PR is to pre-serialize the data to ensure that Django model's are converted to strings using `str(model)`. This is safe for all models because the base Model class implements `__str__`.

### Longer term fix
In the long term I think we should move the 'session' out of the `PipelineState` object and try to keep that object to simple types.

## User Impact
User's can use langfuse again.

### Demo
NA

### Docs and Changelog
TODO

### DIFF
--- a/apps/pipelines/management/commands/import_pipeline.py
+++ b/apps/pipelines/management/commands/import_pipeline.py
@@ -1,0 +1,43 @@
+import json
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+
+from apps.pipelines.models import Pipeline
+from apps.teams.models import Team
+
+
+class Command(BaseCommand):
+    help = "Create a new Pipeline and related models from a JSON file for a specified team."
+
+    def add_arguments(self, parser):
+        parser.add_argument("team", type=str, help="Slug of the team.")
+        parser.add_argument("name", type=str, help="Name of the new pipeline.")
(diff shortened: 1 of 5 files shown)


SmittieC (approved):
Nice, I think you are right, the upgrade might have just masked the error. Thanks for looking into this

apps/pipelines/management/commands/import_pipeline.py
```diff
@@ -0,0 +1,43 @@
+import json
```
SmittieC, line 1: It wouldn't be hard to take this code and create, say 1 or 2 default pipelines for each team where pipelines is being enabled. Those can serve as examples.

snopoke: I just made this because it was the quickest way to import one of the pipelines from prod. But I can imagine an import / export / copy feature being useful.

PR #1323, opened 2025-03-14
Multimedia | 3 | Ability to select collection on node

SmittieC:
## Description

Builds on https://github.com/dimagi/open-chat-studio/pull/1317 and resolves #1220 

## User Impact

Users will be able to select which collection to attach to their node(s)

### Demo

image

### Docs and Changelog

N/A

PR #1324, opened 2025-03-14
Multimedia | 4 | Show attachments for AI messages

### LINKED ISSUE #1221: [Multimedia support] Search and attachments tool | phase  1

### Requirements
- Build two [tools](https://github.com/dimagi/open-chat-studio/blob/a90f847a0b105c09c0d64ddc20e2e94271872afb/apps/chat/agent/tools.py) (Consult [this section](https://docs.google.com/document/d/1rXIbpHf-coprjIo-qsCS0A8PjsVEvoEP7Y2F0K7P4kI/edit?tab=t.nnf8cwfrza7y#heading=h.geihlyr7f09e) in the spec):
1. SearchMediaFiles: This must return the set of file summaries in the collection.
2. CreateAttachment: This must take a file id as input and add the file id to the message's metadata. Use a callback specified in the history manager to achieve this.

The tools should be added to the set of node tools when the node has a collection attached to it. The user must not be able to manually add these tools to the bot, since they depend on a collection. The `SearchMediaFiles` tool should be instantiated with the collection so that it can easily find the set of file summaries that it should return. In phase 1, it should always return the summaries of all files in the collection. In phase 2, this tool will be updated to use more sophisticated search methods.

- The callback from the history manager should append the file id to the [output_message_metadata](https://github.com/dimagi/open-chat-studio/blob/5e262bbb4ce9c1a4b4917a3f909e5e84ed587c9a/apps/service_providers/llm_service/history_managers.py#L165)
- Ensure that the pipeline state’s `message_metadata` key is additive so that other nodes do not override this.
- Refactor: Split the input message metadata from output message metadata

SmittieC:
## Description

Builds on https://github.com/dimagi/open-chat-studio/pull/1323 and resolves #1221 

## User Impact

Users will be able to see and download attachments

### Demo

Progress so far
https://www.loom.com/share/a5205b14891d4be188e9e9a7d1c8f4de

### Docs and Changelog

N/A


SmittieC:

apps/chat/agent/tools.py
```diff
@@ -171,6 +174,30 @@ def action(self, key: str, value: Any):
         return "Success"
 
 
+class AttachMediaTool(CustomBaseTool):
+    name: str = AgentTools.ATTACH_MEDIA
+    description: str = "Attach media to the chat"
```
snopoke, line 179: maybe?

```suggestion
    description: str = "Attach a media file to your response"
```

apps/chat/models.py
```diff
@@ -179,14 +179,23 @@ def get_attached_files(self):
...
-        if file_ids := self.metadata.get("openai_file_ids", []):
-            # We should not show files that are on the assistant level. Users should only be able to download
-            # those on the thread (chat) level, since they uploaded them
-            return [file for file in self.chat.get_attached_files() if file.external_id in file_ids]
-        return []
+        files = []
+        allowed_file_ids = []
+
+        if external_file_ids := self.metadata.get("openai_file_ids", []):
+            allowed_file_ids.extend(File.objects.filter(external_id__in=external_file_ids).values_list("id", flat=True))
```
snopoke, line 190: you could skip this query if you pass the external IDs to `get_attached_files`:

```python
def get_attached_files(self, file_ids: list[int] = None, external_ids: list[str] = None):

File: apps/service_providers/llm_service/history_managers.py [hunk 3]
```diff
@@ -163,8 +164,12 @@ def __init__(
         self.trace_service = session.experiment.trace_service if session else None
         self.ai_message = None
 
-        self.input_message_metadata = None
-        self.output_message_metadata = None
+        self.input_message_metadata = defaultdict(list)
+        self.output_message_metadata = defaultdict(list)
```
snopoke, line 168: not sure about using `defaultdict` here since different metadata keys might want different types

apps/service_providers/llm_service/history_managers.py
```diff
@@ -223,7 +228,7 @@ def add_messages_to_history(
         self, input: str, input_message_metadata: dict, output: str, output_message_metadata: dict, *args, **kwargs
     ):
         self.input_message_metadata = input_message_metadata
-        self.output_message_metadata = output_message_metadata
+        self.output_message_metadata = self.output_message_metadata | output_message_metadata
```
snopoke, line 231: This won't merge the lists. It will just overwrite so if 2 nodes try to attach files, only the last node will succeed

apps/pipelines/nodes/base.py
```diff
@@ -45,6 +45,10 @@ def add_temp_state_messages(left: dict, right: dict):
     return output
 
 
+def merge_dicts(left: dict, right: dict):
+    return {**left, **right}
```
snopoke, line 49: This is fine for dicts with simple values but if the value is a list, set or dict it won't merge them. I think you need a recursive merge strategy here.

apps/service_providers/llm_service/prompt_context.py
```diff
@@ -38,6 +40,23 @@ def get_source_material(self):
         except SourceMaterial.DoesNotExist:
             return ""
 
+    def get_media_summaries(self):
+        """
+        Example output:
+        File (27): summary1
+        ----------
+        File (28): summary2
+        ----------
```
snopoke, line 49: more compact would be to use markdown list notation:
```suggestion
        * File (27): summary1
        * File (28): summary2
```

If multiline summaries are an issue we could use yaml or json

snopoke: Might also want to include the content type so that it knows if it's an image or not.

snopoke (commented):
For the rendering, we could instruct the LLM how to reference the attachment in it's response. I think we could add this to the tool description or else the tool response like:

```
File 3 added. You can use this markdown link to reference it in your response: `[text](file:3)` or `![](file:3)` if it is an image.
```

apps/experiments/models.py
```diff
@@ -595,6 +595,7 @@ class AgentTools(models.TextChoices):
     DELETE_REMINDER = "delete-reminder", gettext("Delete Reminder")
     MOVE_SCHEDULED_MESSAGE_DATE = "move-scheduled-message-date", gettext("Move Reminder Date")
     UPDATE_PARTICIPANT_DATA = "update-user-data", gettext("Update Participant Data")
+    ATTACH_MEDIA = "attach-media", gettext("Attach Media")
```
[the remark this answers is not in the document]
SmittieC: Thanks for flagging, but this is by design

apps/chat/agent/tools.py
```diff
@@ -171,6 +181,37 @@ def action(self, key: str, value: Any):
         return "Success"
 
 
+class AttachMediaTool(CustomBaseTool):
+    name: str = AgentTools.ATTACH_MEDIA
+    description: str = "Attach a media file to your response"
+    requires_session: bool = True
+    args_schema: type[schemas.AttachMediaSchema] = schemas.AttachMediaSchema
+    callback: Callable | None = None
```
snopoke, line 189: shouldn't this be a required field? i.e. not None?

Also you can add type hints to callable:
```python
Callable[[int], Any]
```

apps/pipelines/nodes/base.py
```diff
@@ -45,6 +45,22 @@ def add_temp_state_messages(left: dict, right: dict):
...
 
+def merge_dicts(left: dict, right: dict):
+    """
+    Merge two dictionaries, combining values for the same key into a list. The value of any key is expected to be a list
+    """
+    output = {**left}
+    for key, value in right.items():
+        if key in output:
+            if isinstance(output[key], list):
+                output[key] = list(set(output[key]) | set(value))
```
snopoke, line 56: what made you decide to merge using sets?

apps/pipelines/models.py
```diff
@@ -260,13 +260,18 @@ def invoke(
             output = PipelineState(**output).json_safe()
             pipeline_run.output = output
             if save_run_to_history and session is not None:
-                metadata = output.get("message_metadata", {})
                 if save_input_to_history:
                     self._save_message_to_history(
-                        session, input["messages"][-1], ChatMessageType.HUMAN, metadata=metadata.get("input", {})
+                        session,
```
snopoke, line 265: changes from `main` need to be merged in here

apps/pipelines/nodes/base.py
SmittieC: To deduplicate the values

stephherbers (approved):
looks good to me

PR #1326, opened 2025-03-17
Generate missing encryption keys

SmittieC:
## Description

See [this support ticket](https://dimagi.atlassian.net/browse/SUPPORT-23565)
Missing encryption keys should be generated. The mobile app will ask OCS for an encryption key first, and only then try to decrypt the messages. This means that OCS can simply return the key that was used to encrypt the message.

## User Impact

OCS will be able to send messages to users before the `generate_encryption_key` API was hit.

### Demo

N/A

### Docs and Changelog

N/A


SmittieC:
There's UAT tomorrow, so merging this so long

Issue #1307, opened 2025-03-11
[Chatbots phase1] details page


snopoke:
With reference to this doc: https://docs.google.com/document/d/1k3LlbhWac3YmHKlHJCXBLNFeaj3cDFW-nBQEo1PZHGI/edit?tab=t.6va3nm2ju3o8#heading=h.larqeoclx90w

* Implement the 'Add New' dialog
* Implement the chatbot details page

PR #1327, opened 2025-03-17
Chatbots top level navigation and list view

### LINKED ISSUE #1306: [Chatbots phase1] Nav and list view
With reference to this doc: https://docs.google.com/document/d/1k3LlbhWac3YmHKlHJCXBLNFeaj3cDFW-nBQEo1PZHGI/edit?tab=t.6va3nm2ju3o8#heading=h.larqeoclx90w

* Add the feature flag
* Add the top level navigation
* Add the chatbot list view

barora05:
## Description
Add chatbot navigation option and chatbot list view. Closes #1306 

### Demo
image


apps/chatbots/views.py
```diff
@@ -0,0 +1,30 @@
...
+from apps.chatbots.tables import ChatbotTable
+from apps.experiments.models import Experiment
+from apps.experiments.views.experiment import BaseExperimentView
+from apps.teams.decorators import login_and_team_required
+from apps.utils.BaseTableView import BaseTableView
+from apps.utils.helpers import generic_home
+
+
+@login_and_team_required
+@permission_required("chatbots.view_chatbot", raise_exception=True)
```
snopoke, line 13: Django permissions are based on Django app names and their DB models. Since there is no `ChatBot` DB model there also isn't a `view_chatbot` permission. This should use the same permissions as the current experiment list view `experiments.view_experiment`

apps/chatbots/views.py
```diff
@@ -0,0 +1,30 @@
...
+def chatbots_home(request, team_slug: str):
+    return generic_home(request, team_slug, "Chatbots", "chatbots:table", "chatbots:new")
```
snopoke, line 15: 👍🏻

apps/utils/BaseTableView.py
```diff
@@ -0,0 +1,32 @@
...
+from django.contrib.postgres.search import TrigramSimilarity
+from django.db.models import Q
+from django_tables2 import SingleTableView
+
+
+class BaseTableView(SingleTableView, PermissionRequiredMixin):
+    paginate_by = 25
+    template_name = "table/single_table.html"
+
+    def get_queryset(self):
```
snopoke, line 11: This filtering here is quite specific to the `ExperimentModel` so I think this class should be renamed `BaseExperimentTableView`

apps/utils/helpers.py
```diff
@@ -0,0 +1,20 @@
+from django.template.response import TemplateResponse
+from django.urls import reverse
+
+from apps.generics.help import render_help_with_link
+
+
+def generic_home(request, team_slug: str, title: str, table_url_name: str, new_url: str):
```
snopoke, line 7: This probably belongs in `generics/views.py`

templates/web/components/team_nav.html
```diff
@@ -47,6 +47,12 @@
       {% endif %}
     </ul>
   </li>
+  <li>
```
snopoke, line 50: This needs to be hidden with a feature flag:

```
{% flag "chatbots" %}
...
{% endflag %}
```

apps/utils/helpers.py
```diff
@@ -0,0 +1,20 @@
...
+    return TemplateResponse(
+        request,
+        "generic/object_home.html",
+        {
+            "active_tab": title.lower(),
+            "title": title,
+            "title_help_content": render_help_with_link("", title.lower()),
```
snopoke, line 14: Using `title.lower()` seems risky since it would be easy for someone to change the title not realizing it also impacts the help text. It also makes it hard to find usages of a specific help text key.

apps/chatbots/tests/test_chatbot_views.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@pytest.mark.django_db()
+def test_chatbot_experiment_table_view(client):
+    User = get_user_model()
+    user = User.objects.create_user(username="testuser", password="testpass")
+    team = Team.objects.create(name="Test Team", slug="test-team")
+    content_type = ContentType.objects.get_for_model(Experiment)
+    permission = Permission.objects.get(codename="view_experiment", content_type=content_type)
+    user.user_permissions.add(permission)
```
snopoke, line 42: FYI you could use the `team_with_users` fixture:

```suggestion
def test_chatbot_experiment_table_view(client, team_with_users):
    team = team_with_users
    user = team.members.first()
```

(Both user's in this team have the necessary permissions - see `TeamWithUsersFactory`)

templates/sinlge_chatbot_home.html
```diff
@@ -0,0 +1,391 @@
...
+    function selectTabFromHash() {
+      const tabId = window.location.hash.replace('#', '');
+      const tab = document.getElementById(`tab-${tabId}`);
+      if (tab) {
+        tab.checked = true;
+        showTabContent(tabId);
+      }
+    }
+
+    document.addEventListener('DOMContentLoaded', () => {
```
snopoke, line 336: This code has since been replaced with the new session filter widget

snopoke (dismissed):
I'm OK with this being merged and additional work being done in another PR but also fine to keep address the current comment regarding the filter widget in this PR if you want.

apps/chatbots/tests/test_chatbot_views.py
barora05: done

barora05:
@snopoke I have removed sigle_chatbot_home file from this MR as that would be required for #1307. We can merge this

PR #1329, opened 2025-03-17
Centralize participant data access

### LINKED ISSUE #1181: Expand participant data available to the Python Node

stephherbers:
## Description

Prt 3 and resolves #1181

## User Impact

N/A

### Demo

N/A

### Docs and Changelog

N/a


apps/service_providers/tests/test_prompt_context.py
```diff
@@ -16,8 +16,9 @@ def mock_session():
     session.participant.user = None
 
     proxy_mock = Mock()
-    proxy_mock.get_data.return_value = "participant data"
+    proxy_mock.get.return_value = "participant data"
```
SmittieC, line 19: Should the return value not be a `dict`?

apps/pipelines/nodes/base.py
```diff
@@ -176,6 +177,9 @@ def get_output_map(self) -> dict:
         """A mapping from the output handles on the frontend to the return values of _process_conditional"""
         raise NotImplementedError
 
+    def get_participant_proxy(self, state: PipelineState) -> "ParticipantDataProxy":
```
SmittieC, line 180: `get_participant_data_proxy` rather?

apps/service_providers/llm_service/prompt_context.py
```diff
@@ -134,3 +136,66 @@ def __eq__(self, other):
...
+
+class ParticipantDataProxy:
+    """Allows multiple access without needing to re-fetch from the DB"""
+
+    @classmethod
+    def from_state(cls, pipeline_state) -> Self:
+        # using `.get` here for the sake of tests. In practice the session should always be present
+        return cls(pipeline_state.get("experiment_session"))
+
+    def __init__(self, experiment_session):
```
SmittieC, line 149: Nit: Keep the `__init__` method at the top

SmittieC (commented):
A few minor comments, but otherwise looks good to me

apps/pipelines/nodes/base.py
snopoke:
```suggestion
    def get_participant_data_proxy(self, state: PipelineState) -> "ParticipantDataProxy":
```

apps/pipelines/nodes/nodes.py
```diff
@@ -92,7 +91,7 @@ def _process(self, input, node_id: str, state: PipelineState, **kwargs) -> Pipel
                             or [],
                         }
                     )
-                proxy = ParticipantDataProxy(exp_session)
+                proxy = self.get_participant_proxy(state)
```
snopoke, line 94:
```suggestion
                proxy = self.get_participant_data_proxy(state)
```

apps/pipelines/nodes/nodes.py
```diff
@@ -455,7 +454,7 @@ def _process_conditional(self, state: PipelineState, node_id=None):
         from apps.service_providers.llm_service.prompt_context import SafeAccessWrapper
 
         if self.data_source == self.DataSource.participant_data:
-            data = ParticipantDataProxy.from_state(state).get()
+            data = self.get_participant_proxy(state).get()
```
snopoke, line 457:
```suggestion
            data = self.get_participant_data_proxy(state).get()
```

apps/pipelines/nodes/nodes.py
```diff
@@ -862,7 +861,7 @@ def _get_custom_globals(self, state: PipelineState):
 
         custom_globals = safe_globals.copy()
 
-        participant_data_proxy = ParticipantDataProxy.from_state(state)
+        participant_data_proxy = self.get_participant_proxy(state)
```
snopoke, line 864:
```suggestion
        participant_data_proxy = self.get_participant_data_proxy(state)
```

apps/service_providers/llm_service/prompt_context.py
```diff
@@ -134,3 +136,66 @@ def __eq__(self, other):
...
+        return self.session.participant.global_data | data
+
+    def set(self, data):
+        participant_data = self._get_db_object()
+        participant_data.data = data
+        participant_data.save(update_fields=["data"])
+
+        self.session.participant.update_name_from_data(data)
+
+    def get_schedules(self):
```
snopoke, line 173: I think this should be cached

apps/service_providers/tests/test_prompt_context.py
snopoke: +1

stephherbers: nice thank you

SmittieC (commented):
LGTM

SmittieC (approved):

apps/service_providers/llm_service/prompt_context.py
snopoke: I should have been clearer. We can't use this cache decorator here since it will create a global cache which will result in a memory leak. Options are to create a `cached_property` which is called by this function or create a private variable which you set once on first call.

snopoke (changes requested):
One comment but otherwise this looks good.

apps/service_providers/llm_service/prompt_context.py
```diff
@@ -137,3 +139,70 @@ def __eq__(self, other):
...
+                is_complete=False,
+                cancelled_at=None,
+            )
+            .select_related("action")
+            .order_by("created_at")
+        )
+        scheduled_messages = []
+        for message in messages:
+            scheduled_messages.append(message.as_dict())
+        return scheduled_messages
```
snopoke, line 203: This should be inverted so that the cache attribute is set if it is None:
```suggestion
        if self._scheduled_messages is None:
            from apps.events.models import ScheduledMessage
            experiment = self.session.experiment_id
            participant = self.session.participant_id
            team = self.session.experiment.team
            messages = (
                ScheduledMessage.objects.filter(
                    experiment_id=experiment,
                    participant_id=participant,
                    team=team,
                    is_complete=False,
                    cancelled_at=None,
                )
                .select_related("action")
                .order_by("created_at")
            )
            self._scheduled_messages = [message.as_dict() for message in messages]
        return self._scheduled_messages
```

snopoke (changes requested):
One last change

snopoke (approved):

PR #1335, opened 2025-03-18
Added Tag Validation on UI

### LINKED ISSUE #1282: Update tag UI input to restrict the maximum length of tags according to the DB field.
This will need to be done by adjusting the tom-select code here: https://github.com/dimagi/open-chat-studio/blob/e24caeadd307f1ebcf827d60c3a40ac95d1dacdb/assets/javascript/tag-multiselect.js#L35-L44

I think the best option will be to se a ['createFilter'](https://tom-select.js.org/docs/#createfilter) function to restrict the allowed length but that will need to be tested.

A useful error message should be shown if the tag name is too long.

Sentry Issue: [OPEN-CHAT-STUDIO-R7](https://dimagi.sentry.io/issues/6362535494/?referrer=github_integration)

```
StringDataRightTruncation: value too long for type character varying(100)
  File "django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
  File "psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)

DataError: value too long for type character varying(100)
(18 additional frame(s) were not displayed)
...
  File "apps/annotations/views/tag_views.py", line 149, in post
    obj.tags.create(team=request.team, name=tag_name, created_by=request.user)
```

vaibhavK1199:
Fixes #1282

### DIFF
--- a/assets/javascript/tag-multiselect.js
+++ b/assets/javascript/tag-multiselect.js
@@ -36,6 +36,13 @@ function configureTomSelect() {
       plugins: ["remove_button", "caret_position", "input_autogrow"],
       maxItems: null,
       create: true,
+      createFilter: (input) => {
+        if (input.length > 100) {
+          el.tomselect.dropdown_content.innerHTML = `<div class="ts-error-message" style="color: red; padding: 5px;">Tag name too long. Maximum 100 characters allowed.</div>`;
+          return false;
+        }
+        return true;
+      },
       onItemAdd: addTag('onItemAdd', objectInfo),
       onItemRemove: removeTag('onItemRemove', objectInfo),
       onBlur: () => {


SmittieC (commented):
While your approach is ok, I am wondering if we shouldn't rather set the max length on the input element instead. I can't quite articulate why I prefer this above your approach.

If you inspect the select widget that TomSelect builds, you'll see that it uses an input element for the new tag:

image

The id is whatever the select's id is with `-ts-control` appended. So we can find this element for each tom select instance and add the `max_length` attribute to it. Something like:

```python
    ....
    let inputEle = document.getElementById(el.id + "-ts-control");
    inputEle.setAttribute('maxlength', '100');
```

vaibhavK1199:
> While your approach is ok, I am wondering if we shouldn't rather set the max length on the input element instead. I can't quite articulate why I prefer this to your approach 🤔
> 
> If you inspect the select widget that TomSelect builds, you'll see that it uses an input element for the new tag:
> 
> image
> 
> The id is whatever the select's id is with `-ts-control` appended. So we can find this element for each tom select instance and add the `max_length` attribute to it. Something like:
> 
> ```python
>     ....
>     let inputEle = document.getElementById(el.id + "-ts-control");
>     inputEle.setAttribute('maxlength', '100');
> ```

Thats another way to keep this minimal, but i was under the impression that the user must see a error msg in case characters exceed, but by considering your suggestion we can simply restricting the user from entering such max length tags, will update this !

snopoke (commented):
I actually think this is a reasonable way to do it and display a user friendly error message. @vaibhavK1199 can you record a short demo and attach it to the PR please.

vaibhavK1199:
> I actually think this is a reasonable way to do it and display a user friendly error message. @vaibhavK1199 can you record a short demo and attach it to the PR please.

I have actually added a screenshot for the same on the PR https://github.com/dimagi/open-chat-studio/issues/1282#issuecomment-2735455017
let me know if this works or should i record a demo video for this.

snopoke (approved):

PR #1341, opened 2025-03-20
only return the final message from assistants

### LINKED ISSUE #1340: Assistants are showing 'thinking' messages
As shown below, the assistant's thinking messages are being shown to users in OCS. Also, the messages are in reverse order.

OCS should only show the final message from the assistant.

Image

Image

snopoke:
## Description
Resolves #1340

Assume that all messages prior to the final message are 'thinking' messages related to tool usage and only record the last message.

I'm not if this changed at some point (that threads now include thinking messages) or whether we just never noticed it before.

## User Impact
Users won't see 'thinking' messages from assistants.

### Docs and Changelog
https://github.com/dimagi/open-chat-studio-docs/pull/72


SmittieC (approved):
Looks fine to me

stephherbers (approved):

PR #1350, opened 2025-03-20
Paginate Chats Messages

### LINKED ISSUE #1244: Paginate chat transcript message list view
* Page size: 100
* Django template file: https://github.com/dimagi/open-chat-studio/blob/00ec45f71730911eea66823270c267c2c613a632/templates/experiments/components/experiment_chat.html#L3

This will require converting the table to load asynchronously using HTMX. It will also require updating features like the search to be done server side.

stephherbers:
## Description

resolves #1244
Screenshot 2025-03-20 at 2 02 28 PM
Screenshot 2025-03-20 at 2 02 37 PM

- previous page
- next page
- first page
- last page
- compatible with search bar
### Demo

[Loom video](https://www.loom.com/share/34f779a769db4329a098c5d382764b3a?sid=f4c19e7a-4361-4eea-b89e-c05656e25602)

### Docs and Changelog

yes to changelog


apps/experiments/templatetags/pagination_tags.py
```diff
@@ -0,0 +1,23 @@
...
+    return int(value) * int(arg)
+
+
+@register.filter
+def subtract(value, arg):
+    return int(value) - int(arg)
+
+
+@register.filter
+def add(value, arg):
```
SmittieC, line 17: Looks like django has an [add](https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#add) filter

apps/experiments/views/experiment.py
```diff
@@ -1271,6 +1271,50 @@ def _experiment_chat_ui(request, embedded=False):
...
+    session = request.experiment_session
+    experiment = request.experiment
+    page = int(request.GET.get("page", 1))
+    search = request.GET.get("search", "")
+    page_size = 100
+    all_messages = session.get_messages_for_display()
+
+    if search:
+        filtered_messages = [
+            msg for msg in all_messages if any(search.lower() in tag.lower() for tag in msg.all_tag_names())
```
SmittieC, line 1287: Feels like this can be done with a query?

SmittieC (commented):
This is great. Another way that I've seen sites do is to load the next bunch as the user scrolls down. I'm curious how easy/difficult that would be

apps/experiments/views/experiment.py
snopoke: +1

apps/experiments/views/experiment.py
```diff
@@ -1271,6 +1271,50 @@ def _experiment_chat_ui(request, embedded=False):
...
+        ]
+    else:
+        filtered_messages = all_messages
+
+    total_messages = len(filtered_messages)
```
snopoke, line 1292: this should use a `count` query to avoid loading the full dataset into memory

templates/experiments/components/pagination_buttons.html
```diff
@@ -0,0 +1,32 @@
+{% if total_pages > 1 %}
+  <div class="flex {% if btn_size == 'sm' %}gap-1{% else %}justify-right{% endif %}">
```
snopoke, line 2: this should use a join to create a more cohesive set of buttons: https://v4.daisyui.com/components/join/#join

templates/experiments/components/pagination_buttons.html
```diff
@@ -0,0 +1,32 @@
...
+    {% if page > 1 %}
+      <a href="#table-top"
+         class="btn {% if btn_size == 'sm' %}btn-sm{% endif %}"
+         hx-get="{% url 'experiments:experiment_session_messages_view' request.team.slug experiment.public_id experiment_session.external_id %}?page=1{% if search %}&search={{ search }}{% endif %}"
```
snopoke, line 6: Would it work to use the [querystring](https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#querystring) tag for these?

```suggestion
         hx-get="{% url 'experiments:experiment_session_messages_view' request.team.slug experiment.public_id experiment_session.external_id %}{% querystring page=1 search=search %}"
```

templates/experiments/experiment_session_view.html
```diff
@@ -85,4 +89,28 @@ <h3 class="text-base font-semibold leading-7 mt-8 pl-4">Experiment details</h3>
   >
   </script>
   <script>SiteJS.tagMultiselect.setupTagSelects();</script>
+  <script>
+    document.addEventListener('DOMContentLoaded', function() {
```
snopoke, line 93: What is the purpose of this code? It seems like it could be done with an `htmx` `load` trigger.

templates/experiments/experiment_session_view.html
```diff
@@ -85,4 +89,28 @@ <h3 class="text-base font-semibold leading-7 mt-8 pl-4">Experiment details</h3>
   >
   </script>
   <script>SiteJS.tagMultiselect.setupTagSelects();</script>
```
snopoke, line 91: Does the tag UI work after paginating?

templates/experiments/components/experiment_chat.html
```diff
@@ -1,103 +1,158 @@
...
+            hx-target="#messages-container"
+            hx-include="[name='search']"
+          >
+          {% if search %}
+            <button
+              class="btn btn-ghost btn-sm"
+              hx-get="{% url 'experiments:experiment_session_messages_view' request.team.slug experiment.public_id experiment_session.external_id %}"
+              hx-target="#messages-container"
+            >
+              <i class="fa-solid fa-times"></i> Clear
```
snopoke, line 28: nit: remove 'Clear' to be consistent with other UIs
```suggestion
              <i class="fa-solid fa-times"></i>
```

snopoke (commented):
Generally looks good. One general comment is that we should add status / loading indicators for all ajax actions and where possible disable the buttons during the actions (See https://htmx.org/attributes/hx-disabled-elt/)

templates/experiments/experiment_session_view.html
stephherbers: yes still works

SmittieC (approved):
LGTM :rocket:

PR #1351, opened 2025-03-21
Migration to add a default history mode

SmittieC:
## Description

1. Reverts [this revert](https://github.com/dimagi/open-chat-studio/pull/1346)
2. Added a migration to add a default history mode

I could not reproduce the issue really, but did the see error that @stephherbers reported initially:
```
{'LLMResponseWithPrompt-UcgXz': {'history_mode': "Input should be 'summarize', 'truncate_tokens' or 'max_history_length'"}})}
```

Strange that when I built a pipeline on main and checked out the history mode changes, I didn't see the error anymore, even with new pipelines where I confirmed that the node params did not have a history mode param.

## User Impact

This _should_ fix the pipeline errors

### Demo

N/A

### Docs and Changelog

<Stay tuned>


apps/pipelines/migrations/0014_set_default_history_mode.py
```diff
@@ -0,0 +1,30 @@
...
+from apps.pipelines.nodes.nodes import StartNode, EndNode
+
+CHUNK_SIZE = 100
+
+def _set_default_history_mode(apps, schema_editor):
+    Node = apps.get_model('pipelines', 'Node')
+    node_queryset = Node.objects.exclude(type__in=[StartNode.__name__, EndNode.__name__])
+    nodes = []
+    for node in node_queryset.iterator(chunk_size=CHUNK_SIZE):
+        node.params['history_mode'] = PipelineChatHistoryModes.SUMMARIZE
```
snopoke, line 14: not all nodes support history mode - this should only apply updates to known node types

SmittieC: Ah yes of course, what was I thinking

barora05 (approved):

PR #1363, opened 2025-03-26
add security decorators and checks

snopoke:
## Description
Prompted by [OPEN-CHAT-STUDIO-SF](https://dimagi.sentry.io/issues/6459913444/events/6656277c4a44419089ee611fa509c53b/) I did an audit of all views and found a number that didn't have correct security checks.

I've also added a `@team_required` decorator for views that don't require auth but do require a real team.

## Testing
I tested all these views with the exception of `verify_public_chat_token` and `download_file`

## User Impact
None

### Docs and Changelog
NA

### DIFF
--- a/apps/service_providers/views.py
+++ b/apps/service_providers/views.py
@@ -2,7 +2,8 @@ from collections import defaultdict
 
 from django.conf import settings
 from django.contrib.auth.decorators import login_required, permission_required
-from django.core.exceptions import ValidationError
+from django.contrib.auth.mixins import PermissionRequiredMixin
+from django.core.exceptions import PermissionDenied, ValidationError
 from django.db import transaction
 from django.http import HttpResponse, HttpResponseBadRequest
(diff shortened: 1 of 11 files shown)


apps/service_providers/views.py
```diff
@@ -48,8 +55,11 @@ def matches_blocking_deletion_condition(obj):
 
 
 @require_http_methods(["DELETE"])
+@login_and_team_required
 def delete_service_provider(request, team_slug: str, provider_type: str, pk: int):
     provider = ServiceProvider[provider_type]
+    if not request.user.has_perm(provider.get_permission("delete")):
```
SmittieC, line 61: Why not use the `@permission_required` decorator here?

snopoke: Because the permission is dependent on the service provider type.

SmittieC (approved):

PR #1378, opened 2025-03-27
Multimedia | 5 | Send attachments as links for non-web channels

### LINKED ISSUE #1225: [Multimedia support] Send attachments to users

### Requirements
- For the web channel, image attachments should be rendered
- For all external channels, append the list of attachments with download links

SmittieC:
## Description

Resolves #1225

- For **web channels**, the bot will include a link (or the picture in the case of an image) in the chat, and we attach download links to the message at the bottom.
- For **non-web channels**, the bot **will not** include links in the chat itself, but we append the list of attachment download links to the message.
- For API channels, we include a new `attachments` key that is a list of download links.

## User Impact

Participants on channels like Whatsapp, telegram etc will be able to download attachments by clicking on the links.
Participants on external channels will be able to download referenced files

#### Telegram example
image

#### API response example
image

### Docs and Changelog

N/A


SmittieC:
@coderabbitai please do a full in-depth review

apps/chat/channels.py
```diff
@@ -484,8 +514,12 @@ def _transcribe_audio(self, audio: BytesIO) -> str:
...
         return "Unable to transcribe audio"
 
-    def _get_bot_response(self, message: str) -> str:
-        return self.bot.process_input(message, attachments=self.message.attachments)
+    def _get_bot_response(self, message: str) -> ChatMessage:
+        response = self.bot.process_input(message, attachments=self.message.attachments)
+        if ai_message_id := self.bot.get_ai_message_id():
+            return ChatMessage.objects.get(id=ai_message_id)
+
+        return ChatMessage(content=response, message_type=ChatMessageType.AI)
```
snopoke, line 522: why not have the bot return the actual chat message?

snopoke (approved):

apps/chat/channels.py
SmittieC: This reminds me, I wanted to check when we don't save the AI message, which seems to be only in the case when using a router (without pipelines). Considering this, we _can_ just return the `ChatMessage`. This will mean _a lot_ of tests should be updated. I'll rather do this as a separate ticket if you're fine with us doing it this way?

apps/channels/views.py
```diff
@@ -147,15 +136,26 @@ def _new_api_message(request, experiment_id: uuid, version=None):
...
         experiment_channel,
         message_data["message"],
         participant_id,
         session,
     )
-    return Response(data={"response": response})
+
+    attachments = []
+    if attached_files := ai_response.get_attached_files():
+        session_id = ExperimentSession.objects.get(chat=ai_response.chat).id
```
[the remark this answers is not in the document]
SmittieC: Lol, yes it is, thanks

apps/chat/channels.py
SmittieC: See [this PR](https://github.com/dimagi/open-chat-studio/pull/1414) where I'm returning the `ChatMessage` instance from `process_input`. The pipeline now also returns the chat message instead of the pipeline state

snopoke (dismissed):

Issue #1166, opened 2025-02-07
Increase real estate for pipelines


snopoke:
When editing a pipeline we should increase the screen space available to the editor.

We could:

* remove the global gutter (remove the container class [here](https://github.com/dimagi/open-chat-studio/blob/fdab29f128f9ef319136c3fdec98a6032ef6394e/templates/web/app/app_base.html#L10))
  * I tried this but the rest of the UI isn't great - it would need more tweaking
* Make pipelines UI fill the right gutter (keeping the left gutter in place)
* Provide a full screen mode for pipelines

Davemaina1:
.take

Davemaina1:
Design doc posted on #4408, the sub-issue covering this: https://github.com/dimagi/open-chat-studio/issues/4408#issuecomment-5592233177

PR #1380, opened 2025-03-28
Expand Pipeline Viewport

vaibhavK1199:
Providing a full screen toggle mode for pipeline building as a part of - #1166 

adding a demo for reference -

https://github.com/user-attachments/assets/8f02bc47-6c87-4f2b-8dbd-e186cbf3c1da

### DIFF
--- a/templates/pipelines/pipeline_builder.html
+++ b/templates/pipelines/pipeline_builder.html
@@ -1,5 +1,6 @@
 {% extends 'web/app/app_base.html' %}
 {% load static %}
+
 {% block page_head %}
   <link rel="stylesheet" href="{% static 'css/pipeline.css' %}">
   <style>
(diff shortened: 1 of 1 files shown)


templates/pipelines/pipeline_builder.html
```diff
@@ -18,25 +18,86 @@
...
     document.addEventListener('DOMContentLoaded', () => {
       SiteJS.pipeline.renderPipeline("#pipelineBuilder", "{{ request.team.slug }}", {{ pipeline_id }});
-    }
-    )
+      const toggleButton = document.querySelector('.fullscreen-toggle');
+      const container = document.getElementById('pipelines-container');
+      const sidebar = document.querySelector('.w-64');
+      const icon = document.getElementById('fullscreen-enter-icon');
+
+      function toggleFullScreen() {
```
SmittieC, line 75: have you tried doing this using alpineJS? You might find it easier and less verbose

templates/pipelines/pipeline_builder.html
```diff
@@ -18,25 +18,86 @@
...
+    }
   </style>
 {% endblock page_head %}
 {% block breadcrumbs %}
 {% endblock %}
 {% block app %}
-  <div class="max-w-7xl mx-auto" id="pipelineBuilder">
+  <div class="flex-1 relative" id="pipelines-container">
+    <button class="fullscreen-toggle" title="Toggle Full Screen">
+      <svg class="w-5 h-5" fill="none" stroke="currentColor" viewBox="0 0 24 24">
```
SmittieC, line 48: I'd suggest using the `fa-solid fa-expand` font awesome icon here and the `fa fa-compress` icon for the "shrink" one. I think it will fit a bit better with the rest of the look

templates/pipelines/pipeline_builder.html
vaibhavK1199: Yes gave it a shot you can check my latest commit.

templates/pipelines/pipeline_builder.html
```diff
@@ -2,6 +2,8 @@
 {% load static %}
 {% block page_head %}
   <link rel="stylesheet" href="{% static 'css/pipeline.css' %}">
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.5.1/css/all.min.css" integrity="sha512-..." crossorigin="anonymous" referrerpolicy="no-referrer" />
```
SmittieC, line 5: No need to include this here. We have these modules included through the webpack bundler

templates/pipelines/pipeline_builder.html
```diff
@@ -18,25 +20,71 @@
...
+    </button>
+    <div class="max-w-7xl mx-auto" id="pipelineBuilder"></div>
   </div>
+  <aside class="w-64 hidden" x-ref="sidebar">Sidebar content here</aside>
 {% endblock %}
 {% block page_js %}
   {{ parameter_values|json_script:"parameter-values" }}
   {{ default_values|json_script:"default-values" }}
   {{ node_schemas|json_script:"node-schemas" }}
-  {{ flags_enabled|json_script:"flags-enabled" }}
```
SmittieC, line 33: This was probably removed by accident?

templates/pipelines/pipeline_builder.html
```diff
@@ -18,25 +20,71 @@
...
+      class="fullscreen-toggle"
+      title="Toggle Full Screen"
+      @click="isFullscreen = !isFullscreen"
+      x-on:fullscreenchange="isFullscreen = !!document.fullscreenElement"
+    >
+      <i :class="isFullscreen ? 'fas fa-compress' : 'fas fa-expand'" class="w-5 h-5"></i>
+    </button>
+    <div class="max-w-7xl mx-auto" id="pipelineBuilder"></div>
   </div>
+  <aside class="w-64 hidden" x-ref="sidebar">Sidebar content here</aside>
```
SmittieC, line 74: What is this side bar?

templates/pipelines/pipeline_builder.html
```diff
@@ -18,25 +20,71 @@
...
 {% endblock page_head %}
 {% block breadcrumbs %}
 {% endblock %}
 {% block app %}
-  <div class="max-w-7xl mx-auto" id="pipelineBuilder">
+  <div class="flex-1 relative" id="pipelines-container" x-data="{ isFullscreen: false }" x-init="
+                                                                                                 $watch('isFullscreen', (value) => {
+                                                                                                 if (value) {
+                                                                                                 $el.requestFullscreen().catch(err => console.error('Fullscreen failed:', err));
+                                                                                                 if (sidebar) sidebar.classList.add('hidden');
```
SmittieC, line 57: The sidebar can be referenced using `x-ref`. For instance, you can put `x-ref=sidebar` on the sidebar element [here](https://github.com/dimagi/open-chat-studio/blob/6d378239c60a7198a35499a97f9014ced307d04b/templates/web/app/app_base.html). That said, in alpine there's an easy way to show/hide an element based on some value in `data`, and that is to use `x-show=!isFullscreen` (see the [x-show docs](https://alpinejs.dev/directives/show)).

In order to do this, you will have to put the data directive with the `isFullscreen` value on a level that contains both the sidebar and app elements. I think an [alpine store](https://alpinejs.dev/magics/store) might come in more handy here, since the store allows components on different parts of the page to communicate with each other. We can initialize the store maybe somewhere [this file](https://github.com/dimagi/open-chat-studio/blob/d8cfc96a4d4704e2070d8a62d67870e2978c3970/templates/web/base.html) by including something like

```JS
    <script>
      document.addEventListener('alpine:init', () => {
          Alpine.store('fullscreen', {
              on: false,
              toggle() {
                  this.on = ! this.on
              }
          })
      })
    </script>
```
 in the header. This way the store is there for all pages. Then you can easily create bindings by simply referencing the store's value. For example, to show/hide the sidebar, add the following attributes on the sidebar element:
```HTML
x-data x-show="!$store.fullscreen.on"
```
(It's important to keep the `x-data` attirubte to indicate to alpineJS that this element wants to use Alpine).

PR #1387, opened 2025-03-31
Suggestion: Shell helpers

SmittieC:
Better support for importing extra modules and helpers [are coming](https://docs.djangoproject.com/en/5.2/howto/custom-shell/#top), but in the meantime I was thinking we can do with these helpers.

Feedback welcome

### DIFF
--- a/gpt_playground/scripts/shell_helpers.py
+++ b/gpt_playground/scripts/shell_helpers.py
@@ -1,0 +1,37 @@
+from apps.assistants.models import OpenAiAssistant
+from apps.experiments.models import Experiment, ExperimentSession
+
+"""
+These methods are meant to be imported by the shell for ease of use. Run like this:
+PYTHONSTARTUP=./gpt_playground/scripts/shell_helpers.py python manage.py shell
+"""
+
+
+def get_experiment(id=None, public_id=None):
+    if id:
+        return Experiment.objects.get(id=id)
+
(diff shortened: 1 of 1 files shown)


SmittieC:
> no objection to this approach, but have you looked into [shell_plus](https://django-extensions.readthedocs.io/en/latest/shell_plus.html#additional-imports)?

I initially wanted to go that route, but having to install a whole new app. I think migrating to the new way that django will allow us to do imports might be easier with this route? Happy to be wrong or wait until django comes around, but I do think this will save a tad bit of time while digging in the shell in the meantime, be it this or the shell_plus approach

stephherbers (approved):

gpt_playground/scripts/shell_helpers.py
```diff
@@ -0,0 +1,37 @@
+from apps.assistants.models import OpenAiAssistant
+from apps.experiments.models import Experiment, ExperimentSession
```
snopoke, line 2: these will need to be moved to inline imports since this script will run before Django has run it's setup code

SmittieC:
Django 5.2 has [been released](https://docs.djangoproject.com/en/5.2/releases/5.2/) which supports auto imports. I think let's upgrade and then revisit this PR

SmittieC:
Closing this for now

PR #1388, opened 2025-03-31
Use a custom type to coerce empty strings to `None`

SmittieC:
## Description

Added a custom type `OptionalInt` that is the equivalent of `int | None`, but also coerces empty strings to `None`. This comes in handy when users deselect something from a select input, like source material. The value from the UI comes back as an emtpy string.

## User Impact

Deselecting source material for instance

### Demo

N/A

### DIFF
--- a/apps/pipelines/tests/test_nodes.py
+++ b/apps/pipelines/tests/test_nodes.py
@@ -1,8 +1,8 @@
 import pytest
-from pydantic import BaseModel
+from pydantic import BaseModel, TypeAdapter
 from pydantic_core import ValidationError
 
-from apps.pipelines.nodes.nodes import SendEmail, StructuredDataSchemaValidatorMixin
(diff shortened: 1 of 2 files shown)


stephherbers (approved):

apps/pipelines/tests/test_nodes.py
```diff
@@ -45,3 +45,16 @@ def test_valid_recipient_list(self, recipient_list):
     def test_invalid_recipient_list(self, recipient_list):
         with pytest.raises(ValidationError, match="Invalid list of emails addresses"):
             SendEmail(name="email", recipient_list=recipient_list, subject="Test Subject")
+
+
+def test_optional_int_type():
```
snopoke, line 50: 👏🏻

snopoke (commented):
👍🏻

Issue #971, opened 2024-12-11
Use json mode for router nodes


snopoke:
Currently we rely on parsing the output of the LLM for routers. Considering that the output is always expected to be a single keyword, we could use a structured output approach. This would also better guarantee that the returned value is one of the required values.

Using Langchain's [structured output](https://python.langchain.com/v0.2/docs/how_to/structured_output/#the-with_structured_output-method) method should make it simpler to do for any of our supported models.

Building the desired schema can be done in a similar way to how the [ExtractStructuredDataNode](https://github.com/dimagi/open-chat-studio/blob/e705ca6aef5b8a1c8872853d1407ed08e0be17f4/apps/pipelines/nodes/nodes.py#L540-L568) does it now.

Currently the schema would just be a single literal field:

```python
routes = ["a", "b"]
create_model("name", route=Literal[tuple(routes)])
```

Related: #1232

PR #1403, opened 2025-04-01
Use json mode for router nodes

stephherbers:
## Description

Resolves [#971](https://github.com/dimagi/open-chat-studio/issues/971)

## User Impact

no user-visible impact, but better reliability
 
### Demo

no

### Docs and Changelog

No?

### DIFF
--- a/apps/pipelines/nodes/nodes.py
+++ b/apps/pipelines/nodes/nodes.py
@@ -13,7 +13,7 @@ from django.core.validators import validate_email
 from django.db.models import TextChoices
 from jinja2.sandbox import SandboxedEnvironment
 from langchain_core.messages import BaseMessage
-from langchain_core.prompts import MessagesPlaceholder, PromptTemplate
+from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder, PromptTemplate
 from langchain_core.runnables import RunnableLambda, RunnablePassthrough
 from langchain_text_splitters import RecursiveCharacterTextSplitter
 from pydantic import BaseModel, BeforeValidator, Field, create_model, field_validator, model_validator
@@ -41,6 +41,7 @@ from apps.pipelines.tasks import send_email_from_pipeline
 from apps.service_providers.exceptions import ServiceProviderConfigError
 from apps.service_providers.llm_service.adapters import AssistantAdapter, ChatAdapter
 from apps.service_providers.llm_service.history_managers import PipelineHistoryManager
+from apps.service_providers.llm_service.prompt_context import PromptTemplateContext
 from apps.service_providers.llm_service.runnables import (
     AgentAssistantChat,
     AgentLLMChat,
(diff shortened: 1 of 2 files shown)


stephherbers:
@coderabbit please review

apps/pipelines/nodes/nodes.py
```diff
@@ -447,35 +456,49 @@ class RouterNode(RouterMixin, Passthrough, HistoryMixin):
...
         result = chain.invoke(context, config=self._config)
-        keyword = self._get_keyword(result.content)
+
+        valid_keywords = [k.lower() for k in self.keywords]
+        default_keyword = self.keywords[0].lower()
+        try:
+            keyword = (
+                getattr(result, "route", None)
+                or (result.get("route") if isinstance(result, dict) else None)
+                or self._get_keyword(str(result))
```
SmittieC, line 491: Should `result.get("route")` not always work?

apps/pipelines/nodes/nodes.py
```diff
@@ -430,6 +431,14 @@ def _get_keyword(self, result: str):
         else:
             return self.keywords[0].lower()
 
+    def _create_router_schema(self):
```
SmittieC, line 434: Do we think a test is needed for this method?

SmittieC (approved):
Generally looks good to me. Minor comments

apps/pipelines/nodes/nodes.py
stephherbers: it failed in tests without adding the rest which is why I added, but open to suggestions

stephherbers: great point: [fd566ab](https://github.com/dimagi/open-chat-studio/pull/1403/commits/fd566abf41dcc75e8f8bfd5f8c93165d63d048e3)

apps/utils/langchain.py
```diff
@@ -155,6 +174,21 @@ def _call(self, messages: list[BaseMessage], *args, **kwargs) -> str | BaseMessa
...
+            if isinstance(input_value, ChatPromptValue):
+                messages = input_value.messages
+            else:
+                messages = [input_value]
+            self._call(messages, *args, **kwargs)
+            user_message = messages[-1].content.lower() if messages else ""
+            if hasattr(schema, "__annotations__") and "route" in schema.__annotations__:
+                return type("RouterOutput", (), {"route": user_message})
+
+        return RunnableLambda(_structured_output_handler)
```
snopoke, line 190: do you need to re-define this on this class? Can't you put this implementation on the base `FakeLlm` class?

stephherbers: [de59d0a](https://github.com/dimagi/open-chat-studio/pull/1403/commits/de59d0a39e4a3a0c6633f0415f26845a8b0ff089) right thanks

snopoke (approved):
@stephherbers please confirm that you tested this with an actual pipeline before merging.

stephherbers:
@snopoke yes confirming I tested on a real pipeline

PR #1405, opened 2025-04-02
Multimedia | 6 | Collection versioning

### LINKED ISSUE #1219: [Multimedia support] Version support for collections

- When a pipeline is versioned, all collections that are used in the pipeline should be versioned as well.
- If there are no changes between the "working" version of the collection and its latest version (if there is one), the latest version should be used in the new node's version
- Implement `version_details` on the collection's model. See [this](https://github.com/dimagi/open-chat-studio/blob/255dffa440ce0d4c088ebdbbcd5b5ccc8cac7e43/apps/assistants/models.py#L241) as an example.

SmittieC:
## Description

Resolves #1219 

Files are also now versioned, since their summaries can change, which can alter the bahaviour of a bot.

## User Impact

Users can version collections

### Demo

N/A

### Docs and Changelog

N/A yet

### DIFF
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -1,3 +1,4 @@
+import logging
 from collections import defaultdict
 from collections.abc import Iterator
 from dataclasses import dataclass
@@ -27,6 +28,8 @@ from apps.pipelines.nodes.helpers import temporary_session
 from apps.teams.models import BaseTeamModel
 from apps.utils.models import BaseModel
 
+versioning_logger = logging.getLogger("ocs.versioning")
+
 
 @dataclass
 class ModelParamSpec:
@@ -34,6 +37,9 @@ class ModelParamSpec:
 
     param_name: str
     model_cls: VersionsMixin
+
+    def get_object(self, id: int):
+        return self.model_cls.objects.get(id=id)
 
 
 class PipelineManager(VersionsObjectManagerMixin, models.Manager):
@@ -557,9 +563,13 @@ class Node(BaseModel, VersionsMixin, CustomActionOperationMixin):
 
         for spec in model_param_specs.get(self.type, []):
             if instance_id := self.params[spec.param_name]:
-                instance_cls = spec.model_cls
-                obj = instance_cls.objects.get(id=instance_id)
-                obj.archive()
+                try:
+                    obj = spec.get_object(instance_id)
+                    obj.archive()
+                except self.model_cls.DoesNotExist:
+                    versioning_logger.exception(
+                        f"Failed to archive {spec.param_name} with id {instance_id}, since it could not be found"
+                    )
 
 
 class PipelineRunStatus(models.TextChoices):


SmittieC:

apps/pipelines/models.py
```diff
@@ -418,19 +426,28 @@ def create_new_version(self):
...
             if not assistant.is_a_version:
                 assistant_version = assistant.create_new_version()
                 # convert to string to be consistent with values from the UI
                 new_version.params["assistant_id"] = str(assistant_version.id)
 
+        if self.type == LLMResponseWithPrompt.__name__:
+            if collection_id := self.params.get("collection_id"):
+                collection = Collection.objects.get(id=collection_id)
+
+                if collection.has_versions is False or collection.compare_with_latest():
```
stephherbers, line 445: whats the runtime of collection.compare_with_latest() with large files? I assume .compare() would take a significant amount of time?

SmittieC: Good question. Since the file model also implements `version_details`, we're using that when comparing, and since version details consists only of of the name and the summary of each file, it should have a small memory footprint. The only reason why one file might be a version of another is when the name or summary of the file changed.

apps/documents/models.py
```diff
@@ -1,16 +1,47 @@
...
-        constraints = [models.UniqueConstraint(fields=["team", "name"], name="unique_collection_per_team")]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["team", "name", "version_number", "working_version_id"],
+                name="unique_collection_version_per_team",
+            )
+        ]
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
snopoke, line 44: f-string not required:

```suggestion
        return self.name
```

apps/documents/views.py
```diff
@@ -141,14 +155,14 @@ def edit_file(request, team_slug: str, pk: int):
...
+    _update_collection_membership(file=file, collection_ids=request.POST.getlist("collections"))
 
     return redirect("documents:collections", team_slug=team_slug, tab_name="files")
 
 
-def _update_collection_membership(file: File, collection_names: list[str]):
+def _update_collection_membership(file: File, collection_ids: list[str]):
     """Handles updating the collections a file belongs to"""
-    collections = file.team.collection_set.filter(name__in=collection_names).values_list("id", flat=True)
+    collections = file.team.collection_set.filter(id__in=collection_ids, is_version=False).values_list("id", flat=True)
```
snopoke, line 165: ❤️

apps/pipelines/models.py
```diff
@@ -27,6 +28,13 @@
 from apps.utils.models import BaseModel
 
 
+class ModelParamSpec(TypedDict):
+    """A helper class to hold the parameter name and model of those that are data base records"""
```
snopoke, line 32:
```suggestion
    """A helper class to hold the parameter name and model of those that are database records"""
```

apps/pipelines/models.py
```diff
@@ -27,6 +28,13 @@
 from apps.utils.models import BaseModel
 
 
+class ModelParamSpec(TypedDict):
```
snopoke, line 31: Making this a dataclass would allow attribute access instead of having to use item getters:

`spec.param_name` instead of `spec["param_name"]`

apps/pipelines/models.py
```diff
@@ -499,6 +519,28 @@ def requires_attachment_tool(self) -> bool:
...
+            nodes.LLMResponseWithPrompt.__name__: [
+                ModelParamSpec(param_name="collection_id", model_cls=Collection)
+                # TODO: Custom actions needed
+            ],
+        }
+
+        for spec in model_param_specs.get(self.type, []):
+            if instance_id := self.params[spec["param_name"]]:
+                instance_cls = spec["model_cls"]
+                obj = instance_cls.objects.get(id=instance_id)
```
snopoke, line 541: Could be refactored to `spec.get_object(sepf.params)`

Should probably also add error handling with logging in case the object doesn't exist.

snopoke (approved):
LGTM. Just confirm that versioning a file does not copy the file itself right - it just creates a new DB record pointing to the same file in the file storage.

apps/pipelines/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1405/commits/1b4d3a27b4dd576f83c1ee756dc0cbe00db851d7

apps/pipelines/models.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1405/commits/889296f1be686ac56c2889c7a6d20a02cce7d33e

SmittieC:
> LGTM. Just confirm that versioning a file does not copy the file itself right - it just creates a new DB record pointing to the same file in the file storage.

Correct. When calling `file.url`, it has the same value, so it's pointing to the same file

PR #1409, opened 2025-04-03
Ba/common filters

barora05:
## Description
Create common template for filters code and use it in both experiment and chatbots


barora05:
@snopoke I have updated the single_chatbot_home file code with single_experiment_home code and modified the urls to chatbots wherever required.

snopoke (approved):

PR #1411, opened 2025-04-03
Multimedia | Twilio support for media messages

### LINKED ISSUE #1223: [Multimedia support] Send image attachments | Twilio

Depends on [this ticket](https://github.com/dimagi/open-chat-studio/issues/1225)

### Requirements
Attachments should be sent directly to the chat. If it is not supported for any reason, a link should be appended to the message instead.
See the [documentation](https://www.twilio.com/en-us/blog/how-to-send-picture-whatsapp-twilio-python).

This should cover Whatsapp and Facebook

SmittieC:
## Description

Resolves #1223 

## User Impact

Participants will receive attached files over whatsapp (currently only the Twilio provider is supported)

### Demo

demo

### Docs and Changelog

N/A


apps/chat/channels.py
```diff
@@ -201,16 +207,26 @@ def participant_user(self):
...
 
+    def send_file_to_user(self, files: list[File]):
+        """
+        Sends the file to the user. This is a channel specific way of sending files.
+        The default implementation does nothing.
+        """
+        pass
+
+    def _can_send_file(self, file: File) -> bool:
+        return False
```
snopoke, line 228: how is this different from `supports_multimedia`?

apps/chat/channels.py
```diff
@@ -479,19 +534,8 @@ def _reply_voice_message(self, text: str, attached_files: list[File] = None):
             synthetic_voice = self.bot.processor_experiment.synthetic_voice
 
         speech_service = voice_provider.get_speech_service()
-        try:
-            synthetic_voice_audio = speech_service.synthesize_voice(text, synthetic_voice)
-            self.send_voice_to_user(synthetic_voice_audio)
-        except AudioSynthesizeException as e:
```
snopoke, line 485: why remove the exception handling?

snopoke (approved):
some comments but not blocking

apps/chat/channels.py
SmittieC: `supports_multimedia` indicated that we can send files to the channel, but since the channel impose restrictions on file size and types, we need to check each file individually to see if they can be send to the channel. This is where `_can_send_file comes in

PR #1414, opened 2025-04-04
process_input to return the `ChatMessage` instance

SmittieC:
This started from [here](https://github.com/dimagi/open-chat-studio/pull/1378#discussion_r2027039307)

[process_input](https://github.com/dimagi/open-chat-studio/blob/12d30c1de84c47ef7b22065a7dc69eac7392b16b/apps/chat/bots.py#L189) now returns the `ChatMessage`

Best to go through it commit-wise

Manual tests that passed
- [x] Legacy Routers
- [x] Terminal bots
- [x] Scheduled messages
- [x] Normal experiments
- [x] API
- [x] Pipeline
- [x] Pipeline test message
- [x] Seed message
- [x] Pipeline event
- [x] Safety Layer

### DIFF
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -280,8 +280,7 @@ class Pipeline(BaseTeamModel, VersionsMixin):
                 )
                 pipeline_output = ai_message
             else:
-                # Return the pipeline state. This is currently for logging purposes only.
-                pipeline_output = output
+                pipeline_output = ChatMessage(content=output)
         finally:
             if trace_service:
                 trace_service.end()
--- a/apps/events/actions.py
+++ b/apps/events/actions.py
@@ -133,6 +133,7 @@ class PipelineStartAction(EventActionHandlerBase):
             messages = [session.chat.messages.last().to_langchain_message()]
 
         input = "\n".join(f"{message.type}: {message.content}" for message in messages)
-        return pipeline.invoke(
+        output = pipeline.invoke(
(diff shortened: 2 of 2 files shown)


SmittieC:
@coderabbitai please check and make sure I didn't miss a place that expects a string output from `process_input`, but should be expecting a `ChatMessage` instance

apps/pipelines/models.py
```diff
@@ -277,7 +278,10 @@ def invoke(
                 ai_message = self._save_message_to_history(
                     session, output["messages"][-1], ChatMessageType.AI, metadata=output_metadata
                 )
-                output["ai_message_id"] = ai_message.id
+                pipeline_output = ai_message
+            else:
+                # Return the pipeline state. This is currently for logging purposes only.
+                pipeline_output = output
```
SmittieC, line 284: Should we rather instantiate `ChatMessage` with the state as the content here?

snopoke: Yes, I think we should do that

SmittieC: https://github.com/dimagi/open-chat-studio/pull/1414/commits/d8e6e512ea7278cdbdb1b1de383b892c0cb96e44

SmittieC:
@snopoke if you are happy with these changes, I'll merge them into https://github.com/dimagi/open-chat-studio/pull/1378

snopoke (approved):

SmittieC:
I'll merge this once the base branch is approved so as to not interfere with the diffs

PR #1431, opened 2025-04-09
use immer to avoid nested shallow copying

### LINKED ISSUE #1241: Update state modification in pipelines to use immer
See https://immerjs.github.io/immer/

Example:
https://github.com/dimagi/open-chat-studio/blob/main/assets/javascript/apps/pipeline/nodes/widgets.tsx#L220-L231

```typescript
function getNewNodeData(old: Node, updatedList: Array<string>) {
    return {
      ...old,
      data: {
        ...old.data,
        params: {
          ...old.data.params,
          [props.name]: updatedList,
        },
      },
    };
  }
```

Becomes
```typescript
function getNewNodeData(old: Node, updatedList: Array<string>) {
    return produce(old, new => {
        new.data.params[props.name] = updatedList
    });
  }
```

barora05:
## Description
Use immer to avoid shallow copy of nested objects manually. Closes #1241

### DIFF
--- a/assets/javascript/apps/pipeline/PipelineNode.tsx
+++ b/assets/javascript/apps/pipeline/PipelineNode.tsx
@@ -31,9 +31,7 @@ export function PipelineNode(nodeProps: NodeProps<NodeData>) {
     }
 
     setNode(id, produce((next) => {
-      if (next?.data?.params) {
-        next.data.params[name] = updateValue;
-      }
+      next.data.params[name] = updateValue;
     }));
   };
 
--- a/assets/javascript/apps/pipeline/PipelineNode.tsx
+++ b/assets/javascript/apps/pipeline/PipelineNode.tsx
@@ -8,6 +8,7 @@ import {getWidgetsForNode} from "./nodes/GetInputWidget";
 import NodeInput from "./nodes/NodeInput";
 import NodeOutputs from "./nodes/NodeOutputs";
 import {HelpContent} from "./panel/ComponentHelp";
+import { produce } from "immer";
 
 export type PipelineNode = Node<NodeData>;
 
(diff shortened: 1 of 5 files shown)


SmittieC (commented):
These are great! :rocket: 

I think there are still some areas left to change? If I search for "..." across the tsx files, I see a few other places that might need to be immer-ized

assets/javascript/apps/pipeline/PipelineNode.tsx
```diff
@@ -28,17 +29,11 @@ export function PipelineNode(nodeProps: NodeProps<NodeData>) {
...
-      data: {
-        ...old.data,
-        params: {
-          ...old.data.params,
-          [name]: updateValue,
-        },
-      },
+
+    setNode(id, produce((next) => {
+      if (next?.data?.params) {
```
snopoke, line 34: How come this has an if statement and the others don't?

snopoke (commented):
One question but otherwise looks good to me. I don't see any other obvious places to change in the code.

barora05:
> These are great! 🚀
> 
> I think there are still some areas left to change? If I search for "..." across the tsx files, I see a few other places that might need to be immer-ized

@SmittieC I have only changes where nested copying was present.

assets/javascript/apps/pipeline/PipelineNode.tsx
barora05: EditPanel also has this. Missed it in widgets.tsx. Added it

snopoke: Just following up on this, when would params not exist? And if they don't exist then this update is not going to anything right?

If there is a case where data or params is empty then we should likely set them but I don't think that case can happen. 

I think my core question is what made you add the if statement in the first place?

barora05: Added it for safety. I have not checked if there can be a case of params being empty. But if it is then params[name] might thrown a runtime exception

snopoke: I think the challenge is that previously if there weren't `params` then it would set them where as now it won't. I think it's safe to assume that `n.data.params` will exist based on the data model of `Node` and [`NodeData`](https://github.com/dimagi/open-chat-studio/blob/b7c3b91c85fdc2ef8bbd1c76031caf8c417c0f31/assets/javascript/apps/pipeline/types/nodeParams.ts#L37-L42)

barora05: got it. removed the null check

snopoke (approved):

PR #1433, opened 2025-04-09
add script to filter logs

snopoke:
## Description
Add a script for filtering WAF logs.

(written by Claude with guidance and tweaks from me)


SmittieC (commented):
What's the context for this script? (Not sure if you need an approval, but you can merge it if you need to. I'm fine with the changes)

PR #1436, opened 2025-04-10
Fix: Set team context when creating tags

### LINKED ISSUE #1196: CustomTaggedItem.DoesNotExist
See this [sentry error](https://dimagi.sentry.io/issues/5614046360/?project=4505001320316928&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=24h&stream_index=0). Looks like it's not preventing tags from being created, but somehow we're getting this issue.

SmittieC:
## Description

Resolves #1196 

I was not able to reproduce the issue, so these changes are hopefully going to fix the issue. From Sentry we don't get a full stack trace, so I'm not exactly sure where to make the changes.

I moved the `current_team` context manager into the base channel class where we call `new_user_message` so that each caller doesn't have to do that (the API channel didn't).

I also wrapped the code in `ad_hoc_bot_message` to set the manager, so callers doesn't have to do that. If they do, no worries, we reset the team context to the previous value when the context manager exits.

## User Impact

N/A

### Demo

N/A

### Docs and Changelog

N/A

### DIFF
--- a/apps/service_providers/llm_service/history_managers.py
+++ b/apps/service_providers/llm_service/history_managers.py
@@ -10,6 +10,7 @@ from apps.chat.conversation import compress_chat_history, compress_pipeline_chat
 from apps.chat.models import ChatMessage, ChatMessageType
 from apps.experiments.models import Experiment, ExperimentSession
 from apps.pipelines.models import PipelineChatHistory, PipelineChatHistoryTypes
+from apps.teams.utils import set_current_team
 
 
 class BaseHistoryManager(metaclass=ABCMeta):
@@ -123,13 +124,14 @@ class ExperimentHistoryManager(BaseHistoryManager):
         )
 
         if experiment_tag:
-            tag, _ = Tag.objects.get_or_create(
-                name=experiment_tag,
-                team=self.session.team,
-                is_system_tag=True,
-                category=TagCategories.BOT_RESPONSE,
-            )
-            chat_message.add_tag(tag, team=self.session.team, added_by=None)
+            with set_current_team(self.session.team):
+                tag, _ = Tag.objects.get_or_create(
+                    name=experiment_tag,
+                    team=self.session.team,
+                    is_system_tag=True,
+                    category=TagCategories.BOT_RESPONSE,
+                )
+                chat_message.add_tag(tag, team=self.session.team, added_by=None)
 
         if type_ == ChatMessageType.AI:
             self.ai_message = chat_message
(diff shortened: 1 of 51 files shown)


snopoke (approved):

apps/service_providers/llm_service/history_managers.py
```diff
@@ -123,13 +124,14 @@ def save_message_to_history(
...
 
         if experiment_tag:
-            tag, _ = Tag.objects.get_or_create(
-                name=experiment_tag,
-                team=self.session.team,
-                is_system_tag=True,
-                category=TagCategories.BOT_RESPONSE,
-            )
-            chat_message.add_tag(tag, team=self.session.team, added_by=None)
+            with current_team(self.session.team):
```
snopoke, line 127: This should already be set at a higher level I think.

Also I just realized that if the current team IS already set then this will unset it after the context manager exits. Maybe we should update 'current_team' to reset back to what it was before.

Or else make sure that current team is set higher up the chain.

snopoke (changes requested):

apps/service_providers/llm_service/history_managers.py
SmittieC: Good points. I'll have a looksy

snopoke (dismissed):

apps/teams/utils.py
```diff
@@ -30,24 +30,32 @@ def set_current_team(team):
         get_current_team(team)
     ```
     """
-    _context.set(team)
+    token = _context.set(team)
```
snopoke, line 33: I wonder if we should validate here that the new team is the same as the current team in the case where there is already a team set. 

I can't think of a scenario in the current code base where we would be switching teams within a single context.

SmittieC: https://github.com/dimagi/open-chat-studio/pull/1436/commits/5b9be13b4753b4b6ef005bec06c96cb0d25e7c79

snopoke (approved):

SmittieC:
> This broke some tests:
> 
> ```
> FAILED apps/chatbots/tests/test_chatbot_forms.py::test_chatbot_form_pipeline_creation - ValueError: Cannot set a different team in the current context
> FAILED apps/chatbots/tests/test_chatbot_views.py::test_chatbot_experiment_table_view - ValueError: Cannot set a different team in the current context
> FAILED apps/teams/tests/test_view_mixins.py::TeamMixinTest::test_anonymous_user_redirect_to_login - ValueError: Cannot set a different team in the current context
> FAILED apps/teams/tests/test_view_mixins.py::TeamMixinTest::test_member_view_logged_in - ValueError: Cannot set a different team in the current context
> ```
> 
> Perhaps we should have a pytest fixture that resets the team context before every test?

Yeah that's a good idea

snopoke (approved):

PR #1442, opened 2025-04-10
Tracing refactor part II

snopoke:

### DIFF
--- a/apps/experiments/export.py
+++ b/apps/experiments/export.py
@@ -71,6 +71,7 @@ def filtered_export_to_csv(experiment, sessions_queryset):
 
     for session in queryset:
         for message in session.chat.messages.all():
+            # TODO: fix trace info
             trace_id = message.trace_info.get("trace_id", "") if message.trace_info else ""
             row = [
                 message.id,
--- a/apps/service_providers/llm_service/history_managers.py
+++ b/apps/service_providers/llm_service/history_managers.py
@@ -10,6 +10,7 @@ from apps.chat.conversation import compress_chat_history, compress_pipeline_chat
 from apps.chat.models import ChatMessage, ChatMessageType
 from apps.experiments.models import Experiment, ExperimentSession
 from apps.pipelines.models import PipelineChatHistory, PipelineChatHistoryTypes
+from apps.service_providers.tracing import TracingService
 
 
 class BaseHistoryManager(metaclass=ABCMeta):
@@ -35,7 +36,7 @@ class ExperimentHistoryManager(BaseHistoryManager):
     def __init__(
         self,
         session: ExperimentSession,
-        experiment: Experiment | None = None,
+        experiment: Experiment,
         max_token_limit: int | None = None,
         chat_model: BaseChatModel | None = None,
         trace_service=None,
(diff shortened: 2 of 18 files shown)


snopoke:
@coderabbit review

snopoke:
@coderabbit resolve

apps/service_providers/llm_service/history_managers.py
```diff
@@ -64,12 +65,12 @@ def for_llm_chat(
             experiment=experiment,
             max_token_limit=experiment.max_token_limit,
             chat_model=experiment.get_chat_model(),
-            trace_service=trace_service or experiment.trace_service,
+            trace_service=trace_service,
```
SmittieC, line 68: The param list indicates that `trace_service` might be `None`, but this is not the case anymore, since it will only be a passthrough when the underlying trace service is `None`?

apps/service_providers/llm_service/runnables.py
```diff
@@ -50,7 +50,9 @@ def create_experiment_runnable(
     """Create an experiment runnable based on the experiment configuration."""
 
     if assistant := experiment.assistant:
-        history_manager = ExperimentHistoryManager.for_assistant(session=session, experiment=experiment)
+        history_manager = ExperimentHistoryManager.for_assistant(
```
SmittieC, line 53: Need to update the function signature so that `trace_service` is not `None` by default

apps/experiments/export.py
```diff
@@ -93,3 +94,21 @@ def filtered_export_to_csv(experiment, sessions_queryset):
             ]
             writer.writerow(row)
     return csv_in_memory
+
+
+def _get_trace_info_for_export(message):
```
SmittieC, line 99: Nit: Rename to `_get_trace_id_for_export` since this returns not the full trace info, but the `trace_id`.

SmittieC (commented):
Nice. Just one or two comments

apps/service_providers/tracing/service.py
```diff
@@ -0,0 +1,204 @@
...
+
+        callbacks = []
+        for tracer in self._active_tracers:
+            callback = tracer.get_langchain_callback()
+            if callback:
+                callbacks.append(wrap_callback(callback))
+
+        return callbacks
+
+    def get_langchain_config(self, *, callbacks: list = None, configurable: dict = None) -> RunnableConfig:
```
SmittieC, line 126: Just curious: When we implement our own tracers, we'd have to implement `get_langchain_callback`. How do you envision this would be used by ocs tracers?

apps/service_providers/llm_service/history_managers.py
snopoke: It can be None in tests though we could change that.

apps/service_providers/llm_service/runnables.py
snopoke: It is None in tests

apps/service_providers/tracing/service.py
snopoke: It would be similar to how we currently use the `PipelineLoggingCallbackHandler`

apps/service_providers/llm_service/runnables.py
snopoke: It should never be None in prod code - I can update the tests so that it is always required

snopoke: updated

SmittieC: Thanks, and sorry for being so pedantic. I like the philosophy of not having code written in a certain way to accomodate tests...as far as reasonably possible of course. By all means do hold me to this standard as well, should we choose to go with it

SmittieC (approved):
Super nice. This will make the tracing code lovely to work with

PR #1434, opened 2025-04-09
RouterOutput: pydantic case sensitivty

stephherbers:
## Description

Sentry Issue: [OPEN-CHAT-STUDIO-TQ](https://dimagi.sentry.io/issues/6531618931/?referrer=github_integration)

## User Impact

fixes pipelines with routers

### Demo

n/a

### Docs and Changelog

changelog yes

PR #1443, opened 2025-04-10
Restore Failing Test

stephherbers:
## Description

followup to [P2 PR](https://github.com/dimagi/open-chat-studio/pull/1434) where this test kept failing so it was removed to unblock the deploy

## User Impact

This has been tested locally on a pipeline with router nodes with uppercase, lowercased, and mixedcase keywords

### Demo

N/A

### Docs and Changelog

no

### DIFF
--- a/apps/pipelines/nodes/nodes.py
+++ b/apps/pipelines/nodes/nodes.py
@@ -425,11 +425,11 @@ class RouterMixin(BaseModel):
         return value
 
     def _get_keyword(self, result: str):
-        keyword = result.lower().strip()
-        if keyword in [k.lower() for k in self.keywords]:
-            return keyword.lower()
-        else:
-            return self.keywords[0].lower()
+        input_keyword = result.lower().strip()
+        for keyword in self.keywords:
+            if keyword.lower() == input_keyword:
+                return keyword
+        return self.keywords[0]
 
     def _create_router_schema(self):
         """Create a Pydantic model for structured router output"""
@@ -486,6 +486,8 @@ class RouterNode(RouterMixin, Passthrough, HistoryMixin):
 
         if not keyword:
             keyword = self.keywords[0]
+        else:
+            keyword = self._get_keyword(keyword)
 
         if session:
             self._save_history(session, node_id, node_input, keyword)
--- a/apps/pipelines/tests/test_runnable_builder.py
+++ b/apps/pipelines/tests/test_runnable_builder.py
@@ -26,6 +26,7 @@ from apps.pipelines.tests.utils import (
     llm_response_with_prompt_node,
     passthrough_node,
     render_template_node,
+    router_node,
     start_node,
     state_key_router_node,
 )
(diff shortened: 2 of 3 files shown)


SmittieC (approved):

apps/pipelines/nodes/nodes.py
```diff
@@ -486,6 +486,8 @@ def _process_conditional(self, state: PipelineState, node_id=None):
 
         if not keyword:
             keyword = self.keywords[0]
+        else:
+            keyword = self._get_keyword(keyword)
```
snopoke, line 490: I don't think this is necessary since the pydantic validation will already have verified that the keyword is one from the list specified so we should just be able to return it as is.

apps/pipelines/tests/test_runnable_builder.py
```diff
@@ -1026,3 +1027,80 @@ def test_input_with_format_strings():
...
+        },
+        {
+            "id": "D -> END",
+            "source": template_d["id"],
+            "target": end["id"],
+        },
+    ]
+    runnable = create_runnable(pipeline, nodes, edges)
+
+    output = runnable.invoke(PipelineState(messages=["a"], experiment_session=experiment_session))
```
snopoke, line 1095: This case should really results in a Pydantic validation error when calling `with_structured_output` since the case is incorrect (same with the 'z' case).

I think that's the root of the test issue - our mock `with_structured_output` isn't behaving the same as langchains one.

apps/pipelines/nodes/nodes.py
```diff
@@ -425,11 +425,11 @@ def ensure_keywords_exist(cls, value, info: FieldValidationInfo):
...
-        keyword = result.lower().strip()
-        if keyword in [k.lower() for k in self.keywords]:
-            return keyword.lower()
-        else:
-            return self.keywords[0].lower()
+        input_keyword = result.lower().strip()
+        for keyword in self.keywords:
+            if keyword.lower() == input_keyword:
+                return keyword
+        return self.keywords[0]
```
snopoke, line 432: I think fixes the static router which became case sensitive after the previous change. I think static routers should also be case sensitive but open to others thoughts.

Here's a test that would be good to add for it:


```python

@django_db_with_data(available_apps=("apps.service_providers",))
@mock.patch("apps.pipelines.nodes.base.PipelineNode.logger", mock.Mock())
def test_static_router_case_sensitive(pipeline, experiment_session):
    start = start_node()
    router = state_key_router_node("route_to", ["first", "SECOND"], data_source=StaticRouterNode.DataSource.temp_state)
    template_a = render_template_node("A")
    template_b = render_template_node("B")
    end = end_node()
    nodes = [start, router, template_a, template_b, end]
    edges = [
        {"id": "start -> code", "source": start["id"], "target": router["id"]},
        {
            "id": "router -> A",
            "source": router["id"],
            "target": template_a["id"],
            "sourceHandle": "output_0",
        },
        {
            "id": "router -> B",
            "source": router["id"],
            "target": template_b["id"],
            "sourceHandle": "output_1",
        },
        {"id": "A -> end", "source": template_a["id"], "target": end["id"]},
        {"id": "B -> end", "source": template_b["id"], "target": end["id"]},
    ]
    runnable = create_runnable(pipeline, nodes, edges)
    output = runnable.invoke(PipelineState(messages=[""], experiment_session=experiment_session, temp_state={"route_to": "SECOND"}))
    assert output["messages"][-1] == "B"
```

snopoke: I've made a fix for this here: https://github.com/dimagi/open-chat-studio/pull/1448

stephherbers:
closed in favor of https://github.com/dimagi/open-chat-studio/pull/1448/

PR #1444, opened 2025-04-10
Add additional pruned_memory check

stephherbers:
## Description

Sentry Issue: [OPEN-CHAT-STUDIO-TY](https://dimagi.sentry.io/issues/6535116621/?referrer=github_integration)
Theory: 

The error shows pruned_memory is `[]`
The error happens at line 334 where it raises ChatException("Unable to compress history")
This only happens when context["new_lines"] is empty and first_call is `False`

This suggests the code execution path might be:

`_get_new_summary` is called with empty pruned_memory and first_call=False
The early return check if not pruned_memory: is somehow bypassed
The code calls _get_summary_tokens_with_context which returns empty new_lines
Then it enters the condition if not context["new_lines"]: and raises the exception

idea: add extra pruned_memory check after modification during first call

*untested unsure how to test this edge case specifically

## User Impact

fixes case so conversation can continue

### Demo

n/a

### Docs and Changelog

changelog yes

### DIFF
--- a/apps/chat/conversation.py
+++ b/apps/chat/conversation.py
@@ -320,6 +320,8 @@ def _get_new_summary(llm, pruned_memory, summary, max_token_limit, first_call=Tr
             for msg in pruned_memory
             if msg.content
         ]
+        if not pruned_memory:
+            return summary
 
     tokens, context = _get_summary_tokens_with_context(llm, summary, pruned_memory)
     next_batch = []


barora05:
Ideally, the pruned memory check at the start of the function should have handled this case. I'm not sure if this will work

snopoke:
Closed in favour of https://github.com/dimagi/open-chat-studio/pull/1445

Issue #1143, opened 2025-02-05
[design] Allow nodes in a pipeline to create an OCS tag


bderenzi:
# Task
1. Add a checkbox option to all nodes: "Tag output message"
    - This should have the following help text: "Add a [tag](https://docs.openchatstudio.com/concepts/tags/) to the generated message with this node's name."
2. For router nodes, both llm and static, add an additional checkbox: "Tag output message with route". The help text for this should be similar to above: "Add a tag to the generated message with the route keyword selected by this node."

## Original notes

some more context [here](https://dimagi.slack.com/archives/C05PK63Q89H/p1738698576222059)

I can imagine this working in a number of ways, including:
* Allowing LLM routers to save a route as a tag
* Allowing individual LLM nodes to tag a message with their node name
* Allowing any node to create add an OCS tag to the current pipeline execution?

I'm not sure what's best (and imagine there are others). I think the router saving routes as a tag would replicate the previous functionality

snopoke:
@barora05 for the first part the task is to add text input where the user can type a tag value. If the value is present then a tag will be added to the output message (with a blank tag category).

PR #1450, opened 2025-04-14
Ba/add route tags

barora05:
## Description
Added a checbox widget.
Completed 2nd pt of #1143 i.e
For router nodes, both llm and static, add an additional checkbox: "Tag output message with route". The help text for this should be similar to above: "Add a tag to the generated message with the route keyword selected by this node."

**BugFix**
Fixed an issue with the Toggle widget where it couldn't be unchecked once selected.
### Demo
image

### DIFF
--- a/apps/pipelines/nodes/base.py
+++ b/apps/pipelines/nodes/base.py
@@ -79,6 +79,7 @@ class PipelineState(dict):
     input_message_metadata: Annotated[dict, merge_dicts]
     output_message_metadata: Annotated[dict, merge_dicts]
     attachments: list = Field(default=[])
+    tag_output_message: Annotated[dict, merge_dicts]
 
     def json_safe(self):
         # We need to make a copy of `self` so as to not change the actual value of `experiment_session` forever
@@ -183,7 +184,7 @@ class PipelineNode(BaseModel, ABC):
         output_handle = next((k for k, v in output_map.items() if v == conditional_branch), None)
         state["outputs"][node_id]["output_handle"] = output_handle
         if self.tag_output_message:
-            state["output_message_metadata"].setdefault("tags", []).append(conditional_branch)
+            state["tag_output_message"].setdefault("tags", []).append(f"{self.name}:{conditional_branch}")
         return conditional_branch
 
     def _process(self, input: str, state: PipelineState, node_id: str) -> PipelineState:
--- a/apps/pipelines/nodes/base.py
+++ b/apps/pipelines/nodes/base.py
@@ -223,6 +223,7 @@ class Widgets(StrEnum):
     float = "float"
     range = "range"
     multiselect = "multiselect"
+    checkbox = "checkbox"
     none = "none"
 
     # special widgets
--- a/apps/pipelines/nodes/base.py
+++ b/apps/pipelines/nodes/base.py
@@ -182,6 +182,9 @@ class PipelineNode(BaseModel, ABC):
         output_map = self.get_output_map()
         output_handle = next((k for k, v in output_map.items() if v == conditional_branch), None)
         state["outputs"][node_id]["output_handle"] = output_handle
+        if self.tag_output_message:
+            state["output_message_metadata"].setdefault("tags", []).append(conditional_branch)
+        breakpoint()
         return conditional_branch
 
     def _process(self, input: str, state: PipelineState, node_id: str) -> PipelineState:
--- a/apps/pipelines/nodes/base.py
+++ b/apps/pipelines/nodes/base.py
@@ -184,7 +184,6 @@ class PipelineNode(BaseModel, ABC):
         state["outputs"][node_id]["output_handle"] = output_handle
         if self.tag_output_message:
             state["output_message_metadata"].setdefault("tags", []).append(conditional_branch)
-        breakpoint()
         return conditional_branch
 
     def _process(self, input: str, state: PipelineState, node_id: str) -> PipelineState:
@@ -226,7 +225,6 @@ class Widgets(StrEnum):
     float = "float"
     range = "range"
     multiselect = "multiselect"
-    checkbox = "checkbox"
     none = "none"
 
     # special widgets
--- a/apps/pipelines/nodes/base.py
+++ b/apps/pipelines/nodes/base.py
@@ -79,7 +79,7 @@ class PipelineState(dict):
     input_message_metadata: Annotated[dict, merge_dicts]
     output_message_metadata: Annotated[dict, merge_dicts]
     attachments: list = Field(default=[])
-    tag_output_message: Annotated[dict, merge_dicts]
+    output_message_tags: Annotated[list[str], operator.add]
 
     def json_safe(self):
         # We need to make a copy of `self` so as to not change the actual value of `experiment_session` forever
@@ -184,7 +184,8 @@ class PipelineNode(BaseModel, ABC):
         output_handle = next((k for k, v in output_map.items() if v == conditional_branch), None)
         state["outputs"][node_id]["output_handle"] = output_handle
         if self.tag_output_message:
-            state["tag_output_message"].setdefault("tags", []).append(f"{self.name}:{conditional_branch}")
+            state.setdefault("output_message_tags", [])
+            state["output_message_tags"].append(f"{self.name}:{conditional_branch}")
         return conditional_branch
 
     def _process(self, input: str, state: PipelineState, node_id: str) -> PipelineState:
--- a/apps/pipelines/nodes/nodes.py
+++ b/apps/pipelines/nodes/nodes.py
@@ -418,10 +418,10 @@ class BooleanNode(Passthrough):
 
 class RouterMixin(BaseModel):
     keywords: list[str] = Field(default_factory=list, json_schema_extra=UiSchema(widget=Widgets.keywords))
-    tag_message: str = Field(
-        default="False",
+    tag_message: bool = Field(
+        default=False,
         description="Tag output message with route",
-        json_schema_extra=UiSchema(widget=Widgets.toggle)
+        json_schema_extra=UiSchema(widget=Widgets.checkbox)
     )
 
     @field_validator("keywords")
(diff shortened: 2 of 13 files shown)


apps/pipelines/nodes/nodes.py
```diff
@@ -413,6 +413,9 @@ def get_output_map(self):
 
 class RouterMixin(BaseModel):
     keywords: list[str] = Field(default_factory=list, json_schema_extra=UiSchema(widget=Widgets.keywords))
+    tag_output_message: bool = Field(
+        default=False, description="Tag output message with route", json_schema_extra=UiSchema(widget=Widgets.checkbox)
```
SmittieC, line 417: Maybe elaborate a little bit more? "Tag the output message with chosen route" or something like that

apps/pipelines/nodes/nodes.py
```diff
@@ -476,9 +486,10 @@ def _process_conditional(self, state: PipelineState, node_id=None):
             keyword = getattr(result, "route", None)
         except ValidationError:
             keyword = None
-
         if not keyword:
             keyword = self.keywords[0]
+        if self.tag_output_message:
```
SmittieC, line 491: You would be able to move this logic [here](https://github.com/dimagi/open-chat-studio/blob/6c1437ec90b24e2c21b737af2a2a9ca7a2dd96d9/apps/pipelines/nodes/base.py#L180-L185) and only have to do this once?

apps/service_providers/utils.py
```diff
@@ -152,8 +152,6 @@ def get_first_llm_provider_model(llm_provider, team_id):
     try:
         if llm_provider:
             model = LlmProviderModel.objects.filter(type=llm_provider.type, team_id=team_id).order_by("id").first()
-            breakpoint()
```
SmittieC, line 155: :1st_place_medal:

assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -38,6 +38,8 @@ export function getWidget(name: string, params: PropertySchema) {
       return KeywordsWidget
     case "node_name":
       return NodeNameWidget
+    case "checkbox":
```
SmittieC, line 41: See https://github.com/dimagi/open-chat-studio/pull/1450/files#r2042148019

SmittieC (commented):
Looking good. I think a test to ensure that the output message is indeed tagged would be nice as well

apps/pipelines/nodes/base.py
```diff
@@ -223,6 +223,7 @@ class Widgets(StrEnum):
     float = "float"
     range = "range"
     multiselect = "multiselect"
+    checkbox = "checkbox"
```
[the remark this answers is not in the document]
barora05: Can we use toggle in this case and keep checkbox code also for any future use?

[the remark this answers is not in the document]
SmittieC: The checkbox and toggle code is mostly the same, with the only difference being how the checkbox is ultimately rendered. I doubt we'll want to use the checkbox if we have a toggle, since 1. they do the same thing and 2. this would create a discrepancy in the UI. Maybe in the future if we want the traditional checkbox UI instead of the toggle, we can just swap out the `toggle` classname for `checkbox` and/or update all references from toggle to checkbox, but for now, I think let's use a toggle and remove the checkbox code.

apps/pipelines/nodes/base.py
```diff
@@ -182,6 +182,9 @@ def process_conditional(self, state: PipelineState, node_id: str | None = None)
         output_map = self.get_output_map()
         output_handle = next((k for k, v in output_map.items() if v == conditional_branch), None)
         state["outputs"][node_id]["output_handle"] = output_handle
+        if self.tag_output_message:
+            state["output_message_metadata"].setdefault("tags", []).append(conditional_branch)
+        breakpoint()
```
SmittieC, line 187: Just flagging the breakpoint

apps/pipelines/nodes/nodes.py
```diff
@@ -538,11 +535,10 @@ def _process_conditional(self, state: PipelineState, node_id=None):
             result = ""
 
         result_lower = result.lower()
+        breakpoint()
```
SmittieC, line 538: This one as well

barora05: sure. @SmittieC I am trying to write a test. but in test the conditional_branch returns prompt + tags whereas when I am actually trying conditional_branch only retruns the keyword i.e tag.

apps/pipelines/nodes/nodes.py
barora05: @SmittieC I have done this. Tests are failing. Should I add tag_outpute_message in Boolean Node as well?

SmittieC: The boolean node is deprecated, so I'm thinking it's not necessary

apps/pipelines/nodes/nodes.py
```diff
@@ -413,6 +413,11 @@ def get_output_map(self):
 
 class RouterMixin(BaseModel):
     keywords: list[str] = Field(default_factory=list, json_schema_extra=UiSchema(widget=Widgets.keywords))
+    tag_output_message: bool = Field(
+        default=False,
+        description="Tag the output message with chosen route",
```
snopoke, line 418: nit:
```suggestion
        description="Tag the output message with the selected route",
```

apps/pipelines/nodes/base.py
```diff
@@ -182,6 +182,8 @@ def process_conditional(self, state: PipelineState, node_id: str | None = None)
         output_map = self.get_output_map()
         output_handle = next((k for k, v in output_map.items() if v == conditional_branch), None)
         state["outputs"][node_id]["output_handle"] = output_handle
+        if self.tag_output_message:
+            state["output_message_metadata"].setdefault("tags", []).append(conditional_branch)
```
snopoke, line 186: I think we should use a separate data structure for tags and metadata. Also the tag name should include the node name:

```suggestion
            state["output_message_tags"].setdefault("tags", []).append(f"{self.name}:{conditional_branch}")
```

snopoke (commented):
The tags still need to be saved to the message. This should be done in [`Pipeline.invoke`](https://github.com/dimagi/open-chat-studio/blob/main/apps/pipelines/models.py#L347-L349). You can look at the code in the history managers as a reference: https://github.com/dimagi/open-chat-studio/blob/main/apps/service_providers/llm_service/history_managers.py#L126-L138

barora05:
image
The tag is saved like this currently

snopoke (commented):
Few small tweaks but otherwise looks good. 

Please also add a test for the tagging.

apps/service_providers/utils.py
SmittieC: FYI https://github.com/dimagi/open-chat-studio/pull/1469/files

apps/pipelines/tests/test_pipeline_runs.py
```diff
@@ -183,3 +186,18 @@ def test_save_trace_metadata(pipeline: Pipeline, session: ExperimentSession):
     assert "trace_info" in human_message.metadata
     ai_message = session.chat.messages.filter(message_type=ChatMessageType.AI).first()
     assert "trace_info" in ai_message.metadata
+
+
+@pytest.mark.django_db()
+def test_save_metadata_and_tagging(pipeline: Pipeline, session: ExperimentSession):
```
snopoke, line 192: You should also test that the router output set's the tag's correctly. You could use https://github.com/dimagi/open-chat-studio/blob/main/apps/pipelines/tests/test_runnable_builder.py#L383C5-L383C38 as an example

barora05: @snopoke Tried adding a test for this multiple times like the example test but the test is failing.

```
@pytest.mark.django_db()
def test_router_sets_tags_correctly(pipeline: Pipeline, session: ExperimentSession):
    start = start_node()
    router = state_key_router_node(
        "route_to", ["first", "second", "third"], data_source=StaticRouterNode.DataSource.temp_state
    )
    template_a = render_template_node("A")
    template_b = render_template_node("B")
    template_c = render_template_node("C")
    end = end_node()

    nodes = [start, router, template_a, template_b, template_c, end]
    edges = [
        {"id": "start -> router", "source": start["id"], "target": router["id"]},
        {
            "id": "router -> A",
            "source": router["id"],
            "target": template_a["id"],
            "sourceHandle": "output_0",
        },
        {
            "id": "router -> B",
            "source": router["id"],
            "target": template_b["id"],
            "sourceHandle": "output_1",
        },
        {
            "id": "router -> C",
            "source": router["id"],
            "target": template_c["id"],
            "sourceHandle": "output_2",
        },
        {"id": "A -> end", "source": template_a["id"], "target": end["id"]},
        {"id": "B -> end", "source": template_b["id"], "target": end["id"]},
        {"id": "C -> end", "source": template_c["id"], "target": end["id"]},
    ]
    runnable = create_runnable(pipeline, nodes, edges)

    def _check_routing_and_tags(route_to, expected_tag):
        with mock.patch.object(ChatMessage, "add_system_tag") as mock_add_system_tag:
            runnable.invoke(
                PipelineState(messages=["Test message"], experiment_session=session, temp_state={"route_to": route_to}, tag_output_message= True)
            )
            mock_add_system_tag.assert_any_call(f"route_{route_to.lower()}", TagCategories.BOT_RESPONSE)
            assert mock_add_system_tag.call_count >= 1
    _check_routing_and_tags("first", "route_first")
    _check_routing_and_tags("second", "route_second")
    _check_routing_and_tags("third", "route_third")
```

snopoke: I've added a test

snopoke (approved):

snopoke:
@barora05 please add a changelog entry for this

barora05:
> @barora05 please add a changelog entry for this

https://github.com/dimagi/open-chat-studio-docs/pull/93

PR #1458, opened 2025-04-15
Add back in Router Node Test

stephherbers:
## Description

This is the current state that is passing. There has been so much back and forth between the test and the `FakeLlm` Class that I realize that `PydanticToolsParser` is not even used at this current state


snopoke:
Replaced by https://github.com/dimagi/open-chat-studio/pull/1471

Issue #1462, opened 2025-04-16
web chat polling 403


snopoke:
Anonymous chats in on the web are showing 403 errors when the intermittent polling requests are made:

![Image](https://github.com/user-attachments/assets/ac955267-b856-4629-80ac-b5720f2b8fc2)

There were recent changes to this code: https://github.com/dimagi/open-chat-studio/pull/1439

Issue #1463, opened 2025-04-16
Security improvements to participants


snopoke:
There are some issues with how we create / select participants for the web channel. The following list shows the set of variables that need to be considered:

* User logged in or not
* Consent enabled or not
    * Consent set to capture identifier
    * Consent type set to 'email'
* Identifier matches existing participant or not

For the web channel specifically there are issues caused by using a single field for the identifier when in reality there are three of identifier for web participants:

* Email address
* UUID (random)
* User provided text

In most cases we treat these the same however we should not since they have different security requirements.

## Recommendations
* Add a 'type' field to the `Participant` model which captures the identifier type: `email`, `random`, `other`. This will be blank for all platform types except the web platform.
* When loading participants for web sessions, filter by the identifier type.
* Add a 'verified' field to the `Participant` model to track whether email identifiers have been verified or not.
* Update the flow for email consent so that verification happens before the session is created not after.
    * See https://github.com/dimagi/open-chat-studio/blob/05306a50a211c9f938a97da9f07b1f9d66061704/apps/experiments/views/experiment.py#L907-L919
* For participants linked to logged in OCS users the identifier type will be `email` and `verified` will be `True`.

## Test Scenario
(This scenario was revealed during debugging for #1462)

#### Setup

* An bot exists with consent enabled
* The consent is set to capture the participant identifier
* The identifier type is set to `text`

* A participant already exists for an OCS user on the team (identifier = `test@test.com`)

#### Scenario steps

* The bot's public link is accessed either in an incognito window or a browser where the user is not logged in to OCS. 
* The user enters `test@test.com` as the identifier on the consent page.
    * Because the consent type is set to 'text' there is no verification flow
* A session is created and linked to the participant `test@test.com` for that team.

### Problem

The participant that was linked to the session is the same one that is attached to the OCS user event though the bot was accessed in a browser that wasn't logged in to OCS. This is because the identifier's matched.

snopoke:
Closing this after reaching a decision to remove the public chat UI entirely.

PR #1464, opened 2025-04-16
check user before comparison

snopoke:
## Description
Resolves https://github.com/dimagi/open-chat-studio/issues/1462

This is a band aid fix. I think more work should be done on the participant identifier validation to prevent further issues and improve security.

More details in https://github.com/dimagi/open-chat-studio/issues/1463


SmittieC (approved):
General question: Should we not use the `verify_session_access_cookie` decorator here?

snopoke:
> General question: Should we not use the `verify_session_access_cookie` decorator here?

probably

PR #1465, opened 2025-04-16
load based on hash id

### LINKED ISSUE #1457: Creating a new version should redirect to versions tab
After submitting the 'create new version' form the browser should be redirected back to the experiment / chatbot details page with the versions tab active.

I think this was broken by some of the session filter widget changes. [This line](https://github.com/dimagi/open-chat-studio/blob/053198912057b330208d91511d007d98970aa1dc/templates/experiments/single_experiment_home.html#L324) in particular looks suspect.

I wasn't able to test this after submitting the form the 'create new version' button in the versions table is meant to indicate that the version is being created (See https://github.com/dimagi/open-chat-studio/blob/081a1d308b7726f09d088ed7dfdd6a5ee0119ae9/templates/experiments/create_version_button.html)

barora05:
## Description
Try to load the tab based on hash, if not present then default to allsessions. Closes #1457

### DIFF
--- a/templates/experiments/single_experiment_home.html
+++ b/templates/experiments/single_experiment_home.html
@@ -321,7 +321,16 @@
     }
 
     document.addEventListener('DOMContentLoaded', () => {
-      showTabContent('allsessions');
+        // Initially show the tab based on the hash, if present, otherwise default to allsessions
+        const initialTabIdFromHash = window.location.hash.replace('#', '');
+        if (initialTabIdFromHash && document.getElementById(`tab-${initialTabIdFromHash}`)) {
+          document.getElementById(`tab-${initialTabIdFromHash}`).checked = true;
+          showTabContent(initialTabIdFromHash);
+        } else {
+          document.getElementById('tab-allsessions').checked = true;
+          showTabContent('allsessions');
+        }
+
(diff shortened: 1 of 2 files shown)


templates/experiments/single_experiment_home.html
```diff
@@ -321,18 +321,24 @@ <h3 class="font-bold text-lg">Parent Experiments:
...
-      showTabContent('allsessions');
+        // Initially show the tab based on the hash, if present, otherwise default to allsessions
+      const initialTabIdFromHash = window.location.hash.replace('#', '');
+      if (initialTabIdFromHash && document.getElementById(`tab-${initialTabIdFromHash}`)) {
+        document.getElementById(`tab-${initialTabIdFromHash}`).checked = true;
+        showTabContent(initialTabIdFromHash);
+      } else {
+        document.getElementById('tab-allsessions').checked = true;
+        showTabContent('allsessions');
+      }
```
snopoke, line 332: I think this could be simplified to:

```
if (window.location.hash) {
    selectTabFromHash();
} else {
    showTabContent('allsessions');
}

snopoke (approved):

PR #1466, opened 2025-04-16
remove sorting from some params

### LINKED ISSUE #1456: Remove sorting for session table columns
Sorting of the session table should only be possible using the 'last message' and 'participant' columns.

Attempts to sort by the other columns result in 500 errors ([OPEN-CHAT-STUDIO-V6](https://dimagi.sentry.io/issues/6543211069/events/22597ad8bb5f4400bc193f24cee5c921/))

barora05:
## Description
Remove sorting from tags, actions, versions in session table. Closes #1456 
### Demo
image

### DIFF
--- a/apps/chatbots/tables.py
+++ b/apps/chatbots/tables.py
@@ -61,4 +61,5 @@ class ChatbotSessionsTable(ExperimentSessionsTable):
             ),
         ],
         align="right",
+        orderable=False
     )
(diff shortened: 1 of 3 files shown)


SmittieC (commented):
We'll need to order by participant also, otherwise looks good to me

snopoke (approved):

barora05:
> We'll need to order by participant also, otherwise looks good to me

We are ordering by participants and last message.

PR #1467, opened 2025-04-16
unwrap lazy objects before setting in context

snopoke:
## Description
Sentry [OPEN-CHAT-STUDIO-VA](https://dimagi.sentry.io/issues/6545279442/events/c98af42251744e0790967d377ed74728/)

1. The validation should not apply when either the existing team or the new team are None
2. Unwrap `SimpleLazyObjects` to avoid any proxy issues

### DIFF
--- a/apps/teams/utils.py
+++ b/apps/teams/utils.py
@@ -32,9 +32,11 @@ def set_current_team(team) -> Token:
         get_current_team(team)
     ```
     """
-    if existing_team := get_current_team():
+    team = _unwrap_lazy(team)
+    existing_team = get_current_team()
+    if team is not None and existing_team is not None:
         if existing_team != team:
-            log.error("Trying to set a different team in the current context")
+            log.error("Trying to set a different team in the current context: %s != %s", team, existing_team)
 
     token = _context.set(team)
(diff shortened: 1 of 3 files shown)


apps/teams/utils.py
```diff
@@ -32,9 +32,11 @@ def set_current_team(team) -> Token:
         get_current_team(team)
     ```
     """
-    if existing_team := get_current_team():
+    team = _unwrap_lazy(team)
+    existing_team = get_current_team()
```
SmittieC, line 36: The existing team can be a lazy object as well. See [here](https://dimagi.sentry.io/issues/6545279442/?project=4505001320316928&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&stream_index=1)

snopoke: I can only be a lazy object if it's set to a lazy object and since we're now unwrapping lazy objects before we set them I don't think that's an issue.

SmittieC: Ah of course, you're right. I was using a single brain cell

Issue #1138, opened 2025-02-04
Expose routing info in pipeline state


snopoke:
This should be exposed via the temporary state so that python nodes can access it

stephherbers:
doc for review: https://docs.google.com/document/d/1fYLycZe_P45qv1eTd05Vh1eiPbY1FCtbttVl6IbzC5U/edit?tab=t.0

stephherbers:
FYI this is ready to go, except for tests that are blocked by the reintroduction from last week's P2 where the test had to be removed. This would have to modify that exact test

PR #1470, opened 2025-04-16
Pipeline routing info

stephherbers:
## Description

resolves [#1138](https://github.com/dimagi/open-chat-studio/issues/1138)
## User Impact

User will be able to access the routing information of the pipeline

### Demo

[Loom Video](https://www.loom.com/share/2a59c3337f3b469087536a84f7874d73?sid=b4add846-77bc-4271-a0de-a4e2d4283d29)

### Docs and Changelog

definite yes to both


snopoke:
@stephherbers this should be good to pick up again. Take note of the new field I added to the state class which might be useful: https://github.com/dimagi/open-chat-studio/blob/main/apps/pipelines/nodes/base.py#L86

apps/pipelines/nodes/base.py
```diff
@@ -248,6 +294,10 @@ def router(state: PipelineState, config: RunnableConfig) -> ReturnType:
 
             conditional_branch = self._process_conditional(state, node_id)
             output_handle = next((k for k, v in output_map.items() if v == conditional_branch), None)
+            state["outputs"][self.name] = {
+                "route": conditional_branch,
+                "output": state["node_input"],
+            }
```
snopoke, line 300: can this be moved to `PipelineState.from_router_output`

apps/pipelines/nodes/nodes.py
```diff
@@ -984,6 +984,9 @@ def _get_custom_globals(self, state: PipelineState):
                 "set_temp_state_key": self._set_temp_state_key(state),
                 "get_session_state_key": self._get_session_state_key(state["experiment_session"]),
                 "set_session_state_key": self._set_session_state_key(state["experiment_session"]),
+                "get_route": lambda node_name: state.get_route(node_name),
+                "get_node_path": lambda node_name: state.get_node_path(node_name),
+                "get_all_routes": lambda: state.get_all_routes(),
```
snopoke, line 989: I think 'state' here is just a dict - if you want access to these functions you'll need to wrap it: `state = PipelineState(**state)` (untested).

Also, you can drop the lambdas:
```suggestion
                "get_route": state.get_route,
                "get_node_path": state.get_node_path,
                "get_all_routes": state.get_all_routes,
```

apps/pipelines/nodes/base.py
```diff
@@ -108,10 +108,113 @@ def from_node_output(cls, node_name: str, node_id: str, output: Any = None, **kw
...
 
+    def _get_node_mappings(self):
+        """
+        Helper method to get node ID to name mappings.
+        Returns a tuple of (node_id_to_name, name_to_node_id) dictionaries.
+        """
+        from apps.pipelines.models import Node
+
+        pipeline_version = self.get("pipeline_version")
+        nodes = Node.objects.filter(pipeline__version_number=pipeline_version)
```
SmittieC, line 119: I'm wondering if the DB calls should not be cached, since they cannot really change during a run. The pipeline state can, but not the DB results

apps/pipelines/nodes/base.py
```diff
@@ -108,10 +108,113 @@ def from_node_output(cls, node_name: str, node_id: str, output: Any = None, **kw
...
+
+    def get_route(self, node_name: str) -> str | None:
+        """
+        Gets the route selected by a specific router node.
+        """
+        outputs = self.get("outputs", {})
+        node_id_to_name, name_to_node_id = self._get_node_mappings()
+
+        if node_name in outputs and "route" in outputs[node_name]:
+            return outputs[node_name].get("route")
```
SmittieC, line 148: Would the code after this ever be necessary? We're always saving the `route` in an object after the router is executed. If the router did not execute, no other information would be available as far as I can see? I might be wrong.

apps/pipelines/nodes/base.py
```diff
@@ -108,10 +108,113 @@ def from_node_output(cls, node_name: str, node_id: str, output: Any = None, **kw
...
     @classmethod
-    def from_router_output(cls, node_id, node_name, output, output_handle, tags, route_path) -> Self:
+    def from_router_output(
+        cls, node_id, node_name, output, output_handle, tags, route_path, conditional_branch
+    ) -> Self:
         return cls(
-            outputs={node_id: {"output_handle": output_handle, "message": output}},
+            outputs={
+                node_id: {"output_handle": output_handle, "message": output},
+                node_name: {"route": conditional_branch, "output": output},
```
SmittieC, line 216: Can't we add the route key in the node_id object rather?

apps/pipelines/nodes/base.py
```diff
@@ -108,10 +108,113 @@ def from_node_output(cls, node_name: str, node_id: str, output: Any = None, **kw
...
+        node_id_to_name, name_to_node_id = self._get_node_mappings()
+        current_node_id = name_to_node_id.get(node_name, node_name)
+
+        while current_node_id:
+            current_name = node_id_to_name.get(current_node_id, current_node_id)
+            path.insert(0, current_name)
+
+            parent_node_id = None
+            for _, current, targets in self.get("path", []):
+                if current_node_id in targets or current_node_id == targets[0]:
```
snopoke, line 180: The first condition is a more general case of the second so this can be simplified to:
```suggestion
                if current_node_id in targets:
```

apps/pipelines/tests/test_runnable_builder.py
```diff
@@ -1220,3 +1222,38 @@ def _tool_call(route):
     assert output["messages"][-1] == "D d"
     output = runnable.invoke(PipelineState(messages=["z"], experiment_session=experiment_session))
     assert output["messages"][-1] == "A z"
+
+
+@pytest.mark.django_db()
+def test_router_node_output_structure(provider, provider_model, pipeline, experiment_session):
```
SmittieC, line 1228: Would you mind adding tests for the new functions as well, with happy and sad cases?

A suggestion:
I'm envisioning tests where the pipeline's state is defined in raw JSON (this is the test input for instance). The JSON is loaded into `PipelineState`, then each of the new functions are invoked for that state. This way it will (hopefully) be super clear what the expected output of each should be, given a specific state.  Ultimately it's up to you how you want to do testing.

apps/pipelines/nodes/base.py
snopoke, line 216: I think these can be merged (if you also update this code: https://github.com/dimagi/open-chat-studio/blob/0f91709252a6657e75f4039f67ba4dbc25329c14/assets/javascript/apps/pipeline/panel/TestMessageBox.tsx#L68-L69)

```suggestion
                node_name: {"node_id": node_id, "output_handle": output_handle, "route": conditional_branch, "output": output},
```

apps/pipelines/nodes/base.py
```diff
@@ -108,10 +108,113 @@ def from_node_output(cls, node_name: str, node_id: str, output: Any = None, **kw
...
+                    parent_node_id = current
+                    break
+
+            if not parent_node_id:
+                for node, data in outputs.items():
+                    if "route" in data and data["route"] == current_node_id:
+                        parent_node_id = node
+                        break
```
snopoke, line 188: This block doesn't really make sense. `data["route"]` can't ever match here since it's never set to a node ID.

apps/pipelines/nodes/base.py
```diff
@@ -108,10 +108,113 @@ def from_node_output(cls, node_name: str, node_id: str, output: Any = None, **kw
...
+        """
+        Gets all routing decisions in the pipeline.
+        """
+        routes_dict = {}
+        outputs = self.get("outputs", {})
+        node_id_to_name, _ = self._get_node_mappings()
+
+        for node_id, node_data in outputs.items():
+            if "route" in node_data:
+                node_name = node_id_to_name.get(node_id, node_id)
```
snopoke, line 204: won't `node_id` already be a node name (according to the line below: `node_name: {"route": conditional_branch, "output": output},`

apps/pipelines/nodes/nodes.py
```diff
@@ -984,6 +985,9 @@ def _get_custom_globals(self, state: PipelineState):
                 "set_temp_state_key": self._set_temp_state_key(state),
                 "get_session_state_key": self._get_session_state_key(state["experiment_session"]),
                 "set_session_state_key": self._set_session_state_key(state["experiment_session"]),
+                "get_route": lambda node_name: pipeline_state.get_route(node_name),
+                "get_node_path": lambda node_name: pipeline_state.get_node_path(node_name),
+                "get_all_routes": lambda: pipeline_state.get_all_routes(),
```
snopoke, line 990: You can drop the lambdas:
```suggestion
                "get_route": pipeline_state.get_route,
                "get_node_path": pipeline_state.get_node_path,
                "get_all_routes": pipeline_state.get_all_routes,
```

SmittieC (commented):
Overall looks great! This might be a good time to define the expected structure of the `output` key of the `PipelineState`, instead of just being `dict`. Maybe not for this PR though

apps/pipelines/nodes/base.py
```diff
@@ -108,10 +108,113 @@ def from_node_output(cls, node_name: str, node_id: str, output: Any = None, **kw
...
+        outputs = self.get("outputs", {})
+        node_id_to_name, name_to_node_id = self._get_node_mappings()
+
+        if node_name in outputs and "route" in outputs[node_name]:
+            return outputs[node_name].get("route")
+
+        node_id = name_to_node_id.get(node_name)
+
+        if node_id and node_id in outputs and "route" in outputs[node_id]:
+            return outputs[node_id].get("route")
```
snopoke, line 153: this also doesn't makes sense because the outputs with 'route' key are keyed by node name, not node_id. Also the value of the 'route' key is not a node name, it is the route keyword.

apps/pipelines/nodes/base.py
```diff
@@ -108,10 +108,113 @@ def from_node_output(cls, node_name: str, node_id: str, output: Any = None, **kw
...
+
+        for node_identifier, data in outputs.items():
+            if node_id_to_name.get(node_identifier) == node_name and "route" in data:
+                return data.get("route")
```
snopoke, line 157: Similar comment to above, `route` isn't a node ID or node name so this doesn't make sense.

apps/pipelines/tests/test_runnable_builder.py
```diff
@@ -1220,3 +1222,38 @@ def _tool_call(route):
...
+            experiment_session=experiment_session,
+            temp_state={"user_input": "hello world", "outputs": {}},
+            path=[],
+        )
+        with mock.patch.object(node, "_process_conditional", return_value="A"):
+            node_id = "123"
+            edge_map = {"A": "next_node_a", "B": "next_node_b"}
+            incoming_edges = ["123"]
+            router_func = node.build_router_function(node_id, edge_map, incoming_edges)
+            command = router_func(state, {"metadata": {"langgraph_triggers": []}})
```
snopoke, line 1250: nit: the metadata is unnecessary
```suggestion
            command = router_func(state, {})
```

apps/pipelines/nodes/base.py
SmittieC: What will be the scope of your refactor @snopoke?

I'm feeling a bit iffy about how we currently interact with the pipeline state. Not sure how to exactly express my thoughts or what the correct course of action should be at this point, but I think we should probably get a neat API in place to interact with it sooner rather than later. Anycase, curious to hear your refactoring thoughts

snopoke: We could switch to using a pydantic model for state (https://langchain-ai.github.io/langgraph/how-tos/state-model/#how-to-use-pydantic-model-as-graph-state).

> What will be the scope of your refactor @snopoke?

I want to move all fields out of the state which aren't actually part of the runtime state. Currently this is just `experiment` and `pipeline_version`. But I think it would be useful to provide nodes with a bit more like the DB node object itself and the trace service.

assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -536,6 +536,26 @@ function CodeNodeEditor(
       detail: "Sets the given key in the session's state. Overwrites the current value",
       boost: 1
     }),
+    get_route: snip("get_route(\"${node_name}\")", {
```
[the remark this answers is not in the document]
stephherbers: thank you I made both changes-- I agree they will cause less confusion 🌟

snopoke:
@SmittieC @stephherbers I've made some additional commits here to fix things up. Let me know what you think.

stephherbers:
@snopoke thank you, Simon-- the changes you made make sense

snopoke (approved):

snopoke:
@stephherbers don't forget to update the docs for these changes

PR #1471, opened 2025-04-16
Sk/fake with structured output

snopoke:
## Description
Fix the test bench for testing structured output & restore tests removed in https://github.com/dimagi/open-chat-studio/pull/1434


SmittieC:
> I had the same 3 tests failing: test_extract_structured_data_no_chunking, test_extract_structured_data_with_chunking, and test_extract_participant_data

Looks like it's because the extract data and extract participant data nodes passes a dict to `with_structured_data`, so it returns the same. The new code checks if the schema is a subclass of a pydantic base model, which it isn't. We'll just have to pass it the class instead of its json schema

stephherbers (approved):

PR #1476, opened 2025-04-17
ensure emojis get included in uniqueness constait and fail gracefully

### LINKED ISSUE #1475: Duplicate emoji tags can be added
I noticed that while a duplicate error is thrown when a duplicate string tag is created, an duplicate emoji tag can be created which causes errors in other areas of OCS such as the filters widget

stephherbers:
## Description

resolves #1475
1) bug: duplicate emojis tags will not be added
2) nice add in: all duplicate tags will fail more gracefully in the UI

Tested on string & emoji tags for both duplicates and non-duplicates

## User Impact

A user will not be notified when tag is a duplicate rather than being shown an "oh shucks" page

### Demo

Screenshot 2025-04-17 at 1 59 42 PM

### Docs and Changelog

yes to changelog

### DIFF
--- a/apps/annotations/views/tag_views.py
+++ b/apps/annotations/views/tag_views.py
@@ -1,4 +1,5 @@
 import json
+import unicodedata
 
 from django.contrib import messages
 from django.contrib.auth.mixins import PermissionRequiredMixin
@@ -46,6 +47,19 @@ class CreateTag(CreateView, PermissionRequiredMixin):
     def form_valid(self, form):
         form.instance.team = self.request.team
         form.instance.created_by = self.request.user
+        form.instance.name = unicodedata.normalize("NFC", form.instance.name)
+
+        existing_tags = Tag.objects.filter(
+            team=self.request.team,
+            name=form.instance.name,
+            is_system_tag=getattr(form.instance, "is_system_tag", False),
+            category=getattr(form.instance, "category", ""),
+        )
+
+        if existing_tags.exists():
+            form.add_error("name", "A tag with this name already exists for this team, system status, and category.")
+            return self.form_invalid(form)
+
         return super().form_valid(form)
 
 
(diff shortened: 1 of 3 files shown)


apps/annotations/views/tag_views.py
```diff
@@ -46,6 +47,19 @@ def get_success_url(self):
     def form_valid(self, form):
         form.instance.team = self.request.team
         form.instance.created_by = self.request.user
+        form.instance.name = unicodedata.normalize("NFC", form.instance.name)
```
snopoke, line 50: This is a nice touch - what made you add it?

apps/annotations/views/tag_views.py
```diff
@@ -46,6 +47,19 @@ def get_success_url(self):
...
+            name=form.instance.name,
+            is_system_tag=getattr(form.instance, "is_system_tag", False),
+            category=getattr(form.instance, "category", ""),
+        )
+
+        if existing_tags.exists():
+            form.add_error("name", "A tag with this name already exists for this team, system status, and category.")
+            return self.form_invalid(form)
+
         return super().form_valid(form)
```
snopoke, line 63: Could you not also wrap this in a try / catch and rely on the DB constraint for the duplicate check?

stephherbers: [42ccbbd](https://github.com/dimagi/open-chat-studio/pull/1476/commits/42ccbbd8c4d3b88999c894ae2cd26ea0018fa910)

apps/annotations/views/tag_views.py
stephherbers: it caused weird bugs elsewhere in the platform (ex the filter widget) so I tracked it down-- I didn't know the emojis would need special handling but good to know!

SmittieC (commented):
Ah, nice. Would you mind adding a tiny test as well?

snopoke (commented):
LGTM pending a test

snopoke (approved):

PR #1478, opened 2025-04-18
add channels column in filters

### LINKED ISSUE #1372: Channel Accessor to Filter widget
Followup to https://github.com/dimagi/open-chat-studio/issues/1339
Use case: as a user in the experiment sessions table I want to be able to view _only_ the API requests

Currently with the toggle, users can view just non-API (default) and non-API and API

Suggested change:
- Add a new column dropdown option for "Channels". This will be the same type as "tags" and "versions" where the user can select multiple options. 
- For the operator, it will have the same and tags and versions, but add "exclude" (add exclude to tags and versions too). 
- Options will be API, web, facebook… etc all the options not just configured for the experiment (this is because they could have been configured at one point). The default for the table should remain the same where view just non-API. 
- Remove the "Include API Sessions" toggle

Image

barora05:
## Description
Add channels in column dropdown in filters.
Add exclude operator for filtering by channels, tags, versions.
Remove include API sessions toggle
Closes #1372

### Demo
[Demo](https://drive.google.com/file/d/1Cd4TReH_A5QEhdsBmzImFKIAiFnMxii6/view?usp=drive_link)


stephherbers:
there's a few other remains of the toggle such as [here](https://github.com/dimagi/open-chat-studio/blob/main/apps/experiments/views/experiment.py#L1077) and in the `async_export_chat` function and [here](https://github.com/dimagi/open-chat-studio/blob/main/templates/experiments/components/exports.html#L7) and in single_chatbot_home.html

apps/experiments/filters.py
```diff
@@ -135,6 +140,45 @@ def build_versions_filter(operator, value):
                 ).values("id")
                 q_objects &= Q(Exists(tag_exists))
             return q_objects
+
+        elif operator == "excludes":
+            combined_exclude_query = Q()
```
stephherbers, line 145: not: refactor to reuse some of the duplicate logic here

templates/experiments/filters.html
```diff
@@ -114,6 +114,7 @@
 <script>
   const tags = JSON.parse(document.getElementById('available-tags-data').textContent);
   const versionsList = JSON.parse(document.getElementById('experiment-versions-data').textContent);
+  const channelsList = ['Telegram', 'Web', 'Whatsapp', 'Facebook', 'SureAdhere', 'API', 'Slack', 'CommCare Connect'];
```
stephherbers, line 117: nit: alphabetize

stephherbers (changes requested):
overall: looks great. Can you confirm 1) export functionality is functional with this change
2) default behavior of API are excluded at the initial table rendering remained unchanged

I'm debating whether including the first filter of "excludes API" at the start since that's the behavior of the results makes sense-- I might ask Brian about what he thinks. Otherwise there might be a confusion that the API sessions are included from the start

barora05:
> overall: looks great. Can you confirm 1) export functionality is functional with this change 2) default behavior of API are excluded at the initial table rendering remained unchanged
> 
> I'm debating whether including the first filter of "excludes API" at the start since that's the behavior of the results makes sense-- I might ask Brian about what he thinks. Otherwise there might be a confusion that the API sessions are included from the start

1. Export functionality works with these changes, I have tested it on my local.
2. API sessions are visible at the initial table rendering for now. Do we not want to show them until API filter is selected?

apps/experiments/filters.py
barora05: done

templates/experiments/filters.html
barora05: done

barora05:
> > API sessions are visible at the initial table rendering for now. Do we not want to show them until API filter is selected?
> 
> The current functionality is that the API sessions are _not_ shown by default, so we want to keep that same default behavior. And yes we do want the filter to reflect the API behavior, so there should be a "channel excludes API" filter also displayed at the beginning to reflect that default state.

@stephherbers 
1. I have maintained the default behavior i.e API sessions would be exlcude by defalt.
2. For generating chat export, I have included the api sessions for now as we don't have the filters in the request passed to export_chat url. I have added a comment in the code for the same. We can create a separate ticket for that and pass the filters in the request and use them to determine what all channels to include in chat export.

apps/experiments/views/experiment.py
```diff
@@ -1074,8 +1070,9 @@ def generate_chat_export(request, team_slug: str, experiment_id: str):
     experiment = get_object_or_404(Experiment, id=experiment_id)
     parsed_url = urlparse(request.headers.get("HX-Current-URL"))
```
stephherbers, line 1071: the channel filters aren't included here and added in `get_filtered_sessions`? Going back to our convo, I thing the download should be fine since it utilized the `apply_dynamic_filters` function to get the sessions queryset. So because this new filter is added to that function, then it should be applied to the export as long as it's in this parsed_url here. If that's true, then you would just need to remove the `include_api` code

stephherbers:
thank you @barora05. One outstanding question-- I forgot about the refactor I did 😅 but that should make life much easier

apps/experiments/views/experiment.py
barora05: My bad. Only the default filter which is set was not getting set in the url and therefore the queryparams were null for that. I have fixed it so we can remove the code for include_api

stephherbers (dismissed):

snopoke:
I tested this locally and it is working well 👍🏻

templates/experiments/filters.html
```diff
@@ -114,6 +114,7 @@
 <script>
   const tags = JSON.parse(document.getElementById('available-tags-data').textContent);
   const versionsList = JSON.parse(document.getElementById('experiment-versions-data').textContent);
+  const channelsList = ['API', 'CommCare Connect', 'Facebook', 'Slack', 'SureAdhere', 'Telegram', 'Web', 'Whatsapp'];
```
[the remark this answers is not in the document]
barora05: Tried it on my local. Got response as this on my local. This does not has API value
{ChannelPlatform.TELEGRAM: True, ChannelPlatform.WHATSAPP: False, ChannelPlatform.FACEBOOK: False, ChannelPlatform.SUREADHERE: False}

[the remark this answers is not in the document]
snopoke: you could add a method to `ChannelPlatform`:

```python
@classmethod
    def for_filter(cls, team) -> dict:
        platforms = list(cls.for_dropdown([], team).keys())
        platforms.append(cls.API)
        return sorted(platforms)
```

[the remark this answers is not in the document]
barora05: got it. But in the ticket it is mentioned to show all options as they could have been configured at some point. So do we really want to remove platforms which aren't available?
 
image

stephherbers:
fyi of this PR which will conflict: https://github.com/dimagi/open-chat-studio/pull/1500

templates/experiments/filters.html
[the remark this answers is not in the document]
snopoke: The options which are removed are ones that aren't enabled for the team e.g. via feature flags.

snopoke (approved):

apps/experiments/filters.py
```diff
@@ -128,6 +132,8 @@ def build_tags_filter(operator, value):
                     )
                 )
             return conditions
+        elif operator == "excludes":
```
snopoke, line 135: nit: 
```suggestion
        elif operator == Operators.EXCLUDES:
```

snopoke (approved):

snopoke (approved):

PR #1485, opened 2025-04-21
Fix: Trigger messages load on intersect event

### LINKED ISSUE #1484: BUG: Messages sometimes doesn't load
It has been reported that sometimes the `messages` tab in the session details view is "stuck"

SmittieC:
## Description

Resolves #1484 

I was able to reproduce this and noticed that the `/messages` call is not being made in the cases when it seems to be stuck. It comes down to the browser not firing the `revealed` event. Why this is, I am not 100% sure. Explanations welcome.

## User Impact

Messages should always be loaded when the user opens up the session details view at the `messages` tab and scrolls down.

### Demo

N/A

### Docs and Changelog

Pending


SmittieC:
Merging, because this is preventing some users from seeing messages

Issue #1237, opened 2025-02-24
Allow specifying the default keyword for router nodes


snopoke:
Currently the first keyword is considered the 'default' but users often set keywords in reverse order and only think of a default at the end. 

The goal is to have a checkbox or button on the router keywords that allows users to specify which is the default keyword.

Related #1232

PR #1486, opened 2025-04-21
Smh/default keyword

stephherbers:
## Description

Resolves [#1237](https://github.com/dimagi/open-chat-studio/issues/1237)

Note: when the default keyword is deleted, the default is changed to the first keyword

## User Impact

Allows users to select the default keyword in router nodes in pipeline

### Demo

[loom video](https://www.loom.com/share/d4ccc291bfc441cc843483a24dfd793f?sid=6a0fc9e7-7da8-4f68-bb62-f0667a0a2b08) (updated with sound)

### Docs and Changelog

yes to both


apps/pipelines/nodes/nodes.py
```diff
@@ -454,10 +453,14 @@ class RouterNode(RouterMixin, Passthrough, HistoryMixin):
     )
 
     def _process_conditional(self, state: PipelineState, node_id=None):
+        default_keyword = self.keywords[self.defaultKeywordIndex] if self.keywords else None
         prompt = OcsPromptTemplate.from_messages(
-            [("system", self.prompt), MessagesPlaceholder("history", optional=True), ("human", "{input}")]
+            [
+                ("system", f"{self.prompt}\nThe default routing destination is: {default_keyword}"),
```
SmittieC, line 459: Why was this necessary?

stephherbers: without this, the default keyword is not used-- and then what's passed in https://github.com/dimagi/open-chat-studio/pull/1486/files#diff-9ff598a128c35d264a8848a5b193d20e69a960eb1e96f69b7003ed197d2b158bR479 is always the first keyword rather than the default

assets/javascript/apps/pipeline/nodes/NodeOutputs.tsx
```diff
@@ -11,16 +11,25 @@ export default function NodeOutputs({data}: {
   const generateOutputHandle = (outputIndex: number) => {
     return multipleOutputs ? `output_${outputIndex}` : "output";
   };
+
   const generateOutputLabel = (outputIndex: number, output_label:string) => {
-    if (multipleOutputs && outputIndex === 0) {
+    // Get the default index from params, defaulting to 0 if not defined
+    const defaultIndex = data.params.defaultKeywordIndex !== undefined ? data.params.defaultKeywordIndex : 0;
```
SmittieC, line 17: Nitty comment: I'm wondering if we should do this, since there is a default value specified in the node params, so the default index will always be specified. Thoughts?

assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -701,14 +713,17 @@ export function KeywordsWidget(props: WidgetParams) {
...
+  const setAsDefault = (index: number) => {
+    setNode(props.nodeId, (old) => {
+      return getNewNodeData(old, [...(old.data.params["keywords"] || [])], index);
+    });
+  }
+
   const length = (Array.isArray(props.nodeParams.keywords) ? props.nodeParams.keywords.length : 1);
-  const keywords = Array.isArray(props.nodeParams.keywords) ? props.nodeParams["keywords"] : []
+  const keywords = Array.isArray(props.nodeParams.keywords) ? props.nodeParams["keywords"] : [];
+  const defaultIndex = props.nodeParams.defaultKeywordIndex !== undefined ? props.nodeParams.defaultKeywordIndex : 0;
```
SmittieC, line 724: Will `props.nodeParams.defaultKeywordIndex` be `undefined` ever?

apps/pipelines/nodes/nodes.py
stephherbers: right, so at first it seems like it would return None, but without modifying the prompt it always returns the first keyword

SmittieC (approved):

PR #1488, opened 2025-04-22
allow superuser privilege access for session views

snopoke:
## Description
Refactor the `verify_session_access_cookie` function to raise the correct exception to allow superuser access.

### DIFF
--- a/apps/experiments/decorators.py
+++ b/apps/experiments/decorators.py
@@ -90,9 +90,8 @@ def verify_session_access_cookie(view):
         if request.user.is_authenticated:
             if request.experiment_session.participant.user_id == request.user.id:
                 return view(request, *args, **kwargs)
-            elif request.team_membership:
-                if request.user.has_perm("chat.view_chat"):
-                    return view(request, *args, **kwargs)
+            elif request.team_membership and request.user.has_perm("chat.view_chat"):
+                return view(request, *args, **kwargs)
 
         try:
             access_value = get_chat_session_access_cookie_data(request)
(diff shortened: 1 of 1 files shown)


apps/experiments/decorators.py
```diff
@@ -86,19 +87,24 @@ def my_view(request, team_slug, experiment_id, session_id):
...
+                    return view(request, *args, **kwargs)
+            else:
+                # If the user is authed but not a team member, initiate superuser access checks
+                allow_superuser_access = True
 
         try:
             access_value = get_chat_session_access_cookie_data(request)
         except (signing.BadSignature, KeyError):
-            raise Http404()
+            raise TeamAccessDenied() if allow_superuser_access else Http404()
```
SmittieC, line 104: Does `request.user.is_superuser` not work?

snopoke: I suppose it would

snopoke: e217c8511db1db02dfd0c25f0d01fc9d7645a6fc

SmittieC (approved):

apps/experiments/decorators.py
```diff
@@ -89,16 +90,17 @@ def _inner(request, *args, **kwargs):
         if request.user.is_authenticated:
             if request.experiment_session.participant.user_id == request.user.id:
                 return view(request, *args, **kwargs)
-            elif request.team_membership and request.user.has_perm("chat.view_chat"):
-                return view(request, *args, **kwargs)
+            elif request.team_membership:
```
SmittieC, line 93: nit: this can be reverted

PR #1497, opened 2025-04-23
ensure channel.new_user_message returns a ChatMessage object

snopoke:
## Description
(Review by commit)

The first few commits are test refactors followed by a change to the base channel class to return ChatMessage objects instead of strings.

Resolves [OPEN-CHAT-STUDIO-VZ](https://dimagi.sentry.io/issues/6561632397/events/8ac29816006747448f92daff6fc380c8/)


SmittieC (approved):
Nice :tophat:

PR #1499, opened 2025-04-23
upgrade langgraph & related dependencies

snopoke:
## Description
I've done some basic testing locally using pipelines and everything seems to be working.

PR #1505, opened 2025-04-23
refactor pipelines conditional edges to use langgraph Command pattern

snopoke:
## Description
PR into https://github.com/dimagi/open-chat-studio/pull/1499

The langgraph docs say that when combining state updates and control flow you should use the `Command` pattern: https://langchain-ai.github.io/langgraph/how-tos/command/

When testing https://github.com/dimagi/open-chat-studio/pull/1470 I was unable to get the state to update correctly which is what prompted this change.

The last commit is needed to fix an issue that arose from the upgrade. Previously we were making use of some langchain internals to determine which input to pass to a node in the case where there are multiple valid options which broke after the upgrade.

The fix I've implemented is to store the path through the pipeline as a list of tuples: `(previous, current, next)`. This can then be used to reliably find the correct input for a node in the case where there are multiple incoming edges.

## User Impact
None

### Demo
NA

### Docs and Changelog
NA


apps/pipelines/nodes/base.py
```diff
@@ -81,8 +82,16 @@ class PipelineState(dict):
     attachments: list = Field(default=[])
     output_message_tags: Annotated[list[str], operator.add]
 
+    # List of (previous, current, next) tuples used for aiding in routing decisions.
+    # `previous` can be None, `current` is always a string, `next` is always a list
+    path: Annotated[Sequence[tuple[str, str, str]], operator.add]
```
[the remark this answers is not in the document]
snopoke: [15a965d](https://github.com/dimagi/open-chat-studio/pull/1505/commits/15a965d0dc363d41f87872e7a21018b990fc2cd6)

SmittieC (approved):

apps/pipelines/tests/test_schemas.py
```diff
@@ -3,13 +3,14 @@
...
 
-BASE = pathlib.Path(__file__).parent / "data"
+BASE = pathlib.Path(__file__).parent / "node_schemas"
 
 
 def test_schemas():
     schemas = _pipeline_node_schemas()
     for schema in schemas:
         title = schema["title"]
+        print(title)
```
SmittieC, line 13:
```suggestion

```

snopoke:
@SmittieC I've updated the base branch for this PR to `main` and closed the other langgraph upgrade branch.

SmittieC (dismissed):

snopoke:
I've done some more testing on this and fixed 2 small issues - I think this is now GTG

snopoke:
I'll merge this next week

apps/pipelines/graph.py
```diff
@@ -190,25 +191,25 @@ def _add_nodes_to_graph(self, state_graph: StateGraph, nodes: list[Node]):
...
-                        continue
-                    state_graph.add_conditional_edges(
-                        edge.source,
-                        partial(node.pipeline_node_instance.process_conditional, node_id=edge.source),
-                        self.conditional_edge_map[edge.source],
-                    )
-                    seen_sources.add(edge.source)
-                else:
+                if not edge.is_conditional():
+                    # conditional edges are handled by router note outputs
```
SmittieC, line 212:
```suggestion
                    # conditional edges are handled by router node outputs
```

PR #1509, opened 2025-04-24
Copy array to prevent changing a potentially immutable object

### LINKED ISSUE #1495: LLM node breaks when toggling tools
See [this Sentry error](https://dimagi.sentry.io/issues/6562268817/?alert_rule_id=14063560&alert_type=issue&notification_uuid=e1950ff8-45a5-40d6-97b2-05580dc82b5f&project=4505001320316928&referrer=issue_alert-slack)

### Reproduce
In an LLM node, select a tool, deselect it, select it (or another) tool again. Observe the console logs.

SmittieC:
## Description

Resolves #1495

The `selectedValues` array becomes immutable after being updated by `immer` (without reloading the page). Since the `immer` library is producing a new, immutable object on update, I think this is causes the `selectedValues` array to ultimately become immutable as well.

## User Impact

Users can toggle tools again

### Demo

N/A

I did manual testing

### Docs and Changelog

Pending...

### DIFF
--- a/assets/javascript/apps/pipeline/nodes/widgets.tsx
+++ b/assets/javascript/apps/pipeline/nodes/widgets.tsx
@@ -216,7 +216,9 @@ function MultiSelectWidget(props: WidgetParams) {
   if (options.length == 0) {
     return <></>
   }
-  let selectedValues = Array.isArray(props.paramValue) ? props.paramValue : [];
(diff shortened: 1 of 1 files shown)


snopoke (approved):
Wow, good detective work!

SmittieC:
> Wow, good detective work!

Thanks, but to be fair, I had some AI help

Issue #1503, opened 2025-04-23
Session filter by start date


snopoke:
Ability to filter sessions by session start date.

PR #1510, opened 2025-04-24
Smh/filter session start

stephherbers:
## Description

resolves [#1503](https://github.com/dimagi/open-chat-studio/issues/1503)

Note: I chose to add "First Message"out of alphabetical order of the drop down to keep it next to the "Last Message"

## User Impact

users can now filter by the start of the experiment session

### Demo

[loom](https://www.loom.com/share/74e6415ae8324951a3f63424a277cb5e?sid=1bda78bf-8de7-4f99-a05c-4437d4ebd528)

### Docs and Changelog

no docs, yes changelog

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -1439,10 +1439,7 @@ class SessionStatus(models.TextChoices):
     UNKNOWN = "unknown", gettext("Unknown")
 
 
-class ExperimentSessionObjectManager(models.Manager):
-    def for_chat_id(self, chat_id: str) -> list["ExperimentSession"]:
-        return self.filter(participant__identifier=chat_id)
-
+class ExperimentSessionQuerySet(models.QuerySet):
     def with_last_message_created_at(self):
         last_message_created_at = (
             ChatMessage.objects.filter(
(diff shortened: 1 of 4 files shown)


apps/experiments/models.py
```diff
@@ -1439,10 +1439,7 @@ class SessionStatus(models.TextChoices):
     UNKNOWN = "unknown", gettext("Unknown")
 
 
-class ExperimentSessionObjectManager(models.Manager):
-    def for_chat_id(self, chat_id: str) -> list["ExperimentSession"]:
-        return self.filter(participant__identifier=chat_id)
-
+class ExperimentSessionQuerySet(models.QuerySet):
```
stephherbers, line 1442: note: had to make this class so I could do the chaining of `ExperimentSession.objects.with_last_message_created_at().with_first_message_created_at()` but the functionality is the same

apps/experiments/models.py
```diff
@@ -1453,6 +1450,30 @@ def with_last_message_created_at(self):
...
+                chat__experiment_session=models.OuterRef("pk"),
+            )
+            .order_by("created_at")
+            .values("created_at")[:1]
+        )
+        return self.annotate(first_message_created_at=models.Subquery(first_message_created_at))
+
+
+class ExperimentSessionObjectManager(models.Manager):
+    def for_chat_id(self, chat_id: str) -> list["ExperimentSession"]:
```
SmittieC, line 1465: Not really related to this PR, but it doesn't look like this method is used anymore, except by tests

apps/experiments/models.py
SmittieC: I'm not against this approach, but am going to put another approach forward for your consideration.
One could add a `with_filter_annotations` (or something in that line) method on the object manager that adds all the necessary annotations to the queryset. In this case the `first_message_created_at` and `last_message_created_at` fields.

```python
def with_filter_annotations(self):
        first_message_created_at = (
            ChatMessage.objects.filter(
                chat__experiment_session=models.OuterRef("pk"),
            )
            .order_by("created_at")
            .values("created_at")[:1]
        )
        last_message_created_at = (
            ChatMessage.objects.filter(
                chat__experiment_session=models.OuterRef("pk"),
            )
            .order_by("-created_at")
            .values("created_at")[:1]
        )

        self.queryset = self.queryset.annotate(first_message_created_at=models.Subquery(first_message_created_at))
        self.queryset = self.queryset.annotate(last_message_created_at=models.Subquery(last_message_created_at))
        return self.queryset
```

SmittieC (commented):
Do we have tests for these filters? Might be good to add some if not?

apps/experiments/models.py
stephherbers: happy to switch

stephherbers:
> Do we have tests for these filters? Might be good to add some if not?

I'm a +1 to this. There are none currently-- I think it might be a big pull for this ticket since I think it might turn into a surprisingly large task-- but I'll write a ticket for it

snopoke (approved):

PR #1513, opened 2025-04-25
show survey and voice tabs for pipeline bots

snopoke:
## Description
Show the survey and voice fields for pipeline based experiments. These configuration options will still work with pipelines since they impact functionality in the base channel.

## User Impact
Users can configure surveys and voice behaviour for pipeline bots.

### Docs and Changelog
https://github.com/dimagi/open-chat-studio-docs/pull/94


SmittieC (approved):
Is there a doc / plan on how these options would be configured for chatbots?

snopoke:
> Is there a doc / plan on how these options would be configured for chatbots?

not yet

PR #1518, opened 2025-04-25
enable bugbear lint rules

snopoke:
## Description
See https://docs.astral.sh/ruff/rules/#flake8-bugbear-b


SmittieC (commented):
This would be a nice touch. The tests for the "fix_vector_store_duplication" command are failing. I don't think this command is useful anymore though, so we can probably remove it and the test

snopoke:
> This is a nice touch. The tests for the "fix_vector_store_duplication" command are failing. I don't think this command is useful anymore though, so we can probably remove it and the test

I'll remove it in a separate PR

SmittieC (approved):

PR #1519, opened 2025-04-28
add state while creating session and add option to update state

### LINKED ISSUE #1508: Allow updates to session state via the APIs
* Update the session create endpoint to allow passing state.

* Update the `ExperimentSessionViewSet` API class to allow updating the state of an existing session.

barora05:
## Description
Ability to pass state while creating session and PATCH endpt to update state of session. Closes #1508

### DIFF
--- a/apps/api/tests/test_session_api.py
+++ b/apps/api/tests/test_session_api.py
@@ -249,3 +249,23 @@ def test_end_experiment_session_success(client, session):
     assert response.status_code == status.HTTP_200_OK
     session.refresh_from_db()
     assert session.status == "pending-review"
+
+
+@pytest.mark.django_db()
+def test_update_experiment_session_state_success(session):
+    team = Team.objects.create(name="Test Team")
+    session.team = team
+    session.save()
+    url = f"/api/sessions/{session.external_id}/update_state/"
+    user = session.experiment.team.members.first()
+    client = ApiTestClient(user, session.experiment.team)
+    new_state = {"some": "new_state", "updated": True}
+
+    response = client.patch(url, data={"state": new_state}, format="json")
+
+    assert response.status_code == status.HTTP_200_OK
+    response_data = response.json()
+    assert response_data["success"] is True
+    assert response_data["new_state"] == new_state
+    session.refresh_from_db()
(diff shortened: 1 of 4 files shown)


apps/api/views.py
```diff
@@ -344,6 +375,22 @@ def end_experiment_session(self, request, id=None):
         session.end()
         return Response(status=status.HTTP_200_OK)
 
+    @action(detail=True, methods=["patch"])
+    def update_state(self, request, id=None):
```
SmittieC, line 379: The id will never be `None`, since it's required. I see the same issue with the `end_experiment_session` view.

barora05: done

SmittieC (approved):

apps/api/views.py
```diff
@@ -245,6 +245,22 @@ def _get_id(data):
...
+    fields={
+        "state": serializers.JSONField(),
+    },
+)
+
+update_state_response_serializer = inline_serializer(
+    name="update_state_response",
+    fields={
+        "success": serializers.BooleanField(),
+        "new_state": serializers.JSONField(),
```
snopoke, line 259: nit: I think this should just be `state`
```suggestion
        "state": serializers.JSONField(),
```

snopoke (commented):
small comment but otherwise looks good

snopoke (dismissed):

apps/api/tests/test_session_api.py
```diff
@@ -249,3 +249,47 @@ def test_end_experiment_session_success(client, session):
...
+    user = session.experiment.team.members.first()
+    client = ApiTestClient(user, session.experiment.team)
+    new_state = {"some": "new_state", "updated": True}
+
+    response = client.patch(url, data={"state": new_state}, format="json")
+
+    assert response.status_code == status.HTTP_200_OK
+    response_data = response.json()
+    assert response_data["success"] is True
+    assert response_data["new_state"] == new_state
```
snopoke, line 269:
```suggestion
    assert response_data["state"] == new_state
```

snopoke (approved):

PR #1522, opened 2025-04-29
Remove the extra url tag

SmittieC:
## Description

When toggling `show_archived` in the experiments / chatbots list, the server responded with a 500

## User Impact

User can see archived versions again.

### Demo

N/A

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/97


snopoke:
Thanks

PR #1545, opened 2025-04-30
Ba/copy chatbot

### LINKED ISSUE #1426: Ability to clone a chatbot / pipeline / assistant
Allow users to create copies of experiments. This is similar to creating a new version except without attaching the new version to the parent experiment.

This should not include some configuration e.g. channels

barora05:
## Description
Add ability to copy chatbots(i.e only pipeline experiments) within the same team.
Added a button on chatbots home to copy a particular chatbot.
Copying redirects to the home of copied chatbot.
Create default and published version of copied chatbot. Copies configuration like pipeline, events.
Append _copy to copied chatbot name.
Closes #1426

### Demo
[Demo](https://drive.google.com/file/d/13kIiXkT4F1lV0_s1SGY8Ykl2pLGnlTuP/view?usp=drive_link)

[Updated demo with dialog box ](https://drive.google.com/file/d/1udm5QhT0uKY6Qz1ejh57ilOYPDmvPFqO/view?usp=drive_link)


barora05:
We can copy a same experiment mutiple times by clicking the button. Moreover, we can also copy an already copied experiment. To avoid accidental duplication, we should consider showing a confirmation prompt when the user clicks the "Copy Experiment" button. The prompt can ask the user to confirm whether they really want to proceed with copying the experiment or if they would prefer to change its name. This would help prevent unnecessary copies caused by unintentional clicks.

snopoke:
> We can copy a same experiment mutiple times by clicking the button. Moreover, we can also copy an already copied experiment. To avoid accidental duplication, we should consider showing a confirmation prompt when the user clicks the "Copy Experiment" button. The prompt can ask the user to confirm whether they really want to proceed with copying the experiment or if they would prefer to change its name. This would help prevent unnecessary copies caused by unintentional clicks.

+1 to this

I think it would also be good to add the 'copy' button on the main chatbot (and experiment) details page next to the edit button.

barora05:
> > We can copy a same experiment mutiple times by clicking the button. Moreover, we can also copy an already copied experiment. To avoid accidental duplication, we should consider showing a confirmation prompt when the user clicks the "Copy Experiment" button. The prompt can ask the user to confirm whether they really want to proceed with copying the experiment or if they would prefer to change its name. This would help prevent unnecessary copies caused by unintentional clicks.
> 
> +1 to this
> 
> I think it would also be good to add the 'copy' button on the main chatbot (and experiment) details page next to the edit button.

So, keep the copy button at chatbots home page also or only keep it at the chatbot details page?

snopoke:
> > > We can copy a same experiment mutiple times by clicking the button. Moreover, we can also copy an already copied experiment. To avoid accidental duplication, we should consider showing a confirmation prompt when the user clicks the "Copy Experiment" button. The prompt can ask the user to confirm whether they really want to proceed with copying the experiment or if they would prefer to change its name. This would help prevent unnecessary copies caused by unintentional clicks.
> > 
> > 
> > +1 to this
> > I think it would also be good to add the 'copy' button on the main chatbot (and experiment) details page next to the edit button.
> 
> So, keep the copy button at chatbots home page also or only keep it at the chatbot details page?

Yes 👍🏻

apps/pipelines/helper.py
```diff
@@ -0,0 +1,70 @@
...
+
+        old_to_new_node_ids[old_id] = new_id
+        new_node = node.copy()
+        new_node["id"] = new_id
+        new_node["data"] = new_node["data"].copy()
+        new_node["data"]["id"] = new_id
+
+        if "params" in new_node["data"]:
+            new_node["data"]["params"] = new_node["data"]["params"].copy()
+            new_node["data"]["params"]["name"] = new_id
```
snopoke, line 29: should check if the name is the same as `old_id` before overwriting it

apps/pipelines/helper.py
```diff
@@ -0,0 +1,70 @@
...
+        new_edge_id = f"reactflow__edge-{'-'.join(edge_id_parts)}-{uuid4().hex[:8]}"
+
+        new_edge = edge.copy()
+        new_edge["id"] = new_edge_id
+        new_edge["source"] = new_source_id
+        new_edge["target"] = new_target_id
+        if "sourceHandle" in edge:
+            new_edge["sourceHandle"] = edge["sourceHandle"]
+        if "targetHandle" in edge:
+            new_edge["targetHandle"] = edge["targetHandle"]
```
snopoke, line 60: aren't these handled by the line `new_edge = edge.copy()`

apps/experiments/models.py
```diff
@@ -830,14 +833,16 @@ def create_new_version(self, version_description: str | None = None, make_defaul
             self.versions.filter(is_default_version=True).update(
                 is_default_version=False, audit_action=AuditAction.AUDIT
             )
-
+        if copy_experiment:
+            new_version.name = new_version.name + "_copy"
+            new_version.version_number = 1
         new_version.save()
 
         self._copy_safety_layers_to_new_version(new_version)
```
snopoke, line 841: For safety layers, we should just reference the working version instead of creating a new version

apps/experiments/models.py
```diff
@@ -830,14 +833,16 @@ def create_new_version(self, version_description: str | None = None, make_defaul
...
         self._copy_routes_to_new_version(new_version)
```
snopoke, line 842: For now this should check if there are routes and raise an error since we aren't supporting copying experiments with child links.

apps/experiments/models.py
```diff
@@ -830,14 +833,16 @@ def create_new_version(self, version_description: str | None = None, make_defaul
...
         self._copy_trigger_to_new_version(trigger_queryset=self.static_triggers, new_version=new_version)
         self._copy_trigger_to_new_version(trigger_queryset=self.timeout_triggers, new_version=new_version)
```
snopoke, line 844: These also need special handling so that we just make copies and don't attach them to their working version.

apps/experiments/models.py
```diff
@@ -830,14 +833,16 @@ def create_new_version(self, version_description: str | None = None, make_defaul
...
-        self._copy_pipeline_to_new_version(new_version)
+        self._copy_pipeline_to_new_version(new_version, copy_experiment)
         self._copy_custom_action_operations_to_new_version(new_experiment=new_version)
         self._copy_assistant_to_new_version(new_version)
```
snopoke, line 847: Similar to routes - we should fail if trying to copy an experiment with an assistant

apps/experiments/models.py
```diff
@@ -830,14 +833,16 @@ def create_new_version(self, version_description: str | None = None, make_defaul
...
+            new_version.version_number = 1
         new_version.save()
 
         self._copy_safety_layers_to_new_version(new_version)
         self._copy_routes_to_new_version(new_version)
         self._copy_trigger_to_new_version(trigger_queryset=self.static_triggers, new_version=new_version)
         self._copy_trigger_to_new_version(trigger_queryset=self.timeout_triggers, new_version=new_version)
-        self._copy_pipeline_to_new_version(new_version)
+        self._copy_pipeline_to_new_version(new_version, copy_experiment)
         self._copy_custom_action_operations_to_new_version(new_experiment=new_version)
```
snopoke, line 846: Same as above, this needs special handling to create a detached copy

apps/pipelines/models.py
```diff
@@ -379,19 +380,23 @@ def _save_message_to_history(
         return chat_message
 
     @transaction.atomic()
-    def create_new_version(self):
+    def create_new_version(self, copy_experiment: bool = False):
```
SmittieC, line 383: Nitty comment: Can we rename the `copy_experiment` param to something like `is_copy`? My thought is that `copy_experiment` doesn't quite fit the description when this method is called on non-experiment models.

templates/chatbots/chatbot_copy_dialog.html
```diff
@@ -0,0 +1,52 @@
...
+    const nameInput = document.getElementById('copied-chatbot-name');
+    const copiedName = nameInput.value.trim();
+
+    if (!copiedName) {
+      alert("Please enter a name for the copied chatbot.");
+      return;
+    }
+
+    const url = `/a/${currentTeamSlug}/chatbots/${currentExperimentId}/copy/?new_name=${encodeURIComponent(copiedName)}`;
+    window.location.href = url;
```
snopoke, line 50: Rather just use a normal form inside the dialog like https://github.com/dimagi/open-chat-studio/blob/47ad28395184121a333bad8241c7e7f419c8fb8e/templates/chatbots/single_chatbot_home.html#L125-L145

Then you can use a POST request and do normal form validation.

apps/pipelines/models.py
```diff
@@ -379,19 +380,23 @@ def _save_message_to_history(
         return chat_message
 
     @transaction.atomic()
-    def create_new_version(self):
+    def create_new_version(self, is_copy: bool = False):
         version_number = self.version_number
-        self.version_number = version_number + 1
+        self.version_number = 1 if is_copy else self.version_number + 1
```
snopoke, line 385: This will update the version of the item being copied which is incorrect:
```suggestion
        if not is_copy:
            self.version_number = self.version_number + 1
```

templates/chatbots/single_chatbot_home.html
```diff
@@ -114,6 +120,7 @@ <h3 class="font-bold text-lg inline mr-4">Channels:</h3>
         </div>
         {% if experiment.is_public %}
           {% include "experiments/share/dialog.html" with origin="chatbots" %}
+          {% include "chatbots/chatbot_copy_dialog.html"%}
```
snopoke, line 123: shouldn't be inside the if block

templates/experiments/components/experiment_actions_column.html
```diff
@@ -1,4 +1,5 @@
 {% if record.is_editable %}
+  {% include "chatbots/chatbot_copy_dialog.html"%}
```
snopoke, line 2: you can pass values into the include:
```suggestion
  {% include "chatbots/chatbot_copy_dialog.html" with chatbot_id=record.id team_slug=record.team.slug %}
```

snopoke, line 2: you can pass values into the include:
```suggestion
  {% include "chatbots/chatbot_copy_dialog.html" with chatbot_id=record.id team_slug=record.team.slug %}
```

templates/experiments/single_experiment_home.html
```diff
@@ -114,6 +123,7 @@ <h3 class="font-bold text-lg inline mr-4">Channels:</h3>
         </div>
         {% if experiment.is_public %}
           {% include "experiments/share/dialog.html" %}
+          {% include "chatbots/chatbot_copy_dialog.html"%}
```
snopoke, line 126: move out of `if` block

templates/experiments/components/experiment_actions_column.html
```diff
@@ -9,6 +10,12 @@
      href="{% url type|add:':edit' record.team.slug record.id %}" title="View">
     <i class="fa-solid fa-pencil"></i>
   </a>
+  {% if type == 'chatbots' %}
+    <a class="btn btn-sm"
+       onclick="openCopyDialog({{ record.id }}, '{{ record.team.slug }}')">
```
snopoke, line 15: if you give the dialog a consistent ID then you can open it using that ID e.g. `experiment_{{ experiment_id }}_copy_modal.showModal()`

snopoke (commented):
Please add a test for creating a copy of the experiment which verifies that the connected objects are copied correctly and not still attached to the previous experiment or their previous working copy.

barora05:
> Please add a test for creating a copy of the experiment which verifies that the connected objects are copied correctly and not still attached to the previous experiment or their previous working copy.

@snopoke I have verified in the db for experiment and pipeline that the working_version_id is NULL for copied experiment and pipeline but while writing test working_version and working_verison_id is present.

barora05:
> > @snopoke I have verified in the db for experiment and pipeline that the working_version_id is NULL for copied experiment and pipeline but while writing test working_version and working_verison_id are present.
> 
> Can you commit what you've got and I'll take a look

Commented the test in test_chatbot_views.py for now.

snopoke (approved):

PR #1558, opened 2025-05-06
add a system tag

### LINKED ISSUE #996: Tag 'unsupported message type' messages
Add a system tag to messages saved in the history due to unsupported message types received from the user.

Relevant code: https://github.com/dimagi/open-chat-studio/blob/9a2913d0b4514820ce1cebe2743ffe512c340e01/apps/chat/channels.py#L685-L693

See https://dimagi.slack.com/archives/C05PK63Q89H/p1733823363680929

barora05:
## Description
Closes #996


apps/chat/channels.py
```diff
@@ -704,6 +705,10 @@ def _unsupported_message_type_response(self) -> str:
             session=self.experiment_session, experiment=self.experiment, trace_service=self.trace_service
         )
         trace_info = TraceInfo(name="unsupported message", metadata={"message_type": self.message.content_type})
+        chat_message = ChatMessage.objects.create(
```
SmittieC, line 708: Is there a case where `self.message.message_text` will not be en empty string?

snopoke: I also wondered that but looking at the message parsing code it seems like it could be possible that the text is not empty.

SmittieC (approved):

Issue #1424, opened 2025-04-08
[RAG] Create attachments for referenced files


SmittieC:
Depends on https://github.com/dimagi/open-chat-studio/issues/1425

See [the spec](https://docs.google.com/document/d/1_BScbohNxwysXdeR9ibeRZ07Ojy3m9yFQ2VBQDBYCBM/edit?tab=t.ik6hmw5sgrpx#heading=h.5riri4ypemqd) and [this section](https://docs.google.com/document/d/1_BScbohNxwysXdeR9ibeRZ07Ojy3m9yFQ2VBQDBYCBM/edit?tab=t.ik6hmw5sgrpx#heading=h.7xqgfkdanuwr).

We need the ability to parse the [response object](https://platform.openai.com/docs/api-reference/responses/object#responses/object-output) and extract any referenced files so that we can create attachments for the chat message. We can probably follow the same approach we use [here](https://github.com/dimagi/open-chat-studio/blob/01fb80c6f8b2b8459640453ed60b01b48836b8e3/apps/service_providers/llm_service/runnables.py#L324-L426). The referenced files can be included in the `openai_file_id` list in the chat message metadata.

### Task breakdown
- [ ] Create attachments from the [Response object](https://platform.openai.com/docs/api-reference/responses/object).
- [ ] Ensure that these attachments are shown and downloadable from the UI. Additionally, download links to these attachments should be appended to bot responses for external channels. This should already be the case, but we need to verify that this is indeed happening.

SmittieC:
Please hold off on doing this until we have full support for the OpenAI responses API

PR #1560, opened 2025-05-07
RAG - Link document index to LLM node

### LINKED ISSUE #1422: [RAG] Uploading files to OpenAI
See [the spec](https://docs.google.com/document/d/1_BScbohNxwysXdeR9ibeRZ07Ojy3m9yFQ2VBQDBYCBM/edit?tab=t.ik6hmw5sgrpx#heading=h.e2u38uq93ams) and [this section](https://docs.google.com/document/d/1_BScbohNxwysXdeR9ibeRZ07Ojy3m9yFQ2VBQDBYCBM/edit?tab=t.ik6hmw5sgrpx#heading=h.7xqgfkdanuwr).

When files are uploaded to a document index, we upload them to the `Collection`'s vector store using a celery task. (See [this ticket](https://github.com/dimagi/open-chat-studio/issues/1421)). See also [this file](https://github.com/dimagi/open-chat-studio/blob/a27025d4da7e027e4f67a4e0094dabd8fa936993/apps/assistants/sync.py). We should have an indicator in the UI to show the [embedding status](https://platform.openai.com/docs/api-reference/vector-stores-files/file-object) of each file. The `status` field on the `through` model for the `Collection` - `File` relationship will contain this information.

**How will this status field be updated?**
I think it would make sense to have an async check in the UI that polls the backend for the status of each file that is **not yet completely embedded**. So we don't use an async check if we know that it is embedded, but if we see that it's still in progress, we need an async check. This check should update the status as well when we see that the file is embedded.

_This approach is debatable_

The chunking strategy should be saved in each file's metadata field.

**Update**
I decided to rather save the chunking strategy on the through model

### LINKED ISSUE #1425: [RAG] Select which document index to attach to the node
See [the spec](https://docs.google.com/document/d/1_BScbohNxwysXdeR9ibeRZ07Ojy3m9yFQ2VBQDBYCBM/edit?tab=t.ik6hmw5sgrpx#heading=h.afk7hycodncu).

See also how [collections are exposed to nodes](https://github.com/dimagi/open-chat-studio/blob/b7c3b91c85fdc2ef8bbd1c76031caf8c417c0f31/apps/pipelines/views.py#L178-L187).

We need users to be able to select a document index / vector store that their bot can use. We should only show the indexes for the selected OpenAI provider, but make it clear that the indexes are OpenAI specific. Note that we need to keep this behind the [`document_management` FF](https://github.com/dimagi/open-chat-studio/blob/fb81005b6081f3367f0e33881443790b4acbd186/apps/pipelines/nodes/nodes.py#L256).

We also need to support builtin tools, like `file_search`. When the user selected a document index, the `file_search` tool should be automatically included in the set of tools that the LLM can use. See the [Langchain docs](https://python.langchain.com/docs/integrations/chat/openai/#built-in-tools). In the future, we can extend this set to include `web_search` as well

### Task breakdown
- [ ] Ability for user to select a document index on the LLM node
- [ ] Update the code around tool selection to automatically include the `file_search` tool with the correct vector store id when the user added a document index to a node.

SmittieC:
## Description

Resolves #1422 - Uploading files to OpenAI
Resolves #1425 - Selecting a document index on a node

Tiny note:
I introduced a `VectorStoreManager` to centralize operations related to OpenAI vector stores

## User Impact

- When users create a document index, a vector store will be created for that index. Whenever they upload files to it, it will be uploaded to OpenAI and linked to the vector store in turn
- Users can link a document index to an LLM node

### Docs and Changelog

Pending


apps/documents/tasks.py
```diff
@@ -0,0 +1,39 @@
...
+            remote_file_id = create_files_remote(client, files=[collection_file.file])
+            collection_file.status = FileStatus.COMPLETED
+            file_ids.extend(remote_file_id)
+            link_files_to_vector_store(
+                client,
+                vector_store_id=collection.openai_vector_store_id,
+                file_ids=file_ids,
+                chunk_size=chuking_strategy["chunk_size"],
+                chunk_overlap=chuking_strategy["chunk_overlap"],
+            )
```
snopoke, line 34: I think this should be outside the loop

apps/documents/views.py
```diff
@@ -82,29 +97,43 @@ def add_collection_files(request, team_slug: str, pk: int):
...
 
     messages.success(request, f"Added {len(files)} files to collection")
     return redirect("documents:single_collection_home", team_slug=team_slug, pk=pk)
 
 
+@require_POST
 @login_and_team_required
 @permission_required("documents.change_collection")
-@require_POST
 def delete_collection_file(request, team_slug: str, pk: int, file_id: int):
```
snopoke, line 126: needs atomic decorator

apps/documents/views.py
```diff
@@ -128,6 +157,22 @@ def get_form_kwargs(self):
         kwargs["request"] = self.request
         return kwargs
 
+    def sync_openai_vector_store(self, collection: Collection, remove_old_vector_store: bool = False):
+        try:
+            if remove_old_vector_store:
```
snopoke, line 162:
```suggestion
            if remove_old_vector_store and collection.openai_vector_store_id:
```

apps/documents/views.py
```diff
@@ -128,6 +157,22 @@ def get_form_kwargs(self):
...
+                delete_vector_store(collection.llm_provider, collection.openai_vector_store_id, fail_silently=True)
```
snopoke, line 163: won't this be the old llm provider?

apps/documents/views.py
```diff
@@ -184,6 +236,13 @@ def delete(self, request, team_slug: str, pk: int):
...
         else:
+            if collection.is_index and collection.openai_vector_store_id:
+                try:
+                    delete_vector_store(collection.llm_provider, collection.openai_vector_store_id)
+                    messages.success(request, "Collection deleted")
+                except Exception as e:
+                    logger.exception(f"Could not delete vector store for collection {collection.id}. {e}")
+                    messages.error(self.request, "Could not delete the vector store at OpenAI. Please try again later")
+
             collection.archive()
```
snopoke, line 247: should this happen if the openai operation fails?

apps/pipelines/nodes/nodes.py
```diff
@@ -261,6 +261,13 @@ class LLMResponseWithPrompt(LLMResponse, HistoryMixin):
             widget=Widgets.select, options_source=OptionsSource.collection, flag_required="document_management"
         ),
     )
+    document_index_id: OptionalInt = Field(
+        None,
+        title="Document Index",
```
snopoke, line 266: Shouldn't this be called 'Collection' to keep it consistent?

apps/documents/views.py
SmittieC: Ah yes, well spotted

apps/documents/views.py
SmittieC: Maybe not, no

apps/documents/tasks.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1560/commits/52b9aa8f67939a57bd493e06d5b3f91d95cfec51

apps/documents/views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1560/commits/52b9aa8f67939a57bd493e06d5b3f91d95cfec51

apps/documents/views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1560/commits/52b9aa8f67939a57bd493e06d5b3f91d95cfec51

apps/documents/views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1560/commits/52b9aa8f67939a57bd493e06d5b3f91d95cfec51

templates/documents/single_collection_home.html
```diff
@@ -19,6 +19,12 @@
       <div class="flex justify-between items-start">
         <div>
           <h1 class="pg-title">{{ collection.name }} {% if collection.is_index %}<span class="badge badge-sm badge-ghost">Document Index</span>{% endif %}</h1>
+          <div class="text-sm text-base-content/60 mt-1">
+            {% if collection.is_index %}
```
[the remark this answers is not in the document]
SmittieC: I think this should stay as-is though, especially when we do our own RAG, in which case there wouldn't be an `openai_vector_store_id`. Granted, maybe some other id, but even then, what if the call to create a vector store at OpenAI fails? I think we should still show the user that this is an index (and add retry logic for the vector store creation).

apps/documents/views.py
```diff
@@ -119,6 +140,29 @@ def get_form_kwargs(self):
         kwargs["request"] = self.request
         return kwargs
 
+    def sync_openai_vector_store(
+        self, collection: Collection, remove_old_vector_store: bool = False, old_llm_provider_id: int | None = None
+    ):
+        # TODO: Upload files to the new vector store
```
SmittieC, line 146: I'm going to do this in a second PR

apps/service_providers/llm_service/runnables.py
```diff
@@ -227,17 +227,30 @@ def _build_chain(self) -> Runnable[dict[str, Any], Any]:
 
 class AgentLLMChat(LLMChat):
     def _parse_output(self, output):
-        return output.get("output", "")
+        output = output.get("output", "")
+        if isinstance(output, list):
+            # Responses API responses are lists
+            return "\n".join([o["text"] for o in output])
```
snopoke, line 233: Is the newline here correct? I would have thought we'd just join directly. Also the langchain docs say you can use `output.text()`:

```
for token in llm_with_tools.stream("..."):
    print(token.text(), end="|")
```

Side note: Are we going to support annotations etc in a future PR?

templates/documents/single_collection_home.html
```diff
@@ -52,7 +58,9 @@ <h2 class="text-lg font-semibold">Files</h2>
               <div class="flex items-center gap-2">
                 {% if collection_file.status %}
                   <div class="tooltip" data-tip="{{ collection_file.status }}">
```
snopoke, line 60: the tooltip should use the status label from the enum instead of the value (`In Progress` instead of `in_progress`). My preferred way for doing that is to add a `<field_name>_enum` property to the model:

```python
@property
def status_enum(self):
    return FileStatus(self.status)
```

snopoke (commented):
This is looking good.

apps/pipelines/nodes/nodes.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1560/commits/fa1f31092fee54328bd2937cdcda911f300ba285

apps/documents/views.py
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/1571

apps/assistants/sync.py
```diff
@@ -548,3 +549,51 @@ def get_and_store_openai_file(client, file_id: str, team_id: int) -> File:
...
     return File.from_external_source(filename, file_content_obj, file_id, "openai", team_id)
+
+
+class VectorStoreManager:
+    def __init__(self, client: openai.OpenAI):
+        self.client = client
+
+    @classmethod
+    def from_llm_provider(cls, llm_provider):
+        return cls(llm_provider.get_llm_service().get_raw_client())
```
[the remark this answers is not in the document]
SmittieC: Do you have a preference? I don't, so up to you

apps/assistants/sync.py
```diff
@@ -548,3 +549,51 @@ def get_and_store_openai_file(client, file_id: str, team_id: int) -> File:
...
+    def create_vector_store(self, name: str, file_ids: list = None) -> str:
+        file_ids = file_ids or []
+        vector_store = self.client.vector_stores.create(name=name, file_ids=file_ids)
+        return vector_store.id
+
+    def delete_vector_store(self, vector_store_id: str, fail_silently: bool = False):
+        try:
+            self.client.vector_stores.delete(vector_store_id=vector_store_id)
+        except (openai.NotFoundError, ValueError) as e:
+            logger.exception("Vector store %s not found in OpenAI", vector_store_id)
```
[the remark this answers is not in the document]
SmittieC: Good idea. https://github.com/dimagi/open-chat-studio/pull/1560/commits/163f28633a9aea026bd0dafdb5fe413bfbe08dd8

apps/documents/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+    collection_files_to_update = []
+    vector_store_manager = VectorStoreManager(client)
+
+    for collection_file in collection_files:
+        try:
+            remote_file_id = create_files_remote(client, files=[collection_file.file])
+            collection_file.status = FileStatus.COMPLETED
+            file_ids.extend(remote_file_id)
+        except Exception as e:
+            logger.exception(f"Failed to upload file {collection_file.file.id} to OpenAI: {e}")
```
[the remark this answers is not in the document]
SmittieC: Good to know, thanks

apps/documents/views.py
```diff
@@ -175,6 +228,19 @@ def delete(self, request, team_slug: str, pk: int):
             )
             return HttpResponse(response, headers={"HX-Reswap": "none"}, status=400)
         else:
+            if collection.is_index and collection.openai_vector_store_id:
+                try:
+                    manager = VectorStoreManager.from_llm_provider(collection.llm_provider)
+                    manager.delete_vector_store(collection.openai_vector_store_id)
```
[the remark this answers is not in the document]
SmittieC: Yeah, I made a PR [here](https://github.com/dimagi/open-chat-studio/pull/1571) that does some more things regarding switching providers

apps/documents/views.py
```diff
@@ -119,6 +140,29 @@ def get_form_kwargs(self):
...
+                    LlmProvider.objects.get(id=old_llm_provider_id)
+                )
+                old_provider_manager.delete_vector_store(collection.openai_vector_store_id, fail_silently=True)
+
+            manager = VectorStoreManager.from_llm_provider(collection.llm_provider)
+            collection.openai_vector_store_id = manager.create_vector_store(
+                name=f"collection-{collection.team.slug}-{collection.name}-{collection.id}"
+            )
+            collection.save(update_fields=["openai_vector_store_id"])
+            messages.success(self.request, "Collection Created")
```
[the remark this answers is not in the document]
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/1560#discussion_r2081737531

templates/documents/single_collection_home.html
SmittieC: Ah yes, thanks. https://github.com/dimagi/open-chat-studio/pull/1560/commits/70df13e715d5c89e83d31ad7cfabf23967c0fd33

apps/service_providers/llm_service/runnables.py
SmittieC: Since we're not invking the llm directly like the docs does, it will not work. The [AgentExecuter.steam](https://api.python.langchain.com/en/latest/agents/langchain.agents.agent.AgentExecutor.html#langchain.agents.agent.AgentExecutor.stream) returns an [AddableDict](https://api.python.langchain.com/en/latest/runnables/langchain_core.runnables.utils.AddableDict.html#langchain_core.runnables.utils.AddableDict)

> Side note: Are we going to support annotations etc in a future PR?
Correct. [This](https://github.com/dimagi/open-chat-studio/issues/1424) is the ticket for this

apps/assistants/sync.py
```diff
@@ -548,3 +549,51 @@ def get_and_store_openai_file(client, file_id: str, team_id: int) -> File:
...
+        if chunk_size and chunk_overlap:
+            chunking_strategy = {
+                "type": "static",
+                "static": {"max_chunk_size_tokens": chunk_size, "chunk_overlap_tokens": chunk_overlap},
+            }
+
+        for chunk in chunk_list(file_ids, 500):
+            self.client.vector_stores.file_batches.create(
+                vector_store_id=vector_store_id, file_ids=chunk, chunking_strategy=chunking_strategy
+            )
```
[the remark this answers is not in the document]
SmittieC: Probably yeah. I'll look into it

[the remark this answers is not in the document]
SmittieC: How's [this](https://github.com/dimagi/open-chat-studio/pull/1571/commits/2efec70efd57170f33774c9f9fd7ef37030550b0)? It is part of [this PR](https://github.com/dimagi/open-chat-studio/pull/1571) though

apps/pipelines/nodes/nodes.py
```diff
@@ -315,6 +335,11 @@ def _process(self, input, state: PipelineState, node_id: str) -> PipelineState:
 
         node = Node.objects.get(flow_id=node_id, pipeline__version_number=pipeline_version)
         tools = get_node_tools(node, session, attachment_callback=history_manager.attach_file_id)
+        if self.collection_index_id:
```
SmittieC, line 338: cc @bahack55

apps/assistants/sync.py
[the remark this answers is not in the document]
snopoke: this is fine for now but once we have multiple I think the other approach is better.

snopoke (approved):

apps/documents/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+
+from apps.assistants.sync import OpenAIVectorStoreManager, create_files_remote
+from apps.documents.models import CollectionFile, FileStatus
+
+logger = logging.getLogger("ocs.documents.tasks.upload_files_to_openai")
+
+
+@shared_task(base=TaskbadgerTask, ignore_result=True)
+def upload_files_to_vector_store_task(collection_file_ids: list[int], chuking_strategy: dict):
+    CollectionFile.objects.filter(id__in=collection_file_ids, status=FileStatus.PENDING).update(
```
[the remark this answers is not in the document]
SmittieC: We could, yeah. I'll make the change in [this PR](https://github.com/dimagi/open-chat-studio/pull/1571)

PR #1564, opened 2025-05-07
Sk/event bugs

snopoke:
## Description
Resolves [OCSBUGS-9](https://dimagi.atlassian.net/browse/OCSBUGS-9) and [OCSBUGS-11](https://dimagi.atlassian.net/browse/OCSBUGS-11)

* When firing triggers, filter static and timeout triggers to only those linked to published chatbot versions
* When firing scheduled messages, load the action params from the published version of the trigger

One issue with these changes is that it makes it impossible to test events without publishing them first. I'm not going to try and address this in the current PR.

## User Impact
* Only triggers from published chatbots will get executed
* Scheduled messages will use params from the published version instead of the version they were created from
* Events can only be tested with published chatbots

### Docs and Changelog
https://github.com/dimagi/open-chat-studio-docs/pull/100

[OCSBUGS-9]: https://dimagi.atlassian.net/browse/OCSBUGS-9?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OCSBUGS-11]: https://dimagi.atlassian.net/browse/OCSBUGS-11?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/events/models.py
+++ b/apps/events/models.py
@@ -6,7 +6,6 @@ import pytz
 from dateutil.relativedelta import relativedelta
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.contenttypes.models import ContentType
-from django.core.exceptions import ObjectDoesNotExist
 from django.db import models, transaction
 from django.db.models import F, Func, OuterRef, Q, Subquery, functions
 from django.utils import timezone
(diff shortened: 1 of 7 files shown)


snopoke:
@coderabbit review

apps/events/models.py
```diff
@@ -496,14 +497,14 @@ def published_action(self):
             trigger = self.action.static_trigger
             trigger = StaticTrigger.objects.get_published_version(trigger)
             return trigger.action
-        except StaticTrigger.DoesNotExist:
+        except ObjectDoesNotExist:
```
SmittieC, line 500: Interesting. Why this change?

SmittieC (dismissed):
Thanks for this

apps/events/models.py
snopoke: I forget why I changed that but I think I'll change it back

PR #1571, opened 2025-05-09
RAG - Migrate providers

SmittieC:
## Description

Switching providers 1. shows a warning message to users and 2. re-uploads the files to the new provider. At the end, the previous vector store and files will be removed from that account

## User Impact

See above.

### Demo

demo

### Docs and Changelog

Pending


apps/assistants/sync.py
```diff
@@ -592,8 +599,15 @@ def link_files_to_vector_store(
...
-            self.client.vector_stores.file_batches.create(
-                vector_store_id=vector_store_id, file_ids=chunk, chunking_strategy=chunking_strategy
+        try:
+            for chunk in chunk_list(file_ids, 500):
+                self.client.vector_stores.file_batches.create(
+                    vector_store_id=vector_store_id, file_ids=chunk, chunking_strategy=chunking_strategy
+                )
+            return vector_store_id
+        except Exception as e:
+            logger.exception(
```
snopoke, line 609: Maybe drop this to a warning?

apps/documents/models.py
```diff
@@ -83,6 +83,10 @@ class Meta:
     def __str__(self) -> str:
         return self.name
 
+    @property
+    def index_name(self) -> str:
+        return f"collection-{self.team.slug}-{self.name}-{self.id}"
```
snopoke, line 88: nit: slugify name:
```suggestion
        return f"collection-{self.team.slug}-{slugify(self.name)}-{self.id}"
```

apps/documents/tasks.py
```diff
@@ -41,10 +98,33 @@ def upload_files_to_vector_store_task(collection_file_ids: list[int], chuking_st
```
snopoke, line 98: If there was an error it could be worth recording it on the file to show to the user.

I'm thinking there could be two kinds of errors, temporary ones like API down / rate limits etc. and permanent ones like unable to read file or bad file type.

For the temporary ones the user could retry the indexing.

apps/documents/tests/test_views.py
```diff
@@ -0,0 +1,72 @@
...
+            {
+                "name": new_name,
+                "is_index": True,
+                "llm_provider": collection.llm_provider.id,
+            },
+        )
+
+        assert response.status_code == 302
+        collection.refresh_from_db()
+        assert collection.name == new_name
```
snopoke, line 72: could check that `migrate_vector_stores` is not called here

snopoke (approved):

apps/documents/tests/test_views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1571/commits/46f8b7d906e6aa2e468196c59501d0c7bf9bf5b4

apps/documents/tasks.py
SmittieC: I think I'll do this in a separate PR. I'll create an issue for this

apps/assistants/sync.py
SmittieC: Sure. https://github.com/dimagi/open-chat-studio/pull/1571/commits/9992537797335f1990c35d78a1af45df57f3cb54

snopoke (approved):

PR #1577, opened 2025-05-12
remove deprecated .Client call

barora05:
## Description
Gemini didn't work for me even after downgrading python version to 3.11. genai.Client() is deprecated and needs to be updated for google-generativeAI ersion > 0.8.4


barora05:
Closing this PR for now as it works sometimes but still gives .client attribute not found error some times

Issue #1559, opened 2025-05-06
Chatbot Setting UI Tab


stephherbers:
Goal, the user should be able to view and edit the global configs from the chatbot homepage.

Task:
- add "Settings" tab
- make view containing:
    - Description
    - Voice 
    - Tracing
    -Consent
    -Surveys
    -Allowlist
    -Seed message
Then, using htmx, the items on the page should be editable to update the configurations

Mockup: 

<img width="988" alt="Image" src="https://github.com/user-attachments/assets/3de91753-bc15-40c9-ad6d-7205c443ea3e" />

snopoke:
@stephherbers I think it'll be simpler to have one form for editing all of these together instead of individual field edits.

stephherbers:
@snopoke updated 👍

PR #1578, opened 2025-05-12
Chatbot Settings UI

stephherbers:
## Description

Resolves [#1559](https://github.com/dimagi/open-chat-studio/issues/1559)

image

Unblocks the rest of the experiments -> chatbots transition work

## User Impact

Users are able to view and edit the chatbot configurations form the chatbot homepage

### Demo

[loom video](https://www.loom.com/share/80199406233a42c79e1ff761d75ae52f?sid=155c63c2-dfc5-4481-aca9-1ce0f6569722)

### Docs and Changelog

changelog


stephherbers:
@code rabbit please review

templates/chatbots/settings_content.html
```diff
@@ -0,0 +1,260 @@
+<div id="settings-wrapper">
+  <div class="flex justify-end mb-6">
+    {% if edit_mode %}
+      <div class="flex gap-2">
+        <button id="save-all-button"
+                class="btn btn-success btn-sm"
+                hx-post="/a/{{ request.team.slug }}/chatbots/{{ experiment.id }}/settings/save-all/"
```
snopoke, line 7: Use `url` template tag for urls:

```suggestion
                hx-post="{% url 'chatbots:chatbots_settings_edit_mode' request.team.slug experiment.id %}"
```

apps/chatbots/urls.py
```diff
@@ -65,4 +65,19 @@
         views.chatbot_chat_embed,
         name="chatbot_chat_embed",
     ),
+    path(
+        "<int:experiment_id>/settings/edit-mode/",
+        views.settings_edit_mode,
+        name="chatbots_settings_edit_mode",
```
snopoke, line 71: nit: drop the `chatbots_` prefix from urls since they are already namespaced: `chatbots:chatbots_settings_edit_mode`
```suggestion
        name="settings_edit_mode",
```

Same below.

snopoke (commented):
This looks great. One thing I would suggest is to rather use a Django form for the field rendering and for the data processing. It will simplify the views and the templates. 

You can still render individual fields (or selected fields) where you want them in the UI:

```
{% render_form_fields form fieldA %}
{% render_form_fields form fieldA fieldB %}
```

apps/chatbots/views.py
```diff
@@ -38,6 +40,129 @@
...
+def settings_edit_mode(request, team_slug, experiment_id):
+    if request.team.slug != team_slug:
+        return HttpResponse("Unauthorized", status=403)
+
+    experiment = get_object_or_404(Experiment, id=experiment_id, team=request.team)
+    available_voice_providers = request.team.voiceprovider_set.all()
+    available_synthetic_voices = SyntheticVoice.get_for_team(request.team)
+    available_trace_providers = request.team.traceprovider_set.all()
+    available_consent_forms = request.team.consentform_set.exclude(is_version=True)
+    available_surveys = request.team.survey_set.exclude(is_version=True)
```
snopoke, line 55: These can be removed now

```suggestion
```

apps/chatbots/views.py
```diff
@@ -38,6 +40,129 @@
...
+        "request": request,
+        "edit_mode": False,
+    }
+    return HttpResponse(render_to_string("chatbots/settings_content.html", context, request=request))
+
+
+@login_and_team_required
+@permission_required("experiments.change_experiment", raise_exception=True)
+@require_POST
+def save_all_settings(request, team_slug, experiment_id):
```
snopoke, line 93: you should also use the form here

apps/chatbots/forms.py
```diff
@@ -40,6 +40,59 @@ def save(self, commit=True):
...
 
+class ChatbotSettingsForm(forms.ModelForm):
+    description = forms.CharField(widget=forms.Textarea(attrs={"rows": 2}), required=False)
+    seed_message = forms.CharField(widget=forms.Textarea(attrs={"rows": 2}), required=False)
+
+    class Meta:
+        model = Experiment
+        fields = [
+            "description",
+            "voice_provider",
```
snopoke, line 51: The querysets for these fields need to be filtered according to the team: e.g. https://github.com/dimagi/open-chat-studio/blob/9fd6e85e77e0408c4d6453050aa01385cdd53305/apps/experiments/forms.py#L186-L198

snopoke (commented):
I thought this was basically done until I realized that the there is a bit more to do with JS context needed by the form.

apps/chatbots/views.py
```diff
@@ -39,6 +41,62 @@
...
 
+@login_and_team_required
+@permission_required("experiments.change_experiment", raise_exception=True)
+@require_GET
+def settings_edit_mode(request, team_slug, experiment_id):
+    if request.team.slug != team_slug:
+        return HttpResponse("Unauthorized", status=403)
+
+    experiment = get_object_or_404(Experiment, id=experiment_id, team=request.team)
+    form = ExperimentForm(request=request, instance=experiment)
```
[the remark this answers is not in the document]
stephherbers: added the allowlist widget, but still getting errors with the synthetic voice-- here is the draft: https://github.com/dimagi/open-chat-studio/pull/1578/commits/4e82dd2785349c2b4c83a2f6df1b4bd11b634976. Getting a 2 alpine js errors of synthetic_voice and one other not being defined

templates/chatbots/settings_content.html
```diff
@@ -0,0 +1,249 @@
...
+        //first instance
+          }
+          window.tomSelectInstance = null;
+        }
+        let element = document.getElementById('allowlist-multiselect');
+        let hiddenInput = document.getElementById('id_participant_allowlist');
+
+        if (element && hiddenInput) {
+          window.tomSelectInstance = new TomSelect(element, {
+            plugins: ["remove_button", "caret_position", "input_autogrow"],
```
snopoke, line 211: `plugins` key is duplicated below

apps/chatbots/forms.py
```diff
@@ -40,6 +41,73 @@ def save(self, commit=True):
...
 
+class ChatbotSettingsForm(forms.ModelForm):
+    description = forms.CharField(widget=forms.Textarea(attrs={"rows": 2}), required=False)
+    seed_message = forms.CharField(widget=forms.Textarea(attrs={"rows": 2}), required=False)
+    participant_allowlist = forms.CharField(widget=forms.HiddenInput(), required=False)
+
+    class Meta:
+        model = Experiment
+        fields = [
+            "description",
```
snopoke, line 52: might as well add `name` in here too

apps/chatbots/views.py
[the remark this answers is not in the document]
snopoke: The issue is that the alpine data isn't being initialised. I think the options are to either move it to the main template so that it's always initialised or to put it inline with the form element.

I think the simplest approach is to put it directly on the form element:

```
<form x-data="{
          type: '{{ experiment_type }}',

          voiceProvider: null,
          synthetic_voice: {{experiment.synthetic_voice_id|default:'null'}},
          synthetic_voice_options: [],

          init() {

            this.$watch('voiceProvider', () => {
              const providerType = voiceProviderTypes[this.voiceProvider];
              this.synthetic_voice_options = voiceOptions.filter(option => {
                return option.type === providerType && (!option.provider_id || option.provider_id == this.voiceProvider);
              });
            });
          }
        }">
```

templates/chatbots/settings_content.html
```diff
@@ -175,3 +188,62 @@ <h4 class="text-md font-bold mb-2">Seed Message</h4>
...
+    <script>
+      (function() {
+        if (window.tomSelectInstance) {
+          try {
+            window.tomSelectInstance.destroy();
+          } catch (e) {
+        //first instance
+          }
+          window.tomSelectInstance = null;
+        }
```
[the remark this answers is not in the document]
stephherbers: Yes-- otherwise it works fine on the first click of edit, but on the second it throws errors

templates/chatbots/settings_content.html
```diff
@@ -223,27 +250,6 @@ <h4 class="text-md font-bold mb-2">Seed Message</h4>
           });
         }
       })();
-      const voiceProviderTypes = JSON.parse(document.getElementById("voiceProviderTypes").textContent);
-      const voiceOptions = JSON.parse(document.getElementById("voiceOptions").textContent);
```
snopoke: Aren't these still necessary?

stephherbers: fix ended up being a bit more involved but confirmed locally it's working: https://github.com/dimagi/open-chat-studio/pull/1578/commits/5be53ba6733cb1934d437c4ea98e7d6c597600d2

templates/chatbots/settings_content.html
```diff
@@ -0,0 +1,264 @@
...
+        {% endif %}
+      </div>
+    {% endif %}
+  </div>
+
+  {% if edit_mode %}
+    <div style="display: none;">
+      <div id="voiceProviderTypes">{{ voice_providers_types|safe }}</div>
+      <div id="voiceOptions">{{ synthetic_voice_options|safe }}</div>
+    </div>
```
snopoke, line 45: Rather than using a div here you could use the `json_script` tag which is designed for doing exactly this: https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#json-script

It would also mean you don't have to convert the data to JSON first, a normal dict will be fine.

apps/chatbots/views.py
```diff
@@ -39,6 +42,96 @@
...
+                "type": voice.service.lower(),
+                "provider_id": voice.voice_provider_id,
+            }
+            for voice in SyntheticVoice.get_for_team(request.team, exclude_services=exclude_services)
+        ],
+        key=lambda v: v["text"],
+    )
+    return {
+        "voice_providers_types": json.dumps(voice_provider_dict),
+        "synthetic_voice_options": json.dumps(synthetic_voices),
```
snopoke, line 67: See comment below about using `json_script` and not dumping these to JSON here.

templates/chatbots/settings_content.html
snopoke, line 1: This template feels like it could benefit from some template partials to reduce the repetition in some places.

snopoke (approved):
Some small comments but nothing blocking.

PR #1579, opened 2025-05-13
add openai web search built in tool

### LINKED ISSUE #1555: Add support for builtin tools (gemini & openai)
* search
* code execution

https://python.langchain.com/docs/integrations/chat/google_generative_ai/#built-in-tools
https://python.langchain.com/docs/integrations/chat/openai/#built-in-tools

barora05:
## Description
Add web-searhc built in tool for OPENAI. Closes #1555


apps/pipelines/nodes/nodes.py
```diff
@@ -335,6 +341,10 @@ def _process(self, input, state: PipelineState, node_id: str) -> PipelineState:
 
         node = Node.objects.get(flow_id=node_id, pipeline__version_number=pipeline_version)
         tools = get_node_tools(node, session, attachment_callback=history_manager.attach_file_id)
+        built_in_tools = node.params.get("built_in_tools")
```
snopoke, line 344:
```suggestion
        built_in_tools = self.built_in_tools
```

apps/experiments/models.py
```diff
@@ -520,6 +520,15 @@ class VoiceResponseBehaviours(models.TextChoices):
     NEVER = "never", gettext("Never")
 
 
+class BuiltInTools(models.TextChoices):
```
snopoke, line 523: Instead of using `TextChoices` you could use a dataclass enum so you could store more info e.g. provider_type. You can also add methods to the enum:

```python
@dataclasses.dataclass
class BuiltinToolType:
    type: str
    label: str
    provider_type: LlmProviderType


class BuiltinTools(BuiltinToolType, Enum):
    OPENAI_WEB_SEARCH = "openai_web_search", "Web Search", LlmProviderType.openai
    GEMENI_WEB_SEARCH = "google_web_search", "Google Search", LlmProviderType.gemini

    def get_tool_definition(self):
        match self:
            case BuiltinTools.OPENAI_WEB_SEARCH:
                 return {"type": "web_search_preview"}
            case BuiltinTools.GEMENI_WEB_SEARCH:
                 from google.ai.generativelanguage_v1beta.types import Tool as GenAITool
                 return GenAITool(google_search={})
```

barora05: If we want to keep all info here, should I remove attach_built_in_tools from llm_service?

apps/pipelines/nodes/nodes.py
barora05: done

apps/experiments/models.py
barora05: As discussed with @snopoke, keeping this as it is now i.e using attach_built_in_tools to append built in tools. We can implement this when we use a new widget on UI which shows different options based on the llm provider

apps/service_providers/llm_service/main.py
```diff
@@ -168,6 +172,15 @@ def transcribe_audio(self, audio: BytesIO) -> str:
         )
         return transcript.text
 
+    def attach_built_in_tools(self, built_in_tools: list[str], tools: list[BaseTool]):
```
SmittieC, line 175: What if `built_in_tools` is a list of type `BuiltInTools`. If we have a mapping in `BuiltInTools` that returns the tool type, it might be a tiny bit easier to manage. Here's what I mean

```python
class BuiltInTools(models.TextChoices):
    WEB_SEARCH = "web-search", gettext("Web Search")
    CODE_EXECUTION = "code-execution", gettext("Code Execution")


    @property
    def tool_type(self) -> str:
        if self == BuiltInTools.WEB_SEARCH:
            return "web-search-preview"
        ...
```

apps/service_providers/llm_service/main.py
```diff
@@ -168,6 +172,15 @@ def transcribe_audio(self, audio: BytesIO) -> str:
...
+        if not built_in_tools:
+            return
+        for tool_name in built_in_tools:
+            if tool_name == "web-search":
+                tools.append({"type": "web_search_preview"})
+            else:
+                raise ValueError(f"Unsupported built-in tool for openai: '{tool_name}'")
```
SmittieC, line 182: A `ValueError` will be raised when `code-execution` is enabled. will it not?

apps/pipelines/nodes/nodes.py
```diff
@@ -335,6 +341,10 @@ def _process(self, input, state: PipelineState, node_id: str) -> PipelineState:
 
         node = Node.objects.get(flow_id=node_id, pipeline__version_number=pipeline_version)
         tools = get_node_tools(node, session, attachment_callback=history_manager.attach_file_id)
+        built_in_tools = self.built_in_tools
+        llm_service = self.get_llm_service()
+        if llm_service:
```
SmittieC, line 346: Not necessary to change, but just as an FYI, you can do:
```python
if llm_service := self.get_llm_service():
   ...
```

instead of 

```python
llm_service = self.get_llm_service()
if llm_service:
   ...
```

apps/service_providers/llm_service/token_counters.py
```diff
@@ -80,7 +80,7 @@ def __post_init__(self):
         if not self.google_api_key:
             raise ValueError("KEY not found!")
 
-        self.client = genai.Client(api_key=self.google_api_key)
+        self.client = genai
```
SmittieC, line 83: Why was this necessary?

apps/service_providers/llm_service/main.py
SmittieC: Oh I'm only now seeing [this comment](https://github.com/dimagi/open-chat-studio/pull/1579/files#r2086236993) that is similar

apps/service_providers/llm_service/main.py
barora05: yes. OpenAI does not support code execution tool as of now.

apps/service_providers/llm_service/token_counters.py
barora05: genai.Client() is deprecated. For me it gives, genai does not have any Client attribute

apps/pipelines/nodes/nodes.py
barora05: done

apps/service_providers/llm_service/main.py
SmittieC: The user will be able to choose that option, which will then always break. Maybe we should add it only when OpenAI supports code interpreter

apps/service_providers/llm_service/token_counters.py
SmittieC: Interesting that it works, since genai is a module

apps/service_providers/llm_service/main.py
```diff
@@ -253,4 +256,12 @@ def get_callback_handler(self, model: str) -> BaseCallbackHandler:
         return TokenCountingCallbackHandler(GeminiTokenCounter(model, self.google_api_key))
 
     def attach_built_in_tools(self, built_in_tools: list[str], tools: list[BaseTool]):
-        pass
+        if not built_in_tools:
+            return
```
snopoke: as long as you always pass in a list, this check isn't really necessary

apps/service_providers/llm_service/main.py
snopoke, line 175: Instead of passing tools in here it's better to keep functions stateless where possible. So you should return the tool value and let the caller append it.

snopoke (approved):

apps/service_providers/llm_service/token_counters.py
snopoke: This change results in the following error:

Error in TokenCountingCallbackHandler.on_llm_end callback: AttributeError("module 'google.generativeai' has no attribute 'count_tokens'")

I'm working on a fix

snopoke (changes requested):

snopoke (dismissed):

snopoke (dismissed):

SmittieC (approved):
LGTM

PR #1582, opened 2025-05-13
add nodeID and icon to nodes

snopoke:
## Description
Add node ID in muted text and node icon to pipeline nodes.

## User Impact
User's can see node IDs which is useful when looking at trace data. Node icons also give an indication of the node type that's easy to see.

### Demo
image

### Docs and Changelog
TODO

### DIFF
--- a/apps/pipelines/nodes/base.py
+++ b/apps/pipelines/nodes/base.py
@@ -321,6 +321,7 @@ class UiSchema(BaseModel):
 
 class NodeSchema(BaseModel):
     label: str
+    icon: str = None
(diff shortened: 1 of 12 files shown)


stephherbers (approved):

SmittieC (approved):
Nice! Would it be easy/quick to add a copy to clipboard option as well? Not super sure if it's going to be useful, so maybe not necessary now

snopoke:
> Nice! Would it be easy/quick to add a copy to clipboard option as well? Not super sure if it's going to be useful, so maybe not necessary now

This is probably a temporary measure until traces show node names (coming soon).

PR #1584, opened 2025-05-14
store the node ID and django node model object on the Pipeline node instance

snopoke:
## Description
Storing the node ID and django node object on the pipeline node instance makes it easier to reference them without doing a DB query or having to pass the `node_id` to all the function calls.

I would like to give 'experiment' and 'session' the same treatment but I'll leave that for a future PR.

## User Impact
None


SmittieC (approved):
Thanks

stephherbers (approved):

PR #1587, opened 2025-05-14
RAG - Versioning

SmittieC:
## Description

- Improvements to collection versioning
- When an index is versioned, we create a new vector store at OpenAI (the index backend if you will) for this collection and re-upload the files to that vector store, effectively creating a vector store version. These new instances are then linked to the new version.
- When the collection is deleted, we prevent the user from doing to in two cases: 1. when the collection is being used by a pipeline and 2. when the collection is being used by a pipeline version / experiment version.

## Testing
Apart from the unittests, I manually tested:
- that when a collection is versioned, it creates a copy of the file
- if it's an indexed collection, a new vector store is created at the provider, the files are uploaded and linked to the new vector store and that that the file versions gets new external ids. Basically, I confirmed that a copy of the current vector store is made.
- When I remove the collection from the Pipeline while it is being used in a chatbot version, that the user cannot delete the collection, since it is being used by some version. I first had to archive that specific version before I could remove the collection.
- When a collection is archived, its vector store is removed, files are deleted and all associated IDs are cleared on the file. Also, the file is archived.

## User Impact

User will be able to create versions of pipelines with collection indexes

### Demo

N/A

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/103


apps/documents/views.py
```diff
@@ -206,31 +205,45 @@ def form_valid(self, form):
...
+            experiment_chips = []
+            for version in collection.versions.all():
+                if experiments := version.get_related_experiments_queryset():
+                    experiment_chips.extend(
+                        [
+                            Chip(
+                                label=(
+                                    f"{experiment.name} [{experiment.get_version_name()}]"
+                                    if experiment.is_working_version
+                                    else f"{experiment.name} {experiment.get_version_name()} [published]"
```
stephherbers, line 233: the version is not necessarily published, right? we just check `self.versions.exists()` not just for published versions

stephherbers (approved):
everything looks great i like the approach.  just added one comment and curious if you could include what manual testing you did in the PR description

apps/documents/views.py
SmittieC: Correct. Unless I'm forgetting a reason why we should not, I think this should be the case for assistants as well. Reason being that if a non-published version of some chatbot is using a collection, but we ignore that while archiving said collection, we'll be breaking that version without telling the user or archiving that version. I think we should inform users if a resource like a collection or assistant is being used in any version of a chatbot.

apps/documents/models.py
```diff
@@ -113,20 +111,86 @@ def version_details(self) -> VersionDetails:
...
+    def get_related_nodes_queryset(self) -> models.QuerySet:
+        index_references = get_related_pipelines_queryset(self, "collection_index_id").distinct()
+        collection_references = get_related_pipelines_queryset(self, "collection_id").distinct()
+        return index_references | collection_references
+
+    def get_related_experiments_queryset(self) -> models.QuerySet:
+        """
+        Get all experiments that reference this collection through a pipeline
+        """
+        return Experiment.objects.filter(
```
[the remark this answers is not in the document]
SmittieC: See [this comment](https://github.com/dimagi/open-chat-studio/pull/1587#discussion_r2092939826). Was there a reason why we opted for not checking non-published versions again?

apps/documents/views.py
snopoke: I had thought that we'd also archive the version but perhaps we didn't implement that.

As it stands however, this is still incorrect since it's not necessarily a published version because the queryset isn't filtering out unpublished versions.

SmittieC: Correct. So the desired behaviour is to archive non-published versions if they are using this collection? (Same goes for assistants, but we don't yet do that)

apps/documents/models.py
```diff
@@ -113,20 +110,92 @@ def version_details(self) -> VersionDetails:
...
+    def archive(self):
+        """
+        Archive the collection with its files and remove the index and the files at the remote service, if it has one
+        """
+        if self.get_related_nodes_queryset().exists():
+            return False
+
+        if self.is_working_version:
+            for version in self.versions.all():
+                if version.get_related_experiments_queryset().exists():
```
snopoke, line 184: Can't this be done without querying in a loop? We could fetch the version IDs and then run the query with that:

```python
version_ids = self.versions.values_list("id", flat=True)
has_references = Node.objects.filter(params__collection_id__in=version_ids).exists()
```

apps/utils/deletion.py
```diff
@@ -224,6 +224,15 @@ def get_related_pipelines_queryset(instance, pipeline_param_key: str = None):
     return pipelines
 
 
+def get_related_pipeline_experiments_queryset(instance, pipeline_param_key: str):
+    from apps.experiments.models import Experiment
+
+    return Experiment.objects.filter(
```
snopoke, line 230:
```suggestion
    return Experiment.objects.exclude(pipeline=None).filter(
```

SmittieC: Experiments without pipelines are excluded by the filter already?

snopoke: They are excluded implicitly. Its just nice to make it explicit.

apps/documents/models.py
```diff
@@ -163,10 +162,14 @@ def get_related_experiments_queryset(self) -> models.QuerySet:
...
-        return Experiment.objects.filter(
-            models.Q(pipeline__node__params__collection_index_id=str(self.id))
-            | models.Q(pipeline__node__params__collection_id=str(self.id))
-        ).distinct()
+        # TODO: Update assistant archive code to use get_related_pipeline_experiments_queryset
+        index_references = get_related_pipeline_experiments_queryset(self, "collection_index_id").filter(
+            is_default_version=True
+        )
+        collection_references = get_related_pipeline_experiments_queryset(self, "collection_id").filter(
+            is_default_version=True
```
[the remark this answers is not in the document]
snopoke: I was mostly thinking about the UI where we show the list of experiments after rejecting deletion and whether we should show working versions there too. My thinking is that if a node matches in the working version but not in the published version then we won't list that experiment in the dialog.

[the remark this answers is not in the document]
SmittieC: Ah right. So in this case it will show the working version of the Pipeline, since that version is linked to the working version of the experiment.

apps/documents/models.py
SmittieC: Thoughts on https://github.com/dimagi/open-chat-studio/pull/1587/commits/be84bd635bf144cb3b11c13a4d0915b52ce6d77d?

apps/utils/deletion.py
SmittieC: Good point. See https://github.com/dimagi/open-chat-studio/pull/1587/commits/be84bd635bf144cb3b11c13a4d0915b52ce6d77d

SmittieC:
> I'm honestly having a hard time figuring out what the correct approach to archiving is when versions are involved.

Mind elaborating?

snopoke:
> > I'm honestly having a hard time figuring out what the correct approach to archiving is when versions are involved.
> 
> Mind elaborating?

* Should we check if any version of model X is referenced or only check if the working version is referenced?
* Should we check for references in any version of the referencing models or only the working / published version?

PR #1590, opened 2025-05-15
Small UI updates

SmittieC:
## Description

- Updated the LLM provider dropdown in the LLM node to show the name of the provider when there are multiple providers of the same type.
- The error message that is shown when different providers are used for a collection index and LLM node is updated to show which provider is expected to be used on the node. In the future we can change it to either 1. auto select the node provider based on the collection index and lock it in, or 2. change the set of available options in the collection index dropdown, based on the selected node provider.

### User impact
Users will know exactly 1. from which provider instance they selected a model from and 2. which provider they should be selecting a model from if there is a mismatch between the provider that the collection index is using and the one currently selected.

#### 1
Here you can see what it looks like when I have 1 Anthorpic provider and 2 OpenAI providers. Only the OpenAI ones have their provider names appended.

image

#### 2
This is what the error message looks like when there's a mismatch between the node and collection providers

image

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/101

### DIFF
--- a/apps/pipelines/nodes/nodes.py
+++ b/apps/pipelines/nodes/nodes.py
@@ -319,8 +319,10 @@ class LLMResponseWithPrompt(LLMResponse, HistoryMixin):
 
         collection = Collection.objects.get(id=value)
         if collection.llm_provider_id != info.data.get("llm_provider_id"):
+            info.data.get("llm_provider_id")
             raise PydanticCustomError(
-                "invalid_collection_index", "The collection index must use the same LLM provider as the node"
+                "invalid_collection_index",
+                f"The collection index and node must use the same LLM provider ({collection.llm_provider.name})",
             )
         return value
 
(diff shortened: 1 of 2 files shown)


apps/pipelines/nodes/nodes.py
```diff
@@ -319,8 +319,10 @@ def validate_collection_index_id(cls, value, info: FieldValidationInfo):
 
         collection = Collection.objects.get(id=value)
         if collection.llm_provider_id != info.data.get("llm_provider_id"):
+            info.data.get("llm_provider_id")
```
snopoke, line 322: nit: unused variable

snopoke (approved):

stephherbers (approved):

apps/pipelines/nodes/nodes.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1590/commits/577d907becb71ca580e8a98470fd70ddfd62e2d6

PR #1593, opened 2025-05-15
Bump z-index for team switcher

proteusvacuum:
## Description

I was having trouble switching teams, as the dropdown was rendering under the menu below

## User Impact

User's can switch teams.

#### Before
image

#### After
image


SmittieC (approved):
Interesting that this is an issue

Issue #1591, opened 2025-05-15
Add more tracing to OCS runnable implementations


snopoke:
Need to call the langchain callbacks to get more tracing info in the OCS runnables e.g. `AssistantChat`

https://github.com/dimagi/open-chat-studio/blob/e7b259d21414a1d4a89f78d26f1dd28c292b4f69/apps/service_providers/llm_service/main.py#L35-L41

PR #1599, opened 2025-05-16
dispatch event with assistant details for tracing

snopoke:
## Description
Resolves https://github.com/dimagi/open-chat-studio/issues/1591

## User Impact
Easier access to assistant details from tracing data.

### DIFF
--- a/apps/service_providers/llm_service/runnables.py
+++ b/apps/service_providers/llm_service/runnables.py
@@ -9,6 +9,7 @@ from django.db import transaction
 from langchain.agents import AgentExecutor, create_tool_calling_agent
 from langchain.agents.openai_assistant.base import OpenAIAssistantFinish
 from langchain_core.agents import AgentFinish
+from langchain_core.callbacks import dispatch_custom_event
 from langchain_core.load import Serializable
 from langchain_core.messages import BaseMessage
 from langchain_core.messages.tool import ToolMessage, tool_call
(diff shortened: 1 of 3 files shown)


apps/service_providers/llm_service/runnables.py
```diff
@@ -330,12 +331,21 @@ def invoke(
...
+                    "run_id": run_id,
+                },
+            )
             ai_message, annotation_file_ids = self._get_output_with_annotations(thread_id, run_id)
             ai_message_metadata = self.adapter.get_output_message_metadata(annotation_file_ids)
 
             if not current_thread_id:
-                self.adapter.thread_id = thread_id
-
+                self.adapter.update_thread_id(thread_id)
```
SmittieC, line 338: Today I learned that one can return inside the try and the `finally` block still gets executed.

apps/service_providers/tracing/langfuse.py
```diff
@@ -213,3 +214,29 @@ def shutdown(self):
 def _shutdown():
     """Shutdown the client manager when the program exits."""
     client_manager.shutdown()
+
+
+class LangfuseCallbackHandler(CallbackHandler):
```
SmittieC, line 219: I'm curious, if we do custom tracing, we'll be able to pick up this event as well?

SmittieC (approved):

PR #1600, opened 2025-05-16
Ba/UI widget tools

barora05:
## Description
Widget for built in tools and config. I have tried to keep the config rendering dynamic i.e we can select which widget to load for the config if there are diff types of config in the future. #1555 

### Demo 
[Demo](https://drive.google.com/file/d/1VvHG9bCiYmUh_ueQA867l0eutgeLFtLq/view?usp=drive_link)


barora05:
I have tried to add allowed_domains and blocked_domains in the LLMResponseWithPrompt and not show them on the UI using exclude = True,
```
allowed_domains: str | None = Field(default=None, exclude=True)
blocked_domains: str | None = Field(default=None, exclude=True)
```
 but this didn't work. If there is a way to not show them on the UI, I would be able to reduce hardocding at couple of places

PR #1601, opened 2025-05-16
Rename 'channels' app to 'ocs_channels'

snopoke:
## Description
This renames the 'channels' app in preparation from adding the Django [channels](https://channels.readthedocs.io/en/latest/installation.html) app to avoid a name clash.

The approach I've gone with is to keep the DB schema untouched but to change the app label. This is possible by setting the `db_table` on the model `Meta` class.

This will require a maintenance window in order to update references to the app label in the database (migrations and content type).


apps/channels/migrations/0022_alter_experimentchannel_table.py
```diff
@@ -0,0 +1,21 @@
...
+    dependencies = [
+        ('ocs_channels', '0021_alter_experimentchannel_platform'),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            state_operations=[
+                migrations.AlterModelTable(
+                    name='experimentchannel',
+                    table='channels_experimentchannel',
```
SmittieC, line 17: Wouldn't the new channels app prefix its tables with "channels_"?

snopoke: no, it will use `ocs_channels` since that's the app name. So since we changed the app name and also set `db_table` on the model, Django things we've renamed the model table. But since we haven't, we need to use `SeparateDatabaseAndState` to let Django update it's internal state without actually modifying the DB.

SmittieC: Oh I meant the Django Channels library. Isn't that prefixing its tables with `channels_`? So shouldn't we rename our tables as well to be prefixed with `ocs_channels_`?

snopoke: Oh right. The Django Channels app doesn't have any models. I'll take another look at renaming the tables - that might just be cleaner in the long run anyway.

SmittieC (commented):
LGTM

PR #1602, opened 2025-05-16
management command to rename app reference in DB tables

snopoke:
## Description
Management command extracted from https://github.com/dimagi/open-chat-studio/pull/1601

### DIFF
--- a/apps/web/management/commands/rename_app.py
+++ b/apps/web/management/commands/rename_app.py
@@ -1,0 +1,37 @@
+import logging
+
+from django.core.management.base import BaseCommand
+from django.db import connection
+from django.db.transaction import atomic
+
+logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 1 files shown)


SmittieC:
Is the plan to merge + run this first, then merge + deploy the [rename channels PR](https://github.com/dimagi/open-chat-studio/pull/1601)?

snopoke:
> Is the plan to merge + run this first, then merge + deploy the [rename channels PR](https://github.com/dimagi/open-chat-studio/pull/1601)?

I actually don't think I'll use this at all but I'm committing it for reference. Originally I thought I would run this command and then deploy but that would result in the running service being out of sync with the DB. So the plan is this:

1. Stop services
2. Run the SQL via the DB shell
3. Deploy the updated version of the servics

PR #1606, opened 2025-05-20
RAG - Retry failed uploads

### LINKED ISSUE #1580: RAG - Retry file uploads
See [this comment](https://github.com/dimagi/open-chat-studio/pull/1571#discussion_r2086147829) for more details

SmittieC:
## Description

Resolves #1580 
User will be able to retry failed uploads.

### Open question
We could catch connection / rate limit erros and tell the user to try again later when they hover over the failed status. For any other errors, we could say something like "Failed to upload due to unexpected error" or something. Do we think this lemon is worth the squeeze?

## User Impact

User will be able to retry failed uploads

### Demo

demo

### Docs and Changelog

Pending

### DIFF
--- a/templates/documents/single_collection_home.html
+++ b/templates/documents/single_collection_home.html
@@ -44,9 +44,19 @@
     <div class="flex flex-col gap-4">
       <div class="flex justify-between items-center">
         <h2 class="text-lg font-semibold">Files</h2>
-        <button class="btn btn-sm btn-primary" onclick="chooseFilesModal.showModal()">
-          <i class="fa-solid fa-plus"></i> Add Files
-        </button>
+        <div class="join">
+          {% if collection.has_failed_index_uploads %}
+            <form class="join-item" action="{% url 'documents:retry_failed_uploads' team.slug collection.id %}" method="POST">
+              {% csrf_token %}
+              <button class="btn btn-sm btn-primary">
+                <i class="fa-solid fa-repeat"></i> Retry Failed Uploads
+              </button>
+            </form>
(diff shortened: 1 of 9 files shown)


templates/documents/single_collection_home.html
```diff
@@ -44,9 +44,19 @@ <h1 class="pg-title">{{ collection.name }} {% if collection.is_index %}<span cla
     <div class="flex flex-col gap-4">
       <div class="flex justify-between items-center">
         <h2 class="text-lg font-semibold">Files</h2>
-        <button class="btn btn-sm btn-primary" onclick="chooseFilesModal.showModal()">
-          <i class="fa-solid fa-plus"></i> Add Files
-        </button>
+        <div class="join">
+          {% if collection.has_failed_index_uploads %}
+            <form class="join-item" action="{% url 'documents:retry_failed_uploads' team.slug collection.id %}" method="POST">
```
snopoke, line 49: the user could click this multiple times and kick of multiple tasks - might be worth storing the task ID somewhere and checking it

SmittieC: I think I'm going to change the way I did it slightly by passing in the collection files that I want to upload instead. This way, the caller will decide which files should be uploaded, preventing possible clashes if an upload and retry task is running at the same time

SmittieC: See https://github.com/dimagi/open-chat-studio/pull/1606/commits/c5295170c0d5f88da2f5527772f19055b0e250dc

Issue #1451, opened 2025-04-14
Refactor history compression code


SmittieC:
We need to refactor the history compression code found [here](https://github.com/dimagi/open-chat-studio/blob/2116049b6d05643456fcb4c8373f6392fae04a7e/apps/chat/conversation.py#L240-L281) and [here](https://github.com/dimagi/open-chat-studio/blob/2116049b6d05643456fcb4c8373f6392fae04a7e/apps/chat/conversation.py#L338-L383). Currently we call a recursive function inside a double nested while-loop.

We migth be able to do away with the firstt while loop, now that we force the summary to fit within the context window.

SmittieC:
See langchain's [map_reduce summarization docs](https://python.langchain.com/docs/how_to/summarize_map_reduce) and trimming. We can replace [our trimming code](https://github.com/dimagi/open-chat-studio/blob/2116049b6d05643456fcb4c8373f6392fae04a7e/apps/chat/conversation.py#L339-L341) to use this instead.

cc @snopoke if you want to add anything

PR #1607, opened 2025-05-20
history compression improvements

snopoke:
## Description
Some various refactors to history compression that was started by looking into [OPEN-CHAT-STUDIO-YR](https://dimagi.sentry.io/issues/6615914890/events/d1a9d80b30a9405e98ae865c3fb0b675/)

This includes part of https://github.com/dimagi/open-chat-studio/issues/1451

### Summary of changes
* short circuit compression if input_tokens > token_limit
* perform simple history truncation in the DB to avoid loading more data into memory than is needed
* record compression markers for `truncate_tokens` mode is being used to avoid loading more messages into memory than is needed

## User Impact
None

### Demo
See test cases

### Docs and Changelog
None


apps/chat/conversation.py
```diff
@@ -223,6 +223,11 @@ def truncate_tokens(history, max_token_limit, llm, input_message_tokens):
 
 
 def summarize_history(llm, history, max_token_limit, input_message_tokens, summary, input_messages, pruned_memory):
+    if input_message_tokens >= max_token_limit:
```
SmittieC, line 254: This reminds me. Now that we have token limits on the models, we can do stuff like prevent users from sending messages where input_message_tokens >= max_token_limit. Shall I make a ticket for this?

SmittieC: See https://github.com/dimagi/open-chat-studio/issues/1615 if you have any comments or concerns

apps/chat/conversation.py
```diff
@@ -158,6 +165,14 @@ def compress_pipeline_chat_history(
     keep_history_len: int = 10,
     history_mode: str = None,
 ) -> list[BaseMessage]:
+    if history_mode == PipelineChatHistoryModes.MAX_HISTORY_LENGTH:
+        limit = keep_history_len // 2 + 1  # each pipeline history message is a pair
```
SmittieC, line 169: Nit: Should the 1 not be added only when `keep_history_len` is uneven?

SmittieC: You should probably also check that the number of returned messages doesn't exceed the token limit

apps/pipelines/tests/test_pipeline_history_summary.py
```diff
@@ -44,6 +44,43 @@ def test_no_summary_returns_all_messages(experiment_session):
...
+    )
+    message3 = history.messages.create(ai_message="I am a robot", human_message="how about some tea")
+    expected_messages = [
+        SystemMessage(content="argument", additional_kwargs={"id": message2.id, "node_id": ""}),
+        HumanMessage(content="sudo, please fetch me a coffee", additional_kwargs={"id": message2.id, "node_id": ""}),
+        AIMessage(content="I can't do that", additional_kwargs={"id": message2.id, "node_id": ""}),
+        HumanMessage(content="how about some tea", additional_kwargs={"id": message3.id, "node_id": ""}),
+        AIMessage(content="I am a robot", additional_kwargs={"id": message3.id, "node_id": ""}),
+    ]
+    summary_messages = history.get_langchain_messages_until_summary()
```
SmittieC, line 62: Doesn't look like the robot respects `sudo` privileges

snopoke:
> Thanks. Does it make sense to respect the compression marker when the history mode is not summarization?

No it doesn't. I did briefly think about this and initially thought it wasn't an issue but I now realize that in pipelines you could be using different modes in different nodes. I'll make some updates to separate them out.

Issue #1610, opened 2025-05-20
Create Chatbot Transition Notification Banners


stephherbers:
see here: https://docs.google.com/document/d/1k3LlbhWac3YmHKlHJCXBLNFeaj3cDFW-nBQEo1PZHGI/edit?tab=t.pcz8tlwgw9y9#heading=h.nd12nd35ipt7 

Add prominent banners on the Experiments and Pipelines pages with clear messaging about:
 - The transition timeline
- Where to find existing content
- Links to updated documentation

PR #1611, opened 2025-05-20
Notification Banners

stephherbers:
## Description

Resolves [#1610](https://github.com/dimagi/open-chat-studio/issues/1610)

Allows the creation of notification banners via django admin 

with title:
Screenshot 2025-05-27 at 5 30 26 PM

Screenshot 2025-05-27 at 5 07 14 PM
without title:
Screenshot 2025-05-27 at 5 08 06 PM

Screenshot 2025-05-27 at 5 09 06 PM

## User Impact

users will see dismissable banner configured by django admins

### Demo

navigate to django admin
create a banner with optional title, markdown message, timeframe, styling
set to it's location (global (all pages) or only on specific pages) 

they also have an is_active flag so they can be created and not released, or removed with out deleting within the specified timeframe

### Docs and Changelog

changelog just as a heads up thy will exist now


apps/banners/middleware.py
```diff
@@ -0,0 +1,23 @@
+import re
+
+
+class BannerLocationMiddleware:
```
snopoke, line 4: you could instead use `request.resolver_match.view_name` which is the name of the view including the namespace e.g. `experiments:experiments_home`


https://docs.djangoproject.com/en/5.2/ref/urlresolvers/#django.urls.ResolverMatch.view_name

templates/banners/banner.html
snopoke, line 1: This needs to be included in a base template still. Probably in `web/base.html` alongside the messages include.

snopoke (commented):
This is neat. Should we include a way to prevent the same banner from showing up for the same user every time they load a page? Maybe once they dismiss it we store a cookie?

templates/banners/banner.html
```diff
@@ -0,0 +1,13 @@
+{% if banners %}
+    <div class="banners-container">
+        {% for banner in banners %}
+            <div class="alert alert-{{ banner.type }} alert-dismissible fade show" role="alert" data-banner-id="{{ banner.id }}">
+                {% if banner.title %}
+                    <strong>{{ banner.title }}</strong>:
+                {% endif %}
+                {{ banner.message }}
```
snopoke, line 8: It would be great to render this as markdown

apps/banners/middleware.py
snopoke: If you're using the view name is it really necessary to have this middleware? You could just use the view name in directly without mapping it to 'banner_location'

templates/banners/banner.html
```diff
@@ -8,18 +8,31 @@
...
         {% endfor %}
     </div>
     <script>
         document.addEventListener('DOMContentLoaded', function() {
-            const closeButtons = document.querySelectorAll('[data-bs-dismiss="alert"]');
-            closeButtons.forEach(function(button) {
+            function updateDismissedBanners(bannerId) {
+                const dismissed = JSON.parse(
+                    document.cookie.match(/dismissed_banners=([^;]*)/)?.[1] || '[]'
+                );
```
snopoke, line 23: You can use `SiteJS.app.Cookies.get('dismissed_banners')` to get the cookie value. You can also use it to set the value. See the docs: https://github.com/js-cookie/js-cookie

apps/banners/context_processors.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Context processor that makes banners available in all templates.
+    Uses the banner_location set by the middleware and excludes dismissed banners.
+    """
+    location = getattr(request, "banner_location", None)
+    context = {"banners": []}
+    if location:
+        location_banners = BannerService.get_banner_context(request=request, location=location)
+        context["banners"].extend(location_banners.get("banners", []))
+    global_banners = BannerService.get_banner_context(request=request, location="global")
```
snopoke, line 14: rather than doing a separate query you could make the location filter and OR filter:

`.filter(Q(location=location) | Q(location="global"))`

apps/banners/services.py
```diff
@@ -0,0 +1,45 @@
...
+
+class BannerService:
+    @staticmethod
+    def get_active_banners(request=None, location=None):
+        now = timezone.now()
+        query = Banner.objects.filter(is_active=True, start_date__lte=now, end_date__gt=now)
+
+        if location:
+            query = query.filter(location=location)
+        if request:
```
snopoke, line 16: I don't think request should be optional.

apps/banners/services.py
```diff
@@ -0,0 +1,45 @@
+import json
+
+from django.utils import timezone
+
+from .models import Banner
+
+
+class BannerService:
+    @staticmethod
+    def get_active_banners(request=None, location=None):
```
snopoke, line 10: This could do with a few small tests

PR #1612, opened 2025-05-21
Ba/llm node tag

### LINKED ISSUE #1143: [design] Allow nodes in a pipeline to create an OCS tag

barora05:
## Description
Tag output message for LLM node. Closes #1143.


apps/pipelines/nodes/base.py
```diff
@@ -317,10 +318,11 @@ def router(state: PipelineState, config: RunnableConfig) -> ReturnType:
             conditional_branch = self._process_conditional(state)
             output_handle = next((k for k, v in output_map.items() if v == conditional_branch), None)
             tags = self.get_output_tags(conditional_branch)
+            tag = self.get_output_tag()
             target_node_id = edge_map[conditional_branch]
             route_path = (state["node_source"], self.node_id, [target_node_id])
             output = PipelineState.from_router_output(
-                self.node_id, self.name, state["node_input"], output_handle, tags, route_path, conditional_branch
+                self.node_id, self.name, state["node_input"], output_handle, tags, route_path, conditional_branch, tag
```
SmittieC, line 325: Rather add `tag` to `tags`

apps/pipelines/nodes/nodes.py
```diff
@@ -284,6 +284,10 @@ class LLMResponseWithPrompt(LLMResponse, HistoryMixin):
         description="Custom actions to enable for the bot",
         json_schema_extra=UiSchema(widget=Widgets.multiselect, options_source=OptionsSource.custom_actions),
     )
+    tag: str = Field(
+        default="Tag to add to output message",
```
SmittieC, line 288: How about this?

```suggestion
        default="The tag that the output message should be tagged",
```

apps/pipelines/nodes/nodes.py
```diff
@@ -380,6 +384,7 @@ def _process(self, input, state: PipelineState) -> PipelineState:
             node_id=self.node_id,
             output=result.output,
             output_message_metadata=history_manager.output_message_metadata,
+            tag=f"{self.name}:{self.tag}" if self.tag and self.tag.strip() else None,
```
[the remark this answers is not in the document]
SmittieC: I see now we do want to use a separate Tag Category for this, and not the `BOT_RESPONSE` one. Hmmm.. @snopoke is this correct?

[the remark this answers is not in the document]
barora05: I have used separate var as while adding system tags the tag category for output_message_tags is BOT_RESPONSE where as we want the tag category as blank for this case

apps/annotations/models.py
```diff
@@ -33,7 +33,7 @@ class Tag(TagBase, BaseTeamModel):
     name = models.CharField(verbose_name=pgettext_lazy("A tag name", "name"), max_length=100)
     created_by = models.ForeignKey("users.CustomUser", on_delete=models.DO_NOTHING, null=True, default=None)
     is_system_tag = models.BooleanField(default=False)
-    category = models.CharField(choices=TagCategories.choices, blank=True, default="")
+    category = models.CharField(choices=TagCategories.choices, blank=True, default="", null=False)
```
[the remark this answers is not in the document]
snopoke: `null=False` is the default value so this change doesn't actually do anything

apps/pipelines/nodes/nodes.py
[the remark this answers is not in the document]
snopoke: I had originally thought we'd reuse output_message_tags and store the tag category along with the value:

```
output_message_tags = [
  ("tag1", None), ("routernode:keyword", TagCategory.BOT_RESPONSE)
]
```

snopoke, line 387: I don't think these tags should have the node name appended to them. I think it should just be the plain value provided by the user.

apps/pipelines/nodes/nodes.py
snopoke: I think this should be on all nodes, not just LLM nodes.

apps/pipelines/nodes/nodes.py
```diff
@@ -56,9 +56,23 @@
 from apps.utils.prompt import OcsPromptTemplate, PromptVars, validate_prompt_variables
 
 OptionalInt = Annotated[int | None, BeforeValidator(lambda x: None if x == "" else x)]
+DEFAULT_TAG = "The tag that the output message should be tagged with"
 
 
-class RenderTemplate(PipelineNode):
+class TagMixin(BaseModel):
```
snopoke, line 62: This could go on the `BasePipelineNode` class which would avoid having to add this mixin on each node.

apps/pipelines/nodes/nodes.py
```diff
@@ -56,9 +56,23 @@
...
+    tag: str = Field(
+        default=DEFAULT_TAG,
+        json_schema_extra=UiSchema(widget=Widgets.expandable_text),
+    )
```
snopoke, line 66: suggested field changes:
```suggestion
    tag: str = Field(
        default="",
        title="Message Tag",
        description="The tag that the output message should be tagged with",
    )
```

apps/pipelines/nodes/nodes.py
barora05: didn't add it there as then it will also be included in start and end node.

apps/pipelines/nodes/nodes.py
barora05: > tag: str = Field(

Tried this. description is not visible as help text inside or outside the field.

snopoke: You need to remove the tag field from the `LlmResponseWithPrompt` node since it's overriding the one in the mixin.

barora05: done

apps/pipelines/models.py
```diff
@@ -341,15 +340,20 @@ def invoke(
             return ChatMessage(content=output)
 
     def _save_message_to_history(
-        self, session: ExperimentSession, message: str, type_: ChatMessageType, metadata: dict, tags: list[str] = None
+        self,
+        session: ExperimentSession,
+        message: str,
+        type_: ChatMessageType,
+        metadata: dict,
+        tags: list[str] = None,
```
SmittieC, line 348:
```suggestion
        tags: list[tuple] = None,
```

apps/pipelines/nodes/base.py
```diff
@@ -79,6 +79,7 @@ class PipelineState(dict):
     output_message_metadata: Annotated[dict, merge_dicts]
     attachments: list = Field(default=[])
     output_message_tags: Annotated[list[str], operator.add]
+    tag: str
```
SmittieC, line 82: This can be removed?

apps/pipelines/models.py
```diff
@@ -341,15 +340,20 @@ def invoke(
...
     ) -> ChatMessage:
         chat_message = ChatMessage.objects.create(
             chat=session.chat, message_type=type_.value, content=message, metadata=metadata
         )
 
         if tags:
-            for tag in tags:
-                chat_message.add_system_tag(tag, TagCategories.BOT_RESPONSE)
+            for tag_value, category in tags:
+                chat_message.add_system_tag(tag_value, category or "")
```
snopoke, line 356: I think we should rename this method to `create_and_add_tag` (`add_tag` is already taken) and only if the category is not empty we can mark it as a system tag.

apps/pipelines/nodes/base.py
```diff
@@ -295,6 +296,11 @@ def process(
         state = self._prepare_state(self.node_id, incoming_edges, state)
         output = self._process(input=state["node_input"], state=state)
         output["path"] = [(state["node_source"], self.node_id, outgoing_edges)]
+        get_output_tags_fn = getattr(self, "get_output_tags", None)
+        if callable(get_output_tags_fn):
+            output["output_message_tags"] = get_output_tags_fn()
```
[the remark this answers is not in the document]
barora05: This is also used in start and end node. This will give error that attribute get_output_tags does not exist

apps/pipelines/nodes/nodes.py
```diff
@@ -56,9 +56,24 @@
...
+class TagMixin(BaseModel):
+    tag: str = Field(
+        default="",
+        title="Message Tag",
+        description="The tag that the output message should be tagged with",
+    )
+
+    def get_output_tags(self) -> list[tuple[str, str | None]]:
+        tags: list[tuple[str, str | None]] = []
+        if self.tag and self.tag != DEFAULT_TAG:
```
snopoke, line 71:
```suggestion
        if self.tag:
```

apps/pipelines/nodes/base.py
barora05: yes

apps/pipelines/nodes/nodes.py
```diff
@@ -56,9 +56,24 @@
...
-class RenderTemplate(PipelineNode):
+class TagMixin(BaseModel):
+    tag: str = Field(
+        default="",
+        title="Message Tag",
+        description="The tag that the output message should be tagged with",
+    )
+
+    def get_output_tags(self) -> list[tuple[str, str | None]]:
+        tags: list[tuple[str, str | None]] = []
```
snopoke, line 70:
```suggestion
    def get_output_tags(self) -> list[tuple[str, None]]:
        tags: list[tuple[str, None]] = []
```

apps/pipelines/nodes/base.py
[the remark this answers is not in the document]
SmittieC: I see. I'm starting to wonder if we shouldn't make a subclass for user editable and non-editable (start + end) nodes. Not for this PR of course.

apps/pipelines/nodes/nodes.py
```diff
@@ -58,7 +58,21 @@
 OptionalInt = Annotated[int | None, BeforeValidator(lambda x: None if isinstance(x, str) and len(x) == 0 else x)]
 
 
-class RenderTemplate(PipelineNode):
+class OutputMessageTagMixin(BaseModel):
+    tag: str = Field(
```
snopoke, line 62: this value should receive the same unicode normalization treatement as in the tag views:

`unicodedata.normalize("NFC", value)`

You can do this with an `after` validator (https://docs.pydantic.dev/dev/concepts/validators/#field-validators)

PR #1618, opened 2025-05-22
API improvement

### LINKED ISSUE #1402: API improvement
Currently the [list experiments](https://chatbots.dimagi.com/api/docs/#tag/Experiments/operation/experiment_list) endpoint returns all experiments, including experiment versions. We need to update this endpoint to only return the "base" experiments, and available versions should be shown in a field inside each experiment object. 

The user should see a list of experiments and it should be clear which one is the published one.

SmittieC:
## Description

Resolves #1402 

Only working versions are returned in the response's toplevel. Inside each working version is a new "versions" field, showing the versions of that object. Here's an example of what the new response looks like:

```json
{
    "next": null,
    "previous": null,
    "results": [
        {
            "id": "95f0f4a8-e43d-47fa-bcf6-62d4e1285306",
            "name": "Tester",
            "url": "http://localhost:8000/api/experiments/95f0f4a8-e43d-47fa-bcf6-62d4e1285306/",
            "version_number": 2,
            "versions": [
                {
                    "name": "Tester",
                    "version_number": 1,
                    "is_default_version": true,
                    "version_description": ""
                }
            ]
        },
        {
            "id": "8481365e-d569-4a33-a918-a8d5ada906ea",
            "name": "Normally",
            "url": "http://localhost:8000/api/experiments/8481365e-d569-4a33-a918-a8d5ada906ea/",
            "version_number": 2,
            "versions": [
                {
                    "name": "Normally",
                    "version_number": 1,
                    "is_default_version": true,
                    "version_description": ""
                }
            ]
        },
        {
            "id": "629cedfb-b74e-4433-9f4a-24a321418013",
            "name": "Colly",
            "url": "http://localhost:8000/api/experiments/629cedfb-b74e-4433-9f4a-24a321418013/",
            "version_number": 3,
            "versions": [
                {
                    "name": "Colly",
                    "version_number": 1,
                    "is_default_version": false,
                    "version_description": ""
                },
                {
                    "name": "Colly",
                    "version_number": 2,
                    "is_default_version": true,
                    "version_description": "asd"
                }
            ]
        },
        {
            "id": "81b21c93-e6bb-44c8-98d0-61c43585eee3",
            "name": "Media Index",
            "url": "http://localhost:8000/api/experiments/81b21c93-e6bb-44c8-98d0-61c43585eee3/",
            "version_number": 3,
            "versions": [
                {
                    "name": "Media Index",
                    "version_number": 2,
                    "is_default_version": true,
                    "version_description": "Remove all collection types"
                }
            ]
        }
    ]
}
```

## User Impact

1. Scripts that expects the version of an experiment at the toplevel will break
2. Better navigation to find a version of an experiment

### Demo

N/A

### Docs and Changelog

Pending


tasks.py
```diff
@@ -66,7 +66,7 @@ def translations(c: Context):
 
 @task
 def schema(c: Context):
-    c.run("python manage.py spectacular --file api_schema.yaml")
+    c.run("python manage.py spectacular --file api-schema.yml --validate")
```
SmittieC, line 69: I synced this to https://github.com/dimagi/open-chat-studio/blob/d3069eb96d5ade023f5555d72d677d41820f97ca/apps/api/tests/test_schema.py#L14

SmittieC:
Do we think change comms on the users channel is necessary before we merge?

snopoke (approved):

tasks.py
snopoke: Maybe we should update this comment to say use `inv schema`

Issue #1620, opened 2025-05-22
Remove sass from codebase


snopoke:
We currently have a mix of vanilla CSS, SASS and SCSS files. We don't need both SASS and SCSS so let's remove SASS and migrate the files to .css files

PR #1621, opened 2025-05-22
remove sass

snopoke:
## Description
Resolves https://github.com/dimagi/open-chat-studio/issues/1620

This speeds up our webpack built and removes a dependency that isn't used much and isn't adding value.

## User Impact
None


SmittieC (approved):
I scanned though. LGTM, thanks

PR #1628, opened 2025-05-23
Add admin views for collections

SmittieC:
## Description

In the Collection edit view, you'll see the CollectionFile and versions of it as well

### Demo

image

image

### Docs and Changelog

N/A

### DIFF
--- a/apps/documents/admin.py
+++ b/apps/documents/admin.py
@@ -1,0 +1,27 @@
+from django.contrib import admin
+
+from .models import Collection, CollectionFile
+
+
+class CollectionFileInline(admin.TabularInline):
+    model = CollectionFile
+    extra = 0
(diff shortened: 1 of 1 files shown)


snopoke (commented):
You can extend `apps.analysis.admin.ReadonlyAdminMixin` to make them read only - I find this useful because it prevents Django from loading every single related model for the select lists in forms.

SmittieC:
> You can extend `apps.analysis.admin.ReadonlyAdminMixin` to make them read only - I find this useful because it prevents Django from loading every single related model for the select lists in forms.

Good point. https://github.com/dimagi/open-chat-studio/pull/1628/commits/f5f155ca55331f564881e4e0217b34c7c4e44073

snopoke (approved):

PR #1630, opened 2025-05-26
Ba/refactor custom parse ai message

### LINKED ISSUE #1604: Refactor custom_parse_ai_message function
Currently, the custom_parse_ai_message function is acting as a wrapper around parse_ai_message_to_tool_action, and introduces a fallback mechanism that handles exceptions related to missing tool_call_id. In this fallback, we manually construct ToolAgentAction objects and assign a default tool_call_id using uuid.uuid4().

The preferred approach is to copy the original implementation of parse_ai_message_to_tool_action and make a direct update to the line assigning the tool_call_id, rather than catching and handling the exception downstream

barora05:
## Description
Copy original implementation of parse_ai_message_to_tool_action and refactor custom_parse_ai_function so that we don't need to update the tool_call_id in the catch block. Added fix for anthropic built in tool. Closes #1604

### DIFF
--- a/apps/service_providers/tests/test_chat_with_tools.py
+++ b/apps/service_providers/tests/test_chat_with_tools.py
@@ -2,6 +2,7 @@ from typing import Any
 from unittest.mock import Mock, patch
 
 import pytest
+from langchain_core.agents import AgentFinish
 from langchain_core.messages import AIMessage, AIMessageChunk
 from langchain_core.messages.tool import ToolMessage, tool_call_chunk
 from langchain_core.tools import BaseTool, Tool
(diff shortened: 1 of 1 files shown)


SmittieC (commented):
Are there tests (maybe in a previous PR) to cover this?

barora05:
> Are there tests (maybe in a previous PR) to cover this?

There are tests added for custom tools previously in the code, I can add the tests for built in tools.

SmittieC:
> > Are there tests (maybe in a previous PR) to cover this?
> 
> There are tests added for custom tools previously in the code, I can add the tests for built in tools.

As long as this method is properly tested, we should be good

apps/service_providers/llm_service/helper.py
```diff
@@ -1,43 +1,92 @@
...
+    """Parse an AI message potentially containing tool_calls.
+    This function is adapted from the `parse_ai_message_to_tool_action`
+    function in the langchain library.
+    1. Added a specific condition to skip malformed built in tool calls that
+       originate from Anthropic models (where `call.get("type") == "tool_call"`,
+       `call.get("name") == ""`, and `call.get("id") is None`).
+    2. Ensure that tool_call_id is present in every tool_call as tool_call_id is not returned by anthropic
+    built_in_tools
+    3. Separate processing of custom tools and built in tools as custom tools are present in tool_calls
+    whereas built_in_tools are present in additional kwargs
```
snopoke, line 23: These 3 cases should be tested directly. The current test appears to only test point 3

snopoke (approved):

PR #1641, opened 2025-05-27
add dropdown for date range filtering

### LINKED ISSUE #1504: Filter sessions by relative date ranges
Allow user to filter sessions by date ranges instead of absolute dates.

For example:

* last 7 days
* last 30 days

barora05:
## Description
Added dropdown for date range based filtering. Closes #1504 

Created a separate date_range column in filters as to filter sessions by last 1 hour I needed datetime instead of only date. Changing type from date to datetime for timestamp columns would have impact on both first_message and last_message and it would be odd on selecting time for on operator whereas on backend we can only search for the date part for on operator.
### Demo
[Demo](https://drive.google.com/file/d/1Gbas3DV21Ff8ZhQPRjwRql0b9Tzpwcxr/view?usp=drive_link)


templates/experiments/filters.html
```diff
@@ -129,7 +133,8 @@
           'last_message': {type: 'timestamp', operators: fieldTypeFilters.timestamp, label: 'Last Message'},
           'tags': {type: 'choice', operators: fieldTypeFilters.choice, options: tags, label: 'Tags' },
           'versions': {type: 'choice', operators: fieldTypeFilters.choice.filter(op => op === 'any of' || op === 'excludes'), options: versionsList, label: 'Versions'},
-          'channels': {type: 'choice', operators: fieldTypeFilters.choice, options: channelsList, label: 'Channels'}
+          'channels': {type: 'choice', operators: fieldTypeFilters.choice, options: channelsList, label: 'Channels'},
+          'date_range': {type: 'datetime', operators: fieldTypeFilters.datetime, label: 'Date Range'}
```
SmittieC, line 137: I'm wondering if it's strictly necessary to have a `date_range` filter. Could we not use the `last_message` filter for selected date range, since it ultimately applies to the last message. If there are separate filters that translates to the same column, there might exist a scenario where I can do something like

```
last_message before today && date_range after today
```

which doesn't really make sense. Happy to hear your thoughts

apps/experiments/filters.py
```diff
@@ -124,17 +127,48 @@ def build_participant_filter(operator, value):
...
-        elif operator == Operators.BEFORE:
-            return Q(**{f"{field}__date__lt": date_value})
-        elif operator == Operators.AFTER:
-            return Q(**{f"{field}__date__gt": date_value})
-    except (ValueError, TypeError):
+        client_tz = pytz.timezone(timezone) if timezone else pytz.UTC
+        now_client = datetime.now(client_tz)
+        # Handle 'range' operator with relative time (e.g., '1h', '7d')
+        if operator == Operators.RANGE:
+            if value.endswith(("h", "d", "m")):
```
SmittieC, line 142: Nit: Do an early return here if `value` does not end with `None`. In the rest of the code you can then assume it is a valid value

templates/experiments/date_range_filters.html
```diff
@@ -0,0 +1,110 @@
...
+    <i class="fa-regular fa-clock"></i>
+    <span x-text="selectedLabel || 'Date Range'" class="dropdown-label"></span>
+    <i class="fa-solid fa-caret-down ml-1"></i>
+  </div>
+  <ul tabindex="0" class="dropdown-content menu p-2 shadow bg-base-100 rounded-box w-52">
+    <template x-for="option in options" :key="option.value">
+      <li>
+        <a href="#"
+           x-text="option.label"
+           @click.prevent="selectedLabel = option.label;
```
SmittieC, line 12: Nit: Move this logic into a function defined in `dateRangeDropdown`. Multiline logic is just easier to read that way

templates/experiments/single_experiment_home.html
```diff
@@ -248,6 +248,7 @@ <h3 class="font-bold text-lg">Parent Experiments:
                 {% include "experiments/components/exports.html" %}
               {% endif %}
               {% include "experiments/filters.html"%}
+              {% include "experiments/date_range_filters.html" %}
```
SmittieC, line 251: Is there a specific reason why you didn't include the date_range_filter's template and JS stuff in the filters.html file? Just curious. It would be nice if whatever we do here, could easily be ported as a single entity to other areas

apps/experiments/filters.py
```diff
@@ -124,17 +127,48 @@ def build_participant_filter(operator, value):
...
+
+                range_starting_client_time = now_client - delta
+                range_starting_utc_time = range_starting_client_time.astimezone(pytz.UTC)
+                return Q(**{f"{field}__gte": range_starting_utc_time})
+
+            else:
+                return None
+        else:
+            # No need to convert the date as we are passing only date and date shown on the UI is in client time only
+            date_value = datetime.strptime(value, "%Y-%m-%d").replace(tzinfo=UTC)
```
snopoke, line 163: This should still be converted from the user's timezone because the session dates are stored in UTC so a session on "2025-03-01 UTC" may not match "2025-03-01+06:00".

You can also use `datetime.fromisoformat` here instead of custom parsing (and you'll get a datetime value instead of a date).

apps/experiments/filters.py
```diff
@@ -124,17 +127,48 @@ def build_participant_filter(operator, value):
...
+
+            if operator == Operators.ON:
+                return Q(**{f"{field}__date": date_value})
+            elif operator == Operators.BEFORE:
+                return Q(**{f"{field}__date__lt": date_value})
+            elif operator == Operators.AFTER:
+                return Q(**{f"{field}__date__gt": date_value})
```
snopoke, line 170: After converting the user date to a UTC datetime these will need to be adjusted by removing the `__date` transformation since we also want to filter based on the time part.

apps/experiments/filters.py
```diff
@@ -124,17 +127,48 @@ def build_participant_filter(operator, value):
...
+                return Q(**{f"{field}__gte": range_starting_utc_time})
+
+            else:
+                return None
+        else:
+            # No need to convert the date as we are passing only date and date shown on the UI is in client time only
+            date_value = datetime.strptime(value, "%Y-%m-%d").replace(tzinfo=UTC)
+
+            if operator == Operators.ON:
+                return Q(**{f"{field}__date": date_value})
```
snopoke, line 166: After converting the user date to a UTC datetime this will need adjusting: `Q(**{f"{field}__date": local_datetime.date()})`

templates/experiments/date_range_filters.html
```diff
@@ -0,0 +1,110 @@
...
+              break;
+            }
+          }
+        }
+
+        if (matchedIndex !== null) {
+          const operatorKey = `filter_${matchedIndex}_operator`;
+          if (params.get(operatorKey) === 'range') {
+            params.set(`filter_${matchedIndex}_value`, value);
+            localStorage.setItem('selected_date_range', value);
```
snopoke, line 84: this duplicates much of the logic above for determining `hasValidFilter`

templates/experiments/filters.html
```diff
@@ -95,7 +95,14 @@
...
               <template x-if="filterData.columns[filter.column]?.type === 'timestamp'">
-                <input type="date" x-model="filter.value" @input="triggerFilterChange()" class="w-40 block px-3 border border-base-200 rounded-lg bg-base-100 text-sm" style="height: 2rem;">
+                <input
+                  :type="filter.operator === 'range' ? 'text' : 'date'"
+                  x-model="filter.value"
+                  @input="triggerFilterChange()"
+                  :placeholder="filter.operator === 'range' ? 'e.g. 1d for 1 day, 1h for 1 hour' : ''"
+                  class="w-60 block px-3 border border-base-200 rounded-lg bg-base-100 text-sm"
+                  style="height: 2rem;"
+                >
```
snopoke, line 105: For the range operator, shouldn't this be a select? If we're going to let users type arbitrary things in here we might need some client side validation.

barora05: yes. I was going to confirm this with you. I will use the same option list here as well.

templates/experiments/date_range_filters.html
barora05: done

apps/experiments/filters.py
barora05: I have used datetime.fromisoformat. The time passed in the date is 00:00:00. If I convert it, the date changes and correct results are not obtained. I have tried the on, after and before operator. works fine for me without converting. datetime values in the db are saved like 2025-05-28 19:04:47.637948+05:30

templates/experiments/date_range_filters.html
barora05: done

templates/experiments/filters.html
barora05: done

barora05:
There is one bug, the selected value in the dropdown and in the range filter is not properly synced. Working on it

barora05:
@snopoke my db timezone is IST. Let me try this after updating the timezone to UTC in db.

apps/experiments/filters.py
barora05: I have updated my db timezone to UTC. on, after and before operator works fine without converting the datetime to utc. 
Issue with converting to UTC
ON Operator - We pass the date as 2025-05-28 00:00:00, after converting it to UTC it becomes 2025-05-27 18:30:00 and we take only the date part, so it shows sessions with last_message as 27th not 28th.
AFTER operator - We would be showing all the datetime value greater than  2025-05-27 18:30:00 so it would include 28th May value as well, whereas currently we only pick 29th May date.

snopoke: > ON Operator - We pass the date as 2025-05-28 00:00:00, after converting it to UTC it becomes 2025-05-27 18:30:00 and we take only the date part, so it shows sessions with last_message as 27th not 28th.

"we take only the date part" we shouldn't do this - we should filter on the full timestamp because of this issue.

I had Claude generate some examples to illustrate the issue: https://claude.ai/share/69824039-aab3-4f4f-997f-06ec9300e592

barora05:
@snopoke tried this https://github.com/dimagi/open-chat-studio/pull/1641#discussion_r2120377995. We can do it for before. If we do it for after it will alter our current behaviour, we don't show the date selected sessions in after operator currently, but we would start showing them as time is 00:00:00.

barora05:
> @barora05 please do a changelog entry for this

https://github.com/dimagi/open-chat-studio-docs/pull/110

PR #1643, opened 2025-05-28
tag output messasge with default keyword

### LINKED ISSUE #1634: Tag messages from router node when default route is chosen
In router nodes, if the LLM isn't able to select a route and the default route is selected as a fallback, the output message should be tagged with `<route name>:default` with tag category of `ERROR`.

barora05:
## Description
Tag output message with default tag and error category if default route is selected. Closes #1634

### DIFF
--- a/apps/pipelines/nodes/nodes.py
+++ b/apps/pipelines/nodes/nodes.py
@@ -551,9 +551,13 @@ class RouterMixin(BaseModel):
         """
         return {f"output_{output_num}": keyword for output_num, keyword in enumerate(self.keywords)}
 
-    def get_output_tags(self, selected_route) -> list[tuple[str, str]]:
+    def get_output_tags(self, selected_route, is_default_keyword: bool) -> list[tuple[str, str]]:
         if self.tag_output_message:
-            return [(f"{self.name}:{selected_route}", TagCategories.BOT_RESPONSE)]
+            tag_name = f"{self.name}:{selected_route}"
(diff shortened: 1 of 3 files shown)


apps/pipelines/nodes/nodes.py
```diff
@@ -618,19 +622,22 @@ def _process_conditional(self, state: PipelineState):
...
             keyword = default_keyword
+            is_default_keyword = True
         if not keyword:
             keyword = default_keyword
+            is_default_keyword = True
 
         if session:
             self._save_history(session, self.node_id, node_input, keyword)
-        return keyword
+        return keyword, is_default_keyword
```
snopoke, line 640: This also needs to be applies to `BooleanNode` and `StaticRouter`

barora05: yes. done that

apps/pipelines/tests/test_runnable_builder.py
```diff
@@ -1347,5 +1347,5 @@ def test_router_node_openai_refusal_uses_default_keyword(get_llm_service, provid
         experiment_session=experiment_session,
     )
 
-    keyword = node._process_conditional(state)
+    keyword, is_default_keyword = node._process_conditional(state)
```
snopoke, line 1350:
```suggestion
    keyword, is_default_keyword = node._process_conditional(state)
    assert is_default_keyword
```

snopoke (approved):
one small comment but otherwise GTG

snopoke (approved):

PR #1661, opened 2025-05-29
RAG - Add Embedding model providers

### LINKED ISSUE #1647: RAG - Add Embedding Provider and Services
Users need to be able to configure Embedding providers in the team settings in the same way they can configure LLM providers.

A provider can have models as well, so we'll need create two DB models:
- EmbeddingProvider
- EmbeddingProviderModel

We'll also need a service layer, much like the subclasses of `LlmService`. For now, we can support two services:
- OpenAI
- VoyageAI

These services will be responsible for creating embeddings. We can use langchain to do the heavy lifting for us here:
- [OpenAI](https://python.langchain.com/v0.1/docs/integrations/text_embedding/openai/)
- [VoyageAI](https://python.langchain.com/v0.1/docs/integrations/text_embedding/voyageai/)

SmittieC:
## Description

Resolves #1647 

## User Impact

User will be able to add embedding models, but cannot do anything with it yet

### Demo

demo

### Docs and Changelog

Don't think this is necessary yet?


apps/service_providers/embedding_service.py
```diff
@@ -0,0 +1,6 @@
+class OpenAIEmbeddingService:
+    openai_api_key: str
+    openai_api_base: str = None
+    openai_organization: str = None
+
+    # TODO: Method to create embeddings
```
SmittieC, line 6: This will be added in a following PR

PR #1662, opened 2025-05-29
RAG - Add Embedding model providers

### LINKED ISSUE #1647: RAG - Add Embedding Provider and Services

SmittieC:
The alternative approach: https://github.com/dimagi/open-chat-studio/pull/1679

## Description

Resolves #1647 

I think for now I'm going to leave the VoyageAI provider out to keep the changes smaller. We can add this in a following PR
## User Impact

User will be able to add embedding models, but cannot do anything with it yet

### Demo

demo

### Docs and Changelog

Don't think this is necessary yet?


snopoke (commented):
I haven't looked at the code but just thinking out loud after watching the demo. This will require users to duplicate their keys if they are using the same provider for LLM and embedding, I wonder if we could allow users to 'copy' keys from an existing provider? We could also handle updating keys so that they stay in sync.

SmittieC:
> I haven't looked at the code but just thinking out loud after watching the demo. This will require users to duplicate their keys if they are using the same provider for LLM and embedding, I wonder if we could allow users to 'copy' keys from an existing provider? We could also handle updating keys so that they stay in sync.

[Here's](https://github.com/dimagi/open-chat-studio/issues/1670) a followup ticket to do this.

apps/service_providers/embedding_service/default_models.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def _update_embedding_provider_models(EmbeddingProviderModel):
+    """
+    This method updates the EmbeddingProviderModel objects in the database to match the
+    DEFAULT_EMBEDDING_PROVIDER_MODELS.
+    """
+    for provider_type, provider_models in DEFAULT_EMBEDDING_PROVIDER_MODELS.items():
+        for model in provider_models:
+            EmbeddingProviderModel.objects.get_or_create(team=None, name=model, type=provider_type)
```
SmittieC, line 24: Should we allow users to make custom embedding models?

SmittieC:
Closing this in favor of https://github.com/dimagi/open-chat-studio/pull/1679

PR #1667, opened 2025-05-29
Add Django management command to find feature flag usages

### LINKED ISSUE #1637: Write a management command to remove old feature flags from the database.
This could use regex's to try and find FF that are referenced in code or fetch records from the DB to search in code.

Either way, it should check for usages in code before removing and optionally prompt the user to confirm deletion showing any existing usages in code first.

We are using Django Waffle for feature flags with a custom model: `apps.teams.models.Flag`

snopoke:
Resolves #1637

## Description
Implements management command to find feature flag usages for the purpose of determining if a flag can be deleted.

This PR also enforces a `flag_` prefix for new feature flags and adds a documentation page on flag usage.

### DIFF
--- a/templates/chatbots/single_chatbot_home.html
+++ b/templates/chatbots/single_chatbot_home.html
@@ -252,7 +252,7 @@
         </div>
       </div>
     </div>
-    {% flag "events" %}
+    {% flag "flag_events" %}
       <div role="tabpanel" class="tab-content" id="content-events">
         <div class="app-card">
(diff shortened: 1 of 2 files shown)


SmittieC (commented):
Just some tests that fails due to the new flag naming convension, but otherwise looks good to me. This is neat

snopoke:
@SmittieC I decided to go all in on this and renamed all the current flags in code and the DB

SmittieC (changes requested):
I think we need a migration to rename the DB flags as well, otherwise the teams would not be on the new flag

PR #1672, opened 2025-05-29
fix tool parsing

snopoke:
Fix [OCSBUGS-14](https://dimagi.atlassian.net/browse/OCSBUGS-14)
## Description
The custom tool parsing was producing duplicate tool actions which causes this error [OPEN-CHAT-STUDIO-XY](https://dimagi.sentry.io/issues/6586713766/events/ea5d948884ec41798a690f6661c20979/)

The specific case can be seen in the response example in `apps/service_providers/tests/data/openai_completions_api_tool_calls.json`. Specifically, the tool calls are present in both `additional_kwargs.tool_calls` and in `tool_calls`. This resulted in duplicate tool calls and duplicate tool call responses to the LLM.

### DIFF
--- a/apps/service_providers/llm_service/helper.py
+++ b/apps/service_providers/llm_service/helper.py
@@ -16,10 +16,6 @@ def custom_parse_ai_message(message) -> list[AgentAction] | AgentFinish:
     1. Added a specific condition to skip malformed built in tool calls that
        originate from Anthropic models (where `call.get("type") == "tool_call"`,
        `call.get("name") == ""`, and `call.get("id") is None`).
-    2. Ensure that tool_call_id is present in every tool_call as tool_call_id is not returned by anthropic
-    built_in_tools
-    3. Separate processing of custom tools and built in tools as custom tools are present in tool_calls
-    whereas built_in_tools are present in additional kwargs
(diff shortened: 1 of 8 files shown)


barora05 (approved):

apps/service_providers/llm_service/helper.py
```diff
@@ -57,6 +53,10 @@ def custom_parse_ai_message(message) -> list[AgentAction] | AgentFinish:
         else:
             tool_input = _tool_input
 
+        # Skip malformed built-in tool calls that originate from Anthropic models
```
SmittieC, line 59: Would this only happen sometimes?

SmittieC (approved):
Nice. Thanks for adding the example tool call responses. It helps

apps/service_providers/llm_service/helper.py
snopoke: It happens every time Anthropic does tool calls for internal tools like 'web search'. I haven't determined if it's a Langchain artefact or not because it's not included in their example response in the docs: https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/web-search-tool#response

PR #1677, opened 2025-05-29
add validation for default keyowrd size

barora05:
## Description
Validate default_keyword_index. Fixes [this exception](https://dimagi.sentry.io/issues/6642865279/?query=is%3Aunresolved&referrer=issue-stream&stream_index=1)


SmittieC (approved):

snopoke (commented):
I think we should hide the default keyword field entirely: https://github.com/dimagi/open-chat-studio/issues/1681

PR #1701, opened 2025-06-03
Ba/telegram multimedia

### LINKED ISSUE #1226: [Multimedia support] Send image attachments | Telegram

Depends on [this ticket](https://github.com/dimagi/open-chat-studio/issues/1225)

### Requirements
Attachments should be sent directly to the chat. If it is not supported for any reason, a link should be appended to the message instead.
See the [documentation](https://pytba.readthedocs.io/en/latest/sync_version/index.html#telebot.TeleBot.send_photo).

barora05:
## Description
Add multimedia support for telegram. Closes #1226

### Demo
WhatsApp Image 2025-06-03 at 3 10 41 PM
WhatsApp Image 2025-06-03 at 3 10 57 PM
12 mb is above the max image size limit so download link is returned.

### DIFF
--- a/apps/channels/tests/message_examples/test_telegram_channel.py
+++ b/apps/channels/tests/message_examples/test_telegram_channel.py
@@ -25,18 +25,14 @@ def telegram_channel():
     return TelegramChannel(experiment, experiment_channel=experiment_channel, experiment_session=session)
 
 
-@pytest.fixture()
-def make_mock_file():
-    def _make_mock_file(name, content_type, size, file_data="file_data"):
-        file = Mock(spec=File)
-        file.name = name
-        file.content_type = content_type
-        file.content_size = size
-        file.file = file_data
-        file.download_link.return_value = f"http://example.com/{name}"
-        return file
-
-    return _make_mock_file
+def make_mock_file(name, content_type, size, file_data="file_data"):
+    file = Mock(spec=File)
+    file.name = name
+    file.content_type = content_type
+    file.content_size = size
+    file.file = file_data
+    file.download_link.return_value = f"http://example.com/{name}"
+    return file
 
 
 @pytest.mark.django_db()
@@ -113,7 +109,7 @@ def test_handle_telegram_api_error_other_errors():
     assert excinfo.value.__cause__ == api_error
 
 
-def test_supported_image_file(telegram_channel, make_mock_file):
+def test_supported_image_file(telegram_channel):
     image_file = make_mock_file("img.jpg", "image/jpeg", 5 * 1024 * 1024)
 
     with (
@@ -126,7 +122,7 @@ def test_supported_image_file(telegram_channel, make_mock_file):
         mock_send_text.assert_called_once_with("Here is your file")
 
 
-def test_large_image_fallback_to_text(telegram_channel, make_mock_file):
+def test_large_image_fallback_to_text(telegram_channel):
     channel = telegram_channel
     large_image = make_mock_file("big_img.jpg", "image/jpeg", 15 * 1024 * 1024)
 
@@ -142,7 +138,7 @@ def test_large_image_fallback_to_text(telegram_channel, make_mock_file):
         mock_send_text.assert_called_once_with(expected_message)
 
 
-def test_unsupported_mime_file(telegram_channel, make_mock_file):
+def test_unsupported_mime_file(telegram_channel):
     channel = telegram_channel
     exe_file = make_mock_file("script.exe", "application/octet-stream", 1 * 1024 * 1024)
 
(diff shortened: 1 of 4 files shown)


apps/chat/channels.py
```diff
@@ -895,6 +895,67 @@ def echo_transcript(self, transcript: str):
...
+        unsupported_files = []
+
+        for file in files:
+            mime = file.content_type
+            size = file.content_size or 0  # in bytes
+
+            is_inline_supported = False
+            if (
+                mime.startswith("image/")
+                and size <= 10 * 1024 * 1024  # 10mb
```
SmittieC, line 913: I see the docs doesn't mention if this should be base 10 or base 2. Event Telegram's own docs doesn't mention that. I suspect we'll have to use base 10, looking at some of the stack overflow posts. Might be good to test this maunally and see when their server prevents an upload when you use 50 MB vs 50 MiB?

If it's base 10, then you can simple get the file size using `file.size_mb`

apps/chat/channels.py
```diff
@@ -895,6 +895,67 @@ def echo_transcript(self, transcript: str):
...
+        file_data = file.file
+
+        # Mapping of MIME type prefixes to Telegram methods and argument names
+        mime_method_map = {
+            "image/": ("send_photo", "photo"),
+            "video/": ("send_video", "video"),
+            "audio/": ("send_audio", "audio"),
+        }
+
+        for prefix, (method_name, arg_name) in mime_method_map.items():
```
SmittieC, line 938: To prevent having to always loop through the `mime_method_map` and maybe use one of the methods in there, I'd suggest taking another approach. Maybe something like this:

```python
method = None
arg_name = ""
match mime.split("/")[0]:
    case "image":
        method = self.telegram_bot.send_photo
        arg_name = "photo"
    case "video":
        method = self.telegram_bot.send_video
        arg_name = "video"
    case "audio":
        method = self.telegram_bot.send_audio
        arg_name = "audio"
    case _:
        method = self.telegram_bot.send_document
        arg_name = "document"

antiflood(method, chat_id, **{arg_name: file_data})
```

apps/chat/channels.py
```diff
@@ -895,6 +895,67 @@ def echo_transcript(self, transcript: str):
...
+            if mime.startswith(prefix):
+                method = getattr(self.telegram_bot, method_name)
+                antiflood(method, chat_id, **{arg_name: file_data})
+                return
+
+        antiflood(self.telegram_bot.send_document, chat_id, document=file_data)
+
+    def _send_files_to_user(self, files: list[File]):
```
SmittieC, line 946: You don't need to override this method. The base class' logic should suffice

apps/channels/tests/test_base_channel_behavior.py
```diff
@@ -945,3 +946,67 @@ def test_chat_message_returned_for_cancelled_generate():
     response = channel.new_user_message(channel.message)
 
     assert type(response) is ChatMessage
+
+
+@pytest.fixture()
```
SmittieC, line 951: These tests should reside in the `test_telegram_channel.py` file rather

apps/chat/channels.py
barora05: Tried it out. It uses base 2
elebot.apihelper.ApiTelegramException: A request to the Telegram API was unsuccessful. Error code: 400. Description: Bad Request: file of size 12720000 bytes is too big for a photo; the maximum size is 10485760 bytes

apps/channels/tests/message_examples/test_telegram_channel.py
```diff
@@ -86,3 +87,67 @@ def test_handle_telegram_api_error_other_errors():
...
+    experiment = Mock()
+    session = Mock()
+    session.id = 123
+    experiment_channel = Mock()
+    experiment_channel.extra_data = {"bot_token": "fake_token"}
+    channel = TelegramChannel(experiment, experiment_channel=experiment_channel, experiment_session=session)
+    return channel, session
+
+
+def make_mock_file(name, content_type, size, file_data="file_data"):
```
SmittieC, line 103: You can use a fixture here

apps/channels/tests/message_examples/test_telegram_channel.py
```diff
@@ -86,3 +111,45 @@ def test_handle_telegram_api_error_other_errors():
 
     assert expected_channel_exception_message == str(excinfo.value)
     assert excinfo.value.__cause__ == api_error
+
+
+def test_supported_image_file(telegram_channel, make_mock_file):
+    image_file = make_mock_file("img.jpg", "image/jpeg", 5 * 1024 * 1024)
```
SmittieC, line 117: My apologies, I didn't see that the `make_mock` is parameterized. In this case, it makes more sense to do what you initially did where you called a function directly. Using a fixture doesn't quite make sense here. My apologies

SmittieC (commented):
One small comment, otherwise looks good!

SmittieC (approved):

PR #1709, opened 2025-06-04
RAG - Embed files locally

SmittieC:
## Description

Users can now choose an embedding model and upload files to OCS after which we'll chunk + index those files. Since we don't yet have the infra in place to query the vectors, the "local" indexes are sort of useless. In the pipeline views where we query all team collections, I added a filter to return only those that are remote, thus usable. Maybe we should prevent users from choosing creating local indexes instead?

A big change that was needed is the concept of a local and remote index. The remote index is one where the service provider is responsible for hosting and managing chunking and embeddings. The local one refers to OCS, who is responsible for handing embeddings / chunking. OpenAI is the only remote index we'll probably support, but it should be easy enough to add more creating new subclasses. Dev docs can be found [here](https://github.com/dimagi/open-chat-studio/pull/1709/commits/14204e434b5fa029c31cbe6aca062464c088ef7f).

Review tip: I would recommend looking at this commit-wise. There are some morphing happening as I went along, but it's generally small changes to existing things.

## User Impact

- Users can choose to use OCS as the index host, in which case OCS will manage chunking / embedding and do RAG itself.
- Since the searh part of the RAG implementation is missing, adding an OCS based collection is not usable yet

### Docs and Changelog

Pending

### DIFF
--- a/apps/pipelines/views.py
+++ b/apps/pipelines/views.py
@@ -133,8 +133,12 @@ def _pipeline_node_parameter_values(team, llm_providers, llm_provider_models):
     collections = (
         Collection.objects.working_versions_queryset().filter(team=team, is_index=False).values("id", "name").all()
     )
+    # Until OCS fully supports RAG, we can only use remote indexes
     collection_indexes = (
-        Collection.objects.working_versions_queryset().filter(team=team, is_index=True).values("id", "name").all()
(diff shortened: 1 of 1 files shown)


SmittieC:
@coderabbitai please review thoroughly

Issue #1656, opened 2025-05-28
Migrate shared pipelines to one-to-one experiment relation


stephherbers:
Any pipeline that are shared between multiple experiments must be copied and have the reference updated for the next transition phase 

Check all experiment in apps.experiment.models to see if any share pipeline field. For any that do share, create a duplicate of the pipeline model apps.pipeline.model and have the experiment with the duplicate reference that new pipeline. The end result should be a one-to-one relation between the experiments and pipelines by copying pipelines

stephherbers:
@claude implement this ticket

PR #1711, opened 2025-06-04
Duplicate Pipeline Management Command

stephherbers:
## Description

Resolves [#1656](https://github.com/dimagi/open-chat-studio/issues/1656)

Tested locally with:
- basic pipelines 
- complex pipelines (10+ nodes of all types)
- archived experiments and pipelines (mix too)
- versioned
- copies and copies of versions

lmk of any other edge cases I'm missing and would like me to test

## User Impact

None for functionality, but more pipelines will appear in their pipeline homepage. The naming convention will change it will be renames to "Copy for (`experiment.name`)"

### Demo

sample output with 2 duplicate pipelines:
`Found 2 shared pipelines to duplicate`
`Do you want to proceed? (y/N): y`
`Processing pipeline: broken pipe (ID: 9)`
  `Shared by 2 experiments`
`  Created pipeline copy 34 for experiment has duplicate pipline`
`Processing pipeline: Language Router (ID: 3)`
`  Shared by 2 experiments`
  `Created pipeline copy 35 for experiment alphabot`
`Successfully created 2 pipeline copies`

### Docs and Changelog

No


apps/experiments/management/commands/duplicate_shared_pipelines.py
```diff
@@ -0,0 +1,91 @@
...
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Show what would be done without making changes",
+        )
+
+    def handle(self, *args, **options):
+        dry_run = options["dry_run"]
+
+        shared_pipelines = Pipeline.objects.annotate(experiment_count=Count("experiment")).filter(
```
snopoke, line 23: I think this should also filter by `working_version=None` since we only want to duplicate working version pipelines right?

apps/experiments/management/commands/duplicate_shared_pipelines.py
```diff
@@ -0,0 +1,91 @@
...
+                continue
+
+            for _, experiment in enumerate(experiments[1:], 1):
+                try:
+                    with transaction.atomic():
+                        new_pipeline = Pipeline.objects.create(
+                            team=shared_pipeline.team,
+                            name=f"{shared_pipeline.name} (Copy for {experiment.name})",
+                            data=shared_pipeline.data,
+                            working_version=shared_pipeline.working_version,
```
snopoke, line 61: working version should be None

apps/experiments/management/commands/duplicate_shared_pipelines.py
```diff
@@ -0,0 +1,91 @@
...
+                            version_number=shared_pipeline.version_number,
```
snopoke, line 62: should this get reset to 1?

apps/experiments/management/commands/duplicate_shared_pipelines.py
SmittieC: I think so yeah. [Pipelines are archived with experiments](https://github.com/dimagi/open-chat-studio/blob/c4651c54c11137fe4ccd14634a175e3a231e4193/apps/experiments/models.py#L911-L928), but in the case of versions, there will always be a 1-1 relationship between the pipeline and experiment, since we create a version of the pipeline when versioning. Only working experiments have 1-N rel to pipelines currently.

stephherbers: right thank you both

SmittieC, line 23: I'm realizing now that the default queryset will not return archived pipelines. You'll have to use `Pipeline.objects.get_all().filter(...)` to include them. I guess for the sake of data consistancy we'll want to target them as well

PR #1719, opened 2025-06-05
move source material, surveys, concent forms, and tags out of experim…

### LINKED ISSUE #1718: Chatbots Nav: Part 2
Move move source material, surveys, consent forms, and tags out from under Experiments when the chatbot FF is enabled

stephherbers:
## Description

resolves #1718

move source material, surveys, consent forms, and tags out from under Experiments when the chatbot FF is enabled

## User Impact

UI only

### Demo

FF enabled:
Screenshot 2025-06-05 at 2 18 07 PM
FF not enabled:
Screenshot 2025-06-05 at 2 23 44 PM

### Docs and Changelog

n/a

### DIFF
--- a/templates/web/components/team_nav.html
+++ b/templates/web/components/team_nav.html
@@ -9,6 +9,36 @@
         {% translate "Chatbots" %}
       </a>
     </li>
+    <li>
+      <a href="{% url 'experiments:source_material_home' request.team.slug %}"
+         {% if active_tab == 'source_material' %}class="menu-active"{% endif %}>
+        <i class="fa fa-book fa-fw"></i>
+        {% translate "Source Material" %}
+      </a>
+    </li>
+    <li>
+      <a href="{% url 'experiments:survey_home' request.team.slug %}"
+         {% if active_tab == 'survey' %}class="menu-active"{% endif %}>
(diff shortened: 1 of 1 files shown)


templates/web/components/team_nav.html
```diff
@@ -9,50 +9,90 @@
...
-      </li>
-      <li>
-        <a href="{% url 'experiments:consent_home' request.team.slug %}"
-           {% if active_tab == 'consent_forms' %}class="menu-active"{% endif %}>
-          <i class="fa-solid fa-file-contract fa-fw"></i>
-          {% translate "Consent Forms" %}
-        </a>
-      </li>
-      {% if perms.annotations.add_tag %}
+      {% flag "flag_chatbots" or %}
```
SmittieC, line 49:
```suggestion
      {% flag "flag_chatbots" %}
```

SmittieC: The desired behaviour is to show safety layers under experiments all the time, but the rest (source material etc) should show under experiments only when the chatbots flag is disabled, otherwise with it enabled, those should show under the chatbots heading?

In which case, it seems to me to work as expected without that "or " clause. Am I missing something?

SmittieC (approved):

PR #1720, opened 2025-06-05
add file support for turn.io

### LINKED ISSUE #1228: [Multimedia support] Support attachment links | API

### Requirements
An attachment key in the response should contain all attachment links to the client can download them from there.

barora05:
## Description
Add multimedia support for Turn.io. Closes #1228

### DIFF
--- a/apps/chat/channels.py
+++ b/apps/chat/channels.py
@@ -975,7 +975,7 @@ class WhatsappChannel(ChannelBase):
             from_=from_number,
             to=to_number,
             platform=ChannelPlatform.WHATSAPP,
-            file_name=file.name,
+            file=file,
             download_link=file.download_link(experiment_session_id=self.experiment_session.id),
         )
 
--- a/apps/service_providers/messaging_service.py
+++ b/apps/service_providers/messaging_service.py
@@ -1,4 +1,5 @@
 import logging
+import mimetypes
 import uuid
 from datetime import datetime, timedelta
 from functools import cached_property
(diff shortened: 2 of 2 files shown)


apps/channels/tests/test_whatsapp_integration.py
```diff
@@ -124,17 +124,24 @@ def test_attachments_are_sent_as_separate_messages(self, twilio_client_mock, exp
...
+        ],
     )
     def test_parse_text_message(self, message, message_type):
         message = TurnWhatsappMessage.parse(message)
         assert message.participant_id == "27456897512"
         if message_type == "text":
             assert message.message_text == "Hi there!"
             assert message.content_type == MESSAGE_TYPES.TEXT
-        else:
+        elif message_type == "voice":
```
SmittieC, line 139: since voice is also `media`, I think you can remove this part and let the else claude catch it

apps/chat/channels.py
```diff
@@ -79,6 +79,7 @@ def strip_urls_and_emojis(text: str) -> tuple[str, list[str]]:
 class MESSAGE_TYPES(Enum):
     TEXT = "text"
     VOICE = "voice"
+    MEDIA = "media"
```
SmittieC, line 82: Why was this added?

barora05: added for test.             assert message.content_type == MESSAGE_TYPES.MEDIA

apps/channels/tests/test_whatsapp_integration.py
barora05: I would need to update the type to media in the mocked voice message. Should I do that?
```
"type": "voice",
                "voice": {
                    "id": "180e1c3f-ae50-481b-a9f0-7c698233965f",
                    "mime_type": "audio/ogg; codecs=opus",
                    "sha256": "407d8ac9d98ddddddd78c7bae4179ea131b55740214ccc42373c85d63aeb55b7",
                    "status": "downloaded",
                },
```

apps/chat/channels.py
SmittieC: Oh I see. Hmm, I don't think that test is quite right.
1. It it a misnomer (that's on me).
3. That is testing incoming message parsing. We don't support media messages from users.

apps/service_providers/messaging_service.py
```diff
@@ -183,17 +185,57 @@ def send_voice_message(
...
+        elif mime.startswith(("video/", "audio/")):
+            return size <= 16 * 1024 * 1024  # 16 MB
+        elif mime.startswith("application/"):
+            return size <= 100 * 1024 * 1024  # 100 MB
+        else:
+            return False
+
+    def send_file_to_user(self, from_: str, to: str, platform: ChannelPlatform, file: File, download_link: str):
+        file_name = file.name
+        mime_type, _ = mimetypes.guess_type(file_name)
```
SmittieC, line 218: Since this is a `File` instance, the `content_type` attribute should already be set which does the same as you're doing here, so you can just use that

snopoke (approved):

PR #1721, opened 2025-06-05
Experiment Migration Management Command

### LINKED ISSUE #1664: Allow users to migrate non-pipeline experiments to pipeline chatbots
Create a tool and a management command that will allow migrating non-pipeline experiments to pipeline experiments.

Display a banner on non-pipeline experiments with information about the deprecation and a button that allows them to migrate their experiment to a 'Chatbot'.

stephherbers:
## Description

resolves #1664
second part here: https://github.com/dimagi/open-chat-studio/pull/1724

## User Impact

when run, all assistant and llm experiments will be converted to pipeline experiments

### Demo

n/a

### Docs and Changelog

n/a until the tool for users to do on their own is implemented

### DIFF
--- a/apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
+++ b/apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
@@ -108,11 +108,9 @@ class Command(BaseCommand):
             return "Unknown"
 
     def _convert_experiment(self, experiment):
-        experiment_type = self._get_experiment_type(experiment)
-
-        if experiment_type == "Assistant":
+        if experiment.assistant:
             pipeline = self._create_assistant_pipeline(experiment)
-        elif experiment_type == "LLM":
+        elif experiment.llm_provider:
             pipeline = self._create_llm_pipeline(experiment)
         else:
             raise ValueError(f"Unknown experiment type for experiment {experiment.id}: {experiment_type}")
--- a/apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
+++ b/apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
@@ -1,0 +1,266 @@
+from uuid import uuid4
+
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from django.db.models import Q
+
+from apps.experiments.models import Experiment
+from apps.pipelines.flow import FlowNode, FlowNodeData
+from apps.pipelines.models import Pipeline
+from apps.pipelines.nodes.nodes import AssistantNode, EndNode, LLMResponseWithPrompt, StartNode
+from apps.teams.models import Flag
+
+
+class Command(BaseCommand):
+    help = "Convert assistant and LLM experiments to pipeline experiments"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Show what would be converted without making changes",
+        )
+        parser.add_argument(
+            "--team-slug",
+            type=str,
+            help="Only convert experiments for a specific team (by slug)",
+        )
+        parser.add_argument(
+            "--experiment-id",
+            type=int,
+            help="Convert only a specific experiment by ID",
+        )
+        parser.add_argument(
+            "--chatbots-flag-only",
+            action="store_true",
+            help='Only convert experiments for teams that have the "flag_chatbots" feature flag enabled',
+        )
+
+    def handle(self, *args, **options):
+        dry_run = options["dry_run"]
+        team_slug = options.get("team_slug")
+        experiment_id = options.get("experiment_id")
+        chatbots_flag_only = options["chatbots_flag_only"]
+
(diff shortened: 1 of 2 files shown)


apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
```diff
@@ -0,0 +1,231 @@
...
+        return pipeline
+
+    def _create_llm_pipeline(self, experiment):
+        """Create a start -> LLMResponseWithPrompt -> end nodes pipeline for an LLM experiment."""
+        llm_params = {
+            "name": "llm",
+            "llm_provider_id": experiment.llm_provider.id,
+            "llm_provider_model_id": experiment.llm_provider_model.id,
+            "llm_temperature": experiment.temperature,
+            "history_type": "none",
```
SmittieC, line 200: Since the history type was inherently "global" in the previous "regime", I'd advocade to make the assistant and LLM node's history types "global" by default.

apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
```diff
@@ -0,0 +1,231 @@
...
+                type=node_type,
+                label=node_label,
+                params=node_params,
+            ),
+        )
+
+        end_node = FlowNode(
+            id=end_id,
+            type="endNode",
+            position={"x": 700, "y": 200},
```
SmittieC, line 165: This is what it looks like when I create a pipeline with the command. Maybe the x-axis can do with 50 or a 100 more?

Screenshot from 2025-06-06 08-31-54

apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
```diff
@@ -0,0 +1,214 @@
...
+                    self.style.WARNING(
+                        f"Experiment {experiment_id} is not a published or unreleased version so does not\
+                        require migration."
+                    )
+                )
+                return
+
+            experiments_to_convert = Experiment.objects.filter(id=experiment.id).select_related(
+                "team", "assistant", "llm_provider", "llm_provider_model"
+            )
```
snopoke, line 78: You could just do the `select_related` in the query on line 60 and then you wouldn't have to re-fetch this here.

You could also avoid the extra count later by assigning `experiment_count = 1`

apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
```diff
@@ -0,0 +1,214 @@
...
+            default_working_version_ids = default_experiments.exclude(working_version__isnull=True).values_list(
+                "working_version_id", flat=True
+            )
+            working_experiments = Experiment.objects.filter(query & Q(working_version__isnull=True)).exclude(
+                id__in=default_working_version_ids
+            )
+            combined_ids = list(default_experiments.union(working_experiments).values_list("id", flat=True))
+            experiments_to_convert = Experiment.objects.filter(id__in=combined_ids).select_related(
+                "team", "assistant", "llm_provider", "llm_provider_model"
+            )
```
snopoke, line 90: It may be that loading all of these into memory at once could be an issue so it would be prudent to use an iterator:
```suggestion
            experiments_to_convert_queryset = Experiment.objects.filter(id__in=combined_ids).select_related(
                "team", "assistant", "llm_provider", "llm_provider_model"
            )
            # do this here since you can't do a count on an iterator
            experiment_count = experiments_to_convert_queryset.count()
            experiments_to_convert = experiments_to_convert_queryset.iterator(20)
```

apps/pipelines/models.py
```diff
@@ -115,36 +115,70 @@ def version_display(self) -> str:
...
+            id=str(uuid4()),
             type="startNode",
-            position={
-                "x": -200,
-                "y": 200,
-            },
-            data=FlowNodeData(id=start_id, type=StartNode.__name__, params={"name": "start"}),
+            position={"x": 100, "y": 200},
+            data=FlowNodeData(
+                id=str(uuid4()),
```
snopoke, line 129: shouldn't this be the same value as `FlowNode.id`?

snopoke (commented):
Looking good.

apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
```diff
@@ -0,0 +1,217 @@
...
+    def _create_pipeline_with_node(self, experiment, node_type, node_label, node_params):
+        """Create a pipeline with start -> custom_node -> end structure."""
+        pipeline_name = f"{experiment.name} Pipeline"
+
+        node = FlowNode(
+            id=str(uuid4()),
+            type="pipelineNode",
+            position={"x": 400, "y": 200},
+            data=FlowNodeData(
+                id=str(uuid4()),
```
snopoke, line 168:
```suggestion
        node_id = str(uuid4())
        node = FlowNode(
            id=node_id,
            type="pipelineNode",
            position={"x": 400, "y": 200},
            data=FlowNodeData(
                id=node_id,
```

PR #1724, opened 2025-06-06
Smh/migration tool

stephherbers:
## Description

part of #1664
After pressing banner button to migrate experiment. it navigates to the newly created chatbot
Screenshot 2025-06-06 at 3 46 20 PM

## User Impact

users are able to migrate LLM and Assistant type experiments themselves

### Docs and Changelog

yes to changelog-- the docs have information about the transition, but I may add something specifically for this migration

### DIFF
--- a/apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
+++ b/apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
@@ -1,0 +1,266 @@
+from uuid import uuid4
+
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from django.db.models import Q
+
+from apps.experiments.models import Experiment
+from apps.pipelines.flow import FlowNode, FlowNodeData
+from apps.pipelines.models import Pipeline
+from apps.pipelines.nodes.nodes import AssistantNode, EndNode, LLMResponseWithPrompt, StartNode
+from apps.teams.models import Flag
+
+
+class Command(BaseCommand):
+    help = "Convert assistant and LLM experiments to pipeline experiments"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Show what would be converted without making changes",
+        )
+        parser.add_argument(
+            "--team-slug",
+            type=str,
+            help="Only convert experiments for a specific team (by slug)",
+        )
+        parser.add_argument(
+            "--experiment-id",
+            type=int,
+            help="Convert only a specific experiment by ID",
+        )
+        parser.add_argument(
+            "--chatbots-flag-only",
+            action="store_true",
+            help='Only convert experiments for teams that have the "flag_chatbots" feature flag enabled',
+        )
+
+    def handle(self, *args, **options):
+        dry_run = options["dry_run"]
+        team_slug = options.get("team_slug")
+        experiment_id = options.get("experiment_id")
+        chatbots_flag_only = options["chatbots_flag_only"]
+
+        query = Q(pipeline__isnull=True) & (Q(assistant__isnull=False) | Q(llm_provider__isnull=False))
+
+        if team_slug:
+            query &= Q(team__slug=team_slug)
+
+        if experiment_id:
+            query &= Q(id=experiment_id)
+
+        if chatbots_flag_only:
+            chatbots_flag_team_ids = self._get_chatbots_flag_team_ids()
+            if not chatbots_flag_team_ids:
+                self.stdout.write(self.style.WARNING('No teams found with the "flag_chatbots" feature flag enabled.'))
+                return
+            query &= Q(team_id__in=chatbots_flag_team_ids)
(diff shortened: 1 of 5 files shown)


apps/experiments/tasks.py
```diff
@@ -153,3 +154,18 @@ def _convert_prompt_builder_history(messages_history):
...
+
+
+@shared_task(bind=True, base=TaskbadgerTask)
+def migrate_experiment_to_pipeline_task(self, experiment_id: int) -> dict:
+    """
+    Migrate a single assistant or llm experiment to pipeline experiment.
+    """
+    try:
+        experiment = Experiment.objects.get(id=experiment_id)
+        call_command("migrate_nonpipeline_to_pipeline_experiments", experiment_id=experiment_id, skip_confirmation=True)
```
snopoke, line 166: A preferable alternative to calling the command would be to extract the logic into functions which can be used here and in the command. That also makes it easier to test.

apps/experiments/views/experiment.py
```diff
@@ -1488,3 +1493,21 @@ def get_release_status_badge(request, team_slug: str, experiment_id: int):
...
+
+def migrate_experiment_view(request, team_slug, experiment_id):
+    experiment = get_object_or_404(Experiment, id=experiment_id, team__slug=team_slug)
+    failed_url = reverse(
+        "experiments:single_experiment_home",
+        kwargs={"team_slug": team_slug, "experiment_id": experiment_id},
+    )
+    try:
+        task = migrate_experiment_to_pipeline_task.delay(experiment_id)
+        task.get(timeout=60)
```
snopoke, line 1506: If we're going to wait for the task to complete then we might as well just run it in the request thread instead of using Celery. Considering that it's just creating a new pipeline from existing data it should be fast so I don't see an issue with doing it synchronously.

apps/experiments/views/__init__.py
```diff
@@ -40,6 +40,7 @@
     get_export_download_link,
     get_message_response,
     get_release_status_badge,
+    migrate_experiment_view,
```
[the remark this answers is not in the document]
stephherbers: forgot to add to commit 🙈

apps/pipelines/helper.py
```diff
@@ -35,3 +35,83 @@ def duplicate_pipeline_with_new_ids(pipeline_data):
         edge["target"] = new_target_id
 
     return new_data, old_to_new_node_ids
+
+
+def convert_non_pipeline_experiment_to_pipeline(experiment):
```
SmittieC, line 40: Should we add a comment or two indicating that these methods are temporary?

apps/experiments/views/experiment.py
```diff
@@ -1497,3 +1501,24 @@ def get_release_status_badge(request, team_slug: str, experiment_id: int):
     experiment = get_object_or_404(Experiment, id=experiment_id, team=request.team)
     context = {"has_changes": experiment.compare_with_latest(), "experiment": experiment}
     return render(request, "experiments/components/unreleased_badge.html", context)
+
+
+def migrate_experiment_view(request, team_slug, experiment_id):
```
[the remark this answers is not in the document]
snopoke: The `permission_required` decorator accepts a tuple of permissions so you can actually do thi:
```suggestion
@login_required
@permission_required(("experiments.change_experiment", "pipelines.add_pipeline"))
def migrate_experiment_view(request, team_slug, experiment_id):
```

snopoke (commented):
I think this needs to be updated with the changes from the parent branch

apps/experiments/management/commands/migrate_nonpipeline_to_pipeline_experiments.py
```diff
@@ -141,3 +146,63 @@ def _get_experiment_type(self, experiment):
     def _get_chatbots_flag_team_ids(self):
         chatbots_flag = Flag.objects.get(name="flag_chatbots")
         return list(chatbots_flag.teams.values_list("id", flat=True))
+
+    def _create_pipeline_with_node(self, experiment, node_type, node_label, node_params):
+        """Create a pipeline with start -> custom_node -> end structure."""
+        pipeline_name = f"{experiment.name} Pipeline"
+
+        node = FlowNode(
+            id=str(uuid4()),
```
snopoke, line 155: see https://github.com/dimagi/open-chat-studio/pull/1721/files#r2144511440

snopoke (dismissed):

apps/experiments/views/experiment.py
```diff
@@ -1464,3 +1468,26 @@ def get_release_status_badge(request, team_slug: str, experiment_id: int):
     experiment = get_object_or_404(Experiment, id=experiment_id, team=request.team)
     context = {"has_changes": experiment.compare_with_latest(), "experiment": experiment}
     return render(request, "experiments/components/unreleased_badge.html", context)
+
+
+@login_required
```
snopoke, line 1473: This view also requires a team:
```suggestion
@login_and_team_required
```

PR #1726, opened 2025-06-10
comment out team context change log

snopoke:


snopoke:
> Is the plan to bring this back some time?

I would like to but it's not that high priority and I don't have the time right now. I think the solution is to refactor how the teams middleware works so that it doesn't automatically select a team unless it's very clear which one to pick.

PR #1727, opened 2025-06-10
version performance

snopoke:
## Description
Improve performance of versioning. Mostly this consists of reducing SQL queries.

Review by commit.

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -233,8 +233,7 @@ class SourceMaterial(BaseTeamModel, VersionsMixin):
         super().archive()
         self.experiment_set.update(source_material=None, audit_action=AuditAction.AUDIT)
 
-    @property
-    def version_details(self) -> VersionDetails:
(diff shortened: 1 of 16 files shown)


apps/experiments/models.py
```diff
@@ -345,8 +344,7 @@ def archive(self):
         self.experiments_pre.update(pre_survey=None, audit_action=AuditAction.AUDIT)
         self.experiments_post.update(post_survey=None, audit_action=AuditAction.AUDIT)
 
-    @property
-    def version_details(self) -> VersionDetails:
+    def _get_version_details(self) -> VersionDetails:
```
SmittieC, line 347: Why not make this a `cached_property`?

SmittieC (approved):
LGTM thanks. Good to get a refresher on points of optimization

apps/experiments/models.py
snopoke: Just so that we can centralise the caching

PR #1734, opened 2025-06-11
remove request from get_filtered_sessions

### LINKED ISSUE #1666: Refactor `apps.experiments.export.get_filtered_sessions` to remove request param
`apps.experiments.export.get_filtered_sessions` is used in views and also in celery tasks where there is no request. The method should not require a request object, rather all params should be passed in as plain data params. This is mostly already done but just requires some refactoring to clean it up:

```diff
- def get_filtered_sessions(request, experiment, query_params, timezone):
- def get_filtered_sessions(experiment, query_params, timezone):
```

barora05:
## Description
Remove request from get_filtered_sessions. Closes #1666

### DIFF
--- a/apps/experiments/views/experiment.py
+++ b/apps/experiments/views/experiment.py
@@ -135,10 +135,7 @@ class ExperimentSessionsTableView(LoginAndTeamRequiredMixin, SingleTableView, Pe
             .select_related("participant__user")
         )
         timezone = self.request.session.get("detected_tz", None)
-        referer = self.request.headers.get("referer") or ""
-        parsed_url = urlparse(referer)
-        query_params = parse_qs(parsed_url.query)
-        query_set = apply_dynamic_filters(query_set, query_params, timezone)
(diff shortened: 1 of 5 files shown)


SmittieC (approved):
LGTM

apps/experiments/views/experiment.py
```diff
@@ -134,7 +134,11 @@ def get_queryset(self):
             .filter(team=self.request.team, experiment__id=self.kwargs["experiment_id"])
             .select_related("participant__user")
         )
-        query_set = apply_dynamic_filters(query_set, self.request)
+        timezone = self.request.session.get("detected_tz", None)
+        referer = self.request.headers.get("referer") or ""
+        parsed_url = urlparse(referer)
+        query_params = parse_qs(parsed_url.query)
```
snopoke, line 140: This is a change from previous functionality right? Before it would just use `request.GET` for the params.

Is this necessary or better?

barora05: No advantage as such in the current usecase. I have reverted it to self.request.GET.

PR #1739, opened 2025-06-11
check session if exists

### LINKED ISSUE #1729: IntegrityError: duplicate key value violates unique constraint "experiments_experimentsession_public_id_bd443193_uniq"
Check slack session external ID uniquenss.

Sentry Issue: [OPEN-CHAT-STUDIO-10R](https://dimagi.sentry.io/issues/6670577345/?referrer=github_integration)

```
UniqueViolation: duplicate key value violates unique constraint "experiments_experimentsession_public_id_bd443193_uniq"
DETAIL:  Key (external_id)=(C08UVLUCSUE:1749569119.126549) already exists.
  File "django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
  File "psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)

IntegrityError: duplicate key value violates unique constraint "experiments_experimentsession_public_id_bd443193_uniq"
DETAIL:  Key (external_id)=(C08UVLUCSUE:1749569119.126549) already exists.
(18 additional frame(s) were not displayed)
...
  File "apps/slack/slack_listeners.py", line 59, in respond_to_message
    _respond_to_message(event, channel_id, thread_ts, experiment_channel, experiment, session, context)
  File "apps/slack/slack_listeners.py", line 67, in _respond_to_message
    session = SlackChannel.start_new_session(
  File "apps/chat/channels.py", line 153, in start_new_session
    return _start_experiment_session(
  File "apps/chat/channels.py", line 1197, in _start_experiment_session
    session = ExperimentSession.objects.create(
  File "apps/experiments/models.py", line 1573, in save
    super().save(*args, **kwargs)

Failed to run listener function (error: duplicate key value violates unique constraint "experiments_experimentsession_public_id_bd443193_uniq"
DETAIL:  Key (external_id)=(C08UVLUCSUE:1749569119.126549) already exists.)
```

barora05:
## Description
Did not encounter this exception locally after connecting to Slack. However, this issue could have been caused by a race condition, where multiple messages with the same thread_ts tried to create an ExperimentSession at the same time.
To handle this, I’ve added a check to see if a session already exists before attempting to create a new one. Additionally, if an IntegrityError is raised during creation (due to a concurrent attempt), the existing session is fetched and used instead.

This ensures that only one session is created per unique external_id, avoiding duplicate key errors.
Closes #1729

### DIFF
--- a/apps/slack/slack_listeners.py
+++ b/apps/slack/slack_listeners.py
@@ -10,7 +10,6 @@ Manage event subscriptions at:
 import logging
 import re
 
-from django.db import IntegrityError
 from django.db.models import Q
 from slack_bolt import BoltContext, BoltResponse
 
@@ -65,17 +64,12 @@ def _respond_to_message(event, channel_id, thread_ts, experiment_channel, experi
 
     if not session:
         external_id = make_session_external_id(channel_id, thread_ts)
-        session = ExperimentSession.objects.filter(external_id=external_id).first()
-        if not session:
-            try:
-                session = SlackChannel.start_new_session(
-                    working_experiment=experiment,
-                    experiment_channel=experiment_channel,
-                    participant_identifier=slack_user,
-                    session_external_id=external_id,
-                )
-            except IntegrityError:
-                session = ExperimentSession.objects.get(external_id=external_id)
+        session = SlackChannel.start_new_session(
+            working_experiment=experiment,
+            experiment_channel=experiment_channel,
+            participant_identifier=slack_user,
+            session_external_id=external_id,
+        )
     # strip out the mention
     message_text = re.sub(rf"<@?{context.bot_user_id}>", "", event["text"])
     message = SlackMessage(
--- a/apps/slack/slack_listeners.py
+++ b/apps/slack/slack_listeners.py
@@ -10,6 +10,7 @@ Manage event subscriptions at:
 import logging
 import re
 
+from django.db import IntegrityError
 from django.db.models import Q
 from slack_bolt import BoltContext, BoltResponse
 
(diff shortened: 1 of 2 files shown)


SmittieC (commented):
We could also make [this query](https://github.com/dimagi/open-chat-studio/blob/a7c400a505d25a480a1170cb4aa80fc9ece82c37/apps/chat/channels.py#L1161-L1169) a `get_or_create` instead

apps/slack/slack_listeners.py
```diff
@@ -64,13 +65,17 @@ def _respond_to_message(event, channel_id, thread_ts, experiment_channel, experi
...
+        if not session:
+            try:
+                session = SlackChannel.start_new_session(
+                    working_experiment=experiment,
+                    experiment_channel=experiment_channel,
+                    participant_identifier=slack_user,
+                    session_external_id=external_id,
+                )
+            except IntegrityError:
+                session = ExperimentSession.objects.get(external_id=external_id)
```
snopoke: I don't think we should optimistically re-fetch the session. This seems like a rare edge case where 2 messages came in one after the other, the 2nd one before the 1st had completed.

What do you think of this approach:
```suggestion
          try:
              session = SlackChannel.start_new_session(
                  working_experiment=experiment,
                  experiment_channel=experiment_channel,
                  participant_identifier=slack_user,
                  session_external_id=external_id,
              )
          except IntegrityError as e:
              if external_id in str(e):
                  # make sure we only do this if there issue is the external ID
                  session = ExperimentSession.objects.get(external_id=external_id)
              else:
                  raise e
```

snopoke (approved):
Looks like I'm late to the party. I think the current version is OK so you can ignore my review comments.

PR #1741, opened 2025-06-12
update particiapnt and session data in trigger bot api

barora05:

### DIFF
--- a/api-schema.yml
+++ b/api-schema.yml
@@ -989,6 +989,12 @@ components:
           type: boolean
           default: false
           title: Starts a new session
+        session_data:
+          type: object
+          additionalProperties: {}
+        participant_data:
+          type: object
+          additionalProperties: {}
       required:
       - experiment
       - identifier
--- a/apps/api/serializers.py
+++ b/apps/api/serializers.py
@@ -197,3 +197,5 @@ class TriggerBotMessageRequest(serializers.Serializer):
     experiment = serializers.UUIDField(label="Experiment ID")
     prompt_text = serializers.CharField(label="Prompt to go to bot")
     start_new_session = serializers.BooleanField(label="Starts a new session", required=False, default=False)
+    session_data = serializers.DictField(label="Session Data", required=False, default=dict)
+    participant_data = serializers.DictField(label="Participant Data", required=False, default=dict)
--- a/apps/api/views.py
+++ b/apps/api/views.py
@@ -548,6 +548,10 @@ def trigger_bot_message(request):
 
     if platform == ChannelPlatform.COMMCARE_CONNECT and not participant_data.has_consented():
         return JsonResponse({"detail": "User has not given consent"}, status=status.HTTP_400_BAD_REQUEST)
+    if participant_data and data["participant_data"]:
+        merged_data = {**participant_data.data, **data["participant_data"]}
+        participant_data.data = merged_data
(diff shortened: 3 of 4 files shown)


api-schema.yml
```diff
@@ -989,6 +989,12 @@ components:
           type: boolean
           default: false
           title: Starts a new session
+        session_data:
```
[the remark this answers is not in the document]
barora05: added help_text. text appears as description

apps/api/views.py
```diff
@@ -548,6 +548,10 @@ def trigger_bot_message(request):
 
     if platform == ChannelPlatform.COMMCARE_CONNECT and not participant_data.has_consented():
         return JsonResponse({"detail": "User has not given consent"}, status=status.HTTP_400_BAD_REQUEST)
+    if participant_data and data["participant_data"]:
```
snopoke, line 551: I know the serializer has `default=dict` but I would still use `data.get("participant_data")` here for safety.

Also, if the request includes participant data but the DB object doesn't exist, we should create one.

api-schema.yml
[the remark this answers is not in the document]
snopoke: I think you can still update the example though:

```
OpenApiExample(
    name="GenerateBotMessageAndSend",
    summary="Generates a bot message and sends it to the user",
    value={
        "identifier": "part1",
        "experiment": "exp1",
        "platform": "connect_messaging",
        "prompt_text": "Tell the user to do something",
        "session_data": {"key": "value"},
        "participant_data": {"key": "value"},
    },
    status_codes=[200],
),
```

apps/api/serializers.py
```diff
@@ -197,3 +197,5 @@ class TriggerBotMessageRequest(serializers.Serializer):
     experiment = serializers.UUIDField(label="Experiment ID")
     prompt_text = serializers.CharField(label="Prompt to go to bot")
     start_new_session = serializers.BooleanField(label="Starts a new session", required=False, default=False)
+    session_data = serializers.DictField(help_text="Update session data", required=False, default=dict)
+    participant_data = serializers.DictField(help_text="Update Participant Data", required=False, default=dict)
```
snopoke, line 201: Add note about data being merged:
```suggestion
    session_data = serializers.DictField(help_text="Update session data. This will be merged with existing data.", required=False, default=dict)
    participant_data = serializers.DictField(help_text="Update participant data. This will be merged with existing data.", required=False, default=dict)
```

apps/api/views.py
```diff
@@ -548,6 +550,16 @@ def trigger_bot_message(request):
 
     if platform == ChannelPlatform.COMMCARE_CONNECT and not participant_data.has_consented():
         return JsonResponse({"detail": "User has not given consent"}, status=status.HTTP_400_BAD_REQUEST)
+    if data.get("participant_data"):
+        if participant_data:
+            merged_data = {**participant_data.data, **data["participant_data"]}
+            participant_data.data = merged_data
+            participant_data.save(update_fields=["data"])
+        else:
+            participant = Participant.objects.filter(identifier=identifier, platform=platform).first()
```
snopoke, line 559: This needs to filter by team. Also rather use `.get` which will fail loudly if the participant doesn't exist.
```suggestion
            participant = Participant.objects.get(team=request.team, identifier=identifier, platform=platform)
```

snopoke (approved):

PR #1753, opened 2025-06-13
Transcript translation

stephherbers:
## Description

resolves [#1732](https://github.com/dimagi/open-chat-studio/issues/1732)

left todo:
demo + docs

## User Impact

users will be able to select which language they would like the chat messages to be in in the analysis tool (currently FF'ed)

### Demo

coming soon

### Docs and Changelog

docs + changelog


apps/analysis/tasks.py
```diff
@@ -61,11 +66,24 @@ def process_transcript_analysis(self, analysis_id):
                 progress_recorder.set_progress(
                     progress_value, 100, description=f"Processing session {index + 1}/{total_sessions}"
                 )
+                messages = list(session.chat.messages.all().order_by("created_at"))
```
SmittieC, line 69: Might be good to use a queryset iterator here so save on memory

apps/analysis/translation.py
```diff
@@ -0,0 +1,83 @@
...
+            Messages to translate:
+            {json.dumps(message_data)}
+            Output only the JSON array, without any additional text or explanation.
+            """
+            response = llm.invoke(prompt)
+            translated_data = json.loads(response.content)
+
+            for i, msg in enumerate(messages_to_translate):
+                if i < len(translated_data):
+                    msg.translations[target_language] = translated_data[i]["translated_text"]
```
SmittieC, line 55: Am I right that this assumes that the LLM's output is in the same order that we gave the messages in? It feelds risky?

I haven't thought the following through, but what about making the original message's id the key and the set of translations the value. For instance:

```JSON
{
"1": {"afr": "Hi", "eng": "Hi"},
"2": {"afr": "Hoe gaan dit?", "eng": "How are you?"},
}
```

apps/analysis/tests/__pycache__/test_translations.py
```diff
@@ -0,0 +1,156 @@
+import json
```
SmittieC, line 1: I think you accidentely put this file in the __pycache__ folder

apps/analysis/translation.py
```diff
@@ -0,0 +1,79 @@
+import json
+
+from apps.teams.utils import current_team
+
+from .const import LANGUAGE_CHOICES
+
+TRANSLATION_PROMPT_TEMPLATE = """Translate these chat messages to {target_lang_name}.
+Return a JSON array where each object has a "translation" field.
```
snopoke, line 8: This needs to also include the ID now since you're using it in the output
```suggestion
Return a JSON array where each object has the following fields:
- id: The ID of the message (must match the ID of of the input message)
- translation: The translated text
```

apps/analysis/translation.py
```diff
@@ -0,0 +1,79 @@
+import json
+
+from apps.teams.utils import current_team
+
+from .const import LANGUAGE_CHOICES
+
+TRANSLATION_PROMPT_TEMPLATE = """Translate these chat messages to {target_lang_name}.
```
snopoke, line 7: Did you mean to change this when you extracted it to a constant. Before that it was more detailed:

```
### Instructions
Translate chat messages to {target_lang_name}. Return a JSON array where each object
contains the translation.
**Response format:** Each object should have:
- "translation": the {target_lang_name} translation of the message content
**Translation rules:**
- If the text is already in {target_lang_name}, return the original text unchanged
- Preserve the order, meaning, and tone of the original messages
### Messages to translate:
```json
{message_data}
```
### Final instructions
Output only the JSON array with translations, without any additional text or explanation.

apps/analysis/translation.py
```diff
@@ -0,0 +1,83 @@
...
+
+            prompt = f"""Please translate these chat messages to {target_lang_name}. Return a JSON array
+            where each object has:
+            - "translated_text": the {target_lang_name} translation
+            If the text is already in {target_lang_name}, the value of "translated_text" should be the original text.
+            Messages to translate:
+            {json.dumps(message_data)}
+            Output only the JSON array, without any additional text or explanation.
+            """
+            response = llm.invoke(prompt)
```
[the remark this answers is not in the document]
snopoke: You can't use `with_structured_output` if the output is a list, it has to be a dict.

apps/analysis/translation.py
```diff
@@ -0,0 +1,79 @@
...
+
+            language_names = dict(choice for choice in LANGUAGE_CHOICES if choice[0])
+            target_lang_name = language_names.get(target_language, target_language)
+
+            prompt = TRANSLATION_PROMPT_TEMPLATE.format(
+                target_lang_name=target_lang_name, message_data=json.dumps(message_data)
+            )
+
+            response = llm.invoke(prompt)
+            translated_data = json.loads(response.content)
```
snopoke, line 57: Need to handle json decode errors here

apps/analysis/translation.py
```diff
@@ -0,0 +1,79 @@
...
+
+            for item in translated_data:
+                message_id = item["id"]
+                for message in messages_to_translate:
```
snopoke, line 61: It would be better to make a dict key'd by ID so that you don't have to iterate every time

apps/analysis/translation.py
```diff
@@ -0,0 +1,79 @@
...
+                    if str(message.id) == message_id:
+                        message.translations[target_language] = item["translation"]
+                        message.save(update_fields=["translations"])
```
snopoke, line 64: It would be more efficient to use `Messages.objects.bulk_update(updated_messages, fields=["translations"])`

apps/analysis/translation.py
```diff
@@ -0,0 +1,94 @@
...
+from apps.teams.utils import current_team
+
+from .const import LANGUAGE_CHOICES
+
+TRANSLATION_PROMPT_TEMPLATE = """### Instructions
+Translate chat messages to {target_lang_name}. Return a JSON array where each object has the following fields:
+- id: The ID of the message (must match the ID of of the input message)
+- translation: The translated text
+**Response format:** Each object should have:
+- "translation": the {target_lang_name} translation of the message content
```
snopoke, line 13:
```suggestion
```

apps/analysis/translation.py
```diff
@@ -0,0 +1,94 @@
...
+**Translation rules:**
+- If the text is already in {target_lang_name}, return the original text unchanged
+- Preserve the order, meaning, and tone of the original messages
+### Messages to translate:
+```json
+{message_data}
```
snopoke, line 19:
```suggestion
<messages>
{message_data}
</messages>
```

apps/analysis/translation.py
```diff
@@ -0,0 +1,94 @@
...
+
+from apps.chat.models import ChatMessage
+from apps.teams.utils import current_team
+
+from .const import LANGUAGE_CHOICES
+
+TRANSLATION_PROMPT_TEMPLATE = """### Instructions
+Translate chat messages to {target_lang_name}. Return a JSON array where each object has the following fields:
+- id: The ID of the message (must match the ID of of the input message)
+- translation: The translated text
```
snopoke, line 11:
```suggestion
- translation: The translated text
<example>
[{"id": "4", "translation": "Hello"}, {"id": "9", "translation": "Hi, how can I help?"}]
</example>
```

apps/analysis/translation.py
```diff
@@ -0,0 +1,94 @@
...
+**Response format:** Each object should have:
+- "translation": the {target_lang_name} translation of the message content
+**Translation rules:**
+- If the text is already in {target_lang_name}, return the original text unchanged
+- Preserve the order, meaning, and tone of the original messages
+### Messages to translate:
+```json
+{message_data}
+Final instructions
```
snopoke, line 20:
```suggestion
Final instructions:
```

apps/analysis/translation.py
```diff
@@ -0,0 +1,94 @@
...
+    pass
+
+
+def translate_messages_with_llm(messages, target_language, llm_provider, llm_provider_model):
+    """
+    Translate chat messages using the specified LLM provider and model
+    Only translates messages that don't already have the target language translation.
+    """
+    messages_to_translate = []
+    message_indices = []
```
snopoke, line 34:
```suggestion
```

apps/analysis/translation.py
```diff
@@ -0,0 +1,94 @@
...
+
+    for i, msg in enumerate(messages):
+        if not msg.translations:
+            msg.translations = {}
+
+        if target_language not in msg.translations:
+            messages_to_translate.append(msg)
+            message_indices.append(i)
```
snopoke, line 42:
```suggestion
```

apps/experiments/export.py
```diff
@@ -83,6 +91,8 @@ def filtered_export_to_csv(experiment, sessions_queryset):
                 _format_comments(message.comments.all()),
                 trace_id,
             ]
+            if translation_language:
+                row.append(translation_language)
```
snopoke, line 95:
```suggestion
                row.append(translation_language)
                row.append(message.content)
```

snopoke (approved):

apps/analysis/translation.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def translate_messages_with_llm(messages, target_language, llm_provider, llm_provider_model):
+    """
+    Translate chat messages using the specified LLM provider and model
+    Only translates messages that don't already have the target language translation.
+    """
+    messages_to_translate = []
+
+    for i, msg in enumerate(messages):
```
snopoke, line 37:
```suggestion
    for msg in messages:
```

snopoke (approved):

PR #1757, opened 2025-06-18
Parallel pipeline nodes

snopoke:
## Description
This PR makes it possible (and mostly safe) to have nodes in a pipeline running in parallel. The primary changes are:

1. Allow dangling nodes (nodes whose outputs are not connected)
2. Allow router nodes outputs to dangle (don't require / any router node outputs to be connected)
3. Additional utilities in the Python Node to support parallel runs
     * `get_node_output`: get the output of any node by name
     * `require_node_outputs`: 'wait' until all nodes listed have been executed
     * `abort_with_message`: abort the pipeline once all running nodes have completed

## User Impact
The primary use case it to allow running safety checks in parallel so that they don't slow down the response to the user. But I'm sure there will be other use cases that emerge.

### Demo
**Safety layer with abort**

```mermaid
flowchart TD
    start["start"] --> Safety["SafetyRouter"] & LLM
    Safety -. unsafe .-> PythonNode["PythonNode *abort_with_message*"]
    LLM --> __end__(["<p>end</p>"])

     start:::first
     __end__:::last
```

**Router for tagging only**
```mermaid
flowchart TD
    start["start"] --> LLM & Router
    LLM --> __end__(["<p>end</p>"])

     start:::first
     __end__:::last
```

### Docs and Changelog
TODO


apps/pipelines/nodes/base.py
```diff
@@ -176,6 +189,43 @@ def get_all_routes(self) -> dict:
...
+        Get the inputs for the given node based on the node's incoming edges.
+
+        Returns:
+            A dictionary mapping incoming node IDs to their respective outputs in the state.
+            If there is no output for an incoming edge or if that edge was not targeted,
+            the value in the output will be None.
+        """
+        inputs = {}
+        for incoming_node_id in incoming_nodes:
+            targets = [step[2] for step in self["path"] if step[1] == incoming_node_id]
```
SmittieC, line 221: I'm having a hard time keeping track of what the `step` indexes mean. How about we use constants to indicate what each index refers to?

apps/pipelines/nodes/base.py
```diff
@@ -297,20 +326,30 @@ def _process(self, state: PipelineState) -> PipelineState:
...
     def process(
-        self, incoming_edges: list, outgoing_edges: list, state: PipelineState, config: RunnableConfig
-    ) -> PipelineState:
+        self, incoming_nodes: list, outgoing_nodes: list, state: PipelineState, config: RunnableConfig
+    ) -> PipelineState | Command:
         self._config = config
-        state = self._prepare_state(self.node_id, incoming_edges, state)
+        self._incoming_nodes = incoming_nodes
+        self._outgoing_nodes = outgoing_nodes
+        state = PipelineState(state)
```
SmittieC, line 334: Is `state` a dict here? The type hint says it's `PipelineState`, but I suppose that wrong then?

apps/pipelines/nodes/nodes.py
```diff
@@ -1146,10 +1157,34 @@ def _get_custom_globals(self, node_id, state: PipelineState):
...
+        def abort_pipeline(message, tag_name: str = None):
+            raise AbortPipeline(message, tag_name)
+
+        return abort_pipeline
+
+    def _require_node_outputs(self, state: PipelineState):
+        """A helper function to require inputs from a specific node"""
+
+        def require_node_outputs(*node_names):
+            if len(node_names) == 1 and isinstance(node_names[0], list):
```
SmittieC, line 1178: When will this be the case?

apps/pipelines/nodes/nodes.py
```diff
@@ -1146,10 +1157,34 @@ def _get_custom_globals(self, node_id, state: PipelineState):
...
+                node_names = node_names[0]
+            if not all(isinstance(name, str) for name in node_names):
+                raise PipelineNodeRunError("node names pass to 'require_node_outputs' must be a string")
```
SmittieC, line 1181:
```suggestion
                raise PipelineNodeRunError("Node names passed to 'require_node_outputs' must be a string")
```

apps/pipelines/nodes/nodes.py
```diff
@@ -1110,10 +1111,20 @@ def _process(self, input: str, state: PipelineState) -> PipelineState:
         kwargs = {}
         try:
             exec(byte_code, custom_globals, custom_locals)
-            result = str(custom_locals[function_name](input, **kwargs))
+            result = custom_locals[function_name](input, **kwargs)
+        except WaitForNextInput:
```
SmittieC, line 1115: What is the effect of this on the pipeline's execution? Will it end the pipeline?

SmittieC (commented):
I'm yet to finish review, but a few comments so far

snopoke:
@sentry review

SmittieC (approved):

Issue #1760, opened 2025-06-19
Allow referencing temp_state and session_state from prompts


snopoke:
This allows context from the state to be used in prompts.

The current use case is to inject context into the prompt from a Python Node.

PR #1761, opened 2025-06-19
support using temp_state and session_state in pipeline prompts

snopoke:
## Description
Resolves https://github.com/dimagi/open-chat-studio/issues/1760

## User Impact
Users can reference data from `temp_state` and `session_state` in the prompts.

### Docs and Changelog
TODO


SmittieC (approved):
LGTM

Issue #1762, opened 2025-06-19
Allow pipeline nodes to end sessions by using a tool


snopoke:
Provide a tool to LLM nodes and python nodes which will end the current session when the tool is used.

The session state should only take place at the end of the pipeline run.

PR #1764, opened 2025-06-19
end session tool

snopoke:
## Description
Resolves https://github.com/dimagi/open-chat-studio/issues/1762

This adds a new tool which allows bots to end the current session.

## User Impact
Bot builders can use the tool to end sessions based on bot activity.

### Docs and Changelog
TODO

### DIFF
--- a/apps/chat/agent/tools.py
+++ b/apps/chat/agent/tools.py
@@ -1,8 +1,7 @@
 import logging
-from collections.abc import Callable
 from datetime import datetime, timedelta
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, Union
+from typing import TYPE_CHECKING, Any, ClassVar, Union
 
 from django.db import transaction, utils
 from langchain_community.utilities.openapi import OpenAPISpec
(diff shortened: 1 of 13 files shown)


apps/chat/agent/tools.py
```diff
@@ -183,12 +187,28 @@ def action(self, key: str, value: Any):
         return "Success"
 
 
+class EndSessionTool(CustomBaseTool):
+    requires_callbacks: ClassVar[bool] = True
+    name: str = AgentTools.END_SESSION
+    description: str = (
+        "End the current chat session. "
+        "This will mark the session as completed. "
+        "New messages will result in a new session being createed."
```
SmittieC, line 196:
```suggestion
        "New messages will result in a new session being created."
```

apps/experiments/models.py
```diff
@@ -566,13 +566,18 @@ class AgentTools(models.TextChoices):
...
     UPDATE_PARTICIPANT_DATA = "update-user-data", gettext("Update Participant Data")
     ATTACH_MEDIA = "attach-media", gettext("Attach Media")
+    END_SESSION = "end-session", gettext("End Session")
 
     @classmethod
     def reminder_tools(cls) -> list[Self]:
         return [cls.RECURRING_REMINDER, cls.ONE_OFF_REMINDER, cls.DELETE_REMINDER, cls.MOVE_SCHEDULED_MESSAGE_DATE]
 
+    @classmethod
+    def legacy_experiment_tools(cls):
```
SmittieC, line 576: Since the legacy tool group is only used in the tests, how do you feel about making this a fixture or something? I was initially confused by why we make a distinction between these tools and the "pipeline specific" tools, but looking at the test it makes sense.

apps/pipelines/nodes/base.py
```diff
@@ -64,6 +64,12 @@ def merge_dicts(left: dict, right: dict):
     return output
 
 
+class Intents(StrEnum):
```
SmittieC, line 67: :tent:

SmittieC (approved):

PR #1766, opened 2025-06-20
add model deprecation flow

barora05:


snopoke (dismissed):

barora05:
@snopoke any idea why tests are faiing. I have added the migration script for deprecated field.
`AttributeError: 'LlmProviderModel' object has no attribute 'deprecated'`

snopoke:
> @snopoke any idea why tests are faiing. I have added the migration script for deprecated field. `AttributeError: 'LlmProviderModel' object has no attribute 'deprecated'`

It's happening because there are DB migrations that are calling `_update_llm_provider_models` before the migration that adds the `deprecated` field. You should remove all of those migration operations except for the one in your migration in this PR.

snopoke:
> > > @snopoke any idea why tests are faiing. I have added the migration script for deprecated field. `AttributeError: 'LlmProviderModel' object has no attribute 'deprecated'`
> > 
> > 
> > It's happening because there are DB migrations that are calling `_update_llm_provider_models` before the migration that adds the `deprecated` field. You should remove all of those migration operations except for the one in your migration in this PR.
> 
> I have only 1 new migration in this PR currently, which is for adding the deprecated filed in table. Why are earlier migrations which calls _update_llm_provider_models running again?

It happens in the tests because they start with an empty database.

apps/analysis/migrations/0012_initial.py
```diff
@@ -13,7 +13,6 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('experiments', '0110_alter_participantdata_data'),
-        ('service_providers', '0026_add_google_gemini_models'),
```
snopoke, line 16: this should be added back

apps/documents/migrations/0003_rename_files_collection_files_old_and_more.py
```diff
@@ -24,7 +24,6 @@ class Migration(migrations.Migration):
     dependencies = [
         ('documents', '0002_remove_collection_unique_collection_per_team_and_more'),
         ('files', '0006_remove_file_schema_file_metadata'),
-        ('service_providers', '0026_add_google_gemini_models'),
```
snopoke, line 27: this should be added back

apps/service_providers/migrations/0031_deprecate_models.py
snopoke, line 1: this migration can be merged with the previous one

apps/service_providers/migrations/0032_alter_llmprovider_type_alter_llmprovidermodel_type.py
snopoke, line 1: I don't see any code changes that correspond to this migration and running `makemigrations` on `main` doesn't produce it either so I think it can be removed.

apps/service_providers/migrations/0026_add_google_gemini_models.py
```diff
@@ -18,34 +18,3 @@ class Migration(migrations.Migration):
...
-            field=models.CharField(choices=[
-                ('openai', 'OpenAI'),
-                ('azure', 'Azure OpenAI'),
-                ('anthropic', 'Anthropic'),
-                ('groq', 'Groq'),
-                ('perplexity', 'Perplexity'),
-                ('deepseek', 'DeepSeek'),
-                ('google', 'Google Gemini')
-            ], max_length=255),
-        ),
```
snopoke, line 50: These should not be deleted, only the `RunPython` operation

apps/service_providers/migrations/0032_alter_llmprovider_type_alter_llmprovidermodel_type.py
snopoke: I see the issue - it's because you deleted some model operations in apps/service_providers/migrations/0026_add_google_gemini_models.py

snopoke (approved):

PR #1770, opened 2025-06-23
Detangle annotated file ids

### LINKED ISSUE #1769: Remote index file citation issue
### Observed behaviour
When using the responses API, file citations doesn't seem to work all the time.

### Investigation notes
- In one langfuse trace I can see two file ids being concatenated together, making up the annotated file id. Ofcourse this file will not exist

SmittieC:
## Description

Resolves #1769 

## User Impact

User should be able to see/download cited files again

### Demo

N/A

### Docs and Changelog

Pending


apps/service_providers/llm_service/main.py
```diff
@@ -42,6 +43,21 @@ class AnthropicBuiltinTool(dict):
...
+    """
+    There is a bug in the OpenAI API where separate file ids are sometimes returned concatenated together.
+
+    Example:
+        file-123Abcfile-123Def
+    Should be parsed as:
+        ['file-123Abc', 'file-123Def']
+    """
+    detangled_file_ids = []
+    for file_id in file_ids:
```
stephherbers, line 56: type mismatch for file_ids-- the input is a list of strings--- but if you do have it as a single string cant you just return the re.findall output?

stephherbers (changes requested):

PR #1772, opened 2025-06-24
feat: Add RAG file chunks view for collections

### LINKED ISSUE #1713: RAG - View to see file chunks
We need to build a view where users can easily look at a file's chunks (if they exist). [Here's some](https://claude.ai/public/artifacts/e7d3c3a4-e8d9-4b27-b906-5971b7fdfdde) inspiration for what the UI might look like.

The user should be able to click on any file listed in a collection to see its chunks.

SmittieC:
## Description

Fixes #1713
Generated with [Claude Code](https://claude.ai/code) (mostly)

## User Impact

A simple view where users can inspect their file chunks

### Demo

demo

### Docs and Changelog

Pending

### DIFF
--- a/apps/documents/views.py
+++ b/apps/documents/views.py
@@ -12,7 +12,7 @@ from django.template.loader import render_to_string
 from django.urls import reverse
 from django.views import View
 from django.views.decorators.http import require_POST
-from django.views.generic import CreateView, FormView, TemplateView, UpdateView
+from django.views.generic import CreateView, FormView, ListView, TemplateView, UpdateView
 from django_tables2 import SingleTableView
 
 from apps.documents import tasks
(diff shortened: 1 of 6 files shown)


apps/documents/views.py
```diff
@@ -349,3 +349,49 @@ def form_valid(self, form):
...
+    paginate_by = 10
+    permission_required = ("documents.view_collection", "files.view_file")
+
+    def get_queryset(self):
+        collection_id = self.kwargs["collection_id"]
+        file_id = self.kwargs["file_id"]
+
+        # Get chunks for this file in this collection, ordered by chunk number
+        return FileChunkEmbedding.objects.filter(
+            collection_id=collection_id, file_id=file_id, is_archived=False
```
snopoke, line 369: This should also filter by team just to be safe

apps/documents/views.py
```diff
@@ -349,3 +349,49 @@ def form_valid(self, form):
...
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        team_slug = self.kwargs["team_slug"]
+        collection_id = self.kwargs["collection_id"]
+        file_id = self.kwargs["file_id"]
+
+        collection_file = get_object_or_404(
+            CollectionFile.objects.select_related("file", "collection"),
```
snopoke, line 380: these `select_related` don't seem necessary for this view

snopoke (approved):
LGTM

apps/documents/views.py
SmittieC: We're injecting the `collection`, `file` into the context below that, but we get them from this object

apps/documents/views.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/1772/commits/7593e11431639c009d5f770e7f54d4caed087c68

templates/documents/partials/pagination_controls.html
snopoke, line 1: no need to change but just FYI we do have django-partials installed which is ideal for this kind of thing

PR #1775, opened 2025-06-24
Ba/attachments outside assistant

### LINKED ISSUE #1736: Support uploading attachments outside of Assistants
Currently chat uploads are limited to assistants however most APIs now support attachments so it would be useful to allow users to upload attachments.

## Requirements

* Add a new experiment level setting to configure whether uploads are allowed.
* If uploads are allowed, display an upload button in the web chat which allows the user to upload files. This will be similar to the existing upload buttons however it does not need to categorise the files as 'file_search' or 'code_interpreter'.

barora05:
## Description
Allow option to attach files outside of assistants.
1. Only allows for experiment with file_uploads_enabled true
2. Pass input as multi modal input
Closes #1736


templates/experiments/chat/input_bar.html
```diff
@@ -114,19 +147,20 @@ <h4 class="font-bold">File Search Files</h4>
       message: "",
       code_interpreter_files: [],
       file_search_files: [],
+      experiment_files: [],
       total_upload_size_bytes: 0,
       disableCodeInterpreterUpload: false,
       init() {
-        let fieldsToWatch = ['message']
+        let fieldsToWatch = ['message', 'experiment_files'];
```
snopoke, line 154: This should only be added if `this.$refs.experimentUpload` exists, same pattern as the others.

snopoke (commented):
Generally looks good. A few small comments.

apps/channels/datamodels.py
```diff
@@ -10,7 +10,7 @@
 
 logger = logging.getLogger("ocs.channels")
 
-AttachmentType = Literal["code_interpreter", "file_search"]
+AttachmentType = Literal["code_interpreter", "file_search", "experiment_files"]
```
snopoke, line 13: I don't think it really matters what this is called but I did prefer what you had before which was `ocs_attachments`

templates/experiments/chat/input_bar.html
```diff
@@ -114,19 +147,23 @@ <h4 class="font-bold">File Search Files</h4>
...
+          fieldsToWatch.push('code_interpreter_files');
+        }
+        if (this.$refs.experimentUpload) {
+          fieldsToWatch.push('ocs_attachments');
         }
-        this.$watchMany(fieldsToWatch, (message, cfiles, ffiles) => {
+        this.$watchMany(fieldsToWatch, (message, experimentFiles, cfiles = [], ffiles = []) => {
           const submitButton = document.getElementById('message-submit');
-          if (message || cfiles.length || ffiles.length) {
+          if (message || experimentFiles.length || cfiles.length || ffiles.length) {
```
snopoke, line 166: I think this is a bug since the function could get called with a different number of arguments depending on which fields are being watched. In this specific case is still works since the function doesn't change if you assign `cfile` to `experimentFiles` but it would good to fix it anyway to prevent future bugs:


```suggestion
        this.$watchMany(fieldsToWatch, (...args) => {
          const submitButton = document.getElementById('message-submit');
          if (args.some(Boolean)) {
```

snopoke (commented):
One comment but otherwise this looks good. Are you able to record a demo of it working?

barora05:
> One comment but otherwise this looks good. Are you able to record a demo of it working?

Here is the [demo link](https://drive.google.com/file/d/10GrQmrY8WPV3WCwJ6AaEESXbdw_rb3GD/view?usp=drive_link) after latest changes.

snopoke (approved):

PR #1783, opened 2025-06-26
Prompt the bot to cite the files it generates output from

### LINKED ISSUE #1650: RAG - Citations
The [spec](https://docs.google.com/document/d/1_BScbohNxwysXdeR9ibeRZ07Ojy3m9yFQ2VBQDBYCBM/edit?tab=t.ik6hmw5sgrpx#heading=h.deyporb0b4oy)

Depends on
- https://github.com/dimagi/open-chat-studio/issues/1649

## Requirement
The files - from the set of files that were retrieved -  that were ultimately used to generate the final response, needs to be attached to the chat and cited in the response. Investigate the [citation methods described in the spec](https://docs.google.com/document/d/1_BScbohNxwysXdeR9ibeRZ07Ojy3m9yFQ2VBQDBYCBM/edit?tab=t.ik6hmw5sgrpx#heading=h.5lyxvdixycsb) and use the one that works best.

### How to attach a file?
You need to ensure that a proper `ChatAttachment` instance exists for the current `chat`. Then, add the cited files to the set of files for this chat attachment. Then, add the files' id to the specific chat message's metadata so that we can know which attachments exists for this message.

SmittieC:
## Description

Resolves #1650
A simple approach to citing chunks that were used to generate the response. I did not use the OCS style markdown link (`[file name](file:<team slug>:<session id>:<file id>)`) for citations, because these links are only resolved for web channels. Some channels don'tt support markdown, so even if we render these links for external channels, it would look weird for those channels. For now, I'm keeping things simple.

## User Impact

Files that were used to generate the response will be attached to the response

### Demo

https://www.loom.com/share/6f13741b6b784005bc59879f708b8d27?sid=efea6157-fb71-4ef9-80af-5b661f99180b

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/120


apps/chat/agent/tools.py
```diff
@@ -26,19 +26,45 @@
...
 {chunk}
 """
 
+CHUNK_RESULT_TEMPLATE = """
+# Retrieved chunks
+{retrieved_chunks}
+
+**CRITICAL REQUIREMENT - MANDATORY CITATIONS:**
+
+You MUST cite all information using this exact format: <CIT file_id=the-file-id />
```
snopoke, line 51: what about just `<CIT the-file-id>`? If you want it to look like XML then maybe add quotes: `<CIT file_id="the-file-id"/>`. Either way you should probably handle quotes and no quotes.

apps/service_providers/tests/llm_service/test_utils.py
```diff
@@ -0,0 +1,45 @@
...
+    ]
+
+
+@pytest.mark.parametrize(
+    ("input_text", "expected_file_ids"),
+    [
+        ("", []),
+        ("No citations here", []),
+        ("<CIT file_id=123 />", ["123"]),
+        ("A citation <CIT file_id=123 />. Another one<CIT file_id=456 />.", ["123", "456"]),
```
snopoke, line 26: also test text with newlines and with other XML tags

snopoke (commented):

apps/service_providers/llm_service/utils.py
```diff
@@ -0,0 +1,34 @@
...
+def extract_file_ids_from_ocs_citations(text: str) -> list[int]:
+    from apps.chat.agent.tools import OCS_CITATION_PATTERN
+
+    file_ids = []
+    for match in re.finditer(OCS_CITATION_PATTERN, text):
+        file_ids.append(match.group("file_id"))
+    return file_ids
+
+
+def remove_citation_tags(text: str) -> str:
```
snopoke, line 28: Wouldn't it be better to replace citations with a markdown link and footer like we do with assistants?

snopoke (approved):

PR #1789, opened 2025-06-27
Extract function

snopoke:
## Description
Extract utility function for future reuse (upcoming PR).


apps/experiments/views/experiment.py
```diff
@@ -809,9 +777,8 @@ def get_message_response(request, team_slug: str, experiment_id: uuid.UUID, sess
             "session": session,
             "task_id": task_id,
             "message_details": message_details,
-            "skip_render": skip_render,
```
snopoke, line 812: This is not used in any templates

PR #1793, opened 2025-06-27
Enable changing is_index field for existing collections

snopoke:
## Summary
- Remove restrictions preventing changes to the `is_index` field in existing collections
- Add proper data management for both indexing transitions (True↔False)
- Implement user-friendly warnings explaining the consequences of each change
- Maintain data integrity with atomic transactions

### Form Updates (`apps/documents/forms.py`)
- Removed the disabled attribute on `is_index` field for existing collections
- Simplified form validation to allow `is_index` changes
- Preserved other restrictions (e.g., `is_remote_index` changes for indexed collections)

### View Logic (`apps/documents/views.py`)
- Enhanced `EditCollection.form_valid()` to detect and handle `is_index` field changes
- **True → False transition**: Deletes all indexed data (embeddings, vector stores) and resets file statuses
- **False → True transition**: Sets up indexing with default chunking strategy and triggers background processing
- Added helper methods `_delete_indexed_data()` and `_setup_indexing()` for clean separation of concerns
- Proper transaction handling to ensure data consistency

### UI Improvements (`templates/documents/collection_form.html`)
- Added contextual warning alerts that appear when users change the `is_index` setting
- Different warning content for enabling vs disabling indexing
- Clear explanations of data impact and processing expectations
- Uses Alpine.js for dynamic show/hide behavior based on form state

## Test Plan
- [x] Verify form allows changing `is_index` field for existing collections
- [x] Test True → False transition properly deletes indexed data without breaking the collection
- [x] Test False → True transition sets up indexing and triggers background tasks
- [x] Confirm warning messages appear and disappear appropriately when toggling the field
- [x] Verify existing functionality (LLM provider changes) still works correctly
- [x] Run existing test suite to ensure no regressions

### DIFF
--- a/apps/documents/views.py
+++ b/apps/documents/views.py
@@ -237,24 +237,74 @@ class EditCollection(LoginAndTeamRequiredMixin, CollectionFormMixin, UpdateView,
         return reverse("documents:single_collection_home", args=[self.request.team.slug, self.object.id])
 
     def form_valid(self, form):
-        response = super().form_valid(form)
+        with transaction.atomic():
+            collection = form.instance
+            old_is_index = form.initial.get("is_index", False)
+            new_is_index = form.cleaned_data.get("is_index", False)
+            old_vector_store_id = collection.openai_vector_store_id
 
-        collection = form.instance
-        old_vector_store_id = collection.openai_vector_store_id
+            # Handle is_index field changes
+            if "is_index" in form.changed_data:
+                if old_is_index and not new_is_index:
+                    # Changing from indexed to non-indexed - delete indexed data
+                    self._delete_indexed_data(collection)
+                    messages.success(self.request, "Collection index data has been deleted")
+                elif not old_is_index and new_is_index:
+                    # Changing from non-indexed to indexed - index files
+                    self._setup_indexing(collection, form)
+                    messages.success(
+                        self.request, "Collection indexing has been enabled. Files will be indexed in the background."
+                    )
 
-        if form.instance.is_index and form.instance.is_remote_index and "llm_provider" in form.changed_data:
-            with transaction.atomic():
+            response = super().form_valid(form)
+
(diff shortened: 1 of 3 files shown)


apps/documents/views.py
```diff
@@ -237,25 +237,75 @@ def get_success_url(self):
...
+            collection.openai_vector_store_id = ""
+            collection.save(update_fields=["openai_vector_store_id"])
+        else:
+            # Delete local index data
+            from apps.files.models import FileChunkEmbedding
+
+            FileChunkEmbedding.objects.filter(collection_id=collection.id).delete()
+
+        # Reset collection file statuses
+        CollectionFile.objects.filter(collection_id=collection.id).update(status="")
```
SmittieC, line 290: I'm thinking this should be set to "Completed"?

apps/documents/views.py
```diff
@@ -237,25 +237,75 @@ def get_success_url(self):
...
+            collection.ensure_remote_index_created()
+
+        # Set up collection files for indexing
+        collection_files = CollectionFile.objects.filter(collection_id=collection.id)
+        if collection_files.exists():
+            # Update metadata and status for existing files
+            chunking_strategy = ChunkingStrategy(chunk_size=800, chunk_overlap=400)
+            metadata = CollectionFileMetadata(chunking_strategy=chunking_strategy)
+
+            collection_files.update(status=FileStatus.PENDING, metadata=metadata)
```
SmittieC, line 304: There might be a mismatch between the set of file types in the non-indexed collection and the index's supported file types. I'm thinking we should filter for only supported file types and only try to index those. We should tell the user about this in the notice we show them when they want to do the switch.

Q: What should happen with the rest?
A: We can either remove them, or leave them with an empty status. Or add a new "unsupported" status. Maybe this makes more sense, because then the user can hover over these files and see why they cannot be indexed.

PR #1796, opened 2025-06-27
Add generic document source integration for Collections

snopoke:
## Summary
Implements extensible system for syncing external document sources with indexed Collections, starting with GitHub repository support and designed for future expansion to Confluence, etc.

## Demo
<div>
    <a href="https://www.loom.com/share/5c4e27c455db4c75b9330bae0381bc5b">
      <p>Document Sources in OpenChess Studio: A Quick Demo - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/5c4e27c455db4c75b9330bae0381bc5b">
      
    </a>
  </div>

## Future TODOs
- [ ] Better display for syncing error status
- [ ] Display sync history
- [ ] Support for other document sources


snopoke:
@coderabbitai review

apps/documents/views.py
```diff
@@ -119,6 +165,127 @@ def query_collection(request, team_slug: str, pk: int):
...
+    permission_required = "documents.change_documentsource"
+
+    def get_queryset(self):
+        return DocumentSource.objects.filter(team=self.request.team)
+
+    @property
+    def source_type(self):
+        return self.object.source_type
+
+    def get_context_data(self, **kwargs):
```
SmittieC, line 250: You can remove this method

apps/documents/views.py
```diff
@@ -119,6 +165,127 @@ def query_collection(request, team_slug: str, pk: int):
...
+        self.object = form.save()
+        task = sync_document_source_task.delay(self.object.id)
+        self.object.sync_task_id = task.task_id
+        self.object.save(update_fields=["sync_task_id"])
+        return HttpResponse(headers={"HX-Redirect": self.get_success_url()})
+
+
+class CreateDocumentSource(BaseDocumentSourceView, CreateView):
+    permission_required = "documents.add_documentsource"
+    template_name = "documents/document_source_form_dialog.html"
```
SmittieC, line 225: nit: the base class already has this as a template

apps/documents/tasks.py
```diff
@@ -146,3 +157,116 @@ def create_collection_from_assistant_task(collection_id: int, assistant_id: int)
...
+
+
+@shared_task(ignore_result=True)
+def sync_document_source_task(document_source_id: int):
+    """Sync a specific document source"""
+    from apps.documents.document_source_service import sync_document_source
+
+    try:
+        document_source = DocumentSource.objects.get(id=document_source_id)
+    except DocumentSource.DoesNotExist:
```
SmittieC, line 169: I'm wondering if we shouldn't let it fail here if the document source doesn't exist. Reason being that `sync_document_source_task` is called either manualy or when one is created. In both scenarios, we expect this to exist, so if it doesn't there's something wrong that we should probably know about. Thougths?

apps/teams/backends.py
```diff
@@ -89,7 +89,7 @@ def _get_group_permissions(self, user_obj):
     "teams": ["invitation", "membership", "team"],
     "annotations": ["tag", "customtaggeditem", "usercomment"],
     "participants": [],
-    "documents": ["collection"],
+    "documents": ["collection", "documentsource"],
```
SmittieC, line 92: `DocumentSourceSyncLog` is missing?

apps/documents/tasks.py
SmittieC: Nit: Pre-load the collection, since it's needed in the service later on

SmittieC:
Generally looks good to me. This is exciting!

apps/documents/models.py
```diff
@@ -166,15 +221,29 @@ def create_new_version(self, save=True):
...
+                file_version.save()
+                file_versions[file.id] = file_version.id
+            new_object_version.files.add(*list(file_versions.values()), through_defaults=through_defaults)
 
-        new_version.files.add(*list(file_versions.values()))
+        file_versions: dict[int, int] = {}
+        _version_files(self.files.filter(collectionfile__document_source=None), new_version)
+
+        for document_source in self.document_sources.all():
+            doc_source_version = document_source.create_new_version(save=False)
```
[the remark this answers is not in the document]
snopoke: Yes it does since changes can affect what documents are present in the index. The document source is really just the config.

apps/documents/tasks.py
snopoke: > hs?

If we got an error from this in Sentry I don't know what we would do. Likely ignore it or track it down to some race condition. I think it's safe to ignore.

apps/assistants/sync.py
```diff
@@ -420,7 +420,7 @@ def remove_files_from_tool(ocs_resource: ToolResources, files: list[int]):
         if file in file_references:
             if ocs_resource.extra.get("vector_store_id") and file.external_id:
                 index_manager = OpenAIRemoteIndexManager(client, index_id=ocs_resource.extra.get("vector_store_id"))
-                index_manager.pluck_file_from_index(file_id=file.external_id)
+                index_manager.delete_file_from_index(file_id=file.external_id)
```
snopoke, line 423: @SmittieC I renamed this back to `delete_` since we don't use `pluck` anywhere else and I'm not sure what the difference is.

apps/documents/tasks.py
SmittieC: That's fair

PR #1805, opened 2025-06-30
Resolve N+1 queries in session table 

### LINKED ISSUE #1779: N+1 queries in session table (tags column and versions column)
The tags column and versions column in the chatbot sessions table both do additional queries per row to fetch the tags.

stephherbers:
## Description

resolves #1779

because of how the page is loaded it, I couldn't see the different in queried via the django debug toolbar so rather I compared in the shell locally
running through the shell I saw a difference of 198 -> 4 queries for one session

## User Impact

n/a

### Demo

n/a

### Docs and Changelog

n/a


snopoke:
There is one other place which also needs adjusting which is the `tags_json` call in the `tag_ui.html` template: https://github.com/dimagi/open-chat-studio/blob/680ecd7106bc63cc93b762af9bceee2ded47d930/templates/annotations/tag_ui.html#L4

I does it's own tag query: https://github.com/dimagi/open-chat-studio/blob/2aa5ee93c8e2598b5ebc3d471d9545acdca8fa75/apps/annotations/models.py#L126-L128

apps/annotations/models.py
```diff
@@ -113,28 +113,27 @@ def add_tag(self, tag: Tag, team: Team, added_by: CustomUser = None):
...
-        return [tag["name"] for tag in self.tags_json]
+        return [tag["name"] for tag in self.prefetched_tags_json]
 
     @cached_property
-    def tags_json(self):
-        tagged_items = CustomTaggedItem.objects.filter(
-            content_type__model=self._meta.model_name, content_type__app_label=self._meta.app_label, object_id=self.id
-        ).prefetch_related("tag", "user")
+    def prefetched_tags_json(self):
+        if not hasattr(self, "prefetched_tagged_items"):
```
snopoke, line 126: let's add a warning log here to make it more visible if the prefetch hasn't been done.

snopoke (approved):

PR #1807, opened 2025-07-01
Ba/scheduled messages attempts

### LINKED ISSUE #1744: Track schedule message sending attempts
Create a new DB table to track ScheduleMessage sending attempts and the results of those attempts.

This data can then be used to limit retries and perform backoff.

The data is also useful for auditing and debugging so it should be exposed in the UI on the Participant data page (maybe a collapsed section under each schedule).

The DB model should track:

* the date of the attempt
* the trigger number (e.g. 2nd trigger of a daily schedule)
* the attempt number (e.g. 3rd attempt for this trigger)
* the attempt result (success / failure)
* a log message in case of an error
* a trace info JSON blob if tracing is enabled (see ChatMessage.trace_info)

## Limit retries
A message should only be retried 3 times with exponential backoff.

barora05:
## Description
Add schedules messages attempts table for saving scheduled messages send attempts and add max retry of 3 with exponential backoff. Closes #1744


apps/events/models.py
```diff
@@ -416,27 +416,45 @@ def generate_external_id(name: str, experiment_id: int, participant_id: int, ins
...
         except Exception as e:
             logger.exception(f"An error occurred while trying to send scheduled message {self.id}. Error: {e}")
+            trace_metadata = getattr(e, "trace_metadata", {})
+            attempt.attempt_result = EventLogStatusChoices.FAILURE
+            attempt.log_message = str(e)
+            attempt.trace_info = trace_metadata
+            attempt.save()
+            if attempt_number < 3:
+                backoff_seconds = 2 ** (attempt_number - 1)
+                retry_scheduled_message.apply_async(args=[self.id, attempt_number + 1], countdown=backoff_seconds)
```
SmittieC, line 443: I'm wondering if we need to prevent the `poll_scheduled_messages` task from picking up this message if it is being throttled. `poll_scheduled_messages` runs every 60 seconds, and the retry stuff should finish in that time, but it feels more robust to have this check. What are your thoughts?

barora05:
@snopoke 
Current UI
ui2
image

UI generated by claude (if we want a fancy UI)
Screenshot 2025-07-03 114449

apps/events/models.py
barora05: I agree, we need this check

snopoke: I don't think it's necessary since after the first attempt the `last_triggered_at` will be updated and so `poll_scheduled_messages` won't re-trigger it (this is how it works now)

snopoke: In fact, now that I say that it seems like this retry mechanism won't work at all because `_trigger` won't raise an exception since `fail_silently=True`

snopoke:
@barora05 The simple UI looks fine to me

> Also, I have not been able to test it without unit tests on my local. I have poll_scheduled_message task enabled in my local db but still that doesn't run automatically

Are you running celery with beat?: `inv celery` or `celery -A gpt_playground worker -l INFO -B`?

apps/events/models.py
```diff
@@ -416,27 +416,45 @@ def generate_external_id(name: str, experiment_id: int, participant_id: int, ins
...
+            attempt.save()
+
         except Exception as e:
             logger.exception(f"An error occurred while trying to send scheduled message {self.id}. Error: {e}")
+            trace_metadata = getattr(e, "trace_metadata", {})
+            attempt.attempt_result = EventLogStatusChoices.FAILURE
+            attempt.log_message = str(e)
+            attempt.trace_info = trace_metadata
+            attempt.save()
+            if attempt_number < 3:
```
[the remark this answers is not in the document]
SmittieC: Yes you're right. We don't need to reschedule a fourth

apps/events/models.py
SmittieC: Ah yes you're right. That silent fail should not be silent

barora05: In case of fail_silently=False, if we don't handle the exception in _trigger last_updated_now would not be updated and polled_scheduled_messages can pick it up again

SmittieC: Maybe something like this should solve it?

```python
    def _trigger(self):
        experiment_session = self.participant.get_latest_session(experiment=self.experiment)
        if not experiment_session:
            # Schedules probably created by the API
            return
        trace_info = TraceInfo(
            name="scheduled message",
            metadata={
                "schedule_id": self.external_id,
                "trigger_number": self.total_triggers,
            },
        )
        try:
            trace_metadata = experiment_session.ad_hoc_bot_message(
                self.params["prompt_text"],
                trace_info,
                fail_silently=False,
                use_experiment=self._get_experiment_to_generate_response(),
            )
        finally:
            utc_now = timezone.now()
            self.last_triggered_at = utc_now
            self.total_triggers += 1

            if self._should_mark_complete():
                self.is_complete = True
            else:
                delta = relativedelta(**{self.params["time_period"]: self.params["frequency"]})
                self.next_trigger_date = utc_now + delta

            self.save()

        return trace_metadata or {}
```


I think it's fine to mark it as completed, even though it failed, since the retry logic will try to send it again.

apps/events/models.py
```diff
@@ -416,43 +416,67 @@ def generate_external_id(name: str, experiment_id: int, participant_id: int, ins
...
+            trace_metadata = experiment_session.ad_hoc_bot_message(
+                self.params["prompt_text"],
+                trace_info,
+                fail_silently=False,
+                use_experiment=self._get_experiment_to_generate_response(),
+            )
+        except Exception as e:
+            trace_metadata = getattr(e, "trace_metadata", {})
+            e.trace_metadata = trace_metadata
+            raise
```
snopoke, line 468: This can be removed since the exception already has the `trace_metadata` attribute set in `ad_hoc_bot_message`

apps/experiments/models.py
```diff
@@ -1700,17 +1702,36 @@ def ad_hoc_bot_message(
...
+                    session_id=str(self.external_id),
+                    user_id=str(self.participant.identifier),
+                    inputs={"input": instruction_prompt},
+                    metadata=trace_info.metadata,
+                ):
+                    bot_message = self._bot_prompt_for_user(
+                        instruction_prompt, trace_info, use_experiment=use_experiment, trace_service=trace_service
+                    )
+                    self.try_send_message(message=bot_message)
+                    raise RuntimeError("Simulated bot failure!")
```
snopoke, line 1718:
```suggestion
```

apps/experiments/models.py
```diff
@@ -1700,17 +1702,36 @@ def ad_hoc_bot_message(
...
+                    trace_service.set_current_span_outputs({"response": bot_message})
+                    trace_info = trace_service.get_trace_metadata()
+                return trace_info
         except Exception as e:
             log.exception(f"Could not send message to experiment session {self.id}. Reason: {e}")
             if not fail_silently:
```
snopoke, line 1724: I don't think there are any places that call this method with `fail_silently = True` so we can remove that param

snopoke (commented):

apps/experiments/models.py
barora05: Oops!!!, forgot to remove it

PR #1821, opened 2025-07-01
Improve UI when creating collections

### LINKED ISSUE #1776: Add more in product help to collections
To understand collections currently one really needs to read the docs. We should have some basic help info in OCS directly:

1. What is a media collection vs an indexed collection
2. Remove vs local
3. How to hook up a collection to a bot

SmittieC:
## Description

Fixes #1776

(I brainstormed with claude, so I used some of its code)

## User Impact

This should help users understand the differences between normal and indexed collectiosn more, as well as when to use which. 

### Demo

demo

### Docs and Changelog

I don't think this warrents a changelog update


snopoke (approved):
🥇

Issue #1824, opened 2025-07-02
[chat widget] Create an API for embedding chat widget


snopoke:
Using an API for the external chat widget will allow the widget to have complete control over the UI.

PR #1831, opened 2025-07-02
chat API and initial chat widget updates

snopoke:
## Description
Resolves https://github.com/dimagi/open-chat-studio/issues/1824

A new API which can be used to integrate OCS bots into external sites. This PR also updates the chat component to use the new API instead of embedding the existing chat UI in a an iframe.

The new API has the following endpoints:

- `POST /api/chat/start/` - Start new chat session
- `POST /api/chat/{session_id}/message/` - Send messages
- `GET /api/chat/{session_id}/{task_id}/poll/` - Poll for responses
- `GET /api/chat/{session_id}/poll/` - Poll for new messages

All endpoints can be accessed without auth allow CORS access (new `django-cors-headers` dependency added).

## User Impact
More flexible chat embedding options.

### Demo
<div>
    <a href="https://www.loom.com/share/f2bc27101c1e479885d2b857515dba6d">
      <p>Stencil Component Starter - 2 July 2025 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/f2bc27101c1e479885d2b857515dba6d">
      
    </a>
  </div>

### Docs and Changelog
TODO


apps/api/views/chat.py
```diff
@@ -0,0 +1,339 @@
...
+        user = request.user
+    else:
+        team = experiment.team
+        user = None
+
+    # Create or get participant
+    if user is not None:
+        participant_id = participant_id or user.email
+        if participant_id != user.email:
+            # TODO: re-evaluate this, it doesn't seem correct in this instance
```
stephherbers, line 112: wouldn't this cover the cases we are trying to catch?

> if user is not None:
>     if participant_id and participant_id != user.email:
>         return Response({"error": "Participant ID must match your email...)
>     participant_id = participant_id or user.email

snopoke: The current logic works but it doesn't feel quite correct for an API. It matches what we do for web sessions but if another systems uses this to integrate chat into their app then they should be able to specify the participant ID (I think)

apps/api/views/chat.py
```diff
@@ -0,0 +1,339 @@
...
+    data = serializer.validated_data
+    experiment_id = data["chatbot_id"]
+    participant_id = data.get("participant_id")
+    session_data = data.get("session_data", {})
+
+    # First, check if this is a public experiment
+    experiment = get_object_or_404(Experiment, public_id=experiment_id)
+    if not experiment.is_working_version:
+        return Response(
+            {"error": "Chatbot ID must reference the working version of an chatbot"},
```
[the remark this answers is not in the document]
snopoke:
```suggestion
            {"error": "Chatbot ID must reference the unreleased version of an chatbot"},
```

Issue #1665, opened 2025-05-29
Filter out pipeline experiments from 'experiments list'


snopoke:
Don't show Chatbots in the Experiments list. Also add a banner / update the banner text informing users that 'pipeline experiments have been moved to Chatbots' with a link.

PR #1832, opened 2025-07-02
Filter pipeline experiments from table

stephherbers:
## Description

Resolves [#1665](https://github.com/dimagi/open-chat-studio/issues/1665)

NOTE: do not merge until approval to move forward

## User Impact

users with chatbots flag will not longer see pipeline experiments from the experiments table

### Demo

n/a

### Docs and Changelog

changelog


SmittieC (approved):

snopoke:
I think we can move forward

PR #1844, opened 2025-07-04
add fix to run finally block everytime

barora05:
## Description
Add complete _trigger inside try block so the finally block gets exeucted everytime and the trigger number and lasst_updated_at gets updated.


snopoke:
You'll have to revert this commit before applying these changes: https://github.com/dimagi/open-chat-studio/pull/1843/commits/4a790d0125cf3c8628f7335298fe9c4efffed3f4

snopoke (approved):

PR #1846, opened 2025-07-04
don't uppercase connect identifiers

snopoke:
## Description
CommCare Connect doesn't require uppercase usernames anymore


SmittieC (dismissed):
Nice

Issue #1597, opened 2025-05-16
Pipeline draging issue


SmittieC:
### Reproduce

1. In a pipeline, add a node
2. Click the output circle of the node and start dragging

Notice that it does a weird drag thing. This did not happen before

PR #1848, opened 2025-07-04
Fix pipeline drag issue + read only state

snopoke:
## Description
Fixes https://github.com/dimagi/open-chat-studio/issues/1597

This PR fixes the pipeline drag issue (first commit) and also another bug which allowed editing of pipeline versions via the codemirror editors (LLM and PythonNode).

In addition I've set the 'readonly' / 'disabled' state on inputs when in readonly mode. Although edits are already blocked at the Zustand store level, having the inputs disabled / readonly prevents changes from getting to the store in the first place and creates a clearer interface for the user. 

## User Impact
Fix edge drag issue in pipelines

### Docs and Changelog
TODO


SmittieC:
Here's the changelog entry: https://github.com/dimagi/open-chat-studio-docs/pull/126

PR #1849, opened 2025-07-07
Semantic Search for your index

SmittieC:
## Description

Since it's a quick win and rough draft. I thought this might come in handy, but am happy to discard it if we think otherwise. (See the demo)

## User Impact

User can see which chunks gets returned for a given query

### Demo

demo


SmittieC:
cc @snopoke @bderenzi @stephherbers thoughts welcome

snopoke:
> cc @snopoke @bderenzi @stephherbers thoughts welcome

I think this is useful

Issue #1833, opened 2025-07-03
Allow translating messages from chat transcript view


snopoke:
At the moment, the only way to translate messages to begin with is via the session analysis feature.

Add a button to the chat transcript view:

* opens a dialog when clicked
* allows the user to select a language and model for translation
* kicks off a backend task to perform the translation
* displays a progress bar to the user
* reloads the chat messages once the process completes

PR #1852, opened 2025-07-07
Translate from transcript

stephherbers:
## Description

resolves https://github.com/dimagi/open-chat-studio/issues/1833

Tested with:
- Brand new translation with "default" language selected
- Brand new translation with different complete translated language selected 
- Brand new translation with incomplete translated language selected 
- Previous translation behaviors

## User Impact

users can translate to entirlely new languages right from the transcript page

### Demo

[loom video](https://www.loom.com/share/5c0e325c895d4172ac9490cdbde72dd8)

### Docs and Changelog

yes and yes


apps/experiments/forms.py
```diff
@@ -285,16 +285,27 @@ class Meta:
...
     )
 
-    def __init__(self, *args, team=None, **kwargs):
+    def __init__(self, *args, team=None, translatable_languages=None, widget_id_suffix="", **kwargs):
         super().__init__(*args, **kwargs)
         if team:
             self.fields["provider_model"].choices = [
                 ("", "Choose a model for translation")
             ] + get_dropdown_llm_model_choices(team)
+            self.fields["provider_model"].widget.attrs["id"] = f"translation-provider-model{widget_id_suffix}"
```
stephherbers, line 307: needed for when the "translate remaining messages" button is also shown because before this change 2 provider_ids were passed and one was blank. I opted for this route rather than filtering out the blank one to ensure we are selecting the one we want

SmittieC: Thoughts on having a flag param indicating whether or not this form is meant for translating remaining or all messages. It feels like the API would be cleaner and then inside the form you can do form-specific things depending on that value. So I'm thinking something like
```python
def __init__(self, *args, team=None, translatable_languages=None, is_translate_all_form: bool, **kwargs):
```

apps/experiments/forms.py
```diff
@@ -285,16 +285,27 @@ class Meta:
...
+        if translatable_languages:
```
SmittieC, line 308: Should `translatable_languages` be an optional param?

SmittieC (approved):

PR #1853, opened 2025-07-08
Ba/optimize dashboard queries

### LINKED ISSUE #1834: Investigate options for improving dashboard performance
Examine the SQL queries that are performed when loading the dashboard and determine if there are improvements that can be made e.g. new sql indexes.

In particular look at the [get_bot_performance_summary](https://github.com/dimagi/open-chat-studio/blob/main/apps/dashboard/services.py#L114) method which has very poor performance. 

Other than indexes it could be worth investigating the option of creating some analytics data tables to store pre-computed data e.g. 

* date
* team
* experiment
* sessions_created
* sessions_ended
* avg_session_duration
* messages_sent
* participants_active
* participants_created

That could be computed daily and then used to build the aggregate metrics for the dashboard.

barora05:
## Description
Optimize dashboard queries. Closes #1834 
I have created some indexes after analyzing the queries which were taking more time than others.
I have reduced the number of db calls from a couple of funcitons which were happening inside loop or N+1 times.
 On my local, botperformaceview was taking 1300ms initially which later reduced to less than 200ms.
 
 If we store pre-computed data, then for the hourly view we would still need to make the db callls and for the daily view, some data would reflect the previous day while other charts (like the bot performance chart) might show data for the current day. This could lead to inconsistencies in what the user sees.


apps/dashboard/services.py
```diff
@@ -326,38 +333,56 @@ def get_channel_breakdown_data(self, **filters) -> dict[str, Any]:
...
         )
-        for platform in platforms_in_use:
-            platform_sessions = querysets["sessions"].filter(experiment_channel__platform=platform)
-            platform_messages = querysets["messages"].filter(
-                chat__experiment_session__experiment_channel__platform=platform
+        sessions_stats = (
+            querysets["sessions"]
+            .values("experiment_channel__platform")
+            .annotate(sessions_count=Count("id"), participants_count=Count("participant", distinct=True))
+        )
```
snopoke, line 346: Because of how the `querysets["sessions"]` is being created this actually produces results grouped by date which isn't what you want here. One trick you can do is use `order_by()` to clear the distinct fields and then add `distinct=True` to the count:

```suggestion
        sessions_stats = (
            querysets["sessions"].order_by()
            .values("experiment_channel__platform")
            .annotate(sessions_count=Count("id", distinct=True), participants_count=Count("participant", distinct=True))
        )
```

snopoke (approved):

PR #1858, opened 2025-07-09
add tags to trace

### LINKED ISSUE #1570: Add message tags to langfuse traces
Tags saved to messages should also be added to the traces in langfuse

barora05:
## Description
Closes #1570
### Demo
image

### DIFF
--- a/apps/chat/bots.py
+++ b/apps/chat/bots.py
@@ -377,7 +377,7 @@ class PipelineBot:
             version_number=self.experiment.version_number, is_a_version=self.experiment.is_a_version
         )
         if self.trace_service and (tags := output.get("output_message_tags")):
-            flat_tags = [f"{tag}:{category}" for tag, category in tags]
+            flat_tags = [f"{category}:{tag}" for tag, category in tags]
             self.trace_service.add_output_message_tags_to_trace(flat_tags)
 
         if session_tags := output.get("session_tags"):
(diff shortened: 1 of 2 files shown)


apps/service_providers/tracing/service.py
```diff
@@ -265,3 +265,11 @@ def _get_current_span_info(self) -> tuple[UUID, str]:
         if self.span_stack:
             return self.span_stack[-1]
         return self.trace_id, self.trace_name
+
+    def add_output_message_tags_to_trace(self, tags: list[str]) -> None:
+        if not self.activated or not tags:
+            return
+        span_id, _ = self._get_current_span_info()
+        self.outputs[span_id].setdefault("tags", [])
```
SmittieC, line 273: I think this line can be removed, since the one beneath it does the same?

barora05: yes

apps/service_providers/tracing/service.py
```diff
@@ -265,3 +265,10 @@ def _get_current_span_info(self) -> tuple[UUID, str]:
         if self.span_stack:
             return self.span_stack[-1]
         return self.trace_id, self.trace_name
+
+    def add_output_message_tags_to_trace(self, tags: list[str]) -> None:
```
snopoke, line 269: These should be added to the langfuse trace directly: https://python.reference.langfuse.com/langfuse/api#Trace.tags

So this needs to call each tracer in `self._active_tracers` and let them handle the tags

snopoke (changes requested):

snopoke (approved):

PR #1862, opened 2025-07-10
[do not merge] temporary evals script

snopoke:
## Description
A script to run evals which we can use temporarily until OCS can do the evals directly.


snopoke:
> This is pretty sweet. Is the plan to make this a standalone python based program like the commcare DET? it's quite well suited for that I think

We're building this into OCS evals framework. I just did this to have something to use before that's available.

PR #1863, opened 2025-07-11
User controlled reference generation

### LINKED ISSUE #1856: Allow bot builders to disable citations
When using collections, citations are included in the output text by default. Bot builders should be able to disable this feature.

It is unclear to me whether this should be a global settings or a per node setting. Also whether we should apply this setting at render time or generation time.

@SmittieC FYI

### Relevant code

* Added to message text: [populate_reference_section_from_citations](https://github.com/dimagi/open-chat-studio/blob/b4f1f8747c16442a010f17a237bb4c565309b365/apps/service_providers/llm_service/utils.py#L31)
* Formatted for output: [ChannelBase._format_reference_section](https://github.com/dimagi/open-chat-studio/blob/64f6faecc5fcc43fafd54db52c65c14cb54ffca9/apps/chat/channels.py#L538)
* Displayed in web UI [ai_message.html](https://github.com/dimagi/open-chat-studio/blob/afe1168c19d2ec468ab7a46b0fdf8a26b3bc0c08/templates/experiments/chat/ai_message.html#L22-L32)

SmittieC:
## Description

Resolves #1856 

- Disabling citations / downloads happens on a collection level. You might have one collection with sensitive data and the other with non-sensitive data
- When citation generation is disabled, the LLM is not being instructed to generate citations
- I added a sanitization step to clean out any citation tags that might have been generated. This is mainly to prevent malicious users from asking the bot to generate citations (OCS is open source) and get their hands on the file. Probably not a big risk even if this happens, but I'll rather play safe.

## User Impact

Users can disable file references

### Demo

demo

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/128


SmittieC:
@claude please review the PR

snopoke (commented):
The changes look good to me. I'm a bit unsure about the user requirements and whether this configuration should be at the Collection level or at the pipeline node level.

snopoke:
> collection level toggle would be next best thing

This can be accomplished either the way you have it now (at the collection) or in the pipeline node. In my mind, having it configured at the node is a bit more flexible since it allows using the same collection elsewhere without the flag. Some other thoughts:

* if the config is at the node level then we don't need to load the collection object during the pipeline run
* if 'debug mode' is on for a chatbot, it might be useful to still show the citations (but perhaps that's what tracing is for)

PR #1864, opened 2025-07-11
Poll file status on collections home

### LINKED ISSUE #1722: RAG - Async checks for UI
The file statuses are static at the moment when viewing a specific collection. It will help a lot if users can see what's going on as it happens without having to reload the page.

SmittieC:
## Description

Resolves #1722 

A progress bar would have been nice, but it's hard to estimate the amount of chunks. (or rather, **way** more effort).

## User Impact

Users will not have to reload to see updates statusus

### Demo

demo

### Docs and Changelog

Pending

### DIFF
--- a/templates/documents/collection_file_status.html
+++ b/templates/documents/collection_file_status.html
@@ -3,7 +3,7 @@
        class="tooltip"
        data-tip="{{ collection_file.status_enum.label }}"
        {% if collection_file.status == 'pending' or collection_file.status == 'in_progress' %}
-         hx-get="{% url 'documents:get_collection_file_status' team.slug collection_id collection_file.id %}"
+         hx-get="{% url 'documents:get_collection_file_status' team.slug collection.id collection_file.id %}"
          hx-trigger="every 5s"
          hx-swap="outerHTML"
        {% endif %}>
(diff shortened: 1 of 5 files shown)


templates/documents/collection_file_status.html
```diff
@@ -0,0 +1,21 @@
...
+     {% if collection_file.status == 'pending' or collection_file.status == 'in_progress' %}
+       hx-get="{% url 'documents:get_collection_file_status' team.slug collection.id collection_file.id %}"
+       hx-trigger="every 5s"
+       hx-swap="outerHTML"
+     {% endif %}>
+  {% if collection_file.status == 'pending' %}
+    <span class="status"></span>
+  {% elif collection_file.status == 'in_progress' %}
+    <div class="inline-grid *:[grid-area:1/1]">
+      <div aria-label="warning" class="status status-warning animate-ping"></div>
```
stephherbers, line 13: nit: I think the ping looks great, but we do use the spinners elsewhere so maybe opt for a spinner?

Also aria label of "in progress" would be more accurate

SmittieC: We can do that yes. here's a mini demo of what that looks like

demo

snopoke (approved):

SmittieC:
I forgot to link the changelog. [Here](https://github.com/dimagi/open-chat-studio-docs/pull/132) it is

PR #1865, opened 2025-07-11
Ba/filter by participants and tags

### LINKED ISSUE #1850: Dashboard Improvements: filter by tag and participant
As a user I want to be able to filter by tag and I also want to separately be able to filter by participant 
(both should be multi-select)

barora05:
## Description
Closes #1850 

### Demo
[Demo](https://drive.google.com/file/d/1odhyIq9oR72MLeeN6iQ7M_oeNYlfGvyV/view?usp=drive_link)


apps/dashboard/services.py
```diff
@@ -73,6 +76,18 @@ def get_filtered_queryset_base(
             sessions = sessions.filter(experiment_channel__platform__in=platform_names)
             messages = messages.filter(chat__experiment_session__experiment_channel__platform__in=platform_names)
 
+        if participant_ids:
+            participants = participants.filter(id__in=participant_ids)
+            sessions = sessions.filter(participant__id__in=participant_ids)
+            messages = messages.filter(chat__experiment_session__participant__id__in=participant_ids)
+
+        if tag_ids:
+            tags = [int(t) for t in tag_ids]
```
snopoke, line 85: I think once https://github.com/dimagi/open-chat-studio/pull/1870#pullrequestreview-3016765431 is resolved this may not be required at all (and if it is, it should be done in the `DashboardFilterForm.clean_tags` method.

apps/dashboard/forms.py
```diff
@@ -64,6 +65,10 @@ class DashboardFilterForm(forms.Form):
         widget=forms.SelectMultiple(),
     )
 
+    participants = forms.MultipleChoiceField(choices=[], required=False, widget=forms.SelectMultiple)
+
+    tags = forms.MultipleChoiceField(choices=[], required=False, widget=forms.SelectMultiple())
```
snopoke, line 70: You can make these `ModelMultipleChoiceField` and follow the same pattern as experiments.

apps/dashboard/forms.py
```diff
@@ -82,6 +87,13 @@ def __init__(self, *args, team=None, **kwargs):
                 (label_to_value[label], label) for label in available_platform_labels if label in label_to_value
             ]
             self.fields["channels"].choices = platform_choices
+            all_participants = Participant.objects.filter(team=team).values("id", "identifier")
+            self.fields["participants"].choices = [(p["id"], p["identifier"]) for p in all_participants]
```
snopoke, line 91:
```suggestion
            self.fields["participants"].queryset = Participant.objects.filter(team=team).select_related("user").order_by("identifier")
```

(similar for tags - you can override `__str__` on the Tag model if necessary)

apps/dashboard/forms.py
```diff
@@ -82,6 +87,13 @@ def __init__(self, *args, team=None, **kwargs):
...
+
+            tags = Tag.objects.filter(team=team).exclude(category=TagCategories.EXPERIMENT_VERSION)
+            self.fields["tags"].choices = [
+                (t.id, t.name.split(":", 1)[1].strip() if ":" in t.name else t.name) for t in tags
+            ]
```
snopoke, line 96: Why do we need to split tag name?

snopoke: yes, I think so

apps/dashboard/forms.py
barora05: The options list works fine without overriding the __str__ method. The names are shown on the UI and the ID is sent in the filter params.

apps/dashboard/forms.py
snopoke: @barora05 This still needs to be addressed

snopoke (approved):

PR #1869, opened 2025-07-14
New dev docs page: Common Practises

SmittieC:
I'd like some thoughts / input on 
1. having a page like this
3. having a dedicated page as apposed to putting this in the CLAUDE.md file?


docs/developer_guides/common_practises.md
```diff
@@ -0,0 +1,49 @@
+# Standard Practices
```
stephherbers, line 1:
```suggestion
# Common Practices
```

nit: just to keep consistency

PR #1870, opened 2025-07-14
parse multiselect correctly

### LINKED ISSUE #1868: Issue in dashbaord filters
Selecting multiple value in channels or chabbots removes filters and displays the data for all chatbots.

https://github.com/user-attachments/assets/c2c04f87-dde7-43c3-abf6-98a9976bed6d

barora05:
## Description
Multiple Select is not working because the params are sent like this experiments = ['517,617']. Closes #1868

### DIFF
--- a/assets/javascript/dashboard/main.js
+++ b/assets/javascript/dashboard/main.js
@@ -267,7 +267,7 @@ function dashboard() {
             }
             
             const sanitizedParams = this.sanitizeParams({...this.filters, ...params});
-            const urlParams = new URLSearchParams(sanitizedParams);
+            const urlParams = new URLSearchParams();
 
             for (const [key, value] of Object.entries(sanitizedParams)) {
                 if (Array.isArray(value)) {


snopoke (commented):
I actually think a better approach to this would be pass the params as normal multi-params instead of as JSON lists. We already have code to do this here: https://github.com/dimagi/open-chat-studio/blob/5142d287ff40599c1367ec9e512976f0638d7825/assets/javascript/dashboard/main.js#L207-L223

This will result in multiple params e.g. `tags=1&tags=3&tags=35`. Django should process those correctly by default.

assets/javascript/dashboard/main.js
```diff
@@ -268,6 +268,15 @@ function dashboard() {
             
             const sanitizedParams = this.sanitizeParams({...this.filters, ...params});
             const urlParams = new URLSearchParams(sanitizedParams);
```
snopoke, line 270: don't initialise the params with the values:

```suggestion
            const urlParams = new URLSearchParams();
```

snopoke (approved):

PR #1886, opened 2025-07-15
Ba/average response time

### LINKED ISSUE #1851: Dashboard Improvements: message response time
As a user I want to be able to see the average message response times form the dashboard

barora05:
## Description
Closes #1851

### DIFF
--- a/apps/chat/tests/test_routing.py
+++ b/apps/chat/tests/test_routing.py
@@ -46,7 +46,7 @@ def test_experiment_routing(with_default, routing_response, expected_tag):
 def test_experiment_routing_with_tracing():
     experiment = _make_experiment_with_routing()
     session = ExperimentSessionFactory(experiment=experiment)
-    trace_service = TracingService([MockTracer()], 1, 1)
+    trace_service = TracingService([MockTracer()], experiment.id, experiment.team_id)
     with (
         trace_service.trace("test", session, "bob"),
         mock_llm(responses=["anything", "How can I help today?"], token_counts=[0]) as fake_service,


barora05:
> @barora05 thanks for this. This implementation has a few draw backs:
> 
>     1. It only captures the duration of the pipeline run, not anything before / after.
> 
>     2. The function to get the aggregated durations isn't that performant since it has to loop over all the messages.
> 
> 
> What do you think about using this as an opportunity to create a top level 'trace' table which we can initially use to store the duration:
> 
>     * trace
>       
>       * timestamp
>       * experiment_id
>       * session_id
>       * participant_id
>       * input_message_id
>       * output_message_id
>       * duration
> 
> 
> We could insert a row into this table from the top level trace:
> 
> https://github.com/dimagi/open-chat-studio/blob/64f6faecc5fcc43fafd54db52c65c14cb54ffca9/apps/chat/channels.py#L356-L364
> . Maybe even from the `TracingService._end_traces` method?

Yes, I agree — the current approach does have drawbacks, the major one being that the function to calculate the aggregated durations isn’t that performant because it has to loop through all the messages.
I think inserting a row into the new trace table directly from TracingService._end_traces makes sense and would solve this cleanly.

apps/trace/models.py
```diff
@@ -0,0 +1,21 @@
+from django.db import models
+
+from apps.utils.models import BaseModel
+
+
+class Trace(BaseModel):
```
snopoke, line 6: This should have a 'team' field as well. Either it can extend `BaseTeamModel` or just `models.Model` and add `team` and `timestamp`. We don't really need the `updated_at` date and also don't need the `VersioningMixin` that comes with the other base models.

apps/chat/channels.py
```diff
@@ -354,13 +356,17 @@ def new_user_message(self, message: BaseMessage) -> ChatMessage:
 
             try:
                 with self.trace_service.trace(
+                    experiment_id=self.experiment.id,
```
snopoke, line 359: What about moving this to the constructor and setting it when calling `TracingService.create_for_experiment`?

apps/chat/channels.py
```diff
@@ -136,7 +136,9 @@ def __init__(
         self.experiment_channel = experiment_channel
         self._experiment_session = experiment_session
         self._message: BaseMessage = None
-        self._participant_identifier = experiment_session.participant.identifier if experiment_session else None
+        participant = experiment_session.participant
+        self._participant_identifier = participant.identifier if experiment_session else None
+        self._participant_id = participant.id
```
snopoke, line 141: This logic doesn't quite work since `experiment_session` could be None here. I suggest following the same pattern as the `participant_user` property: https://github.com/dimagi/open-chat-studio/blob/64f6faecc5fcc43fafd54db52c65c14cb54ffca9/apps/chat/channels.py#L216-L219

apps/chat/channels.py
```diff
@@ -354,13 +356,17 @@ def new_user_message(self, message: BaseMessage) -> ChatMessage:
 
             try:
                 with self.trace_service.trace(
+                    experiment_id=self.experiment.id,
                     trace_name=self.experiment.name,
                     session_id=str(self.experiment_session.external_id),
                     user_id=self.participant_identifier,
                     inputs={"input": self.message.model_dump()},
+                    participant_id=self._participant_id,
+                    session_id_fk=self.experiment_session.id,
```
snopoke, line 365: Instead of passing this in as a separate field we could change the method to take the whole session object and then it can extract both the `external_id` and the `id` itself.

apps/dashboard/services.py
```diff
@@ -423,6 +424,36 @@ def get_tag_analytics_data(self, **filters) -> dict[str, Any]:
...
+            return cached_data
+
+        querysets = self.get_filtered_queryset_base(**filters)
+        sessions = querysets["sessions"]
+
+        trunc_func = self._get_trunc_function(granularity)
+
+        avg_response_stats = (
+            Trace.objects.filter(session__in=sessions)
+            .annotate(period=trunc_func("session__created_at"))
```
snopoke, line 441: This should use the `Trace.created_at` (or whatever date on the trace object) as the anchor date.

apps/service_providers/tracing/service.py
```diff
@@ -72,11 +80,20 @@ def trace(
...
+        experiment_id: int | None = None,
+        session_id_fk: int | None = None,
+        participant_id: int | None = None,
     ):
         self.trace_id = uuid.uuid4()
         self.trace_name = trace_name
         self.session_id = session_id
         self.user_id = user_id
+        self.experiment_id = experiment_id
+        self._start_time = timezone.now()
```
snopoke, line 93: For timing you should use `time.time()` since we don't actually care about dates and timezones etc.

```python
start = time.time()
end = time.time()
duration_ms = int((end - start) * 1000)
```

apps/trace/models.py
```diff
@@ -0,0 +1,21 @@
...
+        "experiments.Experiment", on_delete=models.SET_NULL, null=True, related_name="traces"
+    )
+    session = models.ForeignKey(
+        "experiments.ExperimentSession", on_delete=models.SET_NULL, null=True, related_name="traces"
+    )
+    participant = models.ForeignKey(
+        "experiments.Participant", on_delete=models.SET_NULL, null=True, related_name="traces"
+    )
+    input_message_id = models.CharField(max_length=255, blank=True)
+    output_message_id = models.CharField(max_length=255, blank=True)
```
snopoke, line 17: shouldn't these be foreign keys?

snopoke (commented):
A few small comments but generally looks good.

apps/chat/channels.py
barora05: didn't do it as traceservice.trace is called from actions.py file also and there we are passing session.id in session_id instead of external id. 
```
with trace_service.trace(
            trace_name=f"{session.experiment.name} - event pipeline execution",
            session_id=session.id,
            user_id=session.participant.identifier,
            inputs={"input": input},
            metadata={"action_type": action.action_type, "action_id": action.id, "params": action.params},
        ):
```
Shouldn't this be consistent at both the places?

snopoke: yes it should, both should use the external ID for that field

snopoke (approved):

PR #1894, opened 2025-07-16
use djlint instead of djhtml

snopoke:
## Description
`djlint` has better options for linting HTML

## User Impact
None

See also https://github.com/dimagi/open-chat-studio/pull/1917


snopoke:
After some more testing I'm not sure this is actually better for formatting but I do like the linting so I think I'll just go with that for now.

SmittieC (approved):

SmittieC (approved):

PR #1900, opened 2025-07-17
Responses API: Support for downloading generated files

### LINKED ISSUE #1895: Support downloading files generated by the responses API (code execution tool)
[Related Slack thread](https://dimagi.slack.com/archives/C05PK63Q89H/p1752663055115099)

Users need to be able to download the files that are generated by the code execution tool when using the responses API. See [here](https://github.com/dimagi/open-chat-studio/blob/cf31a475ebe18a35b2cfb8d8aaafb24e44c8e7ca/apps/service_providers/llm_service/runnables.py#L466-L470) for how this was being done for assistants.

[Related OpenAI docs](https://platform.openai.com/docs/guides/tools-code-interpreter)

SmittieC:
## Description

Resolves #1895 
This is a semi-urgent feature. See [this thread](https://dimagi.slack.com/archives/C05PK63Q89H/p1752663055115099).

When using the code execution tool and a file is generated, users will now be able to download it. OCS checks for `container_file_citation` annotations, then create files in OCS for those files and attach them to the message.

This PR also includes a refactor to the code that parses the LLM output.
- `LlmChatResponse` is extended to include generated files
- `_parse_output` now returns `LlmChatResponse`, which is also now addable
```python
LlmChatResponse(text="Hello", cited_files=[File1]) + LlmChatResponse(text=" World", cited_files=[File2]) = LlmChatResponse(text="Hello World", cited_files=[File1, File2])
```

## User Impact

When using the code execution tool and a file is generated, users will be able to download it.

### Demo

Here's some screenshots. As you can see, my spelling is impeccable

image

image

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/137

### DIFF
--- a/apps/service_providers/llm_service/adapters.py
+++ b/apps/service_providers/llm_service/adapters.py
@@ -164,8 +164,10 @@ class ChatAdapter(BaseAdapter):
         """`cited_files` is a list of files that are cited in the response whereas generated files are those generated
         by the LLM
         """
-        self.session.chat.attach_files(attachment_type="file_citation", files=cited_files)
-        self.session.chat.attach_files(attachment_type="openai_file_ids", files=generated_files)
+        if cited_files:
+            self.session.chat.attach_files(attachment_type="file_citation", files=cited_files)
+        if generated_files:
+            self.session.chat.attach_files(attachment_type="openai_file_ids", files=generated_files)
         return {
             "cited_files": [file.id for file in cited_files],
             "openai_file_ids": [file.external_id for file in generated_files],
--- a/apps/service_providers/llm_service/adapters.py
+++ b/apps/service_providers/llm_service/adapters.py
@@ -170,7 +170,7 @@ class ChatAdapter(BaseAdapter):
             self.session.chat.attach_files(attachment_type="code_interpreter", files=generated_files)
         return {
             "cited_files": [file.id for file in cited_files],
-            "openai_file_ids": [file.external_id for file in generated_files],
+            "generated_files": [file.external_id for file in generated_files],
         }
 
     def add_citation_section_from_cited_files(self, ai_message: str, cited_files: list[File]) -> str:
--- a/apps/service_providers/llm_service/adapters.py
+++ b/apps/service_providers/llm_service/adapters.py
@@ -170,7 +170,7 @@ class ChatAdapter(BaseAdapter):
             self.session.chat.attach_files(attachment_type="code_interpreter", files=generated_files)
         return {
             "cited_files": [file.id for file in cited_files],
-            "generated_files": [file.external_id for file in generated_files],
+            "generated_files": [file.id for file in generated_files],
         }
 
     def add_citation_section_from_cited_files(self, ai_message: str, cited_files: list[File]) -> str:
--- a/apps/service_providers/llm_service/adapters.py
+++ b/apps/service_providers/llm_service/adapters.py
@@ -167,7 +167,7 @@ class ChatAdapter(BaseAdapter):
         if cited_files:
             self.session.chat.attach_files(attachment_type="file_citation", files=cited_files)
         if generated_files:
-            self.session.chat.attach_files(attachment_type="openai_file_ids", files=generated_files)
+            self.session.chat.attach_files(attachment_type="code_interpreter", files=generated_files)
         return {
             "cited_files": [file.id for file in cited_files],
             "openai_file_ids": [file.external_id for file in generated_files],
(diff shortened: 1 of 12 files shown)


SmittieC:
@coderabbitai please review

apps/service_providers/llm_service/utils.py
```diff
@@ -108,3 +111,13 @@ def remove_citations_from_text(text: str) -> str:
 
     citation_pattern = re.compile(OCS_CITATION_PATTERN)
     return citation_pattern.sub("", text)
+
+
+def get_openai_container_file_contents(
```
SmittieC, line 116: OpenAI's API only gives us the curl command. I tried using the openai library for this, which works (they added support for it), but Langchain didn't like me using the newer openai version. I think we should wait a  bit before using the openai client for this operation

apps/service_providers/llm_service/main.py
```diff
@@ -229,6 +262,72 @@ def _get_model_kwargs(self):
...
+
+        for entry in annotation_entries:
+            if entry.get("type") != "container_file_citation":
+                continue
+
+            file_external_id = entry["file_id"]
+            container_id = entry["container_id"]
+
+            # Check if file already exists in database
+            existing_file = File.objects.filter(
```
stephherbers, line 288: are we worried about race conditions here? should we add a `with transaction.atomic():`?

stephherbers (dismissed):
added some comments, fyi I don't have full context of this area, but nothing stood out as alarming

apps/service_providers/llm_service/main.py
SmittieC: No, since OpenAI will generate a new external id for it. Actually, I think I'm going to remove this check altogether. This file will never exist in the DB by this time

SmittieC:
FYI I'll check in tomorrow to see if there are any more feedback, then I'll merge it quickly

apps/service_providers/llm_service/adapters.py
```diff
@@ -160,13 +160,16 @@ def check_cancellation(self):
...
+    def get_output_message_metadata(self, cited_files: set[File], generated_files: set[File]) -> dict:
+        """`cited_files` is a list of files that are cited in the response whereas generated files are those genrated
+        by the LLM
+        """
         self.session.chat.attach_files(attachment_type="file_citation", files=cited_files)
-        return {"cited_files": [file.id for file in cited_files]}
+        self.session.chat.attach_files(attachment_type="openai_file_ids", files=generated_files)
+        return {
+            "cited_files": [file.id for file in cited_files],
+            "openai_file_ids": [file.external_id for file in generated_files],
```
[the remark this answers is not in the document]
snopoke: Does this only work for the responses API? What if Gemini or Claude returned an attachment? I'm just wondering if we should do something different. I'm also wondering why we store the external ID instead of the OCS file ID.

apps/service_providers/llm_service/main.py
```diff
@@ -151,31 +161,42 @@ def attach_built_in_tools(self, built_in_tools: list[str], config: dict[str, Bas
...
+    def _default_parser(
+        self, llm_output, session: ExperimentSession, include_citations: bool = True
+    ) -> LlmChatResponse:
+        llm_outputs = llm_output.get("output", "")
+        final_text = ""
+        cited_file_ids_remote = []
+        cited_file_ids = []
+        cited_files = []
+        generated_files: list[File] = []
+        if isinstance(llm_outputs, list):
```
snopoke, line 173: This seems to indicate that the output was from a different API or provider, is that correct?

apps/service_providers/llm_service/adapters.py
```diff
@@ -160,13 +160,16 @@ def check_cancellation(self):
...
-        if not cited_files:
-            return {}
-
+    def get_output_message_metadata(self, cited_files: set[File], generated_files: set[File]) -> dict:
+        """`cited_files` is a list of files that are cited in the response whereas generated files are those generated
+        by the LLM
+        """
         self.session.chat.attach_files(attachment_type="file_citation", files=cited_files)
-        return {"cited_files": [file.id for file in cited_files]}
+        self.session.chat.attach_files(attachment_type="openai_file_ids", files=generated_files)
```
snopoke, line 168: shouldn't these check if there are files to attach?

snopoke (commented):
Partial review. I'll go over it again.

apps/service_providers/llm_service/adapters.py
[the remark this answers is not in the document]
SmittieC: > Does this only work for the responses API?

I realize you might be referring here to the general feature as apposed to the specific code snippet. The answer then is "no", but only because we don't yet support code execution with Claude or gemini. I am not aware of any other situation in which files could be generated by the LLM. In the case of citations, our current implementation of parsing anthropic responses doesn't find the relevant file(s) on OCS, so I've added a todo to look into it. I'll create a ticket to address this, as well as the potential removal of the `tool_type`

snopoke:
> @snopoke @stephherbers any critical feedback, or can I deploy this?

LGTM

snopoke (approved):

PR #1909, opened 2025-07-22
Disable DATA_UPLOAD_MAX_NUMBER_FIELDS

SmittieC:
## Description

Fixes [these kinds](https://dimagi.sentry.io/issues/6760058348/events/316e3e55c9ee4b5ea43bbd8728085150/?end=2025-07-22T09%3A03%3A00&start=2025-07-22T08%3A58%3A00) of errors

Disable [DATA_UPLOAD_MAX_NUMBER_FIELDS](https://docs.djangoproject.com/en/5.2/ref/settings/#data-upload-max-number-fields) to allow users to upload

I am not convinced that this is the correct solution, so feel free to dispute it

### Docs and Changelog

Pending


snopoke (commented):
Do you know what accounts for the fields?

SmittieC:
> Do you know what accounts for the fields?

From what I can see, it's the sum of everything, but especially the sessions. I put print statements in the failing area of django and saw the following for a chat I had 3 session in:

```
field_name: csrfmiddlewaretoken
	value: 3fzj11wOkEYek93b1KJPqaCtorBvebkAtxTLmdZZBUnYfQWkVqfpHUtRkrNtl2Py
field_name: name
	value: sdsd
field_name: description
	value: dsd
field_name: sessions
	value: 76
field_name: sessions
	value: 75
field_name: sessions
	value: 74
field_name: provider_model
	value: 4:13
field_name: translation_language
	value: 
field_name: translation_provider_model
	value: 4:13
Final POST: <QueryDict: {'csrfmiddlewaretoken': ['3fzj11wOkEYek93b1KJPqaCtorBvebkAtxTLmdZZBUnYfQWkVqfpHUtRkrNtl2Py'], 'name': ['sdsd'], 'description': ['dsd'], 'sessions': ['76', '75', '74'], 'provider_model': ['4:13'], 'translation_language': [''], 'translation_provider_model': ['4:13']}>
```

So for each session, there is a field name. What is weird, is that there are < 800 sessions for the failing bot, so the other fields could not have made up more than 200, so I am not sure why this is happening to be honest. That said, if there close to > 1k sessions, we will sit with this same problem again. Maybe we should rather increase this number to 2k?

snopoke:
I think keeping the current value is fine and limiting the number of sessions you can select to say 500 is a better option.

SmittieC:
> I think keeping the current value is fine and limiting the number of sessions you can select to say 500 is a better option.

See https://github.com/dimagi/open-chat-studio/pull/1909/commits/044a8fcf510c4e4762dc899b027ca00094384df1.

Here's what it looks like:

demo

PR #1916, opened 2025-07-22
Markitdown support

SmittieC:
## Description

Uses [markitdown](https://github.com/microsoft/markitdown) as the default (and only for now) reader. It has support for pdf, docx, xlsx, xls, outlook, pptx files, so we get a few file types for free when using this. I am converting file contents to markdown, since
1. this code is only used to read files for LLMs
2. LLMs like markdown

## User Impact

User will be able to read more file types.

### Demo

N/A

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/138


snopoke (approved):
Awesome 👏🏻

PR #1931, opened 2025-07-23
MCP tool support

SmittieC:
## Description

- Added [langchain-mcp-adapters](https://github.com/langchain-ai/langchain-mcp-adapters)
- Added uvicorn to run async code (the mcp code requires this)
- Added the ability to configure a remote MCP server + refresh it manually
- Users can choose tools from the MCP server in LLM nodes

> [!NOTE]
> The MCP tool is an async method, so I initially adapted the code to use async functions, but later pivoted (https://github.com/dimagi/open-chat-studio/pull/1931/commits/9d66be75a5c2820ba1d66fce550f88b8f16a1f63) to instead convert the coroutine function of the `StructuredTool` to a sync function, otherwise we'd have to make a lot of existing methods async methods which is...not nice.

### Supported MCP resources
We currently only support `tools`.

### General approach
I have an `McpServer` class. This has an `available_tools` array where we store the tool names that we fetched from the remote server. This list is shown to users in the pipeline. More accurately, we show
```
<mcp server name>: <tool name>
```
to the user. When a user chooses a tool, we save something like
```
<mcp server id>: <tool name>
```
in the node params, against the `mcp_tools` key. I chose to go with this approach rather than having a separate DB table for these tools, since it's easier to manage.

### Auth
The mcp server has a FK to `service_providers.AuthProvider`, so it supports whatever auth OCS supports. OAuth would be ideal, since this is becoming the standard for auth-ing with MCP servers, but we can do that in a next step.

## User Impact

Users can add tools from remote MCP servers to their bots. 

### Demo

demo

### Docs and Changelog

Pending


apps/pipelines/tests/node_schemas/LLMResponseWithPrompt.json
```diff
@@ -166,6 +166,16 @@
       "title": "Tool Config",
       "ui:widget": "none",
       "type": "object"
+    },
+    "mcp_tools": {
+      "description": "MCP tools to enable for the bot",
+      "items": {
+        "type": "string"
+      },
+      "title": "Mcp Tools",
```
snopoke, line 175: can we make this `MCP Tools`?

snopoke (dismissed):

snopoke (approved):

PR #1934, opened 2025-07-24
Ba/update chat start api

barora05:
## Description
Add remote_id and name param in chat_start_Session api. Show remote_id in participant_view


apps/api/views/chat.py
```diff
@@ -115,11 +119,25 @@ def chat_start_session(request):
...
     else:
-        participant = Participant.create_anonymous(team, ChannelPlatform.API)
-
+        participant = Participant.create_anonymous(team, ChannelPlatform.API, remote_id)
+
+    if remote_id and participant.remote_id != remote_id:
+        participant.remote_id = remote_id
+        participant.save(update_fields=["remote_id"])
+
+    if name:
```
snopoke, line 131: this should also set `Participant.name` if `name` is provided

apps/api/tests/test_chat_api.py
```diff
@@ -194,3 +194,30 @@ def test_start_chat_session_with_auth(team_with_users, authed_client, experiment
...
+        "name": name,
+        "session_data": session_state,
+    }
+
+    response = authed_client.post(url, data=data, format="json")
+    assert response.status_code == 201
+    response_json = response.json()
+
+    participant = Participant.objects.get(identifier=remote_id)
+    assert response_json["participant"]["identifier"] == remote_id
```
snopoke, line 217: This isn't correct. The `remote_id` should not be used as the participant identifier.

I think this should be:
```suggestion
    participant = Participant.objects.get(identifier=response["participant"]["identifier"])
    assert response_json["participant"]["remote_id"] == remote_id
```

(you can add `remote_id` to the output serializer)

apps/api/serializers.py
```diff
@@ -205,6 +205,10 @@ class ChatStartSessionRequest(serializers.Serializer):
         help_text="Optional initial state data for the session. "
         "This field will be ignored if the request is not authenticated.",
     )
+    remote_id = serializers.CharField(
+        label="Remote Id", required=False, help_text="Optional ID for the participant from remote systems"
+    )
+    name = serializers.CharField(label="Name", required=False, help_text="Optional participant name")
```
snopoke, line 211: sorry for only seeing this late. I think these should be prefixed with `participant_` otherwise it's not that clear whether they apply to the session or the participant (without reading the help text).

snopoke (commented):
One late comment but otherwise looks good.

apps/api/serializers.py
barora05: updated names in the serializer

snopoke (approved):

snopoke (approved):

PR #1941, opened 2025-07-24
Evals Generation Step

proteusvacuum:
## Description

This adds the ability to select a specific chatbot version (or "latest released / latest published") that will run on each "input" message in a dataset. 

It creates a new Evaluation channel which a new session is created against. History is populated by creating ChatMessages against this session, and running the bot as normal. To do this, I had to change the way history is currently stored on the EvaluationMessage, along with storing summaries. 

To look into is how to populate participant data properly, although this can be done through the dataset message context currently.

## User Impact

User can first run messages through the bot and use the output in an evaluation. 

### Demo

https://github.com/user-attachments/assets/f8b421b0-67c2-4414-8ec0-0f8a811f9584


snopoke (approved):
I made a few tweaks after playing with it. The most notable one is that I linked the eval session to the run result and displayed that in the UI.

snopoke (dismissed):

PR #1943, opened 2025-07-24
Ba/filter remote

### LINKED ISSUE #1913: Allow filtering session by `remote_id`
Related to https://github.com/dimagi/open-chat-studio/issues/1912

See https://docs.google.com/document/d/1KnhOlgMcJTntq9BMxiUKZs70Hhe8ncsSBYVhvlE0Llc/edit?tab=t.rusxb4sq9p46#heading=h.q6sxyorgied2

barora05:
## Description
Closes #1913 

### Demo
[Demo](https://drive.google.com/file/d/13YuVPJdsXOG2LD9xMwvcoy4f4vfgzhcA/view?usp=drive_link)

### DIFF
--- a/apps/participants/views.py
+++ b/apps/participants/views.py
@@ -180,5 +180,10 @@ def participant_identifiers_by_experiment(request, team_slug: str, experiment_id
 @permission_required("experiments.view_participant")
 @login_and_team_required
 def all_participant_identifiers(request, team_slug: str):
-    identifiers = list(Participant.objects.filter(team__slug=team_slug).values_list("identifier", flat=True).distinct())
+    identifiers = list(
+        Participant.objects
+        .filter(team__slug=team_slug)
+        .values_list("identifier", "remote_id")
+        .distinct()
+    )
     return JsonResponse(identifiers, safe=False)
--- a/templates/experiments/filters.html
+++ b/templates/experiments/filters.html
@@ -265,23 +265,16 @@
         })
           .then(res => res.json())
           .then(data => {
-            let identifiers = [];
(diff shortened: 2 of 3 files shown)


apps/participants/views.py
```diff
@@ -168,10 +168,10 @@ def cancel_schedule(request, team_slug: str, participant_id: int, schedule_id: s
 
 @permission_required("experiments.view_participant")
 @login_and_team_required
-def participant_identifiers_by_experiment(request, team_slug: str, experiment_id: int):
+def participants_data_by_experiment(request, team_slug: str, experiment_id: int):
```
snopoke, line 171: this name change and the url name changes don't seem necessary. The view still returns participant identifiers and not participant data (which is typically used to refer to ParticipantData)

templates/experiments/filters.html
```diff
@@ -263,10 +265,24 @@
...
+            let remoteIds = [];
+
+            if (hasExperiment) {
+              identifiers = data.map(d => d[0]).filter(id => id != null);
+              remoteIds = data.map(d => d[1]).filter(id => id != null);
+              this.filterData.columns.remote_id.options = remoteIds;
+            } else {
+              identifiers = data.filter(id => id != null);
+            }
+            this.filterData.columns.participant.options = identifiers;
```
snopoke, line 278: why not update `all_participant_identifiers` to also return `remote_ids`?

barora05: we can if we want to show remote_id filter in datasets also..

snopoke (approved):

PR #1948, opened 2025-07-25
Remove file input from the file form

### LINKED ISSUE #1940: unicode source files for RAG not displaying properly on view file
To recreate:

1. Create a new collection
2. Add a file for RAG with any embedding model (I used the default openai one)
3. If you click the eyeball it downloads and everything is fine. If you click the puzzle to view chunks it displays fine:

Image
 
5. However, if you click the source material and click the link inside the file:
Image

then it displays incorrectly:
Image

My guess is this has something to do with the web server and not setting headers that it's UTF-8 or something like that.

SmittieC:
## Description

Resolves #1940 

Removes the file input to prevent users from changing the underlying file. I don't think it makes sense to allow them to do it at this level. All the necessary logic around indexing etc sits on the collection level, so users should rather just remove + upload a file if they want to change a file.

## User Impact

User will not see the file input when navigating to a collection's file anymore.

### Demo

N/A

### Docs and Changelog

I don't think a changelog entry is necesary?

### DIFF
--- a/apps/files/forms.py
+++ b/apps/files/forms.py
@@ -54,11 +54,7 @@ class MultipleFileFieldForm(forms.Form):
 class FileForm(forms.ModelForm):
     class Meta:
         model = File
-        fields = ["name", "summary", "file"]
+        fields = ["name", "summary"]
         help_texts = {
             "summary": "This is only needed when the file will not be used for RAG",
         }
-
-    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


SmittieC:
> > Can users see which collection a file is in from this view?
> 
> No, but that might be useful to know. I can add a chip for that

@claude can you implement a chip on the file form / view that points to the collection that this file belongs to please?

PR #1957, opened 2025-07-28
Fix: Cleanup local index embeddings when deleting a file

SmittieC:
## Description

Fixes https://dimagi.sentry.io/issues/6774487268/?alert_rule_id=14063560&alert_type=issue&notification_uuid=55facddf-2efe-4719-86a1-90342be00eaa&project=4505001320316928&referrer=issue_alert-slack

The issue was that we didn't handle file deletions for local indexes properly. A file is considered used when another object is using the file. This can be an assistant, or a `chat_attachment` (when a file was referenced in a response, we create these). We now remove the file chunk embeddings when removing the file while it is being used. Note that if the file is not used, we do the same, but we also clear out the file's external id, which we don't want to do in the case where the file is being used.

## User Impact

Users will now be able to remove _all_ file from local indexes again

### Demo

N/A

### Docs and Changelog

Pending

### DIFF
--- a/apps/assistants/sync.py
+++ b/apps/assistants/sync.py
@@ -420,7 +420,7 @@ def remove_files_from_tool(ocs_resource: ToolResources, files: list[int]):
         if file in file_references:
             if ocs_resource.extra.get("vector_store_id") and file.external_id:
                 index_manager = OpenAIRemoteIndexManager(client, index_id=ocs_resource.extra.get("vector_store_id"))
-                index_manager.delete_file_from_index(file_id=file.external_id)
+                index_manager.pluck_file_from_index(file_id=file.external_id)
         else:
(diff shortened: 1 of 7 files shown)


SmittieC:
> looks good-- wondering if there should be a catch so that it can fail gracefully if there is another error like this rather than showing the shuck screen?

This touches an interesting point: We can, but _should_ we? I mean, we should ideally never show the "oh shucks" screen and handle errors gracefully with appropriate feedback to the user, but if we wrap all logic here in a try-catch, should we do the same for all other views? Probably?

Maybe a better solution would then we be do it in middleware?

(I'm going to merge this in the meantime, but we can keep the conversation going)

snopoke:
I made some changes to this logic in #1796 as well

PR #1959, opened 2025-07-29
add message level filter

### LINKED ISSUE #1919: CCC feature request: ability to filter transcripts by message-level tags, not only session-level tags
Because message-level tags don't appear in the filters, it requires the team to manually add any message-level tags we used to the list of session-level tags at the top. it'd be great if:

1. you could filter sessions by message-level tags
2. and/or there was a better UI for summarizing the message-level tags used throughout a given session, somewhere at the top of the session

barora05:
## Description
Closes #1919

### DIFF
--- a/apps/experiments/filters.py
+++ b/apps/experiments/filters.py
@@ -192,21 +192,41 @@ def build_tags_filter(operator, value):
         if not selected_tags:
             return None
         if operator == Operators.ANY_OF:
-            return Q(chat__tags__name__in=selected_tags)
+            chat_tags_condition = Q(chat__tags__name__in=selected_tags)
+            message_tags_condition = Q(chat__messages__tags__name__in=selected_tags)
+            return chat_tags_condition | message_tags_condition
+
         elif operator == Operators.ALL_OF:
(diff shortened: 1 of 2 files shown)


SmittieC (commented):
LGTM. A test or two might help secure the robustness of this feature

apps/experiments/filters.py
```diff
@@ -192,21 +192,41 @@ def build_tags_filter(operator, value):
...
+                        content_type_id=chat_content_type.id,
                         tag__name=tag,
                     )
                 )
+                message_tag_exists = Exists(
+                    CustomTaggedItem.objects.filter(
+                        content_type_id=chat_message_content_type.id,
+                        tag__name=tag,
+                        object_id__in=Subquery(
+                            ChatMessage.objects.filter(chat_id=OuterRef(OuterRef("chat_id"))).values("id")
```
SmittieC, line 217: Since `ChatMessage` has the `chat_id`, you should be able to use
```python
ChatMessage.objects.filter(chat_id=OuterRef("chat_id")).values("id")
```

SmittieC (dismissed):

PR #1968, opened 2025-07-30
override loading

barora05:
## Description
Currently, whenever we start typing and an API call is made, the loading class is applied, which causes a large-sized loader to appear. I have overridden the loading classd
### Demo
[Demo](https://drive.google.com/file/d/1FfHacIOy2KeJfj61JYaEKi7JsAcSxnyV/view?usp=drive_link)

### DIFF
--- a/assets/styles/site-tailwind.css
+++ b/assets/styles/site-tailwind.css
@@ -179,3 +179,6 @@ input[type=number]::-webkit-outer-spin-button {
 .markdown a {
   @apply underline
 }
+.loading {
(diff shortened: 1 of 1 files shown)


snopoke:
I don't think this is correct. The `loading` class get's added to the main element so with this styling we're changing the width and height of the main element and making it spin.

TomSelect does have a spinner which shows in the dropdown. You can see it if you play with the examples here: https://tom-select.js.org/examples/remote/

It's default styling is https://github.com/orchidjs/tom-select/blob/4522c18a0236061d6af19475e202a796f4a35a99/src/scss/_dropdown.scss#L65C3-L93

PR #1977, opened 2025-07-31
chatbot table improvements

snopoke:
## Description
Update main chatbot table to include usage information and be sorted by last message date.

## User Impact
Most active bots are sorted to the top of the list making it easier to find active bots.

### Demo
image


SmittieC (approved):
Nice!

PR #1995, opened 2025-08-04
Ba/search in files

### LINKED ISSUE #1945: add search to collections file view

barora05:
## Description
Closes #1945

### Demo
[Demo](https://drive.google.com/file/d/1w434a_1u2QB6asGNxOgU6yEPkWxKIK4S/view?usp=drive_link)

### DIFF
--- a/apps/documents/views.py
+++ b/apps/documents/views.py
@@ -108,24 +108,26 @@ def collection_files_view(request, team_slug: str, collection_id: int, document_
         .annotate(count=Count("id"))
         .values_list("count")
     )
-    collection_files = (
-        CollectionFile.objects.filter(collection=collection, document_source=document_source)
-        .annotate(
-            chunk_count=Subquery(chunk_count_query, output_field=IntegerField()),
-            # Extract directory part
-            directory=Case(
-                When(
-                    file__name__contains="/",
-                    then=Func("file__name", Value("/[^/]*$"), Value("/"), function="regexp_replace"),
-                ),
-                default=Value(""),
-                output_field=CharField(),
-            ),
-            # Determine if it's a subdirectory file
-            depth=Func("file__name", Value("/"), function="regexp_count"),
-        )
-        .order_by("directory", "depth", "file__name")
+    search_query = request.GET.get("search", "").strip()
+    collection_files = CollectionFile.objects.filter(
+        collection=collection,
+        document_source=document_source
     )
+    if search_query:
+        collection_files = collection_files.filter(file__name__icontains=search_query)
+    collection_files = collection_files.annotate(
+        chunk_count=Subquery(chunk_count_query, output_field=IntegerField()),
+        directory=Case(
+            When(
+                file__name__contains="/",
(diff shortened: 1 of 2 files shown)


apps/documents/views.py
```diff
@@ -108,24 +108,26 @@ def collection_files_view(request, team_slug: str, collection_id: int, document_
...
-        )
-        .order_by("directory", "depth", "file__name")
+    search_query = request.GET.get("search", "").strip()
+    collection_files = CollectionFile.objects.filter(
+        collection=collection,
+        document_source=document_source
     )
+    if search_query:
+        collection_files = collection_files.filter(file__name__icontains=search_query)
+    collection_files = collection_files.annotate(
```
SmittieC, line 118: Any reason you're we're not doing the annotation in the main query on line 112 anymore?

barora05: collection_files.annotate part is after and outside the search_query condition. Won't it get applied?

SmittieC: Ah I see. On github it looked like it's inside the if block. Fine as-is. Just from a "clean" perspective, I would advocate for adding the annotations to the first query, so that there's only two query construction steps instead of 3, but this is nitty. Up to you

barora05:
> This mostly looks OK to me except that it only filters the main files list and not the files lists for document sources.

@snopoke How do i add a document source and then try for that file list? When i click on add document source, and select github repository same view opens as nested view
image

SmittieC (commented):
Maybe add the search bar on the same level as the buttons?

image

templates/documents/single_collection_home.html
```diff
@@ -50,6 +50,19 @@ <h1 class="pg-title">{{ collection.name }} {% if collection.is_index %}<span cla
           {% endif %}
         </div>
       </div>
+
+      <div class="my-2 w-full max-w-md" x-data="fileSearch()">
+        <input
+          x-model="searchQuery"
+          @input.debounce.500ms="triggerSearch"
```
SmittieC, line 57: This approach is fine, but you can use htmx only to achieve the same. If you dispatch a custom event here instead of calling `triggerSearch`, then you should be able to add that event to the `collection-files-container` and `document_source_{{ document_source.id }}` elements' triggers. You can also use [hx-include](https://htmx.org/attributes/hx-include/) to get the search query.

barora05:
> Maybe add the search bar on the same level as the buttons?
> image

image

templates/documents/single_collection_home.html
```diff
@@ -5,7 +5,7 @@
 {% load form_tags %}
 {% load django_tables2 %}
 {% block breadcrumbs %}
-  <div class="text-sm breadcrumbs" aria-label="breadcrumbs">
+<div class="text-sm breadcrumbs" aria-label="breadcrumbs" xmlns:hx-on="http://www.w3.org/1999/xhtml">
```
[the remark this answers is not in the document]
barora05: removed

SmittieC (approved):

snopoke:
@barora05 there are some merge conflicts that need to be resolved

PR #1996, opened 2025-08-04
Fix: Fix versioning of document stores

SmittieC:
## Description

Fixes an issue where if you use collections, the chatbot always shows that there are changes
Fixes another issue where document store versions would get files that are connected to the main collection.

### Demo

N/A

### Docs and Changelog

Pending


snopoke (approved):
Thanks

PR #2002, opened 2025-08-05
cleanup after file deleted during indexing

snopoke:
## Description
This handles an edge case where a file is deleted during the indexing process

Resolves [OPEN-CHAT-STUDIO-12J](https://dimagi.sentry.io/issues/6726459306/events/0e3f0102c9c44a0eaf3bc4db212c41b4/)


SmittieC (approved):
Thanks

Issue #1928, opened 2025-07-23
Adding tags to a message causes the 'translation' status indicators to show


snopoke:
https://github.com/user-attachments/assets/e9130a54-84bf-48cc-98c4-30674cbb2c5d

PR #2004, opened 2025-08-05
Smh/spinner fix

stephherbers:
## Description

resolves [#1928](https://github.com/dimagi/open-chat-studio/issues/1928)

## User Impact

spinner does not show when tags are added/removed 
Bonus: initial page loading and language change indicator not shown also with tag change/ search

### Docs and Changelog

no

### DIFF
--- a/templates/experiments/components/experiment_chat.html
+++ b/templates/experiments/components/experiment_chat.html
@@ -77,8 +77,7 @@
                   hx-vals='{"translate_all": "true"}'
                   hx-include="[name='search'], #translation-provider-all, [name='target_language'], #translation-provider-model-all"
                   hx-target="#messages-container"
-                  hx-indicator="#chat-message-untranslated"
-                  onclick="document.getElementById('translate-all-messages-modal').checked = false;"
+                  onclick="document.getElementById('translate-all-messages-modal').checked = false; showAllTranslationSpinners();"
                 >Translate All Messages
                 </button>
                 <label for="translate-all-messages-modal" class="btn">Cancel</label>
@@ -135,7 +134,7 @@
                         hx-vals='{"language": "{{ language }}"}'
                         hx-include="[name='search'], [name='language'], [name='show_original_translation'], #translation-provider-remaining, #translation-provider-model-remaining"
                         hx-target="#messages-container"
-                        hx-indicator="#chat-message-untranslated"
+                        hx-indicator=".translation-loading-spinner"
                         onclick="document.getElementById('translate-remaining-messages-modal').checked = false;"
                       >Translate Messages
                       </button>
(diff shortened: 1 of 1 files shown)


templates/experiments/components/experiment_chat.html
```diff
@@ -189,39 +196,39 @@ <h3 class="font-bold text-lg">Translate Remaining Messages?</h3>
...
+                          <div id="chat-message-{{ message.id }}-translation">
+                            <div class="translation-content-{{ message.id }}">
                               {% if message.translation %}
                                 <p>{{ message.translation|render_markdown }}</p>
                               {% else %}
                                 <p>{{ message.content|render_markdown }}</p>
                               {% endif %}
                             </div>
-                            <div class="htmx-show">
+                            <div class="translation-loading-spinner-{{ message.id }}" style="display: none;">
```
snopoke, line 217: nit: use `hidden` class instead of inline style

```suggestion
                            <div class="translation-loading-spinner-{{ message.id }} hidden">
```

templates/experiments/components/experiment_chat.html
```diff
@@ -135,7 +134,7 @@ <h3 class="font-bold text-lg">Translate Remaining Messages?</h3>
                         hx-vals='{"language": "{{ language }}"}'
                         hx-include="[name='search'], [name='language'], [name='show_original_translation'], #translation-provider-remaining, #translation-provider-model-remaining"
                         hx-target="#messages-container"
-                        hx-indicator="#chat-message-untranslated"
+                        hx-indicator=".translation-loading-spinner"
```
snopoke, line 137: I don't see anything with this exact class. I do see `translation-loading-spinner-{{ message.id }}` but that won't match this class.

templates/experiments/components/experiment_chat.html
```diff
@@ -77,8 +77,7 @@ <h3 class="font-bold text-lg">Translate All Messages</h3>
                   hx-vals='{"translate_all": "true"}'
                   hx-include="[name='search'], #translation-provider-all, [name='target_language'], #translation-provider-model-all"
                   hx-target="#messages-container"
-                  hx-indicator="#chat-message-untranslated"
-                  onclick="document.getElementById('translate-all-messages-modal').checked = false;"
+                  onclick="document.getElementById('translate-all-messages-modal').checked = false; showAllTranslationSpinners();"
```
snopoke, line 80: Where is `showAllTranslationSpinners` implemented? I couldn't find it.

snopoke (commented):
I tried this out locally and it still needs some work.

Try adding this middleware to simulate a slower connection:

```python
def delay_middleware(get_response):
    def middleware(request):
        import time
        time.sleep(2)  # Introduce a 2-second delay
        response = get_response(request)
        return response
    return middleware
```

Also check out the other actions on the page:
* show original messages
* load tag editor (click tag edit button)
* add tag, remove tag

PR #2007, opened 2025-08-06
Chat Widget: file upload

stephherbers:
## Description

resolves [#1830](https://github.com/dimagi/open-chat-studio/issues/1830)
## User Impact

Screenshot 2025-08-12 at 1 08 28 PM

### Demo

[Loom video](https://www.loom.com/share/0edc24b152514ae5b304d37603138f5d?sid=3897f41f-fa73-4d8d-9f1f-80c5f2f924ce)

### Docs and Changelog

yes and yes (chat widget changelog)


apps/api/views/chat.py
```diff
@@ -16,15 +19,41 @@
...
 from apps.experiments.models import Experiment, ExperimentSession, Participant, ParticipantData
 from apps.experiments.task_utils import get_message_task_response
 from apps.experiments.tasks import get_response_for_webchat_task
+from apps.files.models import File
 
 AUTH_CLASSES = [ApiKeyAuthentication, BearerTokenAuthentication]
 
+MAX_FILE_SIZE_MB = 50
+MAX_TOTAL_SIZE_MB = 50
+SUPPORTED_FILE_EXTENSIONS = [
```
SmittieC, line 35: Where does this list come from?

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
```diff
@@ -295,13 +318,79 @@ export class OcsChat {
...
+  }
+
   private async sendMessage(message: string): Promise<void> {
     if (!this.sessionId || !message.trim()) return;
 
     // Hide starter questions on any user interaction
     this.showStarterQuestions = false;
 
     try {
+      let attachmentIds: number[] = [];
```
SmittieC, line 389: I'm curious why you decided to do file uploading and message submission as two separate steps?

snopoke: From my perspective I'm fine with having them be separate for two reasons:
1. We could do the upload as soon as the file is added instead of waiting for the user to click 'send'. This will reduce the latency. (I'm not proposing we do this as part of this PR).
2. If we move to websockets, we'd need to upload the files separately anyway.

stephherbers:
fyi updated screenshot with new icon sizing

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
```diff
@@ -309,13 +337,79 @@ export class OcsChat {
...
+      // Only upload if there are new files
+      if (formData.has('files')) {
+        const response = await fetch(`${this.getApiBaseUrl()}/api/chat/${this.sessionId}/upload/`, {
+          method: 'POST',
+          body: formData,
+        });
+
+        if (!response.ok) {
+          const errorData = await response.json();
+          throw new Error(errorData.error || 'Failed to upload files');
```
snopoke, line 376: This should rather update the selecteFiles to add the `error` field so that the UI displays the error instead of closing the whole chat window.

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
```diff
@@ -328,27 +434,41 @@ export class OcsChat {
...
       const userMessage: ChatMessage = {
         created_at: new Date().toISOString(),
         role: 'user',
         content: message.trim(),
-        attachments: []
+        attachments: this.allowAttachments ? this.selectedFiles.map(sf => ({
+          name: sf.file.name,
+          content_type: sf.file.type,
+          size: sf.file.size,
+          content_url: '', // We don't have the URL yet
```
snopoke, line 447: I think this can be removed

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
```diff
@@ -321,13 +348,99 @@ export class OcsChat {
...
     try {
+      let attachmentIds: number[] = [];
+      if (this.allowAttachments && this.selectedFiles.length > 0) {
+        attachmentIds = await this.uploadFiles();
+
+        // Check if any files have errors after upload attempt
+        const hasErrors = this.selectedFiles.some(sf => sf.error);
+        if (hasErrors) {
+          // Don't send the message, let user fix file issues first
+          this.error = 'Please remove or fix file errors before sending your message.';
```
snopoke, line 439: Setting the error here will hide the chat window entirely. If we're going to use this field we need to change how it is rendered.

But perhaps we can do that in a followup PR.

snopoke (dismissed):

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
snopoke: https://github.com/dimagi/open-chat-studio/issues/2063

PR #2011, opened 2025-08-06
make chat widget style more customizable

snopoke:
## Description
* Use classes for components instead of inline styles
* Use CSS vars for configurable styling

## User Impact
Users can customize the widget style without having to resort to hacks.

### Demo
Overall the widget looks the same though I did fix some font size discrepencies.

### Docs and Changelog
https://github.com/dimagi/open-chat-studio-docs/pull/146


stephherbers (approved):
🫡

PR #2027, opened 2025-08-11
Ba/stt voice

barora05:
## Description
WIP

### DIFF
--- a/apps/chatbots/views.py
+++ b/apps/chatbots/views.py
@@ -223,13 +223,17 @@ class EditChatbot(LoginAndTeamRequiredMixin, TemplateView, PermissionRequiredMix
         llm_providers = LlmProvider.objects.filter(team=self.request.team).values("id", "name", "type").all()
         llm_provider_models = LlmProviderModel.objects.for_team(self.request.team).all()
         experiment = get_object_or_404(Experiment.objects.get_all(), id=kwargs["pk"], team=self.request.team)
-
+        exclude_services = [SyntheticVoice.OpenAIVoiceEngine]
+        if flag_is_active(self.request, "flag_open_ai_voice_engine"):
+            exclude_services = []
         return {
             **data,
             "pipeline_id": experiment.pipeline_id,
             "node_schemas": _pipeline_node_schemas(),
             "experiment": experiment,
-            "parameter_values": _pipeline_node_parameter_values(self.request.team, llm_providers, llm_provider_models),
+            "parameter_values": _pipeline_node_parameter_values(
+                self.request.team, llm_providers, llm_provider_models, exclude_services
+            ),
             "default_values": _pipeline_node_default_values(llm_providers, llm_provider_models),
             "origin": "chatbots",
             "flags_enabled": [flag.name for flag in Flag.objects.all() if flag.is_active_for_team(self.request.team)],
--- a/apps/pipelines/nodes/base.py
+++ b/apps/pipelines/nodes/base.py
@@ -433,7 +433,7 @@ class Widgets(StrEnum):
     built_in_tools = "built_in_tools"
     key_value_pairs = "key_value_pairs"
     text_editor = "text_editor_widget"
-
+    voice_widget = "voice_widget"
 
 class OptionsSource(StrEnum):
     source_material = "source_material"
@@ -447,7 +447,8 @@ class OptionsSource(StrEnum):
     built_in_tools_config = "built_in_tools_config"
     text_editor_autocomplete_vars_llm_node = "text_editor_autocomplete_vars_llm_node"
     text_editor_autocomplete_vars_router_node = "text_editor_autocomplete_vars_router_node"
-
+    voice_provider_id = "voice_provider_id"
+    synthetic_voice_id = "synthetic_voice_id"
 
 class UiSchema(BaseModel):
     widget: Widgets = None
--- a/apps/pipelines/nodes/nodes.py
+++ b/apps/pipelines/nodes/nodes.py
@@ -398,6 +398,8 @@ class LLMResponseWithPrompt(LLMResponse, HistoryMixin, OutputMessageTagMixin):
         PipelineChatHistoryTypes.GLOBAL,
         json_schema_extra=UiSchema(widget=Widgets.history, enum_labels=PipelineChatHistoryTypes.labels),
     )
+    voice_provider_id: int = Field(..., title="Voice Model", json_schema_extra=UiSchema(widget=Widgets.voice_widget))
+    synthetic_voice_id: int = Field(..., json_schema_extra=UiSchema(widget=Widgets.none))
 
     @model_validator(mode="after")
     def check_prompt_variables(self) -> Self:
--- a/apps/pipelines/views.py
+++ b/apps/pipelines/views.py
@@ -17,11 +17,12 @@ from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_POST
 from django.views.generic import TemplateView
 from django_tables2 import SingleTableView
+from waffle import flag_is_active
 
 from apps.assistants.models import OpenAiAssistant
 from apps.custom_actions.form_utils import get_custom_action_operation_choices
 from apps.documents.models import Collection
-from apps.experiments.models import AgentTools, BuiltInTools, Experiment, SourceMaterial
+from apps.experiments.models import AgentTools, BuiltInTools, Experiment, SourceMaterial, SyntheticVoice
 from apps.pipelines.flow import FlowPipelineData
 from apps.pipelines.models import Pipeline
 from apps.pipelines.nodes.base import OptionsSource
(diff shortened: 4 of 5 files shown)


barora05:
@coderabbitai please review

apps/pipelines/views.py
```diff
@@ -83,13 +84,21 @@ def get_context_data(self, **kwargs):
...
+        if flag_is_active(self.request, "flag_open_ai_voice_engine"):
+            exclude_services = []
         return {
             **data,
             "pipeline_id": kwargs["pk"],
             "pipeline_name": pipeline.name,
             "node_schemas": _pipeline_node_schemas(),
             "parameter_values": _pipeline_node_parameter_values(
-                self.request.team, llm_providers, llm_provider_models, include_versions=pipeline.is_a_version
+                self.request.team,
```
SmittieC, line 96: Nit: I like to use named parameters whenever there are > 2 args. Makes it easy to read from the caller's side

apps/chat/bots.py
```diff
@@ -302,7 +302,9 @@ def invoke_pipeline(
         else:
             result = ChatMessage(content=output)
         self._process_intents(output)
-        return result
+        voice_provider_id = output.get("voice_provider_id", None)
+        synthetic_voice_id = output.get("synthetic_voice_id", None)
+        return result, voice_provider_id, synthetic_voice_id
```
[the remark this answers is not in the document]
barora05: I have removed returning tuple. I have added 2 id fields in pipeline bot

apps/chatbots/views.py
```diff
@@ -222,14 +222,20 @@ def get_context_data(self, **kwargs):
         data = super().get_context_data(**kwargs)
         llm_providers = LlmProvider.objects.filter(team=self.request.team).values("id", "name", "type").all()
         llm_provider_models = LlmProviderModel.objects.for_team(self.request.team).all()
-        experiment = get_object_or_404(Experiment.objects.get_all(), id=kwargs["pk"], team=self.request.team)
-
+        experiment = get_object_or_404(
+            Experiment.objects.get_all().select_related("voice_provider"), id=kwargs["pk"], team=self.request.team
+        )
+        exclude_services = [SyntheticVoice.OpenAIVoiceEngine]
```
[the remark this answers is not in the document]
barora05: this is called from 2 places, from chatbots and pipelines view. We have experiment only in chatbot view and secondly we would need to pass self.request for exclude services

apps/pipelines/views.py
barora05: done

apps/chat/channels.py
```diff
@@ -680,7 +693,10 @@ def _transcribe_audio(self, audio: BytesIO) -> str:
         return "Unable to transcribe audio"
 
     def _get_bot_response(self, message: str) -> ChatMessage:
-        return self.bot.process_input(message, attachments=self.message.attachments)
+        chat_message = self.bot.process_input(message, attachments=self.message.attachments)
+        self.output_voice_provider_id = getattr(self.bot, "voice_provider_id", None)
```
SmittieC, line 697: I still think it will be valuable to let the bot synthesize the voice, since it has the knowledge of which voice to use. In the `TopicBot` case, we use the `processor_experiment`. In the `Pipeline` bot, we use `voice_provider_id` and `output_synthetic_voice_id`. The only reason that `processor_experiment` is being used is to know which voice to use

snopoke: @SmittieC I'm not quite following what you're saying

snopoke: That seems reasonable

apps/chat/bots.py
```diff
@@ -407,6 +427,25 @@ def _save_message_to_history(
                 chat_message.create_and_add_tag(tag_value, self.session.team, category or "")
         return chat_message
 
+    def synthesize_voice(self) -> tuple["VoiceProvider", "SyntheticVoice"] | None:
+        from apps.experiments.models import SyntheticVoice
+        from apps.service_providers.models import VoiceProvider
+
+        voice_provider = (
+            VoiceProvider.objects.get(id=self.voice_provider_id)
+            if getattr(self, "voice_provider_id", None) is not None
```
snopoke, line 436: nit: `getattr` isn't necessary here since these fields are declared on the class during init.
```suggestion
            if self.voice_provider_id is not None
```

apps/chat/bots.py
```diff
@@ -407,6 +427,25 @@ def _save_message_to_history(
...
+            else None
+        )
+
+        synthetic_voice = (
+            SyntheticVoice.objects.get(id=self.synthetic_voice_id)
+            if getattr(self, "synthetic_voice_id", None) is not None
```
snopoke, line 442:
```suggestion
            if self.synthetic_voice_id is not None
```

apps/pipelines/nodes/base.py
```diff
@@ -122,12 +124,23 @@ def clone(cls, state):
...
+    ) -> Self:
         kwargs["outputs"] = {node_name: {"message": output, "node_id": node_id}}
         kwargs.setdefault("temp_state", {}).update({"outputs": {node_name: output}})
         if output is not None:
             kwargs["messages"] = [output]
-
+        if voice_provider_id is not None:
+            kwargs["voice_provider_id"] = voice_provider_id
+        if synthetic_voice_id is not None:
+            kwargs["synthetic_voice_id"] = synthetic_voice_id
```
snopoke, line 143: These changes don't seem necessary since we're already accepting arbitrary kwargs

apps/pipelines/views.py
```diff
@@ -226,6 +238,24 @@ def _get_assistant_url(assistant_id: int):
...
+            [_option(selected_voice_provider.id, selected_voice_provider.type)] if selected_voice_provider else []
+        ),
+        OptionsSource.synthetic_voice_id: sorted(
+            [
+                _option(
+                    voice.id,
+                    str(voice),
+                    voice.service.lower(),
+                    None,
+                    None,
```
snopoke, line 251: nit: these can be left out since they are the same as the defaults.

apps/pipelines/views.py
```diff
@@ -226,6 +238,24 @@ def _get_assistant_url(assistant_id: int):
...
+                )
+                | {"provider_id": voice.voice_provider_id}
+                for voice in synthetic_voices
+                if selected_voice_provider and voice.service.lower() == selected_voice_provider.type
```
snopoke, line 255: Why not do this filtering when we do the query? You could avoid this if statement entirely.

apps/pipelines/views.py
```diff
@@ -83,13 +84,22 @@ def get_context_data(self, **kwargs):
...
             **data,
             "pipeline_id": kwargs["pk"],
             "pipeline_name": pipeline.name,
             "node_schemas": _pipeline_node_schemas(),
             "parameter_values": _pipeline_node_parameter_values(
-                self.request.team, llm_providers, llm_provider_models, include_versions=pipeline.is_a_version
+                team=self.request.team,
+                llm_providers=llm_providers,
+                llm_provider_models=llm_provider_models,
+                synthetic_voices=synthetic_voices,
```
snopoke, line 100: You could pass in an empty list here instead since the filtering in the function won't match anything. Then you can also remove the query above.
```suggestion
                synthetic_voices=[],
```

assets/javascript/apps/pipeline/nodes/widgets.tsx
```diff
@@ -1099,3 +1101,74 @@ function TextEditorModal({
...
+  const { parameterValues } = getCachedData();
+  const setNode = usePipelineStore((state) => state.setNode);
+
+  const updateParamValue = (event: ChangeEvent<HTMLSelectElement>) => {
+    const { value } = event.target;
+    const [providerId, syntheticVoiceId] = value.split('|:|');
+    setNode(props.nodeId, (old) =>
+      produce(old, (next) => {
+        next.data.params.voice_provider_id = providerId;
+        next.data.params.synthetic_voice_id = syntheticVoiceId;
```
snopoke, line 1115: Isn't this always going to be the value of `experiment.voice_provider_id`?

apps/pipelines/nodes/nodes.py
```diff
@@ -398,6 +398,10 @@ class LLMResponseWithPrompt(LLMResponse, HistoryMixin, OutputMessageTagMixin):
         PipelineChatHistoryTypes.GLOBAL,
         json_schema_extra=UiSchema(widget=Widgets.history, enum_labels=PipelineChatHistoryTypes.labels),
     )
+    voice_provider_id: OptionalInt = Field(
```
snopoke, line 401: I'm not quite following why we need this field in the node since the value is always going to be `experiment.voice_provider_id`. It seems like we can just drop this field and only record the voice ID.

Am I missing something?

apps/chatbots/views.py
```diff
@@ -222,14 +222,27 @@ def get_context_data(self, **kwargs):
...
             "pipeline_id": experiment.pipeline_id,
             "node_schemas": _pipeline_node_schemas(),
             "experiment": experiment,
-            "parameter_values": _pipeline_node_parameter_values(self.request.team, llm_providers, llm_provider_models),
+            "parameter_values": _pipeline_node_parameter_values(
+                team=self.request.team,
+                llm_providers=llm_providers,
+                llm_provider_models=llm_provider_models,
+                synthetic_voices=synthetic_voices,
+                selected_voice_provider=experiment.voice_provider,
```
snopoke, line 244: Can be removed
```suggestion
```

apps/pipelines/views.py
```diff
@@ -89,7 +89,12 @@ def get_context_data(self, **kwargs):
             "pipeline_name": pipeline.name,
             "node_schemas": _pipeline_node_schemas(),
             "parameter_values": _pipeline_node_parameter_values(
-                self.request.team, llm_providers, llm_provider_models, include_versions=pipeline.is_a_version
+                team=self.request.team,
+                llm_providers=llm_providers,
+                llm_provider_models=llm_provider_models,
+                synthetic_voices=[],
+                selected_voice_provider=None,  # Not show voice field for now if opened from pipelines tab
```
snopoke, line 96: Can be removed
```suggestion
```

apps/pipelines/views.py
```diff
@@ -129,7 +134,9 @@ def delete(self, request, team_slug: str, pk: int):
         return HttpResponse(response, headers={"HX-Reswap": "none"}, status=400)
 
 
-def _pipeline_node_parameter_values(team, llm_providers, llm_provider_models, include_versions=False):
+def _pipeline_node_parameter_values(
+    team, llm_providers, llm_provider_models, synthetic_voices, selected_voice_provider, include_versions=False
```
snopoke, line 138: Can be removed
```suggestion
    team, llm_providers, llm_provider_models, synthetic_voices, include_versions=False
```

snopoke (commented):
Tiny bit of cleanup left

snopoke (approved):

snopoke:
@barora05 please add a changelog entry for this

PR #2029, opened 2025-08-11
OCS Tracing phase 1

SmittieC:


SmittieC:
@coderabbitai and @claude please review this PR. Comment especially on security related issues

apps/service_providers/tracing/langfuse.py
```diff
@@ -134,7 +134,11 @@ def get_trace_metadata(self) -> dict[str, str]:
             "trace_provider": self.type,
         }
 
-    def _get_current_span(self) -> StatefulTraceClient | StatefulSpanClient:
+    def _get_current_observation(self) -> StatefulTraceClient | StatefulSpanClient:
```
SmittieC, line 137: I wasn't sure what to call this, since it can return a trace or span. `get _current_span_or_trace` sounds weird, but I'm not set on `_get_current_observation` either

apps/trace/tables.py
```diff
@@ -0,0 +1,60 @@
...
+from apps.trace.models import Trace
+
+
+def _chip_chatbot_url_factory(_, __, record, ___):
+    return reverse("chatbots:single_chatbot_home", args=[record.team.slug, record.experiment_id])
+
+
+def _chip_session_url_factory(_, __, record, ___):
+    return reverse(
+        "experiments:experiment_session_view",
```
snopoke, line 18: I see you're using the experiment url here but the chatbot url above. You might need to check the feature flag and the record type to determine where to redirect.

snopoke (commented):
Looking good

apps/service_providers/tracing/langfuse.py
SmittieC: Hmm, that doesn't sit quite right either. What I'm trying to do - at the very real risk of being way too pedantic about this - is to keep the method name that returns the equivalent of a "trace or span" the same. The `StatefulTraceClient` in the LangFuse tracer represents the underlying `Trace`, which is in essence the same as the `Trace` that is being returned in `OcsTracer`.

assets/styles/site-tailwind.css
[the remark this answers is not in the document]
snopoke: Still curious why you need to use custom classes vs just putting the tailwind classes in the template.

templates/trace/trace_detail.html
```diff
@@ -12,7 +13,7 @@
 {% endblock breadcrumbs %}
 
 {% block app %}
-<div x-data="pageData">
+<div x-data="{activeTab: 'spans'}" x-init="{ activeTab: 'spans' }">
```
snopoke, line 16: `x-init` shouldn't be necessary since you're already setting the default value.

snopoke (approved):
only minor comments

assets/styles/site-tailwind.css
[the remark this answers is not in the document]
SmittieC: I kept it since the line becomes quite long (see https://github.com/dimagi/open-chat-studio/pull/2029/commits/929af010152060d4c3bce4cb5ac79e17f0e983d6). We can probably keep it this way. Probably not needed to have style classes if it's not going to be re-used elsewhere

PR #2036, opened 2025-08-12
de-duplicate and filter participant options

snopoke:
## Description
The SQL query is return distinct combinations of participant ID and remote ID. This PR de-duplicates the values independently and also removes empty values in the view rather than on the client.

This fixes and error that occurs from having duplicate and blank values in the filters.


apps/participants/views.py
```diff
@@ -169,21 +169,30 @@ def cancel_schedule(request, team_slug: str, participant_id: int, schedule_id: s
...
 
 @permission_required("experiments.view_participant")
 @login_and_team_required
 def all_participant_identifiers(request, team_slug: str):
     identifiers = list(
-        Participant.objects
-        .filter(team__slug=team_slug)
-        .values_list("identifier", "remote_id")
-        .distinct()
+        Participant.objects.filter(team__slug=team_slug).values_list("identifier", "remote_id").distinct()
```
barora05, line 196: return this also as sep lists?

SmittieC (approved):
LGTM

Issue #2053, opened 2025-08-14
[chat widget] increase polling timeout + add message on timeout


snopoke:
* Increase the polling timeout to 120s
* If there isn't a response when we hit the timeout, display a message to the user/

PR #2061, opened 2025-08-14
Chat widget: polling modifications

stephherbers:
## Description

resolves [#2053](https://github.com/dimagi/open-chat-studio/issues/2053)

opted for a system message for easy traceability of what happened in the session history and for a nicely formatted response rather than an error message, but let me know if an error message makes more sense to you in this case

## User Impact

polling time increased to 120 seconds (120 intervals of 1 second)

### Docs and Changelog

no docs but yes changelog

### DIFF
--- a/components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
+++ b/components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
@@ -64,7 +64,7 @@ interface SessionStorageData {
 })
 export class OcsChat {
 
-  private static readonly TASK_POLLING_MAX_ATTEMPTS = 120;
+  private static readonly TASK_POLLING_MAX_ATTEMPTS = 120; // Changed from 30 to 120 for 120 second timeout
   private static readonly TASK_POLLING_INTERVAL_MS = 1000;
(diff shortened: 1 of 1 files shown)


components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
```diff
@@ -64,7 +64,7 @@ interface SessionStorageData {
 })
 export class OcsChat {
 
-  private static readonly TASK_POLLING_MAX_ATTEMPTS = 30;
+  private static readonly TASK_POLLING_MAX_ATTEMPTS = 120; // Changed from 30 to 120 for 120 second timeout
```
SmittieC, line 67: Nit: Is the comment necessary?

SmittieC (approved):
LGTM

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
snopoke:
```suggestion
  private static readonly TASK_POLLING_MAX_ATTEMPTS = 120;
```

snopoke (approved):

PR #2068, opened 2025-08-15
Tracing | Add sparklines to show error trends in experiment and chatbot list

SmittieC:
### Demo

image

# When there were no activity, it's a flat line like here
image


SmittieC:
@snopoke I added some caching behaviour. Thoughts on this commit? https://github.com/dimagi/open-chat-studio/pull/2068/commits/2fd6506cbef2bf8a9ce848f61c5ccb5fba5d5194

snopoke (approved):
LGTM.

Personally I would have put all this code in the `traces` app but 🤷🏻‍♂️

PR #2075, opened 2025-08-19
Ba/read only api key

### LINKED ISSUE #2055: Read-only API keys for OCS
* Add a field to the `UserAPIKey` model to specify read only vs read / write permission
* Default the field to read only (including for existing keys)
* Update the UI to allow setting the value when creating the key (do not allow editing the value once the key is created)
* Create a new DRF permissions class to only allow GET, HEAD, OPTIONS requests for read only API keys

barora05:
## Description
Closes #2055

### Docs and Changelog
https://github.com/dimagi/open-chat-studio-docs/pull/151

### DIFF
--- a/apps/channels/tests/test_api_integration.py
+++ b/apps/channels/tests/test_api_integration.py
@@ -145,3 +145,12 @@ def test_attachments_returned(mock_response, experiment):
         "response": "Fido",
         "attachments": [{"file_name": file.name, "link": file.download_link(session.id)}],
     }
+
+@pytest.mark.django_db()
+def test_read_only_key_cannot_post(experiment):
+    user = experiment.team.members.first()
+
+    client = ApiTestClient(user, experiment.team, read_only=True)
+    new_message_url = reverse("channels:new_api_message", kwargs={"experiment_id": experiment.public_id})
+    response = client.post(new_message_url, data={"message": "What should I call my dog?"}, format="json")
+    assert response.status_code == 403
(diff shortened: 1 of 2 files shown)


apps/api/permissions.py
```diff
@@ -60,6 +59,20 @@ class BearerTokenAuthentication(BaseKeyAuthentication):
     def get_key(self, request):
         return ConfigurableKeyParser(keyword=self.keyword).get_from_authorization(request)
 
+class ReadOnlyAPIKeyPermission(BasePermission):
+    """
+    Allows only safe methods (GET, HEAD, OPTIONS) for read-only API keys.
+    """
+
+    def has_permission(self, request, view):
+        if not hasattr(request, "auth") or request.auth is None:
```
snopoke, line 68: This should also check that the user is authenticate:

```python
if not request.user or not request.user.is_authenticated:
    return False
```

apps/users/forms.py
```diff
@@ -30,9 +30,18 @@ class UploadAvatarForm(forms.Form):
...
 
 class ApiKeyForm(forms.ModelForm):
+    read_only = forms.BooleanField(
+        required=False,
+        initial=True,
+        label=gettext("Read Only - Can only view/fetch data"),
+        help_text=gettext("Leave unchecked for Read/Write access. "
+                          "This cannot be changed after the key is created."),
+        widget=forms.CheckboxInput(),
+    )
```
snopoke, line 40: Thinking about this now I want to invert the condition on the UI so that by default it is unchecked (read only) and the user must check the box to allow write access. 

I just think that makes more sense from a UI perspective and gives a clearer signal to the user that they are allowing something extra by checking the box.

snopoke (approved):

snopoke:
@barora05 please add a PR to the docs repo with a changelog entry

Issue #2063, opened 2025-08-15
[chat widget] Update error rendering


snopoke:
Currently setting the 'error' field in the widget results in only the error message being rendered and the chat window being hidden which is a very jarring experience and can only be resolved by refreshing the page.

PR #2078, opened 2025-08-19
Chat widget: error handling

stephherbers:
## Description

resolves [#2063](https://github.com/dimagi/open-chat-studio/issues/2063)

rather than closing out the chat widget window, it will send the error message as a system message to the user. I opted for this approach because it's clean and uses the existing UI. 

## User Impact

users will still be able to use the widget even if it errors (for non-critical errors)

### Demo

n/a

### Docs and Changelog

no and no

### DIFF
--- a/components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
+++ b/components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
@@ -250,7 +250,7 @@ export class OcsChat {
     const errorMessage: ChatMessage = {
       created_at: new Date().toISOString(),
       role: 'system',
-      content: `**Error:** ${errorText}\nPlease try again or contact support if the problem persists.`,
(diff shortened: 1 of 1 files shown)


components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
```diff
@@ -246,6 +246,30 @@ export class OcsChat {
     window.removeEventListener('resize', this.handleWindowResize);
   }
 
+  private addErrorMessage(errorText: string): void {
+    const errorMessage: ChatMessage = {
+      created_at: new Date().toISOString(),
+      role: 'system',
+      content: `**Error:** ${errorText}\nPlease try again or contact support if the problem persists.`,
```
snopoke, line 253: nit: unclear who 'support' is or how to contact them so let's leave it out for now.
```suggestion
      content: `**Error:** ${errorText}\nPlease try again.`,
```

PR #2080, opened 2025-08-20
add refresh button

### LINKED ISSUE #2023: Add a 'refresh' button to the chat session table
Add a small button to the top right of the session table using the `<i class="fa-solid fa-rotate"></i>` icon. When clicked, re-load the table.

(If it makes sense, this could be added to all tables though it might be hard to with the filters).

barora05:
## Description
Add refresh button and loading spinner.
Closes #2023

### DIFF
--- a/templates/chatbots/single_chatbot_home.html
+++ b/templates/chatbots/single_chatbot_home.html
@@ -219,10 +219,31 @@
 
     {% if perms.chat.view_chat %}
       <div role="tabpanel" class="tab-content" id="content-allsessions">
-        {% include "chatbots/components/session_table_actions.html" %}
-        <div id="sessions-table" data-url="{% url 'chatbots:sessions-list' request.team.slug experiment.id %}">
+        <div class="relative">
+          {% include "chatbots/components/session_table_actions.html" %}
+          <div class="absolute top-0 right-0 flex items-center gap-2">
+            <button class="btn btn-ghost btn-sm"
+                    hx-get="{% url 'chatbots:sessions-list' request.team.slug experiment.id %}"
+                    hx-target="#sessions-table"
+                    hx-swap="innerHTML"
+                    hx-indicator="#refresh-loader"
+                    title="Refresh sessions">
+              <i class="fa-solid fa-rotate"></i>
+            </button>
+            <span class="loading loading-spinner loading-sm htmx-indicator" id="refresh-loader"></span>
+          </div>
+
+          <div id="sessions-table"
+               data-url="{% url 'chatbots:sessions-list' request.team.slug experiment.id %}"
+               hx-get="{% url 'chatbots:sessions-list' request.team.slug experiment.id %}"
+               hx-trigger="load"
+               hx-swap="innerHTML">
+
+            <div class="flex justify-center items-center p-6">
+                <span class="loading loading-spinner loading-lg"></span>
+            </div>
+          </div>
         </div>
-      </div>
       </div>
     {% endif %}
     <div role="tabpanel" class="tab-content" id="content-versions">
--- a/templates/chatbots/single_chatbot_home.html
+++ b/templates/chatbots/single_chatbot_home.html
@@ -234,12 +234,7 @@
             <span class="loading loading-spinner loading-sm htmx-indicator htmx-show" id="refresh-loader"></span>
           </div>
 
-          <div id="sessions-table"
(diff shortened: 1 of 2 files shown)


templates/chatbots/single_chatbot_home.html
```diff
@@ -219,11 +219,32 @@ <h3 class="font-bold text-lg">Link with {{ platform.label }}</h3>
...
+        <div class="relative">
+          {% include "chatbots/components/session_table_actions.html" %}
+          <div class="absolute top-0 right-0 flex items-center gap-2">
+            <button class="btn btn-ghost btn-sm"
+                    hx-get="{% url 'chatbots:sessions-list' request.team.slug experiment.id %}"
+                    hx-target="#sessions-table"
+                    hx-swap="innerHTML"
+                    hx-indicator="#refresh-loader"
+                    title="Refresh sessions">
+              <i class="fa-solid fa-rotate"></i>
```
snopoke, line 231: add `htmx-hide` class to hide this icon during the request
```suggestion
              <i class="fa-solid fa-rotate htmx-hide"></i>
```

templates/chatbots/single_chatbot_home.html
```diff
@@ -219,11 +219,32 @@ <h3 class="font-bold text-lg">Link with {{ platform.label }}</h3>
...
+            </button>
+            <span class="loading loading-spinner loading-sm htmx-indicator" id="refresh-loader"></span>
```
snopoke, line 233: use `htmx-show` so that the spinner isn't included in the layout:
```suggestion
            <span class="loading loading-spinner loading-sm htmx-show" id="refresh-loader"></span>
```

(is `loading-sm` right or should it be `loading-xs`?)

snopoke (commented):
Does this work when filters are applied?

templates/chatbots/single_chatbot_home.html
barora05: we are using sm at most places in the project. currently the size of loader is same as of refresh icon. we can use xs if we want smaller than it.

templates/chatbots/single_chatbot_home.html
```diff
@@ -223,7 +223,7 @@ <h3 class="font-bold text-lg">Link with {{ platform.label }}</h3>
           {% include "chatbots/components/session_table_actions.html" %}
           <div class="absolute top-0 right-0 flex items-center gap-2">
             <button class="btn btn-ghost btn-sm"
-                    hx-get="{% url 'chatbots:sessions-list' request.team.slug experiment.id %}"
+                    hx-get="{% url 'chatbots:sessions-list' request.team.slug experiment.id %}{% if request.GET %}?{{ request.GET.urlencode }}{% endif %}"
```
snopoke, line 226: I don't think this will work because the filters update dynamically and this URL will only update on the initial page load

templates/chatbots/single_chatbot_home.html
```diff
@@ -219,11 +219,34 @@ <h3 class="font-bold text-lg">Link with {{ platform.label }}</h3>
...
+        <div class="relative">
+          {% include "chatbots/components/session_table_actions.html" %}
+          <div class="absolute top-0 right-0 flex items-center gap-2">
+            <button class="btn btn-ghost btn-sm"
+                    hx-get="{% url 'chatbots:sessions-list' request.team.slug experiment.id %}"
+                    hx-target="#sessions-table"
+                    hx-swap="innerHTML"
+                    hx-indicator="#refresh-loader"
+                    hx-disabled-elt="this"
+                    hx-disabled-elt="this"
```
snopoke, line 231:
```suggestion
```

templates/chatbots/single_chatbot_home.html
```diff
@@ -219,11 +219,33 @@ <h3 class="font-bold text-lg">Link with {{ platform.label }}</h3>
...
+              <i class="fa-solid fa-rotate htmx-hide"></i>
+            </button>
+            <span class="loading loading-spinner loading-sm htmx-indicator htmx-show" id="refresh-loader"></span>
+          </div>
+
+          <div id="sessions-table"
+               data-url="{% url 'chatbots:sessions-list' request.team.slug experiment.id %}"
+               hx-get="{% url 'chatbots:sessions-list' request.team.slug experiment.id %}"
+               hx-trigger="load"
+               hx-swap="innerHTML">
```
snopoke, line 241: I notice these attributes have been added, are they needed?

snopoke (approved):

PR #2094, opened 2025-08-22
fix font size issue

snopoke:
## Description
I noticed this issue when testing 0.4.6 on the docs site.


components/chat_widget/src/components/ocs-chat/ocs-chat.css
```diff
@@ -662,7 +662,6 @@ textarea {
 }
 
 .message-bubble-assistant .chat-markdown {
-  @apply prose prose-sm prose-gray max-w-none prose-a:no-underline hover:prose-a:underline;
```
snopoke, line 665: This is already applied to the `.chat-markdown` class. Applying it again here messes things up.

PR #2102, opened 2025-08-26
Python Evaluator

proteusvacuum:
## Description

Adds a python evaluator, allowing users to write arbitrary code to evaluate messages.
Uses a similar pattern to the code node, but the input arguments to the main function are the parameters that are available for the llm evaluator. 

The output of the python evaluator should be a dictionary whose keys become columns in the dataset output.

### Demo

image


snopoke:
@proteusvacuum FYI there is a merge conflict

snopoke (approved):

PR #2107, opened 2025-08-26
Confluence document loader

snopoke:
## Description
Enable users to load data into an indexed collection from Confluence.

## User Impact
Users can load documents from Confluence into an indexed collection

### Demo
image

image

### Docs and Changelog
https://github.com/dimagi/open-chat-studio-docs/pull/161


stephherbers (dismissed):
looks good-- is there a place for tests?

PR #2108, opened 2025-08-26
CodeMirror for prompt editor in evals 

proteusvacuum:
## Description

Note: base branch is [fr/python-evaluator](https://github.com/dimagi/open-chat-studio/tree/fr/python-evaluator) because of the changes there in `editors.js`.

This adds something similar to the Prompt Editor from pipelines. It extracts the common functionality, but I decided not to import the react component, as it was simpler to just add a new pure javascript editor this way. 

This also adds the ability to use the tab key in the other types of editor (before, hitting tab in the json editor would lose focus)

## User Impact

Screenshot from 2025-08-26 17-11-14

### DIFF
--- a/assets/javascript/editors.js
+++ b/assets/javascript/editors.js
@@ -144,6 +144,21 @@ class BaseEditor {
   }
 
   /**
+   * Get common extensions shared by all editors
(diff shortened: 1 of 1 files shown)


snopoke (approved):
nice

PR #2127, opened 2025-09-02
Add and edit evaluation message history

proteusvacuum:
## Description
Adds the ability to manually add and edit history for individual dataset messages.
This also changes the UI for the dataset create to more closely match the edit view so the templates could be shared.

## User Impact

Screenshot from 2025-09-02 11-25-58
Screenshot from 2025-09-02 11-26-40


apps/evaluations/views/dataset_views.py
```diff
@@ -51,12 +51,10 @@ class DatasetTableView(SingleTableView, PermissionRequiredMixin):
     template_name = "table/single_table.html"
 
     def get_queryset(self):
-        from django.db.models import Count
-
         return (
             EvaluationDataset.objects.filter(team=self.request.team)
             .annotate(message_count=Count("messages"))
-            .order_by("name")
+            .order_by("-created_at")
```
proteusvacuum, line 57: This will change how the datasets are ordered, but I think this ordering makes more sense?

snopoke (approved):

PR #2133, opened 2025-09-04
fixes to invitation flow for email case and additional emails

snopoke:
## Description
* Ensure emails are compared without case sensitivity.
* Match against all email addresses registered to the user, not only their primary one.

## User Impact
Allow users' to login when the case of the invitation email doesn't match their registered email. Also allow users to accept invitations sent to email addresses that aren't their primary email address.

### Docs and Changelog
https://github.com/dimagi/open-chat-studio-docs/pull/168

### DIFF
--- a/apps/teams/views/invitation_views.py
+++ b/apps/teams/views/invitation_views.py
@@ -1,5 +1,6 @@
 import uuid
 
+from anymail.utils import EmailAddress
 from django.conf import settings
 from django.contrib import messages
 from django.http import HttpResponseRedirect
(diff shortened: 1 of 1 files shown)


apps/teams/views/invitation_views.py
```diff
@@ -19,11 +20,18 @@ def accept_invitation(request, invitation_id: uuid.UUID):
...
+    user_email_matches = False
+    if request.user.is_authenticated:
+        user_email_matches = request.user.email.lower() == invitation.email.lower()
+        if not user_email_matches:
+            # If the current user's email doesn't match the invitation,
+            # check if they have an email that does match.
+            # We don't check verified emails since email verification has the
+            # same level of security as invitations.
+            user_email_matches = EmailAddress.objects.filter(
+                email__iexact=invitation.email, user__id=request.user.id
```
SmittieC, line 32: How does it happen that a user gets multiple email addresses in the first place?

snopoke: You can change your email address in your user profile and you can also manage emails at `/accounts/email/`

Issue #1970, opened 2025-07-30
Implement client key security for chat widget


snopoke:
See spec here: https://docs.google.com/document/d/1KnhOlgMcJTntq9BMxiUKZs70Hhe8ncsSBYVhvlE0Llc/edit?tab=t.rusxb4sq9p46#heading=h.7cm9b84ixznp

PR #2115, opened 2025-08-29
Ba/refactor channel dialog

barora05:
## Description
Closes #2089

### Demo
https://github.com/user-attachments/assets/63fccacd-b45f-4355-83e9-1d7f04977a4a

[original demo](https://drive.google.com/file/d/15AqccRHQ5_AXLrFyURXgPShdl23m6dQg/view?usp=drive_link)

PR #2150, opened 2025-09-08
Client key security for chat widget

stephherbers:
## Description

resolves [#1970](https://github.com/dimagi/open-chat-studio/issues/1970)

### Demo

The copy embed code scrolls so the use can get a preview of the code. This screen is shown when the channel is first created, and whenever modifed
image

### Docs and Changelog

yes and yes

### DIFF
--- a/apps/api/tests/test_chat_api_anon.py
+++ b/apps/api/tests/test_chat_api_anon.py
@@ -126,7 +126,7 @@ def test_session_poll_with_messages(api_client, session):
     }
 
 
-@pytest.skip("This no longer applies to the chat API until we have proper public access implemented.")
+@pytest.mark.skip("This no longer applies to the chat API until we have proper public access implemented.")
 @pytest.mark.django_db()
 def test_start_chat_session_requires_auth_when_not_public(team_with_users, api_client, experiment):
     url = reverse("api:chat:start-session")


snopoke (commented):
@stephherbers there have been some recent refactoring to the channel UI which will impact your changes here: https://github.com/dimagi/open-chat-studio/pull/2115

templates/chatbots/partials/channel_dialog.html
```diff
@@ -63,3 +63,114 @@ <h3 class="font-bold text-lg">
...
+            </div>
+          </div>
+        </div>
+
+        <div class="card bg-base-200 mb-4">
+          <div class="card-body">
+            <div class="flex justify-between items-center mb-3">
+              <h4 class="card-title text-base mb-0">
+                <i class="fa-solid fa-code me-1"></i>
+                Embed Code
```
SmittieC, line 127: I'm thinking it would be useful to show this section (the code) when I edit the channel again so I can find the code easily again with my channel specific values prepopulated

apps/channels/views.py
```diff
@@ -353,3 +373,12 @@ def delete_channel(request, team_slug, experiment_id: int, channel_id: int):
             "experiment": channel.experiment,
         },
     )
+
+
+@require_POST
+@login_and_team_required
+def clear_widget_success_session(request, team_slug: str, experiment_id: int):
+    """Clear the embedded widget success data from session"""
+    if "embedded_widget_success" in request.session:
```
SmittieC, line 382: Where do we set `embedded_widget_success` in the session? Why is this used?

apps/api/views/chat.py
```diff
@@ -55,13 +62,63 @@ def check_experiment_access(experiment, participant_id):
...
+def check_session_access(session, request=None):
     """
-    Check if the request has access to the session based on experiment's public API settings.
+    Check if the request has access to the session.
+    Now handles both authenticated users and embedded widgets.
 
     Returns:
         Response object if access denied, None if access allowed
     """
+    if session.experiment_channel.platform == ChannelPlatform.EMBEDDED_WIDGET:
```
SmittieC, line 110: We might want to prefetch the experiment channel for all callers of `check_session_access`

apps/channels/utils.py
```diff
@@ -1,8 +1,103 @@
...
+        if origin_domain_part.endswith("." + base_domain):
+            return True
+
+    return False
+
+
+def extract_domain_from_headers(request) -> str:
+    origin = request.headers.get("Origin")
+    if origin:
+        return origin.replace("http://", "").replace("https://", "")
```
SmittieC, line 62: You can use [urllib.parse](https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse) for these kinds of operations. Up to you though

apps/api/views/chat.py
```diff
@@ -55,13 +62,63 @@ def check_experiment_access(experiment, participant_id):
     return None
 
 
-def check_session_access(session):
+def handle_embedded_widget_auth(request, experiment_id=None, session=None):
```
SmittieC, line 65: Nitty request: May I request that this method be moved to an "auth.py" (or whatever makes sense) file instead?

apps/api/views/chat.py
```diff
@@ -55,13 +62,63 @@ def check_experiment_access(experiment, participant_id):
...
+    origin_domain = extract_domain_from_headers(request)
+    if not origin_domain:
+        raise PermissionDenied("Origin or Referer header required for embedded widgets")
+
+    if experiment_id:
+        is_valid, experiment_channel = validate_embed_key_for_experiment(
+            token=embed_key, origin_domain=origin_domain, experiment_id=experiment_id
+        )
+    elif session:
+        is_valid, experiment_channel = validate_embedded_widget_request(
```
SmittieC, line 80: Can this not also use the `validate_embed_key_for_experiment` method? We have the experiment on the session. This way, the `ExperimentChannel` query can be scoped to the correct experiment and if the channel exists, we don't have to do the `session.experiment_channel.platform != ChannelPlatform.EMBEDDED_WIDGET` check as well on the following line

apps/channels/utils.py
```diff
@@ -1,8 +1,103 @@
 from apps.channels.exceptions import ExperimentChannelException
-from apps.channels.models import ChannelPlatform
+from apps.channels.models import ChannelPlatform, ExperimentChannel
 from apps.experiments.models import Experiment
 
 
+def validate_embedded_widget_request(token: str, origin_domain: str, team) -> tuple[bool, ExperimentChannel]:
```
snopoke, line 6: This could just return the channel since the only time it returns `True` is when it also returns a channel

apps/channels/utils.py
```diff
@@ -1,8 +1,103 @@
...
+    pattern_domain_part = pattern_parts[0]
+    pattern_port = pattern_parts[1] if len(pattern_parts) > 1 else None
+
+    # If pattern specifies a port, origin must match that port exactly
+    if pattern_port is not None:
+        if origin_port != pattern_port:
+            return False
+
+    if origin_domain_part == pattern_domain_part:
+        return True
```
snopoke, line 49: I don't think this port matching is necessary. In practice we wouldn't expect to get port information outside of a dev environment.

apps/channels/utils.py
```diff
@@ -1,8 +1,103 @@
...
+    if referer:
+        domain = referer.replace("http://", "").replace("https://", "").split("/")[0]
+        return domain
+
+    return ""
+
+
+def validate_embed_key_for_experiment(
+    token: str, origin_domain: str, experiment_id: str
+) -> tuple[bool, ExperimentChannel]:
```
snopoke, line 74: Similar comment as above, this could just return the channel or None

apps/api/views/chat.py
snopoke: +1

apps/api/views/chat.py
```diff
@@ -55,13 +62,63 @@ def check_experiment_access(experiment, participant_id):
...
+    if not is_valid:
+        raise PermissionDenied("Invalid embed key or domain not allowed")
+
+    # Generate id for anon widget users
+    participant_remote_id = None
+    if hasattr(request, "data") and request.data:
+        participant_remote_id = request.data.get("participant_remote_id")
+
+    if not participant_remote_id:
+        participant_remote_id = f"embed_{uuid.uuid4()}"
```
snopoke, line 97: If `remote_id` is not provided we should leave it blank.

apps/api/views/chat.py
```diff
@@ -55,13 +62,63 @@ def check_experiment_access(experiment, participant_id):
...
+    else:
+        raise ValueError("Either experiment_id or session must be provided")
+
+    if not is_valid:
+        raise PermissionDenied("Invalid embed key or domain not allowed")
+
+    # Generate id for anon widget users
+    participant_remote_id = None
+    if hasattr(request, "data") and request.data:
+        participant_remote_id = request.data.get("participant_remote_id")
```
snopoke, line 94: This doesn't seem to belong in this method

apps/api/views/chat.py
snopoke: Generally agree with @SmittieC. I think this should just return the channel (or set the channel on the request). Returning the boolean is equivalent to the channel being returned / set on the request.

Extracting the remote ID should be handed directly in the view or in a separate utility.

apps/api/views/chat.py
```diff
@@ -231,32 +293,47 @@ def chat_start_session(request):
...
-            platform=ChannelPlatform.API,
+            platform=platform,
             defaults={"user": user, "remote_id": remote_id},
         )
     else:
-        participant = Participant.create_anonymous(team, ChannelPlatform.API, remote_id)
+        participant = Participant.create_anonymous(team, platform, remote_id)
+        if is_embedded:
+            participant.identifier = participant_id
+            participant.save(update_fields=["identifier"])
```
snopoke, line 336: Rather than doing this you could pass the prefix to the method and let it handle the rest.

```python
participant = Participant.create_anonymous(team, platform, remote_id, prefix="embed")
```

apps/api/views/chat.py
```diff
@@ -272,16 +349,18 @@ def chat_start_session(request):
         if participant_data.data.get("name") != name:
             participant_data.data["name"] = name
             participant_data.save(update_fields=["data"])
-    api_channel = ExperimentChannel.objects.get_team_api_channel(team)
+
+    metadata = {Chat.MetadataKeys.EMBED_SOURCE: request.headers.get("referer", None)}
+    if is_embedded:
+        metadata["embedded_widget"] = True
```
snopoke, line 355: Do you have a future use case for this?

apps/api/views/chat.py
```diff
@@ -272,16 +349,18 @@ def chat_start_session(request):
...
+        metadata["origin_domain"] = extract_domain_from_headers(request)
```
snopoke, line 356: Fine to add this but you should create a new constant in `Chat.MetadataKeys`

apps/channels/forms.py
```diff
@@ -532,3 +533,92 @@ class CommCareConnectChannelForm(ExtraFormBase):
...
+
+        domains = []
+        for line in domains_text.split("\n"):
+            domain = line.strip()
+            if domain:
+                if not re.match(
+                    r"^(\*\.)?[a-zA-Z0-9]([a-zA-Z0-9-]*[a-zA-Z0-9])?(\.[a-zA-Z0-9]([a-zA-Z0-9-]*[a-zA-Z0-9])?)*(:\d+)?$",
+                    domain,
+                ):
+                    raise forms.ValidationError(f"Invalid domain format: {domain}")
```
snopoke, line 581: You can use `django.core.validators.validate_domain_name`

apps/channels/forms.py
```diff
@@ -532,3 +533,92 @@ class CommCareConnectChannelForm(ExtraFormBase):
...
+        """Save widget data and set success message"""
+        widget_token = self.cleaned_data["widget_token"]
+        embed_code = self._generate_embed_code(channel, widget_token)
+        channel.extra_data.update(
+            {
+                "widget_token": widget_token,
+                "allowed_domains": self.cleaned_data["allowed_domains"],
+            }
+        )
+        channel.save()
```
snopoke, line 608: this is already handled in the channel form: https://github.com/dimagi/open-chat-studio/blob/33f64319ceb4759d63146f490c34276dfbfd01d1/apps/channels/forms.py#L69-L72

apps/channels/forms.py
```diff
@@ -532,3 +533,92 @@ class CommCareConnectChannelForm(ExtraFormBase):
...
+
+
+class EmbeddedWidgetChannelForm(ExtraFormBase):
+    """Form for creating embedded chat widget channels"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Initialize widget data attributes
+        self.widget_token = None
+        self.embed_code = None
```
snopoke, line 545: The way we've handled this before is to have extra fields in the form which are hidden when creating the form and disabled when editing the form: https://github.com/dimagi/open-chat-studio/blob/33f64319ceb4759d63146f490c34276dfbfd01d1/apps/channels/forms.py#L169-L174

That should avoid all the custom code in the views and template.

Another option is to use a custom form template like we do for document sources: https://github.com/dimagi/open-chat-studio/blob/a94707d0c567fa1326d34fbc302dab3105070dcc/templates/documents/document_source_form_dialog.html#L23-L27

In this case I think a combination of both is probably best. You can use a special form field for the widget token and a custom template to allow you to show the embed code.

apps/channels/forms.py
```diff
@@ -532,3 +533,92 @@ class CommCareConnectChannelForm(ExtraFormBase):
...
+                "widget_token": widget_token,
+                "allowed_domains": self.cleaned_data["allowed_domains"],
+            }
+        )
+        channel.save()
+        self.widget_token = widget_token
+        self.embed_code = embed_code
+        self.success_message = f"Embedded widget channel created successfully! Widget token: {widget_token}"
+
+    def _generate_embed_code(self, channel: ExperimentChannel, token: str) -> str:
```
snopoke, line 613: Rather use the `widget.html` template

apps/channels/forms.py
```diff
@@ -532,3 +533,92 @@ class CommCareConnectChannelForm(ExtraFormBase):
...
+        """Generate the embed code for the widget"""
+        base_url = getattr(settings, "SITE_URL", "https://chatbots.dimagi.com")
+        embed_code = f'''<!-- Open Chat Studio Embedded Widget -->
+<script type="module" src="{base_url}/static/js/open-chat-studio-widget.esm.js"></script>
```
snopoke, line 617: this `src` attribute should be the unpkg.com url

apps/channels/forms.py
```diff
@@ -532,3 +533,92 @@ class CommCareConnectChannelForm(ExtraFormBase):
...
+<script nomodule src="{base_url}/static/js/open-chat-studio-widget.js"></script>
+<open-chat-studio-widget
+    chatbot-id="{channel.experiment.public_id}"
+    api-base-url="{base_url}"
```
snopoke, line 621: No need to set this. Rather use the default value.

templates/chatbots/partials/channel_dialog.html
stephherbers: the dialog is shown when updating the channel as well 👍

apps/channels/tests/test_embedded_widget_channel.py
```diff
@@ -0,0 +1,152 @@
...
+    assert len(token1) == 32
+    assert len(token2) == 32
+
+    allowed_chars = string.ascii_letters + string.digits + "-_"
+    assert all(c in allowed_chars for c in token1)
+    assert all(c in allowed_chars for c in token2)
+
+
+@pytest.mark.django_db()
+def test_embedded_widget_integration_with_existing_channels():
```
snopoke, line 140: I don't think this is a useful test

apps/channels/tests/test_embedded_widget_channel.py
```diff
@@ -0,0 +1,152 @@
...
+                ChannelPlatform.EMBEDDED_WIDGET, "existing_token_123456789012345678", new_experiment
+            )
+
+    def test_platform_choices_include_embedded_widget(self):
+        choices_dict = dict(ChannelPlatform.choices)
+        assert "embedded_widget" in choices_dict
+        assert choices_dict["embedded_widget"] == "Embedded Widget"
+
+
+def test_form_token_generation_is_secure():
```
snopoke, line 119: I don't think this is a useful test

apps/channels/tests/test_embedded_widget_channel.py
```diff
@@ -0,0 +1,152 @@
...
+        )
+        new_experiment = ExperimentFactory()
+
+        # Should raise exception for duplicate token usage
+        with pytest.raises(ChannelAlreadyUtilizedException):
+            ExperimentChannel.check_usage_by_another_experiment(
+                ChannelPlatform.EMBEDDED_WIDGET, "existing_token_123456789012345678", new_experiment
+            )
+
+    def test_platform_choices_include_embedded_widget(self):
```
snopoke, line 113: I don't think this is a useful test

apps/channels/tests/test_embedded_widget_channel.py
```diff
@@ -0,0 +1,152 @@
...
+    def test_embedded_widget_platform_available_in_dropdown(self):
+        team = TeamWithUsersFactory()
+        available_platforms = ChannelPlatform.for_dropdown(used_platforms=set(), team=team)
+        assert ChannelPlatform.EMBEDDED_WIDGET in available_platforms
+        assert available_platforms[ChannelPlatform.EMBEDDED_WIDGET] is True
+
+    def test_embedded_widget_channel_identifier_key(self):
+        assert ChannelPlatform.EMBEDDED_WIDGET.channel_identifier_key == "widget_token"
+
+    def test_create_embedded_widget_channel(self):
```
snopoke, line 86: I don't think this is a useful test

apps/channels/tests/test_embedded_widget_channel.py
```diff
@@ -0,0 +1,152 @@
...
+
+@pytest.mark.django_db()
+class TestEmbeddedWidgetChannelModel:
+    def test_embedded_widget_platform_available_in_dropdown(self):
+        team = TeamWithUsersFactory()
+        available_platforms = ChannelPlatform.for_dropdown(used_platforms=set(), team=team)
+        assert ChannelPlatform.EMBEDDED_WIDGET in available_platforms
+        assert available_platforms[ChannelPlatform.EMBEDDED_WIDGET] is True
+
+    def test_embedded_widget_channel_identifier_key(self):
```
snopoke, line 83: I don't think this is a useful test

apps/channels/tests/test_embedded_widget_channel.py
```diff
@@ -0,0 +1,152 @@
...
+        ],
+    )
+    def test_match_domain_pattern(self, origin_domain, allowed_pattern, should_match):
+        result = match_domain_pattern(origin_domain, allowed_pattern)
+        assert result == should_match
+
+
+@pytest.mark.django_db()
+class TestEmbeddedWidgetChannelModel:
+    def test_embedded_widget_platform_available_in_dropdown(self):
```
snopoke, line 77: Duplicates `test_models.test_available_channels`

apps/channels/views.py
snopoke, line 1: I think the changes to this file can be reverted

templates/chatbots/partials/channel_dialog.html
snopoke, line 1: If you revert the view changes you'll also need to revert these changes

templates/experiments/share/widget.html
```diff
@@ -1,7 +1,7 @@
 {# The rendered content of this template must be suitable for a Javascript string constant: #}
 {# - escape script tags #}
 <!-- place this in the <head> section -->
-\<script type="module" src="https://unpkg.com/open-chat-studio-widget@0.4.1/dist/open-chat-studio-widget/open-chat-studio-widget.esm.js"\>\<\/script\>
+\<script type="module" src="https://unpkg.com/open-chat-studio-widget@0.4.8/dist/open-chat-studio-widget/open-chat-studio-widget.esm.js"\>\<\/script\>
```
snopoke, line 4: @stephherbers we'll need to release a new version and update this - maybe we should split out the widget changes into a separate PR that we can release before these changes?

stephherbers:
NOTE: please do not merge before this pr: https://github.com/dimagi/open-chat-studio/pull/2330 and a release of the widget has been made

PR #2165, opened 2025-09-11
upgrade python to 3.13

snopoke:
## Description
This required upgrading the following packages;
* psycopg
* twilio & some of its dependencies
* transformers & some of its dependencies
* billiard (used by celery)

It also required adding `audioop-lts` since `audioop` doesn't support python >3.12.

## User Impact
None

### Tests conducted
- [x] run local dev server & celery
- [x] build docker image
- [x] run docker image
- [x] run celery in docker (worker and beat)


pyproject.toml
```diff
@@ -78,6 +78,7 @@ dependencies = [
     "markitdown[pdf, docx, xlsx, xls, outlook, pptx]",
     "langchain-mcp-adapters",
     "atlassian-python-api>=4.0.7",
+    "audioop-lts>=0.2.2",
```
SmittieC, line 81: Should the other libraries (twilio etc) not be pinned here as well?

snopoke: They could be but it's fine for their versions to be managed in the `uv.lock` file

PR #2166, opened 2025-09-11
force migration of all experiments to chatbots

snopoke:
## Description
This will be a no-op for Dimagi's production instance but will force all other instances to migrate.

This also removes all usages of the `flag_chatbots` feature flag (effectively forcing it to be enabled for all teams on all environments).

This is needed before we can start making backwards incompatible changes.

### Docs and Changelog
TODO


snopoke:
> LGTM. One question though: Should we not do the same for the `pipelines-v2` flag?

I was going to do that in a separate PR but now that I think about it we need to do it at the same time.

SmittieC (approved):
LGTM

PR #2187, opened 2025-09-17
Small trends update

SmittieC:
## Description

See the user description

## User Impact

- Users will not see trends when the tracing flag is disabled
- When a user click on a specific chatbot's trend graph, it takes them to the traces view, filtered by that specific chatbot. It's a quick way to get to the traces of only that chatbot

### Demo

N/A

### Docs and Changelog

N/A

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -1,6 +1,7 @@
 import base64
 import logging
 import secrets
+import urllib
 import uuid
 from datetime import UTC, datetime
 from functools import cached_property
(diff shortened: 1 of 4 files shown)


apps/experiments/models.py
```diff
@@ -891,6 +892,15 @@ def get_trend_data(self) -> tuple[list, list]:
         cache.set(self.trends_cache_key, trend_data, settings.EXPERIMENT_TREND_CACHE_TIMEOUT)
         return trend_data
 
+    def traces_url(self) -> str:
+        """
+        Returns a URL to the traces page, filtered to show only traces for this experiment.
+        """
+        query_params = urllib.parse.urlencode(
+            {"filter_0_column": "experiment", "filter_0_operator": "any of", "filter_0_value": f"[{self.id}]"}
```
SmittieC, line 900: I want to write a mini utility method or something for these kinds of operations once the refactor PR is in

Issue #2071, opened 2025-08-18
Tracing | Add more spans


SmittieC:
See the [OCS tracer](https://github.com/dimagi/open-chat-studio/blob/8764af8dc623f45a15b7235f61701c4875ab03e6/apps/service_providers/tracing/ocs_tracer.py) and how it's used in `channels.py`. Currently the only span that is shown is the "process_message" one. This isn't very useful.

We can now add and use nested spans to allow users to see what steps are being performed as the user message are being processed by various parts of the system. 

## The goal
Identify areas and add nested spans (and/or sibling spans) that allows the user to see what is happening with the user message. Also update the UI to show the captured input/outputs/metadata for each span (if any).

PR #2188, opened 2025-09-17
Tracing | OCS tracer callback and UI updates

SmittieC:
## Description

Part of [this ticket](https://github.com/dimagi/open-chat-studio/issues/2071).

- Create a callback manager that creates spans using the OCS tracer
- Update UI to better show span info. This includes input, output and metadata (I copied Langchain's approach)
- Fixed the trends graph to not only show the deployed version's trends, but the trends across all versions of that chatbot. This will clearly show how one version might have broken or fixed something.
- Added a method on the `ColumnFilterData` data structure to return a url safe query param of the filter. This is used to generate a the url to a specific experimen's traces

TODO:
- Some query optimization

## User Impact

More info available to the user

### Demo

image

image

### Docs and Changelog

N/A


apps/web/dynamic_filters/datastructures.py
```diff
@@ -17,13 +17,9 @@ class ColumnFilterData:
...
 
-    def as_query_string(self) -> str:
-        return urlencode(
-            {
-                "filter_0_column": self.column,
-                "filter_0_operator": self.operator,
-                "filter_0_value": self.value,
-            }
+    def as_query_string(self, filter_number=0) -> str:
+        return "filter_{fn}_column={col}&filter_{fn}_operator={op}&filter_{fn}_value={val}".format(  # noqa: UP032
```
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/2188/commits/2335e2bab23443f3feca9414828d3e091f52a7c3

SmittieC:
Opening this up for review while I look into one small issue I noticied and do some query optimization

apps/experiments/tests/test_models.py
```diff
@@ -1069,6 +1069,31 @@ def test_archived_experiments_are_filtered_out(self):
...
+        version1 = experiment.create_new_version()
+        version2 = experiment.create_new_version()
+
+        curr_time = timezone.now()
+        Trace.objects.create(
+            experiment=experiment, team=experiment.team, status=TraceStatus.SUCCESS, timestamp=curr_time, duration=1
+        )
+        # Trace for version1
+        Trace.objects.create(
+            experiment=version1, team=experiment.team, status=TraceStatus.ERROR, timestamp=curr_time, duration=1
```
snopoke, line 1084: Does this ever happen? I mean do we attach spans to experiment versions?

apps/trace/views.py
```diff
@@ -49,11 +53,30 @@ class TraceDetailView(LoginAndTeamRequiredMixin, DetailView, PermissionRequiredM
     permission_required = "trace.view_trace"
 
     def get_queryset(self):
-        return Trace.objects.select_related(
-            "experiment", "session", "participant", "input_message", "output_message"
-        ).filter(team=self.request.team)
+        return (
+            Trace.objects.select_related("experiment", "session", "participant", "input_message", "output_message")
+            .prefetch_related("spans")
```
snopoke, line 58: is this prefetch necessary if we're doing a separate query in `get_context_data` to fetch the spans?

apps/experiments/tests/test_models.py
snopoke: Yikes, looks like we do actually. That's not great. I think we should probably always use the working version - like we do for sessions.

Probably a separate PR for that though.

snopoke (approved):

apps/trace/views.py
SmittieC: Thanks for catching this. Not necessary anymore

apps/experiments/tests/test_models.py
SmittieC: Hmm yeah maybe we should. We could store the version of the experiment in the metadata then

SmittieC: Thanks. Here's a carrot :carrot:

snopoke:
@SmittieC sorry my other filter changes have caused some conflicts here

snopoke (approved):

Issue #2147, opened 2025-09-08
Restrict Tag Creation to Super Admins


stephherbers:
Full context here: https://docs.google.com/document/d/1onSyJCyYwxciIHrIX22YPeIBVyn_pCEF8Zi3tocK5Lw/edit?tab=t.0#heading=h.bxcqojrzzw28 

 Unless the user is a super admin, they should only be able to select one of the tag values and should not be permitted to enter free text.

PR #2191, opened 2025-09-17
Super Admin: tag permissions

stephherbers:
## Description

resolves: [#2147](https://github.com/dimagi/open-chat-studio/issues/2147)

Only super admins can create, modify, or delete tags. Non-super admins can add tags if they have that permission, and can search of them, but they can no longer see the "Add tag...." functionality

Test performed:
As Super Admin:
- creating, editing, deleting tag functionally all remains the same as prior to this change

As Non-Super admin with add permission:
- can add tags but cannot create them
- can search for tags when adding them, but pressing enter for a tag that doesn't exist does not create a tag
- cannot edit or delete tags
- manually entering URLs cannot grant access

As Non-Super admin without add permission:
- cannot add tags as before

## User Impact

non-super admins are still able to view tags and add them if they were able to before, just not create new ones, edit or delete them

### Docs and Changelog

yes to changelog

### DIFF
--- a/apps/annotations/tables.py
+++ b/apps/annotations/tables.py
@@ -8,18 +8,22 @@ from apps.generics import actions
 from .models import Tag
 
 
-def display_condition(request: HttpRequest, record: Context) -> bool:
-    return not record.is_system_tag
+def edit_display_condition(request: HttpRequest, record: Context) -> bool:
+    return not record.is_system_tag and request.user.has_perm("annotations.change_tag")
+
+
+def delete_display_condition(request: HttpRequest, record: Context) -> bool:
+    return not record.is_system_tag and request.user.has_perm("annotations.delete_tag")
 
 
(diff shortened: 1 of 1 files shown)


apps/teams/backends.py
```diff
@@ -183,7 +183,7 @@ def get_permissions(self):
         [
             AppPermSetDef("experiments", ALL),
             AppPermSetDef("bot_channels", ALL),
-            AppPermSetDef("annotations", ALL),
+            AppPermSetDef("annotations", [VIEW]),
```
snopoke, line 186: To get the write permissions for creating tags you'll need to add a `ModelPermSetDef`:

```python
ModelPermSetDef("annotations", "customtaggeditem", ALL)
ModelPermSetDef("annotations", "usercomment", ALL)
```

PR #2198, opened 2025-09-18
filters refactor: include options, operators, labels etc in column schema

snopoke:
## Description
This changes how filter column schemas are defined and unifies them with the `ColumnFilter` classes.

A column filter should now include all the data required to display the filter including:
* query_param
* label
* options (optional)
* operators
* type

## User Impact
None


SmittieC (approved):
Cool. LGTM

Issue #2146, opened 2025-09-08
Link tags to a filtered view of the chatbot session page from dashboard


stephherbers:
Full context here: https://docs.google.com/document/d/1onSyJCyYwxciIHrIX22YPeIBVyn_pCEF8Zi3tocK5Lw/edit?tab=t.0#heading=h.y1wf6zqxsiam 

Enable message-level tag summaries on the OCS dashboard (e.g. to surface the subset of messages tagged “unsafe”)
From the dashboard, create a clickable link to the bot messages tagged unsafe. This should open the experiment session page with a filter set to ‘unsafe’: https://chatbots.dimagi.com/a/Shujaaz/experiments/e/652/?filter_0_column=tags&filter_0_operator=any+of&filter_0_value=%5B%22Unsafe%22%5D#allsessions 

this involves creating a clickable link with the current filters of the dashboard if possible, if not then it should remain unclickable

_Limitations: this can only apply when there is one chatbot filtered in the dashboard_

SmittieC:
What if we build a new view showing all matching experiments + sessions. I'm imagining something similar to what the participant view looks like:

<img width="1150" height="567" alt="Image" src="https://github.com/user-attachments/assets/e33a8d98-2393-4743-b978-e9907de3f16a" />

We list all experiments, then when you click on one, it opens up and shows all matching sessions.

snopoke:
I was thinking the same thing - but just a single table (basically what we already when creating datasets that also shows the chatbot in the table).

PR #2215, opened 2025-09-24
All sessions view with filters

SmittieC:
## Description

This is part 1 of this issue: https://github.com/dimagi/open-chat-studio/issues/2146

This adds a new "all sessions" view with filters. This view shows all sessions for this team. This table is the same one used to render the sessions per chatbot, but only with a new column showing the chatbot it belongs to.

## User Impact

All users will see

### Demo

image


snopoke (approved):
Are you going to do the linking from the dashboard in a separate PR?

PR #2217, opened 2025-09-24
return to celery connection pooling

snopoke:
## Description
Removing the connection pool was worse:

image

### DIFF
--- a/gpt_playground/settings.py
+++ b/gpt_playground/settings.py
@@ -226,7 +226,7 @@ else:
     DATABASES["default"].pop("CONN_MAX_AGE", None)
     db_options["pool"] = {
         "min_size": env.int("DJANGO_DATABASE_POOL_MIN_SIZE", default=2),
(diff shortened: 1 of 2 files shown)


gpt_playground/settings.py
```diff
@@ -226,7 +226,7 @@
     DATABASES["default"].pop("CONN_MAX_AGE", None)
     db_options["pool"] = {
         "min_size": env.int("DJANGO_DATABASE_POOL_MIN_SIZE", default=2),
-        "max_size": env.int("DJANGO_DATABASE_POOL_MAX_SIZE", default=20),
+        "max_size": env.int("DJANGO_DATABASE_POOL_MAX_SIZE", default=25),
```
snopoke, line 229:
```suggestion
        "max_size": env.int("DJANGO_DATABASE_POOL_MAX_SIZE", default=35),
```

PR #2218, opened 2025-09-24
Link from dashboard to all sessions page

### LINKED ISSUE #2146: Link tags to a filtered view of the chatbot session page from dashboard

SmittieC:
## Description

Resolves #2146 to a large extend.

Q: What do you mean by "to a large extend"?
A: Most of the filters applied on the dasboard, are also pre-applied when redirecting to the all sessions view. The _participants_ and _granularity_ filters are **not**, since their values are inherently incompatible with those of the dynamic filters.

Q: Will this confuse users?
A: In all honesty, it can. They should see that those filters are not applied and apply the respective filters accordingly though. We **could** also just apply the selected tag filter, but then the user would have to apply the rest. :shrug: 

> [!NOTE]  
> The approach in this PR is a temporary workaround.

### A workaround to what?
The issue comes down to the fact that the dashboard does not use the new dynamic filters framework, so the query params that the dasboard produces are incompatible with those that the dynamic filters expects. We have two options here:
1. Rebuild the dashboard using the dynamic filters framework so that it produces the correc query params (long term correct solution)
2. Map the filter params to dynamic filter params where possible.

I opted for approach #2.

## User Impact

User will be able to click on a tag, which then takes them to the "all sessions" view where the filters they used on the dashboard (date range, chatbot etc) will be pre-applied to find the set of session in which this tag was found.

### Demo

https://www.loom.com/share/7bce9f44ce4d4819a077ba4738ce0be9?sid=85b0966f-363b-4efc-b6fa-8ca9f3cfeb7c

### Docs and Changelog

Pending

### DIFF
--- a/assets/javascript/dashboard/main.js
+++ b/assets/javascript/dashboard/main.js
@@ -787,10 +787,12 @@ function dashboard() {
                 let operator = "any of";
                 if (key === "start_date") {
                     operator = "after";
-                    parsedValue = value; // keep date as string
+                    // To account for filter mismatches, we add one day to the start date
+                    parsedValue = this.getDayAfter(value); // keep date as string
                 } else if (key === "end_date") {
                     operator = "before";
-                    parsedValue = value; // keep date as string
+                    // To account for filter mismatches, we add one day to the end date
+                    parsedValue = this.getDayAfter(value); // keep date as string
                 } else if (key === "date_range") {
                     operator = "range";
                     parsedValue = value + "d";
@@ -804,6 +806,13 @@ function dashboard() {
             });
             return allSessionsUrl + "?" + urlParams.toString();
         },
+
+        getDayAfter(dateString) {
+            let date = new Date(dateString);
+            date.setDate(date.getDate() + 1);
+            return date.toISOString().split('T')[0];
+        },
+
         // Cleanup
         destroy() {
             if (this.refreshTimeout) {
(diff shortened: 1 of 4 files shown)


assets/javascript/dashboard/main.js
```diff
@@ -759,7 +759,67 @@ function dashboard() {
...
+                    return;
+                }
+                
+                let parsedValue = "";
+                // Map the filter keys to the expected query params in the all sessions view
+                let keyMapped = dynamicFilterParamMapping[key] || key;
+                let operator = "any of";
+                if (key === "start_date") {
+                    operator = "after";
+                    // To account for filter mismatches, we add one day to the start date
```
SmittieC, line 798:
```suggestion
                    // To account for filter mismatches, we subtract one day from the start date
```

snopoke (approved):
one small comment but LGTM

PR #2222, opened 2025-09-25
Make connect channel participant identifiers lower case

SmittieC:
## Description

Fixes the likes of https://dimagi.sentry.io/issues/6899765887/?alert_rule_id=14063560&alert_type=issue&notification_uuid=895720d3-e699-4bb7-8914-953518595062&project=4505001320316928&referrer=issue_alert-slack

### What happened?
Connect treats connect ids as case insensitive, but OCS did not. So when the first call to register the participant arrived (with uppercase identifier), OCS created the particpant, registered the participant at connect and received a channel ID for this participant. We then stored it in the participant data for this participant.

When the second call came in 28 mins later, the identifier was lower case, so we created a new participant, registered them at connect, but connect returned the same channel ID as before (since it already exists in that world), so we then linked this same channel ID to the new participant's data.

In the failing call, we assume that there's a single participant data record with the received channel ID (which is correct), but OCS now created separate participant data records with the same channel ID

## User Impact

N/A

### Demo

N/A. See the test

### Docs and Changelog

Pending

### DIFF
--- a/apps/api/views/participants.py
+++ b/apps/api/views/participants.py
@@ -89,8 +89,14 @@ def _update_participant_data(request):
     serializer.is_valid(raise_exception=True)
 
     identifier = serializer.data["identifier"]
+
     platform = serializer.data["platform"]
     team = request.team
+
+    if platform == ChannelPlatform.COMMCARE_CONNECT:
+        # CommCare Connect IDs are case insensitive, so we store them in lowercase
+        identifier = identifier.lower()
+
     participant, _ = Participant.objects.get_or_create(identifier=identifier, team=team, platform=platform)
 
     # Update the participant's name if provided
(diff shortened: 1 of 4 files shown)


apps/api/views/participants.py
```diff
@@ -89,8 +89,14 @@ def _update_participant_data(request):
     serializer.is_valid(raise_exception=True)
 
     identifier = serializer.data["identifier"]
+
     platform = serializer.data["platform"]
     team = request.team
+
+    if platform == ChannelPlatform.COMMCARE_CONNECT:
+        # CommCare Connect IDs are case insensitive, so we store them in lowercase
+        identifier = identifier.lower()
```
snopoke, line 98: nit: instead of doing this here I would push it into the `ChannelPlatform` class:

```python
identifier = ChannelPlatform(platform).normalize_identifier(identifier)
```

snopoke (approved):

apps/api/views/participants.py
SmittieC: Updated

PR #2236, opened 2025-09-29
Ensure connect id is always lower case

SmittieC:
## Description

I found another place where the connect id can be upper case. Since we now always expect the lower case version of the connect id, we should always make sure to use the lower case.

## User Impact

Users whose phone sends upper case connect ids should continue to work as normal.

### Demo

N/A

### Docs and Changelog

Pending


snopoke:
The bug that won't die 🐞

PR #2237, opened 2025-09-29
Ensure only lower cased commcare_connect participants

SmittieC:
## Description

We should have included a data migration on Thursday when we forced lower cased commare connect participant identifiers. Rather late than never I guess.

Data migration ensure there are only lower cased commcare_connect channel participant identifiers.
This should resolve the likes of https://dimagi.sentry.io/issues/6789088297/?project=4505001320316928&query=is%3Aunresolved&referrer=issue-stream.

## Is this safe?
It should be. The participant data and system metadata is the same for those with the lower and upper cased identifiers. There's 1 participant whose data field between the two isn't the same (but the system metadata is).

Field 1 - updated at 17 Aug 17:12:28 (this is from the participant with the upper cased identifier):
```
{'ecdcount_id': None, 'external_id': '<some id>'}
```

Field 2 - updated at 17 Aug 17:48:57 (this is from the participant with the lower cased identifier)
```
{'ecdcount_id': None, 'external_id': '<some id, but the same as above>', 'scenario_checkpoint': 2}
```
Since sessions will be ported to the participant with the lower cased identifier, the data from field 2 will be left.

## User Impact

Some connect users who cannot chat to bots, should be able to chat to bots again.

### Demo

N/A see the test

### Docs and Changelog

Pending


apps/experiments/tests/test_data_migrations.py
```diff
@@ -0,0 +1,49 @@
...
+    session4.refresh_from_db()
+    session5.refresh_from_db()
+    session6.refresh_from_db()
+
+    # Check that sessions 1, 2, and 3 are all linked to the same participant (the one with the lowercased identifier)
+    assert session1.participant_id == session2.participant_id == session3.participant_id
+    # Ensure the uppercase participant has been deleted
+    assert Participant.objects.filter(team=team, identifier="ABC123").exists() is False
+
+    assert session4.participant == session5.participant
```
snopoke, line 45: Also check that these are different from the other ones with the same ID in another team:
```suggestion
    assert session4.participant == session5.participant
    assert session4.participant != session1.participant
```

stephherbers (approved):

apps/experiments/tests/test_data_migrations.py
SmittieC: Good point. See https://github.com/dimagi/open-chat-studio/pull/2237/commits/5373a9eee24c132bcd746163f881c0e6e36dcfea

snopoke (approved):

Issue #2129, opened 2025-09-03
[evals] Filter messages when cloning from sessions


snopoke:
When cloning sessions into a dataset, give the user the option to only include messages that match the relevant filters.

e.g.
* User filters the sessions by tag
  * This shows sessions with the tag as well as sessions that contain messages with the tag
* User can import ALL messages from the selected sessions OR only messages matching the tag filter.

### Relevant filters
* tag
* date
* versions

PR #2245, opened 2025-09-30
Filter messages cloning

proteusvacuum:
## Description

The user can now choose to either add all messages from a session matching a filter, or only the messages that match that filter.

Since Evaluation datasets expect a (Human, AI) pair, we also include the corresponding other message from this pair in the dataset.

Fixes: https://github.com/dimagi/open-chat-studio/issues/2129

### Demo

https://github.com/user-attachments/assets/1431cde8-5541-4898-b10d-1489548afbd3

### Docs and Changelog

https://github.com/dimagi/open-chat-studio-docs/pull/183


apps/evaluations/tables.py
```diff
@@ -216,13 +216,22 @@ class Meta:
...
         orderable=False,
-        attrs={
-            "input": {"class": "checkbox checkbox-primary session-checkbox", "@change": "updateSelectedSessions()"},
-            "th__input": {"style": "display: none;"},  # Hide the select all checkbox in header
+        extra_context={
+            "css_class": "checkbox checkbox-primary session-checkbox",
+            "js_function": "updateSelectedSessions()",
+        },
+    )
+    clone_filtered_only = TemplateColumn(
```
SmittieC, line 228: Does it make sense to configure this per session, or should it not be configured once for all sessions? Also, what if both `All Messages` and `Filtered Messages` are selected?

apps/evaluations/utils.py
```diff
@@ -165,3 +168,55 @@ def _clean_context_field_name(field_name):
...
+
+    def get_next_message(message):
+        current_index = message_lookup[message.id]
+        return all_chat_messages[current_index + 1] if current_index < len(all_chat_messages) - 1 else None
+
+    all_messages = set()
+
+    # Handle AI seed message
+    has_ai_seed_message = False
+    if queryset_list[0].is_first_message and queryset_list[0].is_ai_message:
```
SmittieC, line 200: Could you not annotate the queryset with `is_first_message` and inside the loop (line 205) check if it is the first message. This will allow us to use an iterator for the loop incase the queryset is huge, save us from having to fetch all records on line 182 and also not require an extra DB query to check if the message is the first one.

apps/evaluations/tables.py
proteusvacuum: Good question! I interpreted the spec

> User can import ALL messages from the selected sessions OR only messages matching the tag filter.

to be selectable for each session, but I can see how you could interpret it as a switch on the whole cloning operation. Maybe @snopoke has opinions? If the switch is at the clone level, it does remove a lot of complexity here. 

Also, good catch - I'll update to only allow a single checkbox to be checked. I forgot to add that.

apps/evaluations/utils.py
```diff
@@ -165,3 +168,64 @@ def _clean_context_field_name(field_name):
...
+            continue
+
+        all_message_ids.add(message.id)
+
+        # For AI messages, add previous human message
+        if message.message_type == ChatMessageType.AI:
+            if message.prev_message_id and message.prev_message_type == ChatMessageType.HUMAN:
+                all_message_ids.add(message.prev_message_id)
+            else:
+                raise ValueError(f"AI message {message.id} has no corresponding human message")
```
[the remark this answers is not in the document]
SmittieC: Hmm, thinking about it more, I think we should probably just discard messages that are "extra" AI messages. So this method should then output only valid human-ai message pairs. Unless for scheduled messages you want to evaluate the message that triggered the scheduled AI message, but I'm guessing that's not needed for now.

> are there other cases where an AI message following a human message wouldn't be in direct response to the human message?

Not that I can think of no

proteusvacuum:
Thinking about this more, I see an issue with the history generation when you filter the sessions (you want the history of all messages included, not just the filtered messages). I'm going to review how that works and will update here.

apps/evaluations/utils.py
```diff
@@ -225,3 +229,123 @@ def make_message_pairs_from_queryset(queryset: QuerySet) -> list[ChatMessage]:
...
+                "summary": getattr(msg, "summary", None),
+            }
+            for msg in messages_
+        ]
+
+    new_messages = []
+    for session_id, target_message_ids in message_ids_per_session.items():
+        target_ids_set = set(target_message_ids)
+
+        # We need to get all the messages in the session to properly compile the history
```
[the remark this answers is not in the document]
SmittieC: I sat with this exact thing while doing https://github.com/dimagi/open-chat-studio/pull/2281/commits/63f7b7b77a62bd7b3017ee5fcca8cbed1219ce4b (before I saw your comment about using this method). I think we're doomed to fetch all sadly.

SmittieC (approved):
Nice!

PR #2264, opened 2025-10-01
Update pipelines to access participant data and session state from the graph state

snopoke:
## Description
This is a fairly major update to how session state and participant data are managed withing a pipeline. Previously they were accessed directly from the database each time and saved to the DB on each update. 

With the changes in this PR they are accessed via the `PipelineState` and any updates are made directly to the state. At the end of a pipeline run the PD and SS are saved to the DB if necessary.

The primary goal for this is to allow evals to inject arbitrary PD and SS without having to create new PD instances for each eval message.

A bonus is that there are less DB queries during the pipeline run.

#### Reference docs

* https://langchain-ai.github.io/langgraph/how-tos/memory/add-memory/#write-short-term
* https://langchain-ai.github.io/langgraph/reference/agents/#langgraph.prebuilt.tool_node.InjectedState
* https://langchain-ai.github.io/langgraph/how-tos/tool-calling/#short-term-memory

## User Impact
None


snopoke:
@coderabbitai review

PR #2278, opened 2025-10-06
Experiment Cleanup: experiment home redirects

stephherbers:
## Description

removes conditional or replaces experiment home redirect entirely with chatbot home

tested on these relevant pages to ensure they still work as expected

## User Impact

users will not be redirected to experiment_home and rather chatbot home

### Docs and Changelog

no

### DIFF
--- a/apps/channels/views.py
+++ b/apps/channels/views.py
@@ -246,18 +246,15 @@ class BaseChannelDialogView(View):
         return context
 
     def get_success_url(self):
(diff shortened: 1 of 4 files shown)


barora05:
This origin flag is still used in experiment_versions_actions.html. If we are removing the experiment nav option, do we need the origin flag in experiment_versions_actions.html? If not, I think we can also remove the OriginDetectionMiddleware.

PR #2293, opened 2025-10-10
Experiment cleanup: redirect to single_chatbot_home

stephherbers:
## Description

rip to "experiments:single_experiments_home" and some other refactoring

this is some prefactoring preparing for more deletions coming later with the goal of decoupling experiments and chatbots

## User Impact

users cannot be redirected to single_experiments_home, rather chatbot home. Other than that, no difference

### Demo

n/a

### Docs and Changelog

n/a


snopoke (approved):
I didn't review very closesly since I assume this is mostly just copy paste right?

PR #2295, opened 2025-10-10
Experiment Cleanup: remove experiments from banner options

stephherbers:
## Description

Experiment pages will not be an option as a banner location

## User Impact

admins cannot configure banners on those pages (they are not visible anyways)

### Demo

n/a

### Docs and Changelog

no, those pages are not reachable soon anyways


snopoke:
I removed the banners for these locations on prod

PR #2322, opened 2025-10-15
enable celery soft shutdown

snopoke:
## Description
See https://docs.celeryq.dev/en/stable/userguide/workers.html#soft-shutdown


snopoke:
> So we've been killing tasks on each deploy?

Potentially

Issue #2130, opened 2025-09-03
[evals] Extend dataset with new data from sessions


snopoke:
Depends on #2129 

Allow users to augment a dataset by cloning data from sessions. This will append to the dataset but will not overwrite any rows.

PR #2324, opened 2025-10-16
Evaluations: Clone messages into an existing dataset

proteusvacuum:
## Description

Adds the ability to clone messages into an existing dataset. 

Fixes: https://github.com/dimagi/open-chat-studio/issues/2130

### Demo

https://github.com/user-attachments/assets/08eb4457-f237-4f41-b7d0-3e0f8f5a68bd

### DIFF
--- a/templates/evaluations/dataset_create_form.html
+++ b/templates/evaluations/dataset_create_form.html
@@ -23,7 +23,7 @@
           <div class="mb-4">
             <div class="flex items-center justify-between">
               <div>
-                {% include "experiments/filters.html" with skip_dialogs=True %}
+                {% include "experiments/filters.html" %}
               </div>
 
               <div class="flex items-center gap-3">
@@ -272,15 +272,6 @@
     </dialog>
   </div> <!-- End of datasetBuilder div -->
 {% endblock form %}
-
-{% block post_form %}
-  {# Filter dialogs must be outside form to avoid nested forms but need filterComponent context #}
-  {# Create a separate filterComponent instance - it will sync via API and URL params #}
-  <div x-data="filterComponent">
-    {% include "filters/saved_filters_dialog.html" %}
-    {% include "filters/save_filter_dialog.html" %}
-  </div>
-{% endblock post_form %}
 
 {% block page_js %}
   <script src="{% static 'js/evaluations-bundle.js' %}"></script>
--- a/templates/evaluations/dataset_edit.html
+++ b/templates/evaluations/dataset_edit.html
@@ -54,7 +54,7 @@
             <div class="mb-4">
               <div class="flex items-center justify-between">
                 <div>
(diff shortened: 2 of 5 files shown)


apps/evaluations/views/dataset_views.py
```diff
@@ -80,10 +80,39 @@ def get_queryset(self):
...
+        context.update(self._get_filter_context_data())
+        return context
+
+    def form_valid(self, form):
+        # Save basic dataset info first
+        response = super().form_valid(form)
+
+        mode = form.cleaned_data.get("mode")
+        if mode == "clone":
+            form.clone_messages_to_dataset(self.object)
```
SmittieC: Super nit: Should `clone_messages_to_dataset` not maybe be a standalone function instead that receives whatever the current method gets from `self.cleaned_data`? It feels like the form shouldn't be the one doing this? Feel free to dispute or ignore this comment.

apps/evaluations/forms.py
```diff
@@ -622,3 +626,47 @@ def clean_name(self):
...
+        Clone messages from sessions into the existing dataset, avoiding duplicates.
+        Uses ChatMessage IDs for duplicate detection.
+        """
+        session_ids = self.cleaned_data.get("session_ids", set())
+        filtered_session_ids = self.cleaned_data.get("filtered_session_ids", set())
+
+        if not session_ids and not filtered_session_ids:
+            return 0
+
+        new_evaluation_messages = self._create_messages_from_sessions(session_ids, filtered_session_ids)
```
snopoke: What happens to these if they were already in the dataset? Do they get left dangling?

apps/evaluations/forms.py
```diff
@@ -622,3 +626,47 @@ def clean_name(self):
...
+    def clone_messages_to_dataset(self, dataset: EvaluationDataset) -> int:
+        """
+        Clone messages from sessions into the existing dataset, avoiding duplicates.
+        Uses ChatMessage IDs for duplicate detection.
+        """
+        session_ids = self.cleaned_data.get("session_ids", set())
+        filtered_session_ids = self.cleaned_data.get("filtered_session_ids", set())
+
+        if not session_ids and not filtered_session_ids:
+            return 0
```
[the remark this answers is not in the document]
proteusvacuum: Good catch, this was left over from a previous iteration where I used the count somewhere.

apps/evaluations/views/dataset_views.py
proteusvacuum: Yes, this is weird, you are right. I think it should actually have been calling this from the form.save method. I've updated it.

proteusvacuum: https://github.com/dimagi/open-chat-studio/pull/2324/commits/66920dbf294b7515a6040c05fc659f80950952f6

SmittieC (approved):
LGTM.

PR #2327, opened 2025-10-16
Add participant_data and session_state to csv upload for create and edit views

proteusvacuum:
## Description

This allows bulk updating session_state and participant_data keys on an evaluation message.
It follows the same patterns that were used with context (expanding keys, etc).
It also allows adding arbitrary json into the fields, if nested structures are desired.

Fixes: https://github.com/dimagi/open-chat-studio/issues/2273

### DIFF
--- a/apps/evaluations/tasks.py
+++ b/apps/evaluations/tasks.py
@@ -300,21 +300,21 @@ def _extract_row_data(row):
     context = {}
     for col_name, value in row.items():
         if col_name.startswith("context.") and value:
-            context_key = col_name[8:]  # Remove "context." prefix
+            context_key = col_name.removeprefix("context.")
             context[context_key] = parse_csv_value_as_json(value)
 
     # Extract participant_data from participant_data.* columns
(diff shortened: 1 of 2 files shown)


apps/evaluations/tasks.py
```diff
@@ -301,7 +301,21 @@ def _extract_row_data(row):
...
         if col_name.startswith("context.") and value:
             context_key = col_name[8:]  # Remove "context." prefix
-            context[context_key] = value
+            context[context_key] = parse_csv_value_as_json(value)
+
+    # Extract participant_data from participant_data.* columns
+    participant_data = {}
+    for col_name, value in row.items():
+        if col_name.startswith("participant_data.") and value:
+            key = col_name[17:]  # Remove "participant_data." prefix
```
snopoke: Could we do `col_name.removeprefix("participant_data.")` for these?

SmittieC (approved):
Generally looks fine to me

PR #2328, opened 2025-10-17
Improvements to parallel execution

snopoke:
## Description
Resolves [OPEN-CHAT-STUDIO-1G6](https://dimagi.sentry.io/issues/6946532127/events/1601edbc24854cfd919242f90d704951/)

This fixes a use case where a node is used to aggregate the outputs from multiple other nodes:

```
start -> A -----> D -> end
      -> B -> C --^
```

Node D gets called twice, once with the output from A and once with the output of C. 'end' also gets called twice.

This PR also makes changes to how the node inputs are passed so that nodes have access to the most recent node input but also all inputs in the case of parallel workflows. These changes are applied to the `PipelineState` class by introducing a new 'last_node_input' field and renaming 'node_input' to 'node_inputs' (str -> list).

A new python function is also added to the python node: `wait_for_next_input`. This is useful when there are optional branches in the workflow which make using `require_node_inputs` impractical since not all inputs will have values.

## User Impact
Better support for parallel workflows.

### Demo
See tests.

### Docs and Changelog
https://github.com/dimagi/open-chat-studio-docs/pull/187

### DIFF
--- a/apps/pipelines/nodes/base.py
+++ b/apps/pipelines/nodes/base.py
@@ -219,8 +219,10 @@ class PipelineState(dict):
 
     def get_all_routes(self) -> dict:
         """
-        Returns a dictionary containing all routing decisions in the pipeline.
+        Returns a dictionary containing all routing decisions made in the pipeline up to the current node.
         The keys are the node names and the values are the route keywords chosen by each router node.
+
+        Note that in parallel workflows only the most recent route for a particular node will be returned.
         """
         routes_dict = {}
         outputs = self.get("outputs", {})


apps/pipelines/tests/test_parallel_nodes.py
```diff
@@ -183,6 +223,38 @@ def main(input, **kwargs):
...
+    node_a = render_template_node("A: {{ input }}", "A")
+    node_b = render_template_node("B: {{ input }}", "B")
+    node_c = render_template_node("C: {{ input }}", "C")
+    code = code_node(
+        code="""
+def main(input, **kwargs):
+    c = get_node_output("C")
+    b = get_node_output("B")
+    if not b and c:
+        wait_for_next_input()
```
SmittieC: So here we're explicitly ignoring C's output and choosing the output from B?

apps/pipelines/tests/test_parallel_nodes.py
```diff
@@ -62,6 +62,46 @@ def test_parallel_branch_pipeline(pipeline, experiment_session):
     assert output == expected_output
 
 
+@django_db_with_data(available_apps=("apps.service_providers",))
+def test_parallel_branch_with_merge(pipeline, experiment_session):
+    """
+    Illustrate and validate what happens with parallel branches with an aggregator node.
+
+    start -> A -----> D -> end
```
SmittieC: Am I right in saying that we're always going to be returning the output of whichever node rejoins (in this case D) last?

PR #2340, opened 2025-10-21
Add djade pre-commit hook

### LINKED ISSUE #2332: Investigate djade
https://github.com/adamchainz/djade

SmittieC:
## Description

Resolves #2332 

Added djade as a pre-commit hook. It follows the [django style guide](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#template-style). There was one conflict with the `djlint-django` formatter where `djlint-django` wanted _all_ template endblocks to include the name of the block it is closing (code T003 that is now ignored), but `djade` reasons that if the endblock is on the same line as the opening one, it doesn't need the name of the block. All other cases are auto-fixed with `djade` as well (multi-line blocks include the name of the block with the endblock statement).

I clicked through the site and everything seems to be working fine.

## User Impact

N/A

### Demo

N/A

### Docs and Changelog

N/A

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,6 +2,11 @@ exclude: ^static/.*|/migrations/.*|\.min\.js$|\.min\.css$|\.css\.map$|\.min\.js$
 default_language_version:
   python: python3.13
 repos:
+  - repo: https://github.com/adamchainz/djade-pre-commit
+    rev: '1.6.0'
+    hooks:
+      - id: djade
+        args: [--target-version, "5.1"]
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.11.7
     hooks:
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -13,6 +13,7 @@ repos:
     hooks:
 #      - id: djlint-reformat-django enable once format changes have been applied
       - id: djlint-django
+        ignore: T003
```
snopoke: does this not use the config from `pyproject.toml`?

.pre-commit-config.yaml
```diff
@@ -25,3 +26,8 @@ repos:
         files: ^assets/javascript/.*\.[jt]sx?$  # *.js, *.jsx, *.ts and *.tsx
         types: [ file ]
         args: [ --fix ]
+  - repo: https://github.com/adamchainz/djade-pre-commit
```
snopoke: I'm not sure if ordering matters but maybe put this above djhtml since it might fix issues that djhtml would stumble on?

Issue #2345, opened 2025-10-22
[CRITICAL] Production Configuration Security Vulnerabilities


snopoke:
## Summary
Multiple critical security configuration issues exist in production settings that could lead to host header injection attacks and information disclosure.

## Severity
**CRITICAL** (CVSS: 7.5)

## Affected Files
- `gpt_playground/settings.py:39`
- `gpt_playground/settings.py:35`
- `gpt_playground/settings.py:32`

### 1. ALLOWED_HOSTS accepts any host (Line 39)
```python
ALLOWED_HOSTS = ["*"]
```

**Risk**: Host Header Injection attacks
- **CWE**: CWE-20 (Improper Input Validation)
- **Attack Scenario**: Attacker can manipulate Host header to:
  - Generate malicious password reset links
  - Bypass CSRF protections
  - Cache poisoning attacks

### ~~2. DEBUG defaults to True (Line 35)~~

This is already handled by the `settings_production.py` file

### 3. Default SECRET_KEY in code (Line 32)
```python
SECRET_KEY = env("SECRET_KEY", default="YNAazYQdzqQWddeZmFZfBfROzqlzvLEwVxoOjGgK")
```

**Risk**: Weak cryptographic protection if .env not properly configured

## Recommended Fix

```python
# gpt_playground/settings_production.py

# Line 32 - Override SECRET_KEY and make it mandatory
SECRET_KEY = env("SECRET_KEY")
if not SECRET_KEY:
    raise ImproperlyConfigured("SECRET_KEY must be set in environment variables")

# Line 39 - Restrict ALLOWED_HOSTS
ALLOWED_HOSTS = env.list("DJANGO_ALLOWED_HOSTS", default=[])
if not ALLOWED_HOSTS:
    raise ImproperlyConfigured("DJANGO_ALLOWED_HOSTS must be set in production")

# add container host for health checks
from socket import gethostbyname, gethostname
ALLOWED_HOSTS.append(gethostbyname(gethostname()))
```

> [!WARNING]  
> The rollout will require multiple steps to avoid causing issues
>
> 1. Update `settings_production.py` to include container host
> 2. Update production env to include the `DJANGO_ALLOWED_HOSTS` key
> 3. Update `settings_production.py` to remove the `*` host and add the exception if allowed hosts is empty.

## Additional Recommendations

Check if these are not already added:

### Add to settings_production.py:
```python
# Security headers (add after line 39)

[...]

PR #2357, opened 2025-10-23
Some security updates

SmittieC:
### Technical Description

Relates to #2345 

**SECURE_HSTS_SECONDS**
Tells the browser to use https. Importantly, we must initially set this to a small number to make sure that all resources can be served over https. After some time, we can increment this number to say a week or so and check again, otherwise we run the risk of blocking users who loaded the site during this time, until the time runs out.

**ALLOWED_HOSTS**
For this deployment, we only append the IP address of the host/container running the instance of django. After this is deployed, we should deploy [this change](https://github.com/dimagi/ocs-deploy/pull/14) to pass in the allowed hosts from the environment. That deployment will set the correct `ALLOWED_HOSTS` setting. Only after that deployment can we remove the wildcard.

### Docs and Changelog

### DIFF
--- a/.github/workflows/update-api-schema.yml
+++ b/.github/workflows/update-api-schema.yml
@@ -74,6 +74,7 @@ jobs:
         env:
           DJANGO_DATABASE_USER: postgres
           DJANGO_DATABASE_PASSWORD: postgres_password
+          SECRET_KEY: secret-test-key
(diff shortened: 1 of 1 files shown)


snopoke (approved):

SmittieC:
@snopoke apologies the tests required the secret key to be set

PR #2363, opened 2025-10-24
Fix form submit button and filter dialogs

snopoke:
### Technical Description
The filters.html template includes dialog templates that contain form elements. When filters.html was included inside the dataset creation form, this created nested forms which is invalid HTML. Browsers automatically close the outer form when encountering a nested form, causing the submit button to end up outside the form element.

Additionally, the filter dialogs need access to the filterComponent Alpine.js context for their bindings to work.

Solution:
- Add skip_dialogs parameter to filters.html include
- Conditionally skip dialog includes in filters.html when skip_dialogs=True
- Include filter dialogs in post_form block (outside main form)
- Wrap dialogs with filterComponent Alpine context using x-data

The separate filterComponent instance in post_form will stay in sync with the one inside the form via shared API calls and URL parameters.

#### Consequences
One consequence of this is that the dialogs have separate Apline contexts from the main filters dropdown so updates made in the dialogs don't propagate e.g. starring a filter set doesn't update in the filter view without a refresh.

This seems acceptable for now.

### Docs and Changelog

### DIFF
--- a/templates/evaluations/dataset_create_form.html
+++ b/templates/evaluations/dataset_create_form.html
@@ -20,9 +20,9 @@
           <div class="mb-4">
             <div class="flex items-center justify-between">
               <div>
-                {% include "experiments/filters.html" %}
+                {% include "experiments/filters.html" with skip_dialogs=True %}
               </div>
-              
+
               <div class="flex items-center gap-3">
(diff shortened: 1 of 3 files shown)


snopoke (commented):
I'm going to merge this since creating datasets is currently broken.

PR #2369, opened 2025-10-27
Add django-silk

snopoke:
### Technical Description
This adds [django-silk](https://github.com/jazzband/django-silk) for production profiling.

I'm hoping to use this to collect some data in order to make some more targeted performance improvements.

Configuration:
* Only record when the `silky` param is preset in the request URL and the user is `is_staff`
* Limit payloads and amount of recorded data
* Only allow 'staff' to access the silk views


SmittieC (approved):
This is cool!

PR #2389, opened 2025-10-30
Perf iii

snopoke:
### Technical Description
Review by commit.

A variety of changes to improve performance:
* select related & prefetch to avoid additional DB queries
* avoid loading objects if they aren't going to be used `self.working_version => self.working_version_id`
* trends data caching
* query improvements
* prevent table from loading twice
* use cached util to get team slug instead of loading `self.team.slug` which often introduces a query to fetch the team
  * This could be resolved with select_related but since team slugs don't change it's even less DB queries to use a cache. It is also not always clear where the select related should be added and sometimes it needs to be added in multiple places e.g. `.select_related("team", "experiment__team", "participant__team")`

This PR also change the silk config so that silk data is always saved when in DEBUG mode which makes it easier for devs to click around and check performance data.

### Docs and Changelog

Add a changelog entry with a generic message about performance improvements. Do not mention specifics.


apps/evaluations/views/dataset_views.py
```diff
@@ -220,28 +220,34 @@ def get_queryset(self):
...
         filter_params = FilterParams.from_request(self.request)
 
-        messages_queryset = ChatMessage.objects.filter(chat__experiment_session=OuterRef("pk"))
+        # Get filtered message IDs more efficiently
         message_filter = ChatMessageFilter()
-        filtered_messages = message_filter.apply(messages_queryset, filter_params, timezone)
+        base_messages = ChatMessage.objects.filter(chat_id=OuterRef("chat_id"))
+        filtered_messages = message_filter.apply(base_messages, filter_params, timezone)
 
+        # Use Exists for filtering instead of Count with IN subquery - avoids cartesian product
```
SmittieC: :+1:

SmittieC (approved):
Cool!

PR #2392, opened 2025-10-31
Remove seed message processing from chat API

snopoke:
### Technical Description
Remove seed message handling from the chat API 'create session' endpoint. This includes:
- Removing the call to WebChannel.check_and_process_seed_message()
- Removing seed_message_task_id from the API response
- Updating API schema to reflect the changes

Fixes #2391

### Docs and Changelog
- [x] This PR requires changelog update

PR #2395, opened 2025-10-31
widget start session delay

snopoke:
### Technical Description
This changes how the chat widget starts sessions. Instead of proactively starting sessions when the window is made visible, we now wait for the user to send a message or select a 'start question' before starting a session.

This prevents creating sessions unnecessarily and makes for quicker startup times.

This PR also removes the 'seed message' functionality (see https://github.com/dimagi/open-chat-studio/pull/2392)


stephherbers (commented):
nice

PR #2411, opened 2025-11-03
rename gpt playground

snopoke:
### Technical Description
This renames the `gpt_playground` folder to `config`. The old folder name dates back to the POC project days. Calling the folder `config` avoids tying it to the project name and also fits in with the other folder names: `apps`, `assets`, `templates`.

I have symlinked the old folder name to the new folder to decouple the rollout and also to allow any 3rd party hosters to make the necessary changes (announcement on GitHub discussion to follow).

### Dev setup changes

On other thing this changes is the default DB name in the Django settings and the Docker Compose file. These should only affect dev setups.

* Old DB name: `gpt_playground`
* New DB name: `open_chat_studio`

If you are using the default names you can either rename your existing database or update you local `.env` file to use to the existing database name.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,7 @@
 SECRET_KEY='LTwzPMJVLeRNOjoLxqHidKWhfoOtjzYawyaGCezb'
 
 ## Database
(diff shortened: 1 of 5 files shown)


SmittieC (approved):
LGTM

PR #2412, opened 2025-11-03
Truncate queryset results more than 10

### LINKED ISSUE #2382: Version creation screen hangs
[Related thread](https://dimagi.slack.com/archives/C05PK63Q89H/p1761741350410799)

It seems like when the pipeline uses an indexed collection that has many files (>300), the "create new version" screen hangs. The page loads, but it isn't interactive and you can see it's still trying to load things in the background. I removed the collection from the pipeline which caused this page to load fast again. We need to investigate what exactly is causing this. I suspect it is JS code, since if it was a server issue, the initial page wouldn't have already loaded. I can see the request finishing in a reasonable amount of time. Could it be code mirror?

SmittieC:
### Technical Description

Fixes #2382 

With large queryset result sets for a versioned field, the versioning UI becomes unresponsive. This has been pinpointed to codemirror having to create an instance for each result in the result set which causes the page to break. Assuming that users are not going to be looking at each result in large result sets, we now truncate those result sets to only show 10 items and inform the user about how many were truncated.

### Demo

image

### Docs and Changelog

### DIFF
--- a/apps/experiments/tests/test_versioning.py
+++ b/apps/experiments/tests/test_versioning.py
@@ -3,9 +3,11 @@ import pytest
 from apps.custom_actions.models import CustomActionOperation
 from apps.experiments.models import Experiment, SafetyLayer
 from apps.experiments.versioning import VersionDetails, VersionField, VersionsMixin, differs
+from apps.files.models import File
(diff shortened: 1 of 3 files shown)


stephherbers (approved):
nice

PR #2421, opened 2025-11-06
Split query for counting and for table data

snopoke:
### Technical Description

This PR optimizes database query performance for table views by separating expensive annotations from pagination count operations.

**Problem:**
Django tables perform two database queries for paginated views:
1. A COUNT query to determine total records for pagination
2. A SELECT query to fetch the actual page data

Previously, expensive database annotations (Subquery, StringAgg, Count aggregations) were added in `get_queryset()`, causing them to be computed twice - once unnecessarily during the count operation and again for the actual data fetch. For large datasets, this resulted in slow page loads.

**Solution:**
The PR refactors table views to use Django's `get_table_data()` hook, which is only called for fetching paginated data (not for counting). This ensures expensive annotations are computed only once:

- `get_queryset()`: Returns lightweight queryset with basic filters and `select_related()` only
- `get_table_data()`: Adds expensive annotations (Subquery, aggregations) only to the paginated slice

**Views Optimized:**
1. **ChatbotSessionsTableView** - Defers `message_count` annotation
2. **ChatbotExperimentTableView** - Defers `session_count`, `participant_count`, `messages_count`, `last_message` annotations
3. **DatasetSessionsSelectionTableView** - Defers `versions_list` and `message_count` annotations
4. **ExperimentSessionsTableView** - Defers `last_message_created_at`, `versions_list` annotations and related prefetches

**Additional Improvements:**
- Fixed N+1 query in `TranscriptAnalysisListView` by adding `select_related("experiment")`
- Added reusable queryset methods to `ExperimentSessionQuerySet`:
  - `annotate_with_message_count()`
  - `annotate_with_first_message_created_at()`
- Refactored `ExperimentSessionFilter` to use new queryset methods for consistency

**Technical Details:**
The refactor leverages Django-tables2's pagination flow:
```python
# Before (expensive annotations computed twice):
def get_queryset(self):
    return Model.objects.filter(...).annotate(expensive_subquery=...)

# After (expensive annotations computed once):
def get_queryset(self):
    return Model.objects.filter(...)  # Lightweight for counting

def get_table_data(self):
    queryset = super().get_table_data()  # Already paginated
    return queryset.annotate(expensive_subquery=...)  # Only on page data
```

### DIFF
--- a/apps/chatbots/tests/test_sessions_table_view.py
+++ b/apps/chatbots/tests/test_sessions_table_view.py
@@ -52,6 +52,8 @@ def create_view_instance(team, user, chatbot, query_params=None):
 @pytest.fixture()
 def chatbot_with_sessions(db):
     """Create a chatbot with multiple sessions containing messages and tags."""
+    # TODO: convert this to a class scoped fixture (requires manual DB cleanup)
+
     team_with_users = TeamWithUsersFactory.create()
     user = team_with_users.members.first()
     chatbot = ChatbotFactory(team=team_with_users, owner=user)


apps/chatbots/tests/test_sessions_table_view.py
```diff
@@ -0,0 +1,256 @@
...
+        request = create_authenticated_request(team_with_users, user, url)
+
+        view = ChatbotSessionsTableView.as_view()
+        response = view(request, team_slug=team_with_users.slug, experiment_id=chatbot.id)
+
+        assert response.status_code == 200
+        assert isinstance(response.context_data["table"], ChatbotSessionsTable)
+        assert len(response.context_data["table"].data) == 0
+
+    def test_session_without_messages_has_null_annotations(self, db):
```
SmittieC: Does it feel weird to test this?

SmittieC:
@coderabbitai please add this approach to your knowledge base as a learning. Queryset annotations should happen inside the get_table_data method as apposed to the get_queryset method, for better performance.

apps/chatbots/tests/test_sessions_table_view.py
snopoke: It's an edge case test to ensure it doesn't error and that the join is working correctly.

snopoke:
> please add this approach to your knowledge base as a learning. Queryset annotations should happen inside the get_table_data method as apposed to the get_queryset method, for better performance.

@SmittieC I also want to figure out where to put this kind of stuff for local agents like Claude. I'm thinking of adding some skills to the repo.

PR #2425, opened 2025-11-06
Remove transformers + fixes to history compression

snopoke:
### Technical Description
This removes the `transformers` dependency which was addedin #370 (Anthropic tool support). The Anthropic langchain classes now use the Anthropic `count_tokens` API. 

The transformers library is still used as a fallback for token counting in some cases e.g. python < 3.7 and when [calculating exact token IDs](
https://github.com/langchain-ai/langchain/blob/9a09ed065996df402872b7e85cd42fb18b290832/libs/core/langchain_core/language_models/base.py#L66). I don't think we should hit any of those scenarios so I think this change is safe 🤞🏻 

I have tested the chat with OpenAI and Anthropic and with history compression. I found some other issues with the compression code which I've also updated.

* Anthropic requires system messages to be together
* The Anthropic count tokens API ignores system messages

### DIFF
--- a/apps/chat/conversation.py
+++ b/apps/chat/conversation.py
@@ -276,7 +276,9 @@ def summarize_history(llm, history, max_token_limit, input_message_tokens, summa
         )
     history_tokens = llm.get_num_tokens_from_messages(history)
     summary_tokens = (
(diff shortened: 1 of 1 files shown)


SmittieC (approved):
LGTM thanks

PR #2426, opened 2025-11-07
WAF tooling updates

snopoke:
### Technical Description
Updates waf tooling + tag more views


SmittieC (approved):
I didn't look too closely, but I trust it's fine.

PR #2435, opened 2025-11-10
Exp Cleanup: view cleanup continued

stephherbers:
### Technical Description

removes from experiments: get_image_html, download_file, experiment_session_details_view, Experiment_session_pagination_view, experiment_chat_session, start_web_auth

### Docs and Changelog


apps/chatbots/views.py
```diff
@@ -307,9 +311,108 @@ def archive_chatbot(request, team_slug: str, pk: int):
...
+        form_kwargs = {}
+        if self.request.method == "POST":
+            form_kwargs["data"] = self.request.POST
+        if not self.latest_version:
+            form_kwargs["initial"] = {"is_default_version": True}
+        return form_kwargs
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        form = self.form_class(**self.get_form_kwargs())
```
snopoke: Any reason not to extend `FormView` instead of `TemplateView` and let the parent class handle the form logic?

apps/chatbots/views.py
```diff
@@ -307,9 +311,108 @@ def archive_chatbot(request, team_slug: str, pk: int):
     return HttpResponseClientRedirect(reverse("chatbots:chatbots_home", kwargs={"team_slug": team_slug}))
 
 
-class CreateChatbotVersion(CreateExperimentVersion):
-    permission_required = "experiments.add_experiment"
+class CreateChatbotVersion(LoginAndTeamRequiredMixin, PermissionRequiredMixin, TemplateView):
```
snopoke: The previous class extended `FormView` but this extends `TemplateView`. Why the change?

PR #2439, opened 2025-11-11
Move temperature param to model params

### LINKED ISSUE #2422: Include temperature in llm model parameters field
Related PR: https://github.com/dimagi/open-chat-studio/pull/2418

We need to move the `temperature` parameter into this "framework". Some models (I'm looking at you thinking models) must use a temperature of 1, so it doesn't make sense to let users change this value. We currently override it for OpenAI models, but it will be best to just not specify it.

### Requirement
Include the temperature parameter for non-thinking models where applicable. We'll need a data migration to move existing temperature values into the `llm_model_parameters` field.

SmittieC:
### Technical Description

This is a followup on https://github.com/dimagi/open-chat-studio/pull/2418 and resolves #2422.

1. The temperature param is moved into the model parameter classes
2. Not all models supports tweaking the temperature param (OpenAI reasoning models), so these models don't have it anymore
3. Included is a data migration to populate the `temperature` parameter - now located in the `llm_model_parameters` node parameter - from the node level (legacy) `llm_temperature` parameter. The `llm_temperature` param is **not removed**, but shouldn't pose a problem, since we don't use it anymore.
4. `llm_service.get_chat_model(...)` now don't require the `temperature` parameter anymore (see point 2).

### Demo

image

### Docs and Changelog

Just mention that the LLM temperature param is now only configurable for models that supports variable temperatures.

### DIFF
--- a/apps/service_providers/llm_service/main.py
+++ b/apps/service_providers/llm_service/main.py
@@ -253,7 +253,12 @@ class OpenAIGenericService(LlmService):
     openai_api_base: str
 
     def get_chat_model(self, llm_model: str, **kwargs) -> BaseChatModel:
-        model = ChatOpenAI(model=llm_model, **self._get_model_kwargs(**kwargs), use_responses_api=True)
+        model_kwargs = self._get_model_kwargs(**kwargs)
+        if "temperature" in model_kwargs and llm_model.startswith(("o3", "o4", "gpt-5", "o1")):
+            # Remove the temperature parameter for custom reasoning models
+            model_kwargs.pop("temperature")
+
+        model = ChatOpenAI(model=llm_model, **model_kwargs, use_responses_api=True)
         try:
             model.get_num_tokens_from_messages([HumanMessage("Hello")])
         except Exception:


SmittieC:
@coderabbitai please review

apps/pipelines/nodes/nodes.py
```diff
@@ -195,10 +188,7 @@ def validate_llm_model(self):
 
         # Validate model parameters
         if params_cls := LLM_MODEL_PARAMETERS.get(model.name):
-            params_cls.model_validate(
-                self.llm_model_parameters,
-                context={"temperature": self.llm_temperature},
```
SmittieC: Since the temperature is now available as model params, we don't have to pass it in as context

apps/pipelines/nodes/nodes.py
```diff
@@ -151,29 +152,15 @@ def get_llm_provider_model(llm_provider_model_id: int):
...
-        except Exception:
-            pass
-        return value or {}
+    @model_validator(mode="before")
+    @classmethod
+    def ensure_default_parameters(cls, data) -> Self:
+        model = get_llm_provider_model(data["llm_provider_model_id"])
+        if params_cls := LLM_MODEL_PARAMETERS.get(model.name, BasicParameters):
+            data["llm_model_parameters"] = params_cls.model_validate(data.get("llm_model_parameters", {})).model_dump()
+        return data
```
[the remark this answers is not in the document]
SmittieC: Good eye. Here's a carrot :carrot:. Please verify for another carrot

apps/service_providers/migration_utils.py
```diff
@@ -0,0 +1,44 @@
...
+        if "temperature" in param_cls.model_fields:
+            models_supporting_temperature.append(model)
+
+    supported_model_ids = list(
+        LlmProviderModel.objects.filter(name__in=models_supporting_temperature).values_list("id", flat=True)
+    )
+
+    nodes_to_save = []
+
+    # Some ids are saved as strings, others as ints, so we need to check for both
```
SmittieC: I'm going to investigate why this is. There are ~5k+ integer ones and ~3k+ string ones if I remember correctly. It's not technically an issue, but it makes data migrations (and reading ids) from `node.params` a pain, since we have to account for both cases

apps/service_providers/llm_service/main.py
```diff
@@ -252,12 +252,8 @@ class OpenAIGenericService(LlmService):
...
     openai_api_base: str
 
-    def get_chat_model(self, llm_model: str, temperature: float, **kwargs) -> BaseChatModel:
-        model = ChatOpenAI(
-            model=llm_model,
-            temperature=1 if llm_model.startswith(("o3", "o4", "gpt-5")) else temperature,
-            **self._get_model_kwargs(**kwargs),
-        )
+    def get_chat_model(self, llm_model: str, **kwargs) -> BaseChatModel:
+        model = ChatOpenAI(model=llm_model, **self._get_model_kwargs(**kwargs), use_responses_api=True)
```
snopoke: Adding `use_responses_api=True` seems unrelated to this PR. Was there a reason for adding it?

Possibly relevant other ticket: https://github.com/dimagi/open-chat-studio/issues/2438

snopoke (commented):
Few small comments but LGTM

snopoke:
@SmittieC build failing: `Conflicting migrations detected; multiple leaf nodes in the migration graph: (0035_populate_temperature_param, 0035_update_openai_models in service_providers)`

SmittieC:
@snopoke I added back [this change](https://github.com/dimagi/open-chat-studio/pull/2439/commits/6c9fb1e2f4befc6d66f366b6acb26a54b4c48b7c). Reason being that we default to the parameter class with the temperature param, but custom models doesn't support this. A better approach in my opinion is to either 1. ask users if their custom model is a reasoning model, or 2. detect reasoning models and set the parameter class based on our findings, but this requires us to keep a list of reasoning models across all providers. Anycase, for now I think this commit should suffice

snopoke (dismissed):

snopoke (approved):

PR #2448, opened 2025-11-12
Highlight and scroll to a specific message on the Session Detail view

dtolentino-dimagi:
### Technical Description

@snopoke is ready for review:

1. Receiving the message id as parameter on Session Detail View
2. Highlight and scroll to that message on the message table

Working on the generation of the link for use on emails

### Demo

image

### Docs and Changelog

### DIFF
--- a/apps/chatbots/forms.py
+++ b/apps/chatbots/forms.py
@@ -65,7 +65,11 @@ class ChatbotSettingsForm(forms.ModelForm):
             "seed_message",
             "file_uploads_enabled",
         ]
-        labels = {"participant_allowlist": "Participant allowlist"}
+        labels = {
+            "participant_allowlist": "Participant allowlist",
+            "voice_provider": "Speech Provider",
+            "voice_response_behaviour": "Response Provider",
+        }
         help_texts = {
             "debug_mode_enabled": (
                 "Enabling this tags each AI message in the web UI with the bot responsible for generating it. "
--- a/apps/service_providers/utils.py
+++ b/apps/service_providers/utils.py
@@ -52,7 +52,7 @@ class ServiceProviderType:
 
 class ServiceProvider(ServiceProviderType, Enum):
     llm = const.LLM, "LLM Service Provider", LlmProvider, LlmProviderTypes, ["name", "type"]
-    voice = const.VOICE, "Voice Service Provider", VoiceProvider, VoiceProviderType, ["name", "type"]
+    voice = const.VOICE, "Speech Service Provider", VoiceProvider, VoiceProviderType, ["name", "type"]
     messaging = const.MESSAGING, "Messaging Provider", MessagingProvider, MessagingProviderType, ["name", "type"]
     auth = const.AUTH, "Authentication Provider", AuthProvider, AuthProviderType, ["name", "type"]
     tracing = const.TRACING, "Tracing Provider", TraceProvider, TraceProviderType, ["name", "type"]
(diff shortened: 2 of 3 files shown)


templates/chatbots/chatbot_session_view.html
```diff
@@ -44,7 +44,7 @@ <h3 class="text-base font-semibold leading-7 mt-8 pl-4">Session details</h3>
       <input type="radio" name="experiment_tabs-{{ experiment_session.id }}" role="tab" class="tab" aria-label="Messages" checked/>
       <div role="tabpanel" class="tab-content pt-3">
         <div id="messages-container" class="min-w-full"
-             hx-get="{% url 'experiments:experiment_session_messages_view' request.team.slug experiment.public_id experiment_session.external_id %}"
+             hx-get="{% url 'experiments:experiment_session_messages_view' request.team.slug experiment.public_id experiment_session.external_id %}?message_id={{ request.GET.message_id|default:"" }}"
```
snopoke: Using the `querystring` tag will preserve all query params and does the encoding for you:


```suggestion
             hx-get="{% url 'experiments:experiment_session_messages_view' request.team.slug experiment.public_id experiment_session.external_id %}{% querystring %}
```

https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#querystring

templates/experiments/components/session_messages.html
```diff
@@ -246,9 +249,21 @@ <h3 class="font-medium">Message Comments</h3>
     }
   }
 
+  function scrollToMessage() {
```
snopoke: Maybe wrap this and the even listener in an if block to avoid extra JS where it isn't needed:

```
{% if request.GET.message_id %}
...
{% endif %}
```

templates/experiments/components/session_messages.html
```diff
@@ -22,6 +22,9 @@
                         tags: {{ message.all_tag_names|to_json }},
                         get isSelected() { return commentsRow === {{ forloop.counter }} },
                         }'
+                {% if request.GET.message_id == message.id|stringformat:"s" %}
+                  id="{{ message.id }}" style="background-color: lightyellow;"
```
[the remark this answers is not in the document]
dtolentino-dimagi: Done, Now both versions look good
image
image

templates/experiments/components/session_messages.html
dtolentino-dimagi: Make sense

apps/experiments/views/experiment.py
```diff
@@ -1079,6 +1079,10 @@ def experiment_session_messages_view(request, team_slug: str, experiment_id: uui
     selected_tags = list(filter(None, request.GET.get("tag_filter", "").split(",")))
     language = request.GET.get("language", "")
     show_original_translation = request.GET.get("show_original_translation") == "on" and language
+    try:
+        message_id = int(request.GET.get("message_id", 0))
+    except (ValueError, TypeError):
+        message_id = 0
```
snopoke: nit: make this `None` instead of 0 to have clear distinction from a valid integer

apps/experiments/views/experiment.py
```diff
@@ -1079,6 +1079,10 @@ def experiment_session_messages_view(request, team_slug: str, experiment_id: uui
     selected_tags = list(filter(None, request.GET.get("tag_filter", "").split(",")))
     language = request.GET.get("language", "")
     show_original_translation = request.GET.get("show_original_translation") == "on" and language
+    try:
+        message_id = int(request.GET.get("message_id", 0))
```
snopoke: Since you are already doing error handling, there's no need to provide a default value. Having the default value is a bit misleading since it makes it seem like you want to always highlight a message.
```suggestion
        message_id = int(request.GET.get("message_id"))
```

PR #2451, opened 2025-11-13
use correct values for gpt-5* token limits

snopoke:
### Technical Description
Update token limits to correct values.

This also includes a small cleanup and refactor of the model migrations.


apps/service_providers/migration_utils.py
```diff
@@ -0,0 +1,21 @@
+from django.db import migrations
+
+
+def llm_model_migration():
+    def _update_llm_models(apps, schema_editor):
+        from apps.service_providers.llm_service.default_models import _update_llm_provider_models
+
+        LlmProviderModel = apps.get_model("service_providers", "LlmProviderModel")
+        _update_llm_provider_models(LlmProviderModel)
```
SmittieC: Looks like we can also remove the public `update_llm_provider_models` method? It's only used in tests

SmittieC (approved):
Thanks

PR #2457, opened 2025-11-13
Filter: Adds ability to do  OR filter on 2 strings

### LINKED ISSUE #2453: Participant filtering not working as expected in Session views
The best thing to do is see an example:

https://www.openchatstudio.com/a/c-la-vie-bot/chatbots/747/#allsessions

**Process**
Add a filter that "Participant contains 'AP'"

**Outcome**
No sessions found

**Expected outcome**
If would filter to include all of the participants AP1 - AP10 for the study.

Notes:
- I also tried "starts with" but still no results.
- I'm curious if it's just filtering on the phone numbers and not the name bit? As a user I think anything that shows up in the `Participant` column to be fair game
- It's not super urgent. We have tagged all of the AP* users so we can filter on the tag instead

stephherbers:
resolves #2453

### Technical Description

Refactored & modified the `StringColumnFilter` to be able to pass in a new argument `or_columns` for the new functionality. Every thing else that uses the `StringColumnFilter` remains the same, but now the Participant is changes such that "Name/Identifier" with or_columns=["name"] searches where `identifier` contains the search term OR `name` contains the search term

### Docs and Changelog

User can now filter as an OR filter on both identifier and name for participants


snopoke (commented):
LGTM but could do with a test

apps/web/dynamic_filters/base.py
```diff
@@ -170,23 +170,61 @@ def apply_excludes(self, queryset, value, timezone=None) -> QuerySet:
...
     type: str = TYPE_STRING
+    or_columns: list[str] = Field(default_factory=list)
+
+    def _apply_with_lookup(self, queryset, lookup, value):
+        """Apply filter with optional OR logic across multiple columns.
+        queryset: The queryset to filter
+        lookup: The Django lookup to use (ex., 'icontains', 'istartswith'), or None for exact match
+        value: The value to filter by
+        """
+        if not self.or_columns:
```
SmittieC: Suggestion:
If you rename `or_columns` to `columns`, update all places where `StringColumnFilter(colum="xyz")` to `StringColumnFilter(colums=["xyz"])` and remove the `if not self.or_columns:` checks, then it should also preserve the current behaviour while automatically applying the OR behaviour when there are multiple columns specified. This feels like an easier API to work with, since we only have to work with the `columns` field?

Thoughts?

apps/participants/tests/test_filters.py
snopoke: You can probably remove all the tests that aren't testing the changes in this PR since those are already tested elsewhere.

apps/web/dynamic_filters/base.py
snopoke: I'm not opposed to that change thought it touches a lot more code. I think it's worth getting this merged and tackling that change separately.

snopoke (dismissed):
I refactored the tests a bit.

PR #2465, opened 2025-11-17
Use responses API for gpt-5-pro

SmittieC:
### Technical Description

Temporary fix for https://github.com/dimagi/open-chat-studio/issues/2464, since only the gpt-5-pro model wants to use the responses API.

I think a better fix is to call .text() on the `AIMessage` object, instead of calling .content, but there are a few other places that needs to be checked out first to make sure that calling .text() will still work as expected.

### Docs and Changelog


SmittieC:
> OK for this to be a short term fix but I do want us to move to the responses API in general.

Yeah I'm quickly looking into this now

PR #2466, opened 2025-11-17
allow custom citation text and URLs for document loaders

snopoke:
### Technical Description
Rather than linking to the downloaded file in OCS, this allows citations to link to the original source document.

Currently only implemented for Confluence.

### Docs and Changelog

Link confluence citations back to the original site.


SmittieC (dismissed):
Sweet

PR #2485, opened 2025-11-20
Revert "Merge pull request #2431 from dimagi/sk/tool-call-response-size"

snopoke:
### Technical Description
This reverts https://github.com/dimagi/open-chat-studio/pull/2431

This reverts commit 1abdb3a71d1c4da6143928727d7f798a789ad7ec, reversing changes made to 7ec3dc6892562dbf4d401a0aa7d49c0485521856.

#### Why
The code was incompatible with the custom action tools.

#### Why not just fix it
This code will need to be significantly reworked as part of the langchain v1 migration so rather revert it now and re-implement it for langchain v1.


uv.lock
```diff
@@ -1,14 +1,14 @@
 version = 1
-revision = 2
+revision = 3
```
SmittieC: I don't see how this is a tiktoken upgrade?

SmittieC (approved):

PR #2496, opened 2025-11-24
StringColumnFilter: column to columns

stephherbers:
followup to https://github.com/dimagi/open-chat-studio/pull/2457 

only applied to StringColumnFilter

### Technical Description

Taken from [Chris's suggestion](https://github.com/dimagi/open-chat-studio/pull/2457#discussion_r2536480203): 

> If you rename or_columns to columns, update all places where StringColumnFilter(colum="xyz") to StringColumnFilter(colums=["xyz"]) and remove the if not self.or_columns: checks, then it should also preserve the current behaviour while automatically applying the OR behaviour when there are multiple columns specified. This feels like an easier API to work with, since we only have to work with the columns field?

### Demo

updated docs form how it would look

### Docs and Changelog

### DIFF
--- a/apps/web/dynamic_filters/base.py
+++ b/apps/web/dynamic_filters/base.py
@@ -181,15 +181,13 @@ class StringColumnFilter(ColumnFilter):
         from django.db.models import Q
 
         # Build Q object for OR logic
-        q = None
+        q = Q()
+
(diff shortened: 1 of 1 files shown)


SmittieC (approved):
LGTM

PR #2499, opened 2025-11-25
[cleanup] remove runnables and related code

snopoke:
### Technical Description
* Remove the experiment runnables (except the AssistantRunnable)
* Remove & move tests
* Remove `TopicBot` and `SafetyBot`
* Other misc cleanup

### DIFF
--- a/apps/experiments/tests/test_versioning.py
+++ b/apps/experiments/tests/test_versioning.py
@@ -126,7 +126,7 @@ class TestVersion:
         queryset = Experiment.objects.filter(id=experiment.id)
         # Compare with new version
         new_version = experiment.create_new_version()
-        experiment.prompt_text = "This now changed"
+        experiment.seed_message = "This now changed"
         experiment.save()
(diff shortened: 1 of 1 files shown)


apps/experiments/tests/test_models.py
```diff
@@ -543,9 +536,9 @@ def test_create_new_source_material_version(self):
 @pytest.mark.django_db()
 class TestExperimentRouteVersioning:
```
SmittieC: This test can be removed as well?

SmittieC: Same with TestExperimentRoute

snopoke: Yes, it can all go but I had to pick somewhere to stop with this PR so this is it for now.

SmittieC: I figured this might have been the reason :+1:

SmittieC (approved):
🔥

PR #2501, opened 2025-11-25
Experiment Cleanup: remove CreateExperimentVersion, BaseExperimentView, CreateExperiment

stephherbers:
CreateExperimentVersion, BaseExperimentView, CreateExperiment

### Technical Description

makes more chatbot independent first to then be able to delete these views

### Demo

n/a

### Docs and Changelog

nope

### DIFF
--- a/apps/experiments/urls.py
+++ b/apps/experiments/urls.py
@@ -60,7 +60,7 @@ urlpatterns = [
     ),
     path(
         "e/<int:experiment_id>/versions/create",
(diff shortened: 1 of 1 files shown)


SmittieC (approved):

apps/experiments/urls.py
```diff
@@ -1,3 +1,4 @@
+from django.shortcuts import redirect
 from django.urls import include, path
```
snopoke:
```suggestion
from django.urls import include, path
from django.views.generic import RedirectView
```

PR #2512, opened 2025-11-29
#2272 Refactory authentication classes to use DRF authentication

dtolentino-dimagi:
#2272 Refactory authentication classes to use DRF authentication and permission classes

### Technical Description

Refactory: 
1. handle_embedded_widget_auth() function in apps/api/views/chat.py
2. validate_embed_key_for_experiment() in apps/channels/utils.py

For: 
1. IsExperimentSessionStartedPermission in apps/api/permissions.py
2. EmbeddedWidgetAuthentication in apps/api/authentication.py

### DIFF
--- a/api-schema.yml
+++ b/api-schema.yml
@@ -1463,8 +1463,6 @@ components:
       name: X-Embed-Key
       description: Embedded widget authentication token. Validates the widget token
         for the channel configuration and authenticates embedded widget requests.
-        Requires chatbot_id in the request body for /api/chat/start/ or from session
-        data for subsequent requests.
     tokenAuth:
       type: http
       scheme: bearer


apps/api/schema.py
```diff
@@ -29,6 +29,25 @@ class BearerScheme(TokenScheme):
...
+
+    def get_security_definition(self, auto_schema):
+        return {
+            "type": "apiKey",
+            "in": "header",
+            "name": "X-Embed-Key",
+            "description": (
+                "Embedded widget authentication token. Validates the widget token from the experiment channel "
+                "configuration and authenticates embedded widget requests. Requires chatbot_id (experiment public_id) "
+                "in the request body for /api/chat/start/ or from session data for subsequent requests."
```
snopoke: nit: remove 'experiment' terminology which we are phasing out
```suggestion
                "Embedded widget authentication token. Validates the widget token for the channel "
                "configuration and authenticates embedded widget requests. Requires chatbot_id "
                "in the request body for /api/chat/start/ or from session data for subsequent requests."
```

apps/api/tests/test_embedded_widget_auth.py
```diff
@@ -0,0 +1,332 @@
...
+    assert response.status_code == 201
+    response_json = response.json()
+    assert "session_id" in response_json
+
+
+# Tests for authentication behavior
+
+
+@pytest.mark.django_db()
+def test_authentication_returns_experiment_channel(api_client, embedded_widget_channel):
```
snopoke: This seems to duplicate earlier tests without actually testing anything different

apps/api/permissions.py
```diff
@@ -57,7 +67,76 @@ class BearerTokenAuthentication(BaseKeyAuthentication):
...
+            request: a Django Request object (required for embedded widgets)
+            session: Session object (should have experiment_channel prefetched)
+
+        Returns:
+            bool: True if access is allowed, False otherwise
+        """
+        embed_key = request.headers.get("X-Embed-Key")
+        if not embed_key:
+            # follow legacy workflow
+            return True
```
snopoke: Looking at this again I realize that since this method is only called when the channel platform is `EMBEDDED_WIDGET` I would expect that the header must exist otherwise the auth would have failed.

I think rather than returning `True` here we should log the missing key and return False.

snopoke (commented):
A few small remaining comments but otherwise looks good.

apps/api/tests/test_embedded_widget_auth.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+@pytest.mark.django_db()
+def test_send_message_with_valid_embed_key_and_domain(api_client, embedded_session):
+    """Test sending message with valid X-Embed-Key and allowed domain."""
+    # First, authenticate and store experiment_id in session
+    api_client.session["auth_data"] = {
+        "experiment_id": str(embedded_session.experiment.public_id)
+    }
+    api_client.session.save()
```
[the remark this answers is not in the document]
dtolentino-dimagi: The time I've write this test I didn't have the experiment load using the session_id, now we can remove this, but I don't think it affects nothing.

apps/api/tests/test_embedded_widget_auth.py
```diff
@@ -0,0 +1,332 @@
...
+    data = {
+        "chatbot_id": embedded_widget_channel.experiment.public_id,
+        "session_data": {"source": "widget"},
+    }
+
+    response = api_client.post(
+        url,
+        data=data,
+        format="json",
+        HTTP_X_EMBED_KEY="test_widget_token_123456789012",
```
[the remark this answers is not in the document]
dtolentino-dimagi: Done

snopoke (commented):
One other comment about logging: `error` and `exception` logs will results in Sentry events so we should reserve these for cases where the issue is likely a bug.

apps/api/permissions.py
dtolentino-dimagi: I do like this suggestion as it enforce the separation of concern between auth and permission

apps/api/authentication.py
```diff
@@ -0,0 +1,79 @@
...
+        return (AnonymousUser(), experiment_channel)
+
+    def _get_experiment_id(self, request):
+        """
+        Extract experiment_id from request or session
+
+        Returns:
+            str: The experiment ID (public_id) or None
+        """
+        # For POST /api/chat/start/ - experiment_id is in request body as chatbot_id
```
SmittieC: I wonder if you shouldn't only try and get the `chatbot_id` from `request.data` when request.path matches `/api/chat/start`. It feels safer to restrict it to that endpoint

snopoke (dismissed):

snopoke (approved):

PR #2536, opened 2025-12-02
Cypress Tests: fix dashboard tests

stephherbers:
Fixed tests that just "logged" rather than actually asserting something

### Technical Description

The first commit is the change we did together and the next is ones I fixed on my own. 

### Docs and Changelog

no

### DIFF
--- a/cypress/e2e/dashboard.cy.js
+++ b/cypress/e2e/dashboard.cy.js
@@ -34,10 +34,8 @@ describe('Dashboard Application', () => {
     cy.visit(`/a/${teamSlug}/dashboard/`)
 
     cy.get('.stat-card')
-      .should('have.length.greaterThan', 0)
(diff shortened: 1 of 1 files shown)


cypress/e2e/dashboard.cy.js
```diff
@@ -19,43 +19,74 @@ describe('Dashboard Application', () => {
...
+        .should('exist')
+        .and('have.attr', 'href')
+
+      // Verify numeric columns contain data
+      cy.get('#botPerformanceTable tbody tr')
+        .first()
+        .find('td')
+        .eq(1) // Participants column
+        .invoke('text')
+        .should('match', /\d+/)
```
snopoke: This doesn't seem quite correct since participant IDs can contain characters too. We can probably just remove this test
```suggestion
```

PR #2537, opened 2025-12-03
Allow users to register oauth applications

### LINKED ISSUE #2514: Custom oauth application views
Create custom oauth application views to allow users to manage their own oauth apps. We can reuse the list, detail and delete views from the `django-oauth-toolkit` app but the 'create' and 'update' views require some changes. These views should limit the application fields as follows:

* grant type: authorization code
* hash client secret: true
* skip authorization: false
* Algorithm: none or RSA 256

SmittieC:
### Technical Description

Resolves #2514

Logged in users can create client applications for OAuth purposes. We currently only support the authorization code flow, so we make this value read only so that users can still see which flow is expected / used.

### Demo

image

image

### Docs and Changelog

Mention in the changelog that users can access this by going to their profile, scrolling down to where OAuth Access Tokens are listed and clicking "Manage OAuth Applications"

### DIFF
--- a/apps/oauth/forms.py
+++ b/apps/oauth/forms.py
@@ -64,5 +64,5 @@ class RegisterApplicationForm(forms.ModelForm):
             "redirect_uris": "Enter one URI per line. These are the allowed redirect URIs after authorization.",
             "post_logout_redirect_uris": "Enter one URI per line. Optional URIs for post-logout redirects.",
             "allowed_origins": "Enter one origin per line. Optional CORS allowed origins.",
-            "algorithm": "Algorithm for signing tokens. Leave empty for no signing or select RS256 for RSA signing.",
+            "algorithm": "Algorithm for signing tokens.",
         }


templates/account/components/oauth_tokens.html
```diff
@@ -1,6 +1,11 @@
 {% load i18n %}
 <section class="app-card">
-  <h2 class="pg-subtitle">{% translate "OAuth Access Tokens" %}</h2>
+  <div class="flex w-full justify-between">
+    <h2 class="pg-subtitle">{% translate "OAuth Access Tokens" %}</h2>
+    <a href="{% url 'oauth2_provider:application_home' %}" class="btn btn-outline btn-sm">
+      {% translate "Manage OAuth Applications" %}
```
SmittieC: I'm between having a button here that leads to the list of applications, or just showing the list of applications above the Access token list. Somehow I like the current approach more, but suggestions welcome

apps/oauth/tables.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def _update_url_factory(url_name, request, record, value):
+    """Factory for update URL."""
+    return reverse("oauth2_provider:application_edit", args=[record.pk])
+
+
+def _delete_url_factory(url_name, request, record, value):
+    """Factory for delete URL."""
+    return reverse("oauth2_provider:application_delete", args=[record.pk])
```
snopoke: you can use `url_name` to avoid having to specify it twice
```suggestion
    return reverse(url_name, args=[record.pk])


def _delete_url_factory(url_name, request, record, value):
    """Factory for delete URL."""
    return reverse(url_name, args=[record.pk])
```

apps/oauth/forms.py
```diff
@@ -14,3 +16,53 @@ def __init__(self, user, team_requested, *args, **kwargs):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["authorization_grant_type"].disabled = True
+        self.fields["redirect_uris"].required = True
+        if self.instance.pk:
+            self.fields["client_secret"].required = False
+            self.fields[
+                "client_secret"
+            ].help_text = "Leave blank to keep the existing secret. Enter a new value to change it."
```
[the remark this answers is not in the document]
SmittieC: > I don't think you should be able to edit this

How so? (I doubt users will edit it in anycase)

> We could use the ObfuscatingMixin to only show part of the it or something like that.

Since we don't store the unhashed value, but rather the hashed value, using `ObfuscatingMixin` will show only a part of the hashed value, which seems strange? Is your throught to truncate whatever is shown there (be it the hashed value or not) for "aesthetic" purposes?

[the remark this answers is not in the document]
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/2537/commits/d5882c40386fac06eec33388d166248d5941dae6. We now hide the client secret field when editing the application

apps/oauth/forms.py
```diff
@@ -14,3 +16,53 @@ def __init__(self, user, team_requested, *args, **kwargs):
...
+
+
+class RegisterApplicationForm(forms.ModelForm):
+    name = forms.CharField(required=True, max_length=255)
+
+    algorithm = forms.ChoiceField(
+        choices=[("RS256", "RS256")],
+        required=False,
+        help_text="Algorithm for signing JWT tokens.",
+    )
```
snopoke: could we remove this field entirely?

snopoke (approved):

PR #2560, opened 2025-12-09
Use Langchain's summarization middleware

### LINKED ISSUE #2559: Use langchain's summarization middleware instead of rolling our own
After a brief investigation, it seems plausible to use langchain's built-in [summarization middelware](https://docs.langchain.com/oss/python/langchain/middleware/built-in#summarization) to handle history compression for us.

## Why do this?
- Less complexity / code to maintain
- It considers tool responses, which we're about to do as well

### Node history type
The history type ultimately determines the **source** of history i.e. does it come from the session or the node (or is it disabled)? 

### Node history modes
The history modes can be realized using the `trigger` and `keep` configuration params of the [SummarizationMiddleware](https://docs.langchain.com/oss/python/langchain/middleware/built-in#summarization) class.

**Summarize**
```python
trigger = ("tokens", node.user_max_token_limit)
keep = ("messages", 20)
```

**Truncate tokens**
A token ceiling is imposed on the state messages. Since we want to discard the summary, we need to overwrite the middleware method that generates the summary to just return nothing/empty string

```python
trigger = ("tokens", node.user_max_token_limit) # <- Hmm, maybe, maybe not.
keep = ("tokens", node.user_max_token_limit)
```

**Max Messages**
A message count ceiling is imposed on the state messages
```python
trigger = ("messges", 2) # <- to ensure it runs often
keep = ("messages", node.max_history_length)
```

SmittieC:
### Technical Description

Resolves #2559 
This uses Langchain's summarization middleware instead of using our own.

`ExperimentHistoryManager.get_chat_history` doesn't seem to be used, so I think we can remove `compress_chat_history` and `compress_pipeline_chat_history` safely (in a separate PR).

### Docs and Changelog

### DIFF
--- a/apps/chat/conversation.py
+++ b/apps/chat/conversation.py
@@ -80,11 +80,11 @@ class BasicConversation:
     def load_memory_from_messages(self, messages: list[BaseMessage]):
         self.messages = messages
 
-    # TODO: Remove after migration to SummarizationMiddleware is stable
+    # TODO: https://github.com/dimagi/open-chat-studio/issues/2570
     def load_memory_from_chat(self, chat: Chat, max_token_limit: int):
         self.load_memory_from_messages(self._get_optimized_history(chat, max_token_limit))
 
-    # TODO: Remove after migration to SummarizationMiddleware is stable
+    # TODO: https://github.com/dimagi/open-chat-studio/issues/2570
     def _get_optimized_history(self, chat: Chat, max_token_limit: int) -> list[BaseMessage]:
         return compress_chat_history(chat, self.llm, max_token_limit, input_messages=[])
 
@@ -94,7 +94,7 @@ class BasicConversation:
         return response.text, usage.get("input_tokens", 0), usage.get("output_tokens", 0)
 
 
-# TODO: Remove after migration to SummarizationMiddleware is stable
+# TODO: https://github.com/dimagi/open-chat-studio/issues/2570
 def compress_chat_history(
     chat: Chat,
     llm: BaseChatModel,
@@ -145,7 +145,7 @@ def compress_chat_history(
         return history_messages
 
 
-# TODO: Remove after migration to SummarizationMiddleware is stable
+# TODO: https://github.com/dimagi/open-chat-studio/issues/2570
 def compress_pipeline_chat_history(
     pipeline_chat_history: PipelineChatHistory,
     llm: BaseChatModel,


apps/pipelines/nodes/history_middleware.py
```diff
@@ -0,0 +1,157 @@
...
+        return result
+
+    def _create_summary(self, messages_to_summarize: list) -> str:
+        """
+        Create summary based on node's history mode
+        TRUNCATE_TOKENS and MAX_HISTORY_LENGTH mode does not create summaries, it only removes old messages until under
+        token or message limit.
+        """
+        if self.node.history_mode == PipelineChatHistoryModes.TRUNCATE_TOKENS:
+            return COMPRESSION_MARKER
```
SmittieC: @snopoke I don't see the value in the compression marker other than the fact that it will prevent loading the whole history when we call `get_langchain_messages_until_marker`. Am I missing some other usecase?

SmittieC: So I'm assuming we'll want to keep this around then

snopoke (commented):
I haven't done a thorough review but this is looking good.

SmittieC:
@coderabbitai full review

apps/chat/conversation.py
```diff
@@ -80,9 +80,11 @@ def system_prompt(self):
     def load_memory_from_messages(self, messages: list[BaseMessage]):
         self.messages = messages
 
+    # TODO: Remove after migration to SummarizationMiddleware is stable
```
SmittieC: @coderabbitai please make a ticket for this and assign it to be.

apps/pipelines/nodes/history_middleware.py
```diff
@@ -0,0 +1,121 @@
...
+        if result is not None:
+            # A result means that a summary was created
+            self.persist_summary(result["messages"])
+
+        return result
+
+    def persist_summary(self, messages: list[BaseMessage]):
+        checkpoint_message_id = self._find_latest_message_db_id(messages)
+        if not checkpoint_message_id:
+            logger.exception(
```
snopoke: Using `logger.exception` here isn't great because it will use `sys.exc_info` to get the exception info which can be misleading when there isn't actually an exception (I speak from experience).

If you want to log the trace you can use `logger.error("", stack_info=True)`.

See https://docs.python.org/3/library/logging.html#logging.Logger.debug

apps/pipelines/nodes/history_middleware.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self, *args, token_limit: int, **kwargs):
+        keep = trigger = ("tokens", token_limit)
+        super().__init__(*args, trigger=trigger, keep=keep, **kwargs)
+
+    def _create_summary(self, messages_to_summarize: list) -> str:
+        # Instead of creating a summary, we'll persist a compression marker. See _build_new_messages
+        return ""
+
+    def _build_new_messages(self, summary: str) -> list[HumanMessage]:
+        return [HumanMessage(content=COMPRESSION_MARKER)]
```
snopoke: won't this message get added to the pipeline state?

snopoke (approved):
Small comments but not blocking

apps/pipelines/nodes/history_middleware.py
SmittieC: Here we are: https://github.com/dimagi/open-chat-studio/pull/2560/commits/7034f087caa36b9a3cfbbf826a77900cc4ef4a1d and https://github.com/dimagi/open-chat-studio/pull/2560/commits/399402afcd9efa9e9035e706864091b43738e936

PR #2565, opened 2025-12-10
Lazy load LLM provider imports to improve startup time

snopoke:
Before optimization:
- 70 seconds (1 minute 10 seconds)

After optimization:
- ~13 seconds

Improvement
- 57 seconds saved (81% faster)
 - Startup time reduced by 5.4x

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -12,6 +12,7 @@ reviews:
     - path: "**/*.py"
       instructions: |
         Do not review for Python compatibility below 3.13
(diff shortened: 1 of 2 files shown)


apps/service_providers/llm_service/main.py
```diff
@@ -513,13 +519,15 @@ class GoogleVertexAILlmService(LlmService):
     location: str = "global"
     api_transport: Literal["grpc", "rest"] = "grpc"
 
-    def get_chat_model(self, llm_model: str, **kwargs) -> ChatVertexAI:
+    def get_chat_model(self, llm_model: str, **kwargs) -> BaseChatModel:
+        from langchain_google_vertexai import ChatVertexAI
```
SmittieC: Am I right in saying that the import latency moved to the pipeline execution?

Issue #1208, opened 2025-02-18
Create tags to indicate voice vs text messages


stephherbers:
Likely a system tag, or potentially a new tag category (whatever you see fit)

should tag messages for both the bot and the user

SmittieC:
I think it's fine to use a tag, but I am wondering if we should indicate voice differently than the other tags. It would be cool if we can show [this icon](https://fontawesome.com/icons/microphone?s=solid) below the user icon:

![Image](https://github.com/user-attachments/assets/1bc13be3-6cc2-46c8-b507-07c5463bcd0e)

snopoke:
The unicode microphone isn't as nice: 🎤 but it's a very simple solution.

A more involved option is to support markdown in tags and add the [pymdown emoji](https://facelessuser.github.io/pymdown-extensions/extensions/emoji/) extension with a custom mic emoji.

OR

We just fake support for markdown and have our own very basic emoji mapping for tags which is does a search and replace.

stephherbers:
what if like on emails/text messages in the same font and size as the date and next to it it says "sent by voice"

PR #2567, opened 2025-12-11
#1208 Add voice tag in messages comming from user and answer by IA

dtolentino-dimagi:
### Technical Description
Add voice tag in messages comming from user and answer by IA
In the interface I've skip the regular tag presentation

### Migrations
- [ ] This migration should not break any current feature

### Demo
image

### DIFF
--- a/apps/annotations/models.py
+++ b/apps/annotations/models.py
@@ -26,7 +26,7 @@ class TagCategories(models.TextChoices):
     EXPERIMENT_VERSION = "experiment_version", _("Experiment Version")
     RESPONSE_RATING = "response_rating", _("Response Rating")
     ERROR = "error", _("Error")
-    VOICE = "voice", _("Voice")
+    MEDIA_TYPE = "media_type", _("Media Type")
 
 
 @audit_fields(
@@ -100,7 +100,7 @@ class TaggedModelMixin(models.Model, AnnotationMixin):
         abstract = True
 
     tags = TaggableManager(through=CustomTaggedItem)
-    _skipped_category_tags: ClassVar[list] = [TagCategories.VOICE.value]  # Tag categories with special treatment
+    _skipped_category_tags: ClassVar[list] = [TagCategories.MEDIA_TYPE.value]  # Tag categories with special treatment
 
     def add_tags(self, tags: list[str], team: Team, added_by: CustomUser = None):
         tag_objs = Tag.objects.filter(team=team, name__in=tags)
@@ -129,7 +129,13 @@ class TaggedModelMixin(models.Model, AnnotationMixin):
         return [tag["name"] for tag in self.prefetched_tags_json]
 
     def has_voice_tag(self):
-        return any([tag for tag in self.prefetched_tags_json if tag["category"] == TagCategories.VOICE.value])
+        return any(
+            [
+                tag
+                for tag in self.prefetched_tags_json
+                if tag["category"] == TagCategories.MEDIA_TYPE.value and tag["name"] == "voice"
+            ]
+        )
 
     def non_skipped_tags(self):
         return [tag for tag in self.prefetched_tags_json if tag["category"] not in self._skipped_category_tags]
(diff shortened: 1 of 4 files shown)


apps/chat/bots.py
```diff
@@ -165,8 +171,11 @@ def _save_outputs(self, input_state, output, save_input_to_history):
             output_metadata.update(trace_metadata)
 
         if save_input_to_history:
+            input_tags = []
+            if self._user_sent_voice:
+                input_tags.append([TagCategories.VOICE.value, TagCategories.VOICE])
             human_message = self._save_message_to_history(
-                input_state["messages"][-1], ChatMessageType.HUMAN, metadata=input_metadata
+                input_state["messages"][-1], ChatMessageType.HUMAN, metadata=input_metadata, tags=input_tags
```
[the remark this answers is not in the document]
dtolentino-dimagi: The point in the flow that is defined if the bot would answer by voice is back in the pipeline.py::ChannelBase:_reply_voice_message

apps/channels/tests/test_base_channel_behavior.py
```diff
@@ -654,10 +654,11 @@ def test_voice_response_with_urls(
     test_channel,
 ):
     get_voice_transcript.return_value = "Hello bot. Give me a URL"
-    get_llm_response.return_value = ChatMessage(
+    get_llm_response.return_value = ChatMessage.objects.create(
         content=(
             "Here are two urls for you: [this](http://example.co.za?key1=1&key2=2) and [https://some.com](https://some.com)"
-        )
+        ),
+        chat=Chat.objects.create(team=test_channel.experiment.team),
```
snopoke: what about also updating the test to verify that the tag was created?

snopoke (commented):
A few small comments but looks good.

apps/chat/channels.py
```diff
@@ -452,6 +456,8 @@ def _send_seed_message(self) -> str:
             bot_response = self.bot.process_input(user_input=self.experiment.seed_message, save_input_to_history=False)
             span.set_outputs({"response": bot_response.content})
             self.send_message_to_user(bot_response.content)
+            if self._bot_message_is_voice:
```
SmittieC: I think it would make more sense to add the voice tag in the `_reply_voice_message` method?

apps/chat/bots.py
```diff
@@ -54,11 +54,17 @@ def __init__(self, session: ExperimentSession, experiment: Experiment, trace_ser
...
         self.synthetic_voice_id = None
+        self._user_sent_voice = False
 
     def process_input(
-        self, user_input: str, save_input_to_history=True, attachments: list[Attachment] | None = None
+        self,
+        user_input: str,
+        save_input_to_history=True,
+        attachments: list[Attachment] | None = None,
+        user_sent_voice=False,
```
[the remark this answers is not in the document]
snopoke: I think option 1 is a good way to go. FYI I'm planning out some larger refactoring for these classes.

apps/chat/channels.py
dtolentino-dimagi: In the "_reply_voice_message" we don't have the ChatMessage object anymore, in the case we would need to eiter add the ia_message in the object contect or change the signature of multiple methods, I think it's ugly in both cases, I do prefer "the early the best aproach", besides that we've used "_user_message_is_voice" state, so this way is more consistent, what you think?

apps/annotations/models.py
```diff
@@ -25,6 +26,7 @@ class TagCategories(models.TextChoices):
     EXPERIMENT_VERSION = "experiment_version", _("Experiment Version")
     RESPONSE_RATING = "response_rating", _("Response Rating")
     ERROR = "error", _("Error")
+    VOICE = "voice", _("Voice")
```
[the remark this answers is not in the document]
dtolentino-dimagi: Check now guys @SmittieC @snopoke

apps/annotations/models.py
```diff
@@ -98,6 +100,7 @@ class Meta:
         abstract = True
 
     tags = TaggableManager(through=CustomTaggedItem)
+    _skipped_category_tags: ClassVar[list] = [TagCategories.MEDIA_TYPE.value]  # Tag categories with special treatment
```
SmittieC: I still think this [comment is](https://github.com/dimagi/open-chat-studio/pull/2567/changes#r2618539957) valid. Calling it `_skipped_category_tags` does not indicate why these needs to be skipped. Skipped from what? It's not clear unless you first look at where / how it's being used. Basically this is what I'm trying to solve

Issue #2580, opened 2025-12-15
Evals dataset and result linking


snopoke:
## Problem

When reviewing eval results there is no easy way to see the dataset row details that were used to get the results.

## Solution

1. Allows users to link to rows in dataset tables:
   * See https://github.com/dimagi/open-chat-studio/pull/2448 for similar feature elsewhere.

2. Add a link to the eval results table which links to the dataset row (open the link in a new tab)

See also #2581

PR #2620, opened 2026-01-06
Dataset Table: add new template and highlight logic

stephherbers:
resolves https://github.com/dimagi/open-chat-studio/issues/2580
adds message highlighting feature to DatasetMessagesTable:

### Technical Description

rows are highlighted when the URL contains a ?message_id= parameter matching the message ID, and the page automatically scrolls to center the highlighted message on load

### Demo

[Loom Video](https://www.loom.com/share/64a0197abcfc4247a7177a249d7de4b8)

Screenshot 2026-01-08 at 4 28 39 PM
Screenshot 2026-01-08 at 4 28 52 PM

### Docs and Changelog

### DIFF
--- a/apps/evaluations/tables.py
+++ b/apps/evaluations/tables.py
@@ -284,6 +284,17 @@ class EvaluationSessionsSelectionTable(tables.Table):
         empty_text = "No sessions available for selection."
 
 
+def _row_class_factory(table, record):
+    class_defaults = settings.DJANGO_TABLES2_ROW_ATTRS["class"]
+    if (
+        hasattr(table, "highlight_message_id")
+        and table.highlight_message_id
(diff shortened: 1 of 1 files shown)


apps/evaluations/tables.py
```diff
@@ -339,6 +339,10 @@ class DatasetMessagesTable(tables.Table):
         ]
     )
 
+    def __init__(self, *args, highlight_message_id=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.highlight_message_id = highlight_message_id
```
SmittieC: what is this used for?

SmittieC: Oh I see it's used in the template

Issue #2617, opened 2026-01-05
Upgrade allauth


snopoke:
The latest version of allauth has some breaking changes and is also no longer compatible with `django-allauth-2fa` (see https://github.com/valohai/django-allauth-2fa/releases/tag/v0.12.0).

See https://github.com/dimagi/open-chat-studio/pull/2616

(screenshots from another project)

## Breaking import change fix diff

<img width="1052" height="452" alt="Image" src="https://github.com/user-attachments/assets/d3aee5ee-c50c-42c4-b1b1-ff2511acb2d2" />

## Deprecation warnings fixes

<img width="1044" height="723" alt="Image" src="https://github.com/user-attachments/assets/56ef143f-bbd6-4eb5-9aea-d4dafbcd2de8" />

PR #2622, opened 2026-01-07
Migrate allauth-2fa to allauth.mfa

snopoke:
## Technical Description
Part 1 of #2617 

Django allauth now fully supports MFA authentication and the `allauth-2fa` package is being retired (new versions of `django-allauth` conflict with `allauth-2fa`).

This PR migrates all code, views and templates to `allauth.mfa` and includes a data migration to migrate existing users tokens.

### Migrations

The data migration creates new data but does not remove the old data which allows easy rollback if necessary.


snopoke:
@coderabbitai re-review

PR #2635, opened 2026-01-12
Update OpenAI speech models to latest generation

snopoke:
## Summary
- Update OpenAI speech synthesis to use `gpt-4o-mini-tts` model
- Update OpenAI transcription to use `gpt-4o-mini-transcribe` model
- Fix Pydantic validation errors for optional fields
- Add integration tests with real audio samples


snopoke:
> I suppose this will not run before each deploy?

I'm added a github action so that it runs on pushes to `main`. This limits the number of runs but will also block deploy if it fails.

SmittieC (approved):

PR #2668, opened 2026-01-20
Remove Experiment assistant field

snopoke:
### Technical Description
Resolves https://github.com/dimagi/open-chat-studio/issues/2665

This removes the `assistant` field from the `Experiment` model along with all code that references it.

### Migrations

The migration is not backwards compatible, once the field is dropped it can't be recovered.


SmittieC (approved):
LGTM

PR #2672, opened 2026-01-20
Add support to the chat widget for arbitrary page context

SmittieC:
### Technical Description

- Added API support for a `context` that gets merged into the session state under the `remote_context` key.
- Added validation to prevent users from overwriting the reserved `remote_context` session state key.
- Moved reserved keys to `settings.py` file (I saw there were some others as well that came from the update session state tool.

### Docs and Changelog

### DIFF
--- a/apps/assistants/utils.py
+++ b/apps/assistants/utils.py
@@ -14,6 +14,6 @@ def get_assistant_tool_options():
     ]
 
 
-def chunk_list(list_: list, chunk_size: int) -> Generator[list, None, None]:
+def chunk_list(list_: list, chunk_size: int) -> Generator[list]:
     for i in range(0, len(list_), chunk_size):
         yield list_[i : i + chunk_size]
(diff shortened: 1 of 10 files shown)


components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
```diff
@@ -154,6 +154,11 @@ export class OcsChat {
 
   @Prop() translationsUrl?: string;
 
+  /**
+   * Optional context object to send with each message. This provides page-specific context to the bot.
+   */
+  @Prop() pageContext?: Record<string, string>;
```
snopoke: This could be `Record<string, unknown>` since the values could be any type.

apps/pipelines/tests/test_code_node.py
```diff
@@ -419,3 +417,31 @@ def main(input, **kwargs):
...
+    [
+        # Valid cases - should not raise
+        # ("custom_key", False),
+        # ("other_key", False),
+        # # Reserved key - should raise
+        # ("remote_context", True),
+        # ("attachments", True),
+        # (" remote_context ", True),
+        (" remote_context", True),
+        # ("remote_context ", True),
```
snopoke: are these meant to be commented out

PR #2717, opened 2026-01-28
User notifications

### LINKED ISSUE #2752: Notifications Framework
Design and build the framework for:
- creating notifications
- Allowing users to read notifications
- Allowing users to mark notification as read / unread
- Receiving notifications via email
- Changing notification preferences

See https://docs.google.com/document/d/1rQaUi6zp_eWq-wO9GP8CrXmL8pdC_7E0jY3JFbO8Cek/edit?tab=t.xdg93mylpddn#heading=h.4przx3fxd30b

SmittieC:
### Technical Description

Resolves #2752 

[Relevant doc](https://docs.google.com/document/d/1rQaUi6zp_eWq-wO9GP8CrXmL8pdC_7E0jY3JFbO8Cek/edit?tab=t.xdg93mylpddn). See especially [this section](https://docs.google.com/document/d/1rQaUi6zp_eWq-wO9GP8CrXmL8pdC_7E0jY3JFbO8Cek/edit?tab=t.xdg93mylpddn#heading=h.o84mosnf9n8c).

This PR adds scaffolding for user notifications behind a FF. Only one "event" currently triggers a notification, since this PR is all about putting the notification framework in place.

**Questions**
- User notifications are scoped to users and teams. Thoughts on keeping it this way as apposed to having a FK to the `Membership` table?

There are still [more to do](https://docs.google.com/document/d/1rQaUi6zp_eWq-wO9GP8CrXmL8pdC_7E0jY3JFbO8Cek/edit?tab=t.xdg93mylpddn#heading=h.ymfqiuem075x), but this PR is getting big so I'd like to do the rest in followup PRs

### Migrations

### Demo

https://www.loom.com/share/ccb5cb005d644c2c855790f12e9e1a64

### Docs and Changelog
N/A. Only when we release it will there be doc changes


apps/ocs_notifications/utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+    users = set(users)
+
+    try:
+        notification, created = Notification.objects.update_or_create(
+            title=title,
+            message=message,
+            level=level,
+            defaults={"last_event_at": timezone.now()},
+        )
```
snopoke: I'm wondering about whether we should create new notifications or reuse the old ones. If we create new ones (as long as the old one is unread) then there will be some history of notifications otherwise if we reuse the same one each time we lose the history (unless we store history separately which could also work well).

apps/ocs_notifications/views.py
```diff
@@ -0,0 +1,126 @@
...
+    preferences, created = UserNotificationPreferences.objects.get_or_create(user=request.user)
+
+    form = NotificationPreferencesForm(request.POST, instance=preferences)
+    if form.is_valid():
+        form.save()
+        bust_unread_notification_cache(request.user.id)
+        messages.success(request, _("Notification preferences saved successfully."))
+        return redirect(reverse("users:user_profile"))
+    else:
+        form = NotificationPreferencesForm(instance=preferences)
```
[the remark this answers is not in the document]
SmittieC: I think its fine

apps/users/views.py
```diff
@@ -141,3 +148,17 @@ def revoke_oauth_token(request):
...
+def save_notification_preferences(request):
+    """Save notification preferences from the profile page"""
+    preferences, created = UserNotificationPreferences.objects.get_or_create(user=request.user)
+    form = NotificationPreferencesForm(request.POST, instance=preferences)
+    if form.is_valid():
+        form.save()
+        messages.success(request, _("Notification preferences saved successfully."))
+    else:
+        messages.error(request, _("Failed to save notification preferences."))
+    return HttpResponseRedirect(reverse("users:user_profile"))
```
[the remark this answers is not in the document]
SmittieC: Please check if this is resolved

apps/users/views.py
```diff
@@ -54,6 +56,10 @@ def profile(request):
 
     new_api_key = request.session.pop(SESSION_API_KEY, None)
 
+    # Get or create notification preferences
+    preferences, created = UserNotificationPreferences.objects.get_or_create(user=request.user)
+    notification_preferences_form = NotificationPreferencesForm(instance=preferences)
+
```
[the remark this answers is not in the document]
SmittieC: Please check if this is resolved

apps/ocs_notifications/utils.py
```diff
@@ -0,0 +1,138 @@
...
+            user_notification.read_at = None
+            user_notification.save()
+
+            if should_send_email:
+                send_notification_email(user_notification)
+
+    except Exception:
+        logger.exception("Failed to create notification")
+
+    return notification
```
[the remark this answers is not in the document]
SmittieC: Please check if this is resolved

apps/ocs_notifications/utils.py
snopoke: Discussed offline and decided that most places have history already e.g. traces, collection sync logs etc.

So we can link through to the relevant place.

apps/ocs_notifications/tests/test_utils.py
```diff
@@ -0,0 +1,204 @@
...
+    create_notification,
+    get_user_notification_cache_value,
+    send_notification_email,
+    set_user_notification_cache,
+)
+from apps.utils.factories.notifications import UserNotificationFactory
+
+
+@patch("apps.ocs_notifications.utils.cache.get")
+def test_get_user_notification_cache_value(mock_cache_get):
```
snopoke: These basic cache tests don't seem that useful

apps/ocs_notifications/tests/test_utils.py
```diff
@@ -0,0 +1,204 @@
...
+    user_id = 789
+    bust_unread_notification_cache(user_id, team_slug="test-team")
+
+    expected_key = CACHE_KEY_FORMAT.format(user_id=user_id, team_slug="test-team")
+    mock_cache_delete.assert_called_once_with(expected_key)
+
+
+@pytest.mark.django_db()
+@patch("apps.ocs_notifications.utils.render_to_string")
+@patch("apps.ocs_notifications.utils.send_mail")
```
snopoke: fine to patch these but just FYI there is a [mailoutbox](https://pytest-django.readthedocs.io/en/latest/helpers.html#mailoutbox) fixture that you can also use in tests

apps/ocs_notifications/tests/test_utils.py
```diff
@@ -0,0 +1,204 @@
...
+
+        create_notification(
+            title="Test Notification", message="Test message", level=LevelChoices.INFO, team=team_with_users
+        )
+
+        assert UserNotification.objects.filter(user=user).count() == 1
+
+    @patch("apps.ocs_notifications.utils.bust_unread_notification_cache", wraps=bust_unread_notification_cache)
+    @patch("apps.ocs_notifications.utils.send_notification_email", wraps=send_notification_email)
+    def test_user_is_notified_again(self, mock_send_email, mock_bust_cache, team_with_users):
```
snopoke: I know you are testing the views as well but if you extract utility functions for things like 'get_notification_count' and 'mark_notification_read' then you could test these without having to manipulate the low level cache and notification objects directly.

apps/ocs_notifications/tests/test_utils.py
```diff
@@ -0,0 +1,204 @@
...
+        mock_bust_cache.assert_called()
+
+        assert user.notifications.count() == 1
+        assert UserNotification.objects.filter(user=user).count() == 1
+        user_notification.refresh_from_db()
+        assert user_notification.read is False, "UserNotification should be marked as unread again"
+
+    @patch("apps.ocs_notifications.utils.bust_unread_notification_cache", wraps=bust_unread_notification_cache)
+    @patch("apps.ocs_notifications.utils.send_notification_email", wraps=send_notification_email)
+    def test_create_notification_with_and_without_event_data(self, mock_send_email, mock_bust_cache, team_with_users):
```
snopoke: This feels like it could just be testing the `create_identifier` function

apps/ocs_notifications/utils.py
```diff
@@ -0,0 +1,170 @@
...
+    try:
+        identifier = create_identifier(event_data)
+        notification, created = Notification.objects.update_or_create(
+            team=team,
+            title=title,
+            message=message,
+            level=level,
+            identifier=identifier,
+            defaults={"last_event_at": timezone.now()},
+        )
```
snopoke: Is it intentional to make title and message part of the main query? If we change a message or title in the future this will create a new notification.

I think what we should do is enforce a 'slug' argument in `create_notification` which we use along with event data to create the identifier. Then the uniqueness should only based on `slug` and the `event_data` (and not on the title / message).

apps/ocs_notifications/views.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class ToggleNotificationReadView(LoginRequiredMixin, SingleTableView):
+    model = UserNotification
+    table_class = UserNotificationTable
+    template_name = "table/single_table.html"
+
+    def post(self, request, *args, **kwargs):
+        notification_id = kwargs.get("notification_id")
+        user_notification = get_object_or_404(UserNotification, id=notification_id, user=request.user)
```
snopoke: missing team filter

apps/users/models.py
```diff
@@ -50,3 +53,31 @@ def avatar_url(self) -> str:
...
+        count = get_user_notification_cache_value(self.id, team_slug=team.slug)
+        if count is not None:
+            return count
+
+        preferences, _created = UserNotificationPreferences.objects.get_or_create(user=self, team=team)
+        if preferences.in_app_enabled:
+            level = preferences.in_app_level
+            count = UserNotification.objects.filter(
+                team__slug=team.slug, user_id=self.id, read=False, notification__level__gte=level
+            ).count()
```
snopoke: See my comment above about making the `_level` fields integers. As it stands this is comparing the values using lexicographical order since they are strings.

templates/ocs_notifications/components/read_button.html
```diff
@@ -0,0 +1,21 @@
...
+    title="Mark as unread"
+  >
+    Mark Unread
+  </button>
+{% else %}
+  <button
+    class="btn btn-sm"
+    hx-post="{% url 'ocs_notifications:toggle_notification_read' record.id %}{% querystring %}"
+    hx-swap="innerHTML"
+    hx-target="#data-table"
```
snopoke: could you just target the current row instead of reloading the whole table?

templates/ocs_notifications/home.html
snopoke: I don't think this override template is necessary since there are no trends in the notifications table

apps/ocs_notifications/views.py
```diff
@@ -0,0 +1,126 @@
...
+    form = NotificationPreferencesForm(request.POST, instance=preferences)
+    if form.is_valid():
+        form.save()
+        bust_unread_notification_cache(request.user.id, team_slug=request.team.slug)
+        messages.success(request, _("Notification preferences saved successfully."))
+        return redirect(reverse("users:user_profile"))
+
+    return render(
+        request,
+        "ocs_notifications/notification_preferences_form.html",
```
snopoke: This template doesn't exist

snopoke (commented):
Hey Chris. I think this still has a bit of work to do. The tests in particular could use a review.

apps/ocs_notifications/models.py
```diff
@@ -0,0 +1,74 @@
...
+    def __str__(self):
+        return f"{self.user} - {self.notification.title}"
+
+
+class UserNotificationPreferences(BaseTeamModel):
+    """Store user preferences for in-app and email notifications"""
+
+    user = models.OneToOneField("users.CustomUser", on_delete=models.CASCADE, related_name="notification_preferences")
+
+    # In-app notification preferences
```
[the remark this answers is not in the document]
SmittieC: Thanks. See https://github.com/dimagi/open-chat-studio/pull/2717/commits/71274271145034d3e776559b7de129229c020f4a

apps/ocs_notifications/tests/test_utils.py
```diff
@@ -0,0 +1,187 @@
...
+        (LevelChoices.ERROR, LevelChoices.INFO, True),
+        # When preference is WARNING, only send WARNING and ERROR
+        (LevelChoices.INFO, LevelChoices.WARNING, False),
+        (LevelChoices.WARNING, LevelChoices.WARNING, True),
+        (LevelChoices.ERROR, LevelChoices.WARNING, True),
+        # When preference is ERROR, only send ERROR
+        (LevelChoices.INFO, LevelChoices.ERROR, False),
+        (LevelChoices.WARNING, LevelChoices.ERROR, False),
+        (LevelChoices.ERROR, LevelChoices.ERROR, True),
+    ],
```
snopoke: This is very thorough but also feels overkill, especially considering these are DB tests.

Maybe just these ones is enough:
* ERROR, ERROR, True
* INFO, WARNING, False
* WARNING, INFO, True

apps/ocs_notifications/utils.py
SmittieC: [Here's](https://github.com/dimagi/open-chat-studio/pull/2717/commits/e1d5151bea7cd58bbbc1a0663b0950c49cbf4fc9) the slug approach.

apps/ocs_notifications/models.py
```diff
@@ -0,0 +1,74 @@
...
+from apps.teams.models import BaseTeamModel
+
+
+class LevelChoices(models.IntegerChoices):
+    INFO = 0, "Info"
+    WARNING = 1, "Warning"
+    ERROR = 2, "Error"
+
+
+class Notification(BaseTeamModel):
```
snopoke: It's interesting that you're not saving the event data. I guess it's only current use to to generate a unique ID.

apps/ocs_notifications/tables.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+    level = columns.TemplateColumn(
+        template_name="ocs_notifications/components/level_badge.html",
+        accessor="notification__level",
+        verbose_name="Level",
+        orderable=False,
+    )
+    read = columns.TemplateColumn(
+        template_name="ocs_notifications/components/read_button.html", verbose_name="Read Status", orderable=True
+    )
```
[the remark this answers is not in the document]
SmittieC: I'm not against this approach, since it is also using an existing pattern. The one reason I'd like to keep things as-is is due to the fact that users can reorder the table based on the read status

apps/ocs_notifications/tests/test_views.py
```diff
@@ -0,0 +1,162 @@
...
+            team=team_with_users,
+            read=False,
+            read_at=None,
+        )
+        notification_id = user_notification.id
+
+        # Login user and set team in session
+        client.force_login(user)
+        session = client.session
+        session["team"] = team_with_users.id
```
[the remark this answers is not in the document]
snopoke: FYI this also applies to other tests

snopoke (dismissed):
nothing blocking

PR #2769, opened 2026-02-04
prevent evals sessions from triggering events

snopoke:
### Technical Description
We shouldn't fire and process events for evals sessions. This adds unnecessary load on the system and there is no way to evaluate the outcomes from those events.

### DIFF
--- a/apps/events/models.py
+++ b/apps/events/models.py
@@ -293,7 +293,7 @@ class TimeoutTrigger(BaseModel, VersionsMixin):
                 ended_at=None,
             )
             .exclude(status__in=STATUSES_FOR_COMPLETE_CHATS)
-            .exclude(platform=ChannelPlatform.EVALUATIONS)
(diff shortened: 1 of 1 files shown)


apps/chat/channels.py
```diff
@@ -402,7 +402,8 @@ def _new_user_message(self, span: TraceContext) -> ChatMessage:
                     # status is ACTIVE
                     self.experiment_session.update_status(SessionStatus.ACTIVE)
 
-            enqueue_static_triggers.delay(self.experiment_session.id, StaticTriggerType.NEW_HUMAN_MESSAGE)
+            if self.experiment_channel.platform != ChannelPlatform.EVALUATIONS:
```
SmittieC: Possibly a bad idea in terms of separation of concerns, but what if we ignore the `evaluations` channel inside the events framework entirely?

Are there scenarios you can think of where we might want to have events firing when doing evals? (I guess this depends on the scope of the evaluation)

PR #2787, opened 2026-02-06
Chat API progress messages

snopoke:
### Technical Description

Part 1 of https://github.com/dimagi/open-chat-studio/issues/2774

Companion to https://github.com/dimagi/open-chat-studio/pull/2798

The primary change here is to update the Chat API poll endpoint to return progress messages in its response while the task is still in progress.

There will be a corresponding change to the chat widget to display these messages.

### System agent change

This is the 2nd use of 'system agents' so I took the opportunity to refactor the code to create a reusable system agent that includes retry and fallback configuaration.

I have updated the python node help utils to use this new system agent.

**NOTE**: These changes require updates to the env of deployed applications.


SmittieC (approved):
Looks fine to me. I'm slightly concerned about the session based key though. Wouldn't this blow up the cache?

PR #2812, opened 2026-02-10
Add file attachment support to CodeNode via HTTP client

snopoke:
### Technical Description

Add file attachment support to CodeNode via the HTTP client, enabling sandboxed Python code to fetch files over HTTP and attach them to AI response messages.

**Changes:**

1. **`RestrictedHttpClient` response dict** — Added `content` field exposing raw response bytes. Text decoding is skipped for binary content types (images, PDFs, etc.) to avoid doubling memory usage. JSON parsing now uses bytes directly.

2. **`CodeNode.add_file_attachment` helper** — New function exposed to sandboxed code:
   ```python
   add_file_attachment(filename, content, content_type=None)
   ```
   - Creates a `File` via `File.create()` with auto content-type detection
   - Links the file to the chat via `ChatAttachment` with `tool_type="code_interpreter"`
   - Tracks file IDs in `output_message_metadata["generated_files"]`
   - Validates that `content` is bytes and an active session exists

**Example usage in a CodeNode:**
```python
def main(input: str, **kwargs) -> str:
    response = http.get("https://example.com/report.pdf")
    if response["is_success"]:
        add_file_attachment("report.pdf", response["content"])
    return "Here's your report"
```

### Demo
N/A — backend-only change with no UI impact.

### Docs and Changelog

New `add_file_attachment` helper for CodeNode and new `content` field on HTTP responses should be documented in the Python node / HTTP client reference.

### DIFF
--- a/apps/pipelines/tests/test_code_node.py
+++ b/apps/pipelines/tests/test_code_node.py
@@ -499,21 +499,3 @@ def main(input, **kwargs):
                 node_inputs=["hi"],
             )
         )
-
-
-def test_add_file_attachment_requires_session():
-    code = """
-def main(input, **kwargs):
-    add_file_attachment("test.txt", b"hello", "text/plain")
-    return input
-"""
-    node = CodeNode(name="test", node_id="123", django_node=None, code=code)
-    with pytest.raises(CodeNodeRunError, match="Cannot attach files without an active session"):
-        node._process(
-            PipelineState(
-                outputs={},
-                experiment_session=None,
-                last_node_input="hi",
-                node_inputs=["hi"],
-            )
-        )


apps/utils/restricted_http.py
```diff
@@ -276,18 +276,23 @@ def _do_request(self, httpx_kwargs):
 
                 return self._build_response_dict(response, body=body)
 
+    _TEXT_CONTENT_TYPES = ("text/", "json", "xml", "javascript", "yaml", "csv", "html")
```
SmittieC: Nit: Can this be at the top of the file?

SmittieC (approved):
LGTM

apps/utils/restricted_http.py
snopoke: It could but it only has this one use so seems fine to have it scoped to the class and declared near the usage

PR #2827, opened 2026-02-13
Notifications rework

### LINKED ISSUE #2754: Allow users to mute notification
This could be indefinitely or temporarily (for 8 hours / 1 day etc)

SmittieC:
### Technical Description

Resolves #2754

- Changed the models backing notifications to handle notification/event history better. This translates into users being able to see past notifications/events. (This PR includes changes from a previous PR that was originally built using the old models, so the code changes quite a bit if you follow it commit-wise. Best to not review it commit-wise)
- Users can now also mute notifications, either on a per-event basis or all by turning "do not disturb" on

### Migrations

The code currently running in prod does not use any of the models introduced here.

### Demo

https://www.loom.com/share/e7998829358e49bbade186bdf8de47f3

### Docs and Changelog

### DIFF
--- a/apps/generics/tables.py
+++ b/apps/generics/tables.py
@@ -55,7 +55,9 @@ class ISOTimeAgoColumn(TimeAgoColumn):
     """A TimeAgoColumn that parses ISO datetime strings before rendering."""
 
     def render(self, value, **kwargs):
-        return super().render(value=parse_datetime(value), **kwargs)
+        if isinstance(value, str):
+            value = parse_datetime(value)
+        return super().render(value=value, **kwargs)
 
 
 class ArrayColumn(tables.Column):
--- a/apps/ocs_notifications/tables.py
+++ b/apps/ocs_notifications/tables.py
@@ -7,7 +7,6 @@ from apps.ocs_notifications.models import EventUser, NotificationEvent
 
 
 class UserNotificationTable(tables.Table):
-    # TODO: Rename table
     notification_content = columns.TemplateColumn(
         template_name="ocs_notifications/components/notification_content.html",
         verbose_name="Notification",


SmittieC:
@coderabbitai full review

apps/users/models.py
```diff
@@ -72,8 +72,11 @@ def unread_notifications_count(self, team: Team) -> int:
...
         if preferences.in_app_enabled:
             level = preferences.in_app_level
-            count = UserNotification.objects.filter(
-                team__slug=team.slug, user_id=self.id, read=False, notification__level__gte=level
+            count = EventUser.objects.filter(
+                team__slug=team.slug,
+                user_id=self.id,
+                read=False,
+                event_type__level__gte=level,
             ).count()
```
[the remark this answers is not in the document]
SmittieC: I think this is fine, since we don't want to "hide" already created notifications, but only prevent new ones from being created when it is muted

apps/ocs_notifications/utils.py
```diff
@@ -30,62 +52,110 @@ def create_notification(
...
-        if not permissions:
-            return True
-        return member.has_perms(permissions)
+    links = links or {}
 
-    users = [
-        member.user for member in team.membership_set.select_related("user").all() if _can_receive_notification(member)
-    ]
+    user_info = get_users_to_be_notified(team, permissions)
+    users = list(user_info.keys())
```
snopoke: nit: this is the same as `list(user_info)`

apps/ocs_notifications/utils.py
```diff
@@ -30,62 +52,110 @@ def create_notification(
...
-        )
-        # Uuser will only be notified when notification is created or if the notification was previously read
-        user_should_be_notified = created or user_notification.read is True
-        if user_notification.read is True:
-            user_notification.read = False
-            user_notification.read_at = None
-            user_notification.save()
-
-        # Bust cache when notification is created or when marking previously read notification as unread
+        if is_notification_muted(user, team, event_type):
```
snopoke: this does a re-check of the notification preferences for the do-not-disturb field which was already done in the user query.

I don't see uses anywhere else except tests so maybe we can drop that query.

apps/ocs_notifications/utils.py
```diff
@@ -208,14 +219,65 @@ def toggle_notification_read(user, user_notification: UserNotification, read: bo
...
+
+    if UserNotificationPreferences.objects.filter(user=user, team=team, do_not_disturb_until__gt=now).exists():
+        return True
+
+    return EventUser.objects.filter(
+        muted_until__gt=now,
+        user=user,
+        team=team,
+        event_type=event_type,
+    ).exists()
```
snopoke: I think this could be done as a bulk query instead of once per user e.g. get user_ids that aren't muted.

You could do it as part of `get_users_to_be_notified`

apps/ocs_notifications/utils.py
```diff
@@ -30,62 +52,110 @@ def create_notification(
...
+            event_user.read_at = None
+            event_user.save(update_fields=["read", "read_at"])
+
+        # Bust cache when notification is created or when marking previously read notification as unread.
         if user_should_be_notified:
+            # Busting the cache so that the UI can show the updated unread notifications count immediately
             bust_unread_notification_cache(user.id, team_slug=team.slug)
-            send_notification_email(user_notification)
+            if should_send_email(user_info[user], event_level=level):
+                send_notification_email_async.delay(event_user.id, notification_event_id=notification_event.id)
```
snopoke: **Performance: N+1 queries in per-user loop**

This loop runs 2-4 DB queries per team member:
- `is_notification_muted()` → 1-2 queries (DND check + EventUser mute check)
- `EventUser.objects.get_or_create()` → 1 query
- `event_user.save()` → 1 query (conditional)

For a team with 50 members, that's 100-200 queries per notification call.

Consider:
- Bulk-fetch existing `EventUser` rows with a single query before the loop
- Batch the mute check: fetch all muted EventUser rows + DND preferences for the user set in one query
- Use `bulk_create` / `bulk_update` instead of per-row `get_or_create` + `save`

apps/ocs_notifications/utils.py
snopoke: **Redundant DND check**

`get_users_to_be_notified()` already filters out users with active DND via `.filter(do_not_disturb=False)` (line 144). But `is_notification_muted()` checks DND again here for each user individually — that branch will never be true for users returned by `get_users_to_be_notified()`.

The per-event-type mute check is still needed, but the DND portion of `is_notification_muted()` is redundant in this context and costs an extra query per user.

apps/ocs_notifications/utils.py
```diff
@@ -30,62 +52,110 @@ def create_notification(
...
+            "email_enabled": member.email_enabled,
+            "email_level": member.email_level,
+        }
+        for member in members_qs
+        if _is_notification_target(member)
+    }
+
+
+def should_send_email(user_email_info: dict, event_level) -> bool:
+    return user_email_info["email_enabled"] and event_level >= user_email_info["email_level"]
```
snopoke: **Fragile None handling**

When a user has no `UserNotificationPreferences` row, the subquery annotations in `get_users_to_be_notified()` return `None` for both `email_enabled` and `email_level`. This works today because `None` is falsy and short-circuits the `and`, but if someone reorders the conditions, `event_level >= None` would raise `TypeError`.

Consider using `Coalesce` in the subqueries to provide defaults, or adding an explicit `None` guard here.

apps/ocs_notifications/utils.py
```diff
@@ -30,62 +52,110 @@ def create_notification(
...
-        slug (str): A slug to identify the notification type. Used with event_data for uniqueness.
-        event_data (dict, optional): Additional data to store with the notification. Combined with slug for uniqueness.
+        slug (str): A slug to identify the event type. Used with event_data for uniqueness.
+        event_data (dict, optional): Additional data to store with the event type. Combined with slug for uniqueness.
 
     Returns:
-        Notification: The created Notification instance, or None if creation failed.
+        NotificationEvent: The created NotificationEvent instance, or None if creation failed.
     """
     if not _notifications_flag_is_active(team):
```
snopoke: **Performance: Feature flag DB query on every notification**

`_notifications_flag_is_active()` does a DB query on every `create_notification()` call. In hot paths (e.g., LLM errors during high traffic), this adds unnecessary latency since feature flags change very infrequently.

Consider a short-TTL cache (30-60s) on the flag lookup.

apps/ocs_notifications/utils.py
```diff
@@ -208,14 +219,65 @@ def toggle_notification_read(user, user_notification: UserNotification, read: bo
...
+        event_user.read = True
+        event_user.read_at = timezone.now()
     else:
-        user_notification.read = False
-        user_notification.read_at = None
-    user_notification.save()
-    bust_unread_notification_cache(user.id, team_slug=user_notification.team.slug)
+        event_user.read = False
+        event_user.read_at = None
+    event_user.save()
```
snopoke: **Nit: Missing `update_fields`**

`event_user.save()` without `update_fields` writes all fields. The other save calls in this module correctly use `update_fields` (e.g., lines 101, 272). This should be `event_user.save(update_fields=["read", "read_at"])` for consistency and to avoid unnecessary column writes.

apps/ocs_notifications/models.py
```diff
@@ -69,10 +71,90 @@ class UserNotificationPreferences(BaseTeamModel):
...
+        ]
+
+
+class NotificationEvent(BaseTeamModel):
+    # TODO: rate limiting?
+    # Notification cleanup after a certain period of time?
+    event_type = models.ForeignKey(EventType, on_delete=models.CASCADE)
+    title = models.CharField(max_length=255)
+    message = models.TextField()
+    links = models.JSONField(null=True)
```
snopoke: **Inconsistent JSONField type**

`links` uses `models.JSONField(null=True)` while `EventType.event_data` uses `SanitizedJSONField(default=dict, blank=True)`. If `links` values are rendered in templates without escaping, this inconsistency could lead to XSS. Consider using `SanitizedJSONField` here too for consistency.

snopoke (commented):
Some additional comments from me directly.

apps/ocs_notifications/models.py
```diff
@@ -69,10 +71,90 @@ class UserNotificationPreferences(BaseTeamModel):
...
+            super()
+            .get_queryset()
+            .annotate(
+                is_muted=models.Case(
+                    models.When(muted_until__gt=timezone.now(), then=True),
+                    default=False,
+                    output_field=models.BooleanField(),
+                )
+            )
+        )
```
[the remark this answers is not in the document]
SmittieC: https://github.com/dimagi/open-chat-studio/pull/2827/commits/f36ac36459cb9864130b931a11ea50b3962de649

apps/ocs_notifications/models.py
```diff
@@ -69,10 +71,90 @@ class UserNotificationPreferences(BaseTeamModel):
...
+class EventUser(BaseTeamModel):
+    event_type = models.ForeignKey(EventType, on_delete=models.CASCADE)
+    user = models.ForeignKey("users.CustomUser", on_delete=models.CASCADE)
+    read = models.BooleanField(default=False, db_index=True)
+    read_at = models.DateTimeField(null=True)
+    objects = EventUserManager()
+    muted_until = models.DateTimeField(null=True, help_text="When the mute expires")
+
+    class Meta:
+        unique_together = ("event_type", "user")
```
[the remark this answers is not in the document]
SmittieC: I've updated one model [here](https://github.com/dimagi/open-chat-studio/pull/2827/commits/54a19082f778a67ff6293d77d99ad990dab1b4cc). The other one will be removed in a followup PR

apps/ocs_notifications/views.py
```diff
@@ -71,17 +95,135 @@ def get_queryset(self):
...
+    template_name = "generic/object_home.html"
+
+    def get_context_data(self, **kwargs):
+        event_type = get_object_or_404(EventType, team=self.request.team, id=self.kwargs["event_type_id"])
+        table_url = reverse("ocs_notifications:notification_event_table", args=[self.request.team.slug, event_type.id])
+
+        context = {
+            "active_tab": "notifications",
+            "title": "Notifications",
+            "subtitle": event_type.notificationevent_set.first().title,
```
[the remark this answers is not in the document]
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/2827/commits/637197bc4109289ac4e4f830e3aea670af2f6659 for the ordering and https://github.com/dimagi/open-chat-studio/pull/2827/commits/9f3b1cb9c3fa9040b87b590012341af3df6cac05

PR #2839, opened 2026-02-17
refactor: pass objects instead of IDs to llm_error_notification

### LINKED ISSUE #2810: Refactor notification functions to accept objects instead of IDs
## Context
In PR #2790, @snopoke noted that some notification functions (e.g., `llm_error_notification`) re-fetch experiment and session objects by ID, even though these objects are already available in the calling context.

## Suggested Improvement
Refactor notification functions to accept the full objects instead of IDs. Since these objects are passed in initially (e.g., in tracing), we can avoid the extra database queries.

## Example
`llm_error_notification` currently accepts `experiment_id: int, session_id: int` and re-fetches the objects. It could instead accept `experiment: Experiment, session: ExperimentSession`.

## Affected Functions
- `llm_error_notification`
- Potentially other notification functions that follow a similar pattern

## References
- PR: https://github.com/dimagi/open-chat-studio/pull/2790
- Discussion: https://github.com/dimagi/open-chat-studio/pull/2790#discussion_r2787699461
- Requested by: @SmittieC

SmittieC:
Avoid unnecessary database queries by passing the already-loaded Experiment and ExperimentSession objects directly instead of their IDs.

### Technical Description

 Resolves #2810

  llm_error_notification accepted experiment_id: int and session_id: int, then re-fetched both objects from the database. The full objects are already available in the calling context (`OCSTracer`), so these extra queries are unnecessary. This was the only notification function with this problem — all others already accept full objects.

### Demo

N/A

### Docs and Changelog

### DIFF
--- a/apps/service_providers/tracing/ocs_tracer.py
+++ b/apps/service_providers/tracing/ocs_tracer.py
@@ -16,7 +16,7 @@ from apps.trace.models import Trace, TraceStatus
 from .base import TraceContext, Tracer
 
 if TYPE_CHECKING:
-    from apps.experiments.models import ExperimentSession
+    from apps.experiments.models import Experiment, ExperimentSession
 
 logger = logging.getLogger("ocs.tracing")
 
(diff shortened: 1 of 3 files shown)


apps/service_providers/tracing/ocs_tracer.py
```diff
@@ -27,11 +27,14 @@ class OCSTracer(Tracer):
     """
 
     def __init__(self, experiment_id: int, team_id: int):
+        from apps.experiments.models import Experiment
+
         super().__init__(OCS_TRACE_PROVIDER, {})
-        self.experiment_id = experiment_id
+        self.experiment = Experiment.objects.get(id=experiment_id)
```
snopoke: I don't think you quite followed. In [TracingService.create_from_experiment](https://github.com/dimagi/open-chat-studio/blob/main/apps/service_providers/tracing/service.py#L65) we have the full experiment object so we should just use it instead of just extracting the ID at that point.

PR #2845, opened 2026-02-18
Model Cleanup

### LINKED ISSUE #2837: Notifications cleanup task
Add a task that runs daily to clean up notifications older than 3 months

SmittieC:
### Technical Description

Resolves #2837 

Remove previous models that notifications were built on. Also fixes an issue where the subquery returned more than 1 result by limiting the result set to 1.

### Migrations

This is backwards compatable, since the code currently running uses the new models. The models being removed are not used anymore

### Docs and Changelog

### DIFF
--- a/apps/teams/backends.py
+++ b/apps/teams/backends.py
@@ -104,8 +104,6 @@ CONTENT_TYPES = {
     "mcp_integrations": ["mcpserver"],
     "oauth": ["oauth2application", "oauth2accesstoken", "oauth2grant", "oauth2idtoken", "oauth2refreshtoken"],
     "ocs_notifications": [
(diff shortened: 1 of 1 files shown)


apps/ocs_notifications/migrations/0005_delete_notification.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 5.1.15 on 2026-02-18 11:08
```
SmittieC: I had to remove the UserNotification model in a separate transaction than the Notification one, otherwise it errors while removing the Notification model. I can't say I am exactly sure why this is

Issue #2843, opened 2026-02-18
Implement type checking


snopoke:
Implement type checking for the project.

PR #2856, opened 2026-02-18
Type checking and fixes II

snopoke:
### Technical Description
Part 2 of https://github.com/dimagi/open-chat-studio/issues/2843

Rule: unsupported-operator
Violations: 16
Approach: Fixed bad type annotation, added None guards, suppressed Django admin false
  positives
────────────────────────────────────────
Rule: invalid-return-type
Violations: 32
Approach: Fixed real bugs (missing returns, wrong return types, return NotImplemented),
  wrapped and-chains in bool(), suppressed library false positives
────────────────────────────────────────
Rule: possibly-missing-attribute
Violations: 44
Approach: Added import factory.django to 15 factory files, replaced implicit submodule
  access in migrations

### DIFF
--- a/apps/annotations/models.py
+++ b/apps/annotations/models.py
@@ -5,6 +5,7 @@ from typing import ClassVar
 
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import FieldDoesNotExist
 from django.db import models, transaction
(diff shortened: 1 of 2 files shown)


apps/utils/factories/traces.py
```diff
@@ -1,4 +1,5 @@
 import factory
+import factory.django
```
SmittieC: why is it importing these?

PR #2874, opened 2026-02-20
Sk/port

snopoke:
### Technical Description
I've started using https://port1355.dev/ locally to avoid port conflicts. This change allows Django to use the port assigned by portless.

An added benefit is that different servers don't share cookies since they are using different domain names.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,6 +1,10 @@
 #!/usr/bin/env python
 import os
+import re
 import sys
+
+
+_PORT_ARG_RE = re.compile(r"^\d+$|^[\w.]+:\d+$")
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


manage.py
```diff
@@ -12,4 +12,11 @@
             "available on your PYTHONPATH environment variable? Did you "
             "forget to activate a virtual environment?"
         ) from exc
+    port = os.environ.get("PORT")
+    if port:
+        # If the runserver command is used, insert the port into the arguments
+        if "runserver" in sys.argv:
+            # Remove any existing port argument first
+            sys.argv = [arg for arg in sys.argv if not arg.isdigit()]
```
snopoke:
```suggestion
            sys.argv = [arg for arg in sys.argv if not _PORT_ARG_RE.match(arg)]
```

PR #2890, opened 2026-02-23
Send email only after transaction commits

SmittieC:
### Technical Description

This should fix [this issue](https://dimagi.sentry.io/issues/7285748030/?alert_rule_id=14063560&alert_type=issue&notification_uuid=62d1209a-68af-40ff-8017-bfd1d3280033&project=4505001320316928&referrer=issue_alert-slack). This was a subtle one. The record ID existed (I can see it in Sentry), but when the task executed, it couldn't see the record.

I think what happened is the fact that the object was created, but the transaction in which the code was called did not yet commit, so the async celery task tried to find it and couldn't.

An additional fix is to only remove `NotificationEvent` records after 90 days, not the `EventType`. Removing the `EventType` will remove all `NotificationEvent` records linked to that.

### Docs and Changelog

### DIFF
--- a/apps/ocs_notifications/tasks.py
+++ b/apps/ocs_notifications/tasks.py
@@ -11,7 +11,6 @@ from django.urls import reverse
 from django.utils import timezone
 
 from apps.ocs_notifications.models import (
-    EventType,
     NotificationEvent,
 )
 from apps.web.meta import absolute_url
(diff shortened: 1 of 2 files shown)


SmittieC:
@coderabbitai full review

PR #2925, opened 2026-02-26
perf: eliminate N+1 API calls for chatbot trend sparklines

snopoke:
### Technical Description

The chatbot table's **Trends** column previously made one HTTP fetch per row to `experiments:trends_data`, each of which issued a separate DB query. With N chatbots on the page, that's N browser requests + N DB queries on every page load.

**Fix:** pre-compute all trend data server-side in a single bulk DB query, then embed it as inline JSON in each row's `data-trends` attribute. The browser reads from the attribute directly — no extra requests at all.

Key changes:
- `Experiment.get_bulk_trend_data(experiment_ids)` — new classmethod that fetches hourly success/error counts for all experiments in one query, grouping by `(root_experiment_id, hour_bucket)` to correctly aggregate across all versions of each experiment
- `ChatbotExperimentTableView.get_table_data()` — evaluates the queryset, calls `get_bulk_trend_data` once, attaches `trend_data` to each object
- `ChatbotTable.render_trends()` — reads `record.trend_data` and serialises to JSON for the template
- Template reads from `data-trends` attribute; JS `renderChart()` renders directly from inline data
- Also: improved chart appearance (fixed sizing with `maintainAspectRatio: false`, x-axis baseline, better colours), reduced time window from 48h → 24h
- Fixed `ty` pre-commit hook by pointing it at the project venv (`python = ".venv"` in `pyproject.toml`)

**Result:** N chatbots = 0 extra HTTP requests + 1 DB query (was N + N).

### Demo

Requires `flag_tracing` waffle flag to be enabled to see the Trends column.

### Docs and Changelog

### DIFF
--- a/assets/javascript/trends.js
+++ b/assets/javascript/trends.js
@@ -58,20 +58,4 @@ export const renderChart = (ctx, data) => {
     });
 };
 
-/**
- * Creates a minimal sparkline chart by fetching data from a URL.
- * @deprecated Use renderChart with inline data instead to avoid per-row HTTP requests.
- * @param {CanvasRenderingContext2D} ctx - Canvas 2D context to render the chart
- * @param {string} dataUrl - URL endpoint that returns JSON data with a trends object
- */
-export const trendsChart = (ctx, dataUrl) => {
-    return fetch(dataUrl)
-        .then(response => response.json())
-        .then(data => renderChart(ctx, data.trends))
-        .catch(error => {
-            console.error('Error loading chart data:', error);
-            return null;
-        });
-};
-
-export default { renderChart, trendsChart };
+export default { renderChart };


apps/experiments/models.py
```diff
@@ -717,11 +717,10 @@ def get_trend_data(self) -> tuple[list, list]:
         """
         Get the error/success trends across all versions in this experiment's version family.
         Returns two lists: successes and errors, each containing the count of successful and error traces
-        for each hour in the last 48 hours.
+        for each hour in the last 24 hours.
         """
-        days = 2
         to_date = timezone.now()
-        from_date = to_date - timezone.timedelta(days=days)
+        from_date = to_date - timezone.timedelta(hours=24)
```
snopoke: I dropped this to 24 hours to make the graph more readable.

SmittieC (approved):

PR #2965, opened 2026-03-03
Add get_filter_options tool to FilterAgent for runtime option lookup

snopoke:
### Technical Description

Adds a `get_filter_options` LangChain tool to `FilterAgent` so the agent can look up valid option values for `choice`/`exclusive_choice` filter parameters at runtime. Previously the agent received only the static filter schema (types, operators) with no knowledge of actual options — meaning it had to guess chatbot names, tag values, or channel names. Now it can call the tool to resolve a user's natural-language term to the correct option ID.

**How it works:**
- `FilterInput` gains a `team_id` field, injected server-side by the view from `request.team.id` (never from the client)
- `make_get_options_tool(filter_class, team)` returns a LangChain `@tool` closed over the filter class and team. It calls `ChoiceColumnFilter.prepare(team)` on demand, normalises options to `{id, label}` format, supports case-insensitive search, and caps results at a configurable limit (default 50) while reporting `total` so the LLM knows if results are truncated
- `FilterAgent.run()` is overridden to build the tool and register it with `build_system_agent`
- `filter_system_prompt.md` gets an `## Available Tools` section instructing the LLM when and how to use the tool

**Key design decisions:**
- `langchain_core.tools` is lazy-loaded inside `make_get_options_tool()` per project convention for heavy AI/ML imports
- `team_id` injection at the view layer means no auth changes are needed in the agent — the team is already authenticated before the view runs

### Migrations

N/A — no model changes.

### Demo

https://github.com/user-attachments/assets/151669e6-b508-4426-9614-255190a391c7

### Docs and Changelog

### DIFF
--- a/apps/help/agents/filter.py
+++ b/apps/help/agents/filter.py
@@ -12,7 +12,7 @@ from apps.help.agent import build_system_agent
 from apps.help.base import BaseHelpAgent
 from apps.help.registry import register_agent
 from apps.teams.models import Team
-from apps.web.dynamic_filters.base import ChoiceColumnFilter
+from apps.web.dynamic_filters.base import ChoiceColumnFilter, get_filter_registry, get_filter_schema
 from apps.web.dynamic_filters.datastructures import ColumnFilterData
 
 logger = logging.getLogger(__name__)
@@ -118,8 +118,6 @@ class FilterAgent(BaseHelpAgent[FilterInput, FilterOutput]):
 
     @classmethod
     def get_system_prompt(cls, input: FilterInput) -> str:
-        from apps.web.dynamic_filters.base import get_filter_registry, get_filter_schema
-
         registry = get_filter_registry()
         filter_class = registry.get(input.filter_slug)
         if filter_class is None:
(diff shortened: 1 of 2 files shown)


apps/help/filter_system_prompt.md
```diff
@@ -43,3 +43,29 @@ When the user asks about a general time period (e.g., "last week", "last 30 days
...
+**Rules for tool use:**
+1. Always call this tool before using any choice filter value to confirm valid IDs.
+2. Use the `search` parameter with the user's term to narrow results before selecting.
+3. Use option **IDs** (not labels) as filter values in `ColumnFilterData`.
+4. If `total > returned`, the list is truncated — refine your search to find the right option.
+5. If the tool returns an error, skip that filter and proceed with remaining filters.
+
+**Example:**
+- User says "filter by chatbot Alpha Bot" → call `get_filter_options(param="experiment", search="Alpha Bot")` → get `[{{"id": 42, "label": "Alpha Bot"}}]` → use value `[42]`
+- User says "filter by tag urgent" → call `get_filter_options(param="tags", search="urgent")` → get `[{{"id": "urgent", "label": "urgent"}}]` → use value `["urgent"]`
```
snopoke: update this example to use an emoji
```suggestion
- User says "error traces" → call `get_filter_options(params="status", search="error")` → get `[{{"id": "error", "label": "Error"}}]` → use value `["error"]`
- User says "filter by thumbs up tag" → call `get_filter_options(param="tags", search="👍")` → get `[{{"id": "👍", "label": "👍"}}]` → use value `["👍"]`
```

snopoke: Fixed in 4928718e4. Updated the example to use a thumbs-up emoji tag.

snopoke:
@coderabbitai full review

PR #2983, opened 2026-03-06
Remove orphan Chat records

SmittieC:
### Technical Description

I noticed that there are quite a bit of orphaned `Chat` records in the DB. Auto session deletions for eval sessions could be one source, but the earliest orphan chat is from 2023, so I'm not sure what may have caused them. This PR removes them and ensures that the eval session cleanup task takes the chat with it as well.

### Migrations

### Demo

N/A

### Docs and Changelog


SmittieC:
@coderabbitai full review

PR #3052, opened 2026-03-18
feat: capture print() output from Python Node for tracing

snopoke:
### Product Description
Python Node code in chatbot pipelines can now use `print()` to output debug/diagnostic information. The printed output is captured as `console_output` in the node's trace data, making it visible in Langfuse spans.

### Technical Description
- **`LogCapturePrintCollector`** — A RestrictedPython-compatible print collector that intercepts `print()` calls in user code and appends the output to a shared list. RestrictedPython transforms `print(x)` into `_print_(_print(x))`, so we provide a `_print_` factory that returns our collector.
- **`PipelineState.from_node_output`** — Added a generic `output_metadata` dict parameter. When provided, its contents are merged into the node's `outputs` entry (alongside `message` and `node_id`). This is reusable by any node type.
- **`CodeNode._process`** — Passes a `log_entries` list through to the sandbox globals. After execution, if any entries were captured, they're joined into a single string and passed as `output_metadata={"console_output": ...}`.
- The console output flows through to Langfuse automatically since the LangChain callbacks already serialize the full chain output dict.

### Demo
```python
def main(input, **kwargs):
    print("Processing input:", input)
    result = input.upper()
    print("Result:", result)
    return result
```
The `console_output` key will appear in the node's output in the trace span:
```json
{"message": "HI", "node_id": "...", "console_output": "Processing input: hi\nResult: HI\n"}
```

### Docs and Changelog

### DIFF
--- a/apps/help/utils.py
+++ b/apps/help/utils.py
@@ -84,7 +84,9 @@ def get_python_node_functions():
     node = CodeNode(name="test", node_id="123", django_node=None, code="")
     node._repo = InMemoryPipelineRepository()
     mock_state = PipelineState(outputs={}, experiment_session=None)
-    res = node._get_custom_functions(state=mock_state, context=NodeContext(mock_state), output_state=mock_state)
+    res = node._get_custom_functions(
+        state=mock_state, context=NodeContext(mock_state), output_state=mock_state, print_collectors=[]
+    )
     function_docs = filter(None, [extract_function_signature(name, obj) for name, obj in res.items()])
     return "\n".join(function_docs)
 


apps/pipelines/nodes/nodes.py
```diff
@@ -953,21 +955,37 @@ def _process(self, state: PipelineState, context: "NodeContext") -> PipelineStat
             message = get_code_error_message("<inline_code>", self.code)
             raise CodeNodeRunError(message) from exc
 
+        output_metadata = {}
+        console_output = "".join(collector() for collector in print_collectors)
+        if console_output:
+            output_metadata["console_output"] = console_output
```
snopoke:
```suggestion
        output_metadata = {
            "console_output": "".join(collector() for collector in print_collectors)
        }
```

apps/pipelines/tests/test_code_node.py
```diff
@@ -508,6 +508,48 @@ def main(input, **kwargs):
...
+    code = """
+def main(input, **kwargs):
+    return input
+"""
+    node = CodeNode(name="test", node_id="123", django_node=None, code=code)
+    node._repo = InMemoryPipelineRepository()
+    state = PipelineState(outputs={}, experiment_session=None, last_node_input="hi", node_inputs=["hi"])
+    node_output = node._process(state, NodeContext(state))
+    outputs = node_output.update["outputs"]["test"]  # ty: ignore[not-subscriptable]
+    assert "console_output" not in outputs
```
snopoke:
```suggestion
    assert outputs["console_output"] == ""
```

SmittieC (approved):
Cool!

PR #3068, opened 2026-03-20
Fix file content type detection using magic bytes

SmittieC:
### Product Description
Files with mismatched extensions (e.g. a JPEG image saved with a `.png` extension) were being served with incorrect `Content-Type` headers. This caused Twilio to fail when downloading media files for WhatsApp delivery.

### Technical Description
The root cause was that content type detection relied solely on the filename extension (`mimetypes.guess_type`). When a file's extension didn't match its actual content, the wrong MIME type was stored and served.

Changes:
- **`File.get_content_type()`**: Now reads the first 2048 bytes and uses `python-magic` (`magic.from_buffer`) as primary detection, falling back to filename-based guessing if byte inspection fails
- **`File.create()`**: Same strategy — tries `magic` on content bytes first, then falls back to `mimetypes.guess_type` from filename
- **`download_file` view** (`apps/experiments/views/experiment.py`): Passes stored `content_type` to `FileResponse` instead of letting Django guess from the filename
- **`FileView` view** (`apps/files/views.py`): Same fix for consistency

Note: `python-magic` was already a project dependency and used as a fallback in `File.create`. This change promotes it to the primary detection method.

### Demo
Verified that a JPEG file stored with a `.png` extension is now correctly identified as `image/jpeg`.

#### See it detected the right content type
image

### Docs and Changelog

### DIFF
--- a/apps/documents/tests/test_document_sources.py
+++ b/apps/documents/tests/test_document_sources.py
@@ -90,7 +90,7 @@ class TestDocumentSourceManager:
         assert files[0].status == FileStatus.PENDING
         file = files[0].file
         assert file.name == "test.md"
-        assert file.content_type == "text/markdown"
+        assert file.content_type == "text/plain"
         assert file.file.read() == b"# Test Document"
         assert "sha" in file.metadata
 


SmittieC:
> nice, how did you discover this was the issue?

Claude did really. I was debuggging a weird issue where certain files could not be downloaded by Twilio. There was a file with a png extention that looked 100% fine when compared to other files that could be downloaded by Twilio. The issue wasn't the file size either. Long story short, during debugging I asked claude to download this file, and it noticed that the actual content type of the file was different from what the response header said it would be. The acual content represented jpg or something.

PR #3084, opened 2026-03-24
Channels Refactor Phase 2 | Refactor web channel

SmittieC:
### Product Description
Fixes two broken tests caused by a recent refactor that moved `WebChannel` to a new module, and adds a regression test to guard against silent breakage of the `ChatMessage.rating()` unsaved-instance check.

### Technical Description
- Updated `@patch` decorators in `test_start_new_session` and `test_start_new_session_pipeline` to target `apps.channels.channels_v2.web_channel.WebChannel.new_user_message` — the previous path (`apps.chat.channels.WebChannel`) was stale after the refactor in commit `4bc89fdf6`.
- Added `TestChatMessage.test_rating_unsaved_instance_returns_none` to `apps/chat/tests/test_models.py` to cover the `if not self.pk: return None` guard in `ChatMessage.rating()` (models.py:303–304), ensuring it returns `None` and does not raise when called on an unsaved instance.

### Demo
N/A — test-only changes.

### Docs and Changelog


SmittieC:
@snopoke can I merge this?

apps/channels/channels_v2/web_channel.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def _get_sender(self) -> NoOpSender:
+        return NoOpSender()
+
+    def _get_callbacks(self) -> ChannelCallbacks:
+        return ChannelCallbacks()
+
+    def _get_capabilities(self) -> ChannelCapabilities:
+        return ChannelCapabilities(
+            supports_voice_replies=False,
```
snopoke: This seems to duplicate the `voice_replies_supported` class field. Do they have separate uses?

apps/channels/channels_v2/web_channel.py
```diff
@@ -0,0 +1,137 @@
...
+    by new_user_message() and picked up by periodic polling from the browser.
+    Session is always pre-set (created by start_new_session class method
+    before the pipeline runs).
+
+    start_new_session() and check_and_process_seed_message() are class
+    methods used outside the pipeline (from web views) and remain on
+    the channel class unchanged.
+    """
+
+    voice_replies_supported = False
```
snopoke: This is already the default value so can be left out

apps/chat/channels.py
```diff
@@ -323,7 +323,9 @@ def get_channel_class_for_platform(platform: ChannelPlatform | str) -> type[Chan
         if platform == "telegram":
             channel_cls = TelegramChannel
         elif platform == "web":
-            channel_cls = WebChannel
+            from apps.channels.channels_v2.web_channel import WebChannel as NewWebChannel  # noqa: PLC0415
```
snopoke, line 327: does this need to be inine?

pyproject.toml
```diff
@@ -88,7 +88,7 @@ dependencies = [
     "chardet",
     "dictdiffer>=0.9.0",
     "djlint>=1.36.4",
-    "transformers"
+"transformers"
```
snopoke: ?

snopoke (commented):
Few small comments but nothing blocking

snopoke (dismissed):

apps/channels/channels_v2/web_channel.py
SmittieC: Already addressed in commit 8079e269. `_get_capabilities()` no longer hardcodes `supports_voice_replies=False` — it now reads `self.voice_replies_supported` (the class field), so there's a single source of truth. The two names refer to different things: `voice_replies_supported` is the channel-class declaration, `supports_voice_replies` is the `ChannelCapabilities` dataclass field it feeds into.

apps/channels/channels_v2/web_channel.py
SmittieC: Fixed in 3c410d3b5. Dropped the redundant `voice_replies_supported = False` — `ChannelBase` already declares it as `False` by default.

apps/chat/channels.py
SmittieC: Yes — `apps/channels/channels_v2/web_channel.py` imports `MESSAGE_TYPES` and `_start_experiment_session` from `apps.chat.channels`, so a top-level import here would be circular. Added an explanatory comment in 3c410d3b5 matching the style of the neighboring `ApiChannel` import.

pyproject.toml
SmittieC: `transformers` is required for anthropic token counting via `llm.get_num_tokens` (added in 6b135c9d9). Fixed the broken indentation and missing trailing comma in 3c410d3b5.

snopoke (approved):

PR #3156, opened 2026-04-10
Pipeline as context to the chat widget

SmittieC:
### Product Description

Out of need, I got claude code to implement this:
We now give pipeline + event data as widget context when viewing a pipeline. The aim of this is to be able to ask the bot questions about the pipeline you see.

### Technical Description

We now give pipeline + event data as widget context. This is updated as the pipeline gets updated to stay current.

### Demo

image

### Docs and Changelog

Pipeline structure + event triggers are given as context to the chat widget in OCS when you're viewing a pipeline.

### DIFF
--- a/apps/pipelines/views.py
+++ b/apps/pipelines/views.py
@@ -90,12 +90,15 @@ class CreatePipeline(LoginAndTeamRequiredMixin, PermissionRequiredMixin, Templat
         return redirect(reverse("pipelines:edit", args=args, kwargs={**kwargs, "pk": pipeline.id}))
 
 
-def _get_pipeline_chat_widget_context(pipeline, experiment=None):
-    """Build chat widget context data for the pipeline view.
+def _serialize_event_action(action):
+    return {
+        "action_type": action.action_type,
+        "action_type_label": EventActionType(action.action_type).label,
+        "params": action.params,
+    }
 
-    Returns a dict with pipeline JSON and experiment events to be passed as
-    page context to the chat widget.
-    """
+
+def _get_pipeline_chat_widget_context(pipeline, experiment=None):
     context = {"pipeline_structure": pipeline.data}
 
     if experiment is None:
@@ -117,11 +120,7 @@ def _get_pipeline_chat_widget_context(pipeline, experiment=None):
                     "type": trigger.type,
                     "type_label": StaticTriggerType(trigger.type).label,
                     "is_active": trigger.is_active,
-                    "action": {
-                        "action_type": trigger.action.action_type,
-                        "action_type_label": EventActionType(trigger.action.action_type).label,
-                        "params": trigger.action.params,
-                    },
+                    "action": _serialize_event_action(trigger.action),
                 }
                 for trigger in static_triggers
             ],
@@ -131,11 +130,7 @@ def _get_pipeline_chat_widget_context(pipeline, experiment=None):
                     "total_num_triggers": trigger.total_num_triggers,
                     "is_active": trigger.is_active,
                     "trigger_from_first_message": trigger.trigger_from_first_message,
-                    "action": {
-                        "action_type": trigger.action.action_type,
-                        "action_type_label": EventActionType(trigger.action.action_type).label,
-                        "params": trigger.action.params,
-                    },
+                    "action": _serialize_event_action(trigger.action),
                 }
                 for trigger in timeout_triggers
             ],
@@ -349,7 +344,6 @@ def _pipeline_node_schemas():
 
 
 def _get_node_schema(node_class):
-
     schema = resolve_references(node_class.model_json_schema())
     schema.pop("$defs", None)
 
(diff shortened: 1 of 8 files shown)


SmittieC:
@coderabbitai full review

assets/javascript/apps/pipeline/App.tsx
```diff
@@ -4,6 +4,20 @@ import {apiClient} from "./api/api";
...
+function getWidget(): WidgetElement | null {
+  return document.querySelector("open-chat-studio-widget") as WidgetElement | null;
+}
+
+function syncPipelineToWidget(nodes: unknown[], edges: unknown[]) {
+  const widget = getWidget();
+  if (widget) {
+    const existing = widget.pageContext || {};
+    widget.pageContext = { ...existing, pipeline_structure: { nodes, edges } };
+  }
```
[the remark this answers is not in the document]
SmittieC: Fixed in eacbe7445. The `syncPipelineToWidget` function now spreads the existing `pipeline_structure` before overriding `nodes` and `edges`, preserving any other keys from the server-rendered pipeline data.

apps/pipelines/views.py
```diff
@@ -89,6 +90,55 @@ def get(self, request, *args, **kwargs):
...
+        static_triggers = StaticTrigger.objects.filter(experiment=experiment, is_archived=False).select_related(
+            "action"
+        )
+        timeout_triggers = TimeoutTrigger.objects.filter(experiment=experiment, is_archived=False).select_related(
+            "action"
+        )
+
+        context["experiment_events"] = {
+            "experiment_id": experiment.id,
+            "experiment_name": experiment.name,
```
snopoke: should we rename these to 'chatbot_id' and 'chatbot_name'?

apps/pipelines/views.py
```diff
@@ -89,6 +90,55 @@ def get(self, request, *args, **kwargs):
...
+            "static_triggers": [
+                {
+                    "type": trigger.type,
+                    "type_label": StaticTriggerType(trigger.type).label,
+                    "is_active": trigger.is_active,
```
snopoke: Maybe we should only include active triggers?

snopoke (commented):
This is great. Some thoughts:

I wonder if we could make it more generic e.g. just all the django template variable `widget_page_context` and have some JS in the global siteJS which looks for it and loads it into the widget if present. That would allow us to use the same pattern in other places.

I was also wondering about the structure and thinking of something like:

* 'chatbot'
  * 'name': X
  * 'description': X
  * 'version': Y
  * 'pipeline_structure':
    * 'nodes': X  (not sure if we need ALL the node data)
    * 'edges': Y
    * we can drop the other pipeline data since that's only used for display
  * 'event_triggers':
    * 'static_triggers': []
    * 'timeout_triggers': []
  * 'settings':
    * 'tracing_configured': yes / no
    * 'convert_speech_inputs_to_text': yes / no
    * 'convert_output_text_to_speech': yes / no
    * 'user_consent_required': yes / no
    * 'file_uploads_allowed': yes / no

We could have other structures for evals etc.

apps/pipelines/views.py
SmittieC: yes

SmittieC: Fixed in 3fba836a4. Renamed `experiment_id` and `experiment_name` to `chatbot_id` and `chatbot_name`.

apps/pipelines/views.py
SmittieC: Fixed in 3fba836a4. Now only active triggers are included in the serialized context. Also removed the `is_active` field from the serialized output since it's always true.

SmittieC:
> This is great. Some thoughts:
> 
> I wonder if we could make it more generic e.g. just all the django template variable `widget_page_context` and have some JS in the global siteJS which looks for it and loads it into the widget if present. That would allow us to use the same pattern in other places.
> 
> I was also wondering about the structure and thinking of something like:
> 
> * 'chatbot'
>   
>   * 'name': X
>   * 'description': X
>   * 'version': Y
>   * 'pipeline_structure':
>     
>     * 'nodes': X  (not sure if we need ALL the node data)
>     * 'edges': Y
>     * we can drop the other pipeline data since that's only used for display
>   * 'event_triggers':
>     
>     * 'static_triggers': []
>     * 'timeout_triggers': []
>   * 'settings':
>     
>     * 'tracing_configured': yes / no
>     * 'convert_speech_inputs_to_text': yes / no
>     * 'convert_output_text_to_speech': yes / no
>     * 'user_consent_required': yes / no
>     * 'file_uploads_allowed': yes / no
> 
> We could have other structures for evals etc.

See https://github.com/dimagi/open-chat-studio/pull/3156/commits/ed8a86962e671de9de8eef9ec0f503b1dc3f7a15

snopoke (approved):

PR #3179, opened 2026-04-17
feat: add cloudflare zero trust tunnel deployment option

ezkemboi:
### Product Description

Part of https://github.com/dimagi/open-chat-studio/issues/3183

No user-facing change. Adds optional Cloudflare Zero Trust Tunnel support for self-hosted deployments, enabling HTTPS public access without opening firewall ports or managing TLS certificates manually.

### Technical Description

Adds a `cloudflared` service to both `docker-compose.yml` (dev) and `docker-compose.prod.yml` (prod) that runs a Cloudflare Tunnel via a token.

**Key decisions:**

- **`CLOUDFLARE_TUNNEL_TOKEN`** added to `.env.example` — obtained from the Cloudflare dashboard under Networks > Tunnels.
- **Production (`docker-compose.prod.yml`):** `cloudflared` depends on `web` with `condition: service_healthy`. This required adding a healthcheck to the `web` service that polls `/status/` — ensuring the tunnel only starts routing traffic once gunicorn is fully ready.
- **Dev (`docker-compose.yml`):** uses `condition: service_started` (weaker) since the dev setup has no healthcheck and strict ordering is less critical.
- The `cloudflared` service is opt-in — it does nothing if `CLOUDFLARE_TUNNEL_TOKEN` is unset.

### Demo

**To enable:**
1. Create a tunnel in the Cloudflare dashboard (Networks > Tunnels > Create a tunnel).
2. Copy the tunnel token and set `CLOUDFLARE_TUNNEL_TOKEN=<token>` in `.env`.
3. Run `docker compose -f docker-compose.prod.yml up` — the `cloudflared` service will start automatically.

### Docs and Changelog


docker-compose.prod.yml
```diff
@@ -69,6 +69,12 @@ services:
     depends_on:
       migrate:
         condition: service_completed_successfully
+    healthcheck:
+      test: ["CMD", "python", "-c", "import urllib.request; urllib.request.urlopen('http://localhost:${PORT:-8000}/status/')"]
```
snopoke: You should be able to use curl for this

docker-compose.prod.yml
```diff
@@ -84,6 +90,21 @@ services:
       migrate:
         condition: service_completed_successfully
 
+  cloudflared:
```
snopoke: I think this should go in a separate compose file since not everyone will want to run this. Maybe `docker-compose.cloudflare.yml`?

docker-compose.yml
```diff
@@ -66,6 +66,23 @@ services:
       migrate:
         condition: service_completed_successfully
 
+  cloudflared:
```
snopoke: This isn't needed in this compose file as it is intended for local dev.

snopoke (commented):
Thanks for the PR

ezkemboi:
@snopoke, thanks for the review and feedback.

Updated as suggested:
- Moved Cloudflare tunnel to a separate `docker-compose.cloudflare.yml`
- Removed it from dev and production compose files
- Kept production clean and self-contained
- Updated healthcheck to use curl as suggested

Everything is now properly separated (dev, prod, optional tunnel). Ready for re-review 👍

docker-compose.cloudflare.yml
```diff
@@ -0,0 +1,12 @@
...
+    image: cloudflare/cloudflared:2025.4.0
+    restart: unless-stopped
+    command: tunnel --no-autoupdate run
+    environment:
+      - TUNNEL_TOKEN=${CLOUDFLARE_TUNNEL_TOKEN}
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "10m"
+        max-file: "3"
```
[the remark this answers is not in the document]
ezkemboi: Thanks for the review.

This is not a real issue in this setup.

The cloudflared service is an independent ingress/tunnel layer and is intentionally decoupled from the application stack. It does not require `depends_on` with `web` because:

- Docker Compose `depends_on` does not guarantee application readiness (only container start order in a single compose context)
- cloudflared is designed to handle backend unavailability and will retry connections automatically
- adding `depends_on` would incorrectly couple an optional networking layer to internal app lifecycle and reduce flexibility across dev/prod compositions

The web service healthcheck already ensures application readiness for internal dependencies, but cloudflared is external traffic routing and should not block or depend on app startup.

Therefore, the current design is intentional and safe, and no change is required.

docker-compose.cloudflare.yml
```diff
@@ -0,0 +1,12 @@
+services:
+  cloudflared:
+    image: cloudflare/cloudflared:2025.4.0
+    restart: unless-stopped
+    command: tunnel --no-autoupdate run
+    environment:
+      - TUNNEL_TOKEN=${CLOUDFLARE_TUNNEL_TOKEN}
+    logging:
```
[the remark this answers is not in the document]
ezkemboi: Well noted and fixed.

docker-compose.prod.yml
```diff
@@ -69,6 +69,12 @@ services:
     depends_on:
       migrate:
         condition: service_completed_successfully
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:${PORT:-8000}/status/"]
```
snopoke: This will fail if a health check token is configured in `settings.HEALTH_CHECK_TOKENS`.

ezkemboi: @snopoke, there are two possible approaches:

1. Update the Docker healthcheck to include HEALTH_CHECK_TOKENS, so it properly authenticates when calling /status/.
2. Remove the Docker healthcheck altogether, since it currently fails due to authentication and is not strictly required by any dependency (e.g., cloudflared does not rely on it).

ezkemboi: @snopoke thanks for the review.

I’ve updated the Docker healthcheck to include HEALTH_CHECK_TOKENS so it properly authenticates when calling /status, instead of failing when token protection is enabled.

I’ve also enforced HEALTH_CHECK_TOKENS as a required env var in prod to avoid silent misconfiguration.

This keeps the healthcheck aligned with the current secured /status design without removing or weakening it.

ezkemboi:
@snopoke thanks for the review.

I’ve removed the /status dependency from the production Docker healthchecks as part of simplifying the deployment model after decoupling `cloudflared` into an optional compose file.

Previously, /status was indirectly used to support the `cloudflared service dependency`, but since Zero Trust tunneling is now `optional` and moved to a separate compose setup, that coupling is no longer necessary.

The health checks now focus strictly on core infrastructure dependencies (DB/Redis), keeping the production Compose file simpler and more self-contained for self-hosted deployments.

snopoke (approved):

Issue #1615, opened 2025-05-21
Prevent users from sending too large message payloads


SmittieC:
### The objective
Prevent users from sending messages that will exceed the model's max token limit.

Note that we do have knowledge of a model's [max token limit](https://github.com/dimagi/open-chat-studio/blob/9fd6e85e77e0408c4d6453050aa01385cdd53305/apps/service_providers/models.py#L174-L178)

Since the prompt doesn't simply consist of the user's message - other components includes source material, participant data etc -, we need to first calculate the alotted token space within the prompt that the user's message are allowed to occupy. As a starting point, see the [summarization logic](https://github.com/dimagi/open-chat-studio/blob/29e0ad9a1e5831684583a190a7400038e1ff0110/apps/chat/conversation.py#L224-L260) where we do something similar.

snopoke:
Closing do to complexity that doesn't support the issue.

PR #3180, opened 2026-04-17
feat: validate message size against model token limit

ezkemboi:
### Product Description
Users who type a message that is too long for the configured LLM model now see a clear, actionable error instead of a silent failure or a generic "something went wrong" message. A live character counter in the chat input warns users as they approach and exceed the limit, and the send button is disabled when the limit is breached.

### Technical Description

**Backend**

- Added `MessageTooLargeError(PipelineNodeRunError)` in `apps/pipelines/exceptions.py` — a specific subclass that is safe to surface directly to users (unlike generic `PipelineNodeRunError`).
- `_validate_user_message_size()` in `apps/pipelines/nodes/llm_node.py` raises `MessageTooLargeError` when the incoming message exceeds `max_token_limit - system_prompt_tokens`. Called in `execute_sub_agent` before the agent is built, using the already-rendered system message to avoid double computation.
- `_check_post_compression_token_limit()` in `apps/pipelines/nodes/history_middleware.py` raises `MessageTooLargeError` after any compressor runs, covering the case where compressed history for non-summarisation modes (truncate, max-length) still overflows the limit.
- The Celery task (`get_response_for_webchat_task`) catches `MessageTooLargeError` separately and sets `user_facing_error=True` in the result dict. `get_message_task_response` in `task_utils.py` passes these errors through to the chat UI directly instead of replacing them with the generic message.
- The experiment view does a fast character-based pre-check (`len(message) > max_token_limit × 4`) before queuing the Celery task, returning a structured `400 {"error": "message_too_large", "detail": "…"}` immediately for obviously oversized messages.
- `apps/api/views/chat.py` (`chat_poll_task_response`): returns HTTP 400 for user-facing errors and HTTP 500 for server errors, so the widget can distinguish them.

**Frontend**

- HTMX error handler (`ajax_errors-js.html`) now parses JSON response bodies and shows `detail` or `error` field instead of raw HTML.
- `input_bar.html`: live character counter rendered when `max_char_limit` is in context. Input border turns red and send button is disabled above the limit; amber warning shown at 80%.
- `max_char_limit` (`= llm_provider_model.max_token_limit × 4`) is passed from `_experiment_session_message`, `chatbot_chat_session`, and `_chatbot_chat_ui`. The ×4 factor is a conservative heuristic (1 token ≈ 4 chars); the backend is the authoritative check.
- `web_chat.html`: passes `max-char-limit` to `<open-chat-studio-widget>` when the feature flag is active.
- Pipeline test panel (`TestMessageBox.tsx`): character counter and send-button guard added, reading the LLM node's `max_token_limit` from cached pipeline parameter values.
- Web widget (`ocs-chat.tsx`): new `maxCharLimit` prop, `messageTooLong` and `messageNearLimit` computed getters, character counter UI below the textarea, send button disabled when exceeded. CSS classes `message-textarea-error`, `char-counter-warning`, `char-counter-error` added to `ocs-chat.css`.
- `chat-session-service.ts` (`pollTaskOnce`): reads the JSON body from non-2xx responses so user-facing error messages are propagated to the widget UI instead of the generic browser `statusText`.

### Migrations

None.

### Demo

1. Open any chatbot session.
2. Paste a very long message (exceeding the model's token limit).
3. The character counter turns red and the send button is disabled — no request is made.
4. If a message somehow reaches the backend over the limit, the user sees "Your message is too large for this model…" inline instead of a generic error.

### Docs and Changelog


ezkemboi:
resolves #1615

ezkemboi:
Heads up for reviewers: this PR is larger than ideal and mainly serves as a reference for the full scope of changes. I’ll be breaking it down into smaller, focused PRs for easier review.

Happy to get any high-level feedback here in the meantime.

apps/experiments/views/utils.py
```diff
@@ -1,6 +1,33 @@
...
+
+    limits = list(
+        LlmProviderModel.objects.filter(id__in=model_ids)
+        .exclude(max_token_limit__isnull=True)
+        .exclude(max_token_limit=0)
+        .values_list("max_token_limit", flat=True)
+    )
+    if not limits:
+        return None
+    return min(limits) * 4
```
snopoke: why `*4?`

apps/experiments/views/utils.py
```diff
@@ -1,6 +1,33 @@
 from apps.channels.models import ChannelPlatform, ExperimentChannel
 
 
+def get_max_char_limit(experiment_version) -> int | None:
```
snopoke: we should be able to cache this based on the version and the pipeline modification date

components/chat_widget/src/components/ocs-chat/ocs-chat.css
snopoke, line 1: Changes to the chat widget should be in a separate PR since the widget has a different release cycle. I'll update the contributing docs to make this more clear.

templates/experiments/chat/input_bar.html
snopoke, line 1: This formatting is problematic as Django doesn't support multi-line tags / expressions.

Also, formatting should be handled by the pre-commit hooks: `prek install --install-hooks`

snopoke:
> Heads up for reviewers: this PR is larger than ideal and mainly serves as a reference for the full scope of changes. I’ll be breaking it down into smaller, focused PRs for easier review.
> 
> Happy to get any high-level feedback here in the meantime.

The size is OK though you should separate out changes to the chat widget in a separate PR since the widget has a different release cycle.

apps/experiments/views/utils.py
ezkemboi: I will take a look at this. This could have been heavy on the DB. Good catch.

apps/experiments/views/utils.py
ezkemboi: From an English perspective, I initially used a rough assumption of `~3–4 characters per token` as an approximation for `UI estimation`.

However, after reviewing this more carefully, I agree that this is only a heuristic and not reliable for enforcement.

It is better to rely on actual token counting for correctness and limits, rather than character-based estimation.

components/chat_widget/src/components/ocs-chat/ocs-chat.css
ezkemboi: This is well noted and will be separated.

templates/experiments/chat/input_bar.html
ezkemboi: I will handle this case.

ezkemboi:
@coderabbitai  can you do another review based on the changes made in progress?
And resolve any resolved issues

apps/experiments/views/utils.py
```diff
@@ -1,5 +1,54 @@
...
+        return cached
+
+    result = _compute_max_char_limit(pipeline)
+    if result is not None:
+        cache.set(cache_key, result)
+    return result
+
+
+def _compute_max_char_limit(pipeline) -> int | None:
+    from apps.pipelines.nodes.nodes import LLMResponseWithPrompt  # noqa: PLC0415
```
SmittieC: Does this have to be inline?

apps/pipelines/nodes/llm_node.py
```diff
@@ -92,12 +101,43 @@ def _process_agent_output(node: PipelineNode, session: ExperimentSession, messag
...
-    node: PipelineNode, context: NodeContext, session: ExperimentSession, tool_callbacks: ToolCallbacks
+    node: PipelineNode,
+    context: NodeContext,
+    session: ExperimentSession,
+    tool_callbacks: ToolCallbacks,
+    prompt_context: PromptTemplateContext | None = None,
+    system_message: BaseMessage | None = None,
 ):
-    prompt_context = _get_prompt_context(node, session, context)
+    if prompt_context is None:
```
SmittieC: Since `build_node_agent` is called only by `execute_sub_agent` which we know always passes `prompt_context` and `system_message`, we don't have to check if they exist in here again

apps/pipelines/nodes/llm_node.py
```diff
@@ -92,12 +101,43 @@ def _process_agent_output(node: PipelineNode, session: ExperimentSession, messag
     return ai_message, ai_message_metadata
 
 
+def _validate_user_message_size(user_input: str, node: PipelineNode, system_message: BaseMessage) -> None:
+    from apps.pipelines.repository import RepositoryLookupError  # noqa: PLC0415
```
SmittieC: Does this have to be an inline import?

apps/pipelines/nodes/llm_node.py
```diff
@@ -92,12 +101,43 @@ def _process_agent_output(node: PipelineNode, session: ExperimentSession, messag
...
+
+    try:
+        llm_provider_model = node.repo.get_llm_provider_model(node.llm_provider_model_id)
```
SmittieC: I think it's fine to assume that `llm_provider_model_id` will exist by now. If not and we return, it will just fail later on, so it might as well fail here if for some reason it is missing

apps/experiments/tasks.py
```diff
@@ -93,8 +94,12 @@ def get_response_for_webchat_task(
         response["response"] = chat_message.content
         response["message_id"] = chat_message.id
     except Exception as e:
-        logger.exception(e)
-        response["error"] = str(e)
+        if isinstance(e, MessageTooLargeError):
```
SmittieC: I suggest doing this instead:
```python
except MessageTooLargeError as e:
    response["error"] = str(e)
    response["user_facing_error"] = True
except Exception as e:
    logger.exception(e)
    response["error"] = str(e)
```

Additionally, I'm thinking, we might want to make `MessageTooLargeError` a subclass of `UserReportableError`, so we can do a more generic match on any error that we might want to report to the user:

```python
except UserReportableError as e:
    ...
```

It is also clearer when looking at the definition of `MessageTooLargeError` that it is meant to be seen by users.

apps/pipelines/tests/test_nodes.py
```diff
@@ -641,3 +641,59 @@ def test_recipient_list_undefined_error(self):
...
+            validate("one two three four five six seven eight nine ten", node, system_message)
+
+    def test_zero_token_limit_skips_validation(self):
+        node, validate = self._make_node(max_token_limit=0)
+        system_message = SystemMessage(content="")
+        # Should not raise even with a very long message
+        validate("word " * 10000, node, system_message)
+
+    def test_repo_error_skips_validation(self):
+        from apps.pipelines.nodes.llm_node import _validate_user_message_size  # noqa: PLC0415
```
SmittieC: All imports should be inline unless it causes circular imports

SmittieC (commented):
some initial comments

apps/pipelines/tests/test_nodes.py
snopoke: I think @SmittieC means all imports should be at the **module** unless there is a good reason. See https://github.com/dimagi/open-chat-studio/blob/main/AGENTS.md#dont

SmittieC: @snopoke thanks! :man_facepalming: That is what I meant to say

apps/experiments/tests/test_view_utils.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@pytest.mark.django_db()
+def test_returns_none_when_no_pipeline():
+    experiment = ExperimentFactory(pipeline=None)
+    assert get_max_char_limit(experiment) is None
+
+
+@pytest.mark.django_db()
+def test_returns_none_when_no_llm_nodes(db):
```
snopoke: you don't need the `db` fixture since you've decorated the test with `django_db`

apps/pipelines/nodes/history_middleware.py
```diff
@@ -46,9 +50,20 @@ def before_model(self, state, runtime):
         if result is not None:
             # A result means that a summary was created
             self.persist_summary(result["messages"])
+            self._check_post_compression_token_limit(result["messages"])
 
         return result
 
+    def _check_post_compression_token_limit(self, messages: list[BaseMessage]) -> None:
```
snopoke: I'm on the fence about this - at this point I think it might be better to let the LLM fail. The compression middleware is already taking the token limit into account.

apps/pipelines/nodes/llm_node.py
```diff
@@ -37,7 +39,14 @@ def execute_sub_agent(node: PipelineNode, context: NodeContext):
     user_input = context.input
     session = context.session
     tool_callbacks = ToolCallbacks()
-    agent = build_node_agent(node, context, session, tool_callbacks)
+
+    prompt_context = _get_prompt_context(node, session, context)
+    system_message = get_system_message(prompt_template=node.prompt, prompt_context=prompt_context)
+    _validate_user_message_size(user_input, node, system_message)
```
snopoke: Computing this here is a bit meaningless except for the first message since we're not accounting for the size of any previous messages or any uploaded files (see `format_multimodal_input`).

I'm wondering if it wouldn't be better to do it in a langgraph hook: https://docs.langchain.com/oss/python/langchain/middleware/custom

I'm also a bit concerned about the performance of this and wonder if just handling the LLM response error wouldn't be enough?

ezkemboi:
@snopoke and @SmittieC I will be working on this this week and all feedback. 
The feedback are appreciated.

templates/experiments/chat/input_bar.html
ezkemboi: @snopoke, with the latest push, formatting has been handled.

apps/experiments/views/utils.py
ezkemboi: @snopoke I have made use of langchain's defined max char count from token.

apps/experiments/views/utils.py
ezkemboi: @snopoke added cache for this

ezkemboi:
@coderabbitai please review this PR and resolve any conversation

apps/pipelines/nodes/llm_node.py
```diff
@@ -113,6 +126,25 @@ def build_node_agent(
...
+        max_token_limit = int(user_limit)
+    else:
+        llm_provider_model = node.repo.get_llm_provider_model(node.llm_provider_model_id)
+        max_token_limit = llm_provider_model.max_token_limit
+
+    if not max_token_limit:
+        return None
+
+    system_tokens = count_tokens_approximately([system_message])
+    response_reserve = max_token_limit // 8
```
ezkemboi: @snopoke and @SmittieC, regarding the response limit, I’m trying to approximate token usage so the request payload does not fully consume the model context window and leave no room for the model response. That’s why I reserved roughly 12.5% of the token limit for the response generation.

Is there an existing approximation utility or recommended approach in LangChain for handling this more accurately?

apps/pipelines/nodes/llm_node.py
```diff
@@ -113,6 +117,20 @@ def build_node_agent(
     )
 
 
+def _build_size_validation_middleware(node: PipelineNode, system_message) -> MessageSizeValidationMiddleware | None:
+    user_limit = getattr(node, "user_max_token_limit", None)
```
snopoke: I don't think we should use this limit. This is intended as a limit for compression shouldn't generate errors. I think the model `max_token_limit` is the correct value to use.

apps/pipelines/nodes/llm_node.py
snopoke: I don't know of any utility or recommended approach for this. I also don't think we necessarily need to account for response generation since that doesn't count towards the max token limit until the next request (at which point this validation will run again).

So I think we could drop the reserve althogether.

apps/pipelines/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    if not llm_nodes:
+        return None
+
+    limits = []
+    model_ids_needing_lookup = []
+
+    for node in llm_nodes:
+        user_limit = node.params.get("user_max_token_limit")
+        if user_limit is not None and int(user_limit) > 0:
+            limits.append(int(user_limit))
```
snopoke: rather just use the llm model max

apps/pipelines/utils.py
```diff
@@ -0,0 +1,63 @@
...
+            .exclude(max_token_limit=0)
+            .values_list("max_token_limit", flat=True)
+        )
+        limits.extend(model_limits)
+
+    if not limits:
+        return None
+
+    token_limit = min(limits)
+    response_reserve = token_limit // 8
```
snopoke: also drop the reserve from here

templates/experiments/chat/input_bar.html
```diff
@@ -83,8 +82,20 @@
...
+        <input x-model="message" name="message" type="text" placeholder="Type your message..." aria-label="Message" autocomplete="off"
+               :maxlength="maxCharLimit || false"
+               :class="messageTooLong ? 'input input-error w-full' : 'input input-primary w-full'">
+        {% if max_char_limit %}
+        <div class="flex justify-end text-xs pr-1">
+          <span :class="messageTooLong ? 'text-error font-semibold' : (messageNearLimit ? 'text-warning' : 'text-base-content/50')">
+            <span x-text="message.length"></span> / {{ max_char_limit }}
+          </span>
+        </div>
+        {% endif %}
```
snopoke: This is misleading since it is only comparing against the global limit and not taking the previous messages / system prompt into account. Since that would be tough to do I'd remove this extra display element

apps/pipelines/nodes/llm_node.py
ezkemboi: Removed reserve

apps/pipelines/nodes/history_middleware.py
```diff
@@ -128,3 +131,26 @@ def _build_new_messages(self, summary: str) -> list:
...
+
+    def before_model(self, state, runtime):
+        if not self._token_limit:
+            return None
+        token_count = count_tokens_approximately(state["messages"])
+        if token_count > self._token_limit:
+            raise MessageTooLargeError(
+                f"Your message is too large for this model. "
+                f"It uses approximately {token_count} tokens, but only {self._token_limit} tokens are available "
+                f"after accounting for the system prompt."
```
snopoke:
```suggestion
                f"It uses approximately {token_count} tokens, but only {self._token_limit} tokens are available."
```

templates/experiments/chat/input_bar.html
```diff
@@ -143,13 +147,20 @@ <h4 class="font-bold">Uploaded Files</h4>
...
       file_search_files: [],
       ocs_attachments: [],
       total_upload_size_bytes: 0,
       disableCodeInterpreterUpload: false,
+      get messageTooLong() {
+        return maxCharLimit !== null && this.message.length > maxCharLimit;
+      },
+      get messageNearLimit() {
+        return maxCharLimit !== null && this.message.length > maxCharLimit * 0.8;
+      },
```
snopoke: nit: unused code
```suggestion
```

snopoke (approved):

PR #3219, opened 2026-04-23
Evaluators can tag sessions / messages

### LINKED ISSUE #3193: Allow evals to tag sessions / messages as part of the eval process
* Allow user to specify tag name
* Allow user to determine when the tag is applied
  * when choice eval field matches value
  * when numeric eval field is in range
* Ensure that tags application is idempotent

SmittieC:
### Product Description
Chatbot builders can now configure evaluators to automatically tag sessions or messages based on evaluation results. When creating or editing an LLM evaluator, a new **Tag Rules** section lets you attach tags conditionally:

- **Choice fields** — apply a tag when the evaluator output equals a chosen value (e.g. `sentiment == "negative"`).
- **Numeric fields** (int/float) — apply a tag when the output falls inside a `min..max` range, or equals a specific value.
- **String fields** — apply a tag when the output equals a specific value.

Tag rules run automatically during every non-preview evaluation run. Matching rules tag the target chat message (MESSAGE mode) or the session's chat (SESSION mode), and each application is recorded in an audit table visible on the run results page under the **Applied Tags** column.

Resolves #3193.

### Technical Description
**New models** (`apps/evaluations/models.py`, migration `0014`):
- `EvaluatorTagRule` — a rule bound to an evaluator + tag, with a `field_name`, `condition_type` (`equals`/`range`), and a JSON `condition_value`. Validated against the evaluator's `output_schema` on `clean()`.
- `AppliedTag` — audit row written each time a rule matches; cascades from both rule and evaluator deletion.
- New `EVALUATIONS` category on `TagCategories`; tag-rule tags are created in this category and are _not_ system tags (so users can manage them).

**Rule engine** (`apps/evaluations/tagging.py`, `apps/evaluations/rule_validation.py`):
- `rule_validation.py` holds the pure, DB-free validators + `ConditionType` enum. Kept separate so `models.py` can import it at module level without circular imports.
- `tagging.py` orchestrates DB writes: it evaluates rules against an `EvaluationResult.output`, bulk-creates `CustomTaggedItem` rows on the target, and bulk-creates `AppliedTag` audit rows. Caller (the evaluation task) runs it inside `transaction.atomic()` so tag + audit writes roll back together on failure.

**Task integration** (`apps/evaluations/tasks.py`):
- `evaluate_single_message_task` now calls `_maybe_apply_tag_rules` after each `EvaluationResult` is created.
- Skips tagging on PREVIEW runs and on error outputs.
- `Evaluator.tag_rules` is prefetched with the tag relation to keep the query count flat.

**UI** (`apps/evaluations/forms.py`, `templates/evaluations/evaluator_form.html`, `templates/evaluations/partials/tag_rule_row.html`):
- New `EvaluatorTagRuleFormSet` inline formset on the evaluator create/edit page. Tag name is a plain text field that gets-or-creates a Tag on save.
- Condition-type options are filtered by the selected field's type (choice/string → `equals` only; int/float → `equals` or `range`).
- Value input shape switches on condition type: dropdown for choice fields, single input for `equals` on numeric/string, min+max inputs for `range`.
- Trash icon removes rows; empty rows are ignored on save.
- Section is hidden for non-LLM evaluator types.

**Schema-drift guard**:
- `EvaluatorForm.clean()` refuses to save a new `output_schema` when an existing tag rule would be incompatible with it (field removed, or type changed in a way that invalidates the condition). The blocking error surfaces on the form's non-field errors.

**Run results**:
- The run-results "Tags" column was renamed to **Applied Tags** and now shows the audit rows created by tag rules, distinct from ad-hoc chat tags.

**Admin**:
- `EvaluatorTagRule` and `AppliedTag` are registered in Django admin for debugging.

**Backends** (`apps/teams/backends.py`): `EvaluatorTagRule` and `AppliedTag` added to the team-scoped permission set.

### Migrations
- `apps/annotations/migrations/0010_add_evaluations_tag_category.py` — enum value add, backwards compatible.
- `apps/evaluations/migrations/0014_add_tag_rules_and_applied_tags.py` — two new tables, no changes to existing tables. Backwards compatible.

### Docs and Changelog

Evaluator tag rules: configure LLM evaluators to automatically tag sessions or messages when the evaluator output matches a condition (equals/range). Applied tags are recorded in an audit log and shown on the run results page.


apps/evaluations/forms.py
```diff
@@ -303,8 +307,223 @@ def clean(self):
...
+        tag, _ = Tag.objects.get_or_create(
+            team=self.team,
+            name=tag_name,
+            is_system_tag=False,
+            category=TagCategories.EVALUATIONS,
+        )
+        cleaned["tag"] = tag
+        self.instance.tag = tag
+        self.instance.condition_value = cleaned["condition_value"]
+        return cleaned
```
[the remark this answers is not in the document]
SmittieC: Fixed in a09c90658. Moved `Tag.objects.get_or_create` out of `clean()` — `EvaluatorTagRuleForm.clean()` now only stashes the resolved tag name, and `EvaluatorTagRuleForm.save()` does the `get_or_create`. The formset-level `save()` already runs under the view’s `transaction.atomic()`, so tag creation now lives inside the transactional save path and cannot leak orphan tags on a failed submit.

apps/evaluations/models.py
```diff
@@ -445,3 +459,62 @@ class EvaluationRunAggregate(BaseModel):
...
+
+        if self.evaluator_id and self.team_id and self.evaluator.team_id != self.team_id:
+            raise ValidationError({"team": "Rule team must match evaluator team."})
+
+        if self.tag_id:
+            validate_tag_compatibility(self.tag, self.evaluator)
+
+        if self.evaluator_id:
+            output_schema = self.evaluator.params.get("output_schema", {}) or {}
+            field_def = validate_field_in_schema(self.field_name, output_schema)
```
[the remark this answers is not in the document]
SmittieC: Fixed in a09c90658. Added `self.evaluator_id` to the guard on the `validate_tag_compatibility` call so the clean() path now returns a clean ValidationError instead of raising `RelatedObjectDoesNotExist` when tag is set but evaluator is not yet attached.

apps/evaluations/tagging.py
```diff
@@ -0,0 +1,254 @@
...
+    if not isinstance(field_definition, _NUMERIC_FIELD_TYPES):
+        raise ValidationError({"condition_type": "'range' condition is only valid for numeric (int/float) fields."})
+    py_type = field_definition.python_type
+    try:
+        lo = py_type(condition_value["min"])
+        hi = py_type(condition_value["max"])
+    except (TypeError, ValueError) as err:
+        raise ValidationError({"condition_value": f"min/max must be coercible to {py_type.__name__}."}) from err
+    if lo > hi:
+        raise ValidationError({"condition_value": "'min' must be <= 'max'."})
```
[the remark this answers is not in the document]
SmittieC: Fixed in a09c90658. Added a `_coerce_numeric` helper that both `_validate_equals_condition` and `_validate_range_condition` now delegate to. For int-typed fields it rejects non-integral floats (1.9) and non-integer strings ("5.5") outright, while still accepting integral forms ("5", 5.0). The coerced native value is written back into `condition_value` so `matches()` has a type-accurate payload at runtime.

apps/evaluations/tagging.py
```diff
@@ -0,0 +1,254 @@
...
+
+    content_type = ContentType.objects.get_for_model(type(target))
+
+    if tags_to_remove:
+        CustomTaggedItem.objects.filter(
+            content_type=content_type,
+            object_id=target.pk,
+            tag_id__in=tags_to_remove,
+            tag__category=TagCategories.EVALUATIONS,
+        ).delete()
```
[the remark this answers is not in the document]
SmittieC: Fixed in a09c90658. Before deleting, `apply_rules_to_result` now consults the `AppliedTag` audit for any still-existing application of each candidate tag from a rule on a *different* evaluator whose evaluation result targets the same chat (MESSAGE mode) or session (SESSION mode). Any tag "owned" elsewhere is subtracted from `tags_to_remove` before the delete, so intentional cross-evaluator tag sharing no longer causes clobber.

apps/evaluations/forms.py
```diff
@@ -303,8 +307,229 @@ def clean(self):
...
+        tag_name = (cleaned.get("tag_name") or "").strip()
+        if not tag_name:
+            self.add_error("tag_name", "Tag name is required.")
+        if not cleaned.get("field_name"):
+            self.add_error("field_name", "Field is required.")
+
+        condition_type = cleaned.get("condition_type")
+        field_schema = self.output_schema.get(cleaned.get("field_name") or "", {}) or {}
+        field_type = field_schema.get("type")
+        if condition_type == ConditionType.EQUALS:
```
SmittieC, line 423: Might be a good idea to move the coercion logic into the `ConditionType` class. So we do
```python
value = cleaned.get("condition_value_single")
cleaned["condition_value"] = ConditionType.coerce_value(value)
```

apps/evaluations/forms.py
```diff
@@ -303,8 +307,229 @@ def clean(self):
...
+            else:
+                try:
+                    cleaned["condition_value"] = {"min": float(lo), "max": float(hi)}
+                except (TypeError, ValueError):
+                    self.add_error("condition_value_min", "Min/max must be numeric.")
+
+        if self.errors:
+            return cleaned
+
+        from apps.evaluations.tagging import (  # noqa: PLC0415
```
SmittieC: Move all imports to the module level

apps/evaluations/forms.py
```diff
@@ -303,8 +307,229 @@ def clean(self):
...
+            validate_condition,
+            validate_field_in_schema,
+        )
+
+        try:
+            field_def = validate_field_in_schema(cleaned["field_name"], self.output_schema)
```
SmittieC, line 456: Let's split out the ton of things this method does into smaller methods

apps/evaluations/forms.py
```diff
@@ -303,8 +307,229 @@ def clean(self):
...
+            return cleaned
+
+        # Defer Tag.objects.get_or_create to save() so a failure elsewhere in the formset
+        # cannot leave behind an orphan Tag row that was never attached to a saved rule.
+        self._resolved_tag_name = tag_name
+        self.instance.condition_value = cleaned["condition_value"]
+        return cleaned
+
+    def save(self, commit=True):
+        tag_name = getattr(self, "_resolved_tag_name", None)
```
SmittieC: Why use `getattr` instead of `self.getattr`?

apps/evaluations/models.py
```diff
@@ -445,3 +458,62 @@ class EvaluationRunAggregate(BaseModel):
...
+    field_name = models.CharField(max_length=255)
+    condition_type = models.CharField(max_length=20, choices=ConditionType.choices)
+    condition_value = models.JSONField(default=dict)
+
+    def __str__(self):
+        return f"{self.evaluator.name}: {self.field_name} {self.condition_type} {self.condition_value}"
+
+    def clean(self):
+        super().clean()
+        from apps.evaluations.tagging import (  # noqa: PLC0415 - circular: tagging imports models
```
SmittieC: let's create a service for this so we don't have to deal with circular imports

apps/evaluations/tagging.py
```diff
@@ -0,0 +1,298 @@
...
+    hi = _coerce_numeric(condition_value["max"], py_type)
+    if lo > hi:
+        raise ValidationError({"condition_value": "'min' must be <= 'max'."})
+    condition_value["min"] = lo
+    condition_value["max"] = hi
+
+
+def validate_tag_compatibility(tag: Tag, evaluator: Evaluator) -> None:
+    """Ensure the tag is compatible with this evaluator (category, team)."""
+    if tag.team_id != evaluator.team_id:
```
SmittieC: when will this ever happen?

apps/evaluations/tagging.py
```diff
@@ -0,0 +1,298 @@
...
+        raise ValidationError({"tag": "Tag must belong to the same team as the evaluator."})
+    if tag.category != TagCategories.EVALUATIONS:
+        raise ValidationError({"tag": f"Tag must be in the '{TagCategories.EVALUATIONS}' category."})
+
+
+def matches(condition_type: str, condition_value: dict, field_value: Any) -> bool:
+    """Return True if field_value satisfies the condition. Raises on unknown type."""
+    if condition_type == ConditionType.EQUALS:
```
SmittieC: I think this method should also move to `ConditionType`

apps/evaluations/tagging.py
```diff
@@ -0,0 +1,298 @@
...
+    """Return the subset of rules that match the given evaluator result output.
+
+    Rules whose field_name is absent or whose condition doesn't apply cleanly
+    are skipped with a logged warning (defensive against schema drift).
+    """
+    matched: list[EvaluatorTagRule] = []
+    result_dict = (result_output or {}).get("result") or {}
+
+    for rule in rules:
+        if rule.field_name not in result_dict:
```
SmittieC, line 48: skip with logging

apps/evaluations/tagging.py
```diff
@@ -0,0 +1,298 @@
...
+        return
+
+    matched_rules = evaluate_rules(rules, evaluation_result.output or {})
+    tags_to_apply = {rule.tag_id: rule.tag for rule in matched_rules}
+    rule_tag_ids = {rule.tag_id for rule in rules}
+    tags_to_remove = rule_tag_ids - set(tags_to_apply.keys())
+
+    content_type = ContentType.objects.get_for_model(type(target))
+
+    if tags_to_remove:
```
SmittieC: we should not handle tag removal in this PR. Remove this logic

apps/evaluations/views/evaluator_views.py
```diff
@@ -44,8 +47,64 @@ def get_queryset(self):
...
+        return EvaluatorTagRuleFormSet(
+            data=data,
+            instance=instance,
+            team=self.request.team,
+            output_schema=(instance.params or {}).get("output_schema", {}) if instance else {},
+        )
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if "tag_rule_formset" not in context:
```
SmittieC, line 65: Why check if this is the only mixin adding this?

apps/evaluations/views/evaluator_views.py
```diff
@@ -44,8 +47,64 @@ def get_queryset(self):
...
+            context["tag_rule_formset"] = self._build_tag_rule_formset(context.get("object"))
+        return context
+
+    def post(self, request, *args, **kwargs):
+        self.object = getattr(self, "object", None) or self.get_object_or_none()
```
SmittieC: Will just a call to `self.get_object_or_none()` not make `getattr(self, "object", None)` redundant?

apps/evaluations/views/evaluator_views.py
```diff
@@ -44,8 +47,64 @@ def get_queryset(self):
...
+        return {
+            "evaluator_schemas": _evaluator_schemas(),
+            "parameter_values": _evaluator_parameter_values(self.request.team, llm_providers, llm_provider_models),
+            "default_values": _evaluator_default_values(llm_providers, llm_provider_models),
+        }
+
+
 @waf_allow(WafRule.SizeRestrictions_BODY)
-class CreateEvaluator(LoginAndTeamRequiredMixin, PermissionRequiredMixin, CreateView):
+class CreateEvaluator(EvaluatorFormsetMixin, LoginAndTeamRequiredMixin, PermissionRequiredMixin, CreateView):
```
SmittieC: Should `LoginAndTeamRequiredMixin` not be first?

apps/evaluations/forms.py
SmittieC: Fixed in 3b982d2004ddc3f694557d31287101e0d8a1eba8. Added `ConditionType.coerce_value(raw, field_type)` as a staticmethod in models.py; the form now calls it via `ConditionType.coerce_value(value, field_type)` instead of inlining the int/float try blocks.

apps/evaluations/forms.py
SmittieC: Fixed in 3b982d2004ddc3f694557d31287101e0d8a1eba8. Lifted `validate_condition` / `validate_field_in_schema` to module-level imports at the top of forms.py, plus `construct_instance` and removed the other `# noqa: PLC0415` local-imports in this file. Moving the validators into `apps/evaluations/rule_validation.py` (see the circular-import thread) broke the cycle that required the local imports.

apps/evaluations/forms.py
SmittieC: Fixed in 3b982d2004ddc3f694557d31287101e0d8a1eba8. Split `EvaluatorTagRuleForm.clean()` into three helpers: `_clean_required_row_fields` (tag/field presence), `_clean_condition_value` (builds `condition_value` per condition type), and `_validate_against_schema` (schema-level validation).

apps/evaluations/forms.py
SmittieC: Fixed in 3b982d2004ddc3f694557d31287101e0d8a1eba8. Now initialises `self._resolved_tag_name = None` in `__init__` and uses the direct attribute access `self._resolved_tag_name` in `save()` — no more `getattr` sentinel.

apps/evaluations/models.py
SmittieC: Fixed in 3b982d2004ddc3f694557d31287101e0d8a1eba8. Extracted the pure validators (`parse_field_definition`, `validate_field_in_schema`, `validate_condition`, `validate_tag_compatibility`, and the private `_coerce_*`/`_validate_*` helpers) into a new `apps/evaluations/rule_validation.py` module. It uses "equals"/"range" string literals internally so it doesn't need to import `ConditionType` from models.py, which breaks the cycle. `models.py` now imports the validators at module level, and `tagging.py` re-exports them for existing callers/tests.

apps/evaluations/tagging.py
SmittieC: The UI/form path can't reach it — `EvaluatorTagRuleForm.save()` always does `Tag.objects.get_or_create(team=self.team, …)` where `self.team == evaluator.team`. The defensive check fires on admin / shell / ORM paths where a rule can be constructed with a tag from a different team (our `test_cross_team_tag_rejected` covers exactly that route). I'd rather keep the check than ship a silent data-integrity hole, so I've left it in place — but updated the docstring to call out why it's there. Happy to drop it if you feel strongly it's dead weight.

apps/evaluations/tagging.py
SmittieC: Fixed in 3b982d2004ddc3f694557d31287101e0d8a1eba8. Added a `matches(condition_value, field_value)` instance method on `ConditionType`; `tagging.matches()` is now a thin wrapper that does `ConditionType(condition_type).matches(...)` (kept as a module-level function so existing callers/tests don't have to change).

apps/evaluations/tagging.py
SmittieC: The current implementation already skips with a `logger.warning` — both the missing-field branch and the `ValueError`/`TypeError` branch from `matches()` continue the loop after logging. Let me know if you meant something more specific (e.g. demoting to `info`/`debug`, or widening the caught exception set) and I'll adjust.

apps/evaluations/tagging.py
SmittieC: Fixed in 3b982d2004ddc3f694557d31287101e0d8a1eba8. Dropped the `tags_to_remove` computation, the `other_applications` owner-check, and the `CustomTaggedItem.objects.filter(...).delete()` call from `apply_rules_to_result`. Also removed the two tests that covered removal semantics (`test_removal_semantics_between_runs` and `test_removal_only_touches_evaluations_category`) so the suite stays honest about what the orchestrator now does.

apps/evaluations/views/evaluator_views.py
SmittieC: The `"tag_rule_formset" not in context` check isn't guarding against other mixins — it's guarding against our own `post()` error path, which calls `self.render_to_response(self.get_context_data(form=form, tag_rule_formset=formset))`. The kwargs land in context via `super().get_context_data(**kwargs)`, and without the check we'd replace the formset that already has validation errors with a fresh unbound one. Happy to switch it to `context.setdefault(...)` if you prefer, but leaving the current form for now (no code change needed).

apps/evaluations/views/evaluator_views.py
SmittieC: Fixed in 2cc5037c2e7add4f48e571447eb937ad0b54bf95. You're right — `self.object` isn't set on the instance before this line in either Create or Edit (our `post()` overrides the framework one), so the `getattr` fallback was always `None`. Simplified to `self.object = self.get_object_or_none()`.

apps/evaluations/views/evaluator_views.py
SmittieC: Fixed in 2cc5037c2e7add4f48e571447eb937ad0b54bf95. Reordered to `LoginAndTeamRequiredMixin, PermissionRequiredMixin, EvaluatorFormsetMixin, CreateView` (same for `EditEvaluator`), matching the convention used across `apps/experiments/views/` and the rest of the codebase.

apps/evaluations/tagging.py
```diff
@@ -11,22 +11,17 @@
...
-from apps.evaluations.field_definitions import (
-    ChoiceFieldDefinition,
-    FieldDefinition,
-    FloatFieldDefinition,
-    IntFieldDefinition,
-    StringFieldDefinition,
-)
+from apps.annotations.models import CustomTaggedItem
 from apps.evaluations.models import ConditionType
+from apps.evaluations.rule_validation import (  # noqa: F401 - re-exported for callers/tests
```
SmittieC: > re-exported for callers/tests

No, don't do this. Update the tests and callers rather

SmittieC: Fixed in ab019f597. Dropped the re-export block from `apps/evaluations/tagging.py` and updated `test_tagging_logic` to import `validate_condition` / `validate_field_in_schema` directly from `apps.evaluations.rule_validation`. No other callers were using the re-exports.

apps/evaluations/rule_validation.py
```diff
@@ -0,0 +1,156 @@
...
+) -> None:
+    """Raise ValidationError when the condition is malformed for the given field.
+
+    Mutates condition_value in place to coerce values to the field's native python_type
+    so stored rules match the type of the evaluator output and `==` comparisons behave.
+    """
+    if not isinstance(condition_value, dict):
+        raise ValidationError({"condition_value": "Condition value must be a JSON object."})
+
+    if condition_type == _EQUALS:
```
SmittieC: Ugly. We need to use the enum value here

SmittieC:
@coderabbitai full review

apps/evaluations/tests/test_tagging_logic.py
```diff
@@ -0,0 +1,218 @@
...
+        self.pk = pk
+        self.field_name = field_name
+        self.condition_type = condition_type
+        self.condition_value = condition_value
+
+
+# ---- matches ---------------------------------------------------------------
+
+
+class TestMatches:
```
snopoke, line 31: nit: most of these could be parameterized into a single test

apps/evaluations/tests/test_tagging_logic.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def test_unknown_condition_raises(self):
+        with pytest.raises(ValueError, match="Unknown condition type"):
+            matches("contains", {"value": "x"}, "x")
+
+
+# ---- validate_condition ----------------------------------------------------
+
+
+class TestValidateCondition:
```
snopoke, line 57: could also be parameterized

apps/evaluations/admin.py
```diff
@@ -59,3 +61,24 @@ class EvaluationResultAdmin(ReadonlyAdminMixin, admin.ModelAdmin):
 class EvaluationMessageAdmin(ReadonlyAdminMixin, admin.ModelAdmin):
     list_display = ("id", "input", "output", "context")
     search_fields = ("id",)
+
+
+class EvaluatorTagRuleInline(admin.TabularInline):
```
snopoke:
```suggestion
class EvaluatorTagRuleInline(ReadonlyAdminMixin, admin.TabularInline):
```

apps/evaluations/admin.py
```diff
@@ -59,3 +61,24 @@ class EvaluationResultAdmin(ReadonlyAdminMixin, admin.ModelAdmin):
...
+    model = EvaluatorTagRule
+    extra = 0
+    fields = ("tag", "field_name", "condition_type", "condition_value")
+    autocomplete_fields = ("tag",)
+
+
+@admin.register(EvaluatorTagRule)
+class EvaluatorTagRuleAdmin(admin.ModelAdmin):
```
snopoke:
```suggestion
class EvaluatorTagRuleAdmin(ReadonlyAdminMixin, admin.ModelAdmin):
```

apps/evaluations/forms.py
```diff
@@ -303,8 +309,225 @@ def clean(self):
...
+    condition_value_max = forms.CharField(required=False, label="Max")
+
+    class Meta:
+        model = EvaluatorTagRule
+        fields = ("field_name", "condition_type")
+        widgets = {
+            "field_name": forms.TextInput(attrs={"class": "input input-bordered w-full input-sm"}),
+            "condition_type": forms.Select(
+                attrs={"class": "select select-bordered w-full select-sm"},
+                choices=ConditionType.choices,
```
snopoke: nit: CSS in python. Can this be pushed to the template?

apps/evaluations/tagging.py
```diff
@@ -0,0 +1,127 @@
...
+    evaluation_result: EvaluationResult,
+    evaluator: Evaluator,
+    evaluation_message: EvaluationMessage,
+) -> None:
+    """Apply this evaluator's tag rules to the target and record audit rows.
+
+    Caller is responsible for running this inside a transaction.atomic() block along
+    with the EvaluationResult.create() it corresponds to.
+    """
+    rules = list(evaluator.tag_rules.all())
```
snopoke: this could be cached to avoid running the query for every row in the result

apps/evaluations/forms.py
```diff
@@ -303,8 +309,225 @@ def clean(self):
...
+        output_schema = (params or {}).get("output_schema") or {}
+        for rule in self.instance.tag_rules.all():
+            try:
+                field_def = validate_field_in_schema(rule.field_name, output_schema)
+                validate_condition(rule.condition_type, rule.condition_value, field_def)
+            except DjangoValidationError as err:
+                raise forms.ValidationError(
+                    f"Tag rule on field '{rule.field_name}' is incompatible with the updated "
+                    f"output schema: {err.messages[0] if err.messages else err}"
+                ) from err
```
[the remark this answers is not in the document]
snopoke: Fixed in f9f9a94e. `_validate_existing_tag_rules_against_schema` now excludes rules the formset is deleting in the same submit (via `pending_deleted_rule_pks`, populated by the view from POST data) and collects all incompatible-rule errors instead of raising on the first.

*— Claude Code*

templates/evaluations/partials/tag_rule_row.html
```diff
@@ -0,0 +1,104 @@
...
+      {{ form.condition_type.errors }}
+    </div>
+    <div class="flex items-end justify-end">
+      <button type="button"
+              @click="deleted = true"
+              class="btn btn-outline btn-xs btn-error"
+              aria-label='{% translate "Remove tag rule" %}'>
+        <i class="fa fa-trash"></i>
+      </button>
+    </div>
```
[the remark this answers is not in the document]
snopoke: Fixed in f9f9a94e. Each label now uses `for="{{ form.<field>.id_for_label }}"` and the matching input/select sets that `id`. Applied to all six fields (Tag name, Field, Condition, Value, Min, Max).

*— Claude Code*

apps/evaluations/admin.py
snopoke: Fixed in f9f9a94e.

*— Claude Code*

apps/evaluations/admin.py
snopoke: Fixed in f9f9a94e.

*— Claude Code*

apps/evaluations/rule_validation.py
```diff
@@ -0,0 +1,182 @@
...
+        if self == ConditionType.EQUALS:
+            return field_value == condition_value.get("value")
+        if self == ConditionType.RANGE:
+            try:
+                numeric = float(field_value)
+            except (TypeError, ValueError):
+                return False
+            return float(condition_value["min"]) <= numeric <= float(condition_value["max"])
+        raise ValueError(f"Unknown condition type: {self}")
+
```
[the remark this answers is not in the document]
snopoke: Fixed in f9f9a94e. `ConditionType.RANGE.matches()` now uses `.get("min")` / `.get("max")`; missing or non-numeric bounds raise `TypeError`/`ValueError` from `float()` and are caught by the existing handler, returning `False` instead of `KeyError`.

*— Claude Code*

apps/evaluations/tagging.py
```diff
@@ -0,0 +1,127 @@
...
+                rule.field_name,
+            )
+            continue
+        field_value = result_dict[rule.field_name]
+        try:
+            if matches(rule.condition_type, rule.condition_value or {}, field_value):
+                matched.append(rule)
+        except (ValueError, TypeError) as exc:
+            logger.warning("Skipping tag rule %s due to evaluation error: %s", rule.pk, exc)
+    return matched
```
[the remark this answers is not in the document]
snopoke: Fixed in f9f9a94e. Added `KeyError` to the except tuple in `evaluate_rules` for defense in depth (also addressed at the source by switching `RANGE.matches()` to `.get()`).

*— Claude Code*

apps/evaluations/views/evaluator_views.py
```diff
@@ -44,8 +47,64 @@ def get_queryset(self):
...
+            if formset.is_valid():
+                with transaction.atomic():
+                    evaluator.save()
+                    formset.instance = evaluator
+                    formset.save()
+                return redirect(self.get_success_url())
+        else:
+            formset.is_valid()  # surface errors in the template
+
+        return self.render_to_response(self.get_context_data(form=form, tag_rule_formset=formset))
```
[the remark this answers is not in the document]
snopoke: Fixed in f9f9a94e. When `form.is_valid()` is False the view now rebuilds the formset against the user's submitted `output_schema` (best-effort parse via a new `_submitted_output_schema` helper, falling back to the stored schema if params can't be parsed) so per-row errors reference the new schema.

*— Claude Code*

apps/evaluations/tests/test_tagging_logic.py
snopoke: Fixed in f9f9a94e. Collapsed the equals/range cases into a single parameterized `test_matches` over `(condition_type, condition_value, field_value, expected)`; kept `test_unknown_condition_raises` separate since it asserts a different exception path.

*— Claude Code*

apps/evaluations/tests/test_tagging_logic.py
snopoke: Fixed in f9f9a94e. Split into parameterized `test_valid` and `test_rejected` over `(condition_type, condition_value, field)` cases.

*— Claude Code*

apps/evaluations/forms.py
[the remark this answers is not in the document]
snopoke: Fixed in f9f9a94e.

*— Claude Code*

templates/evaluations/partials/tag_rule_row.html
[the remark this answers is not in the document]
snopoke: Fixed in f9f9a94e. Each label now uses `for="{{ form.<field>.id_for_label }}"` and the matching input/select sets that `id`. Applied to all six fields (Tag name, Field, Condition, Value, Min, Max).

*— Claude Code*

apps/evaluations/admin.py
snopoke: Fixed in f9f9a94e.

*— Claude Code*

apps/evaluations/admin.py
snopoke: Fixed in f9f9a94e.

*— Claude Code*

apps/evaluations/rule_validation.py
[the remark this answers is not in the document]
snopoke: Fixed in f9f9a94e. `ConditionType.RANGE.matches()` now uses `.get("min")` / `.get("max")`; missing or non-numeric bounds raise `TypeError`/`ValueError` from `float()` and are caught by the existing handler, returning `False` instead of `KeyError`.

*— Claude Code*

apps/evaluations/tagging.py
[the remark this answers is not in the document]
snopoke: Fixed in f9f9a94e. Added `KeyError` to the except tuple in `evaluate_rules` for defense in depth (also addressed at the source by switching `RANGE.matches()` to `.get()`).

*— Claude Code*

apps/evaluations/views/evaluator_views.py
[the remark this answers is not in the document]
snopoke: Fixed in f9f9a94e. When `form.is_valid()` is False the view now rebuilds the formset against the user's submitted `output_schema` (best-effort parse via a new `_submitted_output_schema` helper, falling back to the stored schema if params can't be parsed) so per-row errors reference the new schema.

*— Claude Code*

apps/evaluations/tests/test_tagging_logic.py
snopoke: Fixed in f9f9a94e. Collapsed the equals/range cases into a single parameterized `test_matches` over `(condition_type, condition_value, field_value, expected)`; kept `test_unknown_condition_raises` separate since it asserts a different exception path.

*— Claude Code*

apps/evaluations/tests/test_tagging_logic.py
snopoke: Fixed in f9f9a94e. Split into parameterized `test_valid` and `test_rejected` over `(condition_type, condition_value, field)` cases.

*— Claude Code*

apps/evaluations/forms.py
snopoke: Fixed in f9f9a94e. The widget `class` attrs were dead — `tag_rule_row.html` always renders the inputs by hand (it never calls `{{ form.field_name }}`/`{{ form.condition_type }}`), so the widget classes never reached the DOM. Removed the unused attrs from `Meta.widgets` and left the template-side classes as the single source of truth.

*— Claude Code*

apps/evaluations/tagging.py
snopoke: Fixed in f9f9a94e. Added a private `_tag_rules_cache` on the evaluator instance so repeat calls with the same evaluator skip the queryset evaluation. Note that the production call site (`evaluate_single_message_task` in `apps/evaluations/tasks.py`) already does `prefetch_related("tag_rules")`, so the practical effect today is a small speedup for tests/non-prefetched callers — the cache makes the function robust to those paths without relying on the caller.

*— Claude Code*

snopoke (approved):

PR #3233, opened 2026-04-27
perf: stream and chunk session exports for large datasets

### LINKED ISSUE #3232: perf: improve session export performance for large datasets
## Problem

Session exports are slow and memory-intensive for large datasets. The current implementation in `apps/experiments/export.py` has several issues that compound at scale.

### 1. Entire CSV buffered in `io.StringIO`
All rows are accumulated in a single in-memory `StringIO` before being written to disk or returned to the client. For large experiments (e.g. 10k+ traces × 2 rows each) this creates a large, growing memory allocation that isn't released until the task completes.

### 2. All traces loaded into memory at once
The `Trace` queryset uses `prefetch_related` for tags, comments, and chat relations. `prefetch_related` materialises **all** related objects in Python memory simultaneously — so the entire dataset (trace objects, input/output messages, tags, comments, session chats) is in memory at the same time.

### 3. Repeated per-session serialisation
For sessions with many traces, `json.dumps(session.state)`, `session.get_platform_name()`, and participant fields are recomputed/serialised on **every trace row** for the same session. A busy session with 100 exchanges hits those 5 calls 200 times unnecessarily.

### 4. `session__chat` not in `select_related`
`ExperimentSession.chat` is a `OneToOneField` that could be JOINed efficiently. Currently it's implicitly populated via the `prefetch_related("session__chat__tags")` chain, which is less explicit and misses the optimisation of a single JOIN.

### 5. `analysis/views.py: export_sessions` runs synchronously
Unlike the main export (which uses Celery via `async_export_chat`), the analysis `export_sessions` view calls `filtered_export_to_csv` directly in the request/response cycle. This will time out for large datasets.

## Proposed Changes

1. **Chunked processing** — process traces in batches (e.g. 1,000 at a time) using keyset pagination so memory usage stays flat regardless of dataset size
2. **Streaming CSV output** — refactor `filtered_export_to_csv` to yield rows as a generator; use `StreamingHttpResponse` in the sync view and chunked writing in the Celery task
3. **Add `session__chat` to `select_related`** — explicit JOIN rather than relying on intermediate prefetch
4. **Per-session value cache** — compute session-level values once per session, not once per message row
5. **Async-ify the analysis export view** — offload to a Celery task matching the `async_export_chat` pattern (separate issue / follow-up)

## Acceptance Criteria

- [ ] Exporting a large experiment (10k+ sessions) does not cause memory spikes proportional to dataset size
- [ ] Export time scales linearly with dataset size, not quadratically
- [ ] Existing export tests continue to pass
- [ ] No change to CSV output format or column order

ocs-percy:
Closes #3232

## Summary

Session exports were slow and memory-intensive for large datasets. This PR addresses the root causes identified in #3232 by introducing chunked DB reads and streaming CSV output.

### `apps/experiments/export.py`

- **`generate_export_rows()`** (new): generator that yields the header row then one data row per message. Traces are fetched in batches of 1,000 using keyset pagination (`pk__gt` + `LIMIT`), so ORM objects for the full dataset are never in memory simultaneously.
- **Per-session value cache**: `json.dumps(session.state)`, `session.get_platform_name()`, participant fields, and chat tags/comments are computed once per session and reused for every trace in that session. A busy session with 100 exchanges previously hit those 5 serialisation calls 200 times; now it's 1.
- **`session__chat` in `select_related`**: `ExperimentSession.chat` is a `OneToOneField`; adding it to `select_related` gives a single JOIN rather than relying on the implicit intermediate prefetch step.
- **`export_rows_to_csv_stream()`** (new): wraps any row iterator into a generator of CSV-formatted strings, suitable for `StreamingHttpResponse`.
- **`export_to_tempfile()`** (new): writes the export to a `SpooledTemporaryFile` (stays in memory up to 10 MB, spills to disk beyond that) and returns it seeked to 0.
- **`filtered_export_to_csv()`** retained for backward compatibility — now implemented via `generate_export_rows()`.

### `apps/experiments/tasks.py`

- `async_export_chat`: switched from `filtered_export_to_csv` + `StringIO` to `export_to_tempfile`. The intermediate CSV now spills to disk for large exports rather than requiring one large in-memory buffer.

### `apps/analysis/views.py`

- `export_sessions`: switched from `HttpResponse` with a fully-buffered `StringIO` to `StreamingHttpResponse` + `export_rows_to_csv_stream(generate_export_rows())`. The response now streams to the client line-by-line without accumulating the full CSV in memory first.

### Tests

- Updated `test_trace_id_export` mock to support the new `.filter(pk__gt=...).[:chunk_size]` keyset-pagination call chain.

## What's not in this PR (follow-up in #3232)

- Async-ifying the analysis `export_sessions` view (currently synchronous in the request cycle — covered by #3232 notes)
- Database composite index on `Trace(session_id, input_message_id)`

### DIFF
--- a/apps/analysis/views.py
+++ b/apps/analysis/views.py
@@ -2,7 +2,7 @@ from functools import cached_property
 
 from django.conf import settings
 from django.contrib import messages
-from django.http import FileResponse, HttpResponse, JsonResponse
+from django.http import FileResponse, HttpResponse, JsonResponse, StreamingHttpResponse
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 from django.views.decorators.http import require_POST
@@ -11,7 +11,7 @@ from django_htmx.http import HttpResponseClientRedirect, HttpResponseClientRefre
 from django_tables2 import RequestConfig, SingleTableView
 
 from apps.chatbots.tables import ChatbotSessionsTable
-from apps.experiments.export import filtered_export_to_csv
+from apps.experiments.export import export_rows_to_csv_stream, generate_export_rows
 from apps.experiments.models import Experiment, ExperimentSession
 from apps.teams.mixins import LoginAndTeamRequiredMixin
 
@@ -153,11 +153,10 @@ def download_analysis_results(request, team_slug, pk):
 def export_sessions(request, team_slug, pk):
     analysis = get_object_or_404(TranscriptAnalysis, id=pk, team__slug=team_slug)
     sessions = analysis.sessions.all()
-    csv_content = filtered_export_to_csv(
+    rows = generate_export_rows(
         analysis.experiment, sessions, translation_language=analysis.translation_language
     )
-
-    response = HttpResponse(csv_content.getvalue(), content_type="text/csv")
+    response = StreamingHttpResponse(export_rows_to_csv_stream(rows), content_type="text/csv")
     response["Content-Disposition"] = f'attachment; filename="{analysis.name}_sessions_export.csv"'
     return response
 
--- a/apps/experiments/export.py
+++ b/apps/experiments/export.py
@@ -111,6 +111,67 @@ def _get_export_header(translation_language=None):
     return header
 
 
+def _build_session_cache_entry(session) -> dict:
+    """Compute and return the per-session values used in every export row for that session."""
+    return {
+        "platform": session.get_platform_name(),
+        "session_tags": _format_tags(session.chat.tags.all()),
+        "session_comments": _format_comments(session.chat.comments.all()),
+        "external_id": session.external_id,
+        "state": json.dumps(session.state),
+        "participant_name": session.participant.name,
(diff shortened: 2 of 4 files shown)


ocs-percy:
Addressed the CodeRabbit and CodeScene review feedback in the latest commit (ca93cca):

- **`from collections.abc`**: replaced deprecated `typing.Generator` / `typing.Iterator` aliases, matching the codebase convention.
- **Binary-mode tempfile**: switched `export_to_tempfile` to `mode="wb+"` + `io.TextIOWrapper` for CSV writing — the returned file now yields `bytes` directly, removing the redundant `.encode("utf-8")` in `async_export_chat`.
- **Docstring**: updated to recommend context-manager usage and document the binary return type.

The CodeScene advisory flags (Large Method in test, Excess Function Args on `_build_message_row`) are not blocking and those are trade-offs of the design; happy to address if needed.

snopoke (approved):

PR #3238, opened 2026-04-28
perf: gzip-compress chat exports to reduce size and download time

ocs-percy:
## Summary

CSV exports compress very well (~80–90% reduction) because the data is highly repetitive — experiment names, participant IDs, message types, state JSON, etc. appear across thousands of rows.

A **327 MB export** becomes roughly **30–50 MB**, turning a ~20 minute download into a couple of minutes.

## Changes

**`apps/experiments/export.py`**
- Added `compress=False` parameter to `export_to_tempfile()`
- When `compress=True`, rows are written through `gzip.GzipFile` so the spooled temp file holds compressed bytes rather than raw CSV
- `gz.close()` is called explicitly before `tmp.seek(0)` to flush the gzip trailing checksum

**`apps/experiments/tasks.py`**
- `async_export_chat` now calls `export_to_tempfile(compress=True)`
- Filename changed from `.csv` → `.csv.gz`
- Content type changed to `application/gzip`

## Compatibility

- Existing callers of `export_to_tempfile()` are unaffected (`compress` defaults to `False`)
- **macOS / Linux:** `.csv.gz` files auto-open natively (double-click or `gunzip`)
- **Windows 11:** handles `.gz` files natively without third-party tools
- Pairs with the S3 presigned URL redirect from #3235 — the presigned URL already passes `ResponseContentType` from `file.content_type`, so `application/gzip` is set correctly

### DIFF
--- a/apps/experiments/export.py
+++ b/apps/experiments/export.py
@@ -1,4 +1,5 @@
 import csv
+import gzip
 import io
 import json
 import tempfile
@@ -223,7 +224,9 @@ def export_rows_to_csv_stream(rows: Iterator[list]) -> Generator[str, None, None
         buffer.truncate(0)
 
 
-def export_to_tempfile(experiment, sessions_queryset, translation_language=None) -> tempfile.SpooledTemporaryFile:
+def export_to_tempfile(
+    experiment, sessions_queryset, translation_language=None, compress=False
+) -> tempfile.SpooledTemporaryFile:
     """Write the CSV export to a spooled temporary file and return it, seeked to 0.
 
     Use as a context manager (``with export_to_tempfile(...) as tmp:``) so that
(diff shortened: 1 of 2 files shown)


snopoke (approved):

apps/experiments/export.py
```diff
@@ -233,17 +236,37 @@ def export_to_tempfile(experiment, sessions_queryset, translation_language=None)
...
+        # try to close the GzipFile a second time on garbage collection.
+        gz = gzip.GzipFile(fileobj=tmp, mode="wb")
+        text_wrapper = io.TextIOWrapper(gz, encoding="utf-8", newline="")
+        writer = csv.writer(text_wrapper, delimiter=",", quotechar='"', quoting=csv.QUOTE_MINIMAL)
+        for row in generate_export_rows(experiment, sessions_queryset, translation_language):
+            writer.writerow(row)
+        text_wrapper.flush()
+        text_wrapper.detach()
+        gz.close()  # finalises the gzip stream and writes the trailing checksum
+    else:
```
SmittieC, line 259: is this path ever used?

PR #3256, opened 2026-04-30
Add attachment support to the email channel (Phase 2)

### LINKED ISSUE #3246: Add support in inbound and outbound attachments in the email channel
## Background

The email channel shipped as text-only (Phase 1). Attachment support was explicitly deferred:

> **Non-Goals (Phase 1):** Attachment processing (pass-through to LLM, PDF extraction, etc.)

This issue covers Phase 2: bidirectional attachment support — inbound (user emails a file to the bot) and outbound (bot attaches a file to its reply, e.g. via `add_file_attachment()` in a Python node).

---

## Inbound Attachments

When a user sends an email with attached files, those attachments should be parsed, saved, and passed through the pipeline so the LLM can use them (e.g. read a PDF, analyse a CSV).

**What needs to change:**

- `EmailMessage.parse()` in `apps/channels/datamodels.py` — iterate over `inbound.attachments` from `AnymailInboundMessage`, save each to the `File` model, and populate `BaseMessage.attachments`
- `EmailMessage` datamodel — may need fields for raw attachment metadata before `File` records are created
- `handle_email_message` Celery task — ensure the team/storage context is available when saving files during async processing

**Considerations:**

- Size limits — large attachments should be rejected or truncated with a warning
- Unsupported MIME types — decide whether to silently skip or surface to the LLM as metadata
- The `Attachment.send_to_llm` and `upload_to_assistant` flags on the existing `Attachment` model already control how files are passed to LLM/assistant nodes — email attachments should use those same flags

---

## Outbound Attachments

When the pipeline produces file attachments (e.g. a Python node calls `add_file_attachment()`), the email reply should include them as MIME attachments so the user can download them.

**What needs to change:**

- `EmailSender.send_file()` in `apps/channels/channels_v2/email_channel.py` — implement using Django's `EmailMessage.attach(filename, content, mimetype)`
- `EmailChannel._get_capabilities()` — set `supports_files=True` and provide a `can_send_file` callable with any size/type restrictions
- `EmailSender.send_text()` — consider whether files should be sent in the same `EmailMessage` as the text body (preferred for threading) or as a separate follow-up

---

## Files Expected to Change

| File | Change |
|------|--------|
| `apps/channels/datamodels.py` | Parse `inbound.attachments` in `EmailMessage.parse()`, populate `BaseMessage.attachments` |
| `apps/channels/channels_v2/email_channel.py` | Implement `EmailSender.send_file()`; set `supports_files=True` in capabilities |
| `apps/channels/tasks.py` | Ensure file storage context available during async task execution |
| `apps/channels/tests/test_email_channel.py` | Tests for inbound parsing and outbound file attachment |

snopoke:
Closes #3246

### Product Description

The email channel now supports bidirectional attachments.

- **Inbound:** when a user emails files (PDFs, CSVs, images, etc.) to the bot, they are saved and made available to the LLM/pipeline. Files are stored under the team's storage as `File` records with `purpose=MESSAGE_MEDIA`, the same lifecycle as voice notes and other channel media. Rejected attachments (oversized, executable types, claimed-vs-detected mismatches) are surfaced to the LLM as bracketed notes in the user's message text so the bot can explain the rejection.
- **Outbound:** when the pipeline produces files (e.g. via `add_file_attachment()` in a Python node), they are attached as MIME parts to the bot's email reply, combined with the response text in a single threaded message rather than separate follow-ups.

### Technical Description

**Inbound flow.** `email_inbound_handler` was promoted from a best-effort pre-filter to the full router: it resolves channel + session up-front, then persists each accepted attachment synchronously via `File.create()` with team context known. The Celery task carries `channel_id`, `session_id`, and `attachment_file_ids` only — never raw bytes through the Redis broker.

**Intake validation.** `python-magic` detects the canonical content type from bytes (defeating header/extension spoofing); a three-layer denylist checks extension, claimed header type, and magic-detected type independently. A cross-category mismatch check rejects "claimed image, detected executable" lies, with `application/octet-stream` exemptions on either side and a curated `EMAIL_TEXT_LIKE_APPLICATION_TYPES` allowlist (JSON, XML, YAML, etc.) so legitimate textual `application/*` types aren't false-positive rejected when magic returns `text/plain`. 20 MB per-attachment cap. Per-attachment failures are isolated; one storage error doesn't drop the whole email.

**Pipeline integration.** `BaseMessage` gains a generic `attachment_file_ids: list[int]` field and the default pipeline gains a new `AttachmentHydrationStage` that hydrates `Attachment` objects after `SessionActivationStage`, so each `Attachment.download_link` references a real session. The stage is a no-op for channels that don't pre-persist files.

**Outbound flow.** `ChannelSender` gains a `flush()` hook (no-op default; existing channels inherit it without changes) called by `ResponseSendingStage` after all text/voice/file sends. `EmailSender` buffers `send_text`/`send_file` calls and commits a single Django `EmailMessage` with body + MIME attachments + threading headers in `flush()`, then resets state for any subsequent send. Flush failures are wrapped as `MessageDeliveryFailure` so SMTP errors trigger the existing notification path. `EmailChannel.supports_files` is now `True`; `_can_send_file` delegates to a new `can_send_on_email` checker registered in `FILE_SENDABILITY_CHECKERS`. Files that fail the checker (oversize or denylisted) fall back to inline download URLs in the body via the existing `ResponseFormattingStage` logic.

**Backwards compatibility.** `handle_email_message` accepts `channel_id`/`session_id` directly when provided, but falls back to the legacy `get_email_experiment_channel` routing chain when called without them, so in-flight pre-deploy tasks still process correctly. The fallback path can be removed in the next release cycle.

### Migrations

No DB schema changes. The Celery payload shape changes additively — the legacy task signature is preserved for one release cycle.

### Demo

To exercise the inbound path locally, send an inbound email through the AWS SES webhook handler with one or more attachments; check that the bot reply (1) acknowledges any rejected files in its text response and (2) attaches any pipeline-produced files in the same threaded reply.

### Docs and Changelog

Design spec: \`docs/superpowers/specs/2026-04-29-email-attachments-design.md\`
Implementation plan: \`docs/superpowers/plans/2026-04-29-email-attachments.md\`


apps/channels/channels_v2/stages/core.py
```diff
@@ -529,3 +530,36 @@ def _append_attachment_links(self, text: str, files: list, ctx: MessageProcessin
...
+    (e.g. EmailChannel) populate ctx.message.attachment_file_ids; this
+    stage converts those IDs into Attachment objects with download_links
+    that reference a real session. No-op for channels that don't use
+    this pattern.
+    """
+
+    def should_run(self, ctx: MessageProcessingContext) -> bool:
+        return bool(
+            ctx.message
+            and getattr(ctx.message, "attachment_file_ids", None)
```
SmittieC, line 553: Nit: It's not needed to use `getattr` if `attachment_file_ids` is an attribute on `message`?

apps/channels/channels_v2/email_channel.py
```diff
@@ -156,8 +167,86 @@ def _domain_from_address(email_address: str) -> str:
     return email_address
 
 
+def _detect_content_type(content: bytes, fallback: str = "") -> str:
+    try:
+        detected = magic.from_buffer(content[:2048], mime=True)
```
SmittieC, line 172: @claude this is similar to File.get_content_type. Please make a PR to refactor this to use a single utility that detects content type. Then there's another place where a similar pattern is used in File.create

PR #3296, opened 2026-05-06
Remove TypeSelectForm pattern in favour of static type-specific forms

### LINKED ISSUE #3286: Refactor: remove TypeSelectForm pattern in favour of static type-specific forms
## Background

The `TypeSelectForm` pattern (`apps/generics/type_select_form.py`) is used to handle forms where a model has a `type` field that determines which secondary configuration form is shown. The current implementation renders **all** secondary forms on the page simultaneously, using Alpine.js (`x-show`, `x-bind:disabled`, `x-bind:required`) to show/hide the active one based on the selected type.

This is overly complex:
- All forms (and their validation logic) are always present in the DOM
- Alpine.js attribute binding is needed to fake required/disabled state per form
- The `TypeSelectForm` dataclass, `BaseTypeSelectFormView`, and the `generic/type_select_form.html` template add an extra abstraction layer that makes the code harder to follow

## Proposed Approach

Move type selection **before** the form loads — e.g. a "Create" button that opens a dropdown of type options. Each option links to a type-specific create URL. The create/edit view then receives the type as a URL parameter or from the existing object, and renders a single static form for that type.

This gives:
- One form per type — simpler templates, no Alpine binding tricks
- Standard Django form validation with no workarounds
- No special base view or dataclass needed

## Current Usages

The pattern appears in two places:

### 1. Service Providers ()
- `apps/service_providers/utils.py` — `get_service_provider_form()` builds a `TypeSelectForm` with secondary forms per provider type (LLM, messaging, etc.)
- `apps/service_providers/views.py` — `CreateServiceProvider` extends `BaseTypeSelectFormView`
- Template: `templates/service_providers/llm_provider_form.html` extends `generic/type_select_form.html`

### 2. Events / Event Actions (`apps/events/`)
- `apps/events/forms.py` — `EventActionTypeSelectForm` subclasses `TypeSelectForm`

## What Can Be Removed / Simplified

Once migrated:
- `apps/generics/type_select_form.py` — delete entirely
- `apps/generics/exceptions.py` (`TypeSelectFormError`) — no longer needed
- `BaseTypeSelectFormView` in `apps/generics/views.py` — remove
- `templates/generic/type_select_form.html` — remove
- All `x-bind:disabled` / `x-bind:required` Alpine attribute injection from form fields

snopoke:
### Product Description

When creating a new service provider (LLM, voice, messaging, auth, tracing), the **Add new** button is now a dropdown listing each available subtype — clicking a subtype takes the user straight to a form for that one provider type. Editing existing providers works the same as before. The event-trigger create / edit page no longer ships every action's fields to the browser; instead, picking an action type loads just that form server-side via HTMX.

The visible UX is otherwise unchanged. Closes #3286.

### Technical Description

Removes the \`TypeSelectForm\` abstraction (\`apps/generics/type_select_form.py\`), the \`BaseTypeSelectFormView\` base view, the \`templates/generic/type_select_form.html\` template, and the per-field \`x-bind:disabled\` / \`x-bind:required\` Alpine attribute injection.

**Service providers**
- New URL pattern: \`<provider_type>/create/<subtype>/\`. Edit URL unchanged (subtype derived from the instance).
- \`CreateServiceProvider\` is now a single-form view that renders one \`ModelForm\` + one config form. \`LlmProviderView\` is folded back in via a \`provider_type == llm\` branch in the context-builder.
- New \`apps/service_providers/utils.get_available_subtypes(provider, request)\` helper centralises subtype filtering (Waffle \`flag_open_ai_voice_engine\`, \`SLACK_ENABLED\`); used by both the create view's 404 guard and the home-page dropdown.
- New template \`templates/service_providers/provider_form.html\` and a small \`{% service_provider_subtype_choices %}\` template tag wire the dropdown into \`object_home_content.html\`.
- Type field is rendered as a hidden input and \`disabled=True\`, so its value is fixed by the URL on create and by the instance on edit.

**Events / event actions**
- \`EventActionTypeSelectForm\` and \`get_action_params_form\` removed. Replaced by \`ACTION_PARAMS_FORMS\` registry + \`build_action_params_form()\`.
- New HTMX endpoint \`events:action_params_form\` returns the secondary form fragment for a given \`action_type\`. The \`manage_event.html\` action-type select swaps it in on \`change\`. Trigger-type filtering of action options stays in Alpine but now dispatches a native \`change\` event so HTMX picks it up.

**Cleanup**
- Deleted: \`apps/generics/type_select_form.py\`, \`apps/generics/exceptions.py\`, \`templates/generic/type_select_form.html\`, \`BaseTypeSelectFormView\` in \`apps/generics/views.py\`.
- Fixed unrelated factory bugs surfaced by the stricter type-resolution: \`MessagingProviderFactory\` (missing \`type\`), \`TraceProviderFactory\` (was using \`AuthProviderType.commcare\`).

### Migrations

No DB migrations.

### Demo

n/a — UX is unchanged for normal flows. The only visible difference is that **Add new** on the manage-team page is now a dropdown of subtypes.

### Docs and Changelog

## Test Plan

- [ ] Visit \`/a/<team>/manage_team/\` — each provider section shows an **Add new** dropdown listing only enabled subtypes (e.g. \`slack\` is hidden when \`SLACK_ENABLED\` is off).
- [ ] Click an LLM subtype → land on a single-subtype create form with default / custom / embedding model panels populated.
- [ ] Edit an existing provider → the type field is fixed and the correct subtype config form renders.
- [ ] Hit \`/a/<team>/service_providers/voice/create/openaivoiceengine/\` directly with the flag off → 404.
- [ ] Create / edit a static or timeout event trigger; switch action type → params form swaps in via HTMX.
- [ ] Pick \`new_bot_message\` static trigger → the disallowed action options grey out and selection falls back to \`log\`, params form reloads accordingly.


apps/events/forms.py
```diff
@@ -109,30 +108,24 @@ def save(self, commit=True, experiment_id=None):
...
+
+
+def build_action_params_form(action_type, *, data=None, initial=None, team_id, experiment_id):
+    """Build the secondary "params" form for a given EventAction action_type."""
+    form_cls = ACTION_PARAMS_FORMS[action_type]
+    kwargs = {"data": data, "initial": initial}
+    if form_cls is ScheduledMessageConfigForm:
+        kwargs["experiment_id"] = experiment_id
+    elif form_cls is PipelineStartForm:
+        kwargs["team_id"] = team_id
```
snopoke, line 127: Could we standardize these for the forms to avoid having to special case them?

apps/events/views.py
```diff
@@ -32,23 +34,61 @@ def create_static_event_view(request, team_slug: str, experiment_id: str):
...
+            action_type,
+            data=request.POST,
+            team_id=request.team.id,
+            experiment_id=experiment_id,
+        )
         trigger_form = trigger_form_class(request.POST)
-        if action_form.is_valid() and trigger_form.is_valid():
-            saved_action = action_form.save(experiment_id=experiment_id)
+
+        if action_primary_form.is_valid() and action_params_form.is_valid() and trigger_form.is_valid():
```
snopoke: If this statement short circuits the later forms won't have `is_valid` called which won't populate their errors.

apps/events/views.py
```diff
@@ -73,27 +113,38 @@ def _edit_event_view(trigger_type, request, team_slug: str, experiment_id: str,
...
+            initial=trigger.action.params,
+            team_id=request.team.id,
+            experiment_id=experiment_id,
         )
         trigger_form = trigger_form_class(request.POST, instance=trigger)
-
-        if action_form.is_valid() and trigger_form.is_valid():
-            action_form.save(experiment_id=experiment_id)
-            trigger = trigger_form.save(experiment_id=experiment_id)
+        if action_primary_form.is_valid() and action_params_form.is_valid() and trigger_form.is_valid():
```
snopoke: Same comment as above about short circuit behavior.

apps/service_providers/views.py
```diff
@@ -132,86 +138,114 @@ def remove_file(request, team_slug: str, provider_type: str, pk: int, file_id: i
...
+        instance = self._get_instance()
+        primary_form, config_form = get_service_provider_forms(
+            self.provider_type, team=request.team, subtype=subtype, data=request.POST, instance=instance
+        )
+
+        file_formset = None
+        if request.FILES:
+            file_formset = get_file_formset(request, formset_cls=config_form.file_formset_form)
+
+        if primary_form.is_valid() and config_form.is_valid() and (not file_formset or file_formset.is_valid()):
```
snopoke: Similar potential issue with short circuited statement not calling `is_valid` on all forms

templates/experiments/components/prompt_builder_toolbox.html
```diff
@@ -77,7 +77,7 @@
         <div class="flex-none flex flex-col items-center">
             <label class="label" x-show="!$store.promptBuilder.currentState.provider" x-cloak>
                 <span class="text-error">First create an&nbsp;</span>
-                <a class="text-error link" href="{% url "service_providers:new" request.team.slug "llm" %}">LLM Provider</a>
+                <a class="text-error link" href="{% url "single_team:manage_team" request.team.slug %}#llm-service-provider">LLM Provider</a>
```
snopoke:
```suggestion
                <a class="text-error link" href="{% url "single_team:manage_team" request.team.slug %}#llm-and-embedding-model-service-providers">LLM Provider</a>
```

apps/events/views.py
snopoke: Fixed in f233625cb. Now calling `is_valid()` on each form into a local variable before combining, so all forms' errors populate on POST failure.

*— Claude Code*

apps/events/views.py
snopoke: Fixed in f233625cb. Same treatment applied here — `is_valid()` is now called on each form before the combined check.

*— Claude Code*

apps/service_providers/views.py
snopoke: Fixed in f233625cb. Same treatment applied to `CreateServiceProvider.post` — primary, config, and file_formset `is_valid()` are each called into a local variable before the combined check.

*— Claude Code*

templates/experiments/components/prompt_builder_toolbox.html
snopoke: Fixed in f233625cb. Anchor updated to `#llm-and-embedding-model-service-providers` to match the new section title.

*— Claude Code*

apps/service_providers/views.py
```diff
@@ -132,86 +138,113 @@ def remove_file(request, team_slug: str, provider_type: str, pk: int, file_id: i
...
```
snopoke: Fixed in f233625cb — the post() conditional is now decomposed into named locals (`primary_valid`, `config_valid`, `file_formset_valid`) before the combined check.

*— Claude Code*

apps/service_providers/views.py
```diff
@@ -132,86 +138,113 @@ def remove_file(request, team_slug: str, provider_type: str, pk: int, file_id: i
...
+        ctx = {
+            "primary_form": primary_form,
+            "config_form": config_form,
+            "subtype": subtype,
+            "object": instance,
+            "title": f"Edit {instance.name}" if instance else self.provider_type.label,
+            "button_text": "Update" if instance else "Create",
+            "active_tab": "manage-team",
+        }
+        if instance and isinstance(instance, VoiceProvider) and instance.type == VoiceProviderType.elevenlabs.value:
```
snopoke: Fixed in f233625cb — the elevenlabs check is extracted into an `is_elevenlabs_voice` local so the `if` is a single condition.

*— Claude Code*

apps/events/views.py
snopoke: Fixed in f233625cb — `_create_event_view`'s combined `is_valid()` check is now decomposed into named locals (also addresses the short-circuit issue).

*— Claude Code*

apps/events/views.py
snopoke: Fixed in f233625cb — same treatment for `_edit_event_view`.

*— Claude Code*

apps/events/forms.py
snopoke: Noted — declined by author. Keeping the special-case kwarg handling in `build_action_params_form` for now; the registry pattern is small enough that the trade-off didn't seem worth a forms-side refactor.

*— Claude Code*

apps/events/views.py
snopoke: Fixed in f233625cb. Now calling `is_valid()` on each form into a local variable before combining, so all forms' errors populate on POST failure.

*— Claude Code*

apps/events/views.py
snopoke: Fixed in f233625cb. Same treatment applied here — `is_valid()` is now called on each form before the combined check.

*— Claude Code*

apps/service_providers/views.py
snopoke: Fixed in f233625cb. Same treatment applied to `CreateServiceProvider.post` — primary, config, and file_formset `is_valid()` are each called into a local variable before the combined check.

*— Claude Code*

templates/experiments/components/prompt_builder_toolbox.html
snopoke: Fixed in f233625cb. Anchor updated to `#llm-and-embedding-model-service-providers` to match the new section title.

*— Claude Code*

apps/service_providers/views.py
snopoke: Fixed in f233625cb — the post() conditional is now decomposed into named locals (`primary_valid`, `config_valid`, `file_formset_valid`) before the combined check.

*— Claude Code*

apps/service_providers/views.py
snopoke: Fixed in f233625cb — the elevenlabs check is extracted into an `is_elevenlabs_voice` local so the `if` is a single condition.

*— Claude Code*

apps/events/views.py
snopoke: Fixed in f233625cb — `_create_event_view`'s combined `is_valid()` check is now decomposed into named locals (also addresses the short-circuit issue).

*— Claude Code*

apps/events/views.py
snopoke: Fixed in f233625cb — same treatment for `_edit_event_view`.

*— Claude Code*

apps/events/forms.py
snopoke: Noted — declined by author. Keeping the special-case kwarg handling in `build_action_params_form` for now; the registry pattern is small enough that the trade-off didn't seem worth a forms-side refactor.

*— Claude Code*

PR #3309, opened 2026-05-08
feat(evaluations): auto-populate datasets with rule-driven ingestion

### LINKED ISSUE #3044: automated (online) evals - run evals automatically based on criteria
Run evals automatically:

* Add data to the dataset in session mode or message mode depending on the dataset config
  * Depends on #3043
* filter criteria
  * Set filter criteria for importing data e.g. only completed sessions, matching tags etc.
* Allow sampling sessions
* Only run evals on the new data

## Design (from 2026-04-21 discussion)

**Filters live on the dataset.** Configuring filters at the dataset level (rather than on the evaluation config or evaluator) keeps the mental model simple and handles the case where one dataset is shared across multiple evaluation configs with different criteria.

**No explicit user-facing triggers.** Instead of asking users to pick trigger events, sessions are "presented" to continuous datasets at known lifecycle points and the dataset's filter decides whether to accept them. Practical entry points:
* **Session ends** — filter typically checks session state; excludes sessions already in the dataset.
* **Auto-eval finishes** — filter typically checks tags (e.g. the human annotation queue that picks up sessions failing LLM eval).

**Dataset's role is unchanged.** It holds the queue of sessions to be evaluated. Evaluation configs remain linked to a dataset; results remain on the evaluation config. Only the mechanism for how sessions enter the dataset (and which subset gets evaluated) changes.

**Data model impact is minimal.** Reuse the existing saved-filter mechanism, foreign-keyed to the dataset (possibly with a `type` field so these filters don't leak into other filter UIs).

**Checkpointing / dedup (flagged, not solved).** We don't want to re-run filters over all sessions every cycle. Pure timestamp checkpointing doesn't work because an old session can acquire a new tag later. Fallback: `NOT IN (session_ids already in dataset)` — brute force but correct. To be refined.

**Follow-on (not now).** User notifications when items hit the annotation queue, with debounce for bursts.

## Open question

* Confirm with the Connect team: filters at the **dataset level** (our lean) vs. **evaluation-config level**. We agreed to present both options to the team.

## Questions
* Does this create a dataset?
  * Dataset must already exist
* How does it integrate with notifications e.g. if eval fails or result is below threshold.

## Priority / sequencing

1. Session mode (#3043)
2. Tagging (#3193)
3. Auto-evals (this ticket)
4. Annotation-queue refactor — does *not* need to precede auto-emails

snopoke:
### Product Description

Lets teams configure an `EvaluationDataset` to continuously ingest new sessions from a source chatbot via filter criteria, and (per linked evaluation config) automatically run evaluators over only the newly added rows. Closes [#3044](https://github.com/dimagi/open-chat-studio/issues/3044).

### Technical Description

Decisions worth flagging:

- **Session-mode only.** Auto-population is restricted to session-mode datasets — enforced at the model (`DatasetAutoPopulationRule.clean`), the view (`_RuleViewMixin.dispatch` redirects), and the UI (info note in the rules panel). Message-mode is out of scope for v1.
- **Polling, not signals.** A Celery Beat task scans each enabled rule on a 5-minute tick. Bounded `LOOKBACK_DAYS` (default 30, env-tunable) + brute-force `NOT IN` dedup against existing dataset rows. Forward-only floor via `rule.created_at` so new rules don't backfill. Polling was chosen over signal hooks because tag-application doesn't touch `session.updated_at`, so a high-water-mark approach would silently drop late-tagged sessions. Event-based triggers are deferred — see `docs/superpowers/specs/.../#future-changes`.
- **Delta evaluation runs.** New `EvaluationRunType.DELTA` with a `scoped_messages` M2M; `run_evaluation_task` branches explicitly on `DELTA`. The trigger fires from `transaction.on_commit` inside the ingest task so the M2M is committed before the Celery worker picks up the run.
- **Auto-trigger scope.** Only the auto-population path enqueues delta runs in v1. Manual filter-import and CSV-import paths are intentionally unchanged.
- **Concurrency + failure isolation.** Each rule is processed under `select_for_update(skip_locked=True)`. `_handle_rule_failure` runs *after* the atomic block in a fresh transaction so a DB error mid-ingest still records the failure and (after three strikes) auto-disables the rule via `apps.ocs_notifications`.
- **Filter widget reuse.** The rule form uses a new always-visible inline variant (`templates/filters/inline_filter_editor.html`) that shares the Alpine.data factory (`_filter_component_alpine.html`) with the existing button-style widget on session lists. The widget's HTMX preview-table round-trip is skipped when no data source is configured.
- **Periodic task registration.** Via `SCHEDULED_TASKS` + the project's existing `setup_periodic_tasks` post-migrate hook — no migration needed for the schedule itself.

Explicit follow-ups (called out in the spec): lookback-window tuning, filter mode validation in the form, the per-rule "contributed-row count" column, and event-based triggers as an alternative to polling.

Spec: `docs/superpowers/specs/2026-05-07-auto-populate-eval-datasets-design.md`.

### Migrations

`0015_auto_populate_schema` is purely additive (new model + new fields + new enum choice; all defaults preserve existing rows).

### Demo

<div>
    <a href="https://www.loom.com/share/c982ff5d1d3f4778978af4e2d5bfb7aa">
      <p>Live evals demo - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/c982ff5d1d3f4778978af4e2d5bfb7aa">
      
    </a>
  </div>

### Docs and Changelog


apps/evaluations/migrations/0016_register_auto_populate_periodic.py
snopoke, line 1: This migration is unnecessary

apps/evaluations/tests/test_auto_population_models.py
```diff
@@ -0,0 +1,64 @@
...
+from apps.utils.factories.evaluations import (
+    DatasetAutoPopulationRuleFactory,
+    EvaluationDatasetFactory,
+)
+from apps.utils.factories.experiment import ExperimentFactory
+from apps.utils.factories.team import TeamFactory
+
+
+@pytest.mark.django_db()
+def test_rule_defaults():
```
snopoke: unnecessary test

apps/evaluations/tests/test_delta_evaluation_run.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+@pytest.mark.django_db()
+def test_full_run_has_empty_scope_by_default():
+    run = EvaluationRunFactory.create(type=EvaluationRunType.FULL)
+    assert run.scoped_messages.count() == 0
+
+
+@pytest.mark.django_db()
+def test_evaluation_config_auto_run_on_append_defaults_false():
```
snopoke: unnecessary test

apps/evaluations/tests/test_delta_evaluation_run.py
```diff
@@ -0,0 +1,143 @@
...
+    msg = EvaluationMessageFactory.create()
+    run.scoped_messages.add(msg)
+
+    run.refresh_from_db()
+    assert run.type == EvaluationRunType.DELTA
+    assert list(run.scoped_messages.all()) == [msg]
+
+
+@pytest.mark.django_db()
+def test_full_run_has_empty_scope_by_default():
```
snopoke: unnecessary test

apps/evaluations/tests/test_delta_evaluation_run.py
```diff
@@ -0,0 +1,143 @@
...
+from apps.utils.factories.evaluations import (
+    EvaluationConfigFactory,
+    EvaluationMessageFactory,
+    EvaluationRunFactory,
+    EvaluatorFactory,
+)
+
+
+@pytest.mark.django_db()
+def test_evaluation_run_can_be_created_as_delta_with_scope():
```
snopoke: unnecessary test

apps/evaluations/tests/test_delta_evaluation_run.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+@pytest.mark.django_db()
+def test_evaluation_config_auto_run_on_append_defaults_false():
+    config = EvaluationConfigFactory.create()
+    assert config.auto_run_on_append is False
+
+
+@pytest.mark.django_db()
+def test_evaluation_config_auto_run_on_append_can_be_set():
```
snopoke: unnecessary test

apps/evaluations/auto_population.py
```diff
@@ -0,0 +1,196 @@
...
+from apps.evaluations.notifications import auto_population_rule_disabled_notification
+from apps.evaluations.utils import make_session_evaluation_messages
+from apps.experiments.filters import ExperimentSessionFilter
+from apps.experiments.models import ExperimentSession
+from apps.web.dynamic_filters.datastructures import FilterParams
+
+logger = get_task_logger("ocs.evaluations")
+
+
+def _versions_of(experiment):
```
snopoke: Instead of this you can use the existing `experiment.version_family_ids`:

```python
experiment_id__in=experiment.version_family_ids
```

apps/evaluations/auto_population.py
```diff
@@ -0,0 +1,196 @@
...
+    if not rule.is_enabled:
+        auto_population_rule_disabled_notification(
+            rule, reason=f"{rule.consecutive_failure_count} consecutive failures"
+        )
+
+
+def _ingest_rule_session_mode(rule: DatasetAutoPopulationRule, created_floor) -> list[EvaluationMessage]:
+    qs = ExperimentSession.objects.filter(
+        team=rule.team,
+        experiment__in=_versions_of(rule.source_experiment),
```
snopoke: `ExperimentSession.experiment` is always expected to be a working version so we shouldn't need to filter by all the IDs, only the working version ID.

templates/evaluations/components/auto_population_rules_panel.html
```diff
@@ -0,0 +1,61 @@
...
+              </form>
+              <form method="post"
+                    action="{% url 'evaluations:auto_population_rule_delete' request.team.slug rule.id %}"
+                    class="inline"
+                    hx-post="{% url 'evaluations:auto_population_rule_delete' request.team.slug rule.id %}"
+                    hx-confirm="{% translate "Delete this rule?" %}"
+                    hx-on::after-request="if (event.detail.successful) location.reload();">
+                {% csrf_token %}
+                <button type="submit" class="link text-red-600">{% translate "Delete" %}</button>
+              </form>
```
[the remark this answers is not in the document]
snopoke: Not applicable — fixed differently in 1bce29841 by removing `dataset_id` from the `auto_population_rule_delete` URL pattern entirely (matching the toggle URL pattern). The delete view doesn't need it in the URL since it reads `dataset_id` from the rule object for `get_success_url`.

*— Claude Code*

apps/evaluations/auto_population.py
```diff
@@ -0,0 +1,190 @@
...
+    else:
+        eval_messages = EvaluationMessage.create_from_sessions(
+            team=rule.team,
+            external_session_ids=session_external_ids,
+        )
+
+    if not eval_messages:
+        return []
+
+    existing_pairs = set(
```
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. `_handle_rule_failure` now runs after the atomic block in a fresh transaction (re-fetching the rule), so a DB error during ingestion no longer leaves `rule.save()` running inside an aborted transaction.

*— Claude Code*

apps/evaluations/auto_population.py
```diff
@@ -0,0 +1,145 @@
...
+                )
+                if rule is None:
+                    continue  # locked by another worker, or disabled mid-tick
+                try:
+                    _ingest_rule(rule)
+                except Exception as e:  # noqa: BLE001 - per-rule isolation is the point
+                    logger.exception("Auto-population rule %s failed: %s", rule.id, e)
+                    _handle_rule_failure(rule, e)
+        except Exception:  # noqa: BLE001 - last-resort guard
+            logger.exception("Unexpected outer error processing auto-population rule %s", rule_id)
```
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. `_handle_rule_failure` now runs after the atomic block in a fresh transaction (re-fetching the rule), so a DB error during ingestion no longer leaves `rule.save()` running inside an aborted transaction.

*— Claude Code*

apps/evaluations/auto_population.py
```diff
@@ -0,0 +1,190 @@
...
+                    .filter(id=rule_id, is_enabled=True)
+                    .first()
+                )
+                if rule is None:
+                    continue  # locked by another worker, or disabled mid-tick
+                try:
+                    _ingest_rule(rule)
+                except Exception as e:  # noqa: BLE001 - per-rule isolation is the point
+                    logger.exception("Auto-population rule %s failed: %s", rule.id, e)
+                    _handle_rule_failure(rule, e)
```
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. `_handle_rule_failure` now runs after the atomic block in a fresh transaction.

*— Claude Code*

apps/evaluations/auto_population.py
```diff
@@ -0,0 +1,196 @@
...
+    if rule.consecutive_failure_count >= DatasetAutoPopulationRule.AUTO_DISABLE_FAILURE_THRESHOLD:
+        rule.is_enabled = False
+        update_fields.append("is_enabled")
+
+    rule.save(update_fields=update_fields)
+
+    if not rule.is_enabled:
+        auto_population_rule_disabled_notification(
+            rule, reason=f"{rule.consecutive_failure_count} consecutive failures"
+        )
```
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. `_handle_rule_failure` (which calls the notification helper) now runs after the atomic block in a fresh transaction, so the failure-state write doesn\'t get rolled back.

*— Claude Code*

apps/evaluations/tables.py
```diff
@@ -110,6 +110,20 @@ class EvaluationRunTable(tables.Table):
...
+            "<span class='badge badge-info'>delta · {{ record.scoped_messages.count }}</span>"
+            "{% elif record.type == 'preview' %}"
+            "<span class='badge'>preview</span>"
+            "{% else %}"
+            "<span class='badge badge-ghost'>full</span>"
+            "{% endif %}"
+        ),
+        verbose_name="Type",
+        orderable=False,
+    )
```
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. The template column now uses `{{ record.scoped_messages.all|length }}` which hits the prefetch cache rather than issuing a COUNT per row.

*— Claude Code*

apps/evaluations/tables.py
```diff
@@ -110,6 +110,20 @@ class EvaluationRunTable(tables.Table):
         template_name="evaluations/evaluation_run_status_column.html", verbose_name="Status", orderable=False
     )
 
+    type = TemplateColumn(
+        template_code=(
+            "{% if record.type == 'delta' %}"
+            "<span class='badge badge-info'>delta · {{ record.scoped_messages.count }}</span>"
```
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. Switched to `scoped_messages.all|length` so the prefetch wired up in `EvaluationRunTableView` is reused.

*— Claude Code*

apps/evaluations/models.py
```diff
@@ -326,6 +398,8 @@ def run(self, run_type=EvaluationRunType.FULL) -> EvaluationRun:
             status=EvaluationRunStatus.PENDING,
             type=run_type,
         )
+        if scoped_messages is not None:
+            run.scoped_messages.add(*scoped_messages)
 
         from apps.evaluations.tasks import (  # noqa: PLC0415 - circular: evaluations.tasks imports evaluations.models
             run_evaluation_task,
```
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. The run history table now renders the scope size via `scoped_messages.all|length`, which uses the prefetch cache. (The run results page `get_table_data` doesn\'t call `scoped_messages.count` — it filters results by `message_id__in=scoped_ids`.)

*— Claude Code*

apps/evaluations/forms.py
```diff
@@ -1004,3 +1012,46 @@ def save(self, commit=True):
...
+            raise forms.ValidationError("Source chatbot must belong to your team.")
+        return experiment
+
+    def save(self, commit=True):
+        instance = super().save(commit=False)
+        instance.team = self.team
+        instance.dataset = self.dataset
+        if commit:
+            instance.save()
+        return instance
```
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. `__init__` now pre-sets `instance.team` and `instance.dataset` so Django\'s default `_post_clean` runs `full_clean` against a fully-populated instance. The form overrides `_post_clean` to route errors keyed to fields not on the form (dataset/team) into non-field errors so `add_error` doesn\'t blow up.

*— Claude Code*

apps/evaluations/migrations/0016_register_auto_populate_periodic.py
snopoke: Fixed in 9891748b9 — migration deleted. The periodic task registers via `SCHEDULED_TASKS` + the `setup_periodic_tasks` post-migrate hook, so the dependency-only migration was redundant.

*— Claude Code*

apps/evaluations/tests/test_auto_population_models.py
snopoke: Fixed in 9891748b9 — test deleted.

*— Claude Code*

apps/evaluations/tests/test_delta_evaluation_run.py
snopoke: Fixed in 9891748b9 — test deleted.

*— Claude Code*

apps/evaluations/tests/test_delta_evaluation_run.py
snopoke: Fixed in 9891748b9 — test deleted.

*— Claude Code*

apps/evaluations/tests/test_delta_evaluation_run.py
snopoke: Fixed in 9891748b9 — test deleted.

*— Claude Code*

apps/evaluations/tests/test_delta_evaluation_run.py
snopoke: Fixed in 9891748b9 — test deleted.

*— Claude Code*

apps/evaluations/auto_population.py
snopoke: Fixed in 55255ded4 — `_versions_of` was dropped entirely, replaced with `experiment=rule.source_experiment` (since real chat sessions are always on the working version and the form restricts `source_experiment` to working versions).

*— Claude Code*

apps/evaluations/auto_population.py
snopoke: Fixed in 55255ded4 — `_versions_of` was dropped entirely; the queryset now filters by `experiment=rule.source_experiment` directly.

*— Claude Code*

apps/evaluations/views/auto_population_views.py
```diff
@@ -0,0 +1,87 @@
...
+    model = DatasetAutoPopulationRule
+
+    def get_queryset(self):
+        return DatasetAutoPopulationRule.objects.filter(team=self.request.team)
+
+    def get_success_url(self):
+        return reverse(
+            "evaluations:dataset_edit",
+            args=[self.request.team.slug, self.get_object().dataset_id],
+        )
```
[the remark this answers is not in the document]
snopoke: Not applicable — the dedup excludes sessions whose `id` is already linked from `dataset.messages` via `session_id`. An orphaned `EvaluationMessage` (session FK nulled by `SET_NULL`) means the original session was deleted; that session no longer exists in `ExperimentSession` and can\'t be re-ingested. No re-ingest path produces duplicates.

*— Claude Code*

templates/evaluations/components/auto_population_rules_panel.html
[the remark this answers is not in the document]
snopoke: test reply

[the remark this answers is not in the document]
snopoke: (Ignore previous test message.) Not applicable — fixed differently in 1bce29841 by removing `dataset_id` from the `auto_population_rule_delete` URL pattern entirely (matching the toggle URL pattern). The delete view doesn't need it in the URL since it reads `dataset_id` from the rule object for `get_success_url`.

*— Claude Code*

apps/evaluations/auto_population.py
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. `_handle_rule_failure` now runs after the atomic block in a fresh transaction (re-fetching the rule), so a DB error during ingestion no longer leaves `rule.save()` running inside an aborted transaction.

*— Claude Code*

apps/evaluations/auto_population.py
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. `_handle_rule_failure` now runs after the atomic block in a fresh transaction (re-fetching the rule), so a DB error during ingestion no longer leaves `rule.save()` running inside an aborted transaction.

*— Claude Code*

apps/evaluations/auto_population.py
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. `_handle_rule_failure` now runs after the atomic block in a fresh transaction.

*— Claude Code*

apps/evaluations/auto_population.py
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. `_handle_rule_failure` (which calls the notification helper) now runs after the atomic block in a fresh transaction, so the failure-state write doesn't get rolled back.

*— Claude Code*

apps/evaluations/tables.py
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. The template column now uses `{{ record.scoped_messages.all|length }}` which hits the prefetch cache rather than issuing a COUNT per row.

*— Claude Code*

apps/evaluations/tables.py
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. Switched to `scoped_messages.all|length` so the prefetch wired up in `EvaluationRunTableView` is reused.

*— Claude Code*

apps/evaluations/models.py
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. The run history table now renders the scope size via `scoped_messages.all|length`, which uses the prefetch cache. (The run results page `get_table_data` doesn't call `scoped_messages.count` — it filters results by `message_id__in=scoped_ids`.)

*— Claude Code*

apps/evaluations/forms.py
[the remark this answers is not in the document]
snopoke: Fixed in 9891748b9. `__init__` now pre-sets `instance.team` and `instance.dataset` so Django's default `_post_clean` runs `full_clean` against a fully-populated instance. The form overrides `_post_clean` to route errors keyed to fields not on the form (dataset/team) into non-field errors so `add_error` doesn't blow up.

*— Claude Code*

apps/evaluations/migrations/0016_register_auto_populate_periodic.py
snopoke: Fixed in 9891748b9 — migration deleted. The periodic task registers via `SCHEDULED_TASKS` + the `setup_periodic_tasks` post-migrate hook, so the dependency-only migration was redundant.

*— Claude Code*

apps/evaluations/tests/test_auto_population_models.py
snopoke: Fixed in 9891748b9 — test deleted.

*— Claude Code*

apps/evaluations/tests/test_delta_evaluation_run.py
snopoke: Fixed in 9891748b9 — test deleted.

*— Claude Code*

apps/evaluations/tests/test_delta_evaluation_run.py
snopoke: Fixed in 9891748b9 — test deleted.

*— Claude Code*

apps/evaluations/tests/test_delta_evaluation_run.py
snopoke: Fixed in 9891748b9 — test deleted.

*— Claude Code*

apps/evaluations/tests/test_delta_evaluation_run.py
snopoke: Fixed in 9891748b9 — test deleted.

*— Claude Code*

apps/evaluations/auto_population.py
snopoke: Fixed in 55255ded4 — `_versions_of` was dropped entirely, replaced with `experiment=rule.source_experiment` (real chat sessions are always on the working version and the form restricts `source_experiment` to working versions).

*— Claude Code*

apps/evaluations/auto_population.py
snopoke: Fixed in 55255ded4 — `_versions_of` was dropped entirely; the queryset now filters by `experiment=rule.source_experiment` directly.

*— Claude Code*

apps/evaluations/views/auto_population_views.py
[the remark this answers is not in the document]
snopoke: Not applicable — the dedup excludes sessions whose `id` is already linked from `dataset.messages` via `session_id`. An orphaned `EvaluationMessage` (session FK nulled by `SET_NULL`) means the original session was deleted; that session no longer exists in `ExperimentSession` and can't be re-ingested. No re-ingest path produces duplicates.

*— Claude Code*

apps/evaluations/views/auto_population_views.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def delete(self, request, *args, **kwargs):
+        self.get_object().delete()
+        return HttpResponse(status=200)
```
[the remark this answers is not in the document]
snopoke: Noted — declined. The HTMX-style 200 response matches the existing project pattern (`DeleteDataset` does the same thing). Non-JS browsers would see a blank 200 page on submit, which is consistent with how delete works elsewhere in the app.

*— Claude Code*

apps/evaluations/auto_population.py
```diff
@@ -0,0 +1,196 @@
...
+    rule.last_run_at = timezone.now()
+    update_fields = ["last_run_at", "last_run_status"]
+    if appended:
+        rule.last_run_status = AutoPopulationRunStatus.SUCCESS
+        rule.consecutive_failure_count = 0
+        rule.last_error = ""
+        update_fields += ["consecutive_failure_count", "last_error"]
+    else:
+        rule.last_run_status = AutoPopulationRunStatus.NO_OP
+    rule.save(update_fields=update_fields)
```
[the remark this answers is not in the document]
snopoke: Fixed in ce5fd9185. NO_OP now also resets `consecutive_failure_count` and `last_error`, so a healthy-but-empty tick can't contribute to auto-disable.

*— Claude Code*

apps/evaluations/tables.py
```diff
@@ -291,6 +306,52 @@ class Meta:
...
+    last_run_at = columns.DateTimeColumn(verbose_name="Last run", orderable=False)
+    last_run_status = columns.Column(
+        accessor="get_last_run_status_display",
+        verbose_name="Status",
+        orderable=False,
+    )
+    last_error = columns.Column(verbose_name="Last error", orderable=False)
+
+    def render_last_error(self, value):
+        return mark_safe(f'<span class="text-error">{value[:60]}</span>') if value else "—"
```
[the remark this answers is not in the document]
snopoke: Fixed in ce5fd9185. `render_evaluators` and `render_last_error` now use `format_html` / `format_html_join`, which escapes the dynamic strings (`evaluator.name`, `rule.last_error`).

*— Claude Code*

SmittieC:
One tiny suggestion from looking at the demo: When I create a rule, I can choose to specify a filter or not. I know it says on the page that when a filter is not specified, all will match. How about making this more explicit. For instance, we could hide the filter behind a "filter sessions" checkbox which, when clicked, shows the filter and some text making it clear that the sessions will be filtered with this filterset. When the checkbox is unchecked, we make it clear to the user that all sessions will be added. Thoughts?

apps/evaluations/auto_population.py
```diff
@@ -0,0 +1,155 @@
...
+
+    if rule.filter_query_string:
+        params = FilterParams(QueryDict(rule.filter_query_string))
+        qs = ExperimentSessionFilter().apply(qs, params, timezone=None)
+
+    session_external_ids = list(qs.values_list("external_id", flat=True))
+    if not session_external_ids:
+        return []
+
+    eval_messages = make_session_evaluation_messages(session_external_ids, team=rule.team)
```
SmittieC, line 77: I suspect a race condition might exist under these circumstances, albeit I think it's an edge case:
- a dataset have multiple rules
- two workers happen to pick up these rules, so two of those rules run concurrently
- a session matches both rules
- this session will be added to the dataset twice

snopoke: This is prevented by the `DatasetAutoPopulationRule.objects.select_for_update(skip_locked=True)` in the `auto_populate_eval_datasets` task which will skip rules that are already being processed.

snopoke:
> One tiny suggestion from looking at the demo: When I create a rule, I can choose to specify a filter or not. I know it says on the page that when a filter is not specified, all will match. How about making this more explicit. For instance, we could hide the filter behind a "filter sessions" checkbox which, when clicked, shows the filter and some text making it clear that the sessions will be filtered with this filterset. When the checkbox is unchecked, we make it clear to the user that all sessions will be added. Thoughts?

I expect that filtering is the primary use case but I'll add a warning if there are no filters. (in a followup PR)

apps/evaluations/auto_population.py
SmittieC: So it is!

SmittieC (approved):
Coo!

Issue #3303, opened 2026-05-06
MessageSizeValidationMiddleware over-counts tool outputs and blocks legitimate conversations


snopoke:
Follow-up to #3180, which introduced `MessageSizeValidationMiddleware`. The middleware was temporarily disabled in #3302 to unblock production while we redesign it.

## What the middleware does today

`MessageSizeValidationMiddleware.before_model` runs `count_tokens_approximately(state["messages"])` on every model invocation and raises `MessageTooLargeError` when the count exceeds the model's `max_token_limit` (minus the system message).

## Symptom

Users sending a small message see errors like:

> Your message is too large for this model. It uses approximately 768575 tokens, but only 408695 tokens are available.

…even though the human input is a few words. Reproduced on a chatbot that calls a Custom Action tool returning a large structured response. Trace: 7f648ebf0698781c6ea2307bc9ba1655.

## What's actually happening

The validator runs on **every** iteration of the agent loop, not just the first one. Timeline from the trace:

| Iteration | What runs | OpenAI input_tokens | Approx tokens (validator) |
|---|---|---|---|
| 1 | `before_model` → ChatOpenAI | 1,367 | under limit ✅ |
| 1 | tool call returns ~3MB of text as a `ToolMessage` | — | — |
| 2 | `before_model` → ChatOpenAI | (never reached) | **768,575** ❌ raises |

So the user's message was fine — a tool's response was what blew the budget. The validator can't tell the difference.

## Two issues stacked on each other

1. **`count_tokens_approximately` includes ToolMessage content.** The intent of the validator (per the PR description on #3180) is to fail early on oversized user input. But because it counts everything in `state["messages"]`, it also fails on oversized tool output — which is a different problem with different appropriate handling.

2. **The approximation overshoots significantly on JSON-heavy content.** `count_tokens_approximately` is `chars / 4`. JSON and structured tool output tokenizes much more efficiently than that under real BPE — the OpenAI tokenizer might count 1.3–2× fewer real tokens than the approximation reports. So even when the count is "too large", the model may have actually accepted the request. The error message is also user-facing and shows the inflated number.

3. **`SummarizeHistoryMiddleware` cannot rescue this case.** It runs immediately before the validator, but its `keep=("messages", 20)` policy means `_find_safe_cutoff` returns 0 when fewer than 20 messages are in state — which is the typical situation when a single tool message is the offender. Trigger fires on tokens; keep operates on message count; they don't agree on units.

## Scope

[...]

PR #3196, opened 2026-04-21
feat: add max-char-limit support to chat widget and web chat template

ezkemboi:
### Product Description
Adds a live character counter and send-button guard to the OCS chat widget. Users see a warning as they approach the character limit and are blocked from sending once they exceed it, giving early feedback instead of a backend error.

### Technical Description
Part of #3180 (widget changes separated per changelog/docs automation requirements).

Changes are confined to the StencilJS chat widget component (`components/chat_widget/`):

- **`ocs-chat.tsx`**: Reads the new `maxCharLimit` prop (passed from the Django host page). Alpine.js-style reactive getters `messageTooLong` / `messageNearLimit` drive warning/error CSS classes on the counter and disable the send button when the limit is exceeded.
- **`ocs-chat.css`**: Adds `.char-counter`, `.char-counter--warning`, and `.char-counter--error` styles.
- **`chat-session-service.ts`**: Propagates the `maxCharLimit` value through the service layer.
- **`components.d.ts`**: Autogenerated Stencil type — adds `maxCharLimit?: number` prop declaration.
- **`templates/chatbots/chat/web_chat.html`**: Passes `max-char-limit="{{ max_char_limit }}"` attribute to `<open-chat-studio-widget>` when the backend provides a limit.
- **`assets/translations/en.json`**: Adds i18n key for the counter label.

The backend that computes `max_char_limit` and the HTMX input bar changes live in the companion PR #3180.

### Demo
See PR #3180 for screenshots of the character counter and error state in the web chat UI.

### Docs and Changelog
- [ ] This PR requires docs/changelog update

Widget change exposing a new `maxCharLimit` prop on `<open-chat-studio-widget>`. Changelog entry: the chat widget now enforces a configurable character limit with a live counter.

PR #3316, opened 2026-05-11
fix/over-counts tool outputs and blocks legitimate conversations

ezkemboi:
### Product Description

Users sending short messages were sometimes seeing errors like:

> "Your message is too large for this model."

…even though their input was only a few words. This happened when a tool called during the conversation returned a large response (e.g. a Custom Action returning structured JSON). This fix resolves those false positives and re-enables the middleware that was temporarily disabled in #3302.

The frontend character counter (added in #3180) has also been removed. It was built on the same flawed approximation as the middleware and independently produced false positives — blocking the send button for messages the model would have accepted. The server-side error message now surfaces directly in the chat when a message is genuinely too large.

### Technical Description

`MessageSizeValidationMiddleware.before_model` previously ran `count_tokens_approximately(state["messages"])` — counting **all** messages in state on every agent loop iteration, including `ToolMessage` content from tool calls.

Three problems stacked:

1. **ToolMessage inflation** — on iteration 2 (after a tool call returns), the full state including the large `ToolMessage` was counted, blowing past the token limit even though the user's input was small.
2. **System prompt double-counting** — `effective_limit` already deducts system tokens from the budget, but `state["messages"]` can include the `SystemMessage`, so those tokens were subtracted twice.
3. **chars/4 approximation overshoot** — `count_tokens_approximately` uses `chars / 4`, which overshoots JSON and structured content by 1.3–2× under real BPE. Even when the count was "too large", the model may have accepted the request.

**Fix:** filter `state["messages"]` to only the last `HumanMessage` before counting, and switch to `model.get_num_tokens_from_messages` (real BPE tokenizer via tiktoken for OpenAI models).

Key changes:
- `history_middleware.py` — `before_model` counts only `human_messages[-1:]` using `model.get_num_tokens_from_messages`; `token_limit=None` skips validation, `token_limit=0` (system prompt exhausted full budget) correctly rejects
- `llm_node.py` — middleware re-enabled (was commented out in #3302); `model = node.get_chat_model()` extracted once and passed to both the middleware and `create_agent` to avoid double instantiation; system token count also uses real tokenizer
- `test_llm_node.py` — regression tests for ToolMessage false positive, empty HumanMessage state, multi-turn history, zero budget, and None budget
- `utils.py`, `views/utils.py`, `experiment.py`, `chatbots/views.py`, `pipelines/views.py` — `compute_max_char_limit` / `get_max_char_limit` and all `max_char_limit` template context keys removed
- `input_bar.html` — `maxCharLimit`, `messageTooLong`, `:maxlength`, error class bindings removed
- `apps/pipelines/utils.py` — deleted (emptied by counter removal)
- `test_view_utils.py` — deleted (all tests were for the removed `get_max_char_limit`)

### Migrations

No migrations.

### Demo

No visible UI change on the happy path. Users who were previously blocked by false positives will now be able to send messages normally. When a message genuinely exceeds the model's token budget, a user-facing error appears inline in the chat.

### Docs and Changelog


ezkemboi:
@snopoke, I am doing more testing on this PR, later today and tomorrow, for it to be ready.
I will also let you know of the `JSON and structured tool output tokenize` because that has not been added here yet.

ezkemboi:
From the comment https://github.com/dimagi/open-chat-studio/pull/3196#discussion_r3194893212, I have also removed counter on this end.

ezkemboi:
@snopoke and @SmittieC, I’ve removed the frontend counter from this test.

Also, on the current change, we only value the last message and not with history.

The backend now uses _count_tokens, which relies on get_num_tokens_from_messages for accurate token counting, including proper handling of JSON-structured inputs.

For models that don’t support get_num_tokens_from_messages, we fall back to count_tokens_approximately (langchain-based estimation).

I have attached a screenshot of the test done. 
Screenshot 2026-05-17 at 08 27 41

cc. @KipmurkorDev, please help retest the changes updated.

ezkemboi:
@SmittieC, if you get some time today, help review this PR for me. 

cc. @snopoke

SmittieC:
> @SmittieC, if you get some time today, help review this PR for me.
> 
> cc. @snopoke

Sure I'll take a look today

apps/pipelines/nodes/mixins.py
```diff
@@ -242,7 +241,9 @@ def build_history_middleware(self, system_message: BaseMessage) -> BaseNodeHisto
         )
 
         # Reserve space for the system message so trigger/keep thresholds reflect usable context
-        system_message_tokens = count_tokens_approximately([system_message])
+        from apps.pipelines.nodes.history_middleware import _count_tokens  # noqa: PLC0415
```
SmittieC: can this import be moved to the top?

apps/pipelines/nodes/llm_node.py
```diff
@@ -100,30 +99,33 @@ def build_node_agent(
...
 
-def _build_size_validation_middleware(node: PipelineNode, system_message) -> MessageSizeValidationMiddleware | None:
+def _build_size_validation_middleware(
+    node: PipelineNode, system_message, model
+) -> MessageSizeValidationMiddleware | None:
     max_token_limit = node.repo.get_llm_provider_model(node.llm_provider_model_id).max_token_limit
     if not max_token_limit:
         return None
-    system_tokens = count_tokens_approximately([system_message])
+    system_tokens = _count_tokens(model, [system_message])
```
SmittieC: Rather make `_count_tokens` non-private if we're going to be importing it from another module

apps/pipelines/nodes/mixins.py
```diff
@@ -242,7 +241,9 @@ def build_history_middleware(self, system_message: BaseMessage) -> BaseNodeHisto
...
+
+        system_message_tokens = _count_tokens(model, [system_message]) if model is not None else 0
```
SmittieC: Why assume 0 tokens if the model is missing? `_count_tokens` will fall back to approximating it if the model is missing, which seems like something we'd want?

apps/pipelines/nodes/history_middleware.py
```diff
@@ -14,6 +14,20 @@
 from apps.chat.conversation import COMPRESSION_MARKER
 from apps.pipelines.exceptions import MessageTooLargeError
 
+
+def _count_tokens(model, messages: list) -> int:
```
SmittieC: Suggestion:

You're passing the model from the node level down to here just to call `.get_num_tokens_from_messages` on it. What if you instead make a method that accepts an optional `model` param and return the token counter to use. Something like this:

```python
def get_token_counter(model=None):
    return model.get_num_tokens_from_messages if model else count_tokens_approximately
```

That way, the history middleware (and _build_size_validation_middleware) doesn't have to get a model from its callers, but just the token counter to use.

Regarding the try-catch behaviour: What scenarios might there be where `model.get_num_tokens_from_messages` raises and `count_tokens_approximately` doesn't? Maybe the caller should be responsible for doing any try-catching?

apps/pipelines/nodes/history_middleware.py
```diff
@@ -134,22 +148,28 @@ def persist_summary(self, messages: list[BaseMessage]):
...
+        if not human_messages:
             return None
-        token_count = count_tokens_approximately(state["messages"])
+        token_count = _count_tokens(self._model, [human_messages[-1]])
         if token_count > self._token_limit:
             raise MessageTooLargeError(
                 f"Your message is too large for this model. "
-                f"It uses approximately {token_count} tokens, but only {self._token_limit} tokens are available."
+                f"It uses approximately {token_count} tokens, but only {self._token_limit} tokens are available "
+                f"after accounting for the system prompt."
```
SmittieC: > after accounting for the system prompt

Nit: Just double checking, is there a specific reason why we need to include this?

apps/pipelines/nodes/history_middleware.py
```diff
@@ -134,22 +148,28 @@ def persist_summary(self, messages: list[BaseMessage]):
...
 
     def before_model(self, state, runtime):
-        if not self._token_limit:
+        if self._token_limit is None:
+            return None
+        human_messages = [m for m in state["messages"] if isinstance(m, HumanMessage)]
+        if not human_messages:
             return None
-        token_count = count_tokens_approximately(state["messages"])
+        token_count = _count_tokens(self._model, [human_messages[-1]])
```
SmittieC: Do you know if the history middleware will remove all history when the human message is long enough that it effectively takes up all of the token space that is left over after subtracting the system message's tokens?

I have a feeling that a large human message might still fail at the API level, even though it passes here, unless the history middleware is smart enough to throw everything away except the last human and system message. Do you perhaps know if this is the case?

ezkemboi:
@SmittieC, we are checking on the feedback and retesting based on the knowledge shared with @KipmurkorDev

apps/pipelines/nodes/history_middleware.py
ezkemboi: Previously, the validator only checked the last human message against the token limit. The compressors (summarize/truncate) handled reducing history, but if they couldn't e.g. SUMMARIZE with fewer than 20 messages, or MAX_HISTORY_LENGTH which is count-based, a large message + existing history could slip through and fail silently at the API. That's exactly the edge case you caught.

The changes add a second check on the total conversation history (HumanMessages + AIMessages) that runs after the compressor. If the compressor couldn't reduce context enough, this raises a clear error before the API is called. ToolMessages are excluded because they sit between agent-loop iterations and aren't user-controlled, that was actually the root cause of #3303 where the previous total-context check was firing incorrectly.

apps/pipelines/nodes/history_middleware.py
```diff
@@ -134,22 +143,42 @@ def persist_summary(self, messages: list[BaseMessage]):
...
+
+        # Validate the current user message individually.
+        token_count = self._token_counter([human_messages[-1]])
         if token_count > self._token_limit:
             raise MessageTooLargeError(
                 f"Your message is too large for this model. "
                 f"It uses approximately {token_count} tokens, but only {self._token_limit} tokens are available."
             )
+
+        # Validate total history, excluding tool responses and the system prompt.
```
ezkemboi, line 175: @SmittieC you can check on this approach and let me know what are your thoughts?

apps/pipelines/nodes/history_middleware.py
```diff
@@ -155,27 +143,42 @@ def persist_summary(self, messages: list[BaseMessage]):
...
-    """Raises MessageTooLargeError if the current user message exceeds the token budget.
+    """Raises MessageTooLargeError if the current user message or accumulated history
+    exceeds the model's token budget.
 
-    Only validates the last HumanMessage in state. ToolMessages and prior conversation
-    history are intentionally excluded — large tool responses are a separate concern,
-    and history size is managed by the compression middleware that runs before this one.
+    Two checks run on each call: the last HumanMessage is validated individually, and
+    the total conversation history (HumanMessages + AIMessages) is validated together.
+    ToolMessages and SystemMessages are excluded from the total — tool responses are
```
SmittieC, line 151: I don't quite follow why tool responses are excluded. They are also sent to the LLM along with the history + user's new message, so it should be included, no?

ezkemboi: I will include that as well in this PR.

Previously, I thought we were only validating the latest user message against the configured `max_token_limit`. Since some models support hundreds of thousands of tokens, whereas in our case we may configure a lower limit, such as `8192`, I assumed the validation was only applied to the latest message and not to the entire payload.

From your explanation, I now understand that the validation should apply to the full request sent to the LLM, including the conversation history and tool responses, since all of that contributes to actual token usage.

apps/pipelines/nodes/llm_node.py
```diff
@@ -102,11 +101,7 @@ def build_node_agent(
     system_message = get_system_message(prompt_template=node.prompt, prompt_context=prompt_context)
 
     model = node.get_chat_model()
-    middleware = []
-    if history_middleware := node.build_history_middleware(system_message=system_message, model=model):
-        middleware.append(history_middleware)
-    if size_middleware := _build_size_validation_middleware(node, system_message, model):
-        middleware.append(size_middleware)
+    middleware = node.build_history_middleware(system_message=system_message, model=model)
```
SmittieC, line 104: Suggestion: Keep the history middleware as it was before and always include `MessageSizeValidationMiddleware` in the middleware, since there will always be a max token limit.

ezkemboi: Well noted. I will do so.

snopoke (commented):
@ezkemboi @SmittieC I'm wondering about the utility of this feature. The issue it is trying to resolve is quite rare and requires the user to upload a large document and paste a very large amount of text.

Given the complexity that validating this is turning out to be I'd like to propose we drop it entirely.

ezkemboi:
> @ezkemboi @SmittieC I'm wondering about the utility of this feature. The issue it is trying to resolve is quite rare and requires the user to upload a large document and paste a very large amount of text.
> 
> Given the complexity that validating this is turning out to be I'd like to propose we drop it entirely.

@snopoke, since @SmittieC was the creator of the same issue, his feedback carries more weight on the proposal above.

SmittieC:
> @ezkemboi @SmittieC I'm wondering about the utility of this feature. The issue it is trying to resolve is quite rare and requires the user to upload a large document and paste a very large amount of text.
> 
> Given the complexity that validating this is turning out to be I'd like to propose we drop it entirely.

I was starting to think along the same lines. The only benefit we get by checking and failing early is user feedback. If we still want the user feedback - which I do think is nice, because it tells users what they can do about it - we might as well let the API call fail, parse the failure and reraise a custom user friendly error.

ezkemboi:
> > @ezkemboi @SmittieC I'm wondering about the utility of this feature. The issue it is trying to resolve is quite rare and requires the user to upload a large document and paste a very large amount of text.
> > Given the complexity that validating this is turning out to be I'd like to propose we drop it entirely.
> 
> I was starting to think along the same lines. The only benefit we get by checking and failing early is user feedback. If we still want the user feedback - which I do think is nice, because it tells users what they can do about it - we might as well let the API call fail, parse the failure and reraise a custom user friendly error.

I think what you have mentioned above makes sense. 
Let's hear what @snopoke has to say regarding the same.

snopoke:
+1 to parsing the error - Here is the error you get from OpenAI:

```
 {'error': {'message': 'Your input exceeds the context window of this model. Please adjust your input and try again.', 'type': 'invalid_request_error', 'param': 'input', 'code': 'context_length_exceeded'}}
```

I'm not sure what you get from the other providers.

ezkemboi:
> +1 to parsing the error - Here is the error you get from OpenAI:
> 
> ```
>  {'error': {'message': 'Your input exceeds the context window of this model. Please adjust your input and try again.', 'type': 'invalid_request_error', 'param': 'input', 'code': 'context_length_exceeded'}}
> ```
> 
> I'm not sure what you get from the other providers.

I will take a look at this PR this weekend so that we can get it out of line.

ezkemboi:
> +1 to parsing the error - Here is the error you get from OpenAI:
> 
> ```
>  {'error': {'message': 'Your input exceeds the context window of this model. Please adjust your input and try again.', 'type': 'invalid_request_error', 'param': 'input', 'code': 'context_length_exceeded'}}
> ```
> 
> I'm not sure what you get from the other providers.

for copilot model of Gpt is also the same as mentioned

Reason: Request Failed: 400 {"error":{"message":"This model's maximum context length is 128000 tokens. However, your messages resulted in 132438 tokens (108594 in the messages, 23844 in the functions). Please reduce the length of the messages or functions.","code":"invalid_request_body"}}

ezkemboi:
@snopoke noted. 
Will check on another issue and work on it.

PR #3323, opened 2026-05-11
Consolidate content-type detection into a single utility

SmittieC:
### Product Description
Internal refactor, no user-visible behaviour change.

### Technical Description
Three places were running their own `magic.from_buffer` + fallback cascade (`File.create`, `File.get_content_type`, `email_channel._detect_content_type`). This PR consolidates them into a new `apps/files/content_type.py` utility with two helpers: `detect_content_type(content, *, filename="", fallback="")` (bytes-based, returns `str | None`) and `detect_content_type_from_file(file_obj)` (file-object, always returns a string). All three call sites now delegate to it.

Follow-up to ##3256.

### Migrations

No schema changes.

### Docs and Changelog

### DIFF
--- a/apps/files/models.py
+++ b/apps/files/models.py
@@ -9,7 +9,7 @@ from django.urls import reverse
 from pgvector.django import HalfVectorField
 
 from apps.experiments.versioning import VersionDetails, VersionField, VersionsMixin, VersionsObjectManagerMixin
-from apps.files.content_type import detect_content_type, detect_content_type_from_file
+from apps.files.content_type import DEFAULT_CONTENT_TYPE, detect_content_type, detect_content_type_from_file
 from apps.generics.chips import Chip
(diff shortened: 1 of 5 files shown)


apps/files/models.py
```diff
@@ -89,13 +88,8 @@ def create(
...
 
-        if not content_type and content:
-            with contextlib.suppress(Exception):
-                detected = magic.from_buffer(content[:2048], mime=True)
-                if detected and detected != "application/octet-stream":
-                    content_type = detected
-
-        content_type = content_type or mimetypes.guess_type(filename)[0]
+        if not content_type:
+            content_type = detect_content_type(content or b"", filename=filename) or DEFAULT_CONTENT_TYPE
```
SmittieC: @claude perhaps detect_content_type should return DEFAULT_CONTENT_TYPE instead of potentially None

snopoke (approved):
nice cleanup

PR #3339, opened 2026-05-13
Add `inv dev` honcho launcher; default celery to no-beat

snopoke:
### Product Description
No change. Developer tooling only.

### Technical Description
Bundles the three long-running dev processes (Django, Celery via watchfiles, webpack dev-watch) behind a single command so you no longer need three terminals.

- New `Procfile.dev` defines `web`, `worker`, `assets` — each delegates back to existing invoke tasks (`uv run inv runserver`, `uv run inv celery`) so portless detection and the celery command stay in one place.
- New `inv dev` task shells out to `uv run honcho -f Procfile.dev start`. Honcho added to the `dev` dependency group.
- Dropped the `dev` alias from `inv runserver` so the new task can take that name; `inv runserver` and `inv django` still work.
- `inv celery` now defaults `beat=False` (the previous default included `-B`). Run `inv celery --beat` if you need the beat scheduler. The Procfile-driven worker therefore runs without beat by default, matching how most contributors actually run things.

### Demo
N/A — dev-only change. To try: `uv sync --dev && uv run inv dev`.

### Docs and Changelog

Dev-tooling change only; no user-facing changelog entry needed.

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,0 +1,3 @@
+web: uv run inv runserver
+worker: uv run inv celery
+assets: npm run dev-watch
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -173,6 +173,7 @@ dev = [
     "prek>=0.3.2",
     "django-types",
     "ty==0.0.33",
+    "honcho>=2.0.0",
 ]
 docs = [
     "zensical>=0.0.40",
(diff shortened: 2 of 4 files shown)


SmittieC (approved):
Nice

PR #3351, opened 2026-05-14
refactor: centralize ChatMessage.metadata keys in a StrEnum

SmittieC:
### Product Description
No user-facing changes. Internal refactor to make `ChatMessage.metadata` keys discoverable and typo-resistant.

### Technical Description
Introduces a `ChatMessage.MetadataKeys` `StrEnum` (mirroring the existing `Chat.MetadataKeys` pattern) that centralizes the 9 string keys previously sprinkled as literals across the codebase:

- **OpenAI Assistant (internal):** `openai_run_id`, `openai_file_ids`, `openai_thread_checkpoint`
- **Files & attachments:** `ocs_attachment_file_ids`, `cited_files`, `generated_files`
- **Tracing / observability:** `trace_info`, `trace_provider` (legacy top-level)
- **History / compression:** `compression_marker`

`INTERNAL_METADATA_KEYS` is rebuilt as a `frozenset` of enum members so the API-filter list and the source of truth can't drift.

All production call sites in `apps/chat/`, `apps/channels/`, `apps/pipelines/`, and `apps/service_providers/` have been migrated off raw literals. Tests left untouched — they still pass because `StrEnum` values compare and hash identically to the underlying string.

Out of scope (intentionally): `EvaluationMessage.metadata` keys, the `trace_provider` sub-field inside each `trace_info[]` item, and `Trace.trace_metadata` reads in `apps/trace/` / `apps/experiments/export.py` (different model).

### Migrations

No DB migration is required. `StrEnum` members stringify byte-identically to the existing literal keys, so existing rows are unchanged and all reads continue to match.

### Demo
N/A — internal refactor.

### Docs and Changelog

### DIFF
--- a/apps/chat/models.py
+++ b/apps/chat/models.py
@@ -121,14 +121,29 @@ class ChatMessage(BaseModel, TaggedModelMixin, UserCommentsMixin):
     A message in a chat. Analogous to the BaseMessage class in langchain.
     """
 
+    class MetadataKeys(StrEnum):
+        # OpenAI Assistant (internal)
+        OPENAI_RUN_ID = "openai_run_id"
+        OPENAI_FILE_IDS = "openai_file_ids"
+        OPENAI_THREAD_CHECKPOINT = "openai_thread_checkpoint"
+        # Files & attachments
+        OCS_ATTACHMENT_FILE_IDS = "ocs_attachment_file_ids"
+        CITED_FILES = "cited_files"
+        GENERATED_FILES = "generated_files"
+        # Tracing / observability
+        TRACE_INFO = "trace_info"
+        TRACE_PROVIDER = "trace_provider"  # legacy top-level; only read for migration in trace_info property
+        # History / compression
+        COMPRESSION_MARKER = "compression_marker"
+
     # Metadata keys that should be excluded from the API response
-    INTERNAL_METADATA_KEYS = {
-        # ID of the thread run
-        "openai_run_id",
-        "openai_file_ids",
-        # boolean indicating that this message has been synced to the thread
-        "openai_thread_checkpoint",
-    }
+    INTERNAL_METADATA_KEYS = frozenset(
+        {
+            MetadataKeys.OPENAI_RUN_ID,
+            MetadataKeys.OPENAI_FILE_IDS,
+            MetadataKeys.OPENAI_THREAD_CHECKPOINT,
+        }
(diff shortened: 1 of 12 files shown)


apps/chat/models.py
```diff
@@ -121,14 +121,29 @@ class ChatMessage(BaseModel, TaggedModelMixin, UserCommentsMixin):
     A message in a chat. Analogous to the BaseMessage class in langchain.
     """
 
+    class MetadataKeys(StrEnum):
```
SmittieC: `MetadataKeys` should be outside of the ChatMessage class

SmittieC: Fixed in 509f34cc5. Promoted the enum to module-level `ChatMessageMetadataKeys` and updated all callers.

apps/chat/models.py
```diff
@@ -116,19 +116,35 @@ def role(self):
...
 
     # Metadata keys that should be excluded from the API response
-    INTERNAL_METADATA_KEYS = {
-        # ID of the thread run
-        "openai_run_id",
-        "openai_file_ids",
-        # boolean indicating that this message has been synced to the thread
-        "openai_thread_checkpoint",
-    }
+    INTERNAL_METADATA_KEYS = frozenset(
```
SmittieC: @claude rather call ChatMessageMetadataKeys.internal_keys instead of assigning the set to this constant

apps/chat/models.py
```diff
@@ -265,10 +281,15 @@ def get_attached_files(self):
...
 
-        metadata_key = ["openai_file_ids", "ocs_attachment_file_ids", "cited_files", "generated_files"]
-        for source in metadata_key:
+        metadata_keys = [
+            ChatMessageMetadataKeys.OPENAI_FILE_IDS,
+            ChatMessageMetadataKeys.OCS_ATTACHMENT_FILE_IDS,
+            ChatMessageMetadataKeys.CITED_FILES,
+            ChatMessageMetadataKeys.GENERATED_FILES,
+        ]
+        for source in metadata_keys:
```
SmittieC: @claude it would be nice if we can do ChatMessageMetadataKeys.attachment_keys here

SmittieC:
@coderabbitai full review

snopoke (commented):
nice

snopoke (approved):

PR #3352, opened 2026-05-15
Remove logger. Only unhandled exceptions should surface

SmittieC:
### Product Description

N/A

### Technical Description

No need to report the exception at the place where it happens, since it is being reported in the `SendingErrorHandlerStage` only when it is not being handled first, which is what we want.

### Demo

N/A

### Docs and Changelog


snopoke (commented):
👍🏻

PR #3355, opened 2026-05-15
Allow reviewers to edit their own submitted annotations

### LINKED ISSUE #3292: Ability to edit an annotation (human annotation queue)
After submitting an annotation, it's not currently possible to edit it. This is helpful not just in case you make a mistake, but because you may want to get alignment with fellow humans and then go back and edit it for the final determination.

SmittieC:
Closes #3292

### Product Description
Reviewers can now edit annotations they previously submitted on a queue item. An "Edit" button appears next to each of the current user's submitted annotations on the item page, opening the annotation form pre-filled with their existing data. Saving updates the annotation and recomputes the queue's aggregate stats.

### Technical Description
- Adds `EditAnnotation` view (GET/POST) at `queue/<pk>/item/<item_pk>/annotation/<annotation_pk>/edit/`, gated by `human_annotations.add_annotation` permission and a `reviewer_id == request.user.id` check (raises `PermissionDenied` otherwise — no silent 404).
- Annotation lookup is a single query: `Annotation.objects.select_related("item__queue")` with `item_id`, `item__queue_id`, and `item__queue__team` pushed into the WHERE clause, so team/queue/item scoping is enforced in SQL rather than via three sequential `get_object_or_404` round-trips.
- The annotate template gains an `is_edit` branch — same form, swapped action URL, "Save"/"Cancel" buttons instead of "Submit"/"Skip"/"Flag", and an "Edit Annotation" heading. The annotations list on the item page now renders a per-row Edit link when `ann.reviewer_id == request.user.id`.
- The aggregate-recompute path on `Annotation` was extracted from `_update_item_review_count` into a reusable `recompute_queue_aggregates(queue=None)` method so the edit POST can trigger it without going through the create-time review-count update.

### Demo
https://www.loom.com/share/f6afbb9f45bd47e4b3ce1902336de60a

### Docs and Changelog


SmittieC:
@coderabbitai full review

apps/human_annotations/views/annotate_views.py
```diff
@@ -148,9 +149,11 @@ def get(self, request, team_slug: str, pk: int, item_pk: int):
             schema_fields = list(queue.schema.keys())
             annotations = [
                 {
+                    "id": ann.id,
```
snopoke: nit: This would be better with a more explicit name e.g. `annotation_id`

snopoke (dismissed):

PR #3359, opened 2026-05-15
Add llm-stats.com attribution to LLM provider form

SmittieC:
### Product Description

Attribution to llm-stats.com. They asked that we give attribution if we use the API.

### Demo

image

### Docs and Changelog


SmittieC:
We already mention them in the workflow, so I'm not sure if we also want to do it on the UI

PR #3360, opened 2026-05-15
fix: add UTF-8 BOM to CSV transcript exports

### LINKED ISSUE #697: Character Encoding Issue in Exported CSV Transcripts
Issue:

When exporting transcripts using 'Download All' in the OCS web app, special characters are misencoded in the resulting CSV. For example, "I'm" appears as "Iâ€™m."

Steps to Reproduce:

1. Export consolidated transcripts for an experiment by clicking, 'Download All' in OCS.

barry47products:
### Product Description

CSV transcript exports now open correctly in Excel on both Windows and macOS. Previously, special characters such as curly apostrophes (`'`) and accented letters (`é`, `ü`, etc.) were garbled — for example `I'm` appeared as `Iâ€™m`.

### Technical Description

Excel does not respect the `Content-Type: text/csv` charset header when opening files directly. Without a UTF-8 Byte Order Mark (BOM) at the start of the file, Excel defaults to the system locale encoding (Windows-1252 on most systems), causing multi-byte UTF-8 sequences to be misread.

The fix yields `﻿` (UTF-8 BOM) as the first chunk from `export_rows_to_csv_stream` in `apps/experiments/export.py`. This is a one-line change to the streaming CSV generator used by `StreamingHttpResponse` exports.

### Migrations

N/A

### Demo

Before: downloading a CSV export containing `I'm` or `café` and opening in Excel produces `Iâ€™m` / `cafÃ©`.

After: characters render correctly.

### Docs and Changelog

Fixes #697


apps/experiments/export.py
```diff
@@ -213,6 +213,7 @@ def export_rows_to_csv_stream(rows: Iterator[list]) -> Generator[str]:
     Django's StreamingHttpResponse so the response is sent to the client
     incrementally rather than buffered entirely in memory.
     """
+    yield "﻿"
```
snopoke: It's impossible to tell if this is just an empty string or the correct UTF-8 character. We could write this as `"\ufeff"` which makes it clear:
```suggestion
    yield "\ufeff"  # UTF-8 BOM
```

(another option is to extract this to a constant for additional clarity)

snopoke (commented):
Thanks for the contribution @barry47products.

apps/experiments/export.py
barry47products: Good catch, thanks. Agreed, it's much clearer. Picked this one up as a good first issue while I'm learning the contribution process, so feedback is welcome.

apps/experiments/tests/test_export.py
```diff
@@ -358,3 +358,14 @@ def test_session_state_export():
 
     assert json.loads(rows[1][state_index]) == {"key": "value"}
     assert json.loads(rows[2][state_index]) == {"key": "value"}
+
+
+def test_export_rows_to_csv_stream_starts_with_utf8_bom():
+    chunks = list(export_rows_to_csv_stream(iter([])))
+    assert chunks[0] == "﻿", "First chunk must be a UTF-8 BOM so Excel reads the file correctly"
```
snopoke: This should also use the UTF-8 hex char for clarity. Since it used in both places it is probably worth adding a constant to `export.py`

barry47products:
@snopoke I wanted to check in with you on the CodeScene findings in https://github.com/dimagi/open-chat-studio/pull/3360/checks?check_run_id=76382291625

Is this something you typically address with a refactor or suppress?

snopoke:
> @snopoke I wanted to check in with you on the CodeScene findings in https://github.com/dimagi/open-chat-studio/pull/3360/checks?check_run_id=76382291625
> 
> Is this something you typically address with a refactor or suppress?

We are experimenting with CodeScene - you can threat these as quality signal but not compulsory to address, particularly in test code.

apps/experiments/export.py
barry47products: Extracted to a named constant `UTF8_BOM` in `apps/experiments/export.py` and reused at both call sites — pushed in 09e02e5ed.

apps/experiments/tests/test_export.py
barry47products: Done — the test now imports and asserts against `UTF8_BOM` from `apps.experiments.export` instead of the literal character. Pushed in 09e02e5ed.

snopoke (approved):

PR #3363, opened 2026-05-18
Remove dead TokenCountingCallbackHandler

snopoke:
### Product Description
no change

### Technical Description
`TokenCountingCallbackHandler` was attached to `AssistantChat.invoke` via a callback config merge, but the counts it accumulated were never read — `AssistantChat.invoke` always returned `ChainOutput(prompt_tokens=0, completion_tokens=0)`, and no caller introspected the handler instance. Token usage that actually gets recorded for traces flows through `OCSCallbackHandler` and the `n_prompt_tokens` / `n_completion_tokens` fields on `Trace`, not this handler.

Split into two commits:

1. **Stop attaching the handler in `AssistantChat`** — removes `BaseAdapter.callback_handler` and the `merge_configs` callback wiring so the handler is no longer constructed at runtime.
2. **Delete the now-dead code** — removes `TokenCountingCallbackHandler`, the four `TokenCounter` subclasses (`OpenAITokenCounter`, `AnthropicTokenCounter`, `GeminiTokenCounter`, `GoogleVertexAITokenCounter`), every `get_callback_handler` override on `LlmService` subclasses, the `FakeTokenCounter` test helper, the `token_counter` field on the fake services, and the `token_counts` parameter on the fake-service factories. Updates the 9 test files that were passing `token_counts=` to those factories.

`ChainOutput.prompt_tokens` / `completion_tokens` fields are left in place — they're still constructed (always as 0), so removing them would change a `Serializable`'s schema. Worth a follow-up but kept out of scope here.

`scripts/ocs_cost_estimator.py` has its own unrelated local `AnthropicTokenCounter` class — untouched.

### Docs and Changelog


SmittieC (approved):
Sweet

PR #3367, opened 2026-05-19
Authoritative annotation flag for multi-reviewer queues

### LINKED ISSUE #3342: Human annotation: add 'authoritative' flag to resolve multi-reviewer conflicts
## Background

When multiple reviewers annotate the same session their tags may differ. Currently there is no mechanism to:

1. Identify which sessions have conflicting annotations
2. Designate one annotation as the authoritative / gold-label answer
3. Drive concordance metrics and annotation queue metrics from that gold label

The agreed approach is **not** to create a third "gold label" annotation, but to mark one of the existing annotations as authoritative. Reviewers would each submit their own annotation independently, then sit together to align, and the correct annotation would be edited — with one of them flagged as authoritative.

## Proposed change

Add an `is_authoritative` boolean flag to the annotation/review model (default `False`).

### Rules

- At most one annotation per session can be `is_authoritative = True`
- Setting a different annotation as authoritative on the same session should clear the flag on the previous one
- Reviewers (or admins) can toggle the flag from the annotation UI

### Metric behaviour

- **Concordance metrics** should be calculated using only authoritative annotations where one exists; fall back to all annotations for sessions with no authoritative annotation yet
- **Annotation queue metrics** (completion, coverage, etc.) should count a session as "resolved" only when it has an authoritative annotation (for sessions with >1 reviewer)

## Workaround (until this is implemented)

For projects with multiple reviewers, the interim process is:
1. Each reviewer independently annotates every session
2. Reviewers meet and agree on the correct tags
3. **Both** reviewers edit their annotations to match — ensuring no conflicts remain in the current metrics

## Out of scope

- A dedicated "gold label" third annotation type (explicitly rejected)
- Automatic conflict detection / flagging (could be a follow-up)

snopoke:
### Product Description
Lets queue admins resolve multi-reviewer annotation conflicts by marking one annotation as the authoritative answer. Adds a new "Awaiting resolution" state for multi-reviewer items whose reviews are all in but no authoritative pick has been made yet, surfaces resolved / awaiting counts on the queue detail, and includes an `is_authoritative` column in CSV/JSONL exports.

Closes #3342.

### Technical Description
- Adds `is_authoritative` + `authoritative_set_by` FK + `authoritative_set_at` on `Annotation`. Partial unique constraint enforces at-most-one-per-item at the DB layer.
- New `AWAITING_RESOLUTION` item status for multi-reviewer items with all reviews in but no authoritative pick.
- `SetAuthoritative` view (POST, admin-only) wraps the toggle in `transaction.atomic()` + `select_for_update()` and clears siblings before setting.
- `Annotation.save` auto-marks the first submission on single-reviewer queues with `authoritative_set_by=None` so the reverse migration can distinguish auto-marked rows. Item is locked across the auto-mark check + save so two concurrent submissions can't both pass the check and violate the unique constraint.
- Aggregation prefers authoritative annotation(s) per item, falls back to all submitted when none exist.
- HTMX partial swaps the awaiting-resolution banner and annotation list together via a single `#annotate-summary` target to avoid stale banner state.
- Data migration `0004` backfills authoritatives on existing single-reviewer items (earliest submission wins for over-budget cases) and downgrades stale `COMPLETED` multi-reviewer items; reversible.

Spec: `docs/superpowers/specs/2026-05-18-authoritative-annotation-design.md`
Plan: `docs/superpowers/plans/2026-05-18-authoritative-annotation.md`

### Migrations

`0003` is purely additive — default `False` boolean, nullable FK / datetime, partial unique index. Non-blocking on a hot table.

`0004` is a data migration with a working `backwards`: clears auto-marked authoritatives (those with `authoritative_set_by IS NULL`) and reverts `AWAITING_RESOLUTION` items back to `COMPLETED`.

### Demo
<div>
    <a href="https://www.loom.com/share/b76352e3bdda41029975dc84b57dd021">
      <p>How Authoritative Annotations Work in Open Chat Studio - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/b76352e3bdda41029975dc84b57dd021">
      
    </a>
  </div>

### Docs and Changelog

Adds a new "authoritative" workflow for multi-reviewer annotation queues; introduces an `Awaiting resolution` item status; aggregation and queue completion metrics now use authoritative-or-fallback semantics; exports include an `is_authoritative` column.


apps/human_annotations/models.py
```diff
@@ -215,20 +228,51 @@ class Annotation(BaseTeamModel):
...
     )
+    is_authoritative = models.BooleanField(default=False)
+    authoritative_set_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        related_name="authoritative_annotations_set",
+    )
+    authoritative_set_at = models.DateTimeField(null=True, blank=True)
```
snopoke: these should be at the top of the class with the other model fields

apps/human_annotations/models.py
```diff
@@ -215,20 +228,51 @@ class Annotation(BaseTeamModel):
...
 
     class Meta:
```
snopoke, line 241: Let's move this up to just below the model fields

apps/human_annotations/views/annotate_views.py
```diff
@@ -345,3 +355,55 @@ def post(self, request, team_slug: str, pk: int, item_pk: int):
...
+    permission_required = "human_annotations.change_annotationqueue"
+
+    def post(self, request, team_slug: str, pk: int, item_pk: int, annotation_pk: int):
+        queue = get_object_or_404(AnnotationQueue, id=pk, team=request.team)
+        annotation = get_object_or_404(
+            Annotation.objects.select_related("item__queue"),
+            id=annotation_pk,
+            item_id=item_pk,
+            item__queue=queue,
+        )
```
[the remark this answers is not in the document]
snopoke: Fixed in f98d56181. `SetAuthoritative.post` now adds `status=AnnotationStatus.SUBMITTED` to the `get_object_or_404` lookup, so DRAFT annotations 404 instead of being toggled. Added `test_set_authoritative_rejects_draft_annotation` to lock the behaviour in.

*— Claude Code*

apps/human_annotations/models.py
snopoke: Fixed in f98d56181. Reordered `is_authoritative`, `authoritative_set_by`, and `authoritative_set_at` to sit immediately after `reviewer`, before `data` and `status`.

*— Claude Code*

apps/human_annotations/models.py
snopoke: Fixed in f98d56181 — grouped together with the field reorder above.

*— Claude Code*

apps/human_annotations/migrations/0004_backfill_authoritative.py
snopoke, line 1: I think this should be fine to run directly in a Django migration since there isn't very much annotation data collected yet.

apps/human_annotations/models.py
```diff
@@ -219,16 +241,38 @@ class Annotation(BaseTeamModel):
...
         if is_new and self.status == AnnotationStatus.SUBMITTED:
             self._update_item_review_count()
 
+    def _maybe_auto_mark_authoritative(self):
+        """For single-reviewer queues, auto-mark the first submission as authoritative.
+        Skips when another authoritative annotation already exists on the item (handles
+        over-budget submissions and avoids violating the partial unique constraint)."""
+        queue = self.item.queue
+        if queue.num_reviews_required != 1:
+            return
```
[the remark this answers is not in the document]
snopoke: Fixed in e7ca481e9. Moved the existing `select_for_update()` on `AnnotationItem` to span the auto-mark check, the annotation save, and the review-count update — so two concurrent submissions on the same item now serialize on the row lock instead of both passing the pre-check. The narrow inner transaction in `_update_item_review_count` is gone; everything happens inside one atomic block in `save()`.

*— Claude Code*

apps/human_annotations/views/annotate_views.py
```diff
@@ -345,3 +355,56 @@ def post(self, request, team_slug: str, pk: int, item_pk: int):
...
+    """Queue-admin endpoint to mark/unmark an annotation as authoritative.
+
+    Enforces at-most-one-per-item at the application layer too (clears the flag
+    on any sibling annotation before setting) so we never trip the partial unique constraint.
+    """
+
+    permission_required = "human_annotations.change_annotationqueue"
+
+    def post(self, request, team_slug: str, pk: int, item_pk: int, annotation_pk: int):
+        queue = get_object_or_404(AnnotationQueue, id=pk, team=request.team)
```
SmittieC: how about merging this query with the one below so we don't have to do 2

apps/human_annotations/views/annotate_views.py
```diff
@@ -345,3 +355,56 @@ def post(self, request, team_slug: str, pk: int, item_pk: int):
...
+                annotation.authoritative_set_by = None
+                annotation.authoritative_set_at = None
+            annotation.save(
+                update_fields=["is_authoritative", "authoritative_set_by", "authoritative_set_at", "updated_at"]
+            )
+            item.update_status()
+
+        annotation.recompute_queue_aggregates(queue)
+
+        if request.headers.get("HX-Request"):
```
SmittieC: Is there a case where this endpoint is not called by HTMX?

apps/human_annotations/aggregation.py
```diff
@@ -28,7 +29,10 @@ def compute_aggregates_for_queue(queue) -> AnnotationQueueAggregate:
     ).all()
 
     for item in items:
-        for ann in item.annotations.all():
+        submitted = list(item.annotations.all())
+        authoritative = [a for a in submitted if a.is_authoritative]
+        contributing = authoritative if authoritative else submitted
```
SmittieC: Just to make sure I understand: so once the admin selects the first annotation as being authoritative, the metrics will be computed and only that annotation will be used until they add more?

templates/human_annotations/columns/annotations_summary.html
```diff
@@ -2,7 +2,7 @@
   {% if annotations %}
     {% for ann in annotations|slice:":3" %}
       <div class="text-xs">
-        <span class="font-medium">{{ ann.reviewer.get_full_name|default:ann.reviewer.username }}</span>:
+        {% if ann.is_authoritative %}<i class="fa-solid fa-star text-primary" title="Authoritative"></i> {% endif %}<span class="font-medium">{{ ann.reviewer.get_full_name|default:ann.reviewer.username }}</span>:
```
SmittieC: I don't have anything against the star, but have you considered using [fa-bagde-check](https://fontawesome.com/icons/classic/solid/badge-check) or [fa-gavel](https://fontawesome.com/icons/classic/solid/gavel)?

apps/human_annotations/tests/test_views.py
```diff
@@ -1532,3 +1533,84 @@ def test_import_from_dataset_post_all_duplicates_creates_no_new_items(client, te
...
+    response = client.get(url)
+
+    assert response.status_code == 200
+    assert b"1 awaiting resolution" in response.content
+    # And confirm the renamed label is present too.
+    assert b"items resolved" in response.content
+
+
+@pytest.mark.django_db()
+def test_items_table_marks_authoritative_with_star(client, team_with_users, user):
```
SmittieC: Low value test?

apps/human_annotations/migrations/0004_backfill_authoritative.py
```diff
@@ -0,0 +1,53 @@
...
+    2. For each item currently at COMPLETED in a multi-reviewer queue with no
+       authoritative annotation, downgrade to AWAITING_RESOLUTION.
+    """
+    AnnotationItem = apps.get_model("human_annotations", "AnnotationItem")
+    now = timezone.now()
+
+    # (1) Single-reviewer auto-mark.
+    for item in AnnotationItem.objects.filter(queue__num_reviews_required=1):
+        submitted = list(item.annotations.filter(status="submitted"))
+        if len(submitted) == 1 and not submitted[0].is_authoritative:
```
SmittieC: So we skip items where the `num_reviews_required` is 1 but has multiple submissions. Can we not in this case make the first submission the authoritative one?

SmittieC (commented):
a few comments, but looks good to me

apps/human_annotations/aggregation.py
snopoke: That's correct. If an authoritative review is present, only that one counts for aggregations.

templates/human_annotations/columns/annotations_summary.html
snopoke: I'll update to use `fa-badge-check`

apps/human_annotations/migrations/0004_backfill_authoritative.py
snopoke: Acknowledged — keeping it as a Django data migration.

*— Claude Code*

apps/human_annotations/views/annotate_views.py
snopoke: Fixed in 3b66e98a1. Merged into a single `Annotation` lookup that filters on `item__queue_id=pk` and `item__queue__team=request.team`, with `select_related("item__queue")` so `annotation.item.queue` is available without an extra query.

*— Claude Code*

apps/human_annotations/views/annotate_views.py
snopoke: Good catch — the only callers are the HTMX-driven Mark/Clear authoritative buttons in `partials/annotation_list.html`. Removed the non-HTMX redirect fallback in 3b66e98a1; the view now always returns the swap partial.

*— Claude Code*

apps/human_annotations/aggregation.py
snopoke: Yes, that\'s right. Per item: if any submitted annotation has `is_authoritative=True`, only those contribute to aggregates; otherwise we fall back to every submitted annotation on the item. So once an admin picks one, aggregates lock to that pick. When new annotations come in later, the picked one still dominates until the admin clears or re-picks.

*— Claude Code*

templates/human_annotations/columns/annotations_summary.html
snopoke: Switched to `fa-circle-check` in 3b66e98a1 — `fa-badge-check` would have been my first pick but it\'s FA Pro only, and `fa-circle-check` reads similarly as a “verified” mark in FA Free 6.2.

*— Claude Code*

apps/human_annotations/tests/test_views.py
snopoke: Agreed — removed in 3b66e98a1. It just asserted the icon class string appeared in the rendered HTML, which is both fragile (broke as soon as we switched to `fa-circle-check`) and redundant with the auto-mark tests in `test_authoritative.py`.

*— Claude Code*

apps/human_annotations/migrations/0004_backfill_authoritative.py
snopoke: Good call — fixed in 3b66e98a1. Backfill now picks the earliest submitted annotation when a single-reviewer item has multiple submissions (the over-budget case), matching the runtime auto-mark behavior. Added `test_backfill_picks_earliest_when_single_reviewer_item_has_multiple_submissions` for it.

*— Claude Code*

SmittieC (dismissed):

PR #3373, opened 2026-05-19
Return session data from trigger_bot_message endpoint

### LINKED ISSUE #3369: Return session ID from 'trigger bot message' API endpoint

ocs-percy:
### Product Description
The `trigger_bot` endpoint previously returned an empty HTTP 200. It now returns session information so callers can reference the session immediately without a follow-up lookup:

```json
{
  "session_id": "<external_id>",
  "url": "https://.../api/sessions/<external_id>/",
  "team": {"name": "My Team", "slug": "my-team"},
  "channel": "whatsapp"
}
```

### Technical Description
Session creation was previously done inside `trigger_bot_message_task` (a Celery task), which meant the session ID wasn't available to return synchronously. The fix moves session creation into the view — where it was always needed anyway for the response — and the task now receives a `session_external_id` and focuses only on message delivery.

Key changes:
- New `TriggerBotMessageResponse` serializer (`ModelSerializer`) with `session_id`, `url`, `team`, `channel` fields
- `trigger_bot_message_task` signature changes from `(data: dict)` to `(session_external_id, prompt_text, message_text)`; it no longer calls `ensure_session_exists_for_participant`
- `@extend_schema` updated to document the 200 response body
- Existing tests updated with response body assertions

### Migrations
No migrations.

### Docs and Changelog

Fixes #3369

### DIFF
--- a/apps/api/serializers.py
+++ b/apps/api/serializers.py
@@ -130,7 +130,18 @@ class ExperimentSessionSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = ExperimentSession
-        fields = ["url", "id", "team", "experiment", "participant", "created_at", "updated_at", "status", "messages", "tags"]
+        fields = [
+            "url",
+            "id",
+            "team",
+            "experiment",
+            "participant",
+            "created_at",
+            "updated_at",
+            "status",
+            "messages",
+            "tags",
+        ]
 
     def __init__(self, *args, **kwargs):
         self._include_messages = kwargs.pop("include_messages", False)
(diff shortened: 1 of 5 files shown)


apps/api/tasks.py
```diff
@@ -83,37 +83,33 @@ def create_connect_channel_for_participant(channel, connect_client, connect_id,
...
     response via the LLM and sends that.
     """
-    platform = data["platform"]
-    experiment_public_id = data["experiment"]
-    prompt_text = data.get("prompt_text")
-    message_text = data.get("message_text")
-    identifier = data["identifier"]
-    start_new_session = data["start_new_session"]
-    session_data = data.get("session_data")
+    from apps.experiments.models import ExperimentSession  # noqa: PLC0415
```
snopoke: import should be at the module level

apps/api/serializers.py
```diff
@@ -308,3 +319,16 @@ def validate(self, data):  # ty: ignore[invalid-method-override]
...
         return data
+
+
+class TriggerBotMessageResponse(serializers.ModelSerializer):
+    session_id = serializers.ReadOnlyField(source="external_id")
+    url = serializers.HyperlinkedIdentityField(
+        view_name="api:session-detail", lookup_field="external_id", lookup_url_kwarg="id"
+    )
+    team = TeamSerializer(read_only=True)
+    channel = serializers.CharField(source="experiment_channel.platform", read_only=True)
```
snopoke: This should use the `platform` field directly on the session object

apps/api/views/channels.py
```diff
@@ -249,6 +247,21 @@ def post(self, request):
             if error := _ensure_commcare_connect_ready(channel, identifier, participant_data):
                 return error
 
-        trigger_bot_message_task.delay_on_commit(data)
+        from apps.chatbots.version_resolver import resolve_published_or_working  # noqa: PLC0415
```
snopoke: import should be at the module level

snopoke (dismissed):

snopoke (approved):

PR #3383, opened 2026-05-19
feat: add edit and delete actions for evaluations

### LINKED ISSUE #3203: Evals UI feedback
* Move 'add sessions' to existing dataset to a separate page (or load async) to avoid loading the table on every page load)

* Add 'edit' button to evaluation details page

barry47products:
### Product Description

Closes #3203.

Users can now delete an evaluation (EvaluationConfig) from two places:

- The evaluations list table — trash icon on each row, opens the standard alertify confirm dialog, removes the row inline via HTMX on confirm.
- The evaluation detail page header — Delete button (alongside a new Edit button), uses `hx-confirm` and navigates back to the list on success.

Both controls are permission-gated on `evaluations.delete_evaluationconfig` in the UI; the view enforces the same permission server-side. Deleting an EvaluationConfig cascades to its EvaluationRun and EvaluationResult rows — the confirm copy makes that explicit.

The approach was [discussed and approved in the issue thread](https://github.com/dimagi/open-chat-studio/issues/3203#issuecomment-4486883476) prior to implementation.

### Technical Description

- New `DeleteEvaluation` view in `apps/evaluations/views/evaluation_config_views.py`, mirroring the in-app `DeleteEvaluator` / `DeleteDataset` pattern (Django `DeleteView` + `get_queryset()` for team scoping + `PermissionRequiredMixin`).
- Wired up through `make_crud_urls(..., delete=True)` in `apps/evaluations/urls.py`.
- `EvaluationConfigTable` gets a `delete_action` (trash icon, alertify modal, inline row removal via empty 200).
- Detail page (`evaluation_runs_home.html`) gets Edit + Delete buttons. The Delete button passes `?redirect=1` so the view returns an `HX-Redirect` header to the list page (the detail page no longer exists post-delete). The list-table delete omits the query param and gets the default empty-200 → inline row removal.
- The `?redirect=1` check uses strict equality, not truthiness, so values like `?redirect=0` correctly fall through to the inline-removal path.
- No `messages.success` flash: Django messages do not survive `HX-Redirect` in this codebase, so the call was dropped during manual verification. The row vanishing (list view) or the populated list page (detail view) is the visual confirmation — same convention as `DeleteEvaluator` / `DeleteDataset`.

### Migrations

No migrations in this PR.

### Demo

**Evaluations list with edit / view / run / delete actions:**
1-evaluations-list-with-actions

**Evaluation detail page with new Edit and Delete buttons:**
2-evaluation-runs-edit-delete

**Edit evaluation form:**
3-edit-evaluation-form

**Delete confirmation dialog (cascade warning):**
4-delete-confirmation

### Docs and Changelog

### Follow-up issues

Three out-of-scope items surfaced during the design pass were approved for follow-up issues (I will file these):

1. Modernize evaluations `DeleteView`s to use `form_valid()` instead of overriding `delete()`.
2. Guard destructive evaluation operations against in-flight runs (Evaluator / Dataset / EvaluationConfig).
3. Standardize confirmation dialog UX for destructive actions (alertify modal vs `hx-confirm`).


barry47products:
@coderabbit review

snopoke (commented):
Code looks fine. @barry47products please could you include screenshots in the PR to help verify the UI changes.

barry47products:
Thanks @snopoke — pushed `8337f18` to swap the detail-page delete confirm from the native browser dialog to the alertify modal that the table-row action and tag deletion already use, plus added screenshots above.

Before this lands, a few things I'd appreciate a sanity check on — things that occurred to me on a second pass and that the existing tests don't cover:

1. **Versioning interaction.** `EvaluationConfig` uses `VersionsMixin` (working ↔ published versions). The current `DeleteView` does a straight `self.object.delete()` — should deletion of a working version with published descendants (or a published version) be blocked, cascaded explicitly, or warned about in the confirm copy? The message currently only mentions "run history and results".
2. **Related-object cascade.** The confirm promises run history and results are deleted. Worth confirming the FK chain (`EvaluationRun` → results/messages) all cascades and nothing is `PROTECT`ed, otherwise a delete could 500.
3. **Audit logging.** Does `EvaluationConfig` deletion get captured via `field_audit`? Happy to add an assertion in the test if that's the expectation.
4. **Soft delete convention.** Some models in OCS prefer soft delete (`is_active=False`) for traceability. Is hard delete the right call for `EvaluationConfig` or should we follow a soft-delete pattern?

Happy to address any of these in this PR or as follow-ups — whichever you prefer.

SmittieC (approved):
LGTM

snopoke (approved):

snopoke:
> Thanks @snopoke — pushed `8337f18` to swap the detail-page delete confirm from the native browser dialog to the alertify modal that the table-row action and tag deletion already use, plus added screenshots above.
> 
> Before this lands, a few things I'd appreciate a sanity check on — things that occurred to me on a second pass and that the existing tests don't cover:
> 
> 1. **Versioning interaction.** `EvaluationConfig` uses `VersionsMixin` (working ↔ published versions). The current `DeleteView` does a straight `self.object.delete()` — should deletion of a working version with published descendants (or a published version) be blocked, cascaded explicitly, or warned about in the confirm copy? The message currently only mentions "run history and results".

`EvaluationConfig` get's the version mixin from `BaseModel` purely as a convenience. We don't version any of the evaluation models at all so this isn't an issue.

> 2. **Related-object cascade.** The confirm promises run history and results are deleted. Worth confirming the FK chain (`EvaluationRun` → results/messages) all cascades and nothing is `PROTECT`ed, otherwise a delete could 500.

LGTM - also I assume you tested this locally right?

> 3. **Audit logging.** Does `EvaluationConfig` deletion get captured via `field_audit`? Happy to add an assertion in the test if that's the expectation.

These aren't audited models so there's nothing to do here.

> 4. **Soft delete convention.** Some models in OCS prefer soft delete (`is_active=False`) for traceability. Is hard delete the right call for `EvaluationConfig` or should we follow a soft-delete pattern?

I don't think it is necessary.

barry47products:
> > 2. **Related-object cascade.** The confirm promises run history and results are deleted. Worth confirming the FK chain (`EvaluationRun` → results/messages) all cascades and nothing is `PROTECT`ed, otherwise a delete could 500.
>
> LGTM - also I assume you tested this locally right?

Thanks @snopoke — and good to have 1, 3 and 4 confirmed.

On 2: yes, tested locally. I deleted an `EvaluationConfig` that had 2 runs and 6 results — all of them were removed cleanly (no 500), and the dataset and evaluator were correctly left intact.

PR #3396, opened 2026-05-20
Allow Chatbot Admin and Chat Viewer groups to download files

### LINKED ISSUE #925: Unable to download exports for when I have "experiment admin" and "chat viewer" groups
Assign the "experiment admin" and "chat viewer"  groups to a user, generate an export and download it. You are unable to.

barry47products:
### Product Description

Users assigned the **Chatbot Admin** or **Chat Viewer** group could not download
files. A Chatbot Admin could generate a chat export but hit a permission error
when clicking the download link; a Chat Viewer hit the same wall on session
attachment downloads. Members of both groups can now download these files as
expected.

### Technical Description

`FileView` (`apps/files/views.py`) is gated by the `files.view_file` permission,
but neither the `Chatbot Admin` nor `Chat Viewer` group held any `files`
permission — only `Assistant Admin` did. The export-download link (and session
attachment downloads) therefore returned a permission denial.

This adds `ModelPermSetDef("files", "file", [VIEW])` to both groups in
`apps/teams/backends.py`:

- **Chatbot Admin** — for chat-export downloads and collection (documents) file access.
- **Chat Viewer** — for session attachment downloads.

Scoped to `view_file` only, not the full `files` set: collection file add/delete
is already gated by `documents.change_collection`, so the read permission was the
only thing missing. No data migration is required — permission groups re-sync
from `GROUPS` on every `migrate` via the existing post-migrate signal.

Note: with the permission missing, `permission_required` redirects an
already-logged-in user back to the login page, which surfaces in the browser as a
redirect loop (`ERR_TOO_MANY_REDIRECTS`) rather than a clean 403. Granting the
permission resolves the reported bug; tightening that denial to a proper 403 would
be a separate follow-up.

Tests added:

- `apps/teams/tests/test_permissions.py` — regression tests asserting each group's
  definition includes `view_file`.
- `apps/files/tests/test_file_view_permissions.py` — view-level test that a user
  holding only `Chatbot Admin`, and one holding only `Chat Viewer`, can `GET`
  `FileView` and receive the file instead of a redirect.

### Migrations

_No migrations in this PR — permission groups re-sync from `apps/teams/backends.py`
on every `migrate` via the existing post-migrate signal._

### Demo

**Before** — a user in both the Chatbot Admin and Chat Viewer groups generates a
chat export, clicks "Download export", and the download fails (the `files.view_file`
denial bouncing through the login redirect → `ERR_TOO_MANY_REDIRECTS`):

before-3-download-denied

**After** — the same user downloads the export successfully and opens it:
after-1-download-dialog
after-2-csv-in-excel

### Docs and Changelog

Fixes #925

### DIFF
--- a/apps/files/tests/test_file_view_permissions.py
+++ b/apps/files/tests/test_file_view_permissions.py
@@ -1,0 +1,38 @@
+import pytest
+from django.urls import reverse
+
+from apps.teams.backends import (
+    CHAT_VIEWER_GROUP,
+    CHATBOT_ADMIN_GROUP,
+    add_user_to_team,
+    create_default_groups,
+)
+from apps.utils.factories.files import FileFactory
+from apps.utils.factories.team import TeamFactory
+from apps.utils.factories.user import UserFactory
+
+
+@pytest.mark.django_db()
+@pytest.mark.parametrize("group_name", [CHATBOT_ADMIN_GROUP, CHAT_VIEWER_GROUP])
+def test_file_view_accessible_to_group(client, group_name):
+    """Regression test for #925: members of these groups can download a team File.
+
+    FileView is gated by `files.view_file`. Without that permission a logged-in
+    member who can generate a chat export (Chatbot Admin) or view a session
+    attachment (Chat Viewer) is denied on download.
+
+    create_default_groups() is called explicitly so the DB-backed groups reflect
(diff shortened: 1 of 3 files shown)


barry47products:
@coderabbit review

barry47products:
@coderabbitai review

SmittieC (approved):

PR #3397, opened 2026-05-21
BSUID support for WhatsApp

### LINKED ISSUE #3322: WhatsApp BSUID / usernames support — webhook + send-message changes needed (Meta Cloud + Twilio)
WhatsApp is launching **usernames** later in 2026 (rolling out June 2026 per Twilio). Once a user adopts a username, their phone number is hidden in the app and **omitted from webhook payloads** in most cases. Meta replaces the missing phone with a new identifier — **Business-Scoped User ID (BSUID)** — in a new `user_id` field (format: `US.13491208655302741918`).

BSUIDs start appearing in webhooks **early April 2026**, before usernames go live, so integrations can adopt the new field ahead of the consumer rollout. Outbound sending to BSUIDs is enabled by Meta in **May 2026**.

This affects both of our WhatsApp paths:

- **Meta Cloud API**: webhooks may arrive with no `wa_id`/`from`; send-message requests must use a new `recipient` field instead of `to` when only a BSUID is known. Auth templates (one-tap / zero-tap / copy-code) still require phone numbers.
- **Twilio WhatsApp**: Twilio is introducing an `ExternalUserId` field for BSUIDs and will populate `To`/`From` with BSUIDs when phone numbers aren't available.

Without changes, conversations from username-adopters will lose context or fail to deliver entirely. This ticket is to scope and plan the work.

**References**
- Meta: https://developers.facebook.com/documentation/business-messaging/whatsapp/business-scoped-user-ids
- Twilio: https://www.twilio.com/en-us/changelog/whatsapp-usernames--new-business-scoped-user-id--bsuid--field-re

SmittieC:
Resolves #3322 

### Product Description
Adds support for WhatsApp Business-Scoped User IDs (BSUIDs) — the stable, privacy-preserving identifiers Meta introduced as part of the WhatsApp username rollout. Previously OCS relied on phone numbers as participant identifiers; this change updates inbound message parsing, session resolution, and outbound send payloads to handle BSUIDs alongside phone numbers.

> **Note:** Sending messages via BSUID is not yet enabled. For now, participant lookup and session creation still fall back to the phone number as the canonical identifier. The infrastructure is in place to switch once Meta fully supports BSUID-addressed sends.

### Technical Description
**Inbound parsing (`datamodels.py`, `meta_webhook.py`)**
- Added `looks_like_bsuid(value)` regex helper matching Meta's spec (`CC.identifier` / `CC.ENT.identifier`).
- `MetaCloudAPIMessage.parse()` now receives a single `MetaCloudAPIWebhookMessage` (a new TypedDict) instead of a full value dict, and captures `from_user_id` (BSUID) and `from` (phone, when exposed) separately. The `phone_number` field carries the E.164 phone only when the sender has not adopted a username.
- `TwilioMessage.from_message_data()` reads `ExternalUserId` as the BSUID and preserves the phone from `From` when it is not itself a BSUID.
- `extract_message_values()` is replaced by `extract_messages()` which returns `(phone_number_id, message)` tuples — one per message rather than one per value — simplifying the webhook view loop.

**Participant resolution (`channels_v2/callbacks.py`, `whatsapp_channel.py`, `channels_v2/stages/core.py`, `chat/channels.py`)**
- Added `ChannelCallbacks.get_participant_identifier_filter()` returning a simple `Q(identifier=...)` by default.
- `WhatsappCallbacks` overrides it to return `Q(identifier=bsuid) | Q(identifier=phone)` when both are available, preserving continuity for existing phone-keyed participants when a BSUID starts arriving.
- `SessionResolutionStage` uses this filter (via a `_participant_qs` subquery) when looking up active sessions, instead of the previous `participant__identifier=` equality.
- `_start_experiment_session()` accepts an optional `participant_id_filter: Q` argument; when the filter is a disjunction it probes first and creates only with the canonical (normalized) identifier.

**Outbound sends (`messaging_service.py`)**
- Added `_recipient_kwarg(recipient)` helper that routes BSUIDs to `{"recipient": ...}` and phone numbers to `{"to": ...}` in Meta Cloud API payloads. Applied to all three send methods (template, text, audio, media).

**Cleanup**
- Removed `EvalsBotInteractionStage` and `EvaluationChannel` (dead code, superseded by refactored evaluation path).

### Migrations
No migrations.

### Demo
No UI change. The feature is exercised via WhatsApp webhook delivery — existing participants remain matched and conversations continue uninterrupted when Meta starts sending BSUIDs instead of phone numbers.

### Docs and Changelog


apps/channels/datamodels.py
```diff
@@ -214,22 +254,32 @@ class MetaCloudAPIMessage(TurnWhatsappMessage):
...
-
+            body = message_data["text"]["body"]
+
+        # BSUID (`from_user_id`) is the stable identifier — it's present on every post-rollout
+        # webhook regardless of whether the user adopted a username. A missing field means a
+        # malformed payload, so we let the KeyError surface.
+        participant_id = message_data["from_user_id"]
+        from_value = message_data.get("from")
+        phone_number = from_value if from_value and not looks_like_bsuid(from_value) else None
         return MetaCloudAPIMessage(
```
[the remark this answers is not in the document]
SmittieC: The comment references an earlier version of the code where `participant_id` was hard-assigned from `from_user_id` only. The current implementation already handles both fields — BSUID as the stable identifier, with phone as a fallback for legacy matching. No change needed.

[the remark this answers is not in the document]
SmittieC: Fixed in a68b7d4ce. Investigated the concern but found the correct approach is to keep `phone_number` as the raw wa_id (no E.164 normalization) since legacy Meta Cloud API participants are stored in that same format — normalizing to E.164 would actually break the legacy matching. The commit also fixes a related bug introduced in 0e188cd where the phone-first participant_id approach broke username-adopter sessions (`participant_id=None`) and participant resolution scenarios 2–4.

apps/channels/meta_webhook.py
```diff
@@ -1,28 +1,98 @@
...
+        if user_key not in text_groups:
+            first_idx[user_key] = len(result)
+            text_groups[user_key] = []
+            result.append(msg)  # Placeholder; overwritten below if the group has >1 message.
+        text_groups[user_key].append(msg)
+
+    for user_key, group in text_groups.items():
+        if len(group) == 1:
+            continue
+        group.sort(key=lambda m: int(m.get("timestamp", "0")))
```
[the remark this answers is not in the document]
SmittieC: The `group.sort(key=lambda m: int(m.get("timestamp", "0")))` code this comment references does not exist in the current `meta_webhook.py` (the file is 88 lines with no timestamp sorting). This comment appears to be based on a stale version of the file.

apps/chat/channels.py
```diff
@@ -1459,6 +1460,7 @@ def _start_experiment_session(
     timezone: str | None = None,
     session_external_id: str | None = None,
     metadata: dict | None = None,
+    participant_id_filter: Q | None = None,
```
SmittieC, line 1501: Acknowledged. The complexity increase in `_start_experiment_session` is a side-effect of adding the `participant_id_filter` parameter for BSUID/phone dual-matching. Refactoring this function is out of scope for this PR, but the concern is noted for a follow-up.

SmittieC, line 1501: Same as above — the increased nesting is inherent to the BSUID/legacy-phone resolution logic added in this PR. Refactoring is out of scope here.

apps/service_providers/tests/test_messaging_providers.py
```diff
@@ -8,7 +8,7 @@
 from django.utils import timezone
 from pydantic import ValidationError
 
-from apps.channels.datamodels import MetaCloudAPIMessage, TurnWhatsappMessage
+from apps.channels.datamodels import MetaCloudAPIMessage, TurnWhatsappMessage, looks_like_bsuid
```
SmittieC: Acknowledged. The new BSUID resolution and participant-matching tests added in this PR push the file over the threshold. Splitting the test file is out of scope for this PR.

SmittieC: Acknowledged — the new BSUID-related test classes add responsibilities to this file. Same as above: splitting is out of scope for this PR.

apps/channels/datamodels.py
```diff
@@ -148,19 +174,43 @@ def parse(message_data: dict) -> "TwilioMessage":
...
+            # it mirrors the BSUID. Normalize to E.164 for matching against legacy phone-keyed
+            # participants stored in the DB.
+            if not looks_like_bsuid(from_value):
+                phone_number = phonenumbers.format_number(
+                    phonenumbers.parse(from_value), phonenumbers.PhoneNumberFormat.E164
+                )
+
+            # Sending BSUIDs are not yet supported, so we use the phone number for now. Once this is supported,
+            # remove this line
+            participant_id = phone_number
```
SmittieC, line 200: Note this line and a similar one in `MetaCloudAPIMessage`. From the PR description:
> Sending messages via BSUID is not yet enabled. For now, participant lookup and session creation still fall back to the phone number as the canonical identifier. The infrastructure is in place to switch once Meta fully supports BSUID-addressed sends.

We can either merge this PR in the meantime and remove these lines when meta supports BSUIDs, or hold off and only merge these changes once they do.

apps/channels/channels_v2/stages/core.py
```diff
@@ -132,8 +142,12 @@ def _create_session(self, ctx: MessageProcessingContext):
             participant_identifier=ctx.participant_identifier,
             participant_user=ctx.channel_context.get("participant_user"),
             session_status=SessionStatus.SETUP,
+            participant_id_filter=self.participant_id_filter(ctx),
         )
 
+    def participant_id_filter(self, ctx: MessageProcessingContext) -> Q:
+        return ctx.callbacks.get_participant_identifier_filter(ctx.participant_identifier, ctx.message)
```
snopoke: how come this is delegating to the callback instead of the channel implementation iteself?

apps/channels/tests/test_datamodels.py
```diff
@@ -15,3 +18,81 @@ def test_attachment_file_ids_round_trip(self):
...
+        message = {
+            "From": "whatsapp:+27456897512",
+            "To": self.BUSINESS,
+            "Body": "Hello",
+        }
+        with pytest.raises(KeyError):
+            TwilioMessage.parse(message)
+
+
+class TestLooksLikeBsuid:
```
snopoke: should be a parameterized test

apps/channels/datamodels.py
snopoke: Isn't this code used when receiving messages, not when sending? This appears to work differently from the `MetaCloudAPIMessage` class which is confusing.

I thought we were going to store the BSUID but also keep the phone number and keep using the phone number for sending until BSUID is supported?

apps/channels/channels_v2/stages/core.py
SmittieC: At the time it seemed better to have it in the callbacks, since the callacks class can essentially provide runtime configuration instead of build-time configs, but I do think it breaks current convention looking at it again now. I'll change it so we pass in a ref to `SessionResolutionStage` to configure it at build time rather.

SmittieC: Fixed in 41cd2447a. Moved the filter out of callbacks entirely — `ChannelBase` now exposes `_get_participant_id_filter()` (returns `None` → default `Q(identifier=...)` path), and `SessionResolutionStage` accepts it as a constructor arg. `WhatsappChannel` overrides `_get_participant_id_filter()` with the BSUID-or-phone logic. Consistent with the existing `_get_callbacks()` / `_get_sender()` pattern.

apps/channels/tests/test_datamodels.py
SmittieC: Fixed in 41cd2447a. Converted to `@pytest.mark.parametrize`.

apps/channels/datamodels.py
SmittieC: > Isn't this code used when receiving messages, not when sending?

Correct. The reason for this is that the `participant_id` from the message is what we ultimately use for the "to" field when sending reply messages. Since Meta's API [does not yet support receiving messages](https://developers.facebook.com/documentation/business-messaging/whatsapp/business-scoped-user-ids/#phone-numbers) with a BSUID in it last time I checked (which was today), we cannot use the BSUID as the `participant_id` yet, so we  need to temporarily keep the `participant_id` set to the phone number. Overriding it in at the message level was the easiest way I could see to achieve this. When Meta's API accepts BSUIDs, we can simlpy remove these two lines of code where we make the phone number the participant_id.

From the docs:
> BSUIDs will begin appearing in webhooks in early April 2026. However, our APIs will not support sending messages targeted to the BSUIDs until May 2026 (exact date pending). Once our APIs support BSUIDs in May, you will be able to message users using either their BSUID, phone number, or both.

I'd be happy to chat about this more if you'd like. We can also just wait a few days until June and see if anything changed.

> This appears to work differently from the MetaCloudAPIMessage class which is confusing.

It does a bit, but only because Twilio's message payload looks different than Meta's. Which part are you referring to specifically?

snopoke (commented):
@SmittieC this generally looks good to me though I had some thoughts on the channels changes.

What about creating a new 'ParticipantResolutionStage' stage which can easily be overridden by the channels. Currently we do participant resolution in the 'SessionResolutionStage'.

SmittieC:
> @SmittieC this generally looks good to me though I had some thoughts on the channels changes.
> 
> What about creating a new 'ParticipantResolutionStage' stage which can easily be overridden by the channels. Currently we do participant resolution in the 'SessionResolutionStage'.

Good idea

SmittieC:
> @SmittieC this generally looks good to me though I had some thoughts on the channels changes.
> 
> What about creating a new 'ParticipantResolutionStage' stage which can easily be overridden by the channels. Currently we do participant resolution in the 'SessionResolutionStage'.

I'm going to ship this separately (before this PR)

PR #3414, opened 2026-05-22
Reset applied tags when rerunning evals

### LINKED ISSUE #3409: Reverse tag application
AppliedTag keeps track of which evaluation result applied which tag. We can use this to allow reversing tag application and / or clear tags before running an evaluation again.

SmittieC:
Closes #3409 

### Product Description

When an evaluation is rerun, tags that were previously applied by evaluators but are no longer triggered in the new run are automatically removed. This keeps tag state on chat messages and chats in sync with the latest evaluation results.

### Technical Description

Added `reverse_stale_tags(run)` in `apps/evaluations/tagging.py`, called from `mark_evaluation_complete` in `tasks.py` after a run finishes (after aggregates are computed).

**How it works:**
- Collects all tag IDs managed by the run's evaluators (via their `EvaluatorTagRule`s).
- Batches a single `AppliedTag` query for the entire run, grouped by message ID, to avoid an O(N) query per message.
- For each message in scope, removes any managed tags that were not applied in this run from the resolved target object (`ChatMessage` for `MESSAGE` mode, `Chat` for `SESSION` mode).

**Key decisions:**
- `PREVIEW` runs are skipped entirely — they're not authoritative and shouldn't mutate tags.
- `DELTA` runs only clean up their `scoped_messages`; `FULL` runs clean up all messages in the dataset.
- Not wrapped in `transaction.atomic()` — a mid-loop failure leaves some stale tags in place, but a subsequent rerun will finish the cleanup. Wrapping the whole loop in a single transaction would hold locks on potentially large sets of rows.

### Migrations

No migrations.

### Demo (excuse the crying baby in the background)
https://www.loom.com/share/e7bb18b8cff4401db72c886138f14ea4

### Docs and Changelog

When updating the docs, be sure to make it clear that evaluators can remove tags that might have been applied by humans. The philosophy is: a session or message either meet the criteria for a tag, or not.


docs/superpowers/specs/2026-05-22-tag-removal-on-eval-rerun-design.md
```diff
@@ -0,0 +1,78 @@
...
+PREVIEW runs skip cleanup entirely, consistent with the existing behaviour that skips tag application for previews.
+
+## Algorithm
+
+`possible_tags` is computed once per run (shared across all messages):
+
+```
+possible_tags = all Tag IDs referenced by EvaluatorTagRules
+                across all evaluators attached to run.config
+```
```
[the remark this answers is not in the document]
SmittieC: The spec file was removed in commit 7fa0c70c7 ('Remove design spec now that implementation is complete') — the file no longer exists in the branch. Suggestion is no longer applicable.

apps/evaluations/tagging.py
```diff
@@ -134,3 +137,55 @@ def apply_rules_to_result(
...
+            stale_by_target[target.pk] |= stale_tags
+
+    if not stale_by_target or content_type is None:
+        return
+
+    filter_q = Q()
+    for target_pk, tag_ids in stale_by_target.items():
+        filter_q |= Q(object_id=target_pk, tag_id__in=tag_ids)
+
+    CustomTaggedItem.objects.filter(content_type=content_type).filter(filter_q).delete()
```
SmittieC, line 208: Fixed in 51f97b1f8. Extracted three private helpers (_get_possible_tags, _build_applied_by_message, _compute_stale_by_target) from reverse_stale_tags to bring cyclomatic complexity from 13 down to ~6, well below the threshold of 11. All 8 existing tests pass.

snopoke: Shouldn't this also filter on `AppliedTag` so that you don't remove tags that were applied in other ways (user / some other evals).

I think we should filter stale tags down to those that were actually applied in a previous run of this EvaluationConfig. We could also potentially filter by `AppliedTag.rule`.

snopoke (approved):

PR #3421, opened 2026-05-24
Provider-usage tooling: per-provider usages page + cross-team API key search

snopoke:
### Product Description
Two operator tools for tracing where a service provider is used:

- A **Show usages** button on each team-scoped provider edit page that lists the chatbots, assistants, pipelines, collections, and channels referencing the provider. Versioned references are tagged and links route to the working version.
- A staff-only **Find provider by key** admin tool that takes an API key (full, suffix, or substring) and a provider type, and returns matching providers across all teams.

### Technical Description
Shared logic in `apps/service_providers/usages.py`:

- `get_provider_usages(provider)` walks reverse FKs via the existing `get_related_objects` helper, and (for LLM providers) also resolves pipeline node params. Categories are reshaped so users see chatbots first: pipelines/channels/document sources roll up to their owning chatbot or collection, with dedupe.
- `search_providers_by_api_key` iterates providers of the chosen type and matches the supplied key against each subtype's declared `obfuscate_fields` from its config form.
- A single `usage_item.html` partial renders every item: links target the working version's edit URL, version badges read `version_number` directly so `VersionsMixin` subclasses without that field don't crash, and `#versions` is only appended for chatbot links.

### Demo

**Admin Tool**

image

**Provider usages**
image

### Docs and Changelog

Notes for the changelog: adds two operator tools (per-provider "Show usages" and staff-only "Find provider by API key") to help trace where credentials are used; useful for credential-rotation / leaked-key triage.


apps/service_providers/tests/test_usages.py
```diff
@@ -0,0 +1,346 @@
...
+    entry = chatbots_cat.items[0]
+    assert entry["chatbot"].id == chatbot.id
+    assert len(entry["channels"]) == 2
+
+    unlinked = next(c for c in usages.categories if c.label == "Unlinked Channels")
+    assert len(unlinked.items) == 1
+
+
+@pytest.mark.django_db()
+def test_experiment_category_displays_as_chatbots(team_with_users):
```
snopoke: remove this test

apps/admin/tests/test_find_provider_by_key.py
snopoke, line 1: most of the tests in this file are covered by test in `test_usages`. Let's remove this test file.

apps/admin/views.py
```diff
@@ -431,3 +432,38 @@ def configuration(request):
...
+@is_staff
+def find_provider_by_key(request):
+    """Staff tool: paste an API key + pick a provider type to find usages.
+
+    Iterates providers of the chosen type, decrypts each one's config and
+    compares the configured secret fields against the supplied key.
+    """
+    form = FindProviderByKeyForm(request.POST or None)
+    results = None
+    if request.method == "POST" and form.is_valid():
```
SmittieC, line 446: can we use a class based view instead?

SmittieC (commented):
Nice! One comment only from my side

apps/service_providers/tests/test_usages.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+@pytest.mark.django_db()
+def test_document_sources_roll_up_to_collections(team_with_users, client):
+    from apps.utils.factories.documents import CollectionFactory, DocumentSourceFactory  # noqa: PLC0415
+    from apps.utils.factories.service_provider_factories import (  # noqa: PLC0415
+        AuthProviderFactory,
+        EmbeddingProviderModelFactory,
+        LlmProviderFactory,
+    )
```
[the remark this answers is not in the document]
snopoke: Not applicable — these factory imports are already at module scope (apps/service_providers/tests/test_usages.py lines 8-21). Finding was correctly flagged outdated.

*— Claude Code*

templates/admin/find_provider_by_key.html
```diff
@@ -0,0 +1,72 @@
...
+  <section class="app-card">
+    <div class="flex items-center justify-between mb-4">
+      <h1 class="pg-title">Find provider by API key</h1>
+      <a href="{% url 'ocs_admin:home' %}" class="btn btn-ghost btn-sm">Back to admin</a>
+    </div>
+    <p class="text-sm text-gray-500 mb-4">
+      Paste a full API key, the last few characters, or any substring, then pick a provider type.
+      The search iterates all providers of that type and decrypts each config to match against
+      its secret fields. Use "Suffix" or "Contains" when you only have part of the key.
+    </p>
```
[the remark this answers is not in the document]
snopoke: Fixed in 18489918. Loaded the `i18n` template tag and wrapped user-facing strings in `{% translate %}` / `{% blocktranslate %}`.

*— Claude Code*

apps/service_providers/tests/test_usages.py
snopoke: Fixed in 18489918. Removed `test_experiment_category_displays_as_chatbots` — the Chatbots/Experiments label assertion is already covered by `test_voice_provider_excludes_synthetic_voices`.

*— Claude Code*

apps/admin/tests/test_find_provider_by_key.py
snopoke: Fixed in 18489918. Deleted the file.

*— Claude Code*

apps/service_providers/usages.py
```diff
@@ -0,0 +1,280 @@
...
+            matches.append(provider)
+    return matches
+
+
+def _secret_field_names(provider) -> Iterable[str]:
+    """The set of config keys that hold credentials for ``provider``'s subtype."""
+    form_cls = getattr(provider.type_enum, "form_cls", None)
+    if form_cls is None:
+        return ()
+    return getattr(form_cls, "obfuscate_fields", ()) or ()
```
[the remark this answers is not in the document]
snopoke: Fixed in 18489918. `_secret_field_names` now catches `KeyError`/`ValueError` from `provider.type_enum`, logs a warning, and skips the offending row instead of letting it crash the search view.

*— Claude Code*

apps/admin/views.py
snopoke: Noted — declining for now. Every other view in `apps/admin/views.py` is function-based, so converting just this one to a CBV would be inconsistent with the file. The view is a simple form/results handler that doesn't gain much from a class hierarchy. Happy to revisit if the surrounding module gets a CBV pass.

*— Claude Code*

snopoke: @SmittieC I think this was doing more with the form data e.g. if it was a model form then it would warrant it but as such I think it's OK to leave it as is.

PR #3432, opened 2026-05-25
Fix retrieval quality on Voyage/Google indexed collections (#3237)

### LINKED ISSUE #3237: Use embed_documents for indexing and embed_query for retrieval in local embedding providers
## Summary

All local embedding providers (`OpenAILocalIndexManager`, `GoogleLocalIndexManager`, `VoyageAILocalIndexManager`) currently use `embed_query` for both document indexing and retrieval search. Many embedding APIs distinguish between these two use cases and provide task-specific prompts when the correct input type is specified, which can improve retrieval quality.

For example, Voyage AI recommends using `input_type="document"` when embedding content for indexing and `input_type="query"` for search queries. Similar distinctions exist for other providers.

## Proposed Change

Update the `LocalIndexManager` interface and all concrete implementations so that:
- `add_files` (indexing path) uses `embed_documents` instead of `embed_query`
- `query` (retrieval path) continues using `embed_query`

This applies to all three providers in `apps/service_providers/llm_service/index_managers.py`:
- `OpenAILocalIndexManager`
- `GoogleLocalIndexManager`
- `VoyageAILocalIndexManager`

## References

- Voyage AI embeddings docs: https://docs.voyageai.com/docs/embeddings
- Raised in PR #3230 by @Raymond9734: https://github.com/dimagi/open-chat-studio/pull/3230#discussion_r3152253718
- Requested by @snopoke

barry47products:
### Product Description

Fixes two bugs on the local-indexed-collections path.

**1. Retrieval quality (#3237).** Voyage and Google embedding models return different vectors depending on whether text is a document being indexed or a query being matched. OCS was labelling both sides as a query, so for any chatbot using a local Voyage or Google collection, document chunks were embedded in the wrong mode and retrieval was weaker than these providers can deliver. OpenAI's embeddings API has no input-type concept, so OpenAI users are unaffected at the API level.

**2. OpenAI base URL.** `OpenAILocalIndexManager` was constructed without the provider's `openai_api_base`, so local indexing always hit `api.openai.com` even when the provider was configured against an OpenAI-compatible proxy (LiteLLM, self-hosted gateways, etc.) — teams on that setup got a `401` with no actionable error.

| Provider | Indexing                                             | Retrieval                                     | Quality gain                    |
| -------- | ---------------------------------------------------- | --------------------------------------------- | ------------------------------- |
| VoyageAI | `embed_documents` → `input_type="document"`          | `embed_query` → `input_type="query"`          | Yes                             |
| Google   | `embed_documents` + `task_type="RETRIEVAL_DOCUMENT"` | `embed_query` + `task_type="RETRIEVAL_QUERY"` | Yes                             |
| OpenAI   | `embed_documents`                                    | `embed_query`                                 | No — interface consistency only |

> **Action for operators of Voyage/Google local collections:** historical `FileChunkEmbedding` rows were embedded under the old mode and are not rewritten by this PR. The quality improvement applies to newly-indexed files only; existing files need to be re-uploaded (or re-indexed) to benefit.

Closes #3237.

### Technical Description

Reviewer should pay particular attention to:

- **`LocalIndexManager.get_embedding_vector` signature change.** Now `(self, content: str, *, input_type: EmbeddingInputType) -> Vector` — keyword-only, runtime `ValueError` on unknown values, `Literal["document", "query"]` extracted to a module-top type alias (matches `MatchMode` / `AttachmentType` convention).
- **Google `task_type="RETRIEVAL_DOCUMENT"` on `embed_documents` is load-bearing, not redundant.** `langchain-google-genai`'s `embed_documents` does not default `task_type` (only `embed_query` does). In-code comment added so a future cleanup doesn't strip it.
- **Google `output_dimensionality` passed on both paths.** Without it on the document path, Gemini returns its native dimension and the `HalfVectorField(dimensions=1024)` insert fails. Both branches now pass `settings.EMBEDDING_VECTOR_SIZE`.
- **Voyage routing.** The langchain wrapper's `embed_documents` / `embed_query` map to Voyage's `input_type=document` / `input_type=query` API params — this is where the actual quality gain comes from for Voyage users.
- **OpenAI base URL plumbing.** `OpenAILocalIndexManager.__init__` accepts `openai_api_base: str | None = None`; passed as `base_url=...` to `OpenAIEmbeddings` only when truthy (no `base_url=None` short-circuiting langchain's defaults). `OpenAILlmService.get_local_index_manager` threads `self.openai_api_base` through.

**Not in scope:** `LocalIndexManager.query` filters chunks by `collection_id` only, not `FileStatus`, so chunks from a `FAILED` file can surface in retrieval results. Reproducible locally; filing as its own issue.

### Migrations

No DB schema changes. Rollback is a code revert.

### Demo

No UI changes. Evidence below is from a small script that mocks each provider's langchain class and traces what method gets called with what kwargs on the indexing path vs the retrieval path. Same script run against `main` and against this branch.

**On `main` (buggy):**

```
--- OpenAI ---
OpenAI provider with NO custom openai_api_base:
  document  ctor kwargs: {'api_key': 'k', 'model': 'text-embedding-3-small', 'dimensions': 1024}
            -> .embed_query('the cat sat on the mat')
  query     ctor kwargs: {'api_key': 'k', 'model': 'text-embedding-3-small', 'dimensions': 1024}
            -> .embed_query('where did the cat sit?')

OpenAI provider configured with openai_api_base="https://proxy.example.com/v1",
BUT constructor does not accept it — proxy URL silently dropped:
  document  ctor kwargs: {'api_key': 'k', 'model': 'text-embedding-3-small', 'dimensions': 1024}
            -> .embed_query('the cat sat on the mat')
  query     ctor kwargs: {'api_key': 'k', 'model': 'text-embedding-3-small', 'dimensions': 1024}
            -> .embed_query('where did the cat sit?')

--- Google ---
  document  ctor kwargs: {'google_api_key': 'k', 'model': 'models/text-embedding-004'}
            -> .embed_query('the cat sat on the mat', output_dimensionality=1024)
  query     ctor kwargs: {'google_api_key': 'k', 'model': 'models/text-embedding-004'}
            -> .embed_query('where did the cat sit?', output_dimensionality=1024)

--- Voyage ---
  document  ctor kwargs: {'voyage_api_key': 'k', 'model': 'voyage-4-large', 'output_dimension': 1024}
            -> .embed_query('the cat sat on the mat')
  query     ctor kwargs: {'voyage_api_key': 'k', 'model': 'voyage-4-large', 'output_dimension': 1024}
            -> .embed_query('where did the cat sit?')
```

**On this branch:**

```
--- OpenAI ---
OpenAI provider with NO custom openai_api_base:
  document  ctor kwargs: {'api_key': 'k', 'model': 'text-embedding-3-small', 'dimensions': 1024}
            -> .embed_documents(['the cat sat on the mat'])
  query     ctor kwargs: {'api_key': 'k', 'model': 'text-embedding-3-small', 'dimensions': 1024}
            -> .embed_query('where did the cat sit?')

OpenAI provider WITH openai_api_base="https://proxy.example.com/v1":
  document  ctor kwargs: {'api_key': 'k', 'model': 'text-embedding-3-small', 'dimensions': 1024, 'base_url': 'https://proxy.example.com/v1'}
            -> .embed_documents(['the cat sat on the mat'])
  query     ctor kwargs: {'api_key': 'k', 'model': 'text-embedding-3-small', 'dimensions': 1024, 'base_url': 'https://proxy.example.com/v1'}
            -> .embed_query('where did the cat sit?')

--- Google ---
  document  ctor kwargs: {'google_api_key': 'k', 'model': 'models/text-embedding-004'}
            -> .embed_documents(['the cat sat on the mat'], output_dimensionality=1024, task_type='RETRIEVAL_DOCUMENT')
  query     ctor kwargs: {'google_api_key': 'k', 'model': 'models/text-embedding-004'}
            -> .embed_query('where did the cat sit?', output_dimensionality=1024, task_type='RETRIEVAL_QUERY')

--- Voyage ---
  document  ctor kwargs: {'voyage_api_key': 'k', 'model': 'voyage-4-large', 'output_dimension': 1024}
            -> .embed_documents(['the cat sat on the mat'])
  query     ctor kwargs: {'voyage_api_key': 'k', 'model': 'voyage-4-large', 'output_dimension': 1024}
            -> .embed_query('where did the cat sit?')
```

What changed, at a glance:

- Voyage and Google now call `embed_documents` on the indexing path; previously both paths called `embed_query`.
- Google now passes `task_type="RETRIEVAL_DOCUMENT"` / `"RETRIEVAL_QUERY"` per path; previously neither.
- OpenAI's `OpenAIEmbeddings(...)` now receives `base_url=...` when the provider is configured with a custom `openai_api_base`; previously the kwarg was silently dropped.

### Docs and Changelog

`docs/developer_guides/index_managers.md` updated: class hierarchy now lists all three local subclasses; Local Index Operations example uses the new `input_type` kwarg and the actual `chunk_file(file, ...)` method (previous example referenced a non-existent `chunk_content(text=...)`).


barry47products:
@coderabbitai review

barry47products:
@snopoke  Two things came up while looking over the diff one more time, and I wanted to surface them and check what to do with each.

**Mixed embedding generations in existing Voyage/Google collections.** Once this lands, any existing collection that gains new files will hold both old and new vectors under the same `collection_id` - chunks indexed before the
merge are `embed_query` style, chunks added after are `embed_documents` style. Queries score across the whole collection, so users wouldn't see an error, just ranking that's less consistent than before (an older chunk might outrank a more relevant new one, or vice versa). There's currently no version field on `FileChunkEmbedding`, no reindex command, and nothing in the UI to flag the mixed state - same shape as the existing TODO at `apps/documents/forms.py:74`. Worth a new issue for reindex infrastructure, or is there something light enough to fold into this PR?

**Chunks from FAILED files still appear in retrieval.**
`LocalIndexManager.query` filters by `collection_id` only, not `FileStatus`. If a file fails partway through indexing, the chunks that did persist still surface in results - users see content from a file the system thinks didn't
index. Tight fix is adding `status=COMPLETED` to the query's filter chain; broader fix also deletes partial chunks on failure. New issue, or in this PR?

The NUL-only-chunk case (an all-NULs chunk becoming `""` after Postgres sanitisation, then breaking the embed call) was a special case of the second one above - folded into this PR as 1e3fd7e83 with a `logger.warning`
for operator visibility.

SmittieC:
Thanks for these great fixes!
My 2c

> **Mixed embedding generations in existing Voyage/Google collections.** Once this lands, any existing collection that gains new files will hold both old and new vectors under the same `collection_id` - chunks indexed before the merge are `embed_query` style, chunks added after are `embed_documents` style. Queries score across the whole collection, so users wouldn't see an error, just ranking that's less consistent than before (an older chunk might outrank a more relevant new one, or vice versa). There's currently no version field on `FileChunkEmbedding`, no reindex command, and nothing in the UI to flag the mixed state - same shape as the existing TODO at `apps/documents/forms.py:74`. Worth a new issue for reindex infrastructure, or is there something light enough to fold into this PR?

My first instinct is to trigger a re-index as an async task, but this will incur costs that users didn't "sign off" on, so personally I wouldn't go this route. I'm thinking, we can notify team admins who currently have indexed collections and tell them that we made changes that might improve their retrieval quality and ask them to manually trigger a reindex on those collections where they'd like to have this take an effect on. We'd have to give them this ability though.

 
> **Chunks from FAILED files still appear in retrieval.** `LocalIndexManager.query` filters by `collection_id` only, not `FileStatus`. If a file fails partway through indexing, the chunks that did persist still surface in results - users see content from a file the system thinks didn't index. Tight fix is adding `status=COMPLETED` to the query's filter chain; broader fix also deletes partial chunks on failure. New issue, or in this PR?

Adding `status=COMPLETED` will only make things worse in the worse case, since it removes candidates from the pool. Best case, it has no effect. So my 2c here is to leave it as-is. I'd leave it up to the user to retry any failed files. I do see that we don't notify users when indexing fails, which we probably should.

@snopoke, thoughts?

snopoke:
@SmittieC I agree with both of those (notifying users of errors + the need to re-index). This can be handled in a separate PR since we would also need to add the ability to reindex without re-uploading all the files.

snopoke (approved):

PR #3435, opened 2026-05-26
Port CommCare Connect to v2 channel architecture

SmittieC:
### Product Description

No user-facing change. Part of the ongoing channels refactor.

### Technical Description

**`EarlyAbort` + pipeline support**

Introduces `EarlyAbort` as a second short-circuit path in the pipeline. Unlike `EarlyExitResponse` (which still runs terminal stages to deliver a message), `EarlyAbort` halts silently — no terminal stages, no user-facing response. The right behaviour when the platform itself has withdrawn access (consent revoked, bot blocked).

**Platform consent gate (`ConsentCheckStage`)**

A new stage distinct from `ConsentFlowStage`. `ConsentFlowStage` drives the conversational consent state machine (SETUP → PENDING → ACTIVE). `ConsentCheckStage` enforces a platform-level flag in `ParticipantData.system_metadata['consent']` — a simpler binary gate managed outside the chat (e.g. Connect's key-negotiation handshake, Telegram's bot-blocked handler).

Channels opt in via `ChannelCapabilities.consent_config: ConsentConfig | None`. Two boolean knobs: `strict` (abort when no `ParticipantData` row exists) and `default_consent` (fallback when the row exists but has no `consent` key). Connect uses strict/False (must have an explicit grant); Telegram uses lenient/True (only block on explicit revocation).

This approach was chosen over a "channel provides its own stage" design because the consent logic is the same algorithm in both cases — just parameterised differently. The config dataclass keeps the variation visible and the stage testable in isolation.

**CommCare Connect v2**

`CommCareConnectChannel` and `CommCareConnectSender` are ported to `channels_v2`. The old class in `chat/channels.py` is left in place (with a TODO) while the broader refactor is in progress — both the task dispatcher and the legacy `ChannelBase.from_experiment_session()` now route to the new implementation.

A `participant_data` cached property is added to `MessageProcessingContext` for lazy DB access; stages that need it (consent check, Connect sender) read it from there rather than each fetching independently.

### Migrations

N/A

### Demo
No demo, but you can run a mock connect server with [this script](https://github.com/dimagi/open-chat-studio/blob/c9bfcf2efafa5c582a20f4e5e6ade6ee33d23eb1/scripts/mock_connect_server.py) if you want to test it.

### Docs and Changelog


scripts/mock_connect_server.py
```diff
@@ -0,0 +1,387 @@
...
+            send_user_message(channel_id, message)
+
+        elif cmd == "help":
+            print("  send <channel_id> <message>  — encrypt and send a user message")
+            print("  send <message>               — shorthand when only one channel exists")
+            print("  list                         — show known channels and key status")
+            print("  quit                         — stop the server")
+
+        else:
+            print(f"  Unknown command: {cmd!r}. Type 'help' for usage.")
```
SmittieC, line 364: Fixed in ce3c126f9. Refactored `_interactive_loop` into per-command handlers (`_handle_list_command`, `_handle_send_command`, `_handle_help_command`) plus a small dispatch dict, which collapses the complexity and nesting metrics in one go.

SmittieC, line 364: Fixed in ce3c126f9. Refactored `_interactive_loop` into per-command handlers (`_handle_list_command`, `_handle_send_command`, `_handle_help_command`) plus a small dispatch dict, which collapses the complexity and nesting metrics in one go.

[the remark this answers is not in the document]
SmittieC: Fixed in ce3c126f9. Refactored `_interactive_loop` into per-command handlers (`_handle_list_command`, `_handle_send_command`, `_handle_help_command`) plus a small dispatch dict, which collapses the complexity and nesting metrics in one go.

SmittieC, line 364: Fixed in ce3c126f9. Refactored `_interactive_loop` into per-command handlers (`_handle_list_command`, `_handle_send_command`, `_handle_help_command`) plus a small dispatch dict, which collapses the complexity and nesting metrics in one go.

apps/channels/channels_v2/connect_channel.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+
+    def __init__(self) -> None:
+        self._ctx: MessageProcessingContext | None = None
+
+    def bind(self, ctx: MessageProcessingContext) -> None:
+        self._ctx = ctx
+
+    def send_text(self, text: str, recipient: str) -> None:
+        assert self._ctx is not None, "CommCareConnectSender must be bound to a context before sending"
```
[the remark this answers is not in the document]
SmittieC: Fixed in ce3c126f9. Replaced the `assert` with an explicit `if self._ctx is None: raise ChannelException(...)`, so the guard survives `python -O`.

scripts/mock_connect_server.py
```diff
@@ -0,0 +1,387 @@
...
+
+    print(f"Mock CommCare Connect server on http://localhost:{args.port}")
+    print(f"  OCS URL   : {_config['ocs_url']}")
+    print(f"  Server ID : {_config['server_id']}")
+    print()
+    print("OCS .env settings:")
+    print(f"  COMMCARE_CONNECT_SERVER_URL=http://localhost:{args.port}")
+    print(f"  COMMCARE_CONNECT_SERVER_SECRET={_config['secret']}")
+    print(f"  COMMCARE_CONNECT_SERVER_ID={_config['server_id']}")
+    print(f"  COMMCARE_CONNECT_GET_CONNECT_ID_URL=http://localhost:{args.port}/o/userinfo/")
```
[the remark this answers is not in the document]
SmittieC: Already fixed in 808002da0 on this branch (the CodeQL Autofix commit) — the line now prints `COMMCARE_CONNECT_SERVER_SECRET=***REDACTED***` instead of the raw value. No further change needed.

apps/channels/channels_v2/pipeline.py
```diff
@@ -107,6 +109,28 @@ def __post_init__(self):
     def last_activity_at(self):
         return self.experiment_session.last_activity_at if self.experiment_session else None
 
+    @cached_property
+    def participant_data(self):
```
snopoke: This doesn't feel like it belongs here. Maybe add a `ParticipantResolverStage` which can set this on the context as well as the participant.

pyproject.toml
```diff
@@ -174,6 +174,9 @@ dev = [
     "django-types",
     "ty==0.0.38",
     "honcho>=2.0.0",
+    # Used only by scripts/mock_connect_server.py — a dev tool that impersonates
+    # the CommCare Connect backend for local development.
+    "flask",
```
snopoke: Can we add this in the script itself: https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

SmittieC: Good idea: https://github.com/dimagi/open-chat-studio/pull/3435/commits/9432f738465d68a03226839e0f7246bf19316073

apps/channels/channels_v2/stages/core.py
```diff
@@ -51,6 +57,43 @@ def process(self, ctx: MessageProcessingContext) -> None:
...
+            identifier=normalized,
+            platform=ctx.experiment_channel.platform,
+            defaults={"user": participant_user},
+        )
+        if not created and participant_user and ctx.participant.user is None:
+            ctx.participant.user = participant_user
+            ctx.participant.save()
+
+        try:
+            ctx.participant_data = ParticipantData.objects.for_experiment(ctx.experiment).get(
```
SmittieC, line 89: one downside is that this runs even though some channels never needs it. I'll think a bit how we can optimize this

apps/channels/channels_v2/pipeline.py
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/3435/changes/eea423e50e81834c1c1ca6a80bbb1d310c90a18e

apps/channels/channels_v2/stages/core.py
snopoke: I think that's an OK cost.

apps/channels/channels_v2/stages/core.py
```diff
@@ -72,17 +114,20 @@ def process(self, ctx: MessageProcessingContext) -> None:
...
 
-        # Try to load an existing active session (Issue 13: select_related)
-        ctx.experiment_session = (
-            ExperimentSession.objects.filter(
-                experiment=ctx.experiment.get_working_version(),
-                participant__identifier=str(ctx.participant_identifier),
+        # Load the existing active session using the participant FK when
+        # available. If ctx.participant is None the participant is new and
+        # no session can exist yet, so skip the query.
+        if ctx.participant is not None:
```
snopoke: since this runs after the `ParticipantResolverStage` I don't think this could be None since that stage also creates new participants. Maybe this should just be an `assert ctx.participant is not None`?

SmittieC: Yeah this docstring is outdated. I'll update

SmittieC: Good catch — confirmed the docstring was stale. `ParticipantResolverStage` uses `get_or_create` so `ctx.participant` is always set by the time this stage runs. Replaced the guard with `assert ctx.participant is not None` and updated the comment. (2fd95a8)

snopoke (dismissed):

snopoke:
tests failing:

```
=========================== short test summary info ============================
FAILED apps/channels/tests/channels/stages/test_consent_check.py::TestConsentCheckStageShouldRun::test_skips_when_no_session - AssertionError: assert True is False
 +  where True = should_run(MessageProcessingContext(message=BaseMessage(participant_id='123', message_text='Hi', content_type=<MESSAGE_TYPES.TEXT...s=[], early_exit_response=None, sending_exceptions=[], processing_errors=[], human_message_tags=[], channel_context={}))
 +    where should_run = <apps.channels.channels_v2.stages.core.ConsentCheckStage object at 0x7ff3f83608c0>.should_run
PyTorch was not found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
 +      where <apps.channels.channels_v2.stages.core.ConsentCheckStage object at 0x7ff3f83608c0> = <apps.channels.tests.channels.stages.test_consent_check.TestConsentCheckStageShouldRun object at 0x7ff421319090>.stage
```

PR #3460, opened 2026-05-28
Support inbound images and documents from WhatsApp

### LINKED ISSUE #3407: Add support for receiving images over whatsapp
We need to add support for receiving images over WhatsApp. This includes both the Meta and Twilio providers. See [this](https://github.com/dimagi/open-chat-studio/blob/a74dd6967a574d5b833f78883ac8ba2a5547c7b1/apps/experiments/tasks.py#L57-L93) for how the web channel handles this.

SmittieC:
### Product Description
Chatbots now receive inbound WhatsApp images and documents across the Meta Cloud API, Twilio, and Turn.io providers. Attachments are persisted, linked to the chat, and forwarded to the LLM. Size/format limits are left to the provider.

Resolves #3407.

### Technical Description
Core changes:
- **New `WhatsappAttachmentHydrationStage`** — WhatsApp providers deliver media as a reference (`media_id`/`media_url`), not bytes. The stage downloads media *after* session resolution (not in the webhook, to avoid blocking the provider ack and orphaning files). `WhatsappChannel` opts in via `attachment_hydration_stage_class`.
- **`MessagingService.download_message_media` / `get_inbound_media`** — implemented for Twilio, Turn.io, and Meta Cloud API.
- **Datamodel cleanup** — unified Meta/TurnIO into a single `WhatsAppMessage`; renamed `content_type_unparsed` → `attachment_mime_type` (+ new `attachment_filename`) where consumed, dropped where unused.
- **Bug fix (81baa326)** — email hydration built `Attachment` links but skipped the `ChatAttachment` row that `experiments:download_file` authorizes through, causing 404s. Moved `ChatAttachment` creation into the base class so both email and WhatsApp are covered.

### Migrations
No database migrations — new message fields are Pydantic datamodels, not Django models.

### Demo
image

image

### Docs and Changelog


apps/channels/channels_v2/stages/core.py
```diff
@@ -674,10 +675,103 @@ def should_run(self, ctx: MessageProcessingContext) -> bool:
...
+            return []
+
+        if image is None:
+            return []
+
+        raw_bytes, content_type = image
+        size = len(raw_bytes)
+        content_type = content_type.split(";", 1)[0].strip().lower()
+
+        if size > WHATSAPP_INBOUND_MAX_BYTES:
```
SmittieC: It's not necessary to check this. Meta already imposes these limits, so we can receive and store anything really

apps/channels/channels_v2/stages/core.py
```diff
@@ -674,10 +675,103 @@ def should_run(self, ctx: MessageProcessingContext) -> bool:
...
+            limit_mb = WHATSAPP_INBOUND_MAX_BYTES // MB
+            ctx.message.message_text = self._append_skip_note(
+                ctx.message.message_text, f"exceeds {limit_mb} MB image limit", size
+            )
+            return []
+
+        if content_type not in WHATSAPP_ALLOWED_IMAGE_TYPES:
```
SmittieC: same here. We can support any image type. Meta is the limiting factor. Please remove

apps/channels/tests/message_examples/meta_cloud_api_messages.py
```diff
@@ -65,3 +65,36 @@ def audio_message_value(phone_number_id="12345"):
...
+            {
+                "from": "27456897512",
+                "id": "wamid.img001",
+                "timestamp": "1706709716",
+                "type": "image",
+                "image": {
+                    "id": "image-media-id-456",
+                    "mime_type": "image/jpeg",
+                    "sha256": "def456",
+                    "caption": caption,
```
SmittieC: This payload also contains a url key. We should populate this instead of the media_id so we don't have to fetch the media url with the media id, but instead fetch the media contents directly. This is the same for Turn.io. Add a media_url to the WhatsAppMessage datastructure. Check to see if you can remove `media_id` then, but only if nothing else uses it

apps/service_providers/messaging_service.py
```diff
@@ -203,16 +222,26 @@ def send_voice_message(
         public_url = self._upload_audio_file(synthetic_voice)
         self.client.messages.create(from_=from_, to=to, media_url=[public_url])
 
-    def get_message_audio(self, message: TwilioMessage) -> BytesIO:  # ty: ignore[invalid-method-override]
+    def download_message_media(self, message: TwilioMessage) -> tuple[bytes, str]:
+        """Fetch raw bytes and content type for any inbound Twilio media (image, audio, etc.)."""
         auth = (self.account_sid, self.auth_token)
         response = httpx.get(message.media_url, auth=auth, follow_redirects=True)
+        response.raise_for_status()
+        return response.content, response.headers["Content-Type"]
```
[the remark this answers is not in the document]
SmittieC: Addressed in 17236fc00:

- **Content-Type normalization:** added `_normalize_content_type()` at the top of `messaging_service.py` and applied it to all three `download_message_media()` implementations (Twilio, Turn.io, Meta). Now strips params and lowercases (`Image/PNG; charset=binary` → `image/png`).
- **None guard:** each `download_message_media()` now raises a clear `ValueError` if both media_url and media_id are empty, before any httpx / SDK call. Tested in `Test{Twilio,Turnio,MetaCloudApi}DownloadMedia.test_download_raises_when_*`.

Also fixed a related issue from the human review: TwilioService's httpx.get was missing an explicit timeout, now set to the same `MEDIA_DOWNLOAD_TIMEOUT=30` constant.

apps/channels/channels_v2/stages/core.py
SmittieC: Removed in a2df8bf91 — dropped `WHATSAPP_INBOUND_MAX_BYTES` from `file_limits.py`, the size check from `WhatsappAttachmentHydrationStage._get_files`, and the corresponding skip-note path.

apps/channels/channels_v2/stages/core.py
SmittieC: Removed in a2df8bf91 — dropped `WHATSAPP_ALLOWED_IMAGE_TYPES` from `file_limits.py` and the MIME-allowlist check from the stage. Whatever the provider delivers gets persisted with the (now-normalized) Content-Type from the response.

apps/channels/tests/message_examples/meta_cloud_api_messages.py
SmittieC: Done in 17236fc00. Added `media_url: str | None` to `WhatsAppMessage` and `parse()` now picks it up from `message[type]["url"]` for both Meta and Turn.io. Updated the test fixtures to include a `url` key.

Both `TurnIOService.download_message_media` and `MetaCloudAPIService.download_message_media` now prefer `media_url` for a direct fetch (Turn.io uses httpx with a Bearer header; Meta uses the existing auth headers). They fall back to the previous `media_id` path only when no URL is in the payload — coverage added in `Test{Turnio,MetaCloudApi}DownloadMedia.test_download_falls_back_to_media_id_*`.

I kept `media_id` on the data class because `get_message_audio` for both providers still resolves audio via media_id-style paths and because the fallback gives us defense-in-depth for older webhook payload shapes. Happy to remove it in a follow-up if you'd prefer that everything go through media_url.

apps/channels/datamodels.py
```diff
@@ -141,6 +140,13 @@ def determine_content_type(cls, value):
             return MESSAGE_TYPES.TEXT
         if value and value in ["audio/ogg", "video/mp4"]:
             return MESSAGE_TYPES.VOICE
+        if value and value.startswith("audio/"):
```
SmittieC: this should be an OR case in the check above

apps/channels/channels_v2/stages/core.py
```diff
@@ -674,10 +674,99 @@ def should_run(self, ctx: MessageProcessingContext) -> bool:
...
+    _get_files() downloads the media via the messaging service and persists
+    the bytes as a MESSAGE_MEDIA File. The base class then handles
+    ChatAttachment linkage and Attachment construction. Size and
+    content-type policing is the upstream provider's responsibility —
+    Meta already caps what reaches us.
+    """
+
+    # Excluded MIME markers — voice/audio is handled via the audio transcription path,
+    # never persisted as an inbound media attachment.
+    _VOICE_MIMES = {"audio", "voice"}
```
SmittieC: Twilio sends a `MessageType` which is either `audio, image, text or document`. Rather check that key instead of `MediaContentType0` for the content type. If it is audio, make it "voice" so that it is the same as the whatsapp message datastructure. (this is a datastructure update)

apps/service_providers/messaging_service.py
```diff
@@ -203,16 +245,28 @@ def send_voice_message(
...
+            raise ValueError("Cannot download Twilio media: message.media_url is empty")
         auth = (self.account_sid, self.auth_token)
-        response = httpx.get(message.media_url, auth=auth, follow_redirects=True)
+        response = httpx.get(message.media_url, auth=auth, follow_redirects=True, timeout=MEDIA_DOWNLOAD_TIMEOUT)
+        response.raise_for_status()
+        return response.content, _normalize_content_type(response.headers.get("Content-Type"))
+
+    def get_inbound_media(self, message: TwilioMessage) -> tuple[bytes, str] | None:
+        mime = message.attachment_mime_type
+        if not mime or not message.media_url or _is_voice_mime(mime):
```
SmittieC: add a comment on why `_is_voice_mime` is excluded here.

apps/service_providers/messaging_service.py
```diff
@@ -203,16 +245,28 @@ def send_voice_message(
...
+        """Fetch raw bytes and content type for any inbound Twilio media (image, audio, etc.)."""
+        if not message.media_url:
+            raise ValueError("Cannot download Twilio media: message.media_url is empty")
         auth = (self.account_sid, self.auth_token)
-        response = httpx.get(message.media_url, auth=auth, follow_redirects=True)
+        response = httpx.get(message.media_url, auth=auth, follow_redirects=True, timeout=MEDIA_DOWNLOAD_TIMEOUT)
+        response.raise_for_status()
+        return response.content, _normalize_content_type(response.headers.get("Content-Type"))
+
+    def get_inbound_media(self, message: TwilioMessage) -> tuple[bytes, str] | None:
```
SmittieC: Why doesn't this also calls `_has_downloadable_attachment` like `TurnIOService`?

apps/channels/tests/message_examples/turnio_messages.py
```diff
@@ -1,36 +1,59 @@
...
+            "in_reply_to": "None",
+            "inserted_at": inserted_at,
+            "labels": [],
+            "last_status": "None",
+            "last_status_timestamp": "None",
+            "on_fallback_channel": False,
+            "rendered_content": "None",
+            "uuid": message_uuid,
+        }
+    }
```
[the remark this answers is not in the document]
SmittieC: The `_vnd` helper is a test fixture builder — the kwargs are explicit on purpose so each call site reads as a labeled record (the UUIDs/timestamps are all UUID-shaped strings, so positional args would be unreadable). Suppressing this one; the production call paths don't have this issue.

apps/channels/datamodels.py
SmittieC: Addressed in f4c6b84e9 — the MIME-based validator is gone entirely. `TwilioMessage` now reads Twilio's semantic `MessageType` field directly (text/image/audio/document) and the validator collapses to a small category match. See the reply on the line-723 thread for the broader datamodel change.

apps/channels/channels_v2/stages/core.py
SmittieC: Done in f4c6b84e9. `TwilioMessage.parse()` now reads `message_data["MessageType"]` instead of `MediaContentType0` and maps `audio` → `voice` so the marker matches the WhatsApp datamodel. The `determine_content_type` validator was simplified to a small category match (`voice` → VOICE, `text`/`image`/`document` → TEXT). `attachment_mime_type` still holds the raw `MediaContentType0` MIME for the persistence helper. Test fixtures (`twilio_messages.py`) updated to include `MessageType`; full Twilio/WA/Facebook test sweep passes.

apps/service_providers/messaging_service.py
SmittieC: Done in f4c6b84e9 — but as part of also addressing thread #3322734209, the manual mime/url/voice check was replaced with `_has_downloadable_attachment(message)`, and the comment now sits above that call explaining the voice exclusion.

apps/service_providers/messaging_service.py
SmittieC: Good catch — fixed in f4c6b84e9. `TwilioService.get_inbound_media` now uses the same `_has_downloadable_attachment` helper as TurnIOService and MetaCloudAPIService, with a comment explaining why voice/audio attachments skip the inbound-media path (they're transcribed via `get_message_audio`).

SmittieC:
@coderabbitai full review

apps/channels/channels_v2/stages/core.py
```diff
@@ -674,10 +674,96 @@ def should_run(self, ctx: MessageProcessingContext) -> bool:
...
+            )
+        )
+
+
+# ---------------------------------------------------------------------------
+# WhatsappAttachmentHydrationStage
+# ---------------------------------------------------------------------------
+
+
+class WhatsappAttachmentHydrationStage(AttachmentHydrationStage):
```
snopoke: should this be moved to the `whatsapp_channel.py` module?

apps/channels/datamodels.py
```diff
@@ -180,59 +186,53 @@ def parse(message_data: dict) -> "SureAdhereMessage":
...
-            body = message["text"]["body"]
+        match value:
+            case "audio":
+                return MESSAGE_TYPES.VOICE
+            case "image" | "document":
+                # Treat as a text message with an attachment; the caption becomes the message text.
+                # The actual MIME type is preserved in attachment_mime_type for the persistence helper.
+                return MESSAGE_TYPES.TEXT
+            case _ if MESSAGE_TYPES.is_member(value):
+                return MESSAGE_TYPES(value)
```
snopoke, line 210: should this return `MESSAGE_TYPES.OTHER` for non-member values?

apps/service_providers/messaging_service.py
```diff
@@ -293,22 +348,44 @@ def send_voice_message(
...
+            raise ValueError("Cannot download Turn.io media: both media_url and media_id are empty")
+        response.raise_for_status()
+        return response.content, _normalize_content_type(response.headers.get("Content-Type"))
+
+    def get_inbound_media(self, message: WhatsAppMessage) -> tuple[bytes, str] | None:
+        if not _has_downloadable_attachment(message):
+            return None
+        return self.download_message_media(message)
 
+    def get_message_audio(self, message: WhatsAppMessage) -> BytesIO:  # ty: ignore[invalid-method-override]
```
snopoke: this method and the one on the twilio service seem almost identical. Could they be yanked to a mixing or something?

apps/service_providers/messaging_service.py
```diff
@@ -293,22 +348,44 @@ def send_voice_message(
...
-        data = BytesIO(response.content)
-        content_type = response.headers["Content-Type"]
+        data = BytesIO(raw_bytes)
         message.cached_media_data = MediaCache(content_type=content_type, data=data)
 
         # Example header: {'Content-Type': 'audio/ogg'}
         family, sub_type = content_type.split("/", 1)
         if family != "audio":
-            raise AudioConversionError(f"Unexpected content-type for audio: {content_type}")
+            raise MessageMediaError(f"Unexpected content-type for audio: {content_type}")
```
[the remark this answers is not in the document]
snopoke: +1 inconsistent error handling between the services

apps/service_providers/messaging_service.py
```diff
@@ -604,26 +681,44 @@ def send_file_to_user(
...
         response = httpx.get(
             media_url,
             headers=self._headers,
             follow_redirects=True,
             timeout=self.META_API_TIMEOUT,
         )
+        response.raise_for_status()
+        return response.content, _normalize_content_type(response.headers.get("Content-Type"))
+
+    def get_inbound_media(self, message: WhatsAppMessage) -> tuple[bytes, str] | None:
```
snopoke: also, duplicate method with twilio service

apps/service_providers/messaging_service.py
```diff
@@ -604,26 +681,44 @@ def send_file_to_user(
...
+        if not _has_downloadable_attachment(message):
+            return None
+        return self.download_message_media(message)
+
+    def get_message_audio(self, message: WhatsAppMessage) -> BytesIO:  # ty: ignore[invalid-method-override]
```
snopoke: another duplicate

docs/adr/0029-download-whatsapp-inbound-attachments-in-hydration-stage.md
snopoke, line 1: This ADR will need to be re-numbered

snopoke (commented):
Mostly minor stuff. Nice cleanup on the message models.

apps/channels/datamodels.py
SmittieC: Yes. Good catch

apps/channels/channels_v2/stages/core.py
SmittieC: Fixed in 4ffc92981. Moved `WhatsappAttachmentHydrationStage` out of `stages/core.py` and into `whatsapp_channel.py`, co-located with the channel that wires it (`attachment_hydration_stage_class`). The generic `AttachmentHydrationStage` base stays in `core.py`. Updated the two test imports accordingly.

*— Claude Code*

apps/channels/datamodels.py
SmittieC: Fixed in 4ffc92981. Added a `case _:` fallback returning `MESSAGE_TYPES.OTHER`, matching `TwilioMessage.determine_content_type`, so unknown provider types (sticker, location, …) never fall through to `None`.

*— Claude Code*

apps/service_providers/messaging_service.py
SmittieC: Fixed in 4ffc92981. Extracted the duplicated `get_inbound_media` (and `get_message_audio`) into a new `HttpMediaDownloadMixin`, now shared by `TwilioService`, `TurnIOService`, and `MetaCloudAPIService`. Each service keeps only its provider-specific `download_message_media`.

*— Claude Code*

apps/service_providers/messaging_service.py
SmittieC: Fixed in 4ffc92981. `get_message_audio` is now provided by the shared `HttpMediaDownloadMixin`. While consolidating, the shared handler now catches both `requests.HTTPError` and `httpx.HTTPStatusError`, fixing the prior inconsistency where Turn.io only caught `requests.HTTPError`.

*— Claude Code*

docs/adr/0029-download-whatsapp-inbound-attachments-in-hydration-stage.md
SmittieC: Fixed in 4ffc92981. The file was renamed to `0018-…` but three references still said `0015`: the in-file heading, the `index.md` href, and the mkdocs nav entry. All now point to 0018 — this was also the cause of the failing `test_adr_heading_matches_filename` CI check.

*— Claude Code*

apps/channels/channels_v2/stages/core.py
```diff
@@ -674,10 +674,96 @@ def should_run(self, ctx: MessageProcessingContext) -> bool:
...
+
+    def _get_files(self, ctx: MessageProcessingContext) -> list[File]:
+        """Return the Files to hydrate. Default impl resolves pre-persisted Files
+        by ``ctx.message.attachment_file_ids``. Subclasses can override to acquire
+        files differently (e.g. download from a remote channel)."""
+        return list(
+            File.objects.filter(
+                id__in=ctx.message.attachment_file_ids,
+                team_id=ctx.experiment.team_id,
+            )
```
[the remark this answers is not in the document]
SmittieC: Fixed in 4ffc92981. `_get_files` now guards `ctx.experiment_channel.messaging_provider` for `None` (it is nullable on `ExperimentChannel`): if absent it logs a warning and returns `[]` instead of raising `AttributeError`.

*— Claude Code*

apps/service_providers/messaging_service.py
SmittieC: Fixed in 4ffc92981. Yanked into `HttpMediaDownloadMixin` along with the two duplicate methods flagged on the Meta service — Twilio/Turn.io/Meta now share one implementation.

*— Claude Code*

apps/channels/datamodels.py
```diff
@@ -161,9 +167,9 @@ def parse(message_data: dict) -> "TwilioMessage":
             participant_id=message_data["From"].removeprefix(prefix_to_remove),
             to=to,
             message_text=message_data["Body"],
-            content_type=content_type,
+            content_type=message_type,
```
SmittieC, line 170: `content_type` here is a misnomer. It should really be `message_type`. In a separate PR I want to fix this.

snopoke (approved):

PR #3466, opened 2026-05-28
Add Undo Applied Tags button to evaluation runs

### LINKED ISSUE #3439: Ability to undo a run's applied tags
Followup PR on https://github.com/dimagi/open-chat-studio/pull/3414.

We need to give users the ability to undo a specific run's applied tags. We cannot just iterate over the `AppliedTag` set of that run and remove all tags, since the previous run might have applied one of those tags as well. We need to revert the diff between the current and previous run. The way we can do this is to remove all tags applied by the current run, and apply the previous run's tags. This follows similar logic to what the `reverse_stale_tags` method does

SmittieC:
Resolves #3439

### Product Description
Adds an **Undo Applied Tags** button to the evaluation run detail page. If a run produces bad tag changes (e.g. a misconfigured evaluator), users can revert that run's tags and restore the previous completed run's tags in one click.

### Technical Description
Core logic: `apps/evaluations/tagging.py::undo_run_tags`. For each target in scope, it removes the managed tags this run applied (per `AppliedTag` audit records) and re-applies tags from the most recent prior `COMPLETED` `FULL`/`DELTA` run on the same config. Audit rows are preserved — only live `CustomTaggedItem` state is mutated.

Worth flagging for review:
- **Not wrapped in `transaction.atomic()`** (see docstring): a mid-undo failure leaves a partial state that the next full run or re-run reconciles. Happy to make it fail-closed if preferred.
- **Only `COMPLETED` runs** are eligible; PREVIEW runs are skipped/hidden. If no prior run exists, current tags are removed with nothing put back.

The view `undo_evaluation_run_tags` is `POST`-only, CSRF-protected, and gated on `evaluations.change_evaluationrun`.

### Migrations
N/A — no schema changes.

### Demo
https://www.loom.com/share/d236949d69514ecaae6f54e4e2e41aa1

### Docs and Changelog


apps/evaluations/tagging.py
```diff
@@ -206,3 +213,177 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+            run__config=run.config,
+            run__status=EvaluationRunStatus.COMPLETED,
+            run__type__in=[EvaluationRunType.FULL, EvaluationRunType.DELTA],
+            run__created_at__lt=run.created_at,
+            message_id__in=message_ids,
+        )
+        .order_by("message_id", "-run__created_at")
+        .distinct("message_id")
+        .values_list("message_id", "run_id")
+    )
```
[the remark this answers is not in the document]
SmittieC: Good catch — addressed in 41b8c5343. Both concerns were valid:

1. **Tie-breaker**: added `-run_id` to `order_by` so `DISTINCT ON (message_id)` resolves deterministically when two prior runs share a `finished_at`.
2. **`finished_at` over `created_at`**: switched both the filter and ordering to `run__finished_at` (set in `mark_complete()`). Since runs for the same config can overlap (no concurrency guard), completion order — not creation order — reflects which prior tag state is most recent. The undo view guards `status == COMPLETED`, and `mark_complete()` sets `finished_at` + `status` together, so `run.finished_at` and all candidates' `finished_at` are guaranteed non-null here.

Added two regression tests: `test_orders_by_finished_at_not_created_at` (overlapping runs — later-created/earlier-finished run loses) and `test_tie_broken_deterministically_by_run_id`.

apps/evaluations/views/evaluation_config_views.py
```diff
@@ -611,3 +613,30 @@ def generate_evaluation_results_column_suggestions(result_columns, evaluation_ru
...
     return suggestions
+
+
+@login_and_team_required
+@require_POST
+@permission_required("evaluations.change_evaluationrun")
+def undo_evaluation_run_tags(request, team_slug: str, evaluation_pk: int, evaluation_run_pk: int):
+    """Remove all tags applied by this run and restore the previous run's tags."""
+    evaluation_run = get_object_or_404(
+        EvaluationRun, id=evaluation_run_pk, config_id=evaluation_pk, team__slug=team_slug
```
snopoke: rather do `team=request.team` to avoid having to join on the team table to match the slug:
```suggestion
        EvaluationRun, id=evaluation_run_pk, config_id=evaluation_pk, team=request.team
```

apps/evaluations/tagging.py
```diff
@@ -206,3 +213,184 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+    runs_to_message_ids: defaultdict[int, set[int]] = defaultdict(set)
+    for message_id, run_id in qs.iterator():
+        runs_to_message_ids[run_id].add(message_id)
+    return runs_to_message_ids
+
+
+def _previous_applied_by_message(run: EvaluationRun, message_ids: list[int]) -> defaultdict[int, set[int]]:
```
snopoke: type should just be `dict`. The fact that it's a `defaultdict` isn't consequential to the callers

```suggestion
def _previous_applied_by_message(run: EvaluationRun, message_ids: list[int]) -> dict[int, set[int]]:
```

apps/evaluations/tagging.py
```diff
@@ -206,3 +213,184 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+
+    return applied
+
+
+def _compute_undo_target_diffs(
+    run: EvaluationRun,
+    evaluated_message_ids: list[int],
+    possible_tags: frozenset[int],
+    representative_evaluator: Evaluator,
+) -> tuple[defaultdict[int, set[int]], defaultdict[int, set[int]], ContentType | None]:
```
snopoke:
```suggestion
) -> tuple[dict[int, set[int]], dict[int, set[int]], ContentType | None]:
```

apps/evaluations/tagging.py
```diff
@@ -206,3 +213,184 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+      predecessor that did not touch a given message is skipped, and we look
+      further back until a prior run that did evaluate the message is found.
+
+    PREVIEW runs are skipped entirely.
+
+    The set of messages to process is sourced from this run's EvaluationResult rows,
+    not the live dataset — so later edits to the dataset do not affect what undo
+    touches.
+
+    AppliedTag audit rows are never deleted — they remain as history. Only
```
snopoke, line 373: Should we mark the `AppliedTag` as 'archived' or something? Just thinking about tracing from a tag back to evals - we wouldn't know which run added the tag unless we know when tags are 'unapplied'.

apps/evaluations/tagging.py
```diff
@@ -206,3 +213,184 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+        "session__chat", "expected_output_chat_message"
+    )
+    for message in messages_iter:
+        target = resolve_target(representative_evaluator, message)
+        if target is None:
+            continue
+        if content_type is None:
+            content_type = ContentType.objects.get_for_model(type(target))
+        remove_by_target[target.pk] |= current_applied[message.pk] & possible_tags
+        add_by_target[target.pk] |= previous_applied[message.pk] & possible_tags
```
snopoke: Could we diff these to avoid delete and reapply:

```suggestion
        current_tags = current_applied[message.pk] & possible_tags
        previous_tags = previous_applied[message.pk] & possible_tags
        remove_by_target[target.pk] |= current_tags - previous_tags
        add_by_target[target.pk] |= previous_tags - current_tags
```

templates/evaluations/evaluation_result_home.html
```diff
@@ -43,6 +43,25 @@ <h1 class="text-2xl font-bold">{{ title }}</h1>
             <i class="fa-solid fa-upload"></i>
             Edit Results
           </a>
+          {% if evaluation_run.type != "preview" %}
+            <form id="undo-tags-form-{{ evaluation_run.id }}"
+                  style="display: contents;"
+                  method="post"
+                  action="{% url 'evaluations:evaluation_run_undo_tags' request.team.slug evaluation_run.config_id evaluation_run.id %}">
+              {% csrf_token %}
+              <button type="button"
```
snopoke, line 52: this could do with a 'processing' indicator + `hx-disabled-elt="this"` to disable the button

apps/evaluations/tagging.py
```diff
@@ -206,3 +213,184 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+    messages_iter = EvaluationMessage.objects.filter(pk__in=evaluated_message_ids).select_related(
+        "session__chat", "expected_output_chat_message"
+    )
+    for message in messages_iter:
+        target = resolve_target(representative_evaluator, message)
+        if target is None:
+            continue
+        if content_type is None:
+            content_type = ContentType.objects.get_for_model(type(target))
+        remove_by_target[target.pk] |= current_applied[message.pk] & possible_tags
```
snopoke: Why do we need use `defaultdict` and `|=` these? I would have expected that `messages_iter` has unique references in it. Though maybe that isn't enforced anywhere?

Update: claude confirmed that CSV imports can result in duplicate session / message references. I might try and address that separately

apps/evaluations/tests/test_tagging_integration.py
```diff
@@ -501,3 +507,221 @@ def test_missing_value_reports_error(self, team, message_evaluator):
...
+        )
+        result = EvaluationResultFactory.create(
+            team=team,
+            evaluator=rule_that_fires.evaluator,
+            message=message,
+            run=run,
+            output=output,
+        )
+        apply_rules_to_result(result, rule_that_fires.evaluator, message)
+        return run
```
SmittieC: Fixed in e08051f82. Moved `TestUndoRunTags` into the new `test_undo_run_tags.py` and dropped the `team` argument from `_make_run` (now derived from `config.team`), bringing it to 4 args.

*— Claude Code*

apps/evaluations/tests/test_tagging_integration.py
```diff
@@ -501,3 +507,472 @@ def test_missing_value_reports_error(self, team, message_evaluator):
...
+        dataset = EvaluationDatasetFactory.create(team=team, messages=[])
+        return EvaluationConfigFactory.create(team=team, dataset=dataset, evaluators=[evaluator]), evaluator
+
+    def _make_run_with_results(self, team, config, evaluator, messages, run_type=EvaluationRunType.FULL):
+        run = EvaluationRunFactory.create(team=team, config=config, status=EvaluationRunStatus.COMPLETED, type=run_type)
+        for message in messages:
+            EvaluationResultFactory.create(
+                team=team, evaluator=evaluator, message=message, run=run, output={"result": {}}
+            )
+        return run
```
SmittieC: Fixed in e08051f82. `_make_run_with_results` now derives `team` from `config.team`, reducing it to 4 args (also moved to `test_undo_run_tags.py`).

*— Claude Code*

apps/evaluations/tests/test_tagging_integration.py
```diff
@@ -1,23 +1,29 @@
 """DB integration tests for eval-driven tagging."""
 
+from datetime import timedelta
```
SmittieC: Fixed in e08051f82. Split the undo-specific tests (`TestUndoRunTags`, `TestMessageIdsByLatestPriorRun`) out into a dedicated `test_undo_run_tags.py`; `test_tagging_integration.py` is now ~500 lines and the new module is also under the threshold.

*— Claude Code*

apps/evaluations/views/evaluation_config_views.py
SmittieC: Fixed in f7c9d48e1. Switched to `team=request.team`.

*— Claude Code*

apps/evaluations/tagging.py
SmittieC: Fixed in f7c9d48e1. Return annotation is now `dict[int, set[int]]`.

*— Claude Code*

apps/evaluations/tagging.py
SmittieC: Fixed in f7c9d48e1. Return annotation is now `tuple[dict[int, set[int]], dict[int, set[int]], ContentType | None]` (and `_apply_undo_target_diffs` params updated to match).

*— Claude Code*

apps/evaluations/tagging.py
SmittieC: Good point on traceability. The `AppliedTag` audit rows are deliberately never deleted by undo — they remain as history, so you can still trace which run *applied* a given tag. What we do not yet record is the *unapply* event, so you cannot tell from the data that a tag was later undone. Adding an explicit archived/unapplied marker is a schema + write-path change that is broader than this PR, so I would prefer to track it as a separate follow-up rather than fold it in here. Happy to open an issue.

*— Claude Code*

apps/evaluations/tagging.py
SmittieC: Acknowledged — keeping `defaultdict` + `|=` since, as you confirmed, CSV imports can produce duplicate session/message references so a given target can appear more than once. Agreed that de-duplicating those references is best handled separately.

*— Claude Code*

apps/evaluations/tagging.py
SmittieC: Fixed in f7c9d48e1. Now diffing: `remove = current_tags - previous_tags`, `add = previous_tags - current_tags`, so tags common to both runs are left untouched instead of being deleted and recreated.

*— Claude Code*

templates/evaluations/evaluation_result_home.html
SmittieC: Fixed in 7ad533a93. Note this form is a plain POST that redirects (not an htmx request), so `hx-disabled-elt` does not apply directly. Instead, the confirm `onok` callback disables the button and swaps its label for a `loading-spinner` before submitting, which gives the same processing / no-double-submit behaviour.

*— Claude Code*

apps/evaluations/tagging.py
```diff
@@ -206,3 +213,189 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+            continue
+        if content_type is None:
+            content_type = ContentType.objects.get_for_model(type(target))
+        # Diff current vs previous so we only touch tags that actually change:
+        # drop tags this run added that the prior run didn't have, and restore
+        # prior tags this run no longer has. Tags common to both are left alone.
+        current_tags = current_applied[message.pk] & possible_tags
+        previous_tags = previous_applied[message.pk] & possible_tags
+        remove_by_target[target.pk] |= current_tags - previous_tags
+        add_by_target[target.pk] |= previous_tags - current_tags
```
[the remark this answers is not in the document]
SmittieC: Not a valid bug given the current data model — the premise that "multiple messages can share the same `Chat` target in SESSION mode" doesn't hold.

In SESSION mode there's a strict 1:1 mapping between `EvaluationMessage` and `Chat` target, so the per-message diff never aggregates two messages onto the same `target.pk`:

1. **One message per session.** `make_session_evaluation_messages` (`apps/evaluations/utils.py:217`) collapses an entire conversation into a *single* `EvaluationMessage` with the full history. Locked in by tests (`test_happy_path_multi_turn_session`: 4 turns → `len(result) == 1`; `test_multiple_sessions`: one message each).
2. **No duplicate sessions per dataset.** Both clone paths dedupe by `session_id` before insert (`apps/evaluations/tasks.py:961-967`), confirmed by `TestSessionModeDuplicateDetection`.
3. `resolve_target` maps a session-mode message to `session.chat` (`tagging.py:80-84`), and `_compute_undo_target_diffs` iterates *distinct* `evaluated_message_ids` (`tagging.py:386`).

So every `target.pk` is touched by exactly one message; the `|=` union degenerates to assignment and `current - previous` / `previous - current` is exactly the correct per-target diff. No preserved tag gets removed.

**One legitimate point:** correctness silently depends on this 1:1 invariant. If SESSION mode ever switched to per-turn messages (or a race bypassed the dedupe), the described data loss would become real. The Sentry-suggested fix (aggregate current/previous per target, then diff once) is behavior-preserving today and future-proofs it — worth doing as a low-priority robustness hardening, but it's not fixing a live bug. The same pattern in `_compute_stale_by_target` (`tagging.py:176-178`) relies on the identical invariant.

apps/evaluations/tagging.py
snopoke: OK, please open an issue.

apps/evaluations/tagging.py
snopoke: FYI https://github.com/dimagi/open-chat-studio/pull/3486

apps/evaluations/tagging.py
SmittieC: @coderabbitai see the comments and please open an issue for this. Assign it to me

SmittieC: @snopoke I'm thinking, would it make sense to want to undo applied tags for a run that is not  the latest run, especially since those tags are wiped before the latest run anyways. What you're essentially then just doing is tagging the results with a specific run's tags, which is a weird feature in my opinion (why not just fix the evaluator instead?).

I think it only makes sense to be able to undo the latest run's tags to the run before that. In this case, we can archive the `AppliedTag` set for the latest run as a way to keep track of which run's tags are currently being applied.

So it will work like this:
Run 1 applies tags. Run 2 finished, so run 1's applied tags (AppliedTag records) are archived. When run 3 finishes, run 2's applied tags are archived. I can only undo run 3's applied tags, so it becomes archived and run 2's tags are now "active", but I cannot then undo run 2's applied tags as well. If I have to undo a bunch of runs to get my tagging right, I should probably just tweak my evaluator instead.

snopoke (approved):

SmittieC:
@snopoke I think I'm rather going to close this PR and use [this one](https://github.com/dimagi/open-chat-studio/pull/3517), since it contains some important design changes

SmittieC:
Superseded by https://github.com/dimagi/open-chat-studio/pull/3517

PR #3474, opened 2026-05-28
Remove unused Experiment fields: temperature, tools, citations_enabled

### LINKED ISSUE #3467: Remove unused Experiment fields: temperature, tools, citations_enabled
## Goal

Remove three fields from the `Experiment` model that are no longer read at runtime. This is part of the ongoing experiment → chatbot cleanup. All chat behavior is now driven by Pipeline node parameters; these Experiment-level fields are dead.

This PR is **state-only** — the actual DB columns get dropped in a follow-up issue once this and the sibling cleanup PRs are deployed.

## Scope

| Field | Why safe |
|---|---|
| `temperature` | Only referenced by the prompt builder UI's client-side Alpine state and a few test factory kwargs. Never read at chat runtime — pipelines carry their own temperature in `llm_model_parameters`. |
| `tools` | `ArrayField` populated by migration `0080` but never read. The chat tool wiring uses `assistant.tools` and pipeline node params, never `experiment.tools`. |
| `citations_enabled` | `BooleanField` added in migration `0099` and never read. The pipeline `AssistantNode` has its own `citations_enabled` param that stays. |

## Tasks

- [ ] Remove the three field definitions from `apps/experiments/models.py` (lines 559, 620, 647).
- [ ] Remove `"temperature"` from `EXPERIMENT_FIELDS` in `apps/experiments/model_audit_fields.py`.
- [ ] Drop the `temperature=...` kwargs from `ExperimentFactory.create/build` calls in `apps/experiments/tests/test_versioning.py` (lines 15, 26-27, 51, 65, 86).
- [ ] Add a `SeparateDatabaseAndState` migration that removes the three fields from Django state but leaves the DB columns alone (mirrors `0130`). The matching `DROP COLUMN` SQL ships in a separate follow-up PR.
- [ ] Run `uv run pytest apps/experiments/tests/ -v`, `uv run ruff check apps/ --fix`, `uv run ty check apps/`.

## Notes

- Why two PRs: production runs the previous code revision during the rolling deploy. Dropping the columns in the same PR that removes the model fields would break the old workers that still SELECT them. Shipping state-only first, then the column drop in a separate deploy, keeps the rolling window safe.
- Existing django-auditlog history rows referencing these fields stay in place (matches `llm_provider` removal pattern).
- The pipeline-node `citations_enabled` field at `apps/pipelines/nodes/nodes.py:860` is unrelated and stays.

snopoke:
### Product Description
Internal cleanup — no user-facing change. Removes three dead fields from the `Experiment` model (`temperature`, `tools`, `citations_enabled`) that are no longer read at chat runtime — pipelines carry their own equivalents.

### Technical Description
- Removed the three field definitions from `apps/experiments/models.py` and cleaned up the now-unused validator imports.
- Removed `"temperature"` from `EXPERIMENT_FIELDS` in `apps/experiments/model_audit_fields.py`.
- Updated `apps/experiments/tests/test_versioning.py` to use `seed_message` in place of `temperature` as the test value — the tests were exercising `compare_with_model` / `VersionField` mechanics, not the temperature field specifically.
- Added migration `0136_remove_experiment_temperature_and_more.py` that (1) sets DB-level `DEFAULT` values on the three columns via `RunSQL`, then (2) uses `SeparateDatabaseAndState` to remove the fields from Django state while leaving the DB columns in place. The DB defaults are required so that INSERTs from the new code (which no longer references these columns) satisfy the existing NOT NULL constraints during the rolling deploy window.

### Migrations
- [x] The migrations are backwards compatible — state-only removal, DB columns remain. A follow-up PR will ship the `DROP COLUMN` SQL.

### Demo
N/A — internal cleanup.

### Docs and Changelog
- No docs update required.

Fixes #3467


snopoke:
@coderabbit-ai review

SmittieC (approved):

PR #3483, opened 2026-05-29
Remove Experiment.source_material FK (state-only)

### LINKED ISSUE #3470: Remove Experiment.source_material FK
## Goal

Remove the `source_material` FK from the `Experiment` model. Part of the experiment → chatbot cleanup — pipeline nodes carry their own `source_material_id` (`apps/pipelines/nodes/nodes.py:265`, `apps/pipelines/nodes/llm_node.py:151`). The `SourceMaterial` model itself stays; only the Experiment-level FK goes away.

This PR is **state-only** — the actual DB column and FK constraint get dropped in a follow-up issue once this and the sibling cleanup PRs are deployed.

## Background

- Not read at chat runtime. `PipelineBot.process_input` doesn't touch it; pipeline LLM nodes pass their own `source_material_id` into `PromptTemplateContext`.
- Still wired into versioning, admin, clone-team, and test fixtures — all of which need to be untangled.

## Tasks

- [ ] Remove `source_material` field from `apps/experiments/models.py` (lines 569-575).
- [ ] Remove the `self._copy_attr_to_new_version("source_material", new_version)` call from `Experiment.create_new_version` (`apps/experiments/models.py:916`).
- [ ] Remove the `self.experiment_set.update(source_material=None, audit_action=AuditAction.AUDIT)` call on the `SourceMaterial` model (`apps/experiments/models.py:187`).
- [ ] Remove `"source_material"` from `EXPERIMENT_FIELDS` in `apps/experiments/model_audit_fields.py`.
- [ ] Drop `"source_material"` from `list_display` and `list_filter` in `apps/experiments/admin.py` (lines 94, 99).
- [ ] Drop the source_material remap block in `apps/teams/management/commands/clone_team.py:432-438`. The `SourceMaterial` row copying earlier in `clone_team.py` for pipeline node references stays.
- [ ] If `prompt_builder_load_source_material` view reads via `experiment.source_material`, switch it to query `SourceMaterial.objects.filter(team=...)` directly.
- [ ] Delete `apps/experiments/tests/test_source_material_views.py` if its assertions depend on the Experiment FK; otherwise update it. Prune the source_material versioning cases in `apps/experiments/tests/test_models.py:639, 700-763` and `apps/experiments/tests/test_versioning.py:311`.
- [ ] Add a `SeparateDatabaseAndState` migration removing the field from Django state. The DB column and FK constraint remain in place — they will be dropped in the follow-up PR.
- [ ] Run `uv run pytest apps/experiments/tests/ apps/teams/tests/test_clone_team.py -v`, `uv run ruff check apps/ --fix`, `uv run ty check apps/`.

## Why two PRs

Production runs the previous code revision during the rolling deploy. Dropping the column in the same PR that removes the model field would break the old workers that still SELECT `source_material_id`. Shipping state-only first, then the column drop in a separate deploy, keeps the rolling window safe.

## Out of scope

- The `SourceMaterial` model itself, its CRUD views, and its use by pipeline nodes — all stay.
- Migration of existing `experiment.source_material_id` values into the corresponding pipeline node params — out of scope; chatbots that need source material already wire it on the pipeline node side.

snopoke:
### Product Description
Internal cleanup: removes the unused `source_material` FK from the `Experiment` model as part of the experiment → chatbot cleanup. No user-visible behaviour change — pipeline LLM nodes already carry their own `source_material_id` and read it at runtime.

### Technical Description
- Removes `Experiment.source_material` FK at the Django model level only — the DB column and FK constraint stay in place so the previous code revision can keep running during the rolling deploy.
- Untangles versioning (`create_new_version`), `model_audit_fields`, admin, clone-team remap, and the `SourceMaterial.archive` cascade null.
- Prunes source_material versioning test cases; switches `test_copy_attr_to_new_version` to exercise `consent_form`.
- Adds a `SeparateDatabaseAndState` migration (0137) removing the field from state with no DB operations.

The DB column & FK constraint will be dropped in a follow-up PR after this is deployed.

Closes #3470

### Migrations

### Demo
N/A — internal cleanup, no UI change.

### Docs and Changelog


snopoke:
@coderabbitai review

PR #3486, opened 2026-05-29
fix(evaluations): centralise dataset message dedup to prevent duplicate references

snopoke:
### Product Description
no change

### Technical Description
Datasets could end up with two `EvaluationMessage` rows referencing the same session or chat-message pair. Dedup logic was duplicated across the five population paths and applied inconsistently — and the message-mode clone path deduped in memory with **no transaction or row lock**, so two concurrent clones on the same dataset could each read "not a duplicate" and both insert the same pair. (A production scan turned up exactly one such duplicate.)

All paths now route through a single `EvaluationDataset.add_messages()`:
- Locks the dataset row (`select_for_update`) for the duration — this is the actual race fix.
- Dedups by mode: `(input_chat_message_id, expected_output_chat_message_id)` pair for message mode, `session_id` for session mode; against both existing rows and within the incoming batch.
- Messages with null FKs (manual / CSV) are never deduped, so those paths keep their current behaviour.

Worth a reviewer's eye: message-mode dedup stays **pair-based**, so the same input chat message paired with a *different* expected output is intentionally still allowed (distinct eval). The existing single duplicate in production is **not** cleaned up by this PR — it only prevents new ones.

### Migrations
N/A — no schema change.

### Demo
N/A

### Docs and Changelog

### DIFF
--- a/apps/evaluations/auto_population.py
+++ b/apps/evaluations/auto_population.py
@@ -77,9 +77,8 @@ def _scan_for_new_sessions(rule: DatasetAutoPopulationRule, created_floor) -> li
     eval_messages = make_session_evaluation_messages(session_external_ids, team=rule.team)
     if not eval_messages:
         return []
-    created = EvaluationMessage.objects.bulk_create(eval_messages)
-    rule.dataset.messages.add(*created)
-    return list(created)
+    created, _ = rule.dataset.add_messages(eval_messages)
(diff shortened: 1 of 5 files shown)


SmittieC (approved):

SmittieC:
@snopoke this can be merged?

Issue #3458, opened 2026-05-28
Read-only bot inspection: consumer acceptance criteria from an external verifier (companion to #3452)


jjackson:
Companion to #3452 — **consumer-side acceptance criteria**, not a competing design. #3452 already has the `/inspect/` plan moving (and the right instinct per @snopoke: build out the API rather than a read-only *role*, since `UserAPIKey.read_only` already exists). This issue exists so that endpoint is designed against a concrete external consumer instead of in the abstract — and to flag one thing the current plan would miss.

## Who the consumer is

We're standing up an external agent (ACE) that automates the **Connect Interviews** launch + QA workflow. One of its jobs is to read a deployed interview bot and assert it's wired correctly — purely read-only, via a `read_only` `UserAPIKey` scoped to the team. No writes, no secrets, no role-system changes needed. This is the exact "agent inspects a bot's setup" use case in #3452, with a specific checklist.

## The assertions the verifier must make (→ fields the `/inspect/` payload needs)

| # | Assertion | Field(s) needed in payload |
|---|---|---|
| 1 | Bot exists + is the working/published version we expect | `name`, `external_id`/`public_id`, `version_number` ✅ *(already in `ExperimentSerializer`)* |
| 2 | Pipeline has a router node, and its routing keywords match the cohort's interview schedule | digested pipeline graph + per-node detail incl. router `keywords`/`route_key` params — **covered by #3452 plan** |
| 3 | An LLM node points at the expected collection (RAG), and the collection has the expected files | node detail + top-level `media_collections[]` with file list — **covered by #3452 plan** |
| 4 | **A 24-hr inactivity `TimeoutTrigger` exists and its action is the completion handler** | **experiment-level** `events`/`triggers[]` — ⚠️ **NOT covered** (see below) |
| 5 | A "Session Completion" custom action exists and is wired to fire | custom action detail + the node/event it's attached to — partially covered (custom actions listed as node resources; the *event* wiring is the gap in #4) |

## The one gap to flag: events/triggers are experiment-level, not node resources

#3452's plan describes "details of every resource any **node** references." But `StaticTrigger` and `TimeoutTrigger` aren't pipeline nodes — they hang off the experiment directly (`apps/events/models.py`; both managers filter `experiment__is_default_version=True`, and creation lives at `/a/<team>/chatbots/<id>/events/...`, not in the pipeline graph).

[...]

Issue #3452, opened 2026-05-28
Support for read-only role


SmittieC:
[There is a use-case](https://dimagi.slack.com/archives/C04VBQJ1DL3/p1779856829185219) for agent-based workflows to inspect a bot's setup in OCS, but we don't want the agent to accidentally change something in the team. We can achieve this by adding a read-only user role. This is the easy part. The harder part is making sure that this user can

1. see all views
2. not change anything

To do this, we need to check for `change` permissions on POST requests instead of GET requests also (like we currently do).

SmittieC:
@snopoke thoughts?

snopoke:
I would rather spend this effort on building out more APIs

SmittieC:
The requirement is for an external system to inspect a specific bot through the API. We need to expose as much info of a bot as possible, except secret keys, tokens etc. This includes:
- bot name
- external id
- all configured settings (non-configured should be included, but be None)
- a digested view of the pipeline structure (which node is where)
- on the same level as the digested pipeline structure, each node should be listed with its details
- any resources that a node uses should be included in the node's details. For instance, if an LLM node uses a specific media  collection, a list of each file in the collection should also be included. Resources should not be nested, but rather all at the top level with references from the user to the resource. {nodes: [{"id": 123, media_collection: XYZ}], media_collections: [{"id": XYZ, "files": [...]}]}
- Resources include (but is not limited to), LLM providers, collections, source material etc.

This API should be read-only.

@claude please plan this API in detail

snopoke:
@claude these endpoints are read-only for now but the long term plan is to add write capabilities to the APIs. There is also a tangential goal of migrating away from the 'experiment' nomenclature to 'chatbots'. Since API changes are hard to do once they are in use this seems like a good opportunity.

Also evaluate if we should introduce API versioning as a way of rolling these out.

snopoke:
@claude I'm happy with this direction. Please also read https://github.com/dimagi/open-chat-studio/issues/3458 to incorporate those acceptance criteria.

snopoke:
@claude scope out the public ID migration and create a separate ticket for it

snopoke:
### Design finalized → see #3487

The design hashed out in this thread has been consolidated into a stable design doc plus 8 ADRs, opened as **#3487** (draft). The verbose planning comments above are superseded and have been collapsed as outdated.

**What the design lands on:**
[...]

PR #3487, opened 2026-05-29
Design doc: read-only chatbot inspection API

### LINKED ISSUE #3465: Add public_id (UUID) to resource models exposed by the v2 API
## Scope

Prerequisite for the v2 API surface scoped under #3452 — most resources the new `/api/v2/chatbots/{id}/inspect/` endpoint surfaces are currently identified only by their numeric DB primary key. Numeric IDs are unstable across environments (staging vs. prod), leak DB internals, and lock us in once external consumers start dereferencing them. We need to introduce a stable opaque external identifier (`public_id`, `UUIDField`) on every resource model the v2 API exposes, **before** the inspect endpoint ships.

Already done: `Experiment.public_id`, `Participant.public_id` (both `UUIDField(default=uuid.uuid4, unique=True)`), `ExperimentSession.external_id`. The pattern is established — this ticket extends it.

## Models to update

Grouped by app. Each row notes how the model is currently scoped, since that determines whether the new UUID needs any special handling (it doesn't — UUIDs are globally unique — but versioning and team-scoping affect how `create_new_version` and lookups are wired).

| App | Model | Base | Versioned? | Notes |
|---|---|---|---|---|
| `service_providers` | `LlmProvider` | `BaseTeamModel` | no | — |
| `service_providers` | `LlmProviderModel` | `BaseTeamModel` | no | `team` is nullable (global rows) |
| `service_providers` | `VoiceProvider` | `BaseTeamModel` | no | — |
| `service_providers` | `MessagingProvider` | `BaseTeamModel` | no | — |
| `service_providers` | `AuthProvider` | `BaseTeamModel` | no | — |
| `service_providers` | `TraceProvider` | `BaseTeamModel` | no | — |
| `service_providers` | `EmbeddingProviderModel` | `BaseTeamModel` | no | `team` may be nullable like `LlmProviderModel` — confirm during impl |
| `experiments` | `SourceMaterial` | `BaseTeamModel` | yes (`VersionsMixin`) | new UUID per version |
| `experiments` | `Survey` | `BaseTeamModel` | yes | new UUID per version |
| `experiments` | `ConsentForm` | `BaseTeamModel` | yes | new UUID per version |
| `experiments` | `SyntheticVoice` | `BaseModel` | no | **global, not team-scoped** |
| `documents` | `Collection` | `BaseTeamModel` | yes | new UUID per version |
| `files` | `File` | `BaseTeamModel` | yes | new UUID per version |
| `assistants` | `OpenAiAssistant` | `BaseTeamModel` | yes | new UUID per version |
| `custom_actions` | `CustomAction` | `BaseTeamModel` | no | — |
| `events` | `StaticTrigger` | `BaseModel` | yes | attached to experiment, not team-scoped |
| `events` | `TimeoutTrigger` | `BaseModel` | yes | attached to experiment, not team-scoped |
| `events` | `EventAction` | `BaseModel` | yes | attached to triggers |
| `pipelines` | `Pipeline` | `BaseTeamModel` | yes | referenced from inspect; new UUID per version |

**Already have a public/external ID — out of scope:** `Experiment`, `Participant`, `ExperimentSession` (uses `external_id`), `Team` (uses `slug`).

**Intentionally out of scope:**
- `Node` (`apps/pipelines/models.py`) — already has `flow_id`, the canonical graph-level identifier. Reuse that.
- `EventLog`, `ScheduledMessageAttempt`, `CollectionFile` — operational/relationship tables, not surfaced as standalone resources.
- `ScheduledMessage` — only relevant if `/inspect/` decides to expose `schedule_trigger` action params (open question in #3452). Add later if/when needed.

## Migration shape

Per affected model, in this order:

1. **Schema migration — add nullable field.**
   ```python
   public_id = models.UUIDField(default=uuid.uuid4, null=True, unique=True)
   ```
   Nullable initially so the schema migration is decoupled from the backfill.

2. **Data migration — backfill UUIDs.**
   Single `RunPython` operation per app. Iterate the model in chunks (`.iterator(chunk_size=1000)`), assign `uuid.uuid4()`, `bulk_update(..., ["public_id"])`. Reversible via no-op.

3. **Schema migration — drop nullability.**
   ```python
   ```

The three-step pattern is what Django docs recommend for non-nullable additions on tables that may be large. `File`, `Collection`, and the provider tables are the candidates for "large" — the rest are usually small.

**Indexing.** `unique=True` already creates an index. No extra index needed.

**Shared mixin.** Consider adding `PublicIdMixin` to `apps/utils/models.py`:
```python
class PublicIdMixin(models.Model):

    class Meta:
        abstract = True
```
Apply to each affected model. Optional retrofit of `Experiment`/`Participant` to use the mixin is cosmetic and not required for this ticket.

## Versioning interaction

For models with `VersionsMixin`, when `create_new_version` runs, the new version row must receive a **fresh** UUID. This already exists on `Experiment.create_new_version_for_experiment` (`apps/experiments/models.py:899`):

```python
new_version.public_id = uuid4()
```

Apply the same pattern to each versioned model's `create_new_version` override:
- `SourceMaterial`, `Survey`, `ConsentForm` (any custom `create_new_version`? — audit during impl, otherwise add)
- `Collection.create_new_version` (`apps/documents/models.py:175`)
- `File.create_new_version` (likely default `VersionsMixin` behaviour — needs override added)
- `OpenAiAssistant.create_new_version` (`apps/assistants/models.py:109`)
- `Pipeline.create_new_version` (`apps/pipelines/models.py:282`)
- `StaticTrigger`, `TimeoutTrigger`, `EventAction` — default `VersionsMixin` behaviour; need override added

Alternative: centralise the UUID reset in `VersionsMixin.create_new_version` itself — assign a new UUID if `hasattr(new_instance, "public_id")`. **Recommended** — one place, less to forget, future-proof.

## Backwards compatibility / call-site impact

- **Internal code**: zero impact. Existing FK relationships continue to work with numeric PKs. `public_id` is additive.
- **External API (current `/api/experiments/...` surface)**: zero impact. Still uses `Experiment.public_id`. Nested resources are not currently exposed externally, so the numeric-ID concern doesn't bite yet.
- **Future `/api/v2/chatbots/{id}/inspect/`**: this endpoint emits `public_id` for nested resources from day one. No deprecation cycle needed for numeric IDs in the API surface because they were never externally exposed.
- **Database constraints**: Adding `unique=True` columns can fail on backfill if duplicate UUIDs are generated — `uuid4()` collision risk is astronomically low, but the three-step migration plan above guards against transient mid-backfill states.

## Test coverage

- Migration smoke test per app: load the new migration on a populated test DB; assert all rows have a non-null, unique `public_id`.
- Versioning test: for each versioned model, `create_new_version()` must produce a different `public_id` than the working version.
- Factory updates: `apps/utils/factories/...` — add `public_id = factory.Faker("uuid4")` to each affected factory so tests don't fail with unique constraint violations under parallel creation (pattern already used in `apps/utils/factories/experiment.py:68,82`).

## Suggested PR breakdown

To keep individual PRs reviewable:

1. **PR 1** — `PublicIdMixin` + centralised UUID reset in `VersionsMixin.create_new_version`. No model changes yet.
2. **PR 2** — apply mixin to `service_providers` models (7 models, no versioning). Backfill migration.
3. **PR 3** — apply mixin to versioned models (`SourceMaterial`, `Survey`, `ConsentForm`, `Collection`, `File`, `OpenAiAssistant`, `Pipeline`). Backfill migrations.
4. **PR 4** — apply mixin to `CustomAction` + `events` models (`StaticTrigger`, `TimeoutTrigger`, `EventAction`) + `SyntheticVoice`.

Each PR can ship independently and is backwards-compatible per the section above.

## Acceptance criteria

- [ ] Every model listed in the table above has a `public_id` UUID column, populated for all existing rows, marked `unique=True`, indexed implicitly.
- [ ] Versioned models assign a fresh `public_id` when `create_new_version()` runs (verified by test).
- [ ] `PublicIdMixin` (or equivalent) exists in `apps/utils/models.py` and is applied consistently.
- [ ] Factories produce unique `public_id`s for each affected model.
- [ ] No existing API endpoint changes its response shape (this work is preparatory; consumption lands with #3452).
- [ ] Migrations are backwards-compatible per the PR-template checklist (forward-only column addition + reversible data migration).

## Out of scope (do not include in this work)

- Switching any existing API endpoint to use `public_id` instead of numeric IDs — that's the v2 work tracked under #3452.
- Adding `public_id` to operational/log models (`EventLog`, audit records).
- URL slugs or human-readable IDs — UUID is sufficient; slug debate is separate.

---

Parent: #3452 (depends on the public IDs landing first).

snopoke:
Captures the design and decisions for the read-only chatbot inspection API ([#3452](https://github.com/dimagi/open-chat-studio/issues/3452)), with consumer acceptance criteria from [#3458](https://github.com/dimagi/open-chat-studio/issues/3458). Decisions are recorded as ADRs (canonical); the design doc is retained as the implementation companion.

See https://github.com/dimagi/open-chat-studio/pull/3493 for write API design.

### Technical Description
Docs only — no code. Adds:
- **8 ADRs (0019–0026, `proposed`)** capturing the architectural decisions — API-surface-not-a-role, URL-path versioning (v1 frozen / v2 new), experiment→chatbot rename, `/inspect/` as a denormalized read-only projection, inline nested resource tree, deferring public IDs, allowlist-based secrets exclusion, and the team-scoped authorization model.
- `docs/design/read-only-chatbot-inspection-api.md` (`status: extracted`) — the implementation companion: payload shape, node-reference mapping, secrets-exclusion audit, versioning rules, implementation plan, test plan. Has a Decisions→ADR map.
- `CONTEXT.md` glossary updates (Provider vs Model, Triggers/Event Actions, snapshotted-vs-live on publish).
- mkdocs nav + ADR index wiring.

Reviewer focus: the ADRs are the durable artifacts — **0020** (versioning, premise verified from request logs), **0024** (reversal — no public IDs; **closed [#3465](https://github.com/dimagi/open-chat-studio/issues/3465)**), **0023** (inline tree), **0025** (secrets allowlist), **0026** (transitive authorization — flagged as deliberate for security review). The design was stress-tested against the models before extraction; ADRs are `proposed` since nothing's implemented yet.

### Docs and Changelog

N/A — this PR *is* internal design/architecture documentation; no product changelog entry.


docs/adr/0024-inspect-denormalized-readonly-projection.md
```diff
@@ -0,0 +1,24 @@
+# ADR-0022: `/inspect/` as a denormalized read-only projection on a distinct URL
+
+<span class="adr-status adr-status-proposed">PROPOSED</span>
+
+<p class="adr-meta">Author: Simon Kelly · Created: 2026-05-29</p>
+<p class="adr-meta">Extends: <a href="0020-url-path-api-versioning.md">ADR-0020</a></p>
+
+## Context
+
+An external agent needs a chatbot's full configuration in a single read. These endpoints are read-only now, but write capabilities are planned. The risk in shipping reads first is that a rich read shape placed on the canonical resource URL leads clients to assume it is round-trippable — that they can PATCH whatever they GET — which would constrain the eventual write shape.
```
SmittieC, line 10: (Low logaf score) What will the GET `/chatbots/{id}` return once we do support a richer API around chatbots? Personally I think it's fine to return a 403 for anything other than a GET, especially since the API docs will not mention anything about PATCH / DELETE that is possible.

docs/adr/0024-defer-public-ids-to-write-api.md
```diff
@@ -0,0 +1,27 @@
+# ADR-0024: Defer stable public IDs to a future write API
```
SmittieC: When I read this, I got the sense that it's making a case against something that was mentioned in a design doc - not an earlier ADR - which wouldn't be an issue if we're doing `ADR 0023`, so it's kind of redundant to have this ADR? It's just making the implicit fact that we don't have to expose IDs in the inline payloads, explicit, which I'm not convinced is needed. Happy to be wrong though

snopoke: yea, that's reasonable. I kept it because we did explicitly make the decision not to go that route - but maybe re-phrasing it to avoid mentioning the earlier design question would make it clearer.

snopoke: renamed and re-written as 0024-identify-resources-by-primary-key.md

docs/adr/0024-inspect-denormalized-readonly-projection.md
snopoke: Great question. I think it shows a bare view of the chatbot data with the same basic structure as /inspect/ but without expanded relations (except for channels and events which are partially expanded):

```jsonc
{
  "id": "5a3c..",
  "name": "Customer Support Bot",
  "description": "…",
  "version_number": 0,
  "is_working_version": true,
  "is_default_version": false,
  "version_description": null,
  "team_slug": "acme",
  "settings": {
    // non-secret Experiment fields, null if unset.
    "seed_message": null,
    "conversational_consent_enabled": false,
    "voice_response_behaviour": "reciprocal",
    "echo_transcript": false,
    "use_processor_bot_voice": false,
    "debug_mode_enabled": false,
    "file_uploads_enabled": false,
    "participant_allowlist": []
  },

  // chatbot resources references only (numeric db id; null if unset)
  "consent_form_id": 3,
  "pre_survey_id": null,
  "post_survey_id": 9,
  "voice_provider_id": 4,
  "synthetic_voice_id": 6,
  "trace_provider_id":  null,
  // channels embedded with reference ID for messaging provider
  "channels": [
    { "id": 42, "platform": "telegram", "name": "Support TG", "messaging_provider_id": 6 }
  ],
  "pipeline_id": 42,
  // events embedded with raw params
  "events": {
    "static_triggers": [
      {
        "id": 11,
        "type": "conversation_end", 
        "is_active": true,
        "action": {
          "type": "pipeline_start", "params": {...}
        }
      }
    ],
    "timeout_triggers": [
      {
        "id": 22,
        "delay_seconds": 86400,
        "total_num_triggers": 1,
        "trigger_from_first_message": false,
        "is_active": true,
        "action": {
          "type": "send_message_to_bot", "params": { ... }
        }
      }
    ]
  }
}
```

docs/design/read-only-chatbot-inspection-api.md
```diff
@@ -0,0 +1,815 @@
...
+  "version_number": 0,
+  "is_working_version": true,
+  "is_default_version": false,
+  "version_description": null,
+  "team_slug": "acme",
+  "settings": {
+    // non-secret Experiment fields, null if unset. NOTE: prompt/temperature/tools/
+    // source_material/citations live on the LLM node now, NOT on the chatbot — the
+    // legacy Experiment-level fields (temperature, tools, citations_enabled, prompt_text,
+    // input_formatter, source_material) are removed/not surfaced.
```
snopoke:
```suggestion
    // non-secret Experiment fields, null if unset.
```

docs/design/read-only-chatbot-inspection-api.md
```diff
@@ -0,0 +1,815 @@
...
+  },
+
+  "events": {                             // experiment-level — D9
+    "static_triggers": [
+      {
+        "id": 11,
+        "type": "conversation_end",       // StaticTriggerType
+        "is_active": true,
+        "action": {
+          "id": 47, "action_type": "pipeline_start",
```
snopoke: drop "id" field for event actions since they are 1-1 with the event. Also drop the 'action_' prefix on `action_type`: `action_type` -> `type` (since it's already scoped inside the `action` object.

docs/adr/0024-inspect-denormalized-readonly-projection.md
snopoke: See https://github.com/dimagi/open-chat-studio/pull/3493/changes#r3333601130

docs/design/read-only-chatbot-inspection-api.md
snopoke: Fixed in dec6fd87d. Collapsed the comment to the suggested one-liner (`// non-secret Experiment fields, null if unset.`).

*— Claude Code*

docs/design/read-only-chatbot-inspection-api.md
snopoke: Fixed in dec6fd87d. Dropped the `id` field on embedded event actions (1-1 with the event) and renamed `action_type` → `type` in the payload (it's already scoped inside `action`); also updated the D9 prose. Left the `EventAction.action_type` model field unchanged.

*— Claude Code*

docs/adr/0024-inspect-denormalized-readonly-projection.md
SmittieC: I see, so the GET `/chatbots/{id}` endpoint will be more tailored towards returning useful data for chatbot construction whereas the `/inspect` is more intended for "understanding" what goes on in your chatbot (leaving out details like ids etc).

snopoke: Something like that

PR #3495, opened 2026-06-01
Design: live shared index collections (#2123)

### LINKED ISSUE #2123: update published collection when updates are received from document source
When a document source does automatic updates it should apply those updates to the published collection otherwise the automatic update process is pointless.

* Should this be optional via a flag?
* Maybe also a manual option to update the published collection?

snopoke:
Closes #2123.

### Product Description
Changes to a chatbot's linked collections now take effect on **published** chatbots, not just the working version. This covers scheduled document-source updates to RAG **index** collections (the original #2123 bug) and manual edits to **media** (attachment) collections. Previously a published chatbot read a frozen snapshot of its collections captured at publish time.

### Technical Description
Implements [ADR-0031](docs/adr/0031-collection-content-is-live-shared-resource.md): a collection's *content* is a live shared resource rather than something frozen per published bot. Applies to both the index list (`collection_index_ids`) and the single media collection (`collection_id`) on `LLMResponseWithPrompt` nodes.

The core change is a deletion: `Node.create_new_version` no longer versions either collection param, so the published node keeps the live working-collection id verbatim. (`source_material_id` and assistants still version per bot.) Everything else follows:

- `sync_all_document_sources_task` filters `collection__working_version__isnull=True` so snapshots / legacy frozen copies are never auto-synced (indexes only — media collections have no document sources).
- `Node._archive_related_params` archives a referenced collection only when it's a frozen version (`is_a_version`) — never the live working collection — while still cleaning up legacy frozen copies. Media and index params share one guarded loop.

Worth a reviewer's attention (not obvious from the diff):

- **New-only rollout, no data migration.** Existing published bots keep their frozen ids and behave exactly as before on deploy; they pick up live behaviour on their next republish.
- **Two effects are intentional and fall out for free:** (1) collection content/config drift no longer marks a bot as having "unpublished changes" — published and working nodes hold the same id, so the version diff compares the collection to itself; (2) the existing `Collection.archive()` guard already protects a live collection referenced by a published bot, since published params now contain the working id.
- **Explicit snapshots (opt-in freeze + node picker) are deliberately out of scope** — planned follow-up. ADR-0031 (in this PR) has the full rationale and rejected alternatives, including the explicitly-rejected option of keeping media frozen while only indexes go live.

The `is_a_version` archive guard is covered by tests verified to be load-bearing (they fail if the guard is removed) for both the index and media paths.

### Migrations
N/A — no schema migration (a JSONField-content convention plus a task-filter change). One behaviour delta for legacy data: frozen snapshot document sources are no longer auto-synced (they were synced wastefully before); invisible to end users.

### Docs and Changelog

Changelog note: published chatbots now automatically reflect updates to their linked collections — scheduled document-source syncs to RAG indexes and manual edits to media collections — instead of staying pinned to the snapshot captured at publish time.


docs/superpowers/specs/2026-06-01-live-shared-index-collections-design.md
```diff
@@ -0,0 +1,181 @@
...
+In scope:
+
+- Stop creating frozen index-collection versions when a bot is published.
+- An explicit "Snapshot" action to freeze an index on demand, and the ability to
+  point a pipeline node at a snapshot instead of the live index.
+- A guard ensuring snapshots (and legacy frozen copies) never auto-sync.
+
+Out of scope:
+
+- Media collections referenced via `collection_id` (single-collection message
```
snopoke, line 59: I think this should also be in scope and work in the same way.

snopoke:
@SmittieC let me know what you think if this

SmittieC:
> @SmittieC let me know what you think if this

Looks great! It's nice that the changes are very simple to achieve this

snopoke:
@coderabbitai review

snopoke:
@coderabbitai full review

PR #3496, opened 2026-06-01
feat(api): add URL-path versioning, pin v1 against current URLs

snopoke:
### Product Description
No change. Infrastructure only — sets up API versioning so the existing API can be frozen as `v1` while a renamed `v2` surface and new endpoints (the read-only chatbot `/inspect/` endpoint) are added later.

### Technical Description
Track A step 1 of the [read-only chatbot-inspection API design](docs/design/read-only-chatbot-inspection-api.md). `v1` freezes today's surface and is *also* served under the existing unversioned `/api/` alias, so current callers never break. `/api/v2/` is allowed by settings but has no routes yet (404s) — it's where the `experiment`→`chatbot` rename and new endpoints land in a follow-up.

The notable decision is **not** using DRF's stock `URLPathVersioning`:
- it captures `version` as a URL kwarg and passes it to the view, but many of our function-based views (`chat_start_session(request)`, `generate_key`, `consent`, …) take no such arg → `TypeError`;
- its `reverse()` injects `version` into every reverse call, which would break the ~101 `reverse("api:...")` callers and the `HyperlinkedRelatedField`s that must keep emitting unversioned `/api/experiments/{id}/` URLs.

Instead, `apps/api/versioning.py` reads the version from `request.path` (defaulting to `v1`) and leaves `reverse()` plain. The routes are mounted once under an optional, non-capturing prefix — `re_path(r"^(?:v1/)?", include(v1_patterns))` — so `/api/v1/...` and `/api/...` resolve to the same views, `reverse()` deterministically stays unversioned, and there's no view churn or OpenAPI schema duplication.

The pre-existing chatbot-version path segment (`/api/openai/<id>/v<int:version>/chat/completions`) is a different concept and does not collide.

### Migrations
N/A — no migrations.

### Demo
N/A.

Verified: full `apps/api/tests/` suite passes (incl. `test_schema.py`, so `api-schema.yml` is unchanged); all `api:` reverse names still produce unversioned URLs; `/api/v1/experiments/` resolves to the same view as `/api/experiments/`; `/api/v2/...` → 404. New tests in `apps/api/tests/test_versioning.py`.

### Docs and Changelog

### DIFF
--- a/apps/api/tests/test_versioning.py
+++ b/apps/api/tests/test_versioning.py
@@ -1,0 +1,76 @@
+import pytest
+from django.urls import resolve, reverse
+from rest_framework.exceptions import NotFound
+from rest_framework.test import APIRequestFactory
+
+from apps.api.versioning import URLPathVersioning
+from apps.utils.factories.experiment import ExperimentFactory
+from apps.utils.factories.service_provider_factories import LlmProviderFactory
+from apps.utils.factories.team import TeamWithUsersFactory
+from apps.utils.tests.clients import ApiTestClient
+
+
+@pytest.fixture()
+def experiment(db):
+    exp = ExperimentFactory.create(team=TeamWithUsersFactory.create())
+    LlmProviderFactory.create(team=exp.team)
+    return exp
+
+
+class TestURLPathVersioning:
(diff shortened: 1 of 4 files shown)


SmittieC (approved):
Sweet!

Issue #3497, opened 2026-06-01
Allow bulk download of eval results from the evaluation runs page


ocs-percy:
## Summary

Add a bulk download feature to the evaluation runs page that exports evaluation results as a file.

## Behaviour

- Export should include the **most recent result for each dataset item** from the linked dataset
- Download should be triggered from the evaluation runs page (e.g. a download/export button)
- Generation should be **async** (Celery task)
- UI should show **download progress** using `celery-progress` (consistent with existing usage in the project)
- **CSV structure should match the existing eval run export**

## Motivation

Currently there's no way to export eval results in bulk, making it difficult to analyse results outside of the UI.

## Acceptance Criteria

- [ ] A download/export action is available on the evaluation runs page
- [ ] The export contains the most recent result per dataset item from the linked dataset
- [ ] Export is generated asynchronously via a Celery task
- [ ] UI shows download progress via `celery-progress` (matching existing patterns in the project)
- [ ] CSV structure matches the existing eval run export format

PR #3498, opened 2026-06-01
Add async bulk download of latest eval results from runs page

ocs-percy:
## Summary

Implements #3497.

Adds a **Download Latest Results** button to the evaluation runs home page that generates a CSV asynchronously, showing download progress via `celery-progress`.

## What's in the export

- The **most recent result for each dataset item** (per message/evaluator pair), across all completed FULL and DELTA runs for the evaluation config
- CSV structure matches the existing per-run download (same `EVALUATION_RUN_FIXED_HEADERS` column ordering)

## Changes

- `apps/evaluations/tasks.py` — new `export_evaluation_bulk_results_task` (Celery, `ProgressRecorder`, saves to `File` model)
- `apps/evaluations/views/evaluation_config_views.py` — two new views: `start_bulk_download` (POST, triggers task) and `get_bulk_download_link` (GET, polls status + returns download link)
- `apps/evaluations/urls.py` — two new URL patterns
- `templates/evaluations/partials/bulk_download.html` — HTMX polling template (modelled on `experiments/components/exports.html`)
- `templates/evaluations/evaluation_runs_home.html` — includes the new partial in the button row

### DIFF
--- a/apps/evaluations/tasks.py
+++ b/apps/evaluations/tasks.py
@@ -1,7 +1,7 @@
 import contextlib
 import csv
 import math
-from collections import OrderedDict, defaultdict
+from collections import defaultdict
 from collections.abc import Iterable
 from datetime import timedelta
 from io import StringIO
@@ -27,6 +27,7 @@ from apps.evaluations.auto_population import (
 )
 from apps.evaluations.const import PREVIEW_SAMPLE_SIZE
 from apps.evaluations.exceptions import HistoryParseException
+from apps.evaluations.export import build_evaluation_table_data, write_evaluation_csv
 from apps.evaluations.models import (
     DatasetCreationStatus,
     EvaluationConfig,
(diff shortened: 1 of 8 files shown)


snopoke (commented):
Can this export not reuse code used by download_evaluation_run_csv?

ocs-percy:
Thanks for the feedback, all addressed in the latest commit:

**CodeScene / complexity:** Extracted four helper functions from `export_evaluation_bulk_results_task` — `_get_bulk_results_queryset`, `_populate_message_row_fixed_fields`, `_build_evaluation_table_data`, and `_write_evaluation_csv`. The task body is now a simple orchestration function and should be well under the complexity thresholds.

**DB deduplication:** Replaced the two-step `completed_run_ids` + Python-side dedup with a single queryset using `.distinct("message_id", "evaluator_id")` ordered by `-run__created_at` — PostgreSQL `DISTINCT ON` does the dedup in the DB, so only the latest-run row per pair is ever fetched.

**Evaluator-specific errors:** Switched from overwriting a single `error` column to namespaced `error (EvaluatorName)` columns, matching the pattern used for result columns. Multiple evaluator failures on the same message are now preserved.

**Code reuse (@snopoke):** Extracted `_write_evaluation_csv` as a shared helper and updated `download_evaluation_run_csv` to use it too — the per-run and bulk exports now share the same column-ordering logic.

apps/evaluations/tasks.py
```diff
@@ -955,6 +955,96 @@ def create_dataset_from_sessions_task(self, dataset_id, team_id, session_ids):
...
+    return [{"#": index, **row} for index, row in enumerate(table_by_message.values())]
+
+
+def _write_evaluation_csv(writer, table_data: list[dict]) -> None:
+    """Write *table_data* rows to *writer* using the standard evaluation column ordering.
+
+    Fixed headers come first, then alphabetically-sorted dynamic columns, then any
+    error columns last.  This is shared between the per-run and bulk-download exports.
+    """
+    from apps.evaluations.const import EVALUATION_RUN_FIXED_HEADERS
```
snopoke: yank import to module level

apps/evaluations/tasks.py
snopoke, line 1: There still seems to be a reasonable amount of overlap between `EvaluationRun.get_table_data` and the code in this file. They are both generating export data for evaluation results. It would be great if there weren't 2 code paths doing the same thing. Can we try and refactor to reuse code.

apps/evaluations/tasks.py
snopoke: Fixed in a580cff37. The `EVALUATION_RUN_FIXED_HEADERS` import is now at module level — the writer moved into the new `apps/evaluations/export.py`, so the import lives at the top of that module.

*— Claude Code*

apps/evaluations/tasks.py
snopoke: Fixed in a580cff37. Extracted the shared row-builder and CSV writer into a new `apps/evaluations/export.py` (`build_evaluation_table_data` + `write_evaluation_csv`). `EvaluationRun.get_table_data` now keeps only its queryset/delta-scoping and delegates row-building to the shared function, and the bulk export task uses the same two helpers — so there is a single code path. As a side effect the per-run path now namespaces errors per evaluator (`error (EvaluatorName)`) like the bulk path, preserving evaluator-specific failures. All 258 evaluations tests pass.

*— Claude Code*

snopoke (approved):

PR #3517, opened 2026-06-02
Add Undo Applied Tags button to evaluation runs

### LINKED ISSUE #3439: Ability to undo a run's applied tags

SmittieC:
The new version of https://github.com/dimagi/open-chat-studio/pull/3466

### Product Description

Adds an **Undo Applied Tags** button to the evaluation run detail page, so a misbehaving evaluator's tagging can be reverted in one click. The button appears only on the latest completed **full** run of a config, and a run can be undone once.

The tag tooltip now also attributes eval-driven tags to the evaluator and run that applied them (e.g. `evaluator 'Toxicity' (run #42)`) instead of the generic "Participant".

Resolves #3439 and #3488.

### Technical Description

Builds on the eval tagging work in #3414 / #3466. See **[ADR-0031](https://github.com/dimagi/open-chat-studio/blob/3b8e68c8fba7534add73006ffbe6a3a0deac3ed0/docs/adr/0031-eval-tag-undo-full-runs-only.md)** for the design and rejected alternatives.

- **Undo reverts the diff, not the raw tag set.** `undo_run_tags` (in `apps/evaluations/tagging.py`) restores the previous tagging epoch (prior FULL run + DELTAs since), removing only tags this run added and re-applying tags it dropped. This avoids wrongly removing tags a previous run also applied. Restricted to the latest completed FULL run, undoable once (`can_undo_tags`); the view is team-scoped, POST-only, and permission-gated.
- **Live state is one `EvaluationRun.tags_archived` boolean**, not a per-`AppliedTag` marker (the approach #3488 proposed). A FULL completion archives superseded runs; undo flips the flags back. `AppliedTag` audit rows are never deleted — only `CustomTaggedItem` (live tags) is mutated, all in one transaction. The active set is the non-archived runs, which drives both undo eligibility and tooltip attribution.
- Correctness relies on the DELTA invariant (disjoint message-sets); `EvaluationConfig.run` now rejects overlapping DELTA `scoped_messages` rather than silently corrupting undo.

### Migrations

`0016_evaluationrun_tags_archived` adds one `BooleanField(default=False, db_index=True)`; existing rows default to non-archived.

### Demo
Tag attribution

image

### Docs and Changelog

Adds ADR-0031. Changelog: users can undo an evaluation run's applied tags, and tag tooltips now show which evaluator/run applied an eval-driven tag.


apps/annotations/tag_attribution.py
```diff
@@ -0,0 +1,65 @@
+"""Resolve who/what applied an eval-driven tag, for the tag tooltip.
```
SmittieC, line 1: It would be nice if this module can be a service of some kind from the annotations app so that callers can simply call the service and not have to worry about the evaluations app

apps/evaluations/tagging.py
```diff
@@ -206,3 +228,206 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+    if not possible_tags:
+        return
+
+    # All evaluators in a config share the dataset's evaluation_mode.
+    # Use the first as a representative to carry mode into resolve_target.
+    representative_evaluator = evaluators[0]
+
+    restore_runs = _restore_set_for_run(run)
+    remove_by_target, add_by_target, content_type = _compute_undo_target_diffs(
+        run, restore_runs, possible_tags, representative_evaluator
```
SmittieC: Seems like the only reason for passing in representative_evaluator here is to later access the `evaluation_mode` attribute. We should then just pass in the attr itself rather

apps/evaluations/tagging.py
```diff
@@ -143,14 +150,29 @@ def _get_possible_tags(evaluators: list[Evaluator]) -> frozenset[int]:
...
+    if not run_ids:
+        return applied
+    for row in AppliedTag.objects.filter(evaluation_result__run_id__in=run_ids).values(
+        "evaluation_result__message_id", "tag_id"
+    ):
         applied[row["evaluation_result__message_id"]].add(row["tag_id"])
     return applied
 
 
+def _build_applied_by_message(run: EvaluationRun) -> defaultdict[int, set[int]]:
```
SmittieC: this in unnecessary indirection

apps/evaluations/tagging.py
```diff
@@ -206,3 +228,206 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+
+    content_type: ContentType | None = None
+    remove_by_target: defaultdict[int, set[int]] = defaultdict(set)
+    add_by_target: defaultdict[int, set[int]] = defaultdict(set)
+
+    messages_iter = EvaluationMessage.objects.filter(pk__in=message_ids).select_related(
+        "session__chat", "expected_output_chat_message"
+    )
+    for message in messages_iter:
+        target = resolve_target(representative_evaluator, message)
```
SmittieC: resolve_target should return the `Chat` content type if the mode is session, otherwise it should return `ChatMessage`

apps/evaluations/tagging.py
```diff
@@ -206,3 +228,206 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+    """Compute the (remove, add) managed-tag deltas per target for undoing `run`.
+
+    `remove` is tags `run` applied that the restore set does not have; `add` is tags the
+    restore set applied that `run` does not have. Both are intersected with `possible_tags`
+    so we never touch tags this config doesn't manage. The restore set's AppliedTag rows
+    cover disjoint messages (the DELTA invariant), so they compose without per-message run
+    resolution. Returns (remove_by_target, add_by_target, content_type); content_type is
+    None when no message resolved to a target.
+    """
+    current_applied = _build_applied_by_message(run)
```
SmittieC: it would be nice to have a custom type being returned by _build_applied_by_message and _applied_by_message_for_run_ids

apps/evaluations/tagging.py
```diff
@@ -206,3 +228,206 @@ def reverse_stale_tags(run: EvaluationRun) -> None:
...
+    if prev_full is not None:
+        deltas = deltas.filter(finished_at__gt=prev_full.finished_at)
+
+    restore = list(deltas)
+    if prev_full is not None:
+        restore.append(prev_full)
+    return restore
+
+
+def _compute_undo_target_diffs(
```
SmittieC, line 282: Might be nice to add a test or two for this function

apps/evaluations/tests/test_delta_evaluation_run.py
```diff
@@ -25,6 +27,57 @@ def test_run_with_scoped_messages_persists_scope():
     mock_delay.assert_called_once_with(run.id)
 
 
+@pytest.mark.django_db()
+def test_delta_run_rejects_messages_already_evaluated_by_prior_run():
```
SmittieC, line 31: isn't there already a test that tests this logic? same with test_delta_run_allows_messages_not_seen_by_prior_runs

apps/evaluations/tests/test_undo_run_tags.py
```diff
@@ -0,0 +1,346 @@
...
+            message=message,
+            run=run,
+            output=output,
+        )
+        apply_rules_to_result(result, rule_that_fires.evaluator, message)
+        return run
+
+    # ---- tests -------------------------------------------------------------
+
+    def test_preview_run_is_no_op(self, team):
```
SmittieC, line 262: these tests seem very similar to those in apps/evaluations/tests/test_archived_tags.py. Please reconsile and use a single file

apps/evaluations/tagging.py
SmittieC: Fixed in 4f3b522ca. `resolve_target` now takes `evaluation_mode` directly instead of the evaluator, and the stale/undo diff helpers pass the mode through rather than a representative evaluator.

apps/evaluations/tagging.py
SmittieC: Fixed in 4f3b522ca. Added `_content_type_for_mode(evaluation_mode)` which returns the `Chat` ContentType for SESSION and `ChatMessage` otherwise, so callers derive the content type from the mode up front instead of from each resolved instance.

apps/evaluations/tagging.py
SmittieC: Fixed in 4f3b522ca. Added an `AppliedByMessage` type alias (`defaultdict[int, set[int]]`) for the applied-tags mapping. Note `_build_applied_by_message` was already inlined into its single caller in 7b54ce8d1, so `_applied_by_message_for_run_ids` is the only remaining helper and now carries the alias.

apps/evaluations/tagging.py
SmittieC: Fixed in f91ea91f1. Added `TestComputeUndoTargetDiffs` with direct coverage: one case asserting current-only tags are removed and restore-only tags re-added, and one asserting tags common to both produce no diff.

apps/evaluations/tagging.py
SmittieC: Already addressed: the single-use `_build_applied_by_message` wrapper was inlined into its caller in 7b54ce8d1, removing the indirection. `_applied_by_message_for_run_ids` is now called directly.

apps/annotations/tag_attribution.py
SmittieC: Fixed in d7f635e21. Moved the module to `apps/annotations/tag_attribution.py`. Callers now import the service from the annotations app instead of evaluations; `tag_views.py` (already coupled to evaluations.models) imports it at top level, and the remaining callers keep a local import to avoid circular imports through evaluations.models.

apps/evaluations/tests/test_undo_run_tags.py
SmittieC: Fixed in f91ea91f1. Folded test_archived_tags.py into this file, so the full undo/archive surface (archive-on-supersede, undo, eligibility, diff helper) lives in one module with shared helpers. test_archived_tags.py is deleted.

apps/evaluations/tests/test_archived_tags.py
```diff
@@ -0,0 +1,229 @@
...
+        config=config,
+        status=EvaluationRunStatus.COMPLETED,
+        type=run_type,
+        finished_at=timezone.now(),
+    )
+    result = EvaluationResultFactory.create(
+        team=config.team, evaluator=evaluator, message=message, run=run, output=output
+    )
+    apply_rules_to_result(result, evaluator, message)
+    return run
```
SmittieC: Addressed in f91ea91f1. test_archived_tags.py was consolidated into test_undo_run_tags.py, and the unified `_make_run` now bundles fixtures into a `Setup` namedtuple so it takes 4 arguments (setup, output, run_type, message).

apps/evaluations/tests/test_run_set_undo.py
```diff
@@ -0,0 +1,243 @@
...
+            team=config.team, evaluator=evaluator, message=message, run=run, output=output
+        )
+        apply_rules_to_result(result, evaluator, message)
+        scoped.append(message)
+    if run_type == EvaluationRunType.DELTA:
+        run.scoped_messages.add(*scoped)
+
+    reverse_stale_tags(run)
+    archive_superseded_runs(run)
+    return run
```
SmittieC: Fixed in f91ea91f1. `_complete_run` now takes a `Setup` namedtuple and a single `finished_at`, dropping it to 4 arguments (setup, results, run_type, finished_at).

apps/experiments/views/experiment.py
```diff
@@ -668,6 +668,10 @@ def experiment_session_messages_view(request, team_slug: str, experiment_id: uui
     # Add time gap information to messages
     current_page_messages = _add_time_gap_info(current_page_messages)
 
+    from apps.evaluations.tag_attribution import attach_tag_attributions  # noqa: PLC0415 — avoid circular import
+
+    attach_tag_attributions(current_page_messages)
+
```
SmittieC: `experiment_session_messages_view` was already well over the 80-line threshold before this PR; the change here adds only a 2-line `attach_tag_attributions` call. A full decomposition of this view is out of scope for this PR (no code change needed).

apps/evaluations/tests/test_delta_evaluation_run.py
SmittieC: Not duplicates — the DELTA-overlap validation these cover is new in this PR (added with the run-set undo model in b15341046), and these two tests are its only direct coverage. `test_delta_run_rejects_messages_already_evaluated_by_prior_run` asserts the guard raises and creates no run row; `test_delta_run_allows_messages_not_seen_by_prior_runs` asserts a genuinely-new-message DELTA is accepted (no code change needed).

SmittieC:
Superseded by https://github.com/dimagi/open-chat-studio/pull/3546

PR #3520, opened 2026-06-02
Filter traces by session

### LINKED ISSUE #3060: Filter traces by session
Make it easy for users to see all traces for a session

* Click on the session in a trace detail to view traces table filtered by session
* Shortcut from session transcript view to filtered list
* Highlight 'error' traces in session transcript view

snopoke:
### Product Description
Make it easier to find all traces for a given session, chatbot, or participant. The trace table gains a **Session ID** filter, and quick links are added so users can jump from a trace detail page or a chatbot session transcript straight to a pre-filtered trace table.

Closes #3060.

### Technical Description
- `SessionIdFilter` is registered on `TraceFilter` with `columns=["session__external_id"]` — the override is needed because `Trace` reaches the session via FK (the reusable filter defaults to a bare `external_id` path used when filtering sessions directly).
- The navigation links target `trace:home`, **not** `trace:table`. The CodeRabbit plan suggested `trace:table`, but that endpoint only returns the bare HTMX table partial — the Alpine filter UI that reads `filter_N_*` params from `window.location.search` lives on the home page.
- The "View chatbot traces" link uses `working_version_id|default:experiment_id`. The experiment filter's options only list working-version chatbots and its backend clause matches on the working version id, so passing a versioned trace's raw `experiment_id` would not select correctly. The `any of` operator is URL-encoded as `any+of` and the value as a JSON array (`%5B…%5D`), which is what both the JS parser and the backend normalizer expect.
- All links are gated by `trace.view_trace`.

### Migrations
N/A — no migrations.

### Demo

**Quick links on trace view**
Screenshot from 2026-06-02 17-13-04

**Quick link from session view**
Screenshot from 2026-06-02 17-13-14

**Session filter on trace table**
Screenshot from 2026-06-02 17-13-24

### Docs and Changelog

### DIFF
--- a/templates/trace/trace_detail.html
+++ b/templates/trace/trace_detail.html
@@ -27,9 +27,34 @@
   <div>
     <div class="container mx-auto px-4 py-6">
       {# Heading #}
-      <div class="mb-6">
-        <h1 class="text-3xl font-bold mb-1">{{ trace.experiment.name }}</h1>
-        <p class="text-sm text-base-content/60">Trace ID: {{ trace.trace_id }}</p>
+      <div class="flex flex-wrap items-start justify-between gap-2 mb-6">
+        <div>
+          <h1 class="text-3xl font-bold mb-1">{{ trace.experiment.name }}</h1>
+          <p class="text-sm text-base-content/60">Trace ID: {{ trace.trace_id }}</p>
+        </div>
+        {# Quick links to the trace table pre-filtered by this trace's session/chatbot/participant #}
+        {% if perms.trace.view_trace %}
+          <div class="flex flex-wrap gap-2">
+            {% if trace.session %}
+              <a href="{% url 'trace:home' request.team.slug %}?filter_0_column=session_id&filter_0_operator=equals&filter_0_value={{ trace.session.external_id }}"
+                 class="btn btn-sm btn-ghost">
+                <i class="fas fa-stream"></i> View session traces
+              </a>
+            {% endif %}
+            {% if trace.experiment %}
+              <a href="{% url 'trace:home' request.team.slug %}?filter_0_column=experiment&filter_0_operator=any+of&filter_0_value=%5B{{ trace.experiment.working_version_id|default:trace.experiment_id }}%5D"
+                 class="btn btn-sm btn-ghost">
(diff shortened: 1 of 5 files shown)


templates/trace/trace_detail.html
```diff
@@ -27,9 +27,34 @@
...
-      <div class="mb-6">
-        <h1 class="text-3xl font-bold mb-1">{{ trace.experiment.name }}</h1>
-        <p class="text-sm text-base-content/60">Trace ID: {{ trace.trace_id }}</p>
+      <div class="flex flex-wrap items-start justify-between gap-2 mb-6">
+        <div>
+          <h1 class="text-3xl font-bold mb-1">{{ trace.experiment.name }}</h1>
+          <p class="text-sm text-base-content/60">Trace ID: {{ trace.trace_id }}</p>
+        </div>
+        {# Quick links to the trace table pre-filtered by this trace's session/chatbot/participant #}
+        {% if perms.trace.view_trace %}
```
snopoke: This permission check is unnecessary since this is already a view that requires that permission

templates/trace/trace_detail.html
```diff
@@ -27,9 +27,34 @@
...
+          <div class="flex flex-wrap gap-2">
+            {% if trace.session %}
+              <a href="{% url 'trace:home' request.team.slug %}?filter_0_column=session_id&filter_0_operator=equals&filter_0_value={{ trace.session.external_id }}"
```
snopoke: It would be good for these to use the Django `querystring` template tag: https://docs.djangoproject.com/en/6.0/ref/templates/builtins/#basic-usage

Same for the other two buttons.

templates/trace/trace_detail.html
snopoke: Fixed in 3986d93b6. Removed the permission check — `TraceDetailView` already requires `trace.view_trace`, so it was redundant.

*— Claude Code*

templates/trace/trace_detail.html
snopoke: Fixed in 3986d93b6. All three links (plus the "View Traces" button on the session view) now build their query string with `{% querystring %}`.

*— Claude Code*

PR #3526, opened 2026-06-03
feat: add v2 chatbots read API with inspect projection

SmittieC:
### Product Description
New read-only v2 API for chatbots: `GET /api/v2/chatbots/` (list), `GET /api/v2/chatbots/{id}/` (retrieve), and `GET /api/v2/chatbots/{id}/inspect/?version=` — a single-read, denormalized view of a chatbot's full configuration (settings, channels, pipeline nodes with inlined resources, and event triggers) intended for LLM-agent consumers.

### Technical Description
Implements ADR-0024 (inspect as a read-only projection on a distinct URL, separate from the canonical resource so a future PATCH isn't constrained by the read shape) and ADR-0025 (inline nested resource tree — resources are embedded under the node/event that uses them rather than referenced via lookup tables).

Worth reviewer attention:
- **Secrets exclusion** — serializers in `apps/api/v2/inspect/serializers.py` are built from explicit allowlists, with `test_secrets_exclusion.py` and `test_completeness_guard.py` guarding against new model fields silently leaking or being silently dropped.
- **Node walking** is signal-driven (`node_walker.py`): each node type declares which params are resource refs. Empty-string resource ids in node params are treated as absent, and string ids are cast to int for lookup — both came out of real pipeline data.
- **N+1 avoidance** — `InspectCollector` batch-loads each resource type once across the whole walk, then inlines copies (per ADR-0025, duplication is accepted for locality).
- **Channels** always resolve from the working version (channels are only linked to the family head), plus the team-global web and API channels which are linked to no experiment.

### Migrations
N/A — no migrations.

### Docs and Changelog

### Demo
See [this test](https://github.com/dimagi/open-chat-studio/blob/36f65dda965cf019fc6d9146c4728689864e7258/apps/api/v2/tests/test_chatbots_inspect.py#L506) for the example response.


apps/api/v2/inspect/builder.py
```diff
@@ -0,0 +1,151 @@
...
+    if walk is None:
+        return None
+    return {
+        "id": walk.id,
+        "name": walk.name,
+        "version_number": walk.version_number,
+        "graph": walk.graph,
+        "nodes": [
+            {
+                "flow_id": node.flow_id,
```
snopoke: should we expose this as 'flow_id' or 'node_id'? `flow_id` comes from react-flow so it seems a bit weird to use that term since it's not used anywhere else.

apps/api/v2/inspect/collector.py
```diff
@@ -0,0 +1,130 @@
...
+            return SyntheticVoice.objects.filter(id__in=ids).select_related("voice_provider")
+        return None
+
+    def _get(self, kind: str, resource_id) -> object | None:
+        if not resource_id:
+            return None
+        # pipeline.params contains string ids whereas self._objects has int ids.
+        return self._objects.get(kind, {}).get(int(resource_id))
+
+    def inline_refs(self, refs: dict[str, object]) -> dict:
```
snopoke: Feels weird to have rendering be part of the collector. It seems like a separate activity.

snopoke (commented):
Generally LGTM. I haven't thought hard about the collector, walkers design but I don't have any red falgs.

apps/api/v2/inspect/events.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    action_type = action.action_type
+    params = dict(action.params or {})
+
+    if action_type == EventActionType.PIPELINE_START:
+        from apps.pipelines.models import Pipeline  # noqa: PLC0415 - avoid import cycle at module load
+
+        pipeline_id = params.pop("pipeline_id", None)
+        pipeline_walk = None
+        if pipeline_id is not None:
```
SmittieC: we can optimize the checks here please

apps/api/v2/inspect/builder.py
```diff
@@ -0,0 +1,151 @@
...
+    channels = list(
+        ExperimentChannel.objects.filter(experiment_id=experiment.get_working_version_id()).select_related(
+            "messaging_provider"
+        )
+    )
+    # The web and API channels are team-global (linked to no experiment); every chatbot is
+    # reachable through them.
+    channels.append(ExperimentChannel.objects.get_team_web_channel(experiment.team))
+    channels.append(ExperimentChannel.objects.get_team_api_channel(experiment.team))
+    return ChannelSerializer(channels, many=True).data
```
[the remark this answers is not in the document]
SmittieC: Fixed in d33ed6b9. `_collect_channels` now looks up the team-global web/api channels with read-only `filter(...).first()` calls instead of the `get_team_*_channel` helpers (which `get_or_create`), so the GET inspect flow no longer writes `ExperimentChannel` rows. A team that has never exercised those channels simply has them omitted from the payload. Added a regression test: `test_inspect_does_not_create_team_channels`.

apps/api/v2/inspect/node_walker.py
```diff
@@ -0,0 +1,290 @@
...
+            for action_id, _operation_ids in ref.selections:
+                into.setdefault(CUSTOM_ACTION, set()).add(int(action_id))
+        elif isinstance(ref, LlmRef):
+            if ref.provider_id:
+                into.setdefault(LLM_PROVIDER, set()).add(int(ref.provider_id))
+            if ref.model_id:
+                into.setdefault(LLM_PROVIDER_MODEL, set()).add(int(ref.model_id))
+        elif isinstance(ref, VoiceRef):
+            if ref.synthetic_voice_id:
+                into.setdefault(SYNTHETIC_VOICE, set()).add(int(ref.synthetic_voice_id))
```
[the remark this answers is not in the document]
SmittieC: Fixed in d33ed6b9. Added a shared `_as_int` helper and routed `accumulate_refs`, `collector._get`, and `events.walk_action`'s `pipeline_id` through it, so a non-numeric stored id resolves to absent instead of raising. `_coerce_int_list` reuses the same helper, and `accumulate_refs` is flattened into an `_ref_ids` generator. Regression test: `test_malformed_node_param_id_does_not_crash`.

apps/api/v2/inspect/builder.py
SmittieC: It is already exposed as `node_id` — the serializers map `node_id = serializers.CharField(source="flow_id")`, and I have regenerated `api-schema.yml` so the public schema reflects `node_id` too (d33ed6b9). The internal `flow_id` key is kept because it mirrors the `PipelineNode.flow_id` DB column.

apps/api/v2/inspect/collector.py
SmittieC: The collector does not render — it resolves ref ids to loaded model instances and wraps them in small value objects (`VoicePair` / `ProviderModelPair` / `CustomActionSelection`) that the DRF response serializers then render. The wrapping exists only so a payload key like `voice` lines up with the value its serializer expects; no dict/JSON serialization happens here (that all lives in `serializers.py`). If the "rendering" connotation is what reads oddly, happy to rename `_wrap_single` — let me know.

apps/api/v2/inspect/events.py
SmittieC: Done in d33ed6b9. Flattened the two nested `is not None` checks into a single guard clause and coerce `pipeline_id` through `_as_int` (it comes from event-action JSON), so it is now one team-scoped query with an early return when there is no pipeline.

SmittieC (commented):
(FYI I'm trying another approach that doesn't use a walker / builder / collector, but serializers only). See https://github.com/dimagi/open-chat-studio/compare/cs/read_api_v2?expand=1

SmittieC:
> Generally LGTM. I haven't thought hard about the collector, walkers design but I don't have any red falgs.

See https://github.com/dimagi/open-chat-studio/pull/3536. I didn't like the 3 stage design, so in the [new approach](https://github.com/dimagi/open-chat-studio/pull/3536) we rely on serializers more heavily.

SmittieC:
Superseded by https://github.com/dimagi/open-chat-studio/pull/3536

Issue #3529, opened 2026-06-03
Deprecate: Surveys


ocs-percy:
**Feature:** Surveys
**Surfaces affected:** `Survey` model + experiment `pre_survey`/`post_survey` FKs; survey CRUD UI/nav/table; `experiment_pre_survey` endpoint + pre-survey chat flow; user docs
**Replacement:** none
**Usage audit:** known used; feature-owner sign-off to remove (no fresh 90-day script run)
**Tier:** full lifecycle (used)
**Announced removal date:** 2026-07-10 (30-day window — below the 60-day default; accepted given owner sign-off)

## Full lifecycle (used)

- [x] Audit: known used; owner sign-off recorded above
- [x] Phase 1 code: experiment coupling severed (FKs, `PENDING_PRE_SURVEY`, routing, pre-survey view/template/form fields); survey CRUD read-only (create blocked, edit view-only, delete allowed); in-feature deprecation warning on survey pages
- [x] Phase 1 code: in-product notification (`survey_deprecation_notification` + `send_survey_deprecation_notifications` command); changelog + user-docs note drafted for the docs repo (banner + email intentionally skipped)
- [x] Day 0 execute: run `send_survey_deprecation_notifications` in prod; publish the changelog/docs note in open-chat-studio-docs
- [ ] Deprecation window (to 2026-07-10): support export requests
- [x] Removal checkpoint: pre-cleared by owner sign-off
- [ ] Phase 2 — drop `Survey` model, nav, table, views, forms, permissions; `DROP COLUMN` the experiment FK columns (`pre_survey_id`/`post_survey_id`, kept in Phase 1 for deploy safety); drop the `Survey` table
- [ ] Close out: remove/redirect user docs pages; close this issue

Migrations follow the two-phase drop (state-only field removal in Phase 1, mirroring `0139`→`0140`; physical column drop in Phase 2).

Spec: `docs/superpowers/specs/2026-06-10-survey-deprecation-phase-1-design.md`
Plan: `docs/superpowers/plans/2026-06-10-survey-deprecation-phase-1.md`
Process: `docs/developer_guides/feature_deprecation.md`

PR #3493, opened 2026-06-01
Design doc: chatbot write APIs

snopoke:
Design doc that complements the read-only API described in https://github.com/dimagi/open-chat-studio/pull/3487

PR #3536, opened 2026-06-04
feat: add v2 chatbots inspect API

SmittieC:
Alternative to #3526 (serializer-centric), built on top of that branch. Same public API and response shape — the difference is entirely in *how* the inspect projection is assembled: serializers own both resolution and rendering, instead of a separate walker → collector → builder stage.

### Product Description
Read-only v2 API for chatbots: `GET /api/v2/chatbots/` (list), `/{id}/` (retrieve), and `/{id}/inspect/?version=` — a single-read, denormalized view of a chatbot's full config (settings, channels, pipeline nodes with inlined resources, event triggers) intended for LLM-agent consumers.

### Technical Description

> **📖 The response is self-documenting — read the generated API docs.**
> Most fields now carry `help_text`, and node `params` are described **per node type** in the
> OpenAPI schema: a `oneOf` over `InspectNodeParams` → `RenderTemplateParams`, `RouterNodeParams`,
> `LLMResponseWithPromptParams`, `CodeNodeParams`, … (see `apps/api/v2/inspect/param_serializers.py`).
> The schema lives at `api-schemas/v1.yml` and is served at `/api/schema/` (Swagger/Redoc UI too).
>
> - **Reviewers:** the field descriptions are the spec for the keys that aren't self-evident from the
>   JSON — e.g. `voice_response_behaviour`, `history_type`/`history_name`, `user_max_token_limit` vs the
>   model's own `max_token_limit`, the router `source_handle` ↔ `keywords` index mapping, and the
>   `static_triggers` vs `timeout_triggers` overlap. Please review meaning against the docs rather than
>   inferring it from the sample payload.
> - **Agent / LLM consumers:** load the schema and read these descriptions to interpret the payload
>   correctly. Raw key names alone are ambiguous; the descriptions are the contract.

Two design points worth reviewer attention, because neither is obvious from the code:

- **A batch fetcher avoids query bloat.** Pipeline node params reference resources (LLM providers, collections, assistants, custom actions, voices, …) by id. Resolving these as the serializers render is a textbook N+1 — a query per node per ref. Instead, `ResourceFetcher` runs a single id-collection pre-pass over the whole graph, batch-loads each resource kind *once* (team-scoped), and serves the serializers via in-memory dict lookups. Caveat to be aware of: the pre-pass and the render pass must walk the same graph (including `pipeline_start`-embedded pipelines) or a resource silently renders `null`; a `django_assert_num_queries` test over an adversarial fixture pins the query count.

- **Knowing which params are resources is the hard part.** Node params are untyped JSON with no FKs, so the model can't tell us which entries are resource references. We carry that knowledge by hand in `RESOURCE_PARAM_FIELDS` (`apps/api/v2/inspect/nodes.py`) — a flat `param field name → ResourceKind` map. `iter_resource_refs` reads only the fields a node type actually declares; list-vs-scalar is sniffed from the value at load time, and grouping fields into render keys (e.g. `llm` drawing from both `llm_provider*` fields) is left to the serializers. We considered adding explicit FK fields per resource on the node models, which would make this self-describing and drop the manual map — but that's a bigger change, so for now the map is maintained by hand. There's no automated completeness guard: a new resource-bearing node field that's missing from the map silently renders as a raw id in `params` instead of being inlined.

Also: resources load team-scoped (cross-team / malformed ids resolve to absent, never crash); secrets are excluded via explicit allowlists; one intentional contract change vs #3526 — a node always renders the resource keys its type declares, using `null`/`[]` when unset.

### Migrations
N/A — no migrations.

### Demo
[Example response in test](https://github.com/dimagi/open-chat-studio/blob/5a53ee73d1bd751c3dd1a4ef193954fb53bf6e8b/apps/api/v2/tests/test_chatbots_inspect.py#L571)

<details>

<summary>Example response from a live setup</summary>

```json
{
    "id": "3df0b056-c591-437f-973f-463b0dff9694",
    "name": "Customer Support Bot",
    "description": "Test pipeline chatbot #1 for development",
    "version_number": 7,
    "is_unreleased": false,
    "is_published_version": true,
    "version_description": "latest",
    "team_slug": "test-team",
    "settings": {
        "seed_message": "Tell a joke to the user",
        "conversational_consent_enabled": true,
        "voice_response_behaviour": "reciprocal",
        "echo_transcript": true,
        "use_processor_bot_voice": false,
        "debug_mode_enabled": false,
        "file_uploads_enabled": true,
        "participant_allowlist": []
    },
    "consent_form": {
        "id": 2,
        "name": "Default Consent",
        "consent_text": "\nWelcome to this chatbot built on Open Chat Studio! \n\nThe Chatbot is provided \"as-is\" and \"as available.\" Open Chat Studio makes no warranties,\nexpress or implied, regarding the Chatbot's accuracy, completeness, or availability.\n\nYou use the chatbot at your own risk. Open Chat Studio shall not be liable for any harm\nor damages that may result from your use of the chatbot.\n\nYou understand and agree that any reliance on the Chatbot's responses is solely at your own\ndiscretion and risk.\n\nBy selecting “I Agree” below, you indicate that: \n\n* You have read and understood the above information.\n* You voluntarily agree to try out this chatbot.\n* You are 18 years or older.\n",
        "capture_identifier": true,
        "identifier_label": "Email Address",
        "identifier_type": "email"
    },
    "voice": {
        "provider_id": 1,
        "provider_name": "OpenAI",
        "type": "openai",
        "voice_name": "onyx",
        "language": "",
        "neural": false
    },
    "trace_provider": {
        "id": 1,
        "type": "langfuse",
        "name": "Langfuse"
    },
    "channels": [
        {
            "platform": "whatsapp",
            "name": "Customer Support Bot",
            "messaging_provider": {
                "id": 2,
                "type": "turnio",
                "name": "Turn.io"
            },
            "number": "+27826412932"
        },
        {
            "platform": "commcare_connect",
            "name": "Customer Support Bot",
            "messaging_provider": null,
            "commcare_connect_bot_name": "Custom Support bot"
        },
        {
            "platform": "facebook",
            "name": "Customer Support Bot",
            "messaging_provider": {
                "id": 1,
                "type": "twilio",
                "name": "Twilio"
            },
            "page_id": "23123123123"
        },
        {
            "platform": "sureadhere",
            "name": "Customer Support Bot",
            "messaging_provider": {
                "id": 3,
                "type": "sureadhere",
                "name": "SureAdhere"
            },
            "sureadhere_tenant_id": "asdq1232123"
        },
        {
            "platform": "embedded_widget",
            "name": "Customer Support Bot",
            "messaging_provider": null,
            "allow_all_domains": false,
            "allowed_domains": [
                "www.example.com"
            ]
        },
        {
            "platform": "web",
            "name": "test-team-web-channel",
            "messaging_provider": null
        },
        {
            "platform": "api",
            "name": "test-team-api-channel",
            "messaging_provider": null
        }
    ],
    "pipeline": {
        "id": 12,
        "name": "Customer Support Pipeline",
        "version_number": 7,
        "graph": {
            "nodes": [
                {
                    "node_id": "RouterNode-NZgtq",
                    "type": "RouterNode",
                    "label": "LLM Router"
                },
                {
                    "node_id": "RenderTemplate-3m8ft",
                    "type": "RenderTemplate",
                    "label": "Render a template"
                },
                {
                    "node_id": "CodeNode-44ZZ7",
                    "type": "CodeNode",
                    "label": "Python Node"
                },
                {
                    "node_id": "LLMResponseWithPrompt-93f8d",
                    "type": "LLMResponseWithPrompt",
                    "label": "LLM"
                },
                {
                    "node_id": "4a2fd796-17a7-4553-921d-ffeb0c49ae7f",
                    "type": "EndNode",
                    "label": ""
                },
                {
                    "node_id": "146772a6-56d4-48ae-9285-e5c860ec9311",
                    "type": "StartNode",
                    "label": ""
                },
                {
                    "node_id": "LLMResponseWithPrompt-QQnwM",
                    "type": "LLMResponseWithPrompt",
                    "label": "LLM"
                }
            ],
            "edges": [
                {
                    "source": "146772a6-56d4-48ae-9285-e5c860ec9311",
                    "target": "RouterNode-NZgtq",
                    "source_handle": "output",
                    "target_handle": null
                },
                {
                    "source": "RouterNode-NZgtq",
                    "target": "LLMResponseWithPrompt-93f8d",
                    "source_handle": "output_0",
                    "target_handle": null
                },
                {
                    "source": "RouterNode-NZgtq",
                    "target": "LLMResponseWithPrompt-QQnwM",
                    "source_handle": "output_1",
                    "target_handle": null
                },
                {
                    "source": "LLMResponseWithPrompt-93f8d",
                    "target": "4a2fd796-17a7-4553-921d-ffeb0c49ae7f",
                    "source_handle": "output",
                    "target_handle": null
                },
                {
                    "source": "LLMResponseWithPrompt-QQnwM",
                    "target": "CodeNode-44ZZ7",
                    "source_handle": "output",
                    "target_handle": null
                },
                {
                    "source": "CodeNode-44ZZ7",
                    "target": "RenderTemplate-3m8ft",
                    "source_handle": "output",
                    "target_handle": null
                },
                {
                    "source": "RenderTemplate-3m8ft",
                    "target": "4a2fd796-17a7-4553-921d-ffeb0c49ae7f",
                    "source_handle": "output",
                    "target_handle": null
                }
            ]
        },
        "nodes": [
            {
                "node_id": "146772a6-56d4-48ae-9285-e5c860ec9311",
                "type": "StartNode",
                "label": "",
                "params": {}
            },
            {
                "node_id": "LLMResponseWithPrompt-QQnwM",
                "type": "LLMResponseWithPrompt",
                "label": "LLM",
                "params": {
                    "tag": "",
                    "tools": null,
                    "prompt": "You are a helpful assistant. Answer the user's query as best you can.",
                    "keywords": [
                        ""
                    ],
                    "mcp_tools": null,
                    "tool_config": null,
                    "history_mode": "summarize",
                    "history_name": null,
                    "history_type": "global",
                    "built_in_tools": null,
                    "generate_citations": true,
                    "max_history_length": 10,
                    "llm_model_parameters": null,
                    "user_max_token_limit": null,
                    "max_indexed_collection_search_results": 20
                },
                "llm": {
                    "provider_id": 1,
                    "provider_name": "OpenAI",
                    "type": "openai",
                    "model": "o4-mini",
                    "max_token_limit": 200000,
                    "deprecated": false
                },
                "voice": null,
                "source_material": null,
                "custom_actions": [],
                "media_collection": null,
                "indexed_collections": []
            },
            {
                "node_id": "RouterNode-NZgtq",
                "type": "RouterNode",
                "label": "LLM Router",
                "params": {
                    "prompt": "You are an extremely helpful router",
                    "keywords": [
                        "A",
                        "B"
                    ],
                    "history_mode": "summarize",
                    "history_name": null,
                    "history_type": "node",
                    "max_history_length": 10,
                    "tag_output_message": false,
                    "llm_model_parameters": null,
                    "user_max_token_limit": null,
                    "default_keyword_index": 0
                },
                "llm": {
                    "provider_id": 1,
                    "provider_name": "OpenAI",
                    "type": "openai",
                    "model": "o4-mini",
                    "max_token_limit": 200000,
                    "deprecated": false
                }
            },
            {
                "node_id": "RenderTemplate-3m8ft",
                "type": "RenderTemplate",
                "label": "Render a template",
                "params": {
                    "tag": "",
                    "keywords": [
                        ""
                    ],
                    "template_string": "this is a template: {input}"
                }
            },
            {
                "node_id": "CodeNode-44ZZ7",
                "type": "CodeNode",
                "label": "Python Node",
                "params": {
                    "tag": "",
                    "code": "# You must define a main function, which takes the node input as a string.\n# Return a string to pass to the next node.\n\n# Learn more about Python nodes at https://docs.openchatstudio.com/concepts/pipelines/nodes/#python-node\n\ndef main(input: str, **kwargs) -> str:\n    return input\n",
                    "keywords": [
                        ""
                    ]
                }
            },
            {
                "node_id": "LLMResponseWithPrompt-93f8d",
                "type": "LLMResponseWithPrompt",
                "label": "LLM",
                "params": {
                    "tools": [
                        "append-to-participant-data",
                        "update-user-data"
                    ],
                    "prompt": "You are a helpful assistant. Answer the user's query as best you can.\n\n# User data\n{participant_data}\n\n\nHere's source material: {source_material}.\n\nHere's media you can share:\n{media}",
                    "keywords": [
                        ""
                    ],
                    "history_mode": "summarize",
                    "history_name": null,
                    "history_type": "global",
                    "built_in_tools": [
                        "web-search"
                    ],
                    "llm_temperature": 0.7,
                    "generate_citations": true,
                    "max_history_length": 10,
                    "user_max_token_limit": 1000000,
                    "max_indexed_collection_search_results": 20
                },
                "llm": {
                    "provider_id": 1,
                    "provider_name": "OpenAI",
                    "type": "openai",
                    "model": "gpt-4.1-mini",
                    "max_token_limit": 1000000,
                    "deprecated": false
                },
                "voice": {
                    "provider_id": null,
                    "provider_name": null,
                    "type": null,
                    "voice_name": "nova",
                    "language": "",
                    "neural": false
                },
                "source_material": {
                    "id": 2,
                    "topic": "Red schemes",
                    "description": "This is a test source material description",
                    "material": "This is source material"
                },
                "custom_actions": [
                    {
                        "id": 1,
                        "name": "My custom action",
                        "description": "My custom action description",
                        "server_url": "http://localhost:8000/api",
                        "allowed_operations": [
                            "users_get"
                        ],
                        "api_schema": {
                            "paths": [
                                "/users"
                            ]
                        },
                        "auth_provider": null
                    }
                ],
                "media_collection": {
                    "id": 2,
                    "name": "My media collection",
                    "files": [
                        {
                            "id": 5,
                            "name": "annotated_form.pdf",
                            "content_type": "application/pdf",
                            "content_size": 47398,
                            "external_source": "",
                            "external_id": ""
                        },
                        {
                            "id": 6,
                            "name": "meta_pricing.pdf",
                            "content_type": "application/pdf",
                            "content_size": 1864790,
                            "external_source": "",
                            "external_id": ""
                        }
                    ]
                },
                "indexed_collections": [
                    {
                        "id": 3,
                        "name": "My indexed collection",
                        "embedding": {
                            "provider_id": 1,
                            "provider_name": "OpenAI",
                            "type": "openai",
                            "model": "text-embedding-3-small"
                        },
                        "files": [
                            {
                                "id": 7,
                                "name": "meta_pricing.pdf",
                                "content_type": "application/pdf",
                                "content_size": 1864790,
                                "external_source": "",
                                "external_id": ""
                            },
                            {
                                "id": 8,
                                "name": "Old Testament Narratives Course Guide.docx",
                                "content_type": "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
                                "content_size": 20948,
                                "external_source": "",
                                "external_id": ""
                            }
                        ]
                    }
                ]
            },
            {
                "node_id": "4a2fd796-17a7-4553-921d-ffeb0c49ae7f",
                "type": "EndNode",
                "label": "",
                "params": {}
            }
        ]
    },
    "events": {
        "static_triggers": [
            {
                "id": 5,
                "type": "last_timeout",
                "is_active": true,
                "action": {
                    "type": "pipeline_start",
                    "params": {
                        "input_type": "history_last_summary"
                    },
                    "pipeline": {
                        "id": 8,
                        "name": "Goodbye Pipe",
                        "version_number": 1,
                        "graph": {
                            "nodes": [
                                {
                                    "node_id": "b138a79e-e66f-4b17-973d-f09373884ce0",
                                    "type": "StartNode",
                                    "label": ""
                                },
                                {
                                    "node_id": "3a99bc2c-e73c-43e8-986b-4a58b81d86b4",
                                    "type": "EndNode",
                                    "label": ""
                                },
                                {
                                    "node_id": "LLMResponseWithPrompt-rI84j",
                                    "type": "LLMResponseWithPrompt",
                                    "label": "LLM"
                                }
                            ],
                            "edges": [
                                {
                                    "source": "b138a79e-e66f-4b17-973d-f09373884ce0",
                                    "target": "LLMResponseWithPrompt-rI84j",
                                    "source_handle": "output",
                                    "target_handle": null
                                },
                                {
                                    "source": "LLMResponseWithPrompt-rI84j",
                                    "target": "3a99bc2c-e73c-43e8-986b-4a58b81d86b4",
                                    "source_handle": "output",
                                    "target_handle": null
                                }
                            ]
                        },
                        "nodes": [
                            {
                                "node_id": "b138a79e-e66f-4b17-973d-f09373884ce0",
                                "type": "StartNode",
                                "label": "",
                                "params": {}
                            },
                            {
                                "node_id": "LLMResponseWithPrompt-rI84j",
                                "type": "LLMResponseWithPrompt",
                                "label": "LLM",
                                "params": {
                                    "tag": "",
                                    "tools": null,
                                    "prompt": "Tell the user goodbye",
                                    "keywords": [
                                        ""
                                    ],
                                    "mcp_tools": null,
                                    "tool_config": null,
                                    "history_mode": "summarize",
                                    "history_name": null,
                                    "history_type": "global",
                                    "built_in_tools": null,
                                    "generate_citations": true,
                                    "max_history_length": 10,
                                    "llm_model_parameters": null,
                                    "user_max_token_limit": null,
                                    "max_indexed_collection_search_results": 20
                                },
                                "llm": {
                                    "provider_id": 1,
                                    "provider_name": "OpenAI",
                                    "type": "openai",
                                    "model": "o4-mini",
                                    "max_token_limit": 200000,
                                    "deprecated": false
                                },
                                "voice": null,
                                "source_material": null,
                                "custom_actions": [],
                                "media_collection": null,
                                "indexed_collections": []
                            },
                            {
                                "node_id": "3a99bc2c-e73c-43e8-986b-4a58b81d86b4",
                                "type": "EndNode",
                                "label": "",
                                "params": {}
                            }
                        ]
                    }
                }
            }
        ],
        "timeout_triggers": [
            {
                "id": 5,
                "delay_seconds": 3600,
                "total_num_triggers": 1,
                "trigger_from_first_message": true,
                "is_active": true,
                "action": {
                    "type": "end_conversation",
                    "params": {}
                }
            }
        ]
    }
}
```

</details>

### Docs and Changelog


apps/api/v2/inspect/nodes.py
```diff
@@ -0,0 +1,97 @@
...
+    "voice": ResourceField(frozenset({"voice_provider_id", "synthetic_voice_id"}), ResourceKind.SYNTHETIC_VOICE, False),
+    "source_material": ResourceField(frozenset({"source_material_id"}), ResourceKind.SOURCE_MATERIAL, False),
+    "assistant": ResourceField(frozenset({"assistant_id"}), ResourceKind.ASSISTANT, False),
+    "custom_actions": ResourceField(frozenset({"custom_actions"}), ResourceKind.CUSTOM_ACTION, True),
+    "media_collection": ResourceField(frozenset({"collection_id"}), ResourceKind.COLLECTION, False),
+    "indexed_collections": ResourceField(frozenset({"collection_index_ids"}), ResourceKind.COLLECTION, True),
+}
+
+# The payload keys, in render order — used by the node serializer's ``to_representation``.
+RESOURCE_KEYS: tuple[str, ...] = (
```
snopoke: Could this just be: `RESOURCE_KEYS = tuple(RESOURCE_FIELDS)`

apps/api/v2/inspect/resources.py
```diff
@@ -0,0 +1,186 @@
...
+    multi-source keys (``voice``) and list/parsed keys are handled per payload key; everything is
+    driven by which keys the node type declares (``declared_resource_keys``)."""
+    params = params or {}
+    for key in declared_resource_keys(node_class_for(node_type)):
+        rf = RESOURCE_FIELDS[key]
+        if key == "llm":
+            yield ResourceKind.LLM_PROVIDER, params.get("llm_provider_id")
+            yield ResourceKind.LLM_PROVIDER_MODEL, params.get("llm_provider_model_id")
+        elif key == "voice":
+            yield ResourceKind.SYNTHETIC_VOICE, params.get("synthetic_voice_id")
```
snopoke: should this also include the voice provider param?

apps/api/v2/inspect/resources.py
```diff
@@ -0,0 +1,186 @@
...
+        fetcher._load(ids)
+        return fetcher
+
+    def _collect_embedded_pipeline(self, action, ids: dict[ResourceKind, set[int]]) -> None:
+        if action.action_type != EventActionType.PIPELINE_START:
+            return
+        pipeline_id = _as_int((action.params or {}).get("pipeline_id"))
+        if not pipeline_id or pipeline_id in self._pipelines:
+            return
+        # Imported here to avoid an import cycle at module load (pipelines -> api -> pipelines).
```
snopoke: I question this logic

apps/api/v2/views.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+    @action(detail=True, methods=["get"])
+    def inspect(self, request, id=None):
+        """Return a denormalized, read-only projection of the chatbot's full configuration."""
+        family = self.get_object()
+        try:
+            target = resolve_inspect_version(family, request.query_params.get("version"))
+        except InspectVersionError as err:
+            raise NotFound("Requested chatbot version was not found.") from err
+        target = prefetch_inspect_target(target)
```
snopoke: It's unfortunate that we do 3 queries to resolve the experiment - maybe worth considering an optimization.

snopoke (commented):
No major comments.

> We considered adding explicit FK fields per resource on the node models, which would make this self-describing and drop the manual map — but that's a bigger change, so for now the list has to be maintained by hand.

I was thinking of picking this up next week.

Also FYI, we could drop surveys from this API since we'll be removing them soon: https://github.com/dimagi/open-chat-studio/issues/3529

apps/api/v2/inspect/resources.py
```diff
@@ -0,0 +1,186 @@
...
+            )
+        if kind == ResourceKind.LLM_PROVIDER:
+            return LlmProvider.objects.filter(team=team, id__in=ids)
+        if kind == ResourceKind.LLM_PROVIDER_MODEL:
+            return LlmProviderModel.objects.filter(Q(team=team) | Q(team__isnull=True), id__in=ids)
+        if kind == ResourceKind.VOICE_PROVIDER:
+            return VoiceProvider.objects.filter(team=team, id__in=ids)
+        if kind == ResourceKind.SYNTHETIC_VOICE:
+            return SyntheticVoice.objects.filter(id__in=ids).select_related("voice_provider")
+        assert_never(kind)
```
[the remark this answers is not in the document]
SmittieC: Fixed in 18e5cdd67. The SYNTHETIC_VOICE branch now uses `SyntheticVoice.get_for_team(team).filter(id__in=ids).select_related("voice_provider")`, applying the team-or-general condition (so global/general voices are kept) instead of the unscoped `objects.filter(id__in=ids)`. Added a regression test, `test_cross_team_synthetic_voice_not_embedded`, confirming a voice backed by a provider owned by another team resolves to absent.

apps/api/v2/inspect/resources.py
```diff
@@ -0,0 +1,186 @@
...
+        elif key == "custom_actions":
+            for action_id, _operation_ids in parse_custom_actions(params.get("custom_actions")):
+                yield ResourceKind.CUSTOM_ACTION, action_id
+        elif rf.is_list:
+            field = next(iter(rf.consumes))
+            for raw_id in params.get(field) or []:
+                yield rf.kind, raw_id
+        else:
+            field = next(iter(rf.consumes))
+            yield rf.kind, params.get(field)
```
SmittieC: Fixed in acc8d7e46. Extracted the per-key dispatch into a `_refs_for_key` helper, so `iter_resource_refs` is now a thin driver over the declared keys and drops below the cyclomatic-complexity threshold.

SmittieC: Fixed in acc8d7e46. Moving the nested per-key branches into `_refs_for_key` flattens `iter_resource_refs` to a single `yield from` loop, removing the nested conditional blocks.

apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -0,0 +1,867 @@
+import json
```
SmittieC, line 1: This is a test module; the length reflects intentionally comprehensive coverage (auth, cross-team isolation, secret-exclusion, schema validation, query-budget, and the two-layer completeness guards). Splitting it purely to hit the 600-LOC metric would fragment cohesive test groups, so leaving it as-is. (no code change needed)

apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -0,0 +1,867 @@
...
+                    "id": bot.timeout_trigger.id,
+                    "delay_seconds": 86400,
+                    "total_num_triggers": 1,
+                    "trigger_from_first_message": False,
+                    "is_active": True,
+                    "action": {"type": "send_message_to_bot", "params": {"message_to_bot": "Still there?"}},
+                }
+            ],
+        },
+    }
```
SmittieC, line 759: `test_full_response_body` is a single golden-body assertion over the complete inspect response; its length is the snapshot itself. Breaking it apart would obscure the single-snapshot intent, so leaving it as-is. (no code change needed)

apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -0,0 +1,867 @@
...
+        source=source,
+        media_collection=media_collection,
+        media_file=media_file,
+        index_collection=index_collection,
+        index_file=index_file,
+        action=action,
+        assistant=assistant,
+        schedule_trigger=schedule_trigger,
+        timeout_trigger=timeout_trigger,
+    )
```
SmittieC, line 561: `_full_bot` is one cohesive fixture builder for the fully-populated bot, 6 lines over the threshold. Splitting it would scatter the fixture setup across helpers with no real readability gain, so leaving it as-is. (no code change needed)

apps/api/v2/inspect/resources.py
SmittieC: The comment? in that case, it is not a valid comment. I removed it in https://github.com/dimagi/open-chat-studio/pull/3536/changes/1108d361dd1114765fd30c58b369aef793052bb8

apps/api/v2/inspect/resources.py
SmittieC: `voice_provider_id` should not be there. https://github.com/dimagi/open-chat-studio/pull/3536/changes/1108d361dd1114765fd30c58b369aef793052bb8

apps/api/v2/inspect/nodes.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/3536/changes/1108d361dd1114765fd30c58b369aef793052bb8

apps/api/v2/inspect/resources.py
snopoke: I mean I doubt the logic in the comment is correct - I find I have to question claudes 'dependnecy cycle' logic often.

apps/api/v2/inspect/resources.py
```diff
@@ -0,0 +1,197 @@
...
+            operation_ids.append(operation_id)
+    return list(selections.items())
+
+
+def _refs_for_key(key: str, rf: ResourceField, params: dict):
+    """Expand one declared payload key into its ``(ResourceKind, raw_id)`` refs. Splitting this out
+    keeps ``iter_resource_refs`` a thin driver over the declared keys (one dispatch per key)."""
+    if key == "llm":
+        yield ResourceKind.LLM_PROVIDER, params.get("llm_provider_id")
+        yield ResourceKind.LLM_PROVIDER_MODEL, params.get("llm_provider_model_id")
```
snopoke: I'm not quite sure what the reason is for having one resource field that consumes both of these params if we end up splitting it out here anyway. This is the only one that consumes multiple fields so removing it would simplify things. The combined serialization is already handled elsewhere

SmittieC: This was all simplified in https://github.com/dimagi/open-chat-studio/pull/3536/changes/98138ad9fabf37fb346081ea9d10eb7315d1286a#diff-d249bc4502a265c12002575075e5e5b53fab395bad192d85befd8fe7d8f6681d onwards

SmittieC:
@coderabbitai full review

apps/api/v2/inspect/serializers.py
```diff
@@ -0,0 +1,574 @@
...
+
+
+@extend_schema_serializer(component_name="InspectFile")
+class FileSerializer(serializers.ModelSerializer):
+    """Identity-lean file view. Excludes the signed ``file`` storage URL, ``summary`` and
+    ``metadata`` (the latter two are dropped for size, not secrecy — design D8/Q6)."""
+
+    class Meta:
+        model = File
+        fields = ["id", "name", "content_type", "content_size", "external_source", "external_id", "purpose"]
```
SmittieC: @snopoke [here's](https://github.com/dimagi/open-chat-studio/blob/5a53ee73d1bd751c3dd1a4ef193954fb53bf6e8b/apps/api/v2/tests/test_chatbots_inspect.py#L571) an example of the response body. I'm not sure if we want anything but the file names in here? Maybe the ids as well, but the rest doesn't seem useful, unless we see this API as a "deep inspect" kind of API i.e. rather give more than too little info?

apps/api/v2/inspect/tests/test_node_rendering.py
```diff
@@ -0,0 +1,125 @@
...
+    assert data["llm"] is None
+    assert data["voice"] is None
+    assert data["source_material"] is None
+    assert data["media_collection"] is None
+    assert data["custom_actions"] == []
+    assert data["indexed_collections"] == []
+    assert "assistant" not in data  # LLMResponseWithPrompt does not declare assistant
+
+
+def test_llm_provider_model_id_not_leaked_into_params():
```
SmittieC: remove this test. Not valuable

SmittieC: Done — removed in 0b7b5a3b1.

api-schemas/v1.yml
```diff
@@ -1518,6 +1858,417 @@ components:
...
+      - custom_actions
+      - indexed_collections
+      - label
+      - llm
+      - media_collection
+      - node_id
+      - params
+      - source_material
+      - type
+      - voice
```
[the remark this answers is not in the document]
SmittieC: Already addressed on this branch. The `InspectNodeSchemaExtension` / `InspectTriggerActionSchemaExtension` (`_ConditionalRequiredSchemaMixin`) drop each serializer's conditional keys from `required`. Current schema: `InspectNode.required` = [label, node_id, params, type] and `InspectTriggerAction.required` = [params, type], so the conditionally-present fields are already optional.

apps/api/v2/tests/test_utils.py
```diff
@@ -0,0 +1,18 @@
...
+    assert as_int("abc") is None
+    assert as_int(None) is None
+    assert as_int("") is None
+
+
+def test_parse_custom_actions_groups_operations_per_action():
+    assert parse_custom_actions(["3:weather_get", "3:pollen_get", "5:x"]) == [
+        (3, ["weather_get", "pollen_get"]),
+        (5, ["x"]),
+    ]
```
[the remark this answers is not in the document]
SmittieC: Already addressed — `test_utils.py` already asserts `as_int(True) is None` and `as_int(False) is None`, and `test_parse_custom_actions_handles_scalar_string` covers `parse_custom_actions("12:op_a") == [(12, ["op_a"])]`.

apps/api/v2/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    selections: dict[int, list[str]] = {}
+    for entry in value or []:
+        action_part, _, operation_id = str(entry).partition(":")
+        action_id = as_int(action_part)
+        if action_id is None:
+            continue
+        operation_ids = selections.setdefault(action_id, [])
+        if operation_id and operation_id not in operation_ids:
+            operation_ids.append(operation_id)
+    return list(selections.items())
```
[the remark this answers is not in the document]
SmittieC: Already addressed — `as_int` rejects booleans (`isinstance(value, bool)` -> `None`) and `parse_custom_actions` normalizes a bare string to a single entry instead of iterating it character-by-character.

docs/superpowers/specs/2026-06-04-chatbot-inspect-serializer-refactor-design.md
```diff
@@ -0,0 +1,452 @@
...
+)
+```
+
+`resolve_inspect_version`, `InspectVersionError`, and `prefetch_inspect_target` are
+pure target-resolution/preparation helpers (not collector/builder) — they live in a
+small `versioning.py`.
+
+### Serializer tree
+
+```
```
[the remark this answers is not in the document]
SmittieC: Fixed in 3baede215 — labeled both fenced blocks as `text` to satisfy MD040.

apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -0,0 +1,865 @@
...
+                    {"node_id": "assist", "type": "AssistantNode", "label": "Assistant"},
+                ],
+                "edges": [{"source": "llm", "target": "assist", "source_handle": "output", "target_handle": "input"}],
+            },
+            "nodes": [
+                {
+                    "node_id": "llm",
+                    "type": "LLMResponseWithPrompt",
+                    "label": "Answer",
+                    "params": {"prompt": "Answer the user"},
```
SmittieC, line 635: @snopoke thoughts on keeping non-resources (non-FK) entries inside a `params` object instead of surfacing them to the same level as label, type etc?

See the PR description for a better example.

snopoke: I think it's fine to keep them in 'params'. The example in the PR description has more fields:

```
 "params": {
                    "tag": "",
                    "tools": null,
                    "prompt": "You are a helpful assistant. Answer the user's query as best you can.",
                    "keywords": [
                        ""
                    ],
                    "mcp_tools": null,
                    "max_results": 20,
                    "tool_config": null,
                    "history_mode": "summarize",
                    "history_name": null,
                    "history_type": "global",
                    "built_in_tools": null,
                    "generate_citations": true,
                    "max_history_length": 10,
                    "llm_model_parameters": null,
                    "user_max_token_limit": null
                },
```

Not sure why an LLM node has `keywords`.
Also `max_results` is only relevant if there are indexed collections and might need a better name like `max_collection_search_results`?
What is `tool_config` for? Is that for `built_in_tools`?

apps/api/v2/inspect/serializers.py
snopoke: Seems OK to keep it all except 'purpose'.

apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -0,0 +1,865 @@
...
+
+# Distinctive secret values seeded into provider configs / channel extra_data. None of these may
+# appear anywhere in the response.
+LLM_SECRET = "sk-llm-SECRET-xyz"
+AUTH_SECRET = "auth-SECRET-xyz"
+CHANNEL_SECRET = "telegram-token-SECRET-xyz"
+
+
+@pytest.fixture()
+def inspect_bot(db):
```
snopoke: you could make this a module scoped fixture to avoid re-creating it for every test. Here's an example: https://github.com/dimagi/open-chat-studio/blob/main/apps/teams/tests/test_clone_team.py#L29

apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -0,0 +1,865 @@
...
+    channels = _get(inspect_bot)["channels"]
+
+    assert [c["platform"] for c in channels] == ["telegram"]
+    assert not ExperimentChannel.objects.filter(
+        team=team, platform__in=[ChannelPlatform.WEB, ChannelPlatform.API]
+    ).exists()
+
+
+@pytest.mark.django_db()
+def test_channel_allowlisted(inspect_bot):
```
snopoke: what is this testing?

apps/api/v2/serializers.py
```diff
@@ -0,0 +1,27 @@
...
+from apps.experiments.models import Experiment
+
+
+class ChatbotVersionSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Experiment
+        fields = ["name", "version_number", "is_default_version", "version_description"]
+
+
+class ChatbotSerializer(serializers.ModelSerializer):
```
snopoke, line 12: I have some thoughts about the data for this view but we can do that in a followup.

(https://github.com/dimagi/open-chat-studio/pull/3493/changes#r3333601130)

apps/api/v2/tests/test_chatbots_inspect.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/3536/commits/abfa0a6cc912cbeaff9344ee41368c654a2bb115

apps/api/v2/tests/test_chatbots_inspect.py
SmittieC: It's actually a redundant test that makes sure that only the allowed fields are rendered for the channels for the fixture. I'll remove, since it's redundant, considering the `test_channel_identifier_fields_per_platform` test.

SmittieC: https://github.com/dimagi/open-chat-studio/pull/3536/commits/494410ae17d1c4e5907d9853156044bcd81e071b

apps/api/v2/tests/test_chatbots_inspect.py
SmittieC: > Not sure why an LLM node has keywords

Good question. Turn out that this is a mini-bug. See [here](https://github.com/dimagi/open-chat-studio/blob/6dc02aec11a830aa24104e91673181ce2ed1643e/assets/javascript/apps/pipeline/nodes/GetInputWidget.tsx#L190). We initialize keywords on all node's params. Yikes.

> What is tool_config for? Is that for built_in_tools?

yeah

> Also max_results is only relevant if there are indexed collections and might need a better name like max_collection_search_results?

Correct. This touches a design question actually. Do we want to return all params verbatim that are not resource params (see [here](https://github.com/dimagi/open-chat-studio/blob/512841cd8bde52e7f6dc0abf20773eb4fd097b2a/apps/api/v2/inspect/serializers.py#L464-L468)), or should we be more clever in what we return, based on how the node is configured? My take was verbatim, but we can easily change it.

One thing to consider though is that this is another "display layer" that we'd have to manage e.g. "show this when that is configured." We _can_ add code similar to [this](https://github.com/dimagi/open-chat-studio/blob/6dc02aec11a830aa24104e91673181ce2ed1643e/assets/javascript/apps/pipeline/nodes/GetInputWidget.tsx#L12-L23) that understands the `visible_when` params...

snopoke: I think it's ok to leave the params as 'raw' for now.

SmittieC:
@snopoke https://github.com/dimagi/open-chat-studio/pull/3536/commits/e77957ee9df23f0b639f3bd80442e281fa9e058d adds more documentation in the api docs so agents can understand what those parameters mean

apps/api/v2/inspect/serializers.py
```diff
@@ -0,0 +1,724 @@
...
+        if action.action_type != EventActionType.PIPELINE_START:
+            return None
+        pipeline = self._fetcher.embedded_pipeline((action.params or {}).get("pipeline_id"))
+        if pipeline is None:
+            return None
+        return InspectPipelineSerializer(pipeline, context=self.context).data
+
+
+class InspectStaticTriggerSerializer(serializers.ModelSerializer):
+    action = InspectTriggerActionSerializer()
```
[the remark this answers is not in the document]
SmittieC: 👎 False positive — the premise is backwards.

DRF **does** auto-propagate context to nested serializers declared as class-attribute fields. `Field.context` returns `self.root._context`, and a declared field is bound to its parent (`field.parent = self`), so it walks up to the root serializer's context. The empty-`{}` gotcha the bot describes only applies to serializers **manually instantiated inside a method** (e.g. a `SerializerMethodField`) without passing `context=` — and we already pass `context=self.context` in every such spot (lines 613, 664, 712, 717, 782). The `action = InspectTriggerActionSerializer()` declarations are declared fields, so they inherit context for free.

Verified empirically: rendering both `InspectStaticTriggerSerializer` and `InspectTimeoutTriggerSerializer` for a `pipeline_start` action with only the parent given `context={"fetcher": ...}` raises no `RuntimeError`, and `"fetcher" in <nested action field>.context` holds. No change needed.

apps/api/v2/tests/test_chatbots_inspect.py
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/3536/changes/3164aceb569ed062158aa339203804953cbae287 for max_results -> max_indexed_collection_search_results rename

apps/api/v2/inspect/serializers.py
```diff
@@ -0,0 +1,813 @@
...
+        if action.action_type != EventActionType.PIPELINE_START:
+            return None
+        pipeline = self._fetcher.embedded_pipeline((action.params or {}).get("pipeline_id"))
+        if pipeline is None:
+            return None
+        return InspectPipelineSerializer(pipeline, context=self.context).data
+
+
+class InspectStaticTriggerSerializer(serializers.ModelSerializer):
+    type = serializers.CharField(
```
snopoke: Overriding this results in the loss of the type enum values in the output schema. I think if you use `extra_kwargs` to add the help text it will preserve the enum values.

https://www.django-rest-framework.org/api-guide/serializers/#additional-keyword-arguments

apps/api/v2/inspect/serializers.py
```diff
@@ -0,0 +1,813 @@
...
+
+    The ``pipeline`` key only appears for ``pipeline_start`` actions whose pipeline could be found in
+    the team. It's read from the pipelines the fetcher already loaded, so it costs no extra query.
+    """
+
+    # ``pipeline`` is dropped by ``to_representation`` for non-pipeline_start actions, so the schema
+    # extension marks it optional (see ``_ConditionalRequiredSchemaMixin``).
+    CONDITIONAL_RESPONSE_KEYS = ("pipeline",)
+
+    type = serializers.CharField(
```
snopoke: same comment about trying `extra_kwargs` to preserve enum values

apps/api/v2/inspect/serializers.py
SmittieC: Good catch. See https://github.com/dimagi/open-chat-studio/pull/3536/commits/e3b80976ccf26766961ca9cd5bcfa40e9f797f8e

apps/api/v2/inspect/serializers.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/3536/commits/e3b80976ccf26766961ca9cd5bcfa40e9f797f8e

apps/api/v2/inspect/channels.py
```diff
@@ -0,0 +1,26 @@
...
+        )
+    )
+    team = experiment.team
+    for platform, name in (
+        (ChannelPlatform.WEB, f"{team.slug}-web-channel"),
+        (ChannelPlatform.API, f"{team.slug}-api-channel"),
+    ):
+        channel = ExperimentChannel.objects.filter(team=team, platform=platform, name=name).first()
+        if channel is not None:
+            channels.append(channel)
```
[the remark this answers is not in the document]
SmittieC: Fixed in b650624 — added `.select_related("messaging_provider")` to the team-wide WEB/API channel lookup so it's consistent with the experiment-channel query and avoids the extra per-channel provider fetch. Resolving.

snopoke (dismissed):

PR #3549, opened 2026-06-05
Auto-configure Twilio WhatsApp webhooks

snopoke:
### Product Description
WhatsApp channels using a Twilio provider now have their webhook configured at Twilio automatically when the channel is created or edited, and cleared on delete. If auto-configuration fails (e.g. sandbox number), the existing manual setup instructions are shown instead.

### Technical Description
Uses Twilio's WhatsApp Senders API (`messaging.v2.channels_senders`) via a new `supports_webhook_management` capability on `MessagingService` (Twilio only; Meta/Turn webhooks are app-level).

Points worth review attention:
- Sender updates send a minimal `{"webhook": ...}` payload (`TwilioSenderWebhookUpdate`) because the SDK's generated classes serialize unset fields as JSON nulls, which risks clearing sender config.
- Webhook removal on channel delete only happens if the sender's callback URL still points at the channel being deleted, and never blocks deletion.
- Also includes a one-off management command (`migrate_twilio_webhooks`) to migrate existing webhooks off the chatbots.dimagi.com domain. Dry-run by default; it also normalizes the legacy `/channels/whatsapp/incoming_message` path to `/channels/twilio/incoming_message`.

### Docs and Changelog


apps/service_providers/messaging_service.py
```diff
@@ -223,6 +249,36 @@ def _parse_addressing_params(self, platform: ChannelPlatform, from_: str, to: st
...
+
+    def set_incoming_webhook(self, number: str, webhook_url: str):
+        sender = self._get_whatsapp_sender(number)
+        if sender is None:
+            raise ValueError(f"No WhatsApp sender found for {number}")
+        self._update_sender_webhook(sender, webhook_url)
+
+    def remove_incoming_webhook(self, number: str, webhook_url: str):
+        sender = self._get_whatsapp_sender(number)
+        if sender is None or (sender.webhook or {}).get("callback_url") != webhook_url:
```
snopoke: Fixed in 2164b7bc9. Split the compound conditional into separate guard clauses.

*— Claude Code*

snopoke: Fixed in 2164b7bc9. Split the compound conditional into separate guard clauses.

*— Claude Code*

apps/service_providers/tests/test_messaging_providers.py
```diff
@@ -755,3 +755,82 @@ def _test_messaging_provider(team, provider_type: MessagingProviderType, data):
         type=provider_type,
         config=form.cleaned_data,
     )
+
```
snopoke: Fixed in 2164b7bc9. Moved the new webhook management tests into their own file (test_twilio_webhook_management.py).

*— Claude Code*

apps/channels/views.py
```diff
@@ -429,6 +430,20 @@ def delete_channel(request, team_slug, experiment_id: int, channel_id: int):
...
+        return
+    service = channel.messaging_provider.get_messaging_service()
+    number = (channel.extra_data or {}).get("number")
+    if not service.supports_webhook_management or not number:
+        return
+    try:
+        service.remove_incoming_webhook(number, channel.webhook_url)
+    except Exception:
+        log.exception("Error removing webhook for channel %s", channel.id)
+
```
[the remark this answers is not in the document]
snopoke: Fixed in 2164b7bc9. Moved service construction and the capability check inside the try block so deletion is never blocked.

*— Claude Code*

apps/service_providers/messaging_service.py
```diff
@@ -117,6 +117,32 @@ def resolve_number(self, number: str) -> str | None:
         """Returns `number` if the number is verified to belong to the account, otherwise return `None`"""
         return number
 
+    supports_webhook_management: ClassVar[bool] = False
+    """Whether this service can configure the inbound message webhook at the provider."""
+
+    def set_incoming_webhook(self, number: str, webhook_url: str):
```
snopoke: This should accept the full 'channel.extra_data' dict since different channels may require different set's of data for the webhook.

snopoke: Fixed in db435c566. `set_incoming_webhook`/`remove_incoming_webhook` now take the full `extra_data` dict; `TwilioService` reads `extra_data["number"]` itself, so other providers can read whatever fields they need.

*— Claude Code*

apps/channels/views.py
```diff
@@ -429,6 +430,20 @@ def delete_channel(request, team_slug, experiment_id: int, channel_id: int):
...
 
 
+def _clear_remote_webhook(channel: ExperimentChannel):
+    """Best-effort removal of the channel's webhook configuration at the messaging provider."""
+    if not channel.messaging_provider:
+        return
+    try:
+        service = channel.messaging_provider.get_messaging_service()
+        extra_data = channel.extra_data or {}
+        if not service.supports_webhook_management or not extra_data.get("number"):
```
snopoke: This shouldn't check the presence of 'number'. That should be pushed down to the service.

snopoke: Fixed in 406f72f1d. Removed the number check from the view; `remove_incoming_webhook` now returns early when `extra_data` has no number.

*— Claude Code*

apps/service_providers/messaging_service.py
```diff
@@ -223,6 +252,43 @@ def _parse_addressing_params(self, platform: ChannelPlatform, from_: str, to: st
         prefix = self.TWILIO_CHANNEL_PREFIXES[platform]
         return f"{prefix}:{from_}", f"{prefix}:{to}"
 
+    supports_webhook_management: ClassVar[bool] = True
+
+    def set_incoming_webhook(self, extra_data: dict, webhook_url: str):
+        number = extra_data["number"]
```
snopoke:
```suggestion
        number = extra_data.get("number")
        if not number:
            return
```

SmittieC (approved):
Sweet

PR #3550, opened 2026-06-05
Return connect_channel_id from the update participant API

snopoke:
### Product Description
`POST /api/participants` now returns the participant detail, including the CommCare Connect channel ID for each chatbot, so callers no longer need to poll for it.

### Technical Description
- Connect channels are created synchronously during the request (previously a Celery task), so the IDs are available in the response. Upstream failures map to 404/503/400; participant data stays saved and retries are idempotent.
- `connect_channel_id` added to the participant data serializer (also appears in GET).
- `setup_connect_channels_for_bots` is no longer queued; kept deprecated for in-flight tasks, removal to follow.

### Migrations

No migrations.

### Demo
N/A — see updated API schema examples.

### Docs and Changelog

### DIFF
--- a/apps/api/views/participants.py
+++ b/apps/api/views/participants.py
@@ -2,12 +2,12 @@ import uuid
 
 from django.db import transaction
 from django.db.models import Prefetch
-from django.http import HttpResponse
 from django.utils import timezone
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiExample, OpenApiParameter, extend_schema
 from rest_framework.exceptions import NotFound, ValidationError
 from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from apps.api.pagination import CursorPagination
@@ -96,7 +96,7 @@ class ParticipantView(APIView):
         summary="Update Participant Data",
         tags=["Participants"],
         request=ParticipantDataUpdateRequest(),
-        responses={200: {}},
+        responses={200: ParticipantDetailSerializer},
         examples=[
             OpenApiExample(
                 name="CreateParticipantData",
@@ -199,7 +199,7 @@ def _update_participant_data(request):
     if platform == ChannelPlatform.COMMCARE_CONNECT:
         setup_connect_channels_for_bots.delay(connect_id=identifier, experiment_data_map=experiment_data_map)
 
-    return HttpResponse()
+    return Response(ParticipantDetailSerializer(participant).data)
 
 
 def _get_participant_experiments(team, experiment_data) -> dict[str, Experiment]:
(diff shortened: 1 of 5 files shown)


SmittieC (dismissed):
Thoughts on returning the `connect_channel_id` only for connect channels?

apps/api/views/participants.py
```diff
@@ -194,12 +232,12 @@ def _update_participant_data(request):
...
-            experiment_data_map[experiment.id] = participant_data.id
+            connect_data_by_experiment_id[experiment.id] = participant_data
 
-    if platform == ChannelPlatform.COMMCARE_CONNECT:
-        setup_connect_channels_for_bots.delay(connect_id=identifier, experiment_data_map=experiment_data_map)
+    if connect_data_by_experiment_id:
+        _setup_connect_channels(identifier, connect_data_by_experiment_id)
 
-    return HttpResponse()
+    return Response(ParticipantDetailSerializer(participant).data)
```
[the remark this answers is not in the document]
snopoke: Fixed in 19ababfe3. Added `.select_related("experiment")` to the `get_data` fallback query in `ParticipantDetailSerializer`, so the POST response no longer triggers an N+1 when resolving `experiment.name` / `experiment.public_id` per entry.

*— Claude Code*

apps/api/views/participants.py
```diff
@@ -215,6 +253,50 @@ def _get_participant_experiments(team, experiment_data) -> dict[str, Experiment]
...
+                e.response.text,
+            )
+            if e.response.status_code == 404:
+                raise NotFound("Failed to create channel: Participant not found in CommCare Connect") from e
+            elif e.response.status_code >= 500:
+                raise ServiceUnavailable("Failed to create channel: CommCare Connect service error") from e
+            raise BadRequest(f"Failed to create channel: {e.response.text}") from e
+        except httpx.HTTPError as e:
+            logger.error("Failed to create CommCare Connect channel for participant %s: %s", identifier, str(e))
+            raise ServiceUnavailable("Failed to create channel: Unable to connect to CommCare Connect service") from e
```
snopoke: Fixed in 19ababfe3. Extracted the httpx→DRF error mapping into a `_connect_channel_error` helper, dropping `_setup_connect_channels` cyclomatic complexity from 13 to ~4.

*— Claude Code*

apps/api/views/participants.py
```diff
@@ -1,24 +1,39 @@
+import logging
```
snopoke, line 1: Fixed in 19ababfe3. Extracting the error-mapping helper brought the module mean cyclomatic complexity back under threshold; CodeScene now scores the file 9.38.

*— Claude Code*

snopoke:
> Thoughts on returning the `connect_channel_id` only for connect channels?

While that seems nice it means the schema can't be fully specified without knowing the request input data which isn't ideal. I will add a comment to the field in the schema indicating that it's only set for connect channels.

SmittieC (approved):

PR #3552, opened 2026-06-05
Require per-session tokens on chat API session endpoints

snopoke:
### Product Description

Chat transcripts are no longer readable by anyone who obtains a session ID. New chat sessions require a secret session token (issued when the session starts) or an authenticated user with access to the session.

### Technical Description

- Stateless signed token (`django.core.signing`) issued by `POST /api/chat/start/`, sent as `X-Session-Token` on the four session endpoints.
- `ExperimentSession.session_token_required` (default `True`); enforcement via `SessionAccessPermission`, which preserves legacy behavior for opted-out sessions.
- Opt-out: explicit `use_session_token: false`, or implicit for pre-token widgets (detected via `x-ocs-widget-version` header) and the anonymous web-chat pages until the widget supports tokens (separate PR).
- Token access to sessions inactive >7 days is rejected (`CHAT_SESSION_TOKEN_INACTIVITY_WINDOW`).
- **Breaking change** for direct API consumers of `/api/chat/start/`: send the returned `session_token` on subsequent calls, or opt out.

Spec: `docs/superpowers/specs/2026-06-05-chat-session-token-design.md`

### Migrations

Adds the field (default `True`) and backfills sessions active in the last 24h to `False` so live conversations are uninterrupted.

### Docs and Changelog


apps/api/tests/test_chat_session_token.py
```diff
@@ -0,0 +1,225 @@
...
+    if user is None:
+        user = UserFactory.create()
+        session.participant.user = user
+        session.participant.save()
+    api_client.force_login(user)
+    assert api_client.get(poll_url(session)).status_code == 200
+
+
+@pytest.mark.django_db()
+def test_team_member_bypasses_token(api_client, session, team_with_users):
```
snopoke: Not sure why we need this to be true. See comment in `SessionAccessPermission._user_has_session_access`

apps/api/permissions.py
```diff
@@ -97,6 +125,13 @@ def has_permission(self, request, view):
 
         return experiment.is_participant_allowed(participant_id)
 
+    def _user_has_session_access(self, user, session) -> bool:
+        if session.participant and session.participant.user_id == user.id:
+            return True
+        # Team membership is intentionally sufficient: team members may access
+        # any session in their team (matches team-scoped access elsewhere).
+        return session.team.members.filter(id=user.id).exists()
```
snopoke: Where does this requirement come from? I don't think this should be here.

apps/api/session_tokens.py
```diff
@@ -0,0 +1,44 @@
...
+    """Sliding inactivity backstop: reject token access to long-inactive sessions.
+
+    Activity is the latest chat message (polling does not count, so a leaked
+    token cannot keep a session alive), falling back to session creation.
+    """
+    return timezone.now() - _last_activity(session) > settings.CHAT_SESSION_TOKEN_INACTIVITY_WINDOW
+
+
+def _last_activity(session: ExperimentSession) -> datetime:
+    last_message = session.chat.messages.aggregate(last=Max("created_at"))["last"]
```
snopoke: could this use the `session.last_activity_at` field instead of doing a DB query?

apps/chatbots/views.py
```diff
@@ -771,6 +772,14 @@ def chatbot_chat_embed(request, team_slug: str, experiment_id: uuid.UUID, sessio
 
 
 def _chatbot_chat_ui(request, embedded=False):
+    session = request.experiment_session
+    if flag_is_active(request, "flag_chat_widget") and session.session_token_required:
+        # The bundled chat widget cannot send session tokens yet. Give these
+        # sessions the same implicit opt-out as other pre-token widgets.
+        # Remove once the widget supports the session-token prop.
```
snopoke: Why is this necessary - the existing opt-outs for widget version and user authenticated requests should be sufficient  shouldn't they?

apps/api/tests/test_chat_session_token.py
```diff
@@ -0,0 +1,225 @@
...
+    response = start_session(api_client, experiment, {"use_session_token": True}, HTTP_X_OCS_WIDGET_VERSION="0.9.0")
+    body = response.json()
+    assert body["session_token"]
+    assert ExperimentSession.objects.get(external_id=body["session_id"]).session_token_required is True
+
+
+@pytest.mark.django_db()
+def test_old_widget_implicitly_opts_out(api_client, experiment):
+    """Pre-token widgets send the version header but no use_session_token field."""
+    response = start_session(api_client, experiment, HTTP_X_OCS_WIDGET_VERSION="0.8.0")
```
snopoke, line 202: where does this version check happen?

apps/api/permissions.py
```diff
@@ -77,16 +78,43 @@ def has_permission(self, request, view):
...
-            return True
+class SessionAccessPermission(BasePermission):
+    """Object-capability check for chat session endpoints.
+
+    Token-required sessions demand a valid X-Session-Token (or an
+    authenticated user with rights to the session). Legacy sessions
+    (session_token_required=False) keep the historical behavior.
+    """
 
+    def has_permission(self, request, view):
```
snopoke, line 89: Fixed in abd2026e7. Extracted the token-validation block into `_token_grants_access`, so `has_permission` now reads: lookup → legacy branch → participant-user bypass → token check. Drops it back under the complexity threshold.

*— Claude Code*

apps/api/tests/test_chat_session_token.py
snopoke: Fixed in abd2026e7. Removed the team-membership bypass — the chat API session endpoints are only consumed by widgets/external clients, and team members view sessions through the Django HTMX views, not these endpoints. This test now asserts a team member *without* being the participant is denied (`test_team_member_without_participant_denied`).

*— Claude Code*

apps/api/permissions.py
snopoke: Fixed in abd2026e7. Removed. Only the session’s own participant-user now bypasses the token (covers the login-required `chatbot_chat_session` page where the viewer is the participant). Team membership grants nothing here.

*— Claude Code*

apps/chatbots/views.py
snopoke: Fixed in abd2026e7. Removed. You’re right for every flow except the *anonymous* public web-chat (`chatbot_chat`/`chatbot_chat_embed` under `flag_chat_widget`): those create the session server-side (not via `/api/chat/start/`, so the widget-version opt-out never fires) and the visitor is anonymous (so the participant bypass doesn’t apply). Since it’s a POC flag, we agreed to let that flow 403 on poll until the widget gains token support (widget PR passes a `session-token` prop) rather than mutate a security flag in a GET render path.

*— Claude Code*

apps/api/tests/test_chat_session_token.py
snopoke: It’s in `chat_start_session` (apps/api/views/chat.py): when `use_session_token` is absent, it defaults to `"x-ocs-widget-version" not in request.headers`. It’s a header *presence* check, not a version-value comparison — every token-aware widget always sends `use_session_token` explicitly, so a request that sends the version header but omits the field is an old widget and is implicitly opted out.

*— Claude Code*

apps/experiments/migrations/0141_experimentsession_session_token_required.py
```diff
@@ -0,0 +1,37 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="experimentsession",
+            name="session_token_required",
+            field=models.BooleanField(
+                default=True,
+                help_text="Require a signed session token (or authenticated user) for chat API access to this session.",
+            ),
```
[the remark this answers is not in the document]
snopoke: Not applicable — all environments run Postgres 16 (`pgvector/pgvector:pg16`). On PG11+, `ADD COLUMN … DEFAULT <constant>` is a metadata-only operation (no table rewrite), so the large-table concern doesn’t arise. The backfill `UPDATE` only touches sessions active in the last 24h.

*— Claude Code*

apps/api/session_tokens.py
snopoke: Good call — done in abd2026e7. `session_token_expired` now reads `session.last_activity_at or session.created_at` (indexed field, maintained by the post_save signal on human messages) instead of running a `Max(created_at)` aggregate per request. It’s also the better signal: it tracks user activity, not AI replies.

*— Claude Code*

apps/api/session_tokens.py
```diff
@@ -0,0 +1,38 @@
...
+    return signing.dumps({"sid": str(session.external_id)}, salt=SESSION_TOKEN_SALT)
+
+
+def validate_session_token(token: str, session_external_id: str) -> bool:
+    """Check `token`'s signature and that it was issued for this session."""
+    if not token or not isinstance(token, str):
+        return False
+    try:
+        payload = signing.loads(token, salt=SESSION_TOKEN_SALT)
+    except signing.BadSignature:
```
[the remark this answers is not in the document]
snopoke: Hardened in 51188c2a0 — `validate_session_token` now catches `(signing.BadSignature, ValueError)` and fails closed. Note this was defense-in-depth, not a reachable bug: Django verifies the HMAC before decoding the payload, so every forged/malformed token raises `BadSignature` first (confirmed against 13 malformed inputs, all `BadSignature`). The widening just makes the function total against any future decode path.

*— Claude Code*

apps/api/views/chat.py
```diff
@@ -246,6 +247,10 @@ def chat_start_session(request):
     session_data = data.get("session_data", {})
     remote_id = data.get("participant_remote_id", "")
     name = data.get("participant_name")
+    use_session_token = data.get("use_session_token")
+    if use_session_token is None:
+        # Pre-token widgets send the version header but not the field; treat as opt-out.
+        use_session_token = "x-ocs-widget-version" not in request.headers
```
snopoke: Fixed in 51188c2a0. Extracted the session-token issue/opt-out logic into `_issue_or_opt_out_session_token`, removing the two branches this PR added to `chat_start_session` and restoring it to its prior cyclomatic complexity (the function was already well over threshold before this change).

*— Claude Code*

config/settings.py
```diff
@@ -862,6 +862,9 @@ def _before_create(task: dict):
 MAX_SUMMARY_LENGTH = 1024
 MAX_FILES_PER_COLLECTION = 1000
 MAX_FILE_SIZE_MB = 50
+
+# How long after the last message a chat session token remains usable.
+CHAT_SESSION_TOKEN_INACTIVITY_WINDOW = timedelta(days=7)
 EMBEDDING_VECTOR_SIZE = 1024
 SUPPORTED_FILE_TYPES = {
     "file_search": (
```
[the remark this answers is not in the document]
snopoke: Fixed in 709c9c77e. Added `x-session-token` to `CORS_ALLOW_HEADERS` (config/settings.py) alongside the other widget headers (`x-embed-key`, `x-ocs-widget-version`). Real catch — `CORS_ALLOW_HEADERS` is an explicit allowlist, so cross-origin widget poll/message/upload requests carrying `X-Session-Token` would have failed preflight. The test suite missed it because the Django test client does not perform CORS preflight.

*— Claude Code*

snopoke:
@coderabbitai full review

PR #3556, opened 2026-06-06
Empirical inline-import verification + CI gate

snopoke:
### Product Description
No change — developer tooling and import cleanup.

### Technical Description
ruff already requires a `# noqa` comment on every import that lives inside a function (PLC0415). The problem is the *reason* in that comment can be wrong — it goes stale as code changes, and AI-generated code often makes one up. This PR found two imports justified as "circular" where the cycle no longer existed.

`scripts/check_inline_imports.py` checks those reasons instead of trusting them. It copies `apps/` to a temp dir, moves a candidate import up to module level, and tries to import the whole project under `django.setup()`:

- **No reason / "circular" reason** → move it up and see if imports still work. If they do, the reason is false → **FAIL**.
- **"lazy" (load cost) reason** → check whether that module is already imported at startup elsewhere. If it is, the reason is void → **WARN** (doesn't block CI; the heuristic can be wrong).
- **Timing / test-patching reasons** → trusted (can't be checked either way).
- **Modules in ruff's banned list** → skipped (ruff already forces those inline).

A new `inline-imports` CI job runs the checker on a PR's changed `apps/**.py` files only — about 1–2 minutes when Python files change, and it skips setup entirely otherwise.

The rest of the diff is the cleanup a full project sweep turned up: 65 imports either moved to module level or given an accurate reason. That's what lets the CI gate start from a clean slate.

**Worth a closer look during review:**
- `speech_service.py`: moving its imports up flipped the correct `mock.patch` target from the source module to the usage site — the two `test_voice_providers.py` changes are exactly that.
- `channels/tasks.py` ↔ `email_channel.py`, and the `documents` task ↔ service pair, are genuine import cycles. One side of each stays inline, with a corrected reason.
- 9 "dubious lazy" warnings remain (in `llm_service/main.py` and `llm_service/utils.py`). They don't fail CI and are tracked in #3560.

**Verification:**
- Full sweep after cleanup: 0 imports that should move, 0 false reasons.
- Affected-app tests (1,872) pass.
- Startup time is unchanged. Timing `django.setup()` 7× on each branch: main 5.885s median, this branch 5.983s — a 0.1s gap that's within normal run-to-run noise (the ranges overlap). The heavy libraries moved up in `speech_service.py` (pydub, openai, elevenlabs) are already loaded at startup on `main` anyway, so moving them costs nothing — which is the same thing the checker's "dubious lazy" warning is pointing at.

### Docs and Changelog

### DIFF
--- a/.github/workflows/lint_and_test.yml
+++ b/.github/workflows/lint_and_test.yml
@@ -113,6 +113,39 @@ jobs:
       - name: Run ty type checker
         run: uv run ty check apps/
 
+  inline-imports:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v6
+      - id: file_changes
+        uses: trilom/file-changes-action@v1.2.4
+      - id: py_files
+        name: Filter changed Python files
+        run: |
+          files=$(echo '${{ steps.file_changes.outputs.files }}' | tr ' ' '\n' | grep -E '^apps/.*\.py$' | tr '\n' ' ' || true)
+          echo "files=$files" >> $GITHUB_OUTPUT
+      - name: Set up Python
+        if: steps.py_files.outputs.files != ''
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+      - uses: astral-sh/setup-uv@v7
+        if: steps.py_files.outputs.files != ''
+        with:
+          enable-cache: true
+      - name: Install dependencies
+        if: steps.py_files.outputs.files != ''
+        run: |
+          uv venv
+          uv sync --locked --dev
+      - name: Check inline imports in changed files
+        if: steps.py_files.outputs.files != ''
+        env:
+          SECRET_KEY: secret-test-key
+        run: uv run python scripts/check_inline_imports.py apps --files ${{ steps.py_files.outputs.files }}
+
   check-uv-lock:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/lint_and_test.yml
+++ b/.github/workflows/lint_and_test.yml
@@ -123,9 +123,11 @@ jobs:
         uses: trilom/file-changes-action@v1.2.4
       - id: py_files
         name: Filter changed Python files
+        env:
+          CHANGED_FILES: ${{ steps.file_changes.outputs.files }}
         run: |
-          files=$(echo '${{ steps.file_changes.outputs.files }}' | tr ' ' '\n' | grep -E '^apps/.*\.py$' | tr '\n' ' ' || true)
-          echo "files=$files" >> $GITHUB_OUTPUT
+          files=$(printf '%s\n' "$CHANGED_FILES" | tr ' ' '\n' | grep -E '^apps/.*\.py$' | tr '\n' ' ' || true)
+          echo "files=$files" >> "$GITHUB_OUTPUT"
       - name: Set up Python
         if: steps.py_files.outputs.files != ''
         uses: actions/setup-python@v6
@@ -144,7 +146,10 @@ jobs:
         if: steps.py_files.outputs.files != ''
         env:
           SECRET_KEY: secret-test-key
-        run: uv run python scripts/check_inline_imports.py apps --files ${{ steps.py_files.outputs.files }}
+          CHANGED_FILES: ${{ steps.py_files.outputs.files }}
+        run: |
+          read -r -a files <<< "$CHANGED_FILES"
+          uv run python scripts/check_inline_imports.py apps --files "${files[@]}"
 
   check-uv-lock:
     runs-on: ubuntu-latest
--- a/scripts/check_inline_imports.py
+++ b/scripts/check_inline_imports.py
@@ -61,6 +61,25 @@ class InlineImport:
     justification: str | None = None
 
 
+def _read_justification(node: ast.AST, lines: list[str]) -> str | None:
+    """Return the lazy-import/noqa reason on an import statement, if any."""
+    for line in lines[node.lineno - 1 : node.end_lineno]:
+        if match := LAZY_MARKER.search(line) or NOQA_REASON.search(line):
+            return match.group("reason")
+    return None
+
+
+def _absolute_import_modules(node: ast.Import | ast.ImportFrom) -> tuple[str, ...]:
+    """Absolute module names an import targets (relative imports have none)."""
+    if isinstance(node, ast.Import):
+        return tuple(alias.name for alias in node.names)
+    return (node.module,) if node.level == 0 and node.module else ()
+
+
+def _is_function(node: ast.AST) -> bool:
+    return isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef))
+
+
 def find_inline_imports(source: str) -> list[InlineImport]:
     """Return all import statements nested inside function or method bodies.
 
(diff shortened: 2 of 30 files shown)


scripts/check_inline_imports.py
```diff
@@ -0,0 +1,587 @@
...
+            findings += [Finding(c.path, c.inline_import, "inconclusive", baseline.error) for c in empirical]
+        else:
+            for c, hoistable, error in classify_candidates(empirical, lambda cands: verify_hoists(ctx, cands)):
+                if c.inline_import.justification is None:
+                    verdict = "violation" if hoistable else "legitimate"
+                else:
+                    verdict = "stale" if hoistable else "justified"
+                findings.append(Finding(c.path, c.inline_import, verdict, error))
+    findings.sort(key=lambda f: (str(f.path), f.inline_import.lineno))
+    return findings
```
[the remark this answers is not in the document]
snopoke: Fixed in bb42618ac. Extracted `_route_import`, `_classify_lazy_claims`, `_classify_empirical`, and `_empirical_verdict`; `check_package` is now a thin orchestrator (CC well under threshold).

*— Claude Code*

snopoke, line 557: Fixed in bb42618ac. The nested classification passes moved into `_classify_lazy_claims`/`_classify_empirical`, flattening `check_package`.

*— Claude Code*

snopoke, line 557: Fixed in bb42618ac. Nesting depth dropped after extracting the per-bucket helpers from `check_package`.

*— Claude Code*

scripts/check_inline_imports.py
```diff
@@ -0,0 +1,587 @@
...
+            f" {counts['dubious']} dubious, {counts['legitimate']} legitimate,"
+            f" {counts['justified']} justified."
+        )
+    else:
+        print("No inline imports found.")
+    if counts["violation"] or counts["stale"]:
+        return 1
+    if counts["inconclusive"]:
+        return 2
+    return 0
```
snopoke, line 632: Fixed in bb42618ac. The per-verdict print block is now `_format_finding`, leaving `main` a simple loop.

*— Claude Code*

scripts/check_inline_imports.py
```diff
@@ -0,0 +1,587 @@
...
+                    elif node.level:
+                        parts = module_package.split(".")[: len(module_package.split(".")) - (node.level - 1)]
+                        base = ".".join(parts + ([node.module] if node.module else []))
+                        names.add(base)
+                        # `from . import x` may import submodule x
+                        names.update(f"{base}.{alias.name}" for alias in node.names)
+            visit(node, in_function or isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)))
+
+    visit(tree, False)
+    return names
```
snopoke, line 231: Fixed in bb42618ac. Relative-import resolution moved to `_imported_names`, shrinking the visitor.

*— Claude Code*

snopoke, line 231: Fixed in bb42618ac. Same `_imported_names` extraction flattens the visitor branches.

*— Claude Code*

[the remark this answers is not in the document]
snopoke: Fixed in bb42618ac. Nesting reduced via the `_imported_names` helper.

*— Claude Code*

scripts/check_inline_imports.py
```diff
@@ -0,0 +1,587 @@
...
+                    modules = (node.module,) if node.level == 0 and node.module else ()
+                results.append(
+                    InlineImport(
+                        lineno=node.lineno,
+                        statement=ast.unparse(node),
+                        modules=modules,
+                        justification=justification,
+                    )
+                )
+            visit(node, in_function or isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)))
```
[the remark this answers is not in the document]
snopoke: Fixed in bb42618ac. Justification parsing and module extraction moved to `_read_justification`/`_absolute_import_modules`.

*— Claude Code*

snopoke: Fixed in bb42618ac. Visitor body shrunk after extracting the helpers above.

*— Claude Code*

scripts/check_inline_imports.py
snopoke: Fixed in bb42618ac. The nested classification passes moved into `_classify_lazy_claims`/`_classify_empirical`, flattening `check_package`.

*— Claude Code*

scripts/check_inline_imports.py
snopoke: Fixed in bb42618ac. The same `_imported_names` extraction flattens the visitor branches.

*— Claude Code*

scripts/check_inline_imports.py
```diff
@@ -0,0 +1,587 @@
+#!/usr/bin/env python3
```
snopoke, line 1: Addressed in bb42618ac. Extracting the helpers above lowered the file mean; CodeScene now scores the file 9.38 (green) with the per-method smells resolved.

*— Claude Code*

scripts/check_inline_imports.py
snopoke: Fixed in bb42618ac. Nesting depth dropped after extracting the per-bucket helpers from `check_package`.

*— Claude Code*

scripts/check_inline_imports.py
snopoke: Fixed in bb42618ac. Visitor body shrunk after extracting the helpers above.

*— Claude Code*

scripts/check_inline_imports.py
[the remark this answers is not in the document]
snopoke: Fixed in bb42618ac. Nesting reduced via the `_imported_names` helper.

*— Claude Code*

scripts/test_check_inline_imports.py
```diff
@@ -0,0 +1,574 @@
...
+    def test_stale_circular_claim_fails(self, tmp_path):
+        make_pkg(
+            tmp_path,
+            {
+                "pkg/__init__.py": "",
+                "pkg/a.py": ("def f():\n    import json  # noqa: PLC0415 - circular: b imports a\n    return json\n"),
+            },
+        )
+        (finding,) = check_package(tmp_path / "pkg")
+        assert finding.verdict == "stale"
```
snopoke, line 198: Fixed in bb42618ac. The single-import tests now share a `check_one_import` helper; the test file scores 10.0.

*— Claude Code*

.github/workflows/lint_and_test.yml
```diff
@@ -113,6 +113,39 @@ jobs:
...
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v6
+      - id: file_changes
+        uses: trilom/file-changes-action@v1.2.4
+      - id: py_files
+        name: Filter changed Python files
+        run: |
+          files=$(echo '${{ steps.file_changes.outputs.files }}' | tr ' ' '\n' | grep -E '^apps/.*\.py$' | tr '\n' ' ' || true)
+          echo "files=$files" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
snopoke: Fixed in bb42618ac. Both steps now read the file list from a `CHANGED_FILES` env var and split it into a bash array (`read -r -a`) instead of interpolating the template into the shell.

*— Claude Code*

scripts/check_inline_imports.py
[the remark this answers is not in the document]
snopoke: Fixed in bb42618ac. Justification parsing and module extraction moved to `_read_justification`/`_absolute_import_modules`.

*— Claude Code*

scripts/check_inline_imports.py
snopoke: Fixed in bb42618ac. The per-verdict print block is now `_format_finding`, leaving `main` a simple loop.

*— Claude Code*

scripts/check_inline_imports.py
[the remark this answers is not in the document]
snopoke: Fixed in bb42618ac. Extracted `_route_import`, `_classify_lazy_claims`, `_classify_empirical`, and `_empirical_verdict`; `check_package` is now a thin orchestrator.

*— Claude Code*

scripts/check_inline_imports.py
snopoke: Fixed in bb42618ac. Relative-import resolution moved to `_imported_names`, shrinking the visitor.

*— Claude Code*

SmittieC (approved):
Nice. I glossed over it only

PR #3572, opened 2026-06-09
fix: redirect version snapshot URLs and fix notification bot links

### LINKED ISSUE #3571: fix: notification links to chatbots 404/500 when pointing to a version instead of working version
## Problem

Notification links (e.g. "View Bot") can 404/500 because `Experiment.get_absolute_url()` uses `self.id` unconditionally. When the method is called on a **published version** (i.e. `working_version_id IS NOT NULL`) rather than the working-version family head, the URL contains the version's integer PK.

When that URL is loaded, `single_chatbot_home` calls `resolve_published_or_working(experiment)`, which requires a family-head experiment and raises `ValueError` for a version — resulting in an error page.

### Where published versions sneak into notifications

**1. Channel-triggered notifications (trace errors, audio failures, delivery failures, etc.)**

Channels are initialised with `resolve_published_or_working(experiment_channel.experiment)`, so `self.experiment` may be the published version. The tracer already handles this correctly for trace records (`ocs_tracer.py` L75-77) but `_fire_trace_error_notification` passes `self.experiment` directly:

```python
trace_error_notification(
    experiment=self.experiment,  # ← could be a published version
    ...
)
```

Same pattern in `audio_synthesis_failure_notification`, `message_delivery_failure_notification`, etc.

**2. Deprecated/deleted model notifications**

`get_affected_teams_data` in `apps/data_migrations/management/commands/base.py` explicitly queries published versions via `Q(is_default_version=True)` and calls `get_absolute_url()` on them.

### 1. Fix `Experiment.get_absolute_url()`

When called on a published version, resolve to the working-version URL and append `?version_id=X#versions` so the link lands directly on the right version in the versions tab:

```python
def get_absolute_url(self):
    if self.is_a_version:
        url = reverse("chatbots:single_chatbot_home", args=[get_slug_for_team(self.team_id), self.working_version_id])
        return f"{url}?version_id={self.version_number}#versions"
    return reverse("chatbots:single_chatbot_home", args=[get_slug_for_team(self.team_id), self.id])
```

### 2. Add a guard + redirect in `single_chatbot_home`

Defensively redirect any request that arrives with a version's PK to the correct working-version URL, preserving `?version_id=X#versions`:

```python
def single_chatbot_home(request, team_slug: str, experiment_id: int):
    experiment = get_object_or_404(Experiment.objects.get_all(), id=experiment_id, team=request.team)

    # If a version ID was passed, redirect to the working version.
    # Preserve ?version_id=X#versions so the versions tab can link directly to that version.
    if experiment.is_a_version:
        url = reverse("chatbots:single_chatbot_home", args=[team_slug, experiment.working_version_id])
        url += f"?version_id={experiment.version_number}#versions"
        return redirect(url)
    ...
```

### 3. Handle `?version_id=X#versions` in the versions tab

The versions tab / `#versions` section should read the `version_id` query param and highlight or scroll to that version row so users land in the right place when following a notification link to a specific version.

## Acceptance criteria

- [ ] Clicking a "View Bot" notification link never results in a 404/500, regardless of whether the stored URL points to a working version or a published version PK
- [ ] `Experiment.get_absolute_url()` returns the working-version URL with `?version_id=X#versions` appended when called on a published version
- [ ] `single_chatbot_home` redirects version IDs to the working version with `?version_id=X#versions`
- [ ] The versions tab honours `?version_id=X` to highlight/scroll to the relevant version
- [ ] Notification functions that currently pass `self.experiment` (possibly a published version) are also fixed at source (tracer, channel notifications)

ocs-percy:
### Product Description
Users who click a "View Bot" link in a notification that points to a version snapshot are now redirected to the correct chatbot page, with the relevant version highlighted in the Versions tab.

### Technical Description
Builds on #3569 which already addressed the downstream `ValueError`. This PR completes the fix described in #3571:

**`Experiment.get_absolute_url()`** — When called on a published version snapshot, returns the working version URL with `?version_id=X#versions` so notification links land in the right place.

**`single_chatbot_home`** — Replaced the 404 filter with a redirect: if a version ID is requested, redirects to the working version with `?version_id=X#versions`.

**Versions tab UX** — When `?version_id=X` is present, the page auto-selects the Versions tab and scrolls+highlights the matching row after HTMX loads the table. Added `data-version-number` row attr to `ExperimentVersionsTable` to support this.

**Notification sources fixed at origin** — `OCSTracer._fire_trace_error_notification`, `ChannelBase` (audio synthesis, file delivery, transcription), and `channels_v2` (audio synthesis, message delivery) all now pass the working version to notification functions so future links are correct too.

### Docs and Changelog

Closes #3571

### DIFF
--- a/apps/service_providers/tracing/ocs_tracer.py
+++ b/apps/service_providers/tracing/ocs_tracer.py
@@ -263,8 +263,11 @@ class OCSTracer(Tracer):
 
     def _fire_trace_error_notification(self) -> None:
         trace_url = self.trace_record.get_absolute_url() if self.trace_record else None
+        # self.experiment may be a published version; notifications should always link to the
+        # working version so get_absolute_url() resolves correctly.
+        experiment = self.experiment.get_working_version() if self.experiment.is_a_version else self.experiment
         trace_error_notification(
-            experiment=self.experiment,
+            experiment=experiment,
             session=self.session,
             span_name=self.error_span_name,
             error_message=self.error_message,
--- a/apps/service_providers/tracing/ocs_tracer.py
+++ b/apps/service_providers/tracing/ocs_tracer.py
@@ -263,10 +263,8 @@ class OCSTracer(Tracer):
 
     def _fire_trace_error_notification(self) -> None:
         trace_url = self.trace_record.get_absolute_url() if self.trace_record else None
-        # self.experiment may be a published version; notifications should always link to the
-        # working version so get_absolute_url() resolves correctly.
         trace_error_notification(
-            experiment=self.experiment.get_working_version(),
+            experiment=self.experiment,
             session=self.session,
             span_name=self.error_span_name,
             error_message=self.error_message,
--- a/apps/service_providers/tracing/ocs_tracer.py
+++ b/apps/service_providers/tracing/ocs_tracer.py
@@ -265,9 +265,8 @@ class OCSTracer(Tracer):
         trace_url = self.trace_record.get_absolute_url() if self.trace_record else None
         # self.experiment may be a published version; notifications should always link to the
         # working version so get_absolute_url() resolves correctly.
-        experiment = self.experiment.get_working_version() if self.experiment.is_a_version else self.experiment
         trace_error_notification(
-            experiment=experiment,
+            experiment=self.experiment.get_working_version(),
             session=self.session,
             span_name=self.error_span_name,
             error_message=self.error_message,
--- a/apps/channels/channels_v2/stages/core.py
+++ b/apps/channels/channels_v2/stages/core.py
@@ -596,7 +596,7 @@ class ResponseFormattingStage(ProcessingStage):
             except AudioSynthesizeException:
                 # Graceful fallback to text -- not an unrecoverable error
                 logger.exception("Error generating voice response")
-                audio_synthesis_failure_notification(ctx.experiment.get_working_version(), session=ctx.experiment_session)
+                audio_synthesis_failure_notification(ctx.experiment, session=ctx.experiment_session)
                 ctx.voice_audio = None
         else:
             message, uncited_files = self._format_reference_section(message, files, ctx)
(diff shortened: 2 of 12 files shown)


SmittieC (commented):
An ADR with the decision to redirect to the working version instead of returning a 404 might be handy.

apps/service_providers/tracing/ocs_tracer.py
```diff
@@ -263,8 +263,11 @@ def _bust_caches(self):
 
     def _fire_trace_error_notification(self) -> None:
         trace_url = self.trace_record.get_absolute_url() if self.trace_record else None
+        # self.experiment may be a published version; notifications should always link to the
+        # working version so get_absolute_url() resolves correctly.
+        experiment = self.experiment.get_working_version() if self.experiment.is_a_version else self.experiment
```
snopoke: if we are handling the version redirect in the view and also in `Experiment.get_absolute_url` then why do we need to resolve the working version for notifications?

ocs-percy:
Good call @SmittieC — the reasoning is:

- A 404 gives the user no path forward; the version snapshot ID is a valid, permanent identifier for that version
- The redirect preserves intent: if someone bookmarked or copied a version URL from a notification, we should land them somewhere useful, not block them
- The `?version_id=X#versions` destination is specifically useful — it opens the chatbot page and highlights the exact version in the Versions tab, which is the closest approximation to "what you were trying to see"
- Consistency with typical web conventions (canonical URL redirects rather than hard 404s for alternative valid identifiers)

Happy to write that up as a proper ADR if you'd like one in the docs, or a brief comment in the code is probably sufficient given the commit message covers the decision.

apps/service_providers/tracing/ocs_tracer.py
```diff
@@ -263,8 +263,10 @@ def _bust_caches(self):
 
     def _fire_trace_error_notification(self) -> None:
         trace_url = self.trace_record.get_absolute_url() if self.trace_record else None
+        # self.experiment may be a published version; notifications should always link to the
+        # working version so get_absolute_url() resolves correctly.
```
snopoke: This logic doesn't stand up since get_absolute_url handles working versions correctly. Please remove the working version resolution for notifications and leave them linked to the actual version.

ocs-percy: Good point — reverted in 679c719. `get_absolute_url()` does the right thing now, and keeping the version experiment means the link actually points to `?version_id=X#versions` rather than just the plain working version URL, which is more useful context.

PR #3573, opened 2026-06-09
feat(cost-tracking): add PricingRule, UsageRecord, and resolver (PR 1/3)

davidnene:
### Product Description

Foundation for AI cost tracking in OCS: data model, pricing resolver, and write path. Flag-gated and disabled by default.

### Technical Description

**New app: `apps/cost_tracking/`**

Models:
- `PricingRule` — one row per `(team, provider, model, service_kind)`. `team=NULL` is a global rule. `effective_from` / `effective_to` bookend each row's life. `created_by` FK captures admin attribution. Partial `UniqueConstraint` with `nulls_distinct=False` prevents duplicate active global rules.
- `UsageRecord` — fact table, one row per `(trace, model, service_kind)` bucket. Snapshots `unit_price` and `currency` so historical rows are stable across rate changes. `confidence` enum (`EXACT` / `ESTIMATED` / `UNKNOWN`) describes token-count provenance only; "was this priced?" is `unit_price IS NOT NULL`. Indexes on `(team, timestamp)`, `(team, experiment, timestamp)`, `(team, session, timestamp)`, `(team, model_name, timestamp)`, `(team, confidence, timestamp)`.

Services:
- `PricingResolver` — three-step lookup (team override → global rule → `UnpricedRule` sentinel). Redis-backed cache, 24h TTL. Invalidated by `post_save` and `post_delete` signals on `PricingRule` (catches queryset-level deletes too).
- `record_usage_bulk()` — single bulk insert per call, wrapped in try/except so DB hiccups never propagate. Cost calc is `(quantity / 1000) × unit_price` — quantity is raw tokens, `unit_price` is per 1K tokens.

Loader & seed:
- `apps/cost_tracking/seed_data/llm_pricing.json` with the most-used models.
- `manage.py load_ai_pricing` — idempotent upsert with supersession (existence check via `.exists()`, close-old + create-new in `transaction.atomic()`, explicit `PricingResolver.invalidate(...)` on the close path).
- A data migration calls the loader via `RunPython` so `manage.py migrate` applies the seed.

Flag:
- Data migration registers `flag_ai_cost_monitoring` with `everyone=False`.

Tests (`apps/cost_tracking/tests/`):
- `test_pricing_resolver.py`
- `test_record_usage_bulk.py`
- `test_decimal_precision.py`
- `test_seed_coverage.py`
- `test_supersession.py`
- `test_unique_constraint.py`
- `test_cache_invalidation_signals.py`

### Migrations

Additive only: two new tables, one new flag row, idempotent seed via `RunPython`.

### Demo

```bash
uv run python manage.py migrate
uv run python manage.py load_ai_pricing      # idempotent — no-op on second run
uv run python manage.py shell
>>> from apps.cost_tracking.services.pricing import PricingResolver
>>> PricingResolver().resolve(team_id=None, provider_type="openai",
...     model_name="gpt-4o-mini", service_kind="llm_input",
...     at=timezone.now())
<PricingRule: openai/gpt-4o-mini/llm_input @ 0.00015>

### CODE REVIEW

**Comment by snopoke** [comment]
@davidnene is this ready for review? (please resolve any comments that have been addressed)

**Comment by davidnene** [comment]
> [Review Change Stack](https://app.coderabbit.ai/change-stack/dimagi/open-chat-studio/pull/3573?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)
> 
> 📝 Walkthrough
> ### ❌ Failed checks (1 warning)
> Check name	Status	Explanation	Resolution
> Docstring Coverage	⚠️ Warning	Docstring coverage is 16.67% which is insufficient. The required threshold is 80.00%.	Write docstrings for the functions missing them to satisfy the coverage threshold.
> ✅ Passed checks (4 passed)
> ✏️ Tip: You can configure your own custom pre-merge checks in the settings.
> 
> Comment `@coderabbitai help` to get the list of available commands and usage tips.

resolved

**Comment by davidnene** [comment]
@snopoke I have resolved the addressed comments. Ready for review

File: apps/cost_tracking/services/pricing.py [hunk 1]
```diff
@@ -0,0 +1,121 @@
...
+
+        Returning `None` (rather than the UNPRICED sentinel) lets `resolve()`
+        cleanly distinguish "no team override exists" from "found a team
+        override that happens to be unpriced" — only the former should fall
+        through to the global lookup.
+        """
+        if key.team_id is None:
+            return None
+        rule = self._lookup_one(key, at, use_cache)
+        return rule if rule.unit_price is not None else None
```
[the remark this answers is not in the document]
davidnene: resolved

apps/cost_tracking/models.py
```diff
@@ -0,0 +1,117 @@
...
+from apps.teams.models import BaseTeamModel, Team
+from apps.utils.fields import SanitizedJSONField
+
+
+class ServiceKind(models.TextChoices):
+    LLM_INPUT = "llm_input"
+    LLM_OUTPUT = "llm_output"
+    LLM_CACHED_INPUT = "llm_cached_input"
+    LLM_CACHE_WRITE = "llm_cache_write"
+    LLM_REASONING = "llm_reasoning"
```
snopoke: OpenAI's pricing page doesn't distinguish between reasoning tokens and output tokens so I'm not sure we should track them separately:

Image

https://developers.openai.com/api/docs/pricing

apps/cost_tracking/models.py
```diff
@@ -0,0 +1,117 @@
...
+
+class PricingSource(models.TextChoices):
+    SEED = "seed"
+    MANUAL = "manual"
+    IMPORT = "import"
+
+
+class PricingRule(BaseTeamModel):
+    """team=NULL is a global rule. Rate changes use supersession:
+    close the active rule, insert a new one. Never edit in place.
```
snopoke: nit: Simplify this 'claude' language:


```suggestion
    """team=NULL is a global rule. Intended as mostly read-only table.
    Rate changes close the current rule by setting `effective_to` and
    insert a new rule with the new rate.
```

apps/cost_tracking/models.py
```diff
@@ -0,0 +1,117 @@
...
+    UNKNOWN = "unknown"
+
+
+class PricingSource(models.TextChoices):
+    SEED = "seed"
+    MANUAL = "manual"
+    IMPORT = "import"
+
+
+class PricingRule(BaseTeamModel):
```
snopoke: Drop the `BaseTeamModel` parent class since all the fields except 'updated_at' have been overridden and the `VersioningMixin` isn't necessary.

apps/cost_tracking/models.py
```diff
@@ -0,0 +1,117 @@
...
+                fields=["team", "provider_type", "model_name", "service_kind"],
+                condition=Q(effective_to__isnull=True),
+                nulls_distinct=False,
+                name="cost_tracking_unique_active_pricing_rule",
+            ),
+        ]
+
+    def __str__(self):
+        """Compact representation for admin / shell debugging."""
+        scope = self.team.slug if self.team_id else "global"
```
snopoke: nit: use `team_id` to avoid a DB query to load the slug. (since this isn't user facing)
```suggestion
        scope = self.team_id if self.team_id else "global"
```

apps/cost_tracking/models.py
```diff
@@ -0,0 +1,117 @@
...
+    timestamp = models.DateTimeField(auto_now_add=True, db_index=True)
+
+    service_kind = models.CharField(max_length=32, choices=ServiceKind.choices)
+    provider_type = models.CharField(max_length=64)
+    model_name = models.CharField(max_length=128)
+    quantity = models.DecimalField(max_digits=18, decimal_places=4, null=True)
+
+    # Per 1K tokens (the canonical unit for every v4 service kind).
+    # NULL means no rule matched at write time; `cost` will be 0.
+    # Readers infer "priced" from `unit_price IS NOT NULL`.
```
snopoke: I would have expected readers to infer 'priced' from `pricing_rule IS NOT NULL` since that is the primary source. `unit_price` is just a denormalized field.

apps/cost_tracking/models.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __str__(self):
+        """Compact representation for admin / shell debugging."""
+        scope = self.team.slug if self.team_id else "global"
+        return f"[{scope}] {self.provider_type}/{self.model_name}/{self.service_kind} @ {self.unit_price}"
+
+
+class UsageRecord(BaseTeamModel):
+    """One row per (trace, model, service_kind) bucket. Snapshots
+    `unit_price` / `currency` so historical rows are stable across rate changes.
```
snopoke, line 88: I'm wondering about the necessity of this if the pricing rule model is essentially read only. Maybe it's still good to avoid having to join for aggregations but should we update the comment to give the real reason.

apps/cost_tracking/models.py
```diff
@@ -0,0 +1,117 @@
...
+    unit_price = models.DecimalField(max_digits=14, decimal_places=8, null=True)
+    cost = models.DecimalField(max_digits=14, decimal_places=8, default=0)
+    currency = models.CharField(max_length=3, default="USD")
+    confidence = models.CharField(max_length=16, choices=Confidence.choices, default=Confidence.EXACT)
+
+    experiment = models.ForeignKey("experiments.Experiment", null=True, on_delete=models.SET_NULL)
+    session = models.ForeignKey("experiments.ExperimentSession", null=True, on_delete=models.SET_NULL)
+    participant = models.ForeignKey("experiments.Participant", null=True, on_delete=models.SET_NULL)
+    trace = models.ForeignKey("trace.Trace", null=True, on_delete=models.SET_NULL)
+    pricing_rule = models.ForeignKey(PricingRule, null=True, on_delete=models.SET_NULL)
```
snopoke: I think this should be `on_delete=models.PROTECT` to keep the pricing source present as long as there is cost data.

apps/teams/migrations/0010_create_flag_ai_cost_monitoring.py
snopoke, line 1: This isn't necessary - flags are created automatically when needed (`settings.WAFFLE_CREATE_MISSING_FLAGS = True`)

apps/cost_tracking/services/pricing.py
```diff
@@ -0,0 +1,123 @@
...
+        if team_rule is not None:
+            return team_rule
+        return self._lookup_one(replace(key, team_id=None), at, cache_ok)
+
+    def _is_near_now(self, at: datetime) -> bool:
+        """The cache only stores "active right now" results. Historical or
+        future `at` queries can't be safely served from it, so we bypass.
+        """
+        now = timezone.now()
+        return now - self.CACHE_AT_SKEW <= at <= now + self.CACHE_AT_SKEW
```
snopoke, line 65: nit: could be simplified:
```suggestion
        return abs(now - at) > self.CACHE_AT_SKEW
```

apps/cost_tracking/services/pricing.py
```diff
@@ -0,0 +1,123 @@
...
+        global fallback in `resolve()`. Both of these cases return `None`:
+
+        * `key.team_id is None` — caller didn't ask for a team scope at all.
+        * A team-scoped row exists but resolved to `UNPRICED` (unit_price is
+          None) — treat it the same as absent so we still try the global rule.
+        """
+        if key.team_id is None:
+            return None
+        rule = self._lookup_one(key, at, use_cache)
+        return rule if rule.unit_price is not None else None
```
snopoke, line 75: I think it makes more sense to check `rule.price_rule_id is not None` as the source of truth. The two should always match since the `unit_price` DB field can't be `None` but I think it makes the intention clearer.

apps/cost_tracking/services/pricing.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def _maybe_team_rule(self, key: PricingKey, at: datetime, use_cache: bool) -> ResolvedRule | None:
+        """Return a priced team-scoped rule if one is set and matched, else None.
+
+        `None` signals "no applicable priced team override" and triggers the
+        global fallback in `resolve()`. Both of these cases return `None`:
+
+        * `key.team_id is None` — caller didn't ask for a team scope at all.
+        * A team-scoped row exists but resolved to `UNPRICED` (unit_price is
+          None) — treat it the same as absent so we still try the global rule.
```
snopoke: `unit_price` isn't a nullable field so this argument doesn't hold

apps/cost_tracking/services/pricing.py
```diff
@@ -0,0 +1,123 @@
...
+    currency: str
+    pricing_rule_id: int | None
+
+
+UNPRICED = ResolvedRule(unit_price=None, currency="USD", pricing_rule_id=None)
+
+
+class PricingResolver:
+    CACHE_TTL_SECONDS = 24 * 60 * 60  # safety net; signals do the real invalidation
+    CACHE_KEY_PREFIX = "cost_tracking:pricing"
```
snopoke: nit: use a shorter prefix
```suggestion
    CACHE_KEY_PREFIX = "cost"
```

apps/cost_tracking/services/pricing.py
```diff
@@ -0,0 +1,123 @@
...
+            if cached is not None:
+                return cached
+
+        qs = PricingRule.objects.filter(
+            provider_type=key.provider_type,
+            model_name=key.model_name,
+            service_kind=key.service_kind,
+            effective_from__lte=at,
+        ).filter(Q(effective_to__isnull=True) | Q(effective_to__gt=at))
+        qs = qs.filter(team__isnull=True) if key.team_id is None else qs.filter(team_id=key.team_id)
```
snopoke: You don't actually need the if statement here since `team_id=None` is the same as `team__isnull=True`:

```suggestion
        qs = qs.filter(team_id=key.team_id)
```

apps/cost_tracking/services/recorder.py
```diff
@@ -0,0 +1,109 @@
...
+                model_name=event.model_name,
+                service_kind=event.service_kind,
+            ),
+            at=now,
+        )
+        # cost stays at 0 unless we have BOTH a priced rule AND a real quantity.
+        # `event.quantity` is typed as Decimal | int but we still guard against
+        # None / 0 so the future UNKNOWN-confidence path (no token count) can
+        # land here cleanly without a Decimal(None) crash.
+        priced = resolved.unit_price is not None
```
snopoke, line 78: Although this check is correct I would update this to check the `pricing_rule_id` as the source of truth. Then if you really want you can assert that `unit_price` is not None:

```suggestion
        if priced := resolved.pricing_rule_id is not None:
            assert resolved.unit_price, "Unit price unexpectedly None"
``

File: apps/cost_tracking/tests/test_decimal_precision.py [hunk 15]
```diff
@@ -0,0 +1,81 @@
...
+@pytest.mark.parametrize(
+    ("qty", "unit_price", "expected_cost"),
+    [
+        # 1000 tokens at $0.00025/1K -> exactly $0.00025
+        (1000, "0.00025", "0.00025000"),
+        # 500 tokens at $0.00025/1K -> exactly $0.000125
+        (500, "0.00025", "0.00012500"),
+        # 1 token at $0.00025/1K -> $0.00000025 (fits in 8 decimal places)
+        (1, "0.00025", "0.00000025"),
+        # Headline OpenAI gpt-4o-mini rate over 10k tokens
```
snopoke, line 24: nit: prefer `pytest.param` with `id="<name>"` than inline comments since that also provides meaningful test names in console output

apps/cost_tracking/services/recorder.py
snopoke: Alteratives:
* `priced = resolved == UNPRICED`
* Add `ResolvedRule.is_priced` computed property

apps/cost_tracking/seed_data/llm_pricing.json
snopoke, line 1: I assume this will be filled out with the remaining values in a future PR?

apps/cost_tracking/services/recorder.py
```diff
@@ -0,0 +1,109 @@
...
+from dataclasses import dataclass
+from decimal import Decimal
+
+from django.db import transaction
+from django.utils import timezone
+
+from apps.cost_tracking.models import Confidence, ServiceKind, UsageRecord
+from apps.cost_tracking.services.pricing import PricingKey, PricingResolver
+
+logger = logging.getLogger(__name__)
```
snopoke: We use `ocs.` prefixes for loggers:
```suggestion
logger = logging.getLogger("ocs.cost_tracking")
```

apps/cost_tracking/management/commands/load_ai_pricing.py
```diff
@@ -0,0 +1,108 @@
...
+    active = PricingRule.objects.filter(
+        team__isnull=True,
+        provider_type=key.provider_type,
+        model_name=key.model_name,
+        service_kind=key.service_kind,
+        effective_to__isnull=True,
+    )
+
+    # No-op fast path: same rate already active. Push equality into the WHERE
+    # clause so we don't hydrate the row.
```
snopoke: nit: drop meaningless comment
```suggestion
    # No-op fast path: same rate already active.
```

apps/cost_tracking/management/commands/load_ai_pricing.py
```diff
@@ -0,0 +1,108 @@
...
+            source=source,
+            effective_from=now,
+        )
+
+    # `.update()` is a queryset write and does NOT fire post_save, so the
+    # signal-based cache invalidation didn't run for the closed row. The
+    # `create()` above did fire post_save, but the closed row's transition
+    # needs an explicit invalidation too.
+    if closed:
+        PricingResolver.invalidate(replace(key, team_id=None))
```
snopoke: I could be wrong bu I'm pretty sure this will produce the same cache key as the signal from creating the new rule.

apps/cost_tracking/migrations/0002_seed_pricing.py
```diff
@@ -0,0 +1,22 @@
...
+    call_command("load_ai_pricing", verbosity=0)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("cost_tracking", "0001_initial"),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards, migrations.RunPython.noop),
```
snopoke: since we expect to run migrations like this periodically I think it's worth extracting a utility to create this so that you can do this:

```python
operations = [
    load_pricing_data(),
]

snopoke (commented):
This is generally solid. Most of my comments are minor. There is one recurring theme which is anchoring 'priced' vs 'unpriced' on the existence of a pricing rule instead of `unit_price`. While the two should always match I think the intention is clearer to anchor on `pricing_rule != None`.

apps/cost_tracking/seed_data/llm_pricing.json
davidnene: Yeah, sure

apps/cost_tracking/models.py
davidnene: We included it incase of future tracking as langchain has it, but it can be added later. no worries

davidnene:
> This is generally solid. Most of my comments are minor. There is one recurring theme which is anchoring 'priced' vs 'unpriced' on the existence of a pricing rule instead of `unit_price`. While the two should always match I think the intention is clearer to anchor on `pricing_rule != None`.

@snopoke  The two should match at write time but diverge later: pricing_rule_id is on_delete=SET_NULL, so once a rule gets deleted or superseded the FK goes null on every historical row that referenced it. unit_price is a write-time snapshot that's never touched again.

snopoke:
> > This is generally solid. Most of my comments are minor. There is one recurring theme which is anchoring 'priced' vs 'unpriced' on the existence of a pricing rule instead of `unit_price`. While the two should always match I think the intention is clearer to anchor on `pricing_rule != None`.
> 
> @snopoke The two should match at write time but diverge later: pricing_rule_id is on_delete=SET_NULL, so once a rule gets deleted or superseded the FK goes null on every historical row that referenced it. unit_price is a write-time snapshot that's never touched again.

Currently that is true however I would recommend protecting that FK relationship so that you can't delete a pricing rule that has usage data attached.

davidnene:
> > > This is generally solid. Most of my comments are minor. There is one recurring theme which is anchoring 'priced' vs 'unpriced' on the existence of a pricing rule instead of `unit_price`. While the two should always match I think the intention is clearer to anchor on `pricing_rule != None`.
> > 
> > 
> > @snopoke The two should match at write time but diverge later: pricing_rule_id is on_delete=SET_NULL, so once a rule gets deleted or superseded the FK goes null on every historical row that referenced it. unit_price is a write-time snapshot that's never touched again.
> 
> Currently that is true however I would recommend protecting that FK relationship so that you can't delete a pricing rule that has usage data attached.

Okay, that's clear

davidnene:
@snopoke kindly review the fixes

snopoke (commented):
LGTM - please respond to outstanding comments if you aren't going to address them.

Also, please merge the Django migrations to avoid unnecessary migration accumulation.

davidnene:
> LGTM - please respond to outstanding comments if you aren't going to address them.
> 
> Also, please merge the Django migrations to avoid unnecessary migration accumulation.

I believe I have addressed all your comments, some I automatically  applied the suggestions. I have also consolidated the migrations.

snopoke (approved):

PR #3578, opened 2026-06-09
fix: bind task_id to session to prevent IDOR in chat poll endpoint

### LINKED ISSUE #3577: Chat task-poll endpoint does not bind task_id to the session (IDOR)
### Summary

`chat_poll_task_response` (`GET /api/chat/{session_id}/{task_id}/poll/`) does not verify that `task_id` belongs to `{session_id}`. The task result is fetched directly from the Celery result backend with no scoping, so a caller authorized for one session can read another session's task result by supplying its `task_id`.

### Details

`apps/api/views/chat.py`:

```python
def chat_poll_task_response(request, session_id, task_id):
    session = get_experiment_session_cached(session_id)
    ...
    experiment = session.experiment
    task_details = get_message_task_response(experiment, task_id)
```

`apps/experiments/task_utils.py::get_message_task_response`:

```python
def get_message_task_response(experiment, task_id: str):
    progress = Progress(AsyncResult(task_id)).get_info()
    ...
```

`AsyncResult(task_id)` is global to the Celery result backend. The `experiment` argument is used only for rendering the message (`_handle_success_result`), not for authorization. Nothing checks that the task was spawned for `session_id` (nor even for `experiment`). The `{session_id}` in the URL is used only for the `SessionAccessPermission` check; the returned payload is whatever task `task_id` produced.

### Impact

A caller who passes the permission check for session A (post-#3552: holding session A's valid token) can poll `/api/chat/{A}/{task_id}/poll/` with another session's `task_id` and read that task's result (the assistant reply, including attachments). Not blind-enumerable — `task_id`s are Celery UUIDs — but they are handed to the owning client by the send-message endpoint and may leak via logs/proxies.

PR #3552 (per-session tokens) narrows the precondition from "knows session A's id" to "holds a valid token for some session" but does not close the gap; the missing control is binding `task_id` ↔ `session_id`.

### Suggested fix

Bind the task to the session: persist the owning `session_id` when the message task is dispatched and verify it in `chat_poll_task_response` (return 404 on mismatch), or scope `get_message_task_response` by session. Add a regression test that polling session A's endpoint with session B's `task_id` returns 404.

### References

- Found during review of #3552 (chat session tokens).
- `apps/api/views/chat.py` — `chat_poll_task_response`
- `apps/experiments/task_utils.py` — `get_message_task_response`

ocs-percy:
### Product Description
No user-facing change. Security fix.

### Technical Description

`chat_poll_task_response` (`GET /api/chat/{session_id}/{task_id}/poll/`) fetched the Celery task result using only `task_id`, with no check that the task belonged to `{session_id}`. A caller holding a valid token for session A could read session B's assistant reply by supplying session B's `task_id`.

**Fix:**

In `chat_send_message`, after dispatching the task, write a cache binding:
```python
cache.set(f"task_session:{task_id}", str(session_id), 24 * 3600)
```

In `chat_poll_task_response`, verify the binding before hitting Celery:
```python
bound_session = cache.get(f"task_session:{task_id}")
if bound_session is not None and bound_session != str(session_id):
    raise NotFound()
```

A missing binding (`None`) is allowed through for backward compatibility with tasks dispatched before this fix was deployed. The TTL (24 h) matches the Celery result backend default.

**Regression tests added:**
- Task bound to correct session → 200
- Task bound to a different session → 404, Celery backend never queried
- No binding (legacy unbound task) → 200 (backward compatibility)

### Docs and Changelog

Closes #3577

### DIFF
--- a/api-schemas/v1.yml
+++ b/api-schemas/v1.yml
@@ -39,6 +39,12 @@ paths:
               schema:
                 $ref: '#/components/schemas/ChatTaskPollUserError'
           description: ''
+        '404':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ChatTaskPollNotFound'
+          description: ''
         '500':
           content:
             application/json:
@@ -1529,6 +1535,13 @@ components:
           type: string
       required:
       - status
+    ChatTaskPollNotFound:
+      type: object
+      properties:
+        detail:
+          type: string
+      required:
+      - detail
     ChatTaskPollStatusEnum:
       enum:
       - processing
--- a/apps/api/tests/test_chat_poll_api.py
+++ b/apps/api/tests/test_chat_poll_api.py
@@ -145,40 +145,27 @@ def test_chat_poll_generic_error_returns_500(api_client, mock_session, mock_task
 
 
 @pytest.mark.django_db()
-def test_chat_poll_task_bound_to_correct_session_succeeds(api_client, mock_session, mock_task_response):
-    """Polling with a task_id bound to the current session succeeds."""
-    task_id = "bound-task-ok"
-    cache.set(f"task_session:{task_id}", TEST_SESSION_ID, 60)
+@pytest.mark.parametrize(
+    ("cache_owner", "expected_status", "task_called"),
+    [
+        pytest.param(TEST_SESSION_ID, 200, True, id="bound-to-correct-session"),
+        pytest.param(str(uuid.uuid4()), 404, False, id="bound-to-different-session"),
+        pytest.param(None, 200, True, id="no-binding-backward-compat"),
+    ],
+)
+def test_chat_poll_task_session_binding(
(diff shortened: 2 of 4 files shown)


SmittieC (dismissed):

ocs-percy:
**CI fix pushed 🔧**

Found and fixed two issues causing the `python-tests` failure:

1. **Decorator misplacement** — When `_verify_task_belongs_to_session` was inserted, the `@extend_schema`, `@api_view`, `@authentication_classes` and `@permission_classes` decorators that belonged to `chat_poll_task_response` ended up on the helper instead. This stripped auth/permission checking from the poll endpoint and broke `test_task_poll_requires_token`.

2. **`return NotFound()` → `raise NotFound()`** — The early exit on missing session was returning the exception instance instead of raising it.

Also rebased onto `main` to pick up the session-token fix from #3579 (fixes `test_send_message_sunset_headers_for_deprecated_version`).

SmittieC (approved):

Issue #3586, opened 2026-06-10
Add /me API endpoint (alias for /userinfo) returning authenticated user info


ocs-percy:
## Summary

Add a `/api/v2/me/` endpoint returning basic information about the authenticated user and the team the current API key / OAuth token is scoped to.

## Motivation

A `/me` endpoint is a common REST API convention and provides a convenient, discoverable way for API consumers to retrieve their own user details and confirm which team their token is scoped to, without needing to know IDs upfront.

## Proposed Behaviour

- **Endpoint:** `GET /api/v2/me/`
- **Auth:** Requires authentication (same as other API endpoints)
- **Response:** Basic info about the authenticated user and the single team the token is scoped to, e.g.:
  ```json
  {
    "id": 123,
    "username": "example_user",
    "email": "user@example.com",
    "first_name": "Example",
    "last_name": "User",
    "team": {
      "id": 1,
      "name": "My Team",
      "slug": "my-team"
    }
  }
  ```
- Returns `401 Unauthorized` if not authenticated

## Notes

- API access in OCS is scoped to a single team per API key / OAuth token, so `team` is a single object (not a list).
- This endpoint belongs in v2 — v1 is frozen. The unversioned `/api/` alias is intentionally *not* added for this endpoint.

## Acceptance Criteria

- [ ] `GET /api/v2/me/` returns basic user info for the authenticated user
- [ ] Response includes the team the current token is scoped to (id, name, slug at minimum)
- [ ] Endpoint is documented in the API schema (OpenAPI/Swagger)
- [ ] Returns appropriate error for unauthenticated requests
- [ ] Tests cover authenticated and unauthenticated cases, and correct team returned

PR #3588, opened 2026-06-10
Add GET /api/v2/me/ endpoint

ocs-percy:
## Summary

Implements #3586.

Adds `GET /api/v2/me/` — returns basic info about the authenticated user and the team their API key / OAuth token is scoped to.

## Response shape

```json
{
  "id": 123,
  "username": "example_user",
  "email": "user@example.com",
  "first_name": "Example",
  "last_name": "User",
  "team": {
    "id": 1,
    "name": "My Team",
    "slug": "my-team"
  }
}
```

## Changes

- `apps/api/v2/serializers.py` — `MeSerializer` + `MeTeamSerializer`
- `apps/api/v2/views.py` — `MeView` (APIView, `IsAuthenticated` + `TokenHasOAuthResourceScope` with `required_scopes = []`)
- `apps/api/v2/urls.py` — wires up `me/` route
- `apps/api/v2/tests/test_me_api.py` — 6 tests covering URL routing, unauthenticated 401, api_key auth, OAuth auth, and team scoping

## Notes

- Lives in v2 only — no unversioned alias, consistent with v2 policy
- `required_scopes = []` means any valid OAuth token can call it (no specific scope needed)
- `team` is a single object, not a list — API tokens are scoped to one team

### DIFF
--- a/apps/api/v2/serializers.py
+++ b/apps/api/v2/serializers.py
@@ -1,8 +1,8 @@
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 
+from apps.api.serializers import TeamSerializer
 from apps.experiments.models import Experiment
-from apps.teams.models import Team
 from apps.users.models import CustomUser
 
 
@@ -30,12 +30,6 @@ class ChatbotSerializer(serializers.ModelSerializer):
         fields = ["id", "name", "url", "version_number", "versions"]
 
 
-class MeTeamSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = Team
-        fields = ["id", "name", "slug"]
-
-
 class MeSerializer(serializers.ModelSerializer):
     team = serializers.SerializerMethodField()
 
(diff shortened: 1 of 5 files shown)


apps/api/v2/serializers.py
```diff
@@ -25,3 +27,21 @@ class ChatbotSerializer(serializers.ModelSerializer):
...
+
+
+class MeSerializer(serializers.ModelSerializer):
+    team = serializers.SerializerMethodField()
+
+    class Meta:
+        model = CustomUser
+        fields = ["id", "username", "email", "first_name", "last_name", "team"]
+
+    def get_team(self, obj):
```
snopoke, line 41: This needs to use `@extend_schema_field` to add schema details.

apps/api/v2/serializers.py
```diff
@@ -25,3 +28,22 @@ class ChatbotSerializer(serializers.ModelSerializer):
     class Meta:
         model = Experiment
         fields = ["id", "name", "url", "version_number", "versions"]
+
+
+class MeTeamSerializer(serializers.ModelSerializer):
```
snopoke: shouldn't we reuse the `TeamSerializer` in `apps/api/serializers.py` to avoid creating new representations. The existing one doesn't include the `id` field but I don't think we should have that here either since `slug` is used as a natural key.

snopoke (approved):

PR #3603, opened 2026-06-11
Answer participant's original message after conversational consent (v2)

### LINKED ISSUE #3597: Conversational consent: participant's original message is dropped instead of answered after consent
## Summary

When conversational consent is enabled, the participant's **original message — the one that triggered the consent flow — is never processed by the bot**. After the participant consents, the bot either says nothing or sends a configured *seed* message, so the user's actual first question goes unanswered.

This was surfaced while reviewing #3596 (survey deprecation), which made the "nothing happens after consent" path explicit (see [Background](#background)).

## Current behaviour

The conversational consent flow spans at least two inbound messages:

1. **Triggering message** (status `SETUP` → `PENDING`): the participant sends their first message (e.g. *"Hi, how do I reset my password?"*). The flow transitions to `PENDING` and replies with the consent prompt. **The message content is used only as a trigger — it is never sent to the bot.**
2. **Consent token** (status `PENDING` → `ACTIVE`): the participant replies with the consent token (e.g. `1`). The flow transitions to `ACTIVE` and:
   - if the experiment has a **seed message**, the seed message is sent to the bot; otherwise
   - **nothing is sent** (after #3596 the stage raises `EarlyAbort`; before #3596 the consent token leaked to the bot as the first prompt).

Both inbound messages are persisted to chat history as `HUMAN` messages (in v2, `ChatMessageCreationStage` runs before `ConsentFlowStage`), but the bot never responds to either. The net effect: the participant asks a question, is interrupted for consent, consents — and their question is silently dropped.

## Proposed behaviour

After consent is accepted, **send the participant's original (pre-consent) message to the bot** instead of a seed message / nothing, so the bot answers what the participant actually asked.

## Open questions

- **Original vs. seed precedence.** If both an original message and a `seed_message` are configured, which wins? (Likely: original message if present, seed only when the participant didn't send substantive content first — but needs a decision.)
- **Which message counts as "original"?** The first message that triggered `SETUP → PENDING`, or the most recent non-consent message before `ACTIVE`? They're usually the same, but a participant could send several messages before consenting.
- **Retrieval mechanism.** The original message is already in chat history (persisted before the consent stage), but the consent exchange spans multiple requests, so the post-consent request must look it back up (or it must be stashed on the session).
- **Parity across both implementations:**
  - v2: `apps/channels/channels_v2/stages/core.py` — `ConsentFlowStage` (`SETUP` branch discards the trigger; `PENDING` + consent currently → seed or `EarlyAbort`).
  - v1: `apps/chat/channels.py` — `_handle_pre_conversation_requirements` / `start_conversation` (same shape).
- **Empty-message edge case.** If the participant's first message was itself just the consent token (no prior content), keep the current behaviour (consume it, send nothing).

## Background

Found during review of #3596. The consent stage was changed to raise `EarlyAbort` when consent is accepted and there is no seed message — previously the consent token was incorrectly forwarded to the bot as the first prompt. That fix correctly stops the *consent token* from leaking, but it also made it clear that the participant's *real* first message is never answered. This issue tracks the follow-up: process the original message after consent.

## Affected files

- `apps/channels/channels_v2/stages/core.py` (`ConsentFlowStage`)
- `apps/chat/channels.py` (`_handle_pre_conversation_requirements`, `start_conversation`)

snopoke:
### Product Description
When conversational consent is enabled, the participant's first message (the one that triggered the consent flow) was used only as a trigger and never answered. After consent, the bot sent a seed message or nothing, silently dropping the participant's actual question. Now the bot answers the participant's original message after consent.

### Technical Description
`ConsentFlowStage` (v2) now looks up the first substantive HUMAN message from chat history when consent is accepted and forwards it to the bot. Original message wins over the seed message; the seed is used only when there was no substantive first message. If the first message was itself just the consent token, the previous behaviour is preserved. The original message is passed through as the existing human_message so no duplicate HUMAN record is created.

Fixes #3597

### Migrations

### Docs and Changelog

### DIFF
--- a/apps/channels/channels_v2/stages/core.py
+++ b/apps/channels/channels_v2/stages/core.py
@@ -331,13 +331,6 @@ class ConsentFlowStage(ProcessingStage):
     def _start_conversation(self, ctx: MessageProcessingContext) -> str | None:
         ctx.experiment_session.update_status(SessionStatus.ACTIVE)
 
-        # The consent token (e.g. "1") was persisted as a HUMAN message by
-        # ChatMessageCreationStage, but it's a control token, not conversational
-        # content. Drop it before building the bot's history so it doesn't end
-        # up between the original message and the AI reply (which would feed the
-        # LLM two consecutive HUMAN messages and corrupt the context).
-        self._discard_consent_token_message(ctx)
-
         # Original message wins over the seed message: answer what the
         # participant actually asked before they were interrupted for consent.
         original_message = self._get_original_message(ctx)
@@ -347,21 +340,6 @@ class ConsentFlowStage(ProcessingStage):
         if ctx.experiment.seed_message:
             return self._process_message(ctx, ctx.experiment.seed_message)
         return None
-
-    def _discard_consent_token_message(self, ctx: MessageProcessingContext) -> None:
-        """Delete the consent-token HUMAN message created for this request.
-
-        ``_start_conversation`` only runs once consent has been given, so
-        ``ctx.human_message`` is the consent token itself. Removing it keeps the
-        token out of the persisted chat history and the LLM context. The trace's
-        input-message link is re-pointed when the original message is answered
-        (see ``_process_message``).
-        """
-        consent_message = ctx.human_message
-        if consent_message is None:
-            return
-        consent_message.delete()
-        ctx.human_message = None
 
     def _get_original_message(self, ctx: MessageProcessingContext) -> ChatMessage | None:
         """Return the participant's first substantive message -- the one that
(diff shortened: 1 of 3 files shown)


SmittieC (dismissed):

snopoke:
@claude please address the PR comments and resolve CI failures

apps/channels/channels_v2/stages/core.py
```diff
@@ -327,20 +330,85 @@ def _build_consent_prompt(self, ctx: MessageProcessingContext) -> str:
...
+        # content. Drop it before building the bot's history so it doesn't end
+        # up between the original message and the AI reply (which would feed the
+        # LLM two consecutive HUMAN messages and corrupt the context).
+        self._discard_consent_token_message(ctx)
+
+        # Original message wins over the seed message: answer what the
+        # participant actually asked before they were interrupted for consent.
+        original_message = self._get_original_message(ctx)
+        if original_message is not None:
+            return self._process_message(ctx, original_message.content, human_message=original_message)
```
SmittieC: Should we be setting the human message to the original message instead and let the rest execute like normal instead of calling `self._processi_message(..)` here

SmittieC (approved):

PR #3607, opened 2026-06-11
fix(widget): multi-tab fixes — kiosk visible persistence, ended sessions

snopoke:
### Product Description
Two widget fixes for multi-tab use:
- Visiting the chat page (kiosk widget) no longer causes the floating help widget to auto-open on other pages for the same chatbot.
- When a chat session ends elsewhere (another tab, or by the bot), the widget now stops polling, disables the composer, and shows a notice — instead of polling forever and only surfacing an error on send.

### Technical Description
- `saveVisibleState` now skips kiosk mode, mirroring the existing restore guard. The widget's `visible` localStorage key is scoped only by chatbot ID, so kiosk mode (which forces `visible = true`) was leaking into standard-mode widgets for the same bot.
- The message poll loop reacts to `session_status: "ended"` from `/api/chat/<id>/poll/`: polling stops and a new `sessionEnded` state disables the composer with a system notice (new `status.chatEnded` translation string). Unbound widgets recover via the "new chat" button.

Widget-only change; ships with the next widget release (no version bump in this PR).

### Migrations

### Docs and Changelog


SmittieC (approved):
Cool

PR #3630, opened 2026-06-14
feat(cost-tracking): capture path through tracer + pricing auto-seed (PR 2/3)

davidnene:
### Product Description

PR 2 of 3 in the AI Cost Tracking roadmap. Wires the capture path: when a team opts into the `flag_ai_cost_monitoring` flag, every LLM call in a chat turn produces a `UsageRecord` with a resolved per-call cost (or a flagged "unpriced" row when the model lacks a `PricingRule`). Capture is silent — no UI yet — so PR 3 has the data path to read from.

Also extends the daily `auto-update-models` workflow so newly added models get their pricing rules seeded automatically (llm-stats.com primary, LiteLLM fallback).

### Technical Description

- Provider tagging at `LlmService.get_chat_model()` via `metadata["ocs_provider_type"]` so every LangChain callback knows which OCS provider to bill.
- `MetricsCollector` extended to capture per-call provider/model, route missing `usage_metadata` to a fallback bucket (tiktoken for OpenAI/Azure, zero-quantity UNKNOWN otherwise), and drain into `UsageEvent`s with cache sub-buckets netted out.
- `OCSTracer` evaluates the flag once at trace entry and drains the collector into `UsageRecord` rows after `trace_record.save()`. Wrapped in the existing try/except so cost-recording failures never disrupt the trace.
- Workflow prompt teaches the auto-update agent to seed pricing + generate a `load_pricing_data()` migration alongside model registration.

### Deferred to PR 3

- Daily rate-change job (supersession migration + sibling workflow job).
- Cost dashboard, REST endpoints, weekly digest task.
- Unskip `test_seed_coverage` once the seed is backfilled for every registered model.

### Migrations

No schema migrations in this PR. Generated `load_pricing_data()` migrations are additive only.

### Demo

No UI. Verified end-to-end by `apps/service_providers/tracing/tests/test_ocs_tracer_cost.py`.


davidnene:
> [Review Change Stack](https://app.coderabbit.ai/change-stack/dimagi/open-chat-studio/pull/3630?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)
> 
> 📝 Walkthrough
> Comment `@coderabbitai help` to get the list of available commands and usage tips.

resolved

davidnene:
@snopoke kindly review

.github/workflows/auto-update-models.yml
snopoke, line 1: The prompt here is getting very complex and dealing with multiple different files and sources of data. I wonder if we could simplify it with a script to complement the LLM.

@SmittieC what do you think?

apps/service_providers/llm_service/main.py
```diff
@@ -186,7 +200,7 @@ def get_chat_model(self, llm_model: str, **kwargs) -> BaseChatModel:
                     model.tiktoken_model_name = "gpt-3.5-turbo"
                 case _:
                     model.tiktoken_model_name = "gpt-4"
-        return model
+        return self._tag_chat_model(model)
```
snopoke: instead of having to apply this in every subclass I think it would be better to apply this in the base class:

```python
def get_chat_model(....):
    model = self.._get_chat_model(...)  # abstract method in the base class, implemented in subclasses.
    return self._tag_chat_model(model)
```

apps/service_providers/tracing/tests/test_ocs_tracer_cost.py
```diff
@@ -0,0 +1,228 @@
...
+            "input_tokens": input_tokens,
+            "output_tokens": output_tokens,
+            "total_tokens": input_tokens + output_tokens,
+        },
+        response_metadata={"model_name": model},
+    )
+    return LLMResult(generations=[[ChatGeneration(message=message, text="response")]], llm_output=None)
+
+
+def _enable_flag_for(team) -> Flag:
```
snopoke: can you not use the `override_flag` decorator? https://waffle.readthedocs.io/en/stable/testing/automated.html#unit-tests

apps/service_providers/tracing/tests/test_ocs_tracer_cost.py
```diff
@@ -0,0 +1,228 @@
...
+
+class TestRecordCostsShortCircuits:
+    """Each branch where `_record_costs` should bail without touching the recorder.
+
+    Each test sets every guard except the one under inspection to truthy so a
+    regression in the targeted guard would actually fail this test.
+    """
+
+    def _collector_with_events(self):
+        from apps.service_providers.tracing.metrics import MetricsCollector  # noqa: PLC0415 - test scope
```
snopoke: inline imports should only be used to resolve circular dependencies or be in this list: https://github.com/dimagi/open-chat-studio/blob/main/pyproject.toml#L150-L161

apps/service_providers/llm_service/main.py
```diff
@@ -346,13 +360,14 @@ class AzureLlmService(LlmService):
     def get_chat_model(self, llm_model: str, **kwargs) -> BaseChatModel:
         from langchain_openai.chat_models import AzureChatOpenAI  # noqa: PLC0415 - TID253: heavy lib, slow startup
 
-        return AzureChatOpenAI(
+        model = AzureChatOpenAI(
```
SmittieC: How about `get_chat_model` moves to the base class, calls a `_chat_model` method that returns the custom chat model instances and calls `_tag_chat_model` as well? Then we don't have to remember to call `_tag_chat_model` when adding new providers.

Something like:
```python
class LlmService:
    def get_chat_model(self, llm_model: str, **kwargs) -> BaseChatModel:
        model = self._chat_model(llm_model: str, **kwargs):
        self._tag_chat_model(model)
        return model

    def _tag_chat_model(model):
        raise NotImplementedError

class AzureLlmService(LlmService):
    def _chat_model(self, llm_model: str, **kwargs) -> BaseChatModel:
        from langchain_openai.chat_models import AzureChatOpenAI  # noqa: PLC0415 - TID253: heavy lib, slow startup

        return AzureChatOpenAI(
            azure_endpoint=self.openai_api_base,
            openai_api_version=self.openai_api_version,
            openai_api_key=self.openai_api_key,
            deployment_name=llm_model,
            **kwargs,
        ) 
```

apps/service_providers/tracing/metrics.py
```diff
@@ -29,17 +45,117 @@ class MetricsCollector(UsageMetadataCallbackHandler):
...
+        Yields events from two sources:
+          - `self._exact_usage` -> confidence=EXACT (one bucket per (provider, model)).
+          - `self._fallback_usage` -> confidence=ESTIMATED or UNKNOWN.
+
+        Zero-quantity buckets are skipped except for UNKNOWN, which yields
+        a single zero-quantity LLM_INPUT row per (provider, model) so the
+        weekly digest can surface coverage gaps.
+        """
+        with self._counter_lock:
+            exact_snapshot = dict(self._exact_usage)
```
SmittieC, line 158: why is this step necessary?

apps/service_providers/tracing/metrics.py
```diff
@@ -54,14 +170,116 @@ def get_metrics(self) -> TraceMetrics:
...
+            total = sum(u.get("total_tokens", 0) for u in self.usage_metadata.values())
+        return prompt, completion, total
+
+
+def _sum_response_usage(response) -> dict[str, Any]:
+    """Sum `usage_metadata` across all generations on a chat response."""
+    totals: dict[str, Any] = {"input_tokens": 0, "output_tokens": 0, "input_token_details": {}}
+    for gens in response.generations:
+        for g in gens:
+            usage = getattr(getattr(g, "message", None), "usage_metadata", None) or {}
```
SmittieC, line 211: nit: is this check still needed if we established that there is usage metadata?

apps/service_providers/tracing/metrics.py
```diff
@@ -29,17 +45,117 @@ class MetricsCollector(UsageMetadataCallbackHandler):
...
+        provider, model = pending["provider"], pending["model"]
+        if has_usage_metadata(response):
+            self._exact_add(provider, model, _sum_response_usage(response))
+            return
+        if provider in OPENAI_FAMILY and pending["prompts"]:
+            input_tokens = tiktoken_count(model, pending["prompts"])
+            output_tokens = tiktoken_count(model, response_text(response))
+            self._fallback_add((provider, model, Confidence.ESTIMATED), input_tokens, output_tokens)
+        else:
+            self._fallback_add((provider, model, Confidence.UNKNOWN), 0, 0)
```
SmittieC, line 120: in other places we use `langchain_core.messages.utils.count_tokens_approximately` which might be helpful here?

SmittieC (commented):
Some initial thoughts.

apps/service_providers/llm_service/main.py
davidnene: Nice one. Template Method pattern fits this exactly: tagging is policy that belongs to the base class, model construction is the per-provider implementation. Just one small huddle, a future subclass that overrides get_chat_model directly (bypassing the template method) would skip tagging. we can add a comment for that.

The base class can be as below

```python
class LlmService(pydantic.BaseModel):
    _type: str = "unknown"

    def get_chat_model(self, llm_model: str, **kwargs) -> BaseChatModel:
        return self._tag_chat_model(self._chat_model(llm_model, **kwargs))

    def _chat_model(self, llm_model: str, **kwargs) -> BaseChatModel:
        raise NotImplementedError

    def _tag_chat_model(self, model: BaseChatModel) -> BaseChatModel:
        model.metadata = {**(model.metadata or {}), "ocs_provider_type": self._type}
        return model

File: apps/service_providers/tracing/metrics.py (hunk 5 above)
Comment (davidnene) [reply apps/service_providers/tracing/metrics.py:158]: 1. iter_cost_events is a generator. Iterating self._exact_usage directly would mean holding _counter_lock across yield points — any in-flight _exact_add / _fallback_add would block on the lock until the consumer finished. Snapshotting under the lock and releasing before yielding keeps the critical section short (O(n) where n is "distinct (provider, model) pairs", usually 1–5).

2. LangGraph runs nodes in threads (per the class docstring). In practice all callbacks should have joined by trace finalization, but a late on_llm_end would either cause RuntimeError: dictionary changed size during iteration or silently inconsistent reads. Cheap to defend against, expensive to debug if it ever fires.

File: apps/service_providers/tracing/metrics.py (hunk 6 above)
Comment (davidnene) [reply apps/service_providers/tracing/metrics.py:211]: Yes coz has_usage_metadata uses any, not all, so a response with mixed generations (some carrying usage_metadata, some None) passes the upstream check but would AttributeError here on usage.get(...). Rare in chat models but possible (LangChain LLMResult allows multiple generations per response, e.g. n>1), and the or {} is cheap. Could tighten has_usage_metadata to all instead, but that would silently drop partial-usage responses rather than count what we have, current behaviour adds 0 for the missing ones, which is the correct sum.

File: apps/service_providers/tracing/metrics.py (hunk 7 above)
Comment (davidnene) [reply apps/service_providers/tracing/metrics.py:120]: that fits well too. These are Confidence.ESTIMATED rows already so paying for tiktoken's precision here is overkill. We can swap + drop the gate, broaden the ESTIMATED path to all providers.

**Comment by davidnene** [comment]
@SmittieC kindly review the replies and advise on whether to proceed with the changes

**Review by snopoke — COMMENTED** [review]
I wasn't able to complete my review but these are my comments from before the last round of changes.

File: apps/service_providers/tracing/metrics.py (hunk 7 above)
Comment (SmittieC) [reply apps/service_providers/tracing/metrics.py:120]: I think the `provider in OPENAI_FAMILY` branch above is fine, but the `Confidence.UNKNOWN` branch can maybe use `count_tokens_approximately`. It's better than reporting 0 token usage if in fact there were token usage. Something like

```python
if has_usage_metadata(response):
    self._exact_add(provider, model, _sum_response_usage(response))
    return
if provider in OPENAI_FAMILY and pending["prompts"]:
    input_tokens = tiktoken_count(model, pending["prompts"])
    output_tokens = tiktoken_count(model, response_text(response))
    self._fallback_add((provider, model, Confidence.ESTIMATED), input_tokens, output_tokens)
else:
    input_tokens = count_tokens_approximately([pending["prompts"]])
    output_tokens = count_tokens_approximately[response_text(response)]
    self._fallback_add((provider, model, Confidence.ESTIMATED), input_tokens, output_tokens)

```

Thoughts?

.github/workflows/auto-update-models.yml
SmittieC: Good call. I'll make an issue for this

apps/service_providers/tracing/metrics.py
davidnene: This is solid. I agree. Working on it

apps/service_providers/llm_service/main.py
davidnene: this was fixed

apps/service_providers/tracing/tests/test_ocs_tracer_cost.py
davidnene: noted

apps/service_providers/tracing/tests/test_ocs_tracer_cost.py
davidnene: noted

.github/workflows/auto-update-models.yml
davidnene: Nice. @snopoke @SmittieC do we work on this after this merge approval? or after the full task completion? PR3 pending.
I have resolved all your reviews as well.

apps/service_providers/tracing/tests/test_ocs_tracer_cost.py
snopoke: I notice that you're still manually creating flags and updating the cache - is there a reason the waffle util doesn't work?

apps/service_providers/tracing/ocs_tracer.py
```diff
@@ -149,6 +152,49 @@ def _update_trace_metrics(self) -> None:
...
+        )
+
+    def _evaluate_cost_tracking_flag(self) -> bool:
+        """Look up the team-scoped `flag_ai_cost_monitoring` flag once per
+        trace entry. `Team(pk=...)` is a Django shortcut: `is_active_for_team`
+        only reads `.pk`, so we avoid an unnecessary DB fetch of the Team row.
+        """
+        from apps.teams.models import Flag, Team  # noqa: PLC0415 - lazy: avoid circular at import time
+
+        flag = Flag.objects.filter(name="flag_ai_cost_monitoring").first()
```
snopoke: You should use `Flag.get("flag_ai_cost_monitoring")` which is a cached getter.

Or `get_waffle_flag_model().get("flag_ai_cost_monitoring")`

apps/service_providers/tracing/ocs_tracer.py
```diff
@@ -149,6 +152,49 @@ def _update_trace_metrics(self) -> None:
...
+                participant_id=self.trace_record.participant_id,
+            ),
+        )
+
+    def _evaluate_cost_tracking_flag(self) -> bool:
+        """Look up the team-scoped `flag_ai_cost_monitoring` flag once per
+        trace entry. `Team(pk=...)` is a Django shortcut: `is_active_for_team`
+        only reads `.pk`, so we avoid an unnecessary DB fetch of the Team row.
+        """
+        from apps.teams.models import Flag, Team  # noqa: PLC0415 - lazy: avoid circular at import time
```
snopoke: I don't believe that this is a circular reference

snopoke (dismissed):
LGTM - comments not blocking

apps/service_providers/tracing/ocs_tracer.py
davidnene: resolved

apps/service_providers/tracing/ocs_tracer.py
davidnene: fixed

snopoke (dismissed):

snopoke (approved):

apps/service_providers/tracing/tests/test_ocs_tracer_cost.py
davidnene: switched _evaluate_cost_tracking_flag to Flag.get(...) (the cached classmethod) also replaced the wholesale cache.clear() autouse fixture with a targeted flag.flush() inside the enable_team_flag helper. override_flag still doesn't fit (it toggles everyone, which is_active_for_team deliberately ignores), but the cache-handling smell is gone

davidnene:
@snopoke I've resolved your comments

PR #3636, opened 2026-06-15
feat(redesign-new-filter-url-params): adds new url params format

### LINKED ISSUE #2033: Re-design filter url parameters
Currently filter URLs are constructed as a set of `filter_N_column`, `filter_N_operator`, `filter_N_value` triplets where the column and operator are plain parameters and the value is encoded as JSON.

This presents a few issues but most notably the length of the URL can quickly grow to exceed the maximum limit of 2048 characters.

# New Format

This new format uses more native URL parameters rather than JSON encoding and does not separate the column name and value. The operator is kept separate to avoid repetition when there are multiple values being filtered

## Single value
`f_{column_name}={value}`
`op_{column_name}={operator}`

## Multiple values separated by `~`

Use `~` or separator since it is not a reserved URL character.

`f_{column_name}={value1}~{value2}`
`op_{column_name}={operator}`

If a value contains `~`, the value should be quoted with double quotes: 

    f_tag=tag1~"tag~2"~tag3

## Parsing on the backend

Python's CSV parser can be used:

```
csv.reader(StringIO(raw_value), delimiter="~")
```

# Alternative implementation

An alternative approach is to use a POST request to send the data to the backend which responds with a UUID which is then used in the get request and in the URL. 

The downside of this approach is that the URL is unreadable and that the filter URL will expire after some time which means a different solution is needed for creating permalinks.

brian-mecha:
### Product Description

This new URL parameters format uses more native URL parameters rather than JSON encoding and does not separate the column name and value. The operator is kept separate to avoid repetition when there are multiple values being filtered

### Technical Description

This change introduces a new URL filter parameter format designed to reduce URL length and improve readability. The previous filter_N_column, filter_N_operator, and filter_N_value structure has been replaced with column-based parameters using the format f_{column} for values and op_{column} for operators.

The new format supports both single and multi-value filters, with multiple values represented as a ~-separated list. Values containing the separator character are enclosed in double quotes. The implementation maintains backward compatibility by supporting both legacy and new filter formats during the migration period.

#### Old format
```
filter_0_column=status
filter_0_operator=in
filter_0_value=["active","pending"]
```

#### New format
```
f_status=active~pending
op_status=in 
```

### Demo

Screenshot 2026-06-15 at 8 38 25 PM

### Docs and Changelog

Resolves [2023](https://github.com/dimagi/open-chat-studio/issues/2033)


apps/web/dynamic_filters/datastructures.py
```diff
@@ -55,7 +77,7 @@ def __init__(self, query_params: QueryDict | None = None, column_filters: list[C
     @classmethod
     def from_request(cls, request) -> Self:
         query_params = request.GET
-        if not any(key.startswith("filter_") for key in query_params):
+        if not any(key.startswith(("filter_", "f_")) for key in query_params):
```
SmittieC: I think we can drop the `filter_` entry

templates/filters/_filter_component_alpine.html
```diff
@@ -203,17 +203,64 @@
...
+
+        values.push(current);
+
+        return values;
+      },
+
       loadFiltersFromQueryString(queryString) {
         const urlParams = new URLSearchParams(queryString);
         const filters = [];
+        const newFormatColumns = new Set();
```
SmittieC: nit: references to the "old" format only creates unnecessary clutter that needs cleanup down the line. You can just call this `columns`

templates/filters/_filter_component_alpine.html
```diff
@@ -287,10 +359,15 @@
             params.delete(key);
```
SmittieC: any reason to keep this part around?

SmittieC (commented):
Generally looks good to me, thanks! I do think we need a data migration to update saved filters so they use the new format

SmittieC (commented):
Almost there. We just need to update `FilterSet.filter_query_string` to the new format as well

apps/dashboard/migrations/0002_migrate_saved_filters_format.py
```diff
@@ -0,0 +1,21 @@
+from django.db import migrations
+
+from apps.dashboard.filter_format import convert_saved_filter_data
+
+
+def migrate_saved_filters(apps, schema_editor):
+    DashboardFilter = apps.get_model("dashboard", "DashboardFilter")
```
SmittieC: DashboarFilter doesn't use the same filter params that `FilterSet` does, so you should only be updating `FilterSet` in here. (We want to make the dashboard use the same at some time)

SmittieC:
A few more things beyond my inline comments worth looking into (credit to fable 5):

- `DatasetAutoPopulationRule.filter_query_string` also stores old-format query strings — after this ships those rules would silently go from "filtered" to "all sessions", so it needs the same migration treatment as `FilterSet`.
- Some code still generates old-format links, which now land on unfiltered pages: `templates/trace/trace_detail.html`, `templates/chatbots/chatbot_session_view.html`, and the tag drill-down in `assets/javascript/dashboard/main.js`.
- ~23 tests in files that weren't updated still use the old format and fail: `apps/trace/tests/test_filters.py`, `apps/participants/tests/test_filters.py`, `apps/evaluations/tests/test_add_sessions_view.py`.
- Quote escaping mismatch: the JS serializer escapes `"` as `\"`, but Python's csv parser (and the JS parser) expect doubled quotes (`""`), so values containing both `~` and `"` get corrupted. One escaping rule everywhere please — and quote values containing `"` even without `~`, to match `csv.QUOTE_MINIMAL`.
- In `datastructures.py` the CSV parse only runs when the value contains `~`, so quoted values without `~` don't round-trip. Related: with the old format gone, the `json.loads` fallback in `_normalize_list_value` can be dropped, otherwise e.g. `f_tags=[1,2]` gets re-interpreted as JSON.
- Minor: `setLabelFromURL` lost the `else` that cleared a stale date-range label, and the docs describe the old format as "Deprecated" when support was removed entirely.

apps/dashboard/migrations/0002_migrate_saved_filters_format.py
```diff
@@ -1,21 +1,32 @@
...
 
 from apps.dashboard.filter_format import convert_saved_filter_data
 
 
 def migrate_saved_filters(apps, schema_editor):
-    DashboardFilter = apps.get_model("dashboard", "DashboardFilter")
-    for filter_obj in DashboardFilter.objects.all():
-        if not isinstance(filter_obj.filter_data, dict):
+    FilterSet = apps.get_model("filters", "FilterSet")
+    for filter_set in FilterSet.objects.all():
```
SmittieC: For data migrations where we iterate like this, prefer bulk fetching them instead. See an example [here](https://github.com/dimagi/open-chat-studio/blob/17840117c8b955c45aff42e42e95830bc9dc3070/apps/documents/management/commands/populate_unsupported_channels.py#L48-L68). Same with `migrate_auto_population_filter_query_strings`

apps/evaluations/migrations/0016_migrate_auto_population_filter_query_strings.py
```diff
@@ -0,0 +1,33 @@
...
+
+        try:
+            params = FilterParams(QueryDict(raw_query))
+        except Exception:
+            continue
+
+        if not params.filters:
+            continue
+
+        rule.filter_query_string = params.to_query()
```
SmittieC: The string conversion is still needed, no?

SmittieC: @brian-mecha this question still stands. This doesn't seem to do a proper conversion?

SmittieC:
> [!NOTE]
> This comment was written by Claude (on behalf of @SmittieC) after reviewing the branch.

**JSON-array filter values are misinterpreted by the new normalizer, breaking two in-tree filter producers.**

`ColumnFilterData._normalize_list_value` (`apps/web/dynamic_filters/datastructures.py`) only attempts to parse a value as a list when it contains `~`. A JSON-encoded array like `[123]` or `["v1", "v2"]` (no tilde) is wrapped whole as a single literal string. Two production code paths still produce JSON arrays:

1. **`Experiment.traces_url`** (`apps/experiments/models.py:859-862`) constructs `ColumnFilterData(value=json.dumps([self.id]))` directly, and the validator runs on construction. The string-list case (`versions`) happens to get rescued by the double-unwrap in `to_query`, but the int-list case (`experiment`) doesn't: `_try_parse_json_string_list` rejects `[123]` because the items aren't strings, so the generated URL is `f_experiment=[123]`. When that URL is parsed back on request, `"[123]"` contains no `~`, so it's wrapped as a literal → `experiment__in=["[123]"]` → matches nothing. The "View traces" link from an experiment is functionally broken.

2. **Dashboard → sessions link** (`assets/javascript/dashboard/main.js:812`) still does `parsedValue = JSON.stringify(...)` and appends it as `f_{col}`, so the server receives a JSON array string in the query param and hits the same wrapping failure. (`buildBaseTagDynamicFilter` at line 776 was converted; this path wasn't.)

Most of the currently failing filter tests (`apps/trace`, `apps/participants`, `apps/experiments`) are symptoms of this same mismatch — they pass JSON list values.

**Suggested fix:** either convert both producers to the new tilde format, or make `_normalize_list_value` fall back to JSON parsing when the value contains no `~` (matching the old format's behavior). The cleaner long-term shape is to keep one canonical internal representation for `ColumnFilterData.value` (JSON array, as before) and confine the tilde-CSV encoding strictly to the URL encode/decode boundary — that would also remove the need for the double-unwrap in `to_query`.

SmittieC:
> [!NOTE]
> This comment was written by Claude (on behalf of @SmittieC) after reviewing the branch.

Two more issues, both consequences of legacy `filter_N_*` parsing having been removed from `FilterParams` entirely:

**1. The evaluations data migration is a no-op — saved auto-population rules are never migrated.**

`apps/evaluations/migrations/0016_migrate_auto_population_filter_query_strings.py` parses legacy strings with:

```python
params = FilterParams(QueryDict(raw_query))
...
if not params.filters:
    continue
```

But `FilterParams.__init__` at this branch's head only reads `f_*`/`op_*` keys, so a legacy `filter_0_column=...` query string always yields empty `params.filters` and every rule is skipped. Since runtime parsing of the legacy format is also gone, existing rules with legacy `filter_query_string` values will pass empty filters to auto-population — meaning they'd ingest **every** session into their datasets instead of the filtered subset. Silent data corruption.

Suggested fix: give the migration its own self-contained legacy parser instead of importing live app code (this is exactly why migrations shouldn't depend on application code — the app code moved on before the migration ran). Alternatively, restoring legacy parsing in `FilterParams` (see the backward-compat comment above) makes the migration work as written.

**2. "Create dataset" still generates legacy params the backend now ignores.**

`templates/chatbots/table_actions_dropdown.html` (`createDatasetWithFilters`) still appends:

```js
params.set(`filter_${filterIndex}_column`, 'experiment');
params.set(`filter_${filterIndex}_operator`, 'any of');
params.set(`filter_${filterIndex}_value`, JSON.stringify([experimentId]));
```

The backend discards these, so the dataset-creation flow silently loses its experiment scoping. This needs converting to `f_experiment`/`op_experiment` — and note it also JSON-encodes the value, so once converted it will hit the JSON-array normalization issue described in my previous comment.

apps/evaluations/migrations/0016_migrate_auto_population_filter_query_strings.py
brian-mecha: @SmittieC Resolved

apps/evaluations/migrations/0016_migrate_auto_population_filter_query_strings.py
```diff
@@ -14,22 +67,16 @@ def migrate_auto_population_filter_query_strings(apps, schema_editor):
...
-        if not params.filters:
+        if raw_query.startswith(("f_", "op_")):
             continue
 
-        rule.filter_query_string = str(params.to_query())
-        rules_to_update.append(rule)
-        if len(rules_to_update) >= batch_size:
-            DatasetAutoPopulationRule.objects.bulk_update(rules_to_update, ["filter_query_string"])
-            rules_to_update.clear()
+        converted_query_string = _convert_legacy_filter_query_string_to_new_format(raw_query)
```
SmittieC: You can just call `apps.dashboard.filter_format.convert_saved_filter_data` here instead of  adding a function that duplicates that function

SmittieC:
@brian-mecha there are merge conflicts

brian-mecha:
@SmittieC Merge conflicts resolves and tests fixed.

apps/evaluations/migrations/0016_migrate_auto_population_filter_query_strings.py
```diff
@@ -0,0 +1,38 @@
...
+
+    for rule in DatasetAutoPopulationRule.objects.all().iterator(chunk_size=batch_size):
+        raw_query = rule.filter_query_string or ""
+        if not raw_query:
+            continue
+
+        if raw_query.startswith(("f_", "op_")):
+            continue
+
+        converted_query_string = convert_saved_filter_data(raw_query)
```
SmittieC: You need to add the following before this step:
```python
raw_query = filter_set.filter_query_string or ""
if not raw_query:
    continue
```

or move this logic into `convert_saved_filter_data` and make `convert_saved_filter_data` accept strings. If you do this, remember to update apps/dashboard/migrations/0002_migrate_saved_filters_format.py as well

brian-mecha: @SmittieC that validation rule is already present in `apps/evaluations/migrations/0016_migrate_auto_population_filter_query_strings.py` line 16-18. 

Screenshot 2026-07-17 at 3 10 56 PM

SmittieC: Ah, I must have looked at an old commit, thanks. The strings part is still an issue it seems. You're passing in a string to convert_saved_filter_data in this migration, but in apps/dashboard/migrations/0002_migrate_saved_filters_format.py you're passing in a `QueryDict` instance. Maybe we should make convert_saved_filter_data accept strings?

package-lock.json
```diff
@@ -10555,9 +10555,7 @@
       }
     },
     "node_modules/tailwindcss": {
-      "version": "4.3.2",
-      "resolved": "https://registry.npmjs.org/tailwindcss/-/tailwindcss-4.3.2.tgz",
-      "integrity": "sha512-WtctNNSH8A9jlMIqxzuYumOHU5uGZyRv0Q5svQl+oEPy5w84YpBxdb7MdqyiSPQge5jTJ6zFQLq0PFygdccSBA==",
+      "version": "4.1.5",
```
SmittieC: why this change?

apps/evaluations/migrations/0016_migrate_auto_population_filter_query_strings.py
brian-mecha: @SmittieC This has been refactored and resolved

package-lock.json
brian-mecha: Resolved.

SmittieC: Thanks

apps/dashboard/migrations/0002_migrate_saved_filters_format.py
```diff
@@ -0,0 +1,34 @@
+from django.db import migrations
+
+from apps.dashboard.filter_format import convert_saved_filter_data
+
+
+def migrate_saved_filters(apps, schema_editor):
+    FilterSet = apps.get_model("filters", "FilterSet")
```
SmittieC, line 7: I'm only noticing this now this migration mutates FilterSet which is in the `filters` app, so please move this migration to that app. Additionally, the definition and tests of `convert_saved_filter_data` should also be moved

apps/dashboard/tests/test_models.py
```diff
@@ -73,3 +76,45 @@ def test_filter_creation(self, team, user):
         assert dashboard_filter.filter_data == filter_data
         assert dashboard_filter.team == team
         assert dashboard_filter.user == user
+
+    def test_convert_saved_filter_data_to_new_format(self):
```
SmittieC: These tests doesn't belong in this file / class anymore

apps/dashboard/filter_format.py
```diff
@@ -0,0 +1,66 @@
...
+        if "value" in legacy_filter:
+            value = legacy_filter["value"]
+            if isinstance(value, str):
+                try:
+                    parsed = json.loads(value)
+                except (TypeError, json.JSONDecodeError):
+                    parsed = None
+
+                if isinstance(parsed, list):
+                    value = "~".join(str(item) for item in parsed)
```
SmittieC: `"~".join(...)` doesn't quote values that contain the separator (same pattern at line 57). `["tag~2", "a"]` migrates to `f_tags=tag~2~a`, which the new parser reads back as `["tag", "2", "a"]` — silent corruption of saved filters, and the migration's reverse is a noop so it can't be undone.

Use the CSV writer that the parser is the inverse of: `_serialize_csv_tilde_values` in `apps/web/dynamic_filters/datastructures.py` produces `"tag~2"~a`, which round-trips correctly. Worth promoting it to a public name since it becomes the single source of truth for the wire format. Please also add a test that round-trips a `~`-containing value through the converter and `FilterParams` — that's the test that would have caught this.

apps/web/dynamic_filters/datastructures.py
```diff
@@ -18,14 +60,22 @@ class ColumnFilterData(BaseModel):
...
+        """Normalize list-based filter values to JSON arrays for operators that expect lists."""
         if self.operator in _LIST_OPERATORS:
-            try:
-                parsed = json.loads(self.value)
-                if not isinstance(parsed, list):
+            if isinstance(self.value, str):
+                parsed = _coerce_list_filter_values(self.value)
+                if parsed is not None:
+                    self.value = json.dumps(parsed)
+                elif "~" in self.value:
```
SmittieC: Quoting asymmetry with the frontend: the JS `serializeCSVTildeValues` quotes values containing `~` **or** `"` (matching `csv.QUOTE_MINIMAL`), but this branch only CSV-parses when the raw value contains `~`. A value like `fo"o` arrives as `"fo""o"` and is stored as the literal string `"fo""o"` — the filter silently matches nothing, and Python→Python `to_query()` round-trips break the same way. Always run the CSV parse (a bare value without delimiters parses to itself), or also trigger it on a leading `"`.

templates/chatbots/table_actions_dropdown.html
```diff
@@ -26,17 +26,12 @@
...
       const params = new URLSearchParams(window.location.search);
-      let filterIndex = 0;
-      while (params.has(`filter_${filterIndex}_column`)) {
-        filterIndex++;
-      }
 
-      params.set(`filter_${filterIndex}_column`, 'experiment');
-      params.set(`filter_${filterIndex}_operator`, 'any of');
-      params.set(`filter_${filterIndex}_value`, JSON.stringify([experimentId]));
+      params.set('f_experiment', JSON.stringify([experimentId]));
```
SmittieC: This sets the new-format param to a JSON array string: `f_experiment=["123"]` instead of `f_experiment=123`. Two different parsers then see that value:

- The **backend** happens to handle it, because `_coerce_list_filter_values` still accepts JSON lists as a compatibility fallback — so the filtered results come back correctly.
- The **frontend** Alpine loader on the destination page does not: `parseCSVTildeValue` only understands the tilde/CSV format, so it treats `["123"]` as one literal string value, `[123]`. That means the UI can't match it to an experiment id, and the filter pill either displays the raw `[123]` text or (for object-choice filters, which match selected values by id) drops the value entirely.

The result is a page that *is* filtered but doesn't *show* the active filter — confusing for the user and easy to misread as a bug elsewhere. Emit the plain value like the other producers do: `params.set('f_experiment', experimentId)`. Same problem in `assets/javascript/dashboard/main.js:861`, which wraps values in `JSON.stringify(...)` before building `f_` params.

templates/filters/_filter_component_alpine.html
```diff
@@ -203,17 +203,64 @@
...
+        const columns = new Set();
+
+        // Detect columns using new format (f_*) vs old format (filter_*_column)
+        for (const [key] of urlParams.entries()) {
+          if (key.startsWith('f_')) {
+            columns.add(key.substring(2));
+          }
+        }
+
         let maxIndex = -1;
```
SmittieC: `loadFiltersFromQueryString` has accumulated some confusing scaffolding:

- The `maxIndex` loop declares `valueParam`/`operatorParam` and never uses them.
- The outer `let newFormatIndex = 0` (line 278) is shadowed by the `forEach` parameter below it, and the `newFormatIndex++` at the end of the callback mutates the callback parameter — dead code.
- `filterIndex = Object.keys(filters).length - validNewFormatColumns.length + newFormatIndex` always equals `newFormatIndex`, since both loops use the identical f_+op_ criterion.

This can collapse to: collect valid columns, then one `forEach` that pushes a populated filter object. The indentation in the second half of the function also has an extra level, which makes the block hard to review.

templates/filters/_filter_component_alpine.html
```diff
@@ -226,49 +273,63 @@
...
-                  filter.value = valueParam;
+
+        // Process new format filters
+        let newFormatIndex = 0;
+        const validNewFormatColumns = Array.from(columns)
+          .filter(columnName =>
+            urlParams.has(`f_${columnName}`) &&
+            urlParams.has(`op_${columnName}`)
+          )
+          .sort();
```
SmittieC: Sorting the columns alphabetically re-orders the user's filters on every page load. `URLSearchParams` preserves insertion order — iterate in order of first appearance instead.

templates/evaluations/auto_population_rule_form.html
```diff
@@ -25,7 +25,7 @@
     <script>
       (function(){
         const existing = "{{ object.filter_query_string|default_if_none:''|escapejs }}";
-        if (existing && !window.location.search.includes('filter_')) {
+        if (existing) {
```
SmittieC: The old guard (`!window.location.search.includes('filter_')`) was removed rather than updated, so the saved `filter_query_string` now unconditionally clobbers any filter params already present in the URL via `history.replaceState`. The guard should check for `f_`/`op_` params instead.

apps/dashboard/migrations/0002_migrate_saved_filters_format.py
```diff
@@ -0,0 +1,34 @@
...
+    batch_size = 500
+    filter_sets_to_update = []
+
+    for filter_set in FilterSet.objects.all().iterator(chunk_size=batch_size):
+        raw_query = filter_set.filter_query_string or ""
+        if not raw_query:
+            continue
+
+        if raw_query.startswith(("f_", "op_")):
```
SmittieC: A few structural points on the migrations (this one and the evaluations one):

- Both import `convert_saved_filter_data` from live app code, so future edits to that function silently change what the migration does for anyone who hasn't run it yet. Prefer a self-contained copy inside each migration.
- This migration mutates the **filters** app's model, so it belongs in `apps/filters/migrations/`. Relatedly, `filter_format.py` lives in `apps/dashboard` but serves the filters and evaluations apps — it should move next to `datastructures.py`.
- The already-migrated check `raw_query.startswith(("f_", "op_"))` misses new-format strings that begin with a non-filter param (FilterSets store the full query string, e.g. `page=2&f_tags=x`). Detect the legacy format positively (presence of `filter_\d+_column`) instead.
- There are no tests for the RunPython functions — a migration test with a `~`-containing value would have caught the corruption bug above.

SmittieC (changes requested):
Thanks for taking this on — the core format is solid. It matches the design in #2023 exactly (including the `csv.reader` parsing suggestion), all the producer call sites were swept, the stored-data migrations are batched properly, and the URL-length goal is clearly achieved. The happy-path test coverage is good too.

One blocking issue: **the data migration corrupts saved values containing `~`**, and the reverse migration is a noop so it's unrecoverable — details inline. Beyond that, there's a cluster of frontend issues in the Alpine loader (quoting asymmetry with the backend, lost filter ordering, dead code) and some migration-structure concerns, all noted inline.

**Minor nits** (grouped here since some files aren't in the diff):

- `apps/web/dynamic_filters/base.py:131` — `ColumnFilter`'s docstring still describes the `filter_{i}_column` contract.
- `templates/filters/_filter_component_alpine.html` — garbled comment "// Remove old format new parameters".
- `_parse_csv_tilde_values("a~~b")` yields `["a", "", "b"]` — consider dropping empty elements.
- `getDateRangeFilterIndex` in the Alpine template is now dead code — its only caller was rewritten.
- The PR body's example uses operator `in`, which doesn't exist (`any of`/`all of`/`excludes`); the committed docs are correct.

SmittieC:
(Sorry for the plethora of comments!)

apps/evaluations/migrations/0016_migrate_auto_population_filter_query_strings.py
```diff
@@ -0,0 +1,97 @@
...
+    DatasetAutoPopulationRule = apps.get_model("evaluations", "DatasetAutoPopulationRule")
+    batch_size = 500
+    rules_to_update = []
+
+    for rule in DatasetAutoPopulationRule.objects.all().iterator(chunk_size=batch_size):
+        raw_query = rule.filter_query_string or ""
+        if not raw_query:
+            continue
+
+        converted = _convert_legacy_query_string(raw_query)
```
SmittieC: Why not reuse convert_saved_filter_data from apps/web/dynamic_filters/filter_format.py?

apps/filters/migrations/0004_migrate_saved_filters_format.py
```diff
@@ -0,0 +1,97 @@
...
+    FilterSet = apps.get_model("filters", "FilterSet")
+    batch_size = 500
+    filter_sets_to_update = []
+
+    for filter_set in FilterSet.objects.all().iterator(chunk_size=batch_size):
+        raw_query = filter_set.filter_query_string or ""
+        if not raw_query:
+            continue
+
+        converted = _convert_legacy_query_string(raw_query)
```
SmittieC: Same comment from above: "Why not reuse convert_saved_filter_data from apps/web/dynamic_filters/filter_format.py?"

templates/evaluations/auto_population_rule_form.html
```diff
@@ -25,7 +25,7 @@
     <script>
       (function(){
         const existing = "{{ object.filter_query_string|default_if_none:''|escapejs }}";
-        if (existing && !window.location.search.includes('filter_')) {
+        if (existing && !window.location.search.includes('f_/op_')) {
```
SmittieC: I think you meant something like this?
```JS
["f_", "op_"].some(str => window.location.search.includes(str))
```

templates/filters/_filter_component_alpine.html
```diff
@@ -203,72 +203,100 @@
...
           }
         }
-        for (let i = 0; i <= maxIndex; i++) {
-          filters.push({
-            column: '',
+
+        // Keep only columns that have both a value (f_*) and an operator (op_*)
+        const validColumns = Array.from(columns)
+          .filter(columnName => urlParams.has(`f_${columnName}`) && urlParams.has(`op_${columnName}`))
+          .sort();
```
SmittieC: Nit: if we sort, the user loses the order in which they created the filter. Was there a specific reason for adding this `.sort()`?

apps/evaluations/migrations/0016_migrate_auto_population_filter_query_strings.py
brian-mecha: @SmittieC , you had mentioned, in this comment,  that you `Prefer a self-contained copy inside each migration`.

> A few structural points on the migrations (this one and the evaluations one):
> 
> * Both import `convert_saved_filter_data` from live app code, so future edits to that function silently change what the migration does for anyone who hasn't run it yet. Prefer a self-contained copy inside each migration.
> * This migration mutates the **filters** app's model, so it belongs in `apps/filters/migrations/`. Relatedly, `filter_format.py` lives in `apps/dashboard` but serves the filters and evaluations apps — it should move next to `datastructures.py`.
> * The already-migrated check `raw_query.startswith(("f_", "op_"))` misses new-format strings that begin with a non-filter param (FilterSets store the full query string, e.g. `page=2&f_tags=x`). Detect the legacy format positively (presence of `filter_\d+_column`) instead.
> * There are no tests for the RunPython functions — a migration test with a `~`-containing value would have caught the corruption bug above.

SmittieC: Ah! That's on me then, my apologies. (This reply was claude generated, but I clearly missed this detail. It's also interesting that it didn't attribute itself...). I don't agree with claude here. There's no future case I can see where we'd want to change this in a breaking way.

brian-mecha: @SmittieC Fixed and refactored.

apps/web/dynamic_filters/datastructures.py
snopoke:
```suggestion
        if not any(key.startswith("f_") for key in query_params):
```

apps/web/dynamic_filters/datastructures.py
```diff
@@ -18,14 +66,20 @@ class ColumnFilterData(BaseModel):
...
+        """Normalize list-based filter values to JSON arrays for operators that expect lists."""
         if self.operator in _LIST_OPERATORS:
-            try:
-                parsed = json.loads(self.value)
-                if not isinstance(parsed, list):
-                    self.value = json.dumps([self.value])
-            except (json.JSONDecodeError, TypeError):
-                self.value = json.dumps([self.value])
+            if isinstance(self.value, str):
+                parsed = _coerce_list_filter_values(self.value)
```
snopoke, line 76: is this for backwards compatiblity?

snopoke (commented):
@brian-mecha there are still some outstanding comments on this PR. It would be great to get this merged.

apps/web/dynamic_filters/datastructures.py
brian-mecha: @snopoke Yes it is.

snopoke (changes requested):
Running claude `/review` found some issues that need to be addressed:

🔴 Correctness — must fix

1. Stale legacy check silently breaks dataset session pre-population — apps/evaluations/views/dataset_views.py:203
has_explicit_filters = any(key.startswith("filter_") for key in self.request.GET)
This is not updated by the PR and still matches only the old prefix. New-format keys (f_tags, op_tags) do not start with filter_, so has_explicit_filters is now always False when arriving with new-format URL filters. Effect: navigating to "Add sessions to dataset" from a filtered session list will no longer pre-select the filtered sessions — a silent functional regression. Should be key.startswith("f_") (matching FilterParams.from_request). The PR's add-sessions test only covers the POST body path, so this GET path is uncovered — please add a test.

🟡 Design / risk

2. Hard cutover, not "backward compatible" as the PR describes. FilterParams.__init__ now reads only f_/op_; legacy URL params are ignored (and the form actively rejects them per test_form_rejects_legacy_filter_query). The data migrations handle stored values, but any bookmarked or externally-generated URLs using filter_0_* will silently produce an unfiltered result (no error — potentially showing more sessions than the user intends). The PR body ("maintains backward compatibility … supporting both legacy and new") and the docs contradict the actual behavior. Please either (a) confirm this silent drop is acceptable and fix the description, or (b) add a shim that redirects/translates legacy query params on read.

🟢 Minor / cleanup

- Dead code: getDateRangeFilterIndex (templates/filters/_filter_component_alpine.html:502) still uses the legacy filter_(\d+)_column regex. Its only caller (setLabelFromURL) is rewritten in this PR, leaving it orphaned — remove it.
- Doc rot / inconsistency: apps/web/dynamic_filters/base.py:131 docstring still references filter_{i}_column. In docs/developer_guides/code_systems/dynamic_filters.md, the "Legacy Format (No Longer Supported)" section shows new-format examples (copy-paste slip), and the "No Longer Supported" heading conflicts with the "supported during migration" wording elsewhere.
- Migrations import app code (convert_saved_filter_data → datastructures) at module load. It's tested and pragmatic, but migration behavior will drift if that helper changes later; Django convention favors inlining. Acceptable, just flagging.
- Broad f_ prefix: [k for k in query_params if k.startswith("f_")] treats any query param beginning with f_ as a filter column. Low risk today, but worth documenting f_/op_ as reserved prefixes to avoid future collisions.
- CSV robustness: _parse_csv_tilde_values trusts well-formed input. Values are produced by the escaping serializers so round-trips are safe, but a hand-crafted URL with an unbalanced " in a list value could parse oddly. Not blocking.

brian-mecha:
@snopoke Changes effected.

templates/filters/_filter_component_alpine.html
```diff
@@ -203,17 +203,56 @@
...
+            current = '';
+          } else {
+            current += char;
+          }
+        }
+        if (current) {
+          values.push(current.replace(/^"|"$/g, ''));
+        }
+        return values;
+      },
```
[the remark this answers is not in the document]
snopoke: @brian-mecha this seems relevant

assets/javascript/dashboard/main.js
```diff
@@ -6,6 +6,22 @@
 import TomSelect from "tom-select";
 import {formatDistanceToNow} from "date-fns";
 
+/**
+ * Serialize values into the tilde-separated CSV wire format used by the dynamic
+ * filters (f_* params), quoting values containing "~" or '"'. Mirrors the
+ * serializeCSVTildeValues helper in the Alpine filter component and the Python
+ * serialize_csv_tilde_values so producers and parsers agree on the format.
+ */
+function serializeCSVTildeValues(values) {
```
snopoke: can we not reuse the function defined in templates/filters/_filter_component_alpine.html or else extract it somewhere that both can use it.

brian-mecha: @snopoke Fixed

snopoke:
@brian-mecha you'll need to merge changes from https://github.com/dimagi/open-chat-studio/pull/3952

Also please review this review comment: https://github.com/dimagi/open-chat-studio/pull/3636/changes/ffe15464d1a24bb825d6bf031afe021eb7aa5fbd#r3415740166

brian-mecha:
@snopoke Changes from [3952](https://github.com/dimagi/open-chat-studio/pull/3952) have been merged in and referenced comment was fixed by this [commit](https://github.com/dimagi/open-chat-studio/commit/ffe15464d1a24bb825d6bf031afe021eb7aa5fbd#r3415740166) with the `assets/javascript/filters/csvTilde.js` file

templates/filters/_filter_component_alpine.html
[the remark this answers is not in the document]
brian-mecha: @snopoke this has been resolved by this [commit](https://github.com/dimagi/open-chat-studio/pull/3636/changes/ffe15464d1a24bb825d6bf031afe021eb7aa5fbd)

templates/chatbots/table_actions_dropdown.html
snopoke:
```suggestion
      {% endflag "flag_evaluations" %}
```

snopoke:
```suggestion
      {% endflag "flag_session-analysis" %}
```

snopoke (approved):

snopoke:
@brian-mecha I tried to fix the lint error but misunderstood the issue and I don't seem to be able to push changes to your fork branch so please drop this commit: aa1227d and then merge with main to get the actual fix (https://github.com/dimagi/open-chat-studio/pull/3968)

snopoke (approved):

brian-mecha:
@snopoke 
- Commit `aa1227d` has been dropped
- Branch merged with main

Issue #1106, opened 2025-01-30
Store related object FK's on the node itself instead of node params


SmittieC:
Currently node params are saved in a JSON field on the node. We want to save them on the node model instead an use foreign keys for relations like assistants and souce material, llm provider, llm model, collections etc.

SmittieC:
The PR to record FKs is [here](https://github.com/dimagi/open-chat-studio/pull/3645). Areas needing updating afterwards is [here](https://github.com/dimagi/open-chat-studio/blob/c942c2802bcd4281fc9af988aaf570eedcfc03cb/docs/superpowers/plans/2026-06-15-node-resource-fks.md#follow-up-work-not-in-this-pr).

SmittieC:
### The future of Node.params
- FK relationships to be source of truth
- remove node data from pipeline's data so that it isn't stored in Pipeline.data. It does need to be injected / constructed when rendering the pipeline
- don't store the node's resources in the params. Hydrate params with it from the FK when reading, and when writing, pluck it from params before saving the node

PR #3645, opened 2026-06-17
Mirror pipeline node resource references as FK columns

SmittieC:
### Product Description
No user-facing change.

### Technical Description
Related to #1106 

[The plan](https://github.com/dimagi/open-chat-studio/blob/c942c2802bcd4281fc9af988aaf570eedcfc03cb/docs/superpowers/plans/2026-06-15-node-resource-fks.md) (and list of followup changes to be made).

Resources referenced by a pipeline node (LLM provider, provider model, source material, collection + collection indexes, assistant, synthetic voice) were stored only as ids inside the node's `params` JSON — so there was no real FK, no referential integrity, and "which nodes use this resource?" required scanning JSON.

This adds nullable FK/M2M mirror columns on `Node` for those resources. `params` stays **authoritative**; the columns are a derived mirror, re-synced from `params` via `Node._sync_resource_fk_fields()` (called by `set_params()`, `update_from_params()`, and version copy). All on `SET_NULL`, so deleting a resource succeeds: the node's FK is nulled while `params` is left untouched.

Worth a closer look:
- **`_sync_resource_fk_fields()` in `apps/pipelines/models.py`** — the docstring explains why ids aren't pre-validated (a live node can't reference a deleted resource; versions point at possibly-deleted resources but are never re-synced) and why only scalar-FK changes trigger a save.
- **`set_params()` is the new write path.** Callers that previously did `node.params = ...; node.save()` now route through it (`clone_team`, `default_models`). `duplicate_shared_pipelines` was creating cloned nodes with `Node.objects.create(params=...)` and never syncing the mirror — fixed to create then `set_params()`.
- **`default_models._update_llm_provider_models`** now skips objects with no direct FK field to the model (pipeline nodes reference it via `params`, handled separately).
- The `as_int` id-coercion helper moved from `apps.api.v2.utils` to the shared `apps.utils.fields` so non-API code can use it (it also now rejects booleans).

The mirror is maintained but not yet a read source, so existing flows don't depend on it being populated.

### Migrations

`0026_node_resource_fks` only adds nullable columns + an M2M through table — additive, safe against the currently-running code.

**Backfill rollout plan** (populating the mirror for existing nodes):
1. **Manual first (this PR):** after this deploys, run the idempotent `backfill_node_fks` management command **manually** in prod so we can run it without blocking the migration step.
2. **Follow-up PR:** once the manual run is done, a follow-up PR will run the `backfill_node_fks` command as a **data migration**.

We deliberately run it manually first rather than shipping the data migration in this PR.

### Demo
N/A

### Docs and Changelog


apps/service_providers/tests/test_views.py
```diff
@@ -127,6 +128,31 @@ def test_sync_voices_endpoint(team_with_users, authed_client):
     mock_sync.assert_called_once()
 
 
+@pytest.mark.django_db()
+def test_delete_llm_provider_referenced_by_pipeline_nullifies_node_fk(team_with_users, authed_client):
```
SmittieC, line 132: This will only happen when we archived a pipeline and then remove the llm provider that was referenced by it. Add a comment to add this context

apps/service_providers/llm_service/default_models.py
```diff
@@ -306,7 +306,12 @@ def _update_llm_provider_models(LlmProviderModel):
             for custom_model in existing_custom_global[key]:
                 related_objects = get_related_objects(custom_model)
                 for obj in related_objects:
-                    field = [f for f in obj._meta.fields if f.related_model == LlmProviderModel][0]
+                    fields = [f for f in obj._meta.fields if f.related_model == LlmProviderModel]
+                    if not fields:
+                        # Pipelines surfaced via the Node.llm_provider_model reverse FK have no
+                        # direct field to repoint here; their nodes are handled via params below.
+                        continue
+                    field = fields[0]
```
SmittieC: Refactored in 74ee055c6 — extracted `_update_existing_global_model` and `_replace_custom_model_with_global` out of `_update_llm_provider_models`, bringing its cyclomatic complexity back down.

apps/pipelines/models.py
```diff
@@ -368,13 +415,27 @@ def create_new_version(self, is_copy=False, new_flow_id=None, pipeline=None):  #
...
+    def set_params(self, params):
+        """Assign params, persist them, and re-derive the resource FK mirror.
+
+        Prefer this over assigning ``self.params`` and calling ``save()`` directly: it keeps
+        the FK columns (a derived mirror of the ids in params) from drifting away from params.
+        See ``_sync_resource_fk_fields``.
+        """
+        self.params = params
+        self.save()
+        self._sync_resource_fk_fields()
```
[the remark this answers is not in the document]
SmittieC: Fixed in af452466f. `set_params` now persists only `params` (plus `updated_at`, to keep the auto_now bump) via `update_fields` when the row already exists, falling back to a full save on insert. `_sync_resource_fk_fields` still handles the derived FK columns.

apps/pipelines/models.py
```diff
@@ -383,6 +444,42 @@ def update_from_params(self):
...
+        for field_name in self.resource_fk_fields():
+            value = as_int(params.get(f"{field_name}_id"))
+            if getattr(self, f"{field_name}_id") != value:
+                setattr(self, f"{field_name}_id", value)
+                update_fields.append(f"{field_name}_id")
+        if update_fields:
+            self.save(update_fields=update_fields)
+
+        index_ids = params.get("collection_index_ids") or []
+        self.collection_indexes.set(Collection.objects.filter(id__in=index_ids))
```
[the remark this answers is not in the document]
SmittieC: Fixed in fe3f54249. The ids are normalized to a list and coerced through `as_int` (same as the scalar FKs), dropping malformed values before the `id__in` query.

apps/utils/fields.py
```diff
@@ -1,10 +1,23 @@
...
+    """Convert a value to an int, returning None if it can't be (e.g. a malformed id from JSON).
+
+    Booleans are rejected: they're never valid ids, and ``int(True)`` would otherwise coerce to 1.
+    """
+    if isinstance(value, bool):
+        return None
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return None
```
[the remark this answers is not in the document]
SmittieC: Fixed in 9098a0c27. `as_int` now only converts `int` and integer `str` inputs; floats (and everything else) return `None`, matching the documented contract. Added float cases to `test_fields.py`.

apps/service_providers/tests/test_views.py
SmittieC: Done in 44cf90ba3 — added a docstring note that the FK is only nulled for an archived pipeline, since the delete guards block removing a provider a live node still references.

apps/pipelines/models.py
```diff
@@ -427,7 +427,13 @@ def set_params(self, params):
         See ``_sync_resource_fk_fields``.
         """
         self.params = params
-        self.save()
+        if self.pk:
```
SmittieC: I think this should only do `self.save(update_fields=["params"])` instead of check pk etc. just save and sync

SmittieC: Done in 0d5671511. Confirmed all three callers (`duplicate_shared_pipelines`, `clone_team`, `default_models`) operate on already-persisted nodes, so the pk guard was dead code — simplified to `self.save(update_fields=["params"])` + sync. Heads up: this also drops the `updated_at` bump on a params-only write (the FK-sync save already behaved that way); shout if you'd rather keep it.

apps/service_providers/llm_service/default_models.py
```diff
@@ -277,44 +277,54 @@ def _update_llm_provider_models(LlmProviderModel):
...
+            # Pipelines surfaced via the Node.llm_provider_model reverse FK have no
+            # direct field to repoint here; their nodes are handled via params below.
+            continue
+        field = fields[0]
+        setattr(obj, field.attname, global_model.id)
+        obj.save(update_fields=[field.name])
+
+    related_pipeline_nodes = get_related_pipelines_queryset(custom_model, "llm_provider_model_id")
+    for node in related_pipeline_nodes.select_related("pipeline").all():
+        _update_pipeline_node_param(node.pipeline, node, "llm_provider_model_id", global_model.id)
```
snopoke, line 325: This doesn't seem quite right - if we're using FKs we should be able to use the normal 'related_objects' path and not a special path for pipelines.

I think this stems from the fact that we're still storing the IDs in the params and using those on the read path instead of re-injecting them from the FKs.

Edit: I found this in the plan doc: `Migrate all other params-scanning callsites (inspect API, deletion helpers, admin, etc.) to use FK lookups (see "Follow-up work" table).`

I think that will cover what I'm getting at here.

snopoke (approved):

PR #3647, opened 2026-06-17
security: emit email_verified OIDC claim

bderenzi:
## Why

OCS acts as an **OIDC identity provider** (django-oauth-toolkit): downstream products federate against OCS's `/o/userinfo/` and id_token. One such consumer is about to start **trusting the email addresses OCS asserts** to link/merge user accounts across providers (CommCare HQ, CommCare Connect, OCS). That trust is safest when the token itself signals whether the asserted email is verified.

Today the OIDC validator emits no `email_verified` claim: `apps/oauth/validator.py::get_additional_claims` sets `sub` (= the user's email), `name`, `is_active`, and `team`, but never `email_verified`. So a consumer has no protocol-level signal of whether the asserted email was verified — it has to *assume* OCS's config. (django-oauth-toolkit's base `oidc_claim_scope` already lists `email_verified`, but OCS's custom `get_additional_claims` never populated it.)

> **Note (per review):** an earlier revision of this PR also flipped the `ACCOUNT_EMAIL_VERIFICATION` default from `"optional"` to `"mandatory"` (`config/settings.py` + `app.json`). That has been **reverted** — deployed environments already override this to `mandatory`, and changing the default makes local dev harder. This PR is now scoped to the OIDC claim only.

## What changed

- `apps/oauth/validator.py`: `get_additional_claims` now emits `email_verified`, computed from allauth's `EmailAddress.verified` via the existing `apps.users.helpers.user_has_confirmed_email_address`. Added `email_verified` to `oidc_claim_scope` (scoped to `openid`).
- `apps/oauth/tests/test_validator.py` (new): a table-driven test covering `email_verified` = True (confirmed email), False (unverified `EmailAddress`), False (no `EmailAddress` record), plus that the claim is scoped to `openid`.

## Behavioral implications

- **Additive only**: consumers reading `/o/userinfo/` or the id_token now receive `email_verified` (boolean) alongside `sub`. The value reflects allauth's `EmailAddress.verified`. Consumers should treat `email_verified == false` (or a missing claim from older OCS versions) as "do not trust for account linking."
- No change to signup/login behavior; no data migration.
- The claim is gated behind the `openid` scope — no effect on existing API scopes (`chatbots:*`, `sessions:*`, etc.).

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -310,12 +310,9 @@ MFA_ADAPTER = "apps.users.adapter.MfaAdapter"
 MFA_RECOVERY_CODE_COUNT = 10
 MFA_TOTP_ISSUER = "Open Chat Studio"
 
-# User signup configuration. Defaults to "mandatory" so users must confirm ownership of their email
-# address before signing in. This keeps the email addresses OCS asserts (including via the OIDC
-# userinfo/id_token claims that downstream consumers federate against) trustworthy. Override with
-# "optional" (send confirmation emails but don't require them) or "none" (don't verify at all) only
-# for local development where email delivery isn't configured.
-ACCOUNT_EMAIL_VERIFICATION = env("ACCOUNT_EMAIL_VERIFICATION", default="mandatory")
+# User signup configuration: change to "mandatory" to require users to confirm email before signing in.
+# or "optional" to send confirmation emails but not require them
+ACCOUNT_EMAIL_VERIFICATION = env("ACCOUNT_EMAIL_VERIFICATION", default="optional")
 
 AUTHENTICATION_BACKENDS = (
     # check permissions exist (DEBUG only)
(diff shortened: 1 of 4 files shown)


config/settings.py
```diff
@@ -310,9 +310,12 @@
...
 
-# User signup configuration: change to "mandatory" to require users to confirm email before signing in.
-# or "optional" to send confirmation emails but not require them
-ACCOUNT_EMAIL_VERIFICATION = env("ACCOUNT_EMAIL_VERIFICATION", default="optional")
+# User signup configuration. Defaults to "mandatory" so users must confirm ownership of their email
+# address before signing in. This keeps the email addresses OCS asserts (including via the OIDC
+# userinfo/id_token claims that downstream consumers federate against) trustworthy. Override with
+# "optional" (send confirmation emails but don't require them) or "none" (don't verify at all) only
+# for local development where email delivery isn't configured.
+ACCOUNT_EMAIL_VERIFICATION = env("ACCOUNT_EMAIL_VERIFICATION", default="mandatory")
```
snopoke: This change should be reverted. This will make local dev much harder. Deployed environments already override this to make it mandatory.

bderenzi: Oh yeah, good point. That'd be super annoying. I will strip that out.

SmittieC (approved):

snopoke (approved):

Issue #3613, opened 2026-06-12
Test coverage: re-adding an archived node flow_id in Pipeline.update_nodes_from_data


snopoke:
## Parent

Part of #3398.

## What to build

`Pipeline.update_nodes_from_data` (apps/pipelines/models.py) archives removed nodes that have versions rather than deleting them. If flow data later reintroduces the same `flow_id` (which revert will do), `Node.objects.update_or_create` goes through `NodeObjectManager` and may collide with — or need to revive — the archived row. Write tests covering this path first; fix the behavior if broken.

## Acceptance criteria

- [ ] Test: node with versions is removed (archived) then the same flow_id is re-added via `update_nodes_from_data`
- [ ] Resulting working node is editable and correctly linked; no duplicate or constraint error
- [ ] Behavior fixed if the test exposes a bug

## Blocked by

None - can start immediately

Issue #3610, opened 2026-06-12
Declarative registry for versioned node params


snopoke:
## Parent

Part of #3398.

## What to build

`Node.create_new_version` hardcodes per-node-type handling of params that reference versioned models (`assistant_id` on AssistantNode, `source_material_id` on LLMResponseWithPrompt — see `apps/pipelines/models.py`). Extract a declarative registry mapping node type → {param name: model class} and drive the existing working→version param rewriting from it. No behavior change.

This becomes the single source of truth that the revert feature (#3398) will invert (version id → working id), so future versioned params can't silently break revert.

## Acceptance criteria

- [ ] Versioned-param handling in `Node.create_new_version` is driven by a declarative registry, not per-type conditionals
- [ ] Existing publish/version-creation behavior is unchanged (covered by existing + new tests)
- [ ] Registry is documented so new versioned params get added there

## Blocked by

None - can start immediately

PR #3650, opened 2026-06-18
Revert working chatbot to a previous version (core flow)

### LINKED ISSUE #3615: Revert working version to a previous version: core flow (fields + pipeline)
## Parent

Part of #3398.

## What to build

Tracer bullet for revert. From the versions table, a "Revert to this version" button (native confirm dialog for now) posts to a new team-scoped view which calls `Experiment.revert_to_version(version)`:

- copies versioned content fields from the version row onto the working row (`consent_form` remapped to its working version)
- resets the working pipeline: `data` from the version pipeline with node params remapped from versioned refs back to working refs (inverse of the registry from #3610), then `update_nodes_from_data()`
- non-destructive: no version rows are modified; `is_default_version` untouched

Triggers, the diff modal, and async/snapshot come in follow-up slices.

## Acceptance criteria

- [ ] Revert button on each version row reverts the working chatbot to that version's fields and pipeline
- [ ] Round-trip test: publish v1 → modify → revert to v1 → `version_details` compare shows no changes
- [ ] Node params referencing versioned assistants/source material point at working versions after revert
- [ ] Version history and default version are unchanged by the operation
- [ ] View is POST-only, permission-checked, team-scoped; changes audit-logged

## Blocked by

- #3610
- #3611
- #3613

snopoke:
Closes #3615 — tracer bullet for the revert feature (#3398). A "Revert to this version" button on each version row restores the working chatbot to that version's content, non-destructively. Triggers, the diff modal, and async/snapshot handling are deliberately left for follow-up slices.

### Technical Description
Revert is the inverse of publish, built on the groundwork from #3610/#3611/#3613:

- `VersionedParamSpec.revert_referenced_record` inverts the publish-time param rewriting (versioned id → working id) using the same declarative registry, so new versioned params stay covered.
- `Experiment.revert_to_version` copies the `VERSIONED_CONTENT_FIELDS` classification onto the working row (remapping `consent_form` to its working version), then resets the pipeline. `is_default_version` and version rows are never touched.
- `Pipeline.revert_to_version` sources params from `version.flow_data` (the node rows), **not** the stored `data` JSON. The builder saves `data` *before* publish, so it holds working-era refs; only the node rows carry the versioned refs this revert needs to invert. After remapping it calls `update_nodes_from_data`, which revives archived node `flow_id`s (the path #3613 hardened).

Worth a reviewer's attention: the `flow_data`-vs-`data` distinction above is the subtle correctness point. The view is synchronous for now and guards against running while another version operation holds the lock.

### Migrations
N/A — no schema changes.

### Docs and Changelog


apps/chatbots/tests/test_chatbot_views.py
```diff
@@ -781,6 +781,55 @@ def test_web_chat_widget_rendering(client, team_with_users):
...
+    response = client.post(url)
+
+    assert response.status_code == 302
+    assert response.url.endswith("#versions")
+    experiment.refresh_from_db()
+    assert experiment.name == "Original"
+
+
+@pytest.mark.django_db()
+def test_revert_chatbot_version_is_post_only(client, team_with_users):
```
snopoke: remove this test

apps/chatbots/tests/test_chatbot_views.py
```diff
@@ -781,6 +781,55 @@ def test_web_chat_widget_rendering(client, team_with_users):
...
+    client.force_login(user)
+    experiment = ExperimentFactory.create(team=team, owner=user)
+    version = experiment.create_new_version(make_default=True)
+
+    url = reverse("chatbots:revert-version", args=[team.slug, experiment.id, version.version_number])
+    assert client.get(url).status_code == 405
+
+
+@pytest.mark.django_db()
+def test_revert_chatbot_version_requires_permission(client, team_with_users):
```
snopoke: remove this test

apps/chatbots/views.py
```diff
@@ -474,6 +474,30 @@ def get_success_url(self):
...
+        raise Http404() from None
+
+    if experiment.is_archived:
+        raise PermissionDenied("Unable to revert an archived chatbot.")
+
+    if experiment.version_operation_in_progress:
+        messages.error(request, "A version operation is already in progress.")
+    else:
+        experiment.revert_to_version(version)
+        messages.success(request, f"Reverted to v{version_number}.")
```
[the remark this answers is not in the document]
snopoke: Fixed in 9d235ff. The view now acquires the version-operation lock via `acquire_version_operation_lock` (atomic claim) and reverts inside a try/finally that releases it, so there's no check-then-act gap.

*— Claude Code*

apps/experiments/models.py
```diff
@@ -959,6 +959,32 @@ def create_new_version(  # ty: ignore[invalid-method-override]
...
+        # pipeline is reset in place rather than re-pointed, and the consent form is remapped to its
+        # working version rather than the version's frozen copy.
+        for field in self.VERSIONED_CONTENT_FIELDS - {"pipeline", "consent_form"}:
+            setattr(self, field, getattr(version, field))
+
+        self.consent_form = version.consent_form.get_working_version() if version.consent_form else None
+        self.save()
+
+        if version.pipeline:
+            self.pipeline.revert_to_version(version.pipeline)
```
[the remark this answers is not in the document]
snopoke: Fixed in 9d235ff. `revert_to_version` now raises `ValueError` if `version.pipeline` exists but the working experiment has no pipeline, instead of hitting an `AttributeError`.

*— Claude Code*

apps/chatbots/tests/test_chatbot_views.py
snopoke: Fixed in 9d235ff — removed.

*— Claude Code*

apps/chatbots/tests/test_chatbot_views.py
snopoke: Fixed in 9d235ff — removed.

*— Claude Code*

apps/chatbots/tests/test_chatbot_views.py
```diff
@@ -781,6 +781,55 @@ def test_web_chat_widget_rendering(client, team_with_users):
     assert "end-experiment-modal" in content
 
 
+@pytest.mark.django_db()
```
snopoke: Addressed in 9d235ff — removed the two framework-only view tests (post-only/permission) at the author's request, which trims the module.

*— Claude Code*

apps/chatbots/tests/test_chatbot_views.py
```diff
@@ -781,6 +781,55 @@ def test_web_chat_widget_rendering(client, team_with_users):
...
+    version = experiment.create_new_version(make_default=True)
+
+    # A team member without the change_experiment permission is forbidden.
+    no_perm_user = UserFactory.create()
+    MembershipFactory.create(team=team, user=no_perm_user)
+    client.force_login(no_perm_user)
+
+    url = reverse("chatbots:revert-version", args=[team.slug, experiment.id, version.version_number])
+    assert client.post(url).status_code == 403
+
```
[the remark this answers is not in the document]
snopoke: Noted — declined by author. The post-only/permission view tests are being removed (framework behavior) and CodeScene flags this module as too long, so we're not adding more view-guard tests here. The in-progress path is now exercised by the lock acquisition logic.

*— Claude Code*

templates/experiments/components/experiment_version_actions.html
```diff
@@ -23,6 +23,15 @@
             onclick="clearModalAndOpen();">
       View Details
     </button>
+    <form method="post" action="{% url 'chatbots:revert-version' team_slug=team.slug experiment_id=record.working_version_id version_number=record.version_number %}" class="inline"
+          onsubmit="return confirm('Revert the working chatbot to v{{ record.version_number }}? This overwrites the current unpublished changes.');">
```
SmittieC: Nit: using `alertify.confirm` is more in line with how other places does confirmation.

apps/pipelines/models.py
```diff
@@ -247,6 +248,36 @@ def create_new_version(self, is_copy: bool = False):  # ty: ignore[invalid-metho
...
+        node rows hold the versioned references this revert needs to invert. The versioned record
+        for each param is read from the version node's resource FK column.
+        """
+        version_nodes_by_flow_id = {node.flow_id: node for node in version.node_set.all()}
+        data = copy.deepcopy(version.flow_data)
+        for node in data.get("nodes", []):
+            node_data = node.get("data", {})
+            params = node_data.get("params", {})
+            version_node = version_nodes_by_flow_id.get(node.get("id"))
+            if version_node is None:
```
SmittieC: would this ever be true, since version_node comes from version_nodes_by_flow_id, which comes from `version.node_set`, but `node` also comes from there, just through another path i.e `version.flow_data`

apps/pipelines/models.py
```diff
@@ -247,6 +248,36 @@ def create_new_version(self, is_copy: bool = False):  # ty: ignore[invalid-metho
 
         return pipeline_version
 
+    @transaction.atomic()
+    def revert_to_version(self, version: "Pipeline") -> None:
+        """Reset this working pipeline to the state of ``version``.
+
+        The version's flow data (node params sourced from its node rows, which hold the versioned
```
SmittieC: Nit: I found it unnecessarily confusing to read this docstring. I think the reason is that didn't expect the code to use  `pipeline.data` in the first place, and it is making a case for not using it, but in a way that sounds like it is must be counter-intuitive.

SmittieC (dismissed):
LGTM overall.

apps/pipelines/models.py
snopoke: Good catch — it can't. `flow_data` builds its nodes from the same `node_set`, so every flow_id resolves. Removed the dead guard in 4a49125 and index `version_nodes_by_flow_id` directly so a genuine mismatch would surface as a KeyError.

*— Claude Code*

apps/pipelines/models.py
snopoke: Reworded in 4a49125 to state plainly that it reads node-row params via `flow_data`, dropping the counter-intuitive framing.

*— Claude Code*

templates/experiments/components/experiment_version_actions.html
snopoke: Done in 4a49125 — switched to `alertify.confirm` to match the project convention.

*— Claude Code*

PR #3673, opened 2026-06-22
Exclude experiment's own pipeline from PIPELINE_START trigger choices

### LINKED ISSUE #3669: PIPELINE_START trigger can reference the experiment's own pipeline
## Summary

A "Start a pipeline" trigger action lets the user pick any working pipeline in the team. `PipelineStartForm` (`apps/events/forms.py`) builds its choices from:

```python
self.fields["pipeline_id"].queryset = Pipeline.objects.filter(team_id=team_id, working_version_id__isnull=True)
```

This includes the experiment's **own** pipeline. A trigger on a chatbot that starts that same chatbot's pipeline is not a meaningful configuration (re-entrant/self-referential), and it should not be selectable.

## Expected

The pipeline choices for a `PIPELINE_START` trigger exclude the experiment's own pipeline.

## Notes

- Surfaced while working on #3616 / PR #3668: trigger publish/revert had to special-case the self-reference (versioning the experiment's pipeline before syncing triggers so the trigger pins to the version just created rather than spawning a redundant one). Excluding self-references at the form level removes that edge case.
- Existing data may already contain self-referential triggers; a fix should decide whether to leave them or clean them up.

snopoke:
### Product Description
A "Start a pipeline" trigger could previously reference the experiment's own pipeline, which is a re-entrant/self-referential configuration and not meaningful. This PR excludes the experiment's own pipeline from the `PIPELINE_START` trigger choices.

Fixes #3669

### Technical Description
- `PipelineStartForm` now accepts an optional `experiment_id` and excludes the experiment's own pipeline from its queryset.
- `build_action_params_form` passes `experiment_id` through to `PipelineStartForm`.
- Added `apps/events/tests/test_forms.py`.

Existing self-referential triggers in the DB are left untouched (form-level fix only).

### Migrations

### Docs and Changelog

### DIFF
--- a/apps/events/forms.py
+++ b/apps/events/forms.py
@@ -36,8 +36,16 @@ class PipelineStartForm(forms.Form):
 
     def __init__(self, *args, **kwargs):
         team_id = kwargs.pop("team_id")
+        experiment_id = kwargs.pop("experiment_id", None)
         super().__init__(*args, **kwargs)
-        self.fields["pipeline_id"].queryset = Pipeline.objects.filter(team_id=team_id, working_version_id__isnull=True)
+        queryset = Pipeline.objects.filter(team_id=team_id, working_version_id__isnull=True)
+        if experiment_id is not None:
+            # Exclude the experiment's own pipeline: a trigger that starts the chatbot's own
+            # pipeline is re-entrant/self-referential and not a meaningful configuration.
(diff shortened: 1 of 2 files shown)


snopoke (commented):
@claude I think we actually want to exclude any pipeline that is already directly linked to an experiment.

PR #3675, opened 2026-06-22
Team data export and sync

SmittieC:
### Product Description
A team admin can register an RSA public key for their team, and an operator can replicate a team's data (chatbots, pipelines, providers, participants, sessions, …) from one OCS server to another with the new `sync_team` management command. Secrets — provider configs, participant data — are encrypted to the team's public key in transit and only decrypted on the target with the matching private key.

### Technical Description
**Please note**: Not all models are yet exported. This PR serves as a POC. A followup PR will add the rest of the models to the manifest.

Two halves.

**Source — read API (team-admin only):**
- `GET /api/v2/manifest/` declares the models to pull in dependency order plus each model's per-model config, and doubles as the allowlist: the resource endpoint refuses anything not listed in it.
- `GET /api/v2/<resource>/` is generic, team-scoped and keyset-paginated. Per-model serializers are built at runtime so they can't drift from the model.
- Secret fields use envelope encryption (a random Fernet key, wrapped by the team's RSA public key) so values larger than the RSA block (e.g. a service-account JSON) still seal. The endpoint fails closed with 409 if the team has no public key.

**Target — `sync_team` command + import engine:**
- A SQLite-backed FK translation table maps source→target ids and doubles as a resume checkpoint (a null target means "created but not yet filled"), so a run resumes where it left off. Per-model cursors are derived from the rows already synced rather than stored separately.
- The importer rewrites every reference from source ids to target ids — including ids embedded in pipeline/node params — restores source timestamps (bypassing `auto_now`), matches shared/global rows (e.g. LLM provider models) by natural key, maps existing users by username without overwriting them, and emails new users a password reset (passwords are never migrated).
- The manifest carries a checksum of the applied migration set; the sync aborts on a mismatch unless `--skip-schema-check` is passed.

The widget-JOIN-guard commit is a prerequisite: the new `Team.public_key` migration would otherwise break the widget-deprecation data migration on a fresh DB.

Worth a closer look:
- the catch-all `re_path(r"^(?P<resource>[a-z_]+)/$")` sits after the router include — any future plain-lowercase v2 route would need to precede it.
- access control: both endpoints are `IsAuthenticated + IsTeamAdmin`, and secrets only ever leave the source sealed.

### Migrations

`0011_team_public_key` adds a nullable `Team.public_key` `TextField` with a default — additive only.

### Demo
https://www.loom.com/share/5dc056c2473342e6962bdc2002933b8f

Set a public key on the source team (Team → Data Export), then on the target:

\`\`\`
manage.py sync_team --source-url=<src> --api-key=<key> --team-slug=<slug> --private-key-path=<path>
\`\`\`

Rerun to resume / pick up new data.

### Docs and Changelog

New operator command (`sync_team`) and a new team-admin "Data Export" setting (register a public key).


apps/api/general/views.py
```diff
@@ -0,0 +1,21 @@
...
+config the sync needs. It also serves as the allowlist of models the sync is permitted to read."""
+
+from rest_framework.permissions import BasePermission, IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from apps.teams.sync.manifest import build_manifest
+
+
+class IsTeamAdmin(BasePermission):
```
SmittieC: move this to `apps/api/permissions.py`

apps/teams/export/tests/test_manifest.py
```diff
@@ -0,0 +1,97 @@
+import pytest
```
SmittieC, line 1: we also need a test to make sure that between the manifest and a newly added (currently missing) list, all possible models are covered. The aim of that test is to ensure that when we add new models, that we either add it to the manifest, or explicitly ignore it.

apps/api/general/views.py
```diff
@@ -19,3 +33,55 @@ class ManifestView(APIView):
 
     def get(self, request):
         return Response(build_manifest())
+
+
+class ResourceSlugView(APIView):
```
SmittieC: rather call it "ResourceView"

apps/api/general/views.py
```diff
@@ -19,3 +33,55 @@ class ManifestView(APIView):
...
+    permission_classes = [IsAuthenticated, IsTeamAdmin]
+
+    def get(self, request, resource):
+        entry = get_entry(resource)
```
SmittieC: rename get_entry to get_manifest_entry

apps/api/general/views.py
```diff
@@ -19,3 +33,55 @@ class ManifestView(APIView):
...
+        if entry is None:
+            raise NotFound("Unknown content type.")
```
SmittieC: rather do `raise NotFound("Unknown resource.")`

apps/api/general/serializers.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return list(Flag.objects.filter(teams=team).values_list("name", flat=True))
+
+
+def _group_names(self, instance):
+    return list(instance.groups.values_list("name", flat=True))
+
+
+# Per-model SerializerMethodFields for values that aren't a plain field dump.
+_METHOD_FIELDS: dict[str, dict] = {
```
SmittieC: rename: _METHOD_FIELDS -> _FIELD_RESOLVERS or something

apps/api/general/views.py
```diff
@@ -19,3 +33,55 @@ class ManifestView(APIView):
...
+            queryset = queryset.filter(id__gt=int(cursor))
+        rows = list(queryset[: limit + 1])
+        has_more = len(rows) > limit
+        rows = rows[:limit]
+        next_cursor = str(rows[-1].id) if rows else cursor
+        return rows, next_cursor, has_more
+
+    queryset = queryset.order_by("updated_at", "id")
+    if cursor:
+        keyset = json.loads(base64.b64decode(cursor))
```
SmittieC: might be good to have the cursor base64 encode and decode operation behind a mini function or wrapper class. CursorHelper or something. Just so it reads easier

apps/teams/sync/tests/test_seal.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def test_seal_unseal_round_trips_a_dict(keypair):
+    public_pem, private_pem = keypair
+    value = {"api_key": "sk-secret", "nested": {"a": 1}}
+    token = seal_mod.seal(value, seal_mod.load_public_key(public_pem))
+    assert seal_mod.unseal(token, seal_mod.load_private_key(private_pem)) == value
+
+
+def test_seal_unseal_round_trips_a_string(keypair):
```
SmittieC: this test and the one above should be merged and use pytest parameterization

apps/teams/management/commands/sync_team.py
```diff
@@ -0,0 +1,102 @@
...
+        cursor = _start_cursor(model_label, cursor_type, store, model)
+        importer.import_rows(model_label, client.iter_rows(resource, start_cursor=cursor, limit=page_limit))
+        write(f"synced {resource}")
+    return importer
+
+
+class Command(BaseCommand):
+    help = "Sync a team's data from a source OCS server into this one."
+
+    def add_arguments(self, parser):
```
SmittieC, line 98: it would be nice to have an option to force delete the team before running the rest of the command.

apps/teams/export/tests/test_importer.py
```diff
@@ -0,0 +1,237 @@
...
+def test_rerun_does_not_duplicate(store):
+    importer = Importer(store)
+    importer.import_rows("teams.team", [_team_row()])
+    first_pk = store.get_target("teams.team", 9001)
+    importer.import_rows("teams.team", [_team_row()])
+    assert store.get_target("teams.team", 9001) == first_pk
+    assert Team.objects.filter(slug="imported-team-xyz").count() == 1
+
+
+def test_global_row_matches_existing_and_is_not_recreated(store):
```
SmittieC, line 110: add small docstrings to these tests

apps/api/general/serializers.py
```diff
@@ -0,0 +1,54 @@
...
+    "teams.team": {"feature_flags": _feature_flags},
+    "teams.membership": {"groups": _group_names},
```
snopoke: aren't groups systems defined ie. not user editable and shouldn't be synced

api-schemas/v1.yml
```diff
@@ -1045,6 +1045,25 @@ paths:
                       channel
                   summary: Experiment cannot send messages on the specified channel
           description: ''
+  /api/v2/{resource}/:
```
snopoke: I don't love that these are dynamic so you can never see what the available resources or schema's are. Would it be possible to expose the correct set of endpoints and schema? DRF Spectacular does have mechanisms that I think we could make use of to build the endpoints / schema dynamically: https://drf-spectacular.readthedocs.io/en/latest/customization.html

apps/api/general/views.py
SmittieC: Fixed in 0fab71ef2 — moved `IsTeamAdmin` into `apps/api/permissions.py` and imported it back into the view.

apps/teams/export/tests/test_manifest.py
SmittieC: Fixed in 244edee32 — added an explicit `IGNORED_MODELS` set to `manifest.py` plus `test_every_first_party_model_is_synced_or_ignored`, which fails if any first-party model is neither in `MANIFEST_ENTRIES` nor in the ignore list. Also added a disjointness check and folded `IGNORED_MODELS` into the existing resolve-to-models test so stale entries are caught.

Note: the ignore list currently holds 69 models that aren't synced yet — worth a glance to confirm none of them actually belong in the manifest.

apps/api/general/views.py
SmittieC: Fixed in d4fe2386e — renamed to `ResourceView` and updated the v2 urlconf reference.

apps/api/general/views.py
SmittieC: Fixed in 0240b810b — renamed `get_entry` to `get_manifest_entry` across the manifest, the view, and the tests.

apps/api/general/views.py
SmittieC: Fixed in 44832e1dc — now `raise NotFound("Unknown resource.")`.

apps/api/general/serializers.py
SmittieC: Fixed in e12520aa6 — renamed `_METHOD_FIELDS` to `_FIELD_RESOLVERS`.

apps/api/general/views.py
SmittieC: Fixed in 4d8aa7811 — added a `CursorHelper` with `encode`/`decode` static methods and routed the keyset pagination through it.

apps/teams/sync/tests/test_seal.py
SmittieC: Fixed in 976bc7a8e — merged the dict and string cases into a single parametrized `test_seal_unseal_round_trips`. I left the larger-than-RSA-block case as its own named test since it documents a distinct behaviour.

apps/teams/management/commands/sync_team.py
SmittieC: Fixed in 4e7c53e9a — added a `--force-delete` flag that drops the local team (matched by `--team-slug`) and its sync-state DB before syncing. Resetting the state is required, otherwise the derived resume cursor would skip re-importing the rows that were just deleted.

apps/teams/export/tests/test_importer.py
SmittieC: Fixed in 731b2d9c1 — added a one-line docstring to each test in the file.

apps/api/general/serializers.py
SmittieC: They are, but we still need to include the group names that users belong to in the payload of `membership`. This is what this mapping is for. Perhaps [this change](https://github.com/dimagi/open-chat-studio/pull/3675/commits/e12520aa643e8cf7c4b4697ac361eb26b46afa9c) makes that clearer.

apps/teams/export/importer.py
```diff
@@ -0,0 +1,252 @@
...
+        feature flags and auth groups -- aren't synced, so they're matched against existing target
+        rows by name rather than by translated id. Names with no match are skipped."""
+        if model_label == "teams.team":
+            from apps.teams.models import Flag  # noqa: PLC0415
+
+            for name in row.get("feature_flags", []):
+                flag = Flag.objects.filter(name=name).first()
+                if flag:
+                    flag.teams.add(instance)
+        elif model_label == "teams.membership":
```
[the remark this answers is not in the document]
SmittieC: this is fine.

apps/teams/management/commands/sync_team.py
SmittieC: Follow-up in b1e8b6a5b — switched the deletion to `delete_object_with_auditing_of_related_objects` inside the `current_team` context (the same audited cascade the team-delete view / `delete_team_async` use), rather than a bare queryset delete. Notifications are intentionally suppressed: a re-import shouldn't email anyone that their team was deleted. A test asserts the outbox stays empty.

apps/teams/export/importer.py
```diff
@@ -0,0 +1,252 @@
...
+
+        for name, target_pks in m2m_values.items():
+            getattr(instance, name).set([pk for pk in target_pks if pk is not None])
+
+        self._apply_named_links(model_label, instance, row)
+
+        if timestamps:  # keep the source timestamps; auto_now would otherwise overwrite them
+            _bypass_auto_now_update(model, instance.pk, timestamps)
+
+        if created and model_label == "users.customuser" and self.on_user_created:
```
SmittieC: Advisory — the `target_pk is not None and ...exists()` guard is intentional: a translation-map hit must still exist on the target before we reuse it. Left as-is.

*— Claude Code*

apps/api/general/serializers.py
SmittieC: Good question. Groups aren't synced or created here — the importer links each member to *existing* target groups by name (`Group.objects.filter(name__in=...)`), so a member keeps the same role where that role already exists on the target, and unknown names are skipped. That preserves roles on migration without touching the system-defined group definitions. Keeping it as-is for now — happy to drop the group link entirely if you'd rather members land role-less.

*— Claude Code*

api-schemas/v1.yml
```diff
@@ -1045,6 +1045,25 @@ paths:
...
+        schema:
+          type: string
+          pattern: ^[a-z_]+$
+        required: true
+      tags:
+      - api
+      security:
+      - {}
+      - apiKeyAuth: []
+      - tokenAuth: []
```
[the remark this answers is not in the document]
SmittieC: `api-schemas/v1.yml` is generated output and this diff is outdated. The endpoints are `IsAuthenticated + IsTeamAdmin`, and the regenerated schema reflects that (the update-generated-files check is green).

*— Claude Code*

api-schemas/v1.yml
```diff
@@ -1045,6 +1045,25 @@ paths:
...
+      responses:
+        '200':
+          description: No response body
```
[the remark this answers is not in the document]
SmittieC: Outdated — `ManifestView`/`ResourceView` now carry `@extend_schema` with explicit response serializers (`ManifestSerializer`, `resource_responses`), so the regenerated schema documents the response bodies.

*— Claude Code*

apps/api/export/views.py
```diff
@@ -0,0 +1,93 @@
...
+        timestamp = parse_datetime(keyset["updated_at"])
+        queryset = queryset.filter(Q(updated_at__gt=timestamp) | Q(updated_at=timestamp, id__gt=keyset["id"]))
+    rows = list(queryset[: limit + 1])
+    has_more = len(rows) > limit
+    rows = rows[:limit]
+    if rows:
+        next_cursor = CursorHelper.encode({"updated_at": rows[-1].updated_at.isoformat(), "id": rows[-1].id})
+    else:
+        next_cursor = cursor
+    return rows, next_cursor, has_more
```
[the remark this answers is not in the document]
SmittieC: Fixed in 4250882be — `_paginate` now raises a 400 (`ValidationError`) when the cursor's `updated_at` can't be parsed, instead of silently filtering on `None`.

*— Claude Code*

apps/teams/export/client.py
```diff
@@ -0,0 +1,59 @@
...
+                response = self._session.get(url, params=params, headers=headers, timeout=self.timeout)
+            except (requests.ConnectionError, requests.Timeout):
+                if last_attempt:
+                    raise
+                self._sleep(self.backoff * (attempt + 1))
+                continue
+
+            if response.status_code >= 500 and not last_attempt:
+                self._sleep(self.backoff * (attempt + 1))
+                continue
```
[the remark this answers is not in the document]
SmittieC: Fixed in 4250882be — the response is now closed before backing off on a 5xx retry, returning the socket to the pool. Covered by `test_transient_5xx_is_retried_then_succeeds`.

*— Claude Code*

apps/teams/export/importer.py
```diff
@@ -0,0 +1,252 @@
...
+    doesn't copy have no translation and are left as-is."""
+    result = dict(params)
+    for key, label in _node_param_models().items():
+        if key not in result or result[key] in (None, "", 0):
+            continue
+        value = result[key]
+        if isinstance(value, list):
+            result[key] = [store.get_target(label, as_int(v)) or v for v in value]
+        else:
+            result[key] = store.get_target(label, as_int(value)) or value
```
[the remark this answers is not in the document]
SmittieC: Declined. The manifest is dependency-ordered, so every in-scope resource a node references is synced (and translated) before nodes are imported — a missing translation there would be an ordering bug, not a value to silently keep. Out-of-scope refs are intentionally left untouched (covered by `test_remap_node_params_leaves_out_of_scope_refs_untouched`). No change.

*— Claude Code*

apps/teams/sync/tests/test_translation.py
```diff
@@ -0,0 +1,73 @@
...
+    store = FKTranslationStore(tmp_path / "team.sqlite")
+    store.record("chat.chat", 1, 10)
+    store.record("chat.chat", 2)  # uncommitted
+    assert store.committed_targets("chat.chat") == {1: 10}
+    assert store.committed_targets("missing") == {}
+
+
+def test_derive_pk_cursor_is_the_max_committed_key():
+    assert derive_pk_cursor([3, 1, 2]) == "3"
+    assert derive_pk_cursor([]) is None
```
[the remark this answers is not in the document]
SmittieC: Done in 79a3c9a50 — converted to `pytest.mark.parametrize` with readable ids (`returns-max-key`, `empty-returns-none`).

*— Claude Code*

api-schemas/v1.yml
SmittieC: Partly addressed — the URLConf now mounts one documented endpoint per resource via `resource_view` + `@extend_schema` (see `apps/api/general/schema.py`), so the resource set and per-resource response schemas are explicit in the generated spec rather than fully dynamic. Driving the whole thing from drf-spectacular machinery would be a nice follow-up.

*— Claude Code*

apps/teams/management/commands/sync_team.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+
+    importer = Importer(store, private_key=private_key, on_user_created=on_user_created)
+    for entry in manifest["entries"]:
+        model_label, resource, cursor_type = entry["model"], entry["resource"], entry["cursor"]
+        model = entry_model(model_label)
+        cursor = _start_cursor(model_label, cursor_type, store, model)
+        importer.import_rows(model_label, client.iter_rows(resource, start_cursor=cursor, limit=page_limit))
+        write(f"synced {resource}")
+    return importer
```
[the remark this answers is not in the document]
SmittieC: Root cause fixed in e6cd2352a: `.codescene/code-health-rules.json` was invalid JSON, so the tuned `function_max_arguments=6` was ignored and it fell back to the default of 4. `run_sync` still has 7 (one over): they're the command-wiring and test-injection seams (`write`, `enforce_schema`, `on_user_created`), kept explicit for readability. Advisory.

*— Claude Code*

apps/teams/export/client.py
```diff
@@ -0,0 +1,60 @@
...
+
+class SourceClient:
+    def __init__(self, base_url, api_key, *, timeout=30, max_retries=5, backoff=1.0, session=None, sleep=time.sleep):
+        self.base_url = base_url.rstrip("/")
+        self.api_key = api_key
+        self.timeout = timeout
+        self.max_retries = max_retries
+        self.backoff = backoff
+        self._session = session or requests.Session()
+        self._sleep = sleep
```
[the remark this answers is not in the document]
SmittieC: Same root cause as the other arg-count note — the CodeScene config was invalid JSON so the tuned threshold of 6 was ignored (fixed in e6cd2352a). `SourceClient.__init__` keeps 7: the retry knobs (`timeout`/`max_retries`/`backoff`) plus the `session`/`sleep` test seams. Kept explicit; advisory.

*— Claude Code*

apps/teams/export/importer.py
```diff
@@ -0,0 +1,262 @@
+"""The import engine: pull a serialized row, rewrite every reference from source ids to target ids,
```
SmittieC, line 1: Reduced in 7f68a5498 — extracting `_collect_concrete_field`, `_remap_embedded_resource_ids`, and `_build_m2m_values` from `_build_values` lowers the module's mean complexity (and cleared the Complex Method + Bumpy Road findings). Any residual is a small overage of the default mean-complexity threshold, which the custom config doesn't override. Advisory.

*— Claude Code*

apps/teams/export/importer.py
SmittieC: Advisory — the `target_pk is not None and ...exists()` guard is intentional: a translation-map hit must still exist on the target before we reuse it. Left as-is.

*— Claude Code*

apps/api/general/serializers.py
SmittieC: Good question. Groups aren't synced or created here — the importer links each member to *existing* target groups by name (`Group.objects.filter(name__in=...)`), so a member keeps the same role where that role already exists on the target, and unknown names are skipped. That preserves roles on migration without touching the system-defined group definitions. Keeping it as-is for now — happy to drop the group link entirely if you'd rather members land role-less.

*— Claude Code*

api-schemas/v1.yml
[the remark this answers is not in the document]
SmittieC: `api-schemas/v1.yml` is generated output and this diff is outdated. The endpoints are `IsAuthenticated + IsTeamAdmin`, and the regenerated schema reflects that (the update-generated-files check is green).

*— Claude Code*

api-schemas/v1.yml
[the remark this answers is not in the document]
SmittieC: Outdated — `ManifestView`/`ResourceView` now carry `@extend_schema` with explicit response serializers (`ManifestSerializer`, `resource_responses`), so the regenerated schema documents the response bodies.

*— Claude Code*

apps/api/export/views.py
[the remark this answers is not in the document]
SmittieC: Fixed in 4250882be — `_paginate` now raises a 400 (`ValidationError`) when the cursor's `updated_at` can't be parsed, instead of silently filtering on `None`.

*— Claude Code*

apps/teams/export/client.py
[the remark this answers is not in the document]
SmittieC: Fixed in 4250882be — the response is now closed before backing off on a 5xx retry, returning the socket to the pool. Covered by `test_transient_5xx_is_retried_then_succeeds`.

*— Claude Code*

apps/teams/export/importer.py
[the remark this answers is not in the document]
SmittieC: Declined. The manifest is dependency-ordered, so every in-scope resource a node references is synced (and translated) before nodes are imported — a missing translation there would be an ordering bug, not a value to silently keep. Out-of-scope refs are intentionally left untouched (covered by `test_remap_node_params_leaves_out_of_scope_refs_untouched`). No change.

*— Claude Code*

apps/teams/sync/tests/test_translation.py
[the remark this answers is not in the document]
SmittieC: Done in 79a3c9a50 — converted to `pytest.mark.parametrize` with readable ids (`returns-max-key`, `empty-returns-none`).

*— Claude Code*

apps/api/export/views.py
```diff
@@ -0,0 +1,93 @@
...
+                    status=status.HTTP_409_CONFLICT,
+                )
+            context["public_key"] = load_public_key(request.team.public_key)
+
+        limit = min(int(request.query_params.get("limit", DEFAULT_LIMIT)), MAX_LIMIT)
+        queryset = team_scoped_queryset(entry, request.team)
+        rows, next_cursor, has_more = _paginate(queryset, entry.cursor, request.query_params.get("cursor"), limit)
+
+        serializer = build_sync_serializer(entry_model(entry.model))(rows, many=True, context=context)
+        return Response({"cursor": next_cursor, "has_more": has_more, "results": serializer.data})
```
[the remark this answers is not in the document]
SmittieC: Already addressed — `ManifestView` and the per-resource `ResourceView` subclasses now carry `@extend_schema`, and the update-generated-files check passes.

*— Claude Code*

apps/teams/management/commands/sync_team.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def force_delete_team(team_slug, state_dir, write=lambda _m: None):
+    """Delete the local team (matched by slug) and its sync-state DB so the next run re-imports from
+    scratch. Without resetting the state, the derived cursor would skip the rows that were deleted."""
+    deleted, _ = Team.objects.filter(slug=team_slug).delete(audit_action=AuditAction.AUDIT)
+    state_db = Path(state_dir) / f"{team_slug}.sqlite"
+    state_db.unlink(missing_ok=True)
+    write(f"force-deleted team '{team_slug}' and reset sync state ({deleted} objects removed)")
```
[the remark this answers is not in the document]
SmittieC: Declined — `sync_team` is an operator-run admin command with trusted `--team-slug` input, so the path-traversal risk is out of scope here. Noted for a future hardening pass if it ever takes untrusted input.

*— Claude Code*

api-schemas/v1.yml
SmittieC: This is actually addressed in full

apps/teams/export/manifest.py
```diff
@@ -0,0 +1,227 @@
...
+    cursor: str  # pk | updated_at_id
+    secret: bool = False
+
+
+# Pulled in dependency order: a referencing model always follows what it points at. Versioned models
+# (source material, consent form, pipeline, node, chatbot) page by "pk": a working version is always
+# created before its published copies, so it has a lower id and is served -- and imported -- first,
+# letting the self-referential working_version FK resolve. Guarded by
+# test_working_version_always_served_before_its_published_copies.
+MANIFEST_ENTRIES: list[ManifestEntry] = [
```
SmittieC, line 35: Please note: Not all models are yet exported. This PR serves as a POC. A followup PR will add the rest of the models to the manifest.

api-schemas/v2.yml
```diff
@@ -4,6 +4,99 @@ info:
   version: v2
   description: Build, deploy and monitor chatbots.
 paths:
+  /api/v2/auth_provider/:
+    get:
+      operationId: sync_auth_provider
```
snopoke: can we drop the `sync_` prefix on these?

api-schemas/v2.yml
```diff
@@ -4,6 +4,99 @@ info:
...
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ChatExportPage'
+          description: ''
+  /api/v2/chatbot/:
```
snopoke: We already have a `v2/chatbots` API - seems odd to have 2.

The usual convention for endpoints that return multiple objects is to use the plural form e.g. `chatbots`, `sessions`.

api-schemas/v2.yml
```diff
@@ -94,6 +187,176 @@ paths:
...
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/EmbeddingProviderModelExportPage'
+          description: ''
+  /api/v2/experiment_session/:
```
snopoke: Rename to `sessions` (matches v1 API and drops the legacy 'experiment' nomenclature)

api-schemas/v2.yml
```diff
@@ -94,6 +187,176 @@ paths:
...
+    get:
+      operationId: sync_experiment_session
+      summary: Experiment Session
```
snopoke: Use `Chatbot Sessions` (or just `Sessions`)

api-schemas/v2.yml
```diff
@@ -113,47 +376,419 @@ paths:
...
-            type: string
-      required:
-      - paths
-    Assistant:
-      type: object
-      description: Assistant fields. ``instructions`` and ``assistant_id`` are exposed
-        on purpose.
-      properties:
-        id:
+  /api/v2/membership/:
```
snopoke: I'd rename this to qualify what membership it refers to:

```suggestion
  /api/v2/team_memberships/:
```

Alternatively, it might be even better to use a name like `team_users` or even just `users` since that's what we are returning right?

api-schemas/v2.yml
```diff
@@ -113,47 +376,419 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/MessagingProviderExportPage'
+          description: ''
+        '409':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SyncErrorDetail'
+          description: Team has no registered public key; secret data cannot be sealed.
+  /api/v2/node/:
```
snopoke: nit: maybe:

```suggestion
  /api/v2/pipeline_node/:
```

api-schemas/v2.yml
```diff
@@ -113,47 +376,419 @@ paths:
...
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ParticipantExportPage'
+          description: ''
+  /api/v2/participant_data/:
```
snopoke: Just noting that it feels slightly weird to have this separate from the participant endpoint. But it does makes sense for the 'sync' use case.

api-schemas/v2.yml
```diff
@@ -113,47 +376,419 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/TraceProviderExportPage'
+          description: ''
+        '409':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SyncErrorDetail'
+          description: Team has no registered public key; secret data cannot be sealed.
+  /api/v2/user/:
```
snopoke: I expected the 'membership' endpoint to have this data - I don't think there should be separate endpoints for users and team membership since these APIs are team scoped. The user data should include their role within the team.

api-schemas/v2.yml
```diff
@@ -113,47 +376,419 @@ paths:
...
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SyntheticVoiceExportPage'
+          description: ''
+  /api/v2/teams/:
```
snopoke: Surprised to see this here (especially as a 'plural' endpoint).

api-schemas/v2.yml
```diff
@@ -180,6 +815,71 @@ components:
...
+        updated_at:
+          type: string
+          format: date-time
+          readOnly: true
+        type:
+          $ref: '#/components/schemas/AuthProviderExportItemTypeEnum'
+        name:
+          type: string
+          maxLength: 255
+        team:
```
snopoke: Is this necessary? All data is going to be scoped to the current team right?

api-schemas/v2.yml
```diff
@@ -239,23 +939,73 @@ components:
...
+          description: True while more rows remain beyond this page.
+        results:
+          type: array
+          items:
+            $ref: '#/components/schemas/ChatSync'
+      required:
+      - cursor
+      - has_more
+      - results
+    ChatSync:
```
snopoke: Just noting that it feels yuk to have this here since it's mostly a useless table. Could we add these fields to the session data?

api-schemas/v2.yml
```diff
@@ -434,121 +1267,475 @@ components:
...
+          description: True while more rows remain beyond this page.
+        results:
+          type: array
+          items:
+            $ref: '#/components/schemas/CustomUserSync'
+      required:
+      - cursor
+      - has_more
+      - results
+    CustomUserSync:
```
snopoke: Should just be `User`

api-schemas/v2.yml
```diff
@@ -434,121 +1267,475 @@ components:
...
+          type: integer
+          readOnly: true
+        last_login:
           type: string
-          description: JSON object mapping field name to a description of the value
-            to extract.
-        key_name:
+          format: date-time
+          nullable: true
+        is_superuser:
```
snopoke: I don't think we should expose `is_superuser` and `is_staff` since those are crosscutting permissions.

api-schemas/v2.yml
```diff
@@ -434,121 +1267,475 @@ components:
...
+          type: string
+          format: uri
+        language:
+          type: string
+          nullable: true
+          maxLength: 10
+      required:
+      - id
+      - username
+    EmbeddingProviderModelExportPage:
```
snopoke: All of these 'Page` objects are identical except for the results type. Is there a way we can reduce the duplication?

api-schemas/v2.yml
```diff
@@ -434,121 +1267,475 @@ components:
...
+          readOnly: true
+        type:
+          $ref: '#/components/schemas/TypeB81Enum'
+        name:
           type: string
-          nullable: true
-        max_token_limit:
+          description: The name of the model. e.g. 'text-embedding-3-small'
+          maxLength: 128
+        team:
```
snopoke: Instead of putting the team ID here, we could add a field like `is_global` or `is_team_custom` to indicate if it's a global vs team model.

api-schemas/v2.yml
```diff
@@ -434,121 +1267,475 @@ components:
...
+          additionalProperties: {}
+          nullable: true
+          description: Model call parameters (e.g. ``temperature``) keyed by name.
+            Older nodes may instead carry a flat ``llm_temperature``; both feed the
+            same underlying setting.
+        data_schema:
+          type: string
+          description: JSON object mapping field name to a description of the value
+            to extract.
+    FlattenedLlm:
```
snopoke, line 475: nit: rename to `NodeModelConfig` (node prefix to indicate that it's a node config schema)

api-schemas/v2.yml
```diff
@@ -434,121 +1267,475 @@ components:
...
+          type: integer
+          nullable: true
+          description: The model's maximum context window, in tokens.
+        deprecated:
+          type: boolean
+          nullable: true
+          description: True if this model is deprecated and should be migrated off.
+      required:
+      - type
+    FlattenedModelProvider:
```
snopoke, line 501: nit: rename to `CollectionModelConfig`

api-schemas/v2.yml
```diff
@@ -434,121 +1267,475 @@ components:
...
+          nullable: true
+        type:
+          type: string
+          nullable: true
+        model:
+          type: string
+          nullable: true
+      required:
+      - type
+    FlattenedVoice:
```
snopoke, line 522: Similar to others: rename to `NodeVoiceConfig`

api-schemas/v2.yml
```diff
@@ -434,121 +1267,475 @@ components:
...
+          nullable: true
+          description: True if this model is deprecated and should be migrated off.
+      required:
+      - type
+    FlattenedModelProvider:
+      type: object
+      description: |-
+        A collection's embedding provider and model, flattened into one object.
+
+        Expects a :class:`ProviderModelPair`; a missing provider or model renders its fields as null.
```
snopoke, line 506: remove unnecessary detail from description
```suggestion
        A collection's embedding provider and model.
```

api-schemas/v2.yml
```diff
@@ -434,121 +1267,475 @@ components:
...
+          type: string
+          nullable: true
+      required:
+      - type
+    FlattenedVoice:
+      type: object
+      description: |-
+        A voice provider and synthetic voice, flattened into one ``voice`` object (ADR-0025).
+
+        Expects a :class:`VoicePair`.
```
snopoke, line 527: remove unnecessary detail from description
```suggestion
        A voice provider and synthetic voice
```

api-schemas/v2.yml
```diff
@@ -434,121 +1267,475 @@ components:
...
+          type: string
+          nullable: true
+          description: Voice language/locale code; empty when the voice has no explicit
+            language set.
+        neural:
+          type: boolean
+          nullable: true
+          description: Use the provider's neural (higher-quality) voice engine, where
             supported.
+    GenderEnum:
```
snopoke: nit:
```suggestion
    VoiceGenderEnum:
```

api-schemas/v2.yml
```diff
@@ -1042,6 +2229,154 @@ components:
...
+        created_at:
+          type: string
+          format: date-time
+          readOnly: true
+        updated_at:
+          type: string
+          format: date-time
+          readOnly: true
+        type:
+          $ref: '#/components/schemas/TypeB81Enum'
```
snopoke: I expect there will be a warning about the naming here. You can use the `ENUM_NAME_OVERRIDES` setting to give this a better name and it will also stop outputting the warning.

api-schemas/v2.yml
```diff
@@ -1088,38 +2423,396 @@ components:
...
+          description: True while more rows remain beyond this page.
+        results:
+          type: array
+          items:
+            $ref: '#/components/schemas/NodeSync'
+      required:
+      - cursor
+      - has_more
+      - results
+    NodeSync:
```
snopoke:
```suggestion
    PipelineNode:
```

api-schemas/v2.yml
```diff
@@ -1088,38 +2423,396 @@ components:
...
+    PaginatedChatbotList:
+      type: object
+      required:
+      - results
+      properties:
+        next:
+          type: string
+          nullable: true
+          format: uri
+          example: http://api.example.org/accounts/?cursor=cD00ODY%3D"
```
snopoke: nit: shouldn't use example urls unless they are valid.

Also why does this use `next` and `previous` while the others have `cursor` and `has_more`?

apps/api/export/views.py
```diff
@@ -0,0 +1,190 @@
...
+        next_cursor = str(rows[-1].id) if rows else cursor
+        return rows, next_cursor, has_more
+
+    queryset = queryset.order_by("updated_at", "id")
+    if cursor:
+        keyset = CursorHelper.decode(cursor)
+        timestamp = parse_datetime(keyset["updated_at"])
+        if timestamp is None:
+            raise ValidationError("Invalid pagination cursor.")
+        queryset = queryset.filter(Q(updated_at__gt=timestamp) | Q(updated_at=timestamp, id__gt=keyset["id"]))
```
snopoke, line 159: We should make it clear in the docs where resources are paginated by `updated_at` since that implies that you could get duplicates so clients need to be able to handle that.

apps/api/general/views.py
```diff
@@ -0,0 +1,190 @@
...
+    ),
+]
+
+
+def resource_view(entry: ManifestEntry) -> type[ResourceView]:
+    """A ResourceView subclass documenting a single resource, for the URLConf to mount at that
+    resource's literal path. Subclassing keeps the auth and permissions (and therefore the security
+    schemes) identical to the base view; the subclass exists only to carry the per-resource schema."""
+    model = entry_model(entry.model)
+    view = type(f"{model.__name__}ExportView", (ResourceView,), {})
```
snopoke: If we set `verbose_name`, `verbose_name_plural`, `label` on the model Meta we could use those instead of the class name. That would give us more control of the naming without having to have a class -> name mapping.

https://docs.djangoproject.com/en/6.0/ref/models/options/#verbose-name

snopoke (commented):
Got part way though - reviewed the APIs but not the client / import pieces yet.

snopoke (commented):
Some general feedback:
* Use 'chatbot' instead of 'experiment'
* Keep internal details out of API schema descriptions
* Drop 'team' FK from resources

api-schemas/v2.yml
SmittieC: Fixed in 3db2a5453 — dropped the `sync_` prefix, so the operation ids are now `manifest`, `chatbot`, `participant`, etc.

api-schemas/v2.yml
SmittieC: Fixed in 3db2a5453 — added `is_staff` and `is_superuser` to the user export's `EXCLUDE_REGISTRY`, so they're no longer serialized (the importer falls back to the model defaults).

apps/api/export/views.py
SmittieC: Fixed in 3db2a5453 — the `updated_at`-paginated resources now carry a schema description noting that a row modified mid-pagination can reappear on a later page, so clients should treat results as upserts keyed by `id`.

api-schemas/v2.yml
SmittieC: Fixed in 3db2a5453 — added an `ENUM_NAME_OVERRIDES` entry so the provider `type` enum is now `LlmProviderTypeEnum` instead of `TypeB81Enum`. That also clears the collision warning you predicted.

api-schemas/v2.yml
SmittieC: Fixed in 3db2a5453 — renamed via `ENUM_NAME_OVERRIDES`; it's now `VoiceGenderEnum`.

api-schemas/v2.yml
SmittieC: Done in b5c43b868 — the export endpoints now live under `/api/v2/resources/` and the resource names are plural, so the export surface is `/api/v2/resources/chatbots/`, cleanly separated from the `chatbots` CRUD viewset rather than colliding with it.

api-schemas/v2.yml
SmittieC: Done in b5c43b868 — renamed the resource to `sessions` (now `/api/v2/resources/sessions/`), dropping the legacy 'experiment' nomenclature.

api-schemas/v2.yml
SmittieC: Done in b5c43b868 — the summary is derived from the resource name, so it now reads "Sessions".

api-schemas/v2.yml
SmittieC: Done in b5c43b868 — renamed to `pipeline_nodes`.

apps/api/export/serializers.py
```diff
@@ -49,10 +57,26 @@ def get_is_global(self, instance) -> bool:
     return get_is_global
 
 
+@extend_schema_serializer(component_name="SessionChat")
+class _EmbeddedChatSerializer(serializers.Serializer):
+    """The chat fields carried inline in a session row (chat isn't its own synced resource)."""
+
+    name = serializers.CharField()
```
snopoke: drop `name` from `_EmbeddedChatSerializer`

apps/api/export/serializers.py
```diff
@@ -74,13 +75,28 @@ def _session_chat(self, session):
...
 _FIELD_RESOLVERS: dict[str, dict] = {
     "teams.team": {"feature_flags": _feature_flags},
     "users.customuser": {"groups": _team_role_groups},
     "experiments.experimentsession": {"chat": _session_chat},
 }
 
+# Per-model extra serializer fields layered on top of the plain model dump -- nested relations the
+# dump won't surface on its own. Each value is a factory returning a fresh field instance per build.
+_EXTRA_FIELDS: dict[str, dict] = {
+    "experiments.experiment": {"versions": lambda: ChatbotVersionSerializer(many=True)},
```
snopoke: I was thinking this should be a separate endpoint: `/chatbots/<id>/versions`

SmittieC:
The new PR: https://github.com/dimagi/open-chat-studio/pull/3697

Issue #3664, opened 2026-06-19
Export API | Migration Checklist


SmittieC:
See https://docs.google.com/document/d/1GHU59EFizL9OFLZvDbjdQJZMYhUuh8LpmvbG7hir18Y/edit?tab=t.va15f9e7vcxz#heading=h.7oo4mkq1k0dj.

### Acceptance criteria
- A documented, ordered runbook an operator can follow without referring back to the design spec.
- Cutover and re-establishment steps cover every platform/exclusion in the spec.

SmittieC:
Be sure to mention how to create a public / private key pair. An online tool can also be used: https://cryptotools.net/rsagen

SmittieC:
https://docs.openchatstudio.com/tech-hub/migrate_team/

PR #3677, opened 2026-06-22
feat: add team public key for data export sealing

### LINKED ISSUE #3660: Export API | Add the ability to link a public key to a team
[Doc](https://docs.google.com/document/d/1GHU59EFizL9OFLZvDbjdQJZMYhUuh8LpmvbG7hir18Y/edit?tab=t.va15f9e7vcxz#heading=h.wiye95wquhwc).

The team model should get an `public_key` attribute that stores the public key in base64 format.

SmittieC:
### Product Description

Adds a **Public Key** setting to the team admin page under a new \"Data Export\" section. Team admins can register an asymmetric public key that OCS uses to seal secret data exported from the team during a migration.

### Technical Description
Resolves #3660 

- New `public_key` TextField on `Team` (blank, no encryption at rest — it's a public key).
- `TeamPublicKeyForm` + `set_public_key` POST view, gated behind `teams.change_team` permission (admins only). 
- Migration `0011_team_public_key` is additive with a default of `""` — fully backwards compatible.
- Also includes a small fix (`cfbb733`) to guard the `notify_deprecated_widget_versions` data migration command against an empty database (no widget channels), which would otherwise error when running migrations on a fresh DB after this PR due to a missing `Team.public_key` column mid-migration.

### Migrations

### Demo

image

### Docs and Changelog


apps/teams/forms.py
```diff
@@ -105,6 +105,21 @@ class Meta:
         }
 
 
+class TeamPublicKeyForm(forms.ModelForm):
```
snopoke, line 110: What about adding a 'clean' method to validate that it really is a public key? You could use `cryptography.hazmat.primitives.serialization.load_pem_private_key` to load the key (raises exception if not valid).

templates/teams/manage_team.html
```diff
@@ -88,6 +88,19 @@ <h3 class="pg-subtitle">{% translate "Team Members" %}</h3>
     </section>
     {% include "teams/components/team_invitations.html" %}
     {% if request.team_membership.is_team_admin %}
+      <section class="app-card">
+        <h3 class="pg-subtitle">{% translate "Data Export" %}</h3>
+        <p class="my-2 text-muted">
+          {% translate "Register a public key to seal secret data exported from this team during a migration." %}
```
snopoke, line 94: Do you think users will know how to create a key pair?

SmittieC: Most probably yes, but I do think we can / should add docs around this as well. Once the API stuff is in and we have docs, I can come back to this to link users to it. I added a comment on the [docs ticket](https://github.com/dimagi/open-chat-studio/issues/3664) to not forget.

apps/teams/forms.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/3677/commits/de2bc3520427b32512e36e44a2f70092e98c3fb9

PR #3678, opened 2026-06-22
Revert confirmation modal with diff and unreleased-changes warning

### LINKED ISSUE #3617: Revert confirmation modal with diff and unreleased-changes warning
## Parent

Part of #3398.

## What to build

Replace the native confirm from #3615 with an HTMX modal (same pattern as the version-details modal) that shows what revert will do before confirming:

- diff between the current working state and the target version, reusing the existing `VersionDetails.compare()` rendering from the create-version form
- prominent warning when `compare_with_latest()` detects unreleased changes that will be overwritten

## Acceptance criteria

- [ ] Modal shows a field/node-level diff of working state vs target version
- [ ] Unreleased changes produce a clear overwrite warning
- [ ] Confirming performs the revert; cancelling makes no changes

## Blocked by

- #3615

snopoke:
### Product Description
Reverting a chatbot to a previous version now opens a modal showing exactly what will change — a field/node-level diff of the working state against the target version — instead of a bare confirm dialog. A warning appears when the working version has unreleased changes that the revert would overwrite.

### Technical Description
Closes #3617. Follows the existing version-details modal pattern: a GET view (`chatbot_revert_confirm`) renders the diff via the existing `VersionDetails.compare()` rendering; the Revert button loads it over HTMX; confirming POSTs to the unchanged revert endpoint.

Worth a closer look:
- **`version_details` is no longer cached** (`@cached_property` → `@property`). `compare()` mutates the `VersionDetails` in place, so a shared cached instance let two comparisons on the same record clobber each other — the unreleased-changes check (working vs *latest*) was wiping the `changed` flags set by the diff (working vs *target*), silently hiding real differences. Building it is cheap (it holds lazy querysets), and dropping the cache also removes the `save()` → `_clear_version_cache()` staleness machinery across experiments/events/files. This touches shared versioning code; the full versioning/pipeline/event/assistant/file/custom-action suites pass.
- `editors-bundle.js` is now loaded on the chatbot home page and diff-views are reinitialised on `htmx:afterSettle`, so CodeMirror diffs actually render inside the modal (this also fixes the same gap in the existing View Details modal).

### Migrations
N/A — no schema changes.

### Docs and Changelog


apps/experiments/versioning.py
```diff
@@ -380,21 +379,17 @@ def compare_with_latest(self) -> bool:
             version.compare(prev_version.version_details, early_abort=True)
         return version.fields_changed
 
-    @cached_property
+    @property
     def version_details(self) -> VersionDetails:
+        # Intentionally NOT cached. `VersionDetails.compare()` mutates the returned object in place,
+        # so a cached (shared) instance lets independent comparisons on the same record clobber each
+        # other's results, and goes stale on mutation. Building it is cheap (the querysets it holds
+        # are lazy), so callers get a fresh instance each time.
```
snopoke:
```suggestion
```

apps/experiments/versioning.py
```diff
@@ -380,21 +379,13 @@ def compare_with_latest(self) -> bool:
             version.compare(prev_version.version_details, early_abort=True)
         return version.fields_changed
 
-    @cached_property
+    @property
```
snopoke, line 382: dropped this cache since it wasn't necessary and also isn't actually every used

apps/chatbots/tests/test_revert_views.py
```diff
@@ -1,8 +1,30 @@
...
+            if field.changed:
+                names.append(field.name)
+            if field.raw_value_version:
+                walk(field.raw_value_version)
+            for result in field.queryset_results or []:
+                if result.raw_value_version:
+                    walk(result.raw_value_version)
+
+    walk(version_details)
+    return names
```
snopoke, line 28: Addressed in 418ded6 — extracted the nested queryset traversal into a `walk_nested` helper, clearing the Bumpy Road biomarker (CodeScene now scores the file 10.0).

*— Claude Code*

PR #3680, opened 2026-06-22
Emit per-version OpenAPI schemas

snopoke:
### Product Description
API reference docs are now split by version. `/api/docs/` is a landing page linking to v1 and v2 references and the guides site.

### Technical Description
v2 endpoints were leaking into the v1 OpenAPI schema (`api-schemas/v1.yml`). Root cause: the custom `apps/api/versioning.py:URLPathVersioning` subclassed `BaseVersioning`, which drf-spectacular doesn't recognise as a versioning class — it logged "unsupported versioning class, processed as unversioned" (silently swallowed) and dumped every path into one schema.

Fix: subclass DRF's `URLPathVersioning` (keeping our path-regex `determine_version` and unversioned-`reverse()` behaviour) so per-`--api-version` filtering works. `inv schema` now emits `v1.yml` + `v2.yml`.

Worth a closer look:
- `version_param = "api_version"` on the versioning class is deliberate — drf-spectacular substitutes `version_param` into the path when emitting per-version schemas, and the stock default `"version"` collides with the `{version}` *experiment*-version path param on the openai/channels endpoints (would mangle `/v{version}/` → `/vv1/`).
- The live `/api/schema/` now serves **v1 only** (was the combined surface); v2 lives at `/api/v2/schema/`. This is intended per ADR-0022 but is a visible change to anyone reading v2 off the default docs page.
- `prune_unused_tags` postprocessing hook drops tag definitions a version doesn't use, so v2 no longer advertises v1's tags.

### Migrations
N/A — no migrations.

### Demo
- `/api/docs/` — landing page
- `/api/v1/docs/`, `/api/v2/docs/` — per-version Redoc
- `/api/schema/`, `/api/v2/schema/` — raw schemas

image

### Docs and Changelog

The `/api/docs/` URL now points to a landing page rather than the v1 Redoc directly; the v1 reference moved to `/api/v1/docs/`.


apps/api/schema.py
```diff
@@ -6,6 +6,25 @@
...
+    used = {
+        tag
+        for path in result.get("paths", {}).values()
+        for op in path.values()
+        if isinstance(op, dict)
+        for tag in op.get("tags", [])
+    }
+    if "tags" in result:
+        result["tags"] = [tag for tag in result["tags"] if tag["name"] in used]
+    return result
```
snopoke: Fixed in aa4189a75. `prune_unused_tags` only uses `result`, so the unused `generator`/`request`/`public` params are now folded into `**kwargs` (drf-spectacular calls hooks with keyword args).

*— Claude Code*

apps/api/versioning.py
```diff
@@ -27,3 +36,8 @@ def determine_version(self, request, *args, **kwargs):
         if not self.is_allowed_version(version):
             raise NotFound(self.invalid_version_message)
         return version
+
+    def reverse(self, viewname, args=None, kwargs=None, request=None, format=None, **extra):
+        # Skip DRF's version-kwarg injection: our version lives in the path text, not a kwarg, so
+        # always emit the canonical unversioned URL (BaseVersioning's behaviour).
+        return BaseVersioning.reverse(self, viewname, args, kwargs, request, format, **extra)
```
snopoke: Fixed in aa4189a75. `reverse()` is a thin pass-through to `BaseVersioning.reverse`, so it now takes `*args, **kwargs` instead of restating DRF's full signature.

*— Claude Code*

api-schemas/v2.yml
```diff
@@ -0,0 +1,1324 @@
...
+      operationId: me
+      description: Returns basic information about the authenticated user and the
+        team the token is scoped to.
+      summary: Current User
+      tags:
+      - Me
+      security:
+      - {}
+      - apiKeyAuth: []
+      - tokenAuth: []
```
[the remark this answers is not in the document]
snopoke: Fixed in aa4189a75. Good catch — `MeView` requires `IsAuthenticated`, so the `- {}` was wrong. Root cause was our `OAuth2TeamsScheme.get_security_requirement` returning `{}` for views with no required scope; it now returns `{OAuth2: []}` (auth required, any scope). `/api/v2/me/` security is now `[OAuth2: [], apiKeyAuth: [], tokenAuth: []]`.

*— Claude Code*

SmittieC (approved):
Nice. This is something I've been meaning to look into for a time now

PR #3684, opened 2026-06-23
Open Continue Chat in embedded widget when flag is active

snopoke:
### Product Description
When `flag_chat_widget` is on, the **Continue Chat** action in the sessions table now resumes the conversation in the embedded chat widget (a floating popup, same UX as the existing widget) instead of navigating to the full-page chat UI. With the flag off, behaviour is unchanged (full-page link). Applies to both the single-chatbot home page and the All Sessions page.

### Technical Description
The widget binds to its session only during `componentWillLoad` (no `@Watch` on `sessionId` in the published `open-chat-studio-widget@0.9.1`), so resuming a session needs a *freshly mounted* element rather than re-setting props on the existing one. `ocsContinueSessionChat` therefore creates a new session-bound widget on each click (removing any prior one). This also makes it work on the All Sessions page, which spans multiple chatbots and has no pre-existing widget.

The Continue Chat action carries a per-row signed `session_token` (`issue_session_token`) only when the flag is active; otherwise it falls back to the legacy link template, preserving the old path.

No changes to the widget package itself — purely OCS app/template code.

### Migrations
N/A — no migrations.

### Demo

<div>
    <a href="https://www.loom.com/share/097c55ca49ac453d98fe589c98df8602">
      <p>Chat Widget Reuses Previous Bot Sessions - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/097c55ca49ac453d98fe589c98df8602">
      
    </a>
  </div>

### Docs and Changelog

### DIFF
--- a/apps/chatbots/tests/test_chatbot_views.py
+++ b/apps/chatbots/tests/test_chatbot_views.py
@@ -1,3 +1,4 @@
+import re
 from datetime import UTC, datetime
 from unittest.mock import Mock, patch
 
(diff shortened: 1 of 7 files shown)


apps/chatbots/tests/test_chatbot_views.py
```diff
@@ -362,6 +363,65 @@ def test_chatbot_sessions_table_view(team_with_users):
     assert isinstance(response.context_data["table"], ChatbotSessionsTable)
 
 
+@pytest.mark.django_db()
+def test_continue_chat_launcher_present_on_session_pages_when_flag_active(client, team_with_users):
```
snopoke: remove test

snopoke: Removed in 21a6f079a.

*— Claude Code*

apps/chatbots/tests/test_chatbot_views.py
```diff
@@ -1,3 +1,4 @@
+import re
```
snopoke, line 1: Reduced in 21a6f079a by removing the launcher-presence test. The remaining ~16-line parametrized test is the core behavioral coverage (flag on/off branching + session-token validity), and this test module was already close to the threshold before this PR, so it stays just over 600 LOC. Advisory — happy to suppress if acceptable.

*— Claude Code*

PR #3687, opened 2026-06-23
feat(cost-tracking): dashboard, REST, digest, overrides, rate-change (PR3/3)

davidnene:
### Product Description

PR 3 of 3 in the AI Cost Tracking roadmap. Closes out the user-facing surface and the operational tooling on top of PR 1's foundation and PR 2's capture path.

For teams with `flag_ai_cost_monitoring` enabled:
- Cost Tracking panel on the team dashboard: period spend with delta vs prior period, exact / estimated breakdown, top-N chatbots, pricing-last-synced footer. Reacts to the dashboard date filter without a page reload.
- Per-1K rate sub-line under every LLM provider model on the team settings page. HTMX flow to override a global rate at team scope or revert to global. The custom-model creation dialog accepts optional input/output rates.
- Two read-only v2 API endpoints (`/api/v2/cost_tracking/usage/`, `/pricing/`) gated by a new `cost_tracking:read` OAuth scope.

Operations layer (not flag-gated):
- Weekly Celery digest emails the platform team a roll-up of unpriced models and unknown-call coverage gaps.
- The daily auto-update-models workflow gains a sibling job that diffs llm-stats.com pricing against the in-repo seed and opens a "Pricing update" PR when rates move.
- Bulk seed backfill (52 models priced via LiteLLM); `test_seed_coverage` is unskipped with a documented `KNOWN_UNPRICED` allow-list for the six models LiteLLM has no rate for.

### Technical Description

- `apps/cost_tracking/services/reporting.py`: `cost_summary`, `top_n_bots`, `last_synced_at` as single-aggregation, team-scoped reads. N+1 gated by `CaptureQueriesContext` in tests.
- Override / revert / create-with-pricing HTMX flow on the LLM provider page. Form errors re-render the modal in place via `HX-Retarget`; the Save button only closes on `event.detail.successful`. Supersession of team rules runs in a transaction so the unique-active-rule constraint catches racing writers.
- Bulk `_pricing_lookup` joins active rules onto a model list in one query; precomputes `primary` and `has_team_override` synthetic keys so the template avoids Django's eager `|default:` resolution.
- Dashboard panel: server-rendered on initial page load; `loadCostTrackingPanel()` JS swaps the partial on filter change via `api/cost-tracking-panel/`.
- Weekly digest aggregates `UsageRecord` rows cross-team, summing `extra->>missing_usage_calls` via Postgres JSON extraction so the report counts actual failed calls.
- `scripts/diff_seed_pricing.py` (stdlib-only, dependency-injected fetcher) rewrites the seed JSON and emits a `load_pricing_data()` migration when rates differ.
- `backfill_pricing_seed` management command: idempotent against the existing seed JSON; safe to re-run.
- Django admin: `PricingRule` full-edit (operator backfill); `UsageRecord` read-only (tracer-written).
- Flag enum registration so team admins self-toggle via the Feature Flags page.

### Deferred to follow-up

- Production-side `ALTER TABLE cost_tracking_pricingrule ADD COLUMN updated_at` for environments running PR 1's original `0001_initial`. The new `0003_pricingrule_updated_at` uses `ADD COLUMN IF NOT EXISTS` + `state_operations` for fresh installs but a deployed instance skips it (Django sees `0001` as already applied). Such environments need a manual ALTER on the way to this branch.

### Migrations

`0003_pricingrule_updated_at` uses `ADD COLUMN IF NOT EXISTS` + `state_operations` to reconcile model vs DB without breaking legacy installs. Loader-driven data migrations (`load_pricing_data`) are additive only.

### Demo

[Demo link](https://drive.google.com/file/d/1s50xfZotbd293QDknBbACSax7SxLevjv/view?usp=sharing)
- Enable `flag_ai_cost_monitoring` for your team via the Feature Flags page.
- Run a chat turn (UsageRecord rows land via OCSTracer from PR 2).
- Open the team dashboard — the Cost Tracking panel sits below the overview cards and reacts to the date filter.
- Open the team's LLM Provider page — each model row shows its per-1K rate with Override / Revert / Set pricing buttons.
- `GET /api/v2/cost_tracking/usage/` with the `cost_tracking:read` scope.
- `python manage.py send_unpriced_usage_digest` (or wait for Monday 09:00 UTC).

### Docs and Changelog


davidnene:
> [Review Change Stack](https://app.coderabbit.ai/change-stack/dimagi/open-chat-studio/pull/3687?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)
> 
> 📝 Walkthrough
> Thanks for using [CodeRabbit](https://coderabbit.ai?utm_source=oss&utm_medium=github&utm_campaign=dimagi/open-chat-studio&utm_content=3687)! It's free for OSS, and your support helps us grow. If you like it, consider giving us a shout-out.
> 
> ❤️ Share
> Comment `@coderabbitai help` to get the list of available commands.

resolved

apps/cost_tracking/models.py
```diff
@@ -61,6 +60,7 @@ class PricingRule(models.Model):
     notes = models.TextField(blank=True)
 
     created_at = models.DateTimeField(auto_now_add=True)
```
SmittieC, line 62: Nit: Prefer using `BaseModel` which adds the created_at and updated_at fields. It's an established pattern in the code already

apps/cost_tracking/migrations/0003_pricingrule_updated_at.py
snopoke, line 1: This shouldn't be necessary. I don't think we need this field on this model and the migration which contained it was never executed on a production env (only dev).

apps/cost_tracking/models.py
```diff
@@ -61,6 +60,7 @@ class PricingRule(models.Model):
...
+    updated_at = models.DateTimeField(auto_now=True)
```
snopoke: I don't think we need this one here

apps/cost_tracking/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    unknown calls. Skips the email when the digest is empty so an inbox
+    isn't filled with no-news messages."""
+    end = timezone.now()
+    start = end - timedelta(days=_DIGEST_WINDOW_DAYS)
+    summary = build_digest(start, end)
+    if summary.is_empty:
+        logger.info("cost_tracking.digest.skipped_empty", extra={"start": start.isoformat()})
+        return
+    recipient = _operator_email()
+    if not recipient:
```
snopoke: do this check before computing the digest

apps/api/v2/cost_tracking/serializers.py
```diff
@@ -0,0 +1,63 @@
...
+    exact_cost = serializers.DecimalField(max_digits=14, decimal_places=8)
+    estimated_cost = serializers.DecimalField(max_digits=14, decimal_places=8)
+    unknown_call_count = serializers.IntegerField()
+    last_synced = serializers.DateTimeField(allow_null=True)
+
+
+class BotSpendSerializer(serializers.Serializer):
+    """Mirrors `services.reporting.BotSpend`."""
+
+    experiment_id = serializers.IntegerField()
```
snopoke: v2 APIs must use the term 'chatbot' instead of 'experiment'

apps/api/v2/cost_tracking/views.py
```diff
@@ -0,0 +1,121 @@
...
+    """Active PricingRules visible to the caller's team (team overrides
+    plus globals). Read-only snapshot."""
+
+    permission_classes = [IsAuthenticated, TokenHasOAuthResourceScope]
+    required_scopes = ["cost_tracking"]
+
+    @extend_schema(
+        operation_id="cost_tracking_pricing",
+        summary="Cost-tracking pricing",
+        tags=["Cost Tracking"],
```
snopoke: tags should also be declared in settings.py

apps/api/v2/cost_tracking/views.py
snopoke, line 1: Can we drop these API's for now - I think they warrant a bit more design before we publish them (since it's hard to make breaking changes to APIs once they are in use).

snopoke (commented):
This mostly seems OK. I'm a bit unclear on what the usages of the different management commands and scripts is. We now have

* load_ai_pricing (previous PR)
* backfill_pricing_seed
* diff_seed_pricing

Can you walk through what each is for and when they are used (and why we need all 3).

Please also add a demo video to the PR since this PR includes a number of UI changes.

apps/api/v2/cost_tracking/views.py
davidnene: that's okay

davidnene:
> This mostly seems OK. I'm a bit unclear on what the usages of the different management commands and scripts is. We now have
> 
> * load_ai_pricing (previous PR)
> * backfill_pricing_seed
> * diff_seed_pricing
> 
> Can you walk through what each is for and when they are used (and why we need all 3).
> 
> Please also add a demo video to the PR since this PR includes a number of UI changes.


They all at different points in the pricing-data lifecycle
load_ai_pricing - reads seed_data/llm_pricing.json and upserts PricingRule rows with supersession. Called from every pricing data migration  and each NNNN_rate_update_<date>.py that diff_seed_pricing generates. So this is what actually mutates the DB

backfill_pricing_seed - walks DEFAULT_LLM_PROVIDER_MODELS and, for any (provider, model) not yet in the seed, fetches prices from LiteLLM's model_prices_and_context_window.json and writes them into llm_pricing.json. Idempotent and gap-only: existing seed entries are left verbatim.

scripts/diff_seed_pricing.py - this watches the rate change. Runs daily from .github/workflows/auto-update-models.yml (the diff-pricing job). For each model already in the seed, fetches current prices from llm-stats.com. If a price has moved, it rewrites the seed entry and generates a new  apps/cost_tracking/migrations/NNNN_rate_update_<YYYYMMDD>.py. The workflow then opens a Pricing update PR.

the other option would be to edit the json by hand which I don't think is a good idea, quite long and typos etc

davidnene:
@snopoke I have resolve your comments and also attached the demo on PR body. You can also find it [here](https://drive.google.com/file/d/1s50xfZotbd293QDknBbACSax7SxLevjv/view?usp=sharing)

apps/service_providers/views.py
```diff
@@ -286,26 +307,225 @@ def _get_models_by_type(queryset):
...
             "for_type": form.cleaned_data["type"],
+            "cost_tracking_enabled": cost_tracking_enabled,
+            "pricing_lookup": _pricing_lookup(request.team, list(custom_models)) if cost_tracking_enabled else {},
+        },
+    )
+
+
+@login_and_team_required
+@permission_required("service_providers.change_llmprovidermodel", raise_exception=True)
+def pricing_override_form(request, team_slug: str, pk: int):
```
SmittieC: Why not use a class based view for `pricing_override_form` and `pricing_override` instead? It will make the form handling less-verbose as well

apps/service_providers/views.py
```diff
@@ -286,26 +307,225 @@ def _get_models_by_type(queryset):
...
+    `select_for_update()` on the close-out query serialises concurrent
+    overrides so the partial-unique active-rule constraint never trips."""
+    now = timezone.now()
+    with transaction.atomic():
+        for field_name, kind in _FORM_FIELD_TO_KIND.items():
+            per_million = cleaned.get(field_name)
+            if per_million is None:
+                continue
+            list(
+                PricingRule.objects.select_for_update().filter(
```
SmittieC: why not do this and the update query below this in the same query?

apps/cost_tracking/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    """Email the platform team a weekly summary of unpriced models and
+    unknown calls. Skips the email when the digest is empty so an inbox
+    isn't filled with no-news messages."""
+    recipient = _operator_email()
+    if not recipient:
+        logger.warning("cost_tracking.digest.no_recipient")
+        return
+    end = timezone.now()
+    start = end - timedelta(days=_DIGEST_WINDOW_DAYS)
+    summary = build_digest(start, end)
```
SmittieC: if you check the existence of `recipient` at the top of the method, we don't have to do a bunch of computing in `build_digest` just to see that it wasn't necessary

apps/cost_tracking/services/digest.py
```diff
@@ -0,0 +1,92 @@
...
+    def is_empty(self) -> bool:
+        return not self.unpriced_rows and not self.unknown_rows
+
+
+def build_digest(start: datetime, end: datetime) -> DigestSummary:
+    """Two single-query aggregations, cross-team. `extra->>missing_usage_calls`
+    is summed via JSON extraction so the digest reports actual call counts
+    rather than UsageRecord row counts (rows fold multiple failed calls).
+    """
+    unpriced = list(
```
SmittieC: I think we should only consider usage records of models that we (OCS) maintain and ignore custom team scoped models. We can do nothing about them.

apps/cost_tracking/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+
+from apps.cost_tracking.services.digest import DigestSummary, build_digest
+
+logger = logging.getLogger("ocs.cost_tracking")
+
+_DIGEST_WINDOW_DAYS = 7
+
+
+@shared_task(ignore_result=True)
+def send_unpriced_usage_digest() -> None:
```
SmittieC: I wonder about the utility of this task. I'm thinking this type of notification(s) belongs in a github workflow. It is nice to see usage of unpriced models, which creates a sort of urgency per model, but even then I'm not entirely sure if this sort of thing belongs in the codebase itself. I'm curious what @snopoke's thoughts on this are.

apps/cost_tracking/models.py
snopoke: We didn't do this initially because this model overrides most of the fields in the base model.

apps/cost_tracking/tasks.py
snopoke: Yes that probably does make sense. Originally this was here because it was meant to also track pricing drift (though that's not actually possible) and I haven't re-considered it since then. 

If we are only looking at models managed by OCS (present in the default_models.py file) then there is no need to have check against usage data, we can just check the static pricing data.

snopoke:
> @snopoke I have resolve your comments and also attached the demo on PR body. You can also find it [here](https://drive.google.com/file/d/1s50xfZotbd293QDknBbACSax7SxLevjv/view?usp=sharing)

Is there a way you can show what the dashboard looks like with actual usage data?

snopoke:
> backfill_pricing_seed - walks DEFAULT_LLM_PROVIDER_MODELS and, for any (provider, model) not yet in the seed, fetches prices from LiteLLM's model_prices_and_context_window.json and writes them into llm_pricing.json. Idempotent and gap-only: existing seed entries are left verbatim.

How does this related to the `auto-update-models.yml` github workflow?

> scripts/diff_seed_pricing.py - this watches the rate change. Runs daily from .github/workflows/auto-update-models.yml (the diff-pricing job). For each model already in the seed, fetches current prices from llm-stats.com. If a price has moved, it rewrites the seed entry and generates a new apps/cost_tracking/migrations/NNNN_rate_update_.py. The workflow then opens a Pricing update PR.

This seems to overlap a lot with `scripts/fetch_model_updates.py` - I think it would make sense to merge these two.

apps/cost_tracking/tasks.py
davidnene: okay, will focus on the default models

apps/cost_tracking/tasks.py
davidnene: this was solved

davidnene:
> > backfill_pricing_seed - walks DEFAULT_LLM_PROVIDER_MODELS and, for any (provider, model) not yet in the seed, fetches prices from LiteLLM's model_prices_and_context_window.json and writes them into llm_pricing.json. Idempotent and gap-only: existing seed entries are left verbatim.
> 
> How does this related to the `auto-update-models.yml` github workflow?
> 
> > scripts/diff_seed_pricing.py - this watches the rate change. Runs daily from .github/workflows/auto-update-models.yml (the diff-pricing job). For each model already in the seed, fetches current prices from llm-stats.com. If a price has moved, it rewrites the seed entry and generates a new apps/cost_tracking/migrations/NNNN_rate_update_.py. The workflow then opens a Pricing update PR.
> 
> This seems to overlap a lot with `scripts/fetch_model_updates.py` - I think it would make sense to merge these two.

I have refactored, we now have one reconcile file

davidnene:
> > @snopoke I have resolve your comments and also attached the demo on PR body. You can also find it [here](https://drive.google.com/file/d/1s50xfZotbd293QDknBbACSax7SxLevjv/view?usp=sharing)
> 
> Is there a way you can show what the dashboard looks like with actual usage data?

image

it currently looks like this. seeded some dummy data

davidnene:
@snopoke @SmittieC kindly review current updates

.github/workflows/auto-update-models.yml
```diff
@@ -32,47 +35,101 @@ permissions:
...
+            --title "$PR_TITLE" \
+            --body-file "$PR_BODY_PATH"
+
+      - name: Open or update missing-pricing issue
+        if: steps.reconcile.outputs.has_missing_pricing == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ISSUE_BODY_PATH: ${{ steps.reconcile.outputs.missing_pricing_issue_body_path }}
+          MISSING_COUNT: ${{ steps.reconcile.outputs.missing_pricing_count }}
+        run: |
```
snopoke, line 110: Can this also notify the `@dimagi/open-chat-studio` team so that we know about the new issue.

apps/cost_tracking/README.md
```diff
@@ -0,0 +1,68 @@
+# Cost Tracking App
+
+This app records the cost of every LLM call OCS makes and surfaces it to the team that owns the chat. The whole feature is gated by the `flag_ai_cost_monitoring` team-scoped Waffle flag, so teams opt in.
+
+There are two halves: the capture path (record what happened) and the resolution path (price what was recorded). A small operational layer (seed loader, auto-update workflow, weekly digest) keeps pricing data fresh.
+
+The source of truth is the code so [![Ask DeepWiki](https://deepwiki.com/badge.svg)](https://deepwiki.com/dimagi/open-chat-studio)
```
snopoke: nit: unnecessary 
```suggestion
```

apps/cost_tracking/README.md
```diff
@@ -0,0 +1,68 @@
...
+## Surface
+
+Two places consume the data, both gated by `flag_ai_cost_monitoring`:
+
+- **Dashboard panel** (`templates/dashboard/_cost_tracking_panel.html`). Period spend, delta vs prior period, exact/estimated breakdown, top-N chatbots. Reacts to the dashboard date filter via `dashboard:api_cost_tracking_panel`.
+- **LLM Provider page** shows each model's current per-1K rate inline. Admins can override at team scope via an HTMX modal (`pricing_override` view) or revert to global. The custom-model creation dialog accepts optional input/output rates that persist as team-scoped `PricingRule` rows in the same transaction as the model save.
+
+Coverage gaps for OCS-managed models (entries in `default_models.py` missing pricing) are surfaced from `auto-update-models.yml` as a GitHub issue, not as a runtime task. See "Seed Data and Updates" above.
+
+## Layout
```
snopoke, line 48: This is unnecessary and goes out of date easily.

apps/service_providers/forms.py
```diff
@@ -373,6 +374,23 @@ class LangfuseTraceProviderForm(ObfuscatingMixin, ProviderTypeConfigForm):
...
+        required=False,
+        min_value=Decimal("0"),
+        decimal_places=6,
+    )
+    output_price_per_million_tokens = forms.DecimalField(
+        label=_("Output price ($ / 1M tokens)"),
+        required=False,
+        min_value=Decimal("0"),
+        decimal_places=6,
+    )
```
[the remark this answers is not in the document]
snopoke: @davidnene this seems relevant

apps/service_providers/views.py
```diff
@@ -265,12 +278,20 @@ def _get_context(self, primary_form, config_form, subtype, instance):
             default_llm_models_by_type = _get_models_by_type(LlmProviderModel.objects.filter(team=None))
             embedding_models_by_type = _get_models_by_type(EmbeddingProviderModel.objects.filter(team=None))
             custom_llm_models_by_type = _get_models_by_type(LlmProviderModel.objects.filter(team=self.request.team))
+            cost_tracking_enabled = Flag.get("flag_ai_cost_monitoring").is_active_for_team(self.request.team)
```
snopoke: You can use the `waffle.flag_is_active(request, "flag_ai_cost_monitoring")` utility here. 

OR you can use the Django template tag: `{% flag "flag_ai_cost_monitoring" %}` directly in the UI

apps/service_providers/views.py
```diff
@@ -286,26 +307,208 @@ def _get_models_by_type(queryset):
...
-    if form.is_valid():
-        model = form.save(commit=False)
-        model.team = request.team
-        model.save()
-    else:
+    if not form.is_valid():
         if len(form.errors) == 1 and "__all__" in form.errors:
             return HttpResponseBadRequest(", ".join([str(v) for v in form.errors.values()]))
         return HttpResponseBadRequest(str(form.errors))
+    cost_tracking_enabled = Flag.get(COST_TRACKING_FLAG).is_active_for_team(request.team)
```
snopoke: again prefer the `waffle.flag_is_active` utility when a request is available

apps/service_providers/views.py
```diff
@@ -286,26 +307,208 @@ def _get_models_by_type(queryset):
...
+        },
+    )
+
+
+class PricingOverrideView(LoginAndTeamRequiredMixin, PermissionRequiredMixin, django_views.View):
+    """GET renders the override modal; POST persists team-scoped rules.
+    Invalid submissions re-render the form (with field errors) into the modal
+    body via HX-Retarget so the user can correct in place."""
+
+    template_name = "service_providers/components/_pricing_override_form.html"
```
snopoke: why does this file start with an `_`?  (also `_llm_model_row.html`). I don't think that's a pattern that we use elsewhere (at least not intentionally)

apps/service_providers/views.py
```diff
@@ -286,26 +307,208 @@ def _get_models_by_type(queryset):
...
+    ).update(effective_to=timezone.now())
+    return _render_model_row(request, model)
+
+
+def _resolve_model(team, pk: int) -> LlmProviderModel:
+    """Both team-scoped customs and the global defaults are addressable."""
+    return get_object_or_404(LlmProviderModel, Q(team=team) | Q(team__isnull=True), pk=pk)
+
+
+def _require_cost_tracking_flag(team) -> None:
```
snopoke: you can replace this with the `@waffle_flag('flag_name')` decorator: https://waffle.readthedocs.io/en/stable/usage/decorators.html

snopoke (commented):
LGTM - mostly minor feedback

davidnene:
@snopoke I have worked on your comments. Kindly review

snopoke (commented):
@davidnene 
I did some testing locally and had some issues:

**Setup**
1. ran migrations
2. ran `load_ai_pricing`
3. configured a bot to use gpt-5.4-mini
4. confirmed pricing for that model

Image

5. chatted to the bot to generate some usage

**Issues**
* Usage records are created with 'unknown' provider and 0 cost (though confidence is 'exact'). There is no linked pricing rule either.
    * Something wrong in the resolver that isn't resolving pricing. I tried flushing the cache but that didn't work.

Image

* The dashboard shows the 'no usage' message despite there being usage (granted it's all 0 but still usage).
    * I think this is because the confidence for the usage is exact but the cost is 0.

davidnene:
> @davidnene I did some testing locally and had some issues:
> 
> **Setup**
> 
> 1. ran migrations
> 2. ran `load_ai_pricing`
> 3. configured a bot to use gpt-5.4-mini
> 4. confirmed pricing for that model
> 
> Image
> 5. chatted to the bot to generate some usage
> 
> **Issues**
> 
> * Usage records are created with 'unknown' provider and 0 cost (though confidence is 'exact'). There is no linked pricing rule either.
>   
>   * Something wrong in the resolver that isn't resolving pricing. I tried flushing the cache but that didn't work.
> 
> Image
> * The dashboard shows the 'no usage' message despite there being usage (granted it's all 0 but still usage).
>   
>   * I think this is because the confidence for the usage is exact but the cost is 0.

Hello @snopoke lemmi check it out

davidnene:
m

> > @davidnene I did some testing locally and had some issues:
> > **Setup**
> > 
> > 1. ran migrations
> > 2. ran `load_ai_pricing`
> > 3. configured a bot to use gpt-5.4-mini
> > 4. confirmed pricing for that model
> > 
> > Image
> > 5. chatted to the bot to generate some usage
> > **Issues**
> > 
> > * Usage records are created with 'unknown' provider and 0 cost (though confidence is 'exact'). There is no linked pricing rule either.
> >   
> >   * Something wrong in the resolver that isn't resolving pricing. I tried flushing the cache but that didn't work.
> > 
> > Image
> > * The dashboard shows the 'no usage' message despite there being usage (granted it's all 0 but still usage).
> > 
> > * I think this is because the confidence for the usage is exact but the cost is 0.
> 
> Hello @snopoke lemmi check it out

Hello @snopoke there was a bug, LlmService._type was declared as _type: str = "unknown" and pydantic treats leading-underscore class attributes as private and silently drops them from init kwargs. Quite hidden. I have fixed and tested with a chatbot. its now working
image

snopoke (approved):

PR #3697, opened 2026-06-25
Team data export and sync

SmittieC:
Same change as #3675, re-sequenced into a clean commit history — **review commit-by-commit**: shared primitives (`seal` → `manifest`), the source read-API, the target sync/import half, then two `fix:` commits that reconcile leftovers from the original branch's churn.

### Product Description
A team admin registers an RSA public key for their team, and an operator replicates a team's data (chatbots, pipelines, providers, participants, sessions, …) from one OCS server to another with the new `sync_team` management command. Secrets — provider configs, participant data — are encrypted to the team's public key in transit and only decrypted on the target with the matching private key.

### Technical Description
**POC** — not all models are exported yet; a follow-up adds the rest to the manifest.

**Source — read API (team-admin only):** `GET /api/export/manifest/` declares the models to pull in dependency order plus each model's config, and doubles as the allowlist (resource endpoints refuse anything not listed). `GET /api/export/<resource>/` is generic, team-scoped and keyset-paginated; per-model serializers are built at runtime so they can't drift from the model. Secret fields use envelope encryption (a random Fernet key wrapped by the team's RSA key) so values larger than the RSA block still seal; the endpoint fails closed with 409 when the team has no public key.

**Target — `sync_team` + import engine:** a SQLite FK translation table maps source→target ids and doubles as a resume checkpoint, so a run resumes where it left off. The importer rewrites every reference from source to target ids (including ids embedded in pipeline/node params), restores source timestamps past `auto_now`, matches shared/global rows by natural key, maps existing users by username without overwriting them, and emails new users a password reset (passwords are never migrated). It aborts on a manifest schema-checksum mismatch unless `--skip-schema-check` is passed.

Worth a closer look:
- The export surface is standalone and **unversioned** (`/api/export/`, not v1/v2) — it carries no backwards-compatibility guarantee.
- The catch-all `re_path(r"^(?P<resource>[a-z_]+)/$")` sits after the router include, so any future plain-lowercase route must precede it.
- Regenerating the schema (final commit) picked up the `Team.metadata` field (internal staff-only) now on main — so it is currently exported. If that shouldn't leave the source, add it to the manifest's exclusions. Left as a follow-up decision.

### Migrations
N/A — the `Team.public_key` field and its migration already landed on main.

### Demo
https://www.loom.com/share/5dc056c2473342e6962bdc2002933b8f (from #3675)

\`\`\`
manage.py sync_team --source-url=<src> --api-key=<key> --team-slug=<slug> --private-key-path=<path>
\`\`\`
Rerun to resume / pick up new data.

### Docs and Changelog


apps/api/export/views.py
```diff
@@ -0,0 +1,227 @@
...
+    @extend_schema(
+        operation_id="manifest",
+        tags=["Manifest"],
+        summary="Manifest",
+        description=(
+            "Returns the resource manifest: resource call order, per-model config, and a schema"
+            " checksum clients can use to detect schema compatability."
+        ),
+        responses=ManifestSerializer,
+    )
```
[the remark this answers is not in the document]
SmittieC: let's remove the export tag then

apps/api/export/serializers.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def component_name(model) -> str:
+    """The base name for a synced model's OpenAPI components: its ``verbose_name`` PascalCased."""
+    return "".join(word[:1].upper() + word[1:] for word in str(model._meta.verbose_name).split())
+
+
+@cache
+def build_resource_serializer(model):
+    # Cached so every caller (the export resource endpoint and the v2 chatbot retrieve) shares one
```
SmittieC: this comment is outdated

apps/api/export/tests/test_views.py
```diff
@@ -0,0 +1,222 @@
...
+def test_team_url_resolves():
+    assert reverse("api:export:team") == "/api/export/team/"
+    assert resolve("/api/export/team/").url_name == "team"
+
+
+def test_resource_url_resolves():
+    assert _resource_url("users") == "/api/export/users/"
+
+
+def test_manifest_requires_authentication():
```
SmittieC: I don't think auth is required for this. We don't return any sensitive data here

apps/teams/export/client.py
```diff
@@ -0,0 +1,64 @@
...
+
+class SourceClient:
+    def __init__(self, base_url, api_key, *, timeout=30, max_retries=5, backoff=1.0, session=None, sleep=time.sleep):
+        self.base_url = base_url.rstrip("/")
+        self.api_key = api_key
+        self.timeout = timeout
+        self.max_retries = max_retries
+        self.backoff = backoff
+        self._session = session or requests.Session()
+        self._sleep = sleep
```
SmittieC: The extra parameters (`timeout`, `max_retries`, `backoff`, `session`, `sleep`) are keyword-only dependency-injection seams with production defaults; real call sites pass only `base_url`/`api_key`. Keeping them explicit lets tests inject a fake session/clock without patching. No code change needed.

apps/teams/export/importer.py
```diff
@@ -0,0 +1,309 @@
+"""The import engine: pull a serialized row, rewrite every reference from source ids to target ids,
```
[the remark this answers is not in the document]
SmittieC: The mean complexity (4.24) is marginally over the 4.0 threshold across 21 small functions. The module is a cohesive import engine and splitting it would trade readability for the metric, so leaving as-is. No change needed.

api-schemas/export.yml
```diff
@@ -0,0 +1,1984 @@
...
+        name: limit
+        schema:
+          type: integer
+        description: Maximum rows per page (default 100, capped at 1000).
+      tags:
+      - Team
+      security:
+      - {}
+      - apiKeyAuth: []
+      - tokenAuth: []
```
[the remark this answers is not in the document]
SmittieC: Fixed in 719ca20b0. The `- {}` came from the OAuth2 security scheme returning an empty requirement for views that carry no OAuth scope permission. The export views now set `authentication_classes = [ApiKeyAuthentication, BearerTokenAuthentication]` (OAuth2 isn't used on this surface), so the regenerated schema requires `apiKeyAuth`/`tokenAuth` with no public entry. An unauthenticated request still returns 401.

apps/api/export/tests/test_serializers.py
```diff
@@ -0,0 +1,136 @@
...
+    the redundant team FK."""
+    chat = ChatFactory(translated_languages=["en", "fr"], metadata={"k": "v"})
+    data = _serialize(Chat, chat)
+    assert data["translated_languages"] == ["en", "fr"]
+    assert data["metadata"] == {"k": "v"}
+    assert "team" not in data
+
+
+@pytest.mark.parametrize("is_global", [True, False])
+def test_global_able_model_exposes_is_global_flag_instead_of_team(is_global):
```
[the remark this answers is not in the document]
SmittieC: Fixed in a60f9dd3a -- now `pytest.param(True, id="global")` / `pytest.param(False, id="team_scoped")`.

apps/api/export/urls.py
```diff
@@ -0,0 +1,27 @@
...
+    *[
+        path(
+            f"{entry.resource}/",
+            views.resource_view(entry).as_view(),
+            {"resource": entry.resource},
+            name=f"resource-{entry.resource}",
+        )
+        for entry in MANIFEST_ENTRIES
+    ],
+    re_path(r"^(?P<resource>[a-z_]+)/$", views.UnknownResourceView.as_view(), name="resource"),
```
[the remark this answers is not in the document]
SmittieC: Fixed in 719ca20b0 -- the catch-all is now `(?P<resource>[^/]+)`, so non-snake-case paths also reach `UnknownResourceView`'s JSON 404.

apps/api/export/views.py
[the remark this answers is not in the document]
SmittieC: Done in 0c08b0e81 -- removed the unused `Export` tag from `SPECTACULAR_SETTINGS['TAGS']`. The endpoints stay on the Manifest/Team tags, and the `Export` tag was already pruned from every generated schema since no operation referenced it.

apps/api/export/views.py
```diff
@@ -0,0 +1,227 @@
...
+
+    @staticmethod
+    def decode(cursor: str) -> dict:
+        try:
+            keyset = json.loads(base64.b64decode(cursor))
+        except (ValueError, TypeError) as e:  # bad base64 (binascii.Error) or bad JSON
+            raise ValidationError("Invalid pagination cursor.") from e
+        if not isinstance(keyset, dict) or "updated_at" not in keyset or "id" not in keyset:
+            raise ValidationError("Invalid pagination cursor.")
+        return keyset
```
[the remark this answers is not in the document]
SmittieC: Fixed in 719ca20b0 -- `decode` now casts `keyset["id"]` to `int` and raises `ValidationError` on failure, so a non-integer id is a clean 400 rather than a 500 in the ORM filter. Added a `keyset-cursor-non-integer-id` case to `test_malformed_pagination_input_returns_400`.

apps/api/export/views.py
```diff
@@ -0,0 +1,227 @@
...
+            raise ValidationError("Invalid pagination cursor.")
+        queryset = queryset.filter(Q(updated_at__gt=timestamp) | Q(updated_at=timestamp, id__gt=keyset["id"]))
+    rows = list(queryset[: limit + 1])
+    has_more = len(rows) > limit
+    rows = rows[:limit]
+    if rows:
+        next_cursor = CursorHelper.encode({"updated_at": rows[-1].updated_at.isoformat(), "id": rows[-1].id})
+    else:
+        next_cursor = cursor
+    return rows, next_cursor, has_more
```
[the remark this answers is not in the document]
SmittieC: Fixed in 719ca20b0 -- both pagination branches now return `None` when `has_more` is false, matching the documented contract (`cursor` is null once the resource is exhausted). Added `test_cursor_is_null_once_the_resource_is_exhausted`.

apps/api/tests/test_schema.py
```diff
@@ -11,7 +11,7 @@ def test_schema_filters():
     assert "/cms/" not in response_yaml
 
 
-@pytest.mark.parametrize("version", ["v1", "v2"])
+@pytest.mark.parametrize("version", ["v1", "v2", "export"])
```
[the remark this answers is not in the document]
SmittieC: Fixed in a60f9dd3a -- each version now uses `pytest.param(..., id=...)`.

apps/teams/export/importer.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def _import_team_owned_row(
+        self, model_label: str, model: type[models.Model], source_pk: int, row: dict
+    ) -> tuple[models.Model, bool]:
+        with transaction.atomic():
+            field_values, m2m_values, timestamps = self._build_values(model_label, model, row)
+            instance, created = self._get_or_create(model_label, model, source_pk, row, field_values)
+
+            for name, target_pks in m2m_values.items():
+                getattr(instance, name).set([pk for pk in target_pks if pk is not None])
```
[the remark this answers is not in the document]
SmittieC: No change needed for the current model set. The only in-scope m2m is `teams.team.members`, which is in `EXCLUDE_REGISTRY` and handled via the user/membership path, so it never reaches `_build_m2m_values`. Every m2m that does reach it points at a deliberately-unsynced model -- `users.customuser.groups` (named-linked) and `user_permissions` (excluded), `pipelines.node.collection_indexes` -> `documents.collection`, `chat.chat.tags` -> `annotations.tag` -- where `get_target` returning `None` is by design, the same as out-of-scope FKs in `resolve_fk`. So no row is ever checkpointed with a partially-resolved *in-scope* m2m. I'll revisit with a defer/fail path if an in-scope m2m is ever added.

apps/teams/export/manifest.py
```diff
@@ -0,0 +1,162 @@
...
+def schema_checksum() -> str:
+    """Hash of the set of applied migration names; unaffected by apply order. Returns the hash of an
+    empty set when the migrations table is absent (e.g. tests run with --no-migrations); that's
+    consistent as long as the source and target both lack it."""
+    recorder = MigrationRecorder(connection)
+    names = []
+    if recorder.has_table():
+        names = list(recorder.migration_qs.order_by("name").values_list("name", flat=True))
+    data = json.dumps(names)
+    return hashlib.sha256(data.encode("utf-8")).hexdigest()
```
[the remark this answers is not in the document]
SmittieC: Fixed in 719ca20b0 -- `schema_checksum()` now hashes `(app, name)` pairs ordered by both, so migrations sharing a filename like `0001_initial` across apps no longer collide.

apps/teams/export/seal.py
```diff
@@ -0,0 +1,39 @@
...
+def _as_bytes(pem) -> bytes:
+    return pem.encode() if isinstance(pem, str) else pem
+
+
+def load_public_key(pem):
+    return serialization.load_pem_public_key(_as_bytes(pem))
+
+
+def load_private_key(pem, password=None):
+    return serialization.load_pem_private_key(_as_bytes(pem), password=password)
```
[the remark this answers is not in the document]
SmittieC: Fixed in 719ca20b0 -- `load_public_key`/`load_private_key` now reject non-RSA keys (and public keys under 2048 bits) with a `ValueError` at load time. Added `test_load_keys_reject_non_rsa` and `test_load_public_key_rejects_undersized_rsa`. (The live module already imports `Fernet`/`padding` correctly -- only the suggestion diff had the `ferkey`/`ferkey` typos.)

apps/teams/management/commands/sync_team.py
```diff
@@ -0,0 +1,147 @@
...
+        client = SourceClient(options["source_url"], options["api_key"])
+
+        run_sync(
+            client,
+            store,
+            private_key,
+            write=self.stdout.write,
+            page_limit=options["limit"],
+            enforce_schema=not options["skip_schema_check"],
+        )
```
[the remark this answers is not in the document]
SmittieC: Fixed in 719ca20b0 -- `handle()` now loads the private key and runs `check_sync_preconditions` (source reachable + schema match + secret/key check) *before* `--force-delete`, so a bad key path, unreachable source, or schema mismatch aborts while the existing team is still intact. Added `test_force_delete_is_not_reached_when_preconditions_fail`.

apps/teams/export/tests/test_manifest.py
```diff
@@ -0,0 +1,213 @@
...
+    assert not (in_manifest & EMBEDDED_MODELS)
+    assert not (IGNORED_MODELS & EMBEDDED_MODELS)
+
+
+def test_embedded_models_resolve_to_models():
+    for label in EMBEDDED_MODELS:
+        assert _model(label) is not None
+
+
+def test_membership_is_embedded_and_chat_is_a_standalone_resource():
```
SmittieC: this method needs updating. remove reference to the chat

apps/teams/export/tests/test_manifest.py
```diff
@@ -0,0 +1,213 @@
...
+    """Membership rides inside the user export (the team role); chat is its own synced resource."""
+    assert "teams.membership" in EMBEDDED_MODELS
+    assert manifest.get_manifest_entry("memberships") is None
+    assert manifest.get_manifest_entry("chats") is not None
+
+
+def test_chat_is_served_before_sessions():
```
SmittieC: remove this test

apps/teams/export/tests/test_manifest.py
```diff
@@ -0,0 +1,213 @@
...
+
+@pytest.mark.django_db()
+def test_schema_checksum_is_reproducible():
+    first = manifest.schema_checksum()
+    assert isinstance(first, str)
+    assert first == manifest.schema_checksum()
+
+
+@pytest.mark.django_db()
+def test_team_scoped_queryset_isolates_teams_and_includes_globals():
```
SmittieC, line 182: add a mini docstring

apps/api/export/tests/test_schema_hook.py
```diff
@@ -0,0 +1,112 @@
...
+    build_resource_response_serializer,
+    build_resource_serializer,
+    component_name,
+    resource_responses,
+)
+from apps.api.export.views import ResourceView, resource_view
+from apps.teams.export.manifest import MANIFEST_ENTRIES, build_manifest, entry_model, get_manifest_entry
+
+
+def test_each_resource_resolves_to_a_resourceview_carrying_its_name():
```
SmittieC, line 15: these tests should be moved to test_serializers.py?

apps/api/export/serializers.py
SmittieC: Fixed in 805e9f8c0 -- `build_resource_serializer` is now only used within the export surface, so I dropped the stale "v2 chatbot retrieve" reference. The caching rationale (one class per model for stable drf-spectacular component identity) still holds, so the cache stays.

apps/api/export/tests/test_views.py
SmittieC: Agreed -- done in f55d70d5c. `ManifestView` is now `AllowAny` with no authenticators, so the manifest (static resource list + schema checksum, no team data) is public. The team and resource endpoints stay behind IsAuthenticated + IsTeamAdmin. Replaced the two auth tests with `test_manifest_is_public_and_returns_entries`.

apps/teams/management/commands/sync_team.py
```diff
@@ -0,0 +1,161 @@
...
+    # dedicated ``team/`` endpoint rather than as a paginated resource in the manifest.
+    importer.import_rows(TEAM_MODEL, [client.get_team()])
+    write("synced team")
+    for entry in manifest["entries"]:
+        model_label, resource, cursor_type = entry["model"], entry["resource"], entry["cursor"]
+        model = entry_model(model_label)
+        cursor = _start_cursor(model_label, cursor_type, store, model)
+        importer.import_rows(model_label, client.iter_rows(resource, start_cursor=cursor, limit=page_limit))
+        write(f"synced {resource}")
+    return importer
```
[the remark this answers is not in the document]
SmittieC: The extra params on `run_sync` (`write`, `page_limit`, `enforce_schema`, `on_user_created`) are keyword-only injection seams with defaults so the command and the tests can drive the sync without patching globals; real callers pass `client`/`store`/`private_key`. Leaving as-is. No change needed.

apps/teams/export/tests/test_manifest.py
SmittieC: Done in b53bb658b -- the test is now `test_membership_is_embedded_not_a_standalone_resource` and only asserts the membership-embedding invariant; the chat assertion is gone.

apps/teams/export/tests/test_manifest.py
SmittieC: Removed in b53bb658b.

apps/teams/export/tests/test_manifest.py
SmittieC: Done in 752b41811 -- added a one-line docstring describing the team-isolation + globals invariant.

apps/api/export/tests/test_schema_hook.py
SmittieC: `test_schema_hook.py` actually spans three groups: URL/routing (this test plus `test_resource_view_class_name...` and `test_unknown_resource...`), serializer/response shape (`test_response_envelope...`, `test_secret_fields_documented...`, `test_manifest_serializer_matches...`), and generated-schema components (`test_every_export_resource_publishes...`, `test_export_detail_and_curated...`). The test here is a routing test, so it wouldn't fit `test_serializers.py`. I'd suggest keeping routing/schema tests here and, if anything, moving only the serializer-shape subset over -- happy to do that, but it's a judgment call on file layout so I've left it for you to decide rather than reshuffling unprompted.

apps/teams/export/tests/test_importer.py
```diff
@@ -0,0 +1,381 @@
...
+    """Re-importing the same row maps to the existing target instead of creating a duplicate."""
+    importer = Importer(store)
+    importer.import_rows("teams.team", [_team_row()])
+    first_pk = store.get_target("teams.team", 9001)
+    importer.import_rows("teams.team", [_team_row()])
+    assert store.get_target("teams.team", 9001) == first_pk
+    assert Team.objects.filter(slug="imported-team-xyz").count() == 1
+
+
+def test_global_row_matches_existing_and_is_not_recreated(store):
```
SmittieC: we need to test each entry in GLOBAL_CONFIG. For each entry, make sure there is a team scoped and global record, and assert that we only get a single global object back

apps/api/export/tests/test_schema_hook.py
SmittieC: Done in 6971961f0. Moved the serializer-shape subset — `test_response_envelope_has_cursor_has_more_and_results`, `test_secret_fields_documented_as_sealed_strings`, and `test_manifest_serializer_matches_build_manifest_payload` — to `test_serializers.py`. `test_schema_hook.py` keeps the routing tests and the generated-schema-component tests (`test_secret_resources_document_the_no_public_key_conflict`, `test_every_export_resource_publishes_its_row_component`, `test_export_detail_and_curated_serializer_live_in_separate_schemas`).

apps/teams/export/client.py
```diff
@@ -0,0 +1,64 @@
...
+
+class ResourceFetcher:
+    def __init__(self, base_url, api_key, *, timeout=30, max_retries=5, backoff=1.0, session=None, sleep=time.sleep):
+        self.base_url = base_url.rstrip("/")
+        self.api_key = api_key
+        self.timeout = timeout
+        self.max_retries = max_retries
+        self.backoff = backoff
+        self._session = session or requests.Session()
+        self._sleep = sleep
```
SmittieC: Fixed in fd1b80c47. `timeout`/`max_retries`/`backoff` were constructor kwargs but no caller ever overrode them (only `session`/`sleep` are injected, by tests), so they're now fixed instance attributes set in `__init__`. The signature is down to `base_url, api_key, *, session, sleep` — under the threshold.

apps/teams/export/tests/test_importer.py
SmittieC: Done in 81cdf5730. The test is now parametrized over `GLOBAL_CONFIG`, so every global model (`llmprovidermodel`, `embeddingprovidermodel`, `syntheticvoice`) is covered. For each, it creates both a team-scoped twin and a global record sharing the natural key, then asserts the import resolves to the single global record (not the team-scoped one) and recreates nothing. A new `GLOBAL_CONFIG` entry fails the test until a natural-key sample is added.

api-schemas/export.yml
```diff
@@ -0,0 +1,1945 @@
+openapi: 3.0.3
+info:
+  title: Open Chat Studio
+  version: export
+  description: Build, deploy and monitor chatbots.
```
snopoke: Is it possible to customise this for the export API?

apps/api/export/serializers.py
```diff
@@ -0,0 +1,164 @@
...
+        data = super().to_representation(instance)
+        public_key = self.context.get("public_key")
+        for field in self.secret_fields:
+            data[field] = seal(getattr(instance, field), public_key)
+        return data
+
+
+@extend_schema_field(serializers.ListField(child=serializers.CharField()))
+def _feature_flags(self, team):
+    from apps.teams.models import Flag  # noqa: PLC0415 - avoid import cycle at module load
```
snopoke: unlikely to actually cause a circular import

apps/api/export/views.py
```diff
@@ -0,0 +1,247 @@
...
+        return Response(build_team_serializer()(request.team).data)
+
+
+class ResourceView(_ExportAPIView):
+    # The per-resource OpenAPI schema is attached by ``resource_view`` (below), not here.
+    def get(self, request, resource):
+        entry = get_manifest_entry(resource)
+        if entry is None:
+            # Defense-in-depth: routing only mounts manifested resources, so this can only fire
+            # if the view is called directly (e.g. in tests) with a resource not in the manifest.
```
snopoke: I don't quite follow this comment - probably not worth keeping it.

apps/api/export/views.py
```diff
@@ -0,0 +1,247 @@
...
+            raise NotFound("Unknown resource.")
+
+        context = {"public_key": None, "team": request.team}
+        if entry.secret:
+            if not request.team.public_key:
+                return Response(
+                    {"detail": "Team has no registered public key; secret data cannot be sealed."},
+                    status=status.HTTP_409_CONFLICT,
```
snopoke: Is this the right response code? I think 400 would be better.

apps/api/export/views.py
```diff
@@ -0,0 +1,247 @@
...
+        rows = rows[:limit]
+        next_cursor = str(rows[-1].id) if has_more and rows else None
+        return rows, next_cursor, has_more
+
+    queryset = queryset.order_by("updated_at", "id")
+    if cursor:
+        keyset = CursorHelper.decode(cursor)
+        timestamp = parse_datetime(keyset["updated_at"])
+        if timestamp is None:
+            raise ValidationError("Invalid pagination cursor.")
```
snopoke: this could be in the CursorHelper as well (id validation is there already)

apps/api/export/serializers.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+@extend_schema_field(serializers.ListField(child=serializers.CharField()))
+def _team_role_groups(self, user):
+    """The user's role in the exported team: the group names on their membership for that team. Reads
+    the prefetched memberships, so no per-row query. Empty when no team is in the context."""
+    team = self.context.get("team")
+    team_id = team.id if team is not None else None
+    for membership in user.membership_set.all():
+        if membership.team_id == team_id:
```
snopoke, line 48: can't this filtering be pushed to the DB?

apps/teams/management/commands/sync_team.py
```diff
@@ -0,0 +1,161 @@
...
+
+        store = FKTranslationStore(Path(options["state_dir"]) / f"{options['team_slug']}.sqlite")
+        client = ResourceFetcher(options["source_url"], options["api_key"])
+        enforce_schema = not options["skip_schema_check"]
+
+        # Preflight before the destructive --force-delete: a bad key path (read above), an
+        # unreachable source, or a schema mismatch must fail while the existing team is still intact.
+        check_sync_preconditions(client, private_key, enforce_schema)
+
+        if options["force_delete"]:
```
snopoke, line 149: this should also prompt for confirmation before deleting

snopoke (dismissed):
This looks good. Some of the client code is hard to follow but otherwise fine. Most comments are minor.

api-schemas/export.yml
SmittieC: Yes — and it already was for the served \`/api/export/schema/\` via \`custom_settings\`. The committed \`export.yml\` showed the global default because the \`spectacular\` management command reads the global \`SPECTACULAR_SETTINGS\` and ignores per-view \`custom_settings\`. Fixed in 814683d05: moved the description into a \`set_export_description\` postprocessing hook keyed on \`api_version\`, so the served schema and the generated file stay in sync; dropped the duplicated \`custom_settings\`.

apps/api/export/serializers.py
SmittieC: Agreed — confirmed there is no cycle (verified \`apps.teams.models\` imports cleanly alongside the existing module-level export imports). Hoisted to module level and dropped the noqa in 069102e91.

apps/api/export/views.py
SmittieC: Removed in 069102e91.

apps/api/export/views.py
SmittieC: Switched to 400 in 814683d05 — updated the view, the documented per-resource response, and the regenerated \`export.yml\`.

apps/api/export/views.py
SmittieC: Done in 069102e91: moved the \`updated_at\` parse + null-check into \`CursorHelper.decode\` so the whole cursor is validated in one place; \`_paginate\` now consumes the already-parsed datetime. Added an unparseable-timestamp cursor test case.

apps/api/export/serializers.py
SmittieC: Good call — pushed it to the DB in d50c08019, via a team-scoped \`Prefetch\` in \`team_scoped_queryset\`: \`Prefetch("membership_set", queryset=Membership.objects.filter(team=team).prefetch_related("groups"))\`. So the DB only returns the exported team's membership (+ its role groups), no over-fetch.

A note on what I deliberately avoided: doing the filter in the resolver itself (\`user.membership_set.filter(team_id=...)\`) would be an N+1 — \`.filter()\` on a related manager bypasses the prefetch cache and re-queries per row (measured 2 queries/user, ~2000 for a 1000-row page). The team-scoped \`Prefetch\` keeps it at two queries per page. Kept the one-line \`team_id\` guard in the resolver so the field stays correct when a user is serialized outside the export queryset (the serializer unit tests do this). Added a multi-team correctness test and a no-N+1 query-count test.

apps/teams/management/commands/sync_team.py
SmittieC: Added in 5475003db: \`--force-delete\` now prompts for confirmation before deleting, with \`--no-input\`/\`--noinput\` to skip it for non-interactive (CI) runs. Covered by decline-aborts and no-input-skips tests.

snopoke (approved):

PR #3701, opened 2026-06-26
Migrate SureAdhere to v2 channels framework

snopoke:
### Product Description
No change. Internal refactor — SureAdhere messaging behaves identically.

### Technical Description
Moves SureAdhere onto the v2 channels framework (#3329), replacing the old `ChannelBase` subclass in `apps/chat/channels.py` with a pipeline-based channel in `channels_v2/`.

SureAdhere is text-only, so the new channel is minimal: a `SureAdhereSender` that delivers via the messaging service (`from_=tenant_id`, `to=recipient`), default no-op callbacks (no voice/transcription), and capabilities read from the messaging service. Both dispatch points (`handle_sureadhere_message` task and `get_channel_class_for_platform`) now resolve the v2 class.

### Migrations
N/A — no schema changes.

### Docs and Changelog

### DIFF
--- a/apps/channels/channels_v2/sureadhere_channel.py
+++ b/apps/channels/channels_v2/sureadhere_channel.py
@@ -1,0 +1,84 @@
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+from apps.channels.channels_v2.callbacks import ChannelCallbacks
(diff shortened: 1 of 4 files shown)


SmittieC (approved):
cool

PR #3705, opened 2026-06-29
feat: WhatsApp BSUID participant identity

### LINKED ISSUE #3322: WhatsApp BSUID / usernames support — webhook + send-message changes needed (Meta Cloud + Twilio)

SmittieC:
Supersedes #3397. Resolves #3322.

### Product Description
Adds WhatsApp Business-Scoped User ID (BSUID) support. Meta is rolling out BSUIDs — a stable per-business identifier sent on every webhook — and the phone number is becoming optional (omitted once a user adopts a WhatsApp username). Participants keep a stable identity, returning users stay linked to their history, and outbound messages still reach them.

### Technical Description
The identity model is captured in **ADR-0043**. The short version: the BSUID is the participant identifier; the phone number is a secondary attribute stored on the existing `Participant.remote_id` and used as the send recipient.

This supersedes #3397, which keyed participants by phone with the BSUID as a fallback. That direction was rejected because the phone is optional post-rollout and can change, so it isn't a stable key. Here the BSUID is the key and continuity for already-existing phone-keyed participants is handled at lookup time instead.

The change reads bottom-up across three layers, plus the ADR:

- **Parsing** — `WhatsAppMessage`/`TwilioMessage.parse` extract the BSUID (`from_user_id` / `ExternalUserId`) as `participant_id` and capture the phone separately. When no BSUID is present (Turn.io, pre-rollout traffic) `participant_id` falls back to the phone, as before.
- **Resolution** — `get_or_create_participant` matches on the BSUID **or** the inbound phone (oldest row wins), so a user previously keyed by phone is reused rather than forked onto a new BSUID row. The inbound phone is stored on `remote_id`.
- **Sending** — outbound goes to the stored phone, falling back to the identifier. The Meta payload routes a phone to `to` and a BSUID to `recipient`, so the path is ready for when providers accept BSUID recipients (neither Meta nor Twilio does yet).

Worth a closer look: the disjunctive (BSUID-or-phone) participant lookup and the oldest-wins tie-break in `get_or_create_participant` — that's the bit that preserves continuity without a data migration.

**Participant-handling refactor.** Alongside the feature work, session creation was tidied to match the identity model. `start_experiment_session` now takes a single `participant` argument — a `Participant` wrapper that need not be stored (an unstored `Participant(identifier=...)` is resolved/created internally) — instead of separate `participant_identifier`/`participant_user`/`participant` arguments, so one object carries everything we know about the participant. It was also promoted from the private `_start_experiment_session` in `chat.channels` to a public service at `apps/experiments/services.py` that every channel can reuse.

### Migrations
N/A — no schema change. The phone is stored on the pre-existing `remote_id` field.

### Demo
No UI change; exercised via WhatsApp webhook delivery. Covered by unit + integration tests across all three layers, including the end-to-end "existing phone-keyed participant is reused (not forked) when a BSUID-bearing message arrives" case for both Meta Cloud and Twilio.

### Docs and Changelog

Adds ADR-0043 (WhatsApp BSUID participant identity).


apps/channels/datamodels.py
```diff
@@ -149,6 +169,26 @@ def determine_content_type(cls, value):
...
+        return bsuid or None
+
+    @staticmethod
+    def get_phone_number(message_data: dict) -> str | None:
+        """Return the user's E.164 phone number from `From` for WhatsApp messages, else None."""
+        prefix, _, raw = message_data["From"].partition(":")
+        if prefix != "whatsapp" or looks_like_bsuid(raw):
+            return None
+        # Normalize to E.164 (the DB format); numbers usually arrive in this format already.
+        return phonenumbers.format_number(phonenumbers.parse(raw), phonenumbers.PhoneNumberFormat.E164)
```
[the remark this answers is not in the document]
SmittieC: Twilio will always have the BSUID in the `ExternalUserId` field. When the user has a phone number, both bsuid an phone number will exist. If the user only uses external id, the phone number only will fall away. This is a non-issue

apps/channels/tests/channels/stages/test_participant_resolution.py
```diff
@@ -69,3 +77,136 @@ def test_does_not_match_participant_on_different_platform(self):
...
+
+        participant = get_or_create_participant(
+            team=experiment.team,
+            normalized_identifier="bsuid_oldest_wins",
+            platform=ChannelPlatform.WHATSAPP,
+            participant_user=None,
+            participant_id_filter=Q(identifier="bsuid_oldest_wins") | Q(identifier="+15551112222"),
+        )
+
+        assert participant.id == oldest.id
```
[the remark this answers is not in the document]
SmittieC: Fixed in bce21b9dd. The disjunctive lookup now orders by `created_at, id`, so the oldest-wins reuse is deterministic even when two rows share a timestamp.

apps/service_providers/tests/test_meta_cloud_api_bsuid.py
```diff
@@ -0,0 +1,119 @@
...
+            from_="phone123",
+            to=self.BSUID,
+            platform=ChannelPlatform.WHATSAPP,
+            file=file,
+            download_link="https://example.com/x.jpg",
+            last_activity_at=timezone.now() - timedelta(hours=1),
+        )
+        sent = mock_post.call_args_list[1].kwargs["json"]
+        assert sent["recipient"] == self.BSUID
+        assert "to" not in sent
```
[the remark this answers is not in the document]
SmittieC: Fixed in bce21b9dd. The template, voice, and file BSUID tests now also assert `recipient_type == "individual"`, matching the text-path contract. (All four Meta send paths set it via the shared `_recipient_kwarg` helper.)

docs/adr/0043-whatsapp-bsuid-participant-identity.md
```diff
@@ -0,0 +1,59 @@
...
+## Context
+
+Meta is rolling out business-scoped user IDs (BSUIDs) for WhatsApp — a stable per-business
+identifier (e.g. `US.13491208655302741918`) sent on every post-rollout webhook. The phone
+number is now optional: Meta omits it once a user adopts a WhatsApp username, and Twilio
+likewise sends the BSUID (`ExternalUserId`) alongside the phone. Historically OCS keyed
+WhatsApp participants by phone number, so we need an identity model for the BSUID era that
+keeps a stable key, preserves continuity for users already keyed by phone, and can still
+address outbound messages — Meta and Twilio do not yet accept a BSUID as a send recipient.
+
```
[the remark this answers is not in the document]
SmittieC: No change needed — the ADR is internally consistent. The Decision section already documents the BSUID→`recipient` field routing as forward-compatible plumbing ("keeping the path forward-compatible for when Meta accepts BSUID recipients"). The Context/Consequences statements are about Meta *accepting* a BSUID as a deliverable recipient, which it still does not; that's why outbound resolution prefers the stored phone (`remote_id`). The test asserts the payload *shape*, not that Meta delivers to a BSUID-only participant.

apps/channels/channels_v2/stages/core.py
```diff
@@ -32,11 +40,64 @@
...
+                existing.save()
+            return existing
+
+    participant, created = Participant.objects.get_or_create(
+        team=team,
+        identifier=normalized_identifier,
+        platform=platform,
+        defaults={"user": participant_user},
+    )
+    if not created and participant_user and participant.user is None:
```
SmittieC: Won't change. The flagged conditional just backfills the user FK on an existing row (`if not created and participant_user and participant.user is None`) — a plain short-circuit guard that reads clearly. Splitting it to drop under the threshold would add indirection without improving readability.

apps/channels/channels_v2/stages/core.py
```diff
@@ -92,6 +153,15 @@ def process(self, ctx: MessageProcessingContext) -> None:
         except ParticipantData.DoesNotExist:
             ctx.participant_data = None
 
+    def _store_phone_number(self, ctx: MessageProcessingContext, phone_number: str | None) -> None:
+        """Persist the user's phone number on the participant's remote_id so it can be used as
+        the send recipient -- the participant identifier may be a (non-sendable) BSUID."""
+        participant = ctx.participant
+        if not phone_number or participant is None or participant.remote_id == phone_number:
```
SmittieC: Won't change. This is a standard early-return guard (`if not phone_number or participant is None or participant.remote_id == phone_number`) that skips the write when there's nothing to persist. Refactoring purely to satisfy the threshold would reduce clarity.

apps/channels/channels_v2/stages/core.py
```diff
@@ -32,11 +40,64 @@
...
+    participant, created = Participant.objects.get_or_create(
+        team=team,
+        identifier=normalized_identifier,
+        platform=platform,
+        defaults={"user": participant_user},
+    )
+    if not created and participant_user and participant.user is None:
+        participant.user = participant_user
+        participant.save()
+    return participant
```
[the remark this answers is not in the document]
SmittieC: Won't change. All five arguments are meaningful and passed by keyword at the call site (team, normalized_identifier, platform, participant_user, participant_id_filter). Bundling them into a params object would obscure rather than clarify.

apps/channels/channels_v2/whatsapp_channel.py
```diff
@@ -25,6 +25,16 @@
 logger = logging.getLogger("ocs.channels")
 
 
+def _resolve_recipient(ctx: MessageProcessingContext | None, fallback: str) -> str:
+    """Return the WhatsApp send recipient: the phone number stored on the participant's
+    remote_id if we have one, else the given identifier. The participant identifier may be a
+    BSUID, which Meta/Twilio cannot yet accept as an outbound recipient, so we prefer the phone
+    captured on inbound."""
+    if ctx is not None and ctx.participant is not None and ctx.participant.remote_id:
```
SmittieC: Won't change. The guard (`if ctx is not None and ctx.participant is not None and ctx.participant.remote_id`) is a simple presence check before falling back to the passed identifier. It reads clearly as-is.

SmittieC: Fixed in latest push: flipped to an early return so the recipient lookup is a guard (`if ctx is None or ctx.participant is None: return fallback`) plus a single `return ctx.participant.remote_id or fallback`, removing the 3-term conditional.

apps/channels/channels_v2/stages/core.py
SmittieC: Fixed in latest push: split the 3-term guard into two early returns (`not phone_number or participant is None`, then `participant.remote_id == phone_number`), so neither conditional exceeds the threshold.

apps/channels/channels_v2/stages/core.py
SmittieC: Fixed in latest push: extracted the duplicated 'backfill user on first contact' logic into a `_associate_user` helper. The flagged `not created and participant_user and participant.user is None` becomes `if not created: _associate_user(...)`, removing the complex conditional and the duplication.

apps/channels/channels_v2/stages/core.py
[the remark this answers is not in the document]
SmittieC: Leaving as-is (1 over threshold). The five params (team, normalized_identifier, platform, participant_user, participant_id_filter) are each distinct lookup inputs; bundling them into a params object would add indirection without a readability gain.

apps/service_providers/tests/test_messaging_providers.py
```diff
@@ -624,6 +624,7 @@ def test_send_text_within_window_sends_normal(self, mock_post):
         )
         assert mock_post.call_args.kwargs["json"] == {
             "messaging_product": "whatsapp",
+            "recipient_type": "individual",
```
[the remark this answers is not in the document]
SmittieC: Leaving as-is: a test module 3 lines over the 600 threshold. Splitting the file purely to shed 3 lines isn't worth the navigation cost.

apps/experiments/services.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # Record the participant's timezone
+        if timezone:
+            participant.update_memory(data={"timezone": timezone}, experiment=working_experiment)
+
+    if experiment_channel.platform != ChannelPlatform.EVALUATIONS:
+        if participant.experimentsession_set.filter(experiment=working_experiment).count() == 1:
+            enqueue_static_triggers.delay(session.id, StaticTriggerType.PARTICIPANT_JOINED_EXPERIMENT)
+        enqueue_static_triggers.delay(session.id, StaticTriggerType.CONVERSATION_START)
+    return session
```
SmittieC, line 82: Leaving as-is: this is pre-existing code relocated from `apps/chat/channels.py` (`_start_experiment_session`) into the new `services.py` module — flagged as 'new' only because the file is new. The cyclomatic complexity predates this PR; a deeper refactor is out of scope here.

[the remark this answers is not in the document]
SmittieC: Leaving as-is: relocated pre-existing code. This PR actually *reduced* the arg count from 8 → 7 by consolidating `participant_identifier`/`participant_user` into a single `participant` param. Further reduction (options object) is out of scope for this PR.

apps/channels/tests/test_datamodels.py
```diff
@@ -55,3 +62,84 @@ def test_false_for_conversational(self, message_data):
...
+        ("US.has-dash", False),  # Non-alphanumeric in tail.
+        ("US." + "a" * 129, False),  # Tail exceeds 128 chars.
+        ("US.", False),  # Empty tail.
+        ("27456897512", False),  # Plain wa_id.
+        ("+27456897512", False),  # E.164.
+        ("", False),
+    ],
+)
+def test_looks_like_bsuid(value, expected):
+    assert looks_like_bsuid(value) is expected
```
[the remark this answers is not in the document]
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/3705/commits/7d1c6b99bcab552b7af177b6de1982b42bb29c04

apps/channels/channels_v2/stages/core.py
```diff
@@ -75,15 +140,15 @@ class ParticipantResolverStage(ProcessingStage):
     def process(self, ctx: MessageProcessingContext) -> None:
         normalized = ctx.experiment_channel.platform_enum.normalize_identifier(ctx.participant_identifier)
         participant_user = ctx.channel_context.get("participant_user")
-        ctx.participant, created = Participant.objects.get_or_create(
+        phone_number = getattr(ctx.message, "phone_number", None)
```
snopoke: Just thinking of how to make this more generic:

Option 1. Override the stage just for whatsapp channels

Option 2. Create a 'remote_id' support feature: allow passing 'remote_id_field_name' to the stage on creation (in this case it would be 'phone_number'). If that field name is set then we also filter by remote ID```python
no_remote_id_support = ParticipantResolverStage()
with_remote_id_support = ParticipantResolverStage(remote_id_field_name="phone_number")
```

File: apps/experiments/services.py [hunk 13]
```diff
@@ -0,0 +1,88 @@
...
+
+    team = working_experiment.team
+    participant_identifier = participant.identifier
+    participant_user = participant.user
+    if not participant_identifier and not participant_user:
+        raise ValueError("Either participant_identifier or participant_user must be specified!")
+
+    if participant_user and participant_identifier != participant_user.email:
+        # This should technically never happen, since we disable the input for logged in users
+        raise Exception(f"User {participant_user.email} cannot impersonate participant {participant_identifier}")
```
snopoke: These checks feel like they could be skipped if the passed in participant is already saved in the DB. Maybe they should be pushed into `get_or_create_participant`

snopoke (dismissed):
nothing blocking

apps/channels/channels_v2/stages/core.py
SmittieC: Thoughts on 77f30ab72?

apps/experiments/services.py
SmittieC: Done in 1a1b1cf7e — moved the identifier/user validation into `get_or_create_participant`, which only runs on the create path (`participant.pk is None`). A stored participant is now trusted and skips the redundant checks.

snopoke (approved):

PR #3720, opened 2026-06-30
feat: add team files zip export endpoint

SmittieC:
### Product Description

Adds a **Download team files** button to the Team Management page (admin-only). Clicking it starts a background job that zips up every current file belonging to the team. A progress bar tracks the job and, once it finishes, the page shows a link to download the generated archive. If an export is already running for the team, the page resumes tracking that one instead of starting a second.

Also tightens the team-deletion flow: the confirmation modal now requires typing the team's name before the delete button becomes active.

### Technical Description

**Team files export (async).** Replaces the earlier synchronous `zipstream-ng` streaming response with a Celery task, `create_team_files_zip_task` (`apps/teams/tasks.py`):

- Builds a zip of the team's current files (stdlib `zipfile`, streamed to a temp file in chunks) and stores it as a `DATA_EXPORT` `File` with a 24h expiry. The finished archive is served through the normal `FileView` pre-signed URL.
- Reports progress via `celery-progress`; the `download_files_progress.html` partial polls until completion, then swaps in the download link.
- Archive entry paths mirror each file's storage path (`file.file.name`, optionally prefixed by the storage `location`) so the archive can be re-imported into a storage backend with an identical layout.
- The queryset excludes archived files, versioned copies, and previously generated `DATA_EXPORT` artifacts, so export zips don't nest inside future export zips.
- Files with no content, missing from storage, or that fail to read are skipped (and logged) rather than aborting the whole export.

Two new `Team` fields track state: `files_export_task_id` (id of the in-flight task) and `files_export` (FK to the most recently generated export `File`, `on_delete=SET_NULL`). `start_team_files_export` reuses an in-flight export if one exists rather than dispatching a duplicate — a lightweight check-then-set rather than a hard lock, since two concurrent exports for the same team is a fringe case not worth guarding against atomically.

**`is_team_admin` fix.** `request.team_membership.is_team_admin` was read as an attribute (a bound-method reference, always truthy) instead of being called as `is_team_admin()`, so every team member was treated as an admin in the `manage_team` view. It's now called correctly in both places.

**Delete-team confirmation.** The delete modal (`delete_team_modal.html`) now requires typing the team name (Alpine.js `x-model`) before the delete button is enabled.

### Migrations

`teams/0013` adds two additive fields — `files_export` (nullable FK, `SET_NULL`) and `files_export_task_id` (blank `CharField`, default `""`). Both are nullable/defaulted, so no backfill is needed.

### Demo
https://www.loom.com/share/0c30da7d03374d41ae76c231daf52fec

#### My other small change: Deleting a team will show a model like this (test1 is the team's name):
image

### Docs and Changelog


snopoke (commented):
My main concern with this is that if the download is interrupted part way there is no recovery.

Another option would be to generate the ZIP and then create a S3 pre-signed URL (same as we do for data exports). Then we don't need to manage the download at all.

apps/teams/tests/test_file_export.py
```diff
@@ -0,0 +1,234 @@
...
+
+@pytest.fixture()
+def admin(team):
+    user = CustomUser.objects.create(username="admin@acme.com")
+    make_user_team_owner(team, user)
+    return user
+
+
+@pytest.fixture()
+def member(team):
```
SmittieC, line 26: You can use the team_with_users factory

apps/teams/tests/test_file_export.py
```diff
@@ -0,0 +1,234 @@
...
+        ],
+    )
+    def test_button_visibility(self, request, client, team, user_fixture, should_see):
+        client.force_login(request.getfixturevalue(user_fixture))
+        response = client.get(_manage_team_url(team))
+        assert (_download_files_url(team) in response.content.decode()) is should_see
+
+
+@pytest.mark.django_db()
+class TestManageTeamExportState:
```
SmittieC: remove this class's tests

apps/teams/tests/test_file_export.py
SmittieC: Done in 3cbd8ee — replaced the hand-rolled `team`/`admin`/`member` fixtures with `TeamWithUsersFactory`, deriving admin/member from the created memberships.

apps/teams/tests/test_file_export.py
SmittieC: Removed `TestManageTeamExportState` in 3cbd8ee.

apps/teams/tasks.py
```diff
@@ -32,3 +47,102 @@ def delete_team_async(team_id, user_email, notify_recipients="self"):
...
+
+    Returns the id of the created File so the caller can serve it later (via a
+    pre-signed URL). Files with no stored content, missing from storage, or that
+    fail to read are skipped rather than aborting the whole export. Regardless of
+    outcome, clears the in-progress export marker (see
+    Team.mark_files_export_finished) so a new export can be started.
+    """
+    progress_recorder = ProgressRecorder(self)
+    team = Team.objects.get(id=team_id)
+    files = list(get_team_files_queryset(team))
```
snopoke: do we need to load these all into memory?

api-schemas/export.yml
```diff
@@ -5470,6 +5470,14 @@ components:
...
         metadata:
           description: Internal staff-only metadata for this team.
+        files_export_task_id:
+          type: string
+          description: Celery task id of an in-progress team-files export, if any.
+          maxLength: 255
+        files_export:
+          type: integer
+          nullable: true
+          description: Most recently generated team-files export, if any.
```
snopoke: Is this addition intentional? Doesn't seem useful unless you can access the progress / file via the API

SmittieC: No, I wanted to update this, but forgot to. See https://github.com/dimagi/open-chat-studio/pull/3720/commits/1f37082133938a9b2a08ddc0d27ce4724e74dbd0 for the fix.

apps/teams/tasks.py
SmittieC: Good catch. Optimized here https://github.com/dimagi/open-chat-studio/pull/3720/commits/a9f0a9dfc45429bc98c552b8537fa29730c552b5

snopoke (approved):

PR #3725, opened 2026-06-30
Cost tracking UX improvements: pricing display, deprecated models, source badge

### LINKED ISSUE #3721: Cost tracking UX improvements: pricing display, deprecated models, dashboard detail
A collection of UX improvements for the cost tracking feature, from user feedback.

## 1. Display rates in $/1M on the LLM provider page

Currently `llm_model_row.html` shows rates as `$0.00300/1K` (5 decimal places, per-1K). This is not how model pricing is conventionally quoted — the industry standard is per-million tokens (e.g. `$3.00/1M`).

The override form (`pricing_override_form.html`) already uses per-1M in its labels, so this is an inconsistency between the display and the form.

**Fix:** In `llm_model_row.html`, multiply `unit_price` by 1000 and display with 2 decimal places:
```
in ${{ rates.llm_input.unit_price|multiply:1000|floatformat:2 }}/1M
```
(or add a `per_million` template filter to `cost_tracking` tags)

## 2. Disable pricing edits for deprecated models

`LlmProviderModel` has a `deprecated` boolean field (shown as `(Deprecated)` in `display_name`). Setting a pricing override on a deprecated model is pointless and may create confusion.

**Fix:** In `llm_model_row.html`, hide or disable the pencil/"Set pricing" button when `model.deprecated` is true. The existing revert button should similarly be suppressed.

## 3. Only show the source badge when it's not the default (`seed`)

Every model row shows a `seed` / `manual` / `import` badge via `rates.primary.source`. The `seed` value is the default for almost every model and adds visual noise without useful information.

**Fix:** Only render the badge when the source is _not_ `seed` (i.e. when it's `manual` or `import`, or when there's a team override):
```django
{% if rates.primary.source != 'seed' or rates.primary.scope == 'team' %}
  <span class="badge badge-xs badge-outline ml-1">...</span>
{% endif %}
```

_Dashboard panel improvements have been split into a separate issue: #3723_

snopoke:
### Product Description
UX improvements for cost tracking on the LLM provider model list (#3721):

- Rates now display per 1M tokens (e.g. `$3.00/1M`), matching the override form and industry convention, instead of per-1K with 5 decimals.
- The pricing edit / revert / "Set pricing" buttons are hidden for deprecated models.
- The source badge only shows when the source isn't the default `seed` (i.e. `manual`, `import`, or a team override).
- Sort models by date created (descending) with deprecated models at the end.
- Use a more muted color for the 'No pricing configured' text.

### Technical Description
Adds a `per_million` template filter to `cost_tracking` tags and updates `llm_model_row.html`. Render-level tests cover each behaviour.

### Migrations

### Demo

**light mode**
Screenshot from 2026-06-30 17-16-43

image

**dark mode**
image

image

### Docs and Changelog

Closes #3721

### DIFF
--- a/apps/service_providers/tests/test_pricing_override.py
+++ b/apps/service_providers/tests/test_pricing_override.py
@@ -1,6 +1,7 @@
 """Tests for the override / revert / create-with-pricing HTMX flow on
 the LLM provider model list."""
 
+from datetime import UTC, datetime
 from decimal import Decimal
 
 import pytest
@@ -9,6 +10,7 @@ from django.urls import reverse
 
 from apps.cost_tracking.models import PricingRule, PricingSource, ServiceKind
 from apps.service_providers.models import LlmProviderModel
+from apps.service_providers.views import _get_models_by_type
 from apps.teams.models import Flag
 from apps.utils.factories.team import TeamWithUsersFactory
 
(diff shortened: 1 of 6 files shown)


apps/service_providers/tests/test_pricing_override.py
```diff
@@ -187,3 +188,105 @@ def test_pricing_fields_ignored_when_flag_off(self):
...
+        team = TeamWithUsersFactory.create()
+        model = _custom_model(team)
+        model.deprecated = True
+
+        html = _render_row(model, None)
+
+        assert "Set pricing" not in html
+        assert "No pricing configured" in html
+
+    def test_deprecated_model_name_muted(self):
```
SmittieC: low value test?  Same with test_unpriced_uses_yellow_only_when_not_deprecated

SmittieC (dismissed):

apps/service_providers/tests/test_pricing_override.py
snopoke: Yup, I'm still making tweaks. I've made it a draft PR

SmittieC (approved):

PR #3748, opened 2026-07-01
Sync: re-register team webhooks after sync; exclude data-export files from export API

SmittieC:
### Product Description
Two improvements to the team data sync (export/import between servers):

1. **Webhook re-registration after sync** — after a team is synced onto this server, every one of that team's channel webhooks is repointed at this server, not just the channels a given run happened to create. This makes sync robust to interruption: a run that dies partway through no longer leaves channels pointing at the wrong server.
2. **Data-export files are no longer synced** — transient data-export download bundles (the ZIPs produced by "Download files", 24h expiry) are excluded from the export API so they aren't replicated to another server.

### Technical Description
- `sync_team` command now re-registers webhooks for all of the synced team's channels rather than only newly-created ones. Re-registering an already-correct webhook is harmless/idempotent.
- Export sync: added a `files.file` entry to `EXTRA_FILTERS` in `apps/teams/export/manifest.py` — `~Q(purpose=FilePurpose.DATA_EXPORT)` — applied in `team_scoped_queryset` after team scoping. Data-export files have no synced model referencing them, so excluding them dangles nothing on the target. Filtering happens at the queryset level rather than the serializer (serializers can't drop rows).
- Tests: added `test_files_queryset_excludes_data_export_files` (mirrors the pricing-rules exclusion test) and expanded `test_command.py` webhook coverage.

### Demo
image

image

### Docs and Changelog


SmittieC:
Fixed in b6a876cc8. `_channel_label` now falls back to `channel.name` when the channel has no linked experiment (team-global web/API/evaluations channels), and the misleading docstring is corrected. Added a regression test (`test_reregister_webhooks_labels_channel_without_experiment_by_name`) that reproduced the `AttributeError` before the fix.

snopoke (commented):
I think this should be flagged as a specific step so that it can be done independently. 

It should also confirm that the Site URL is correct before doing the update.

apps/teams/export/tests/test_command.py
```diff
@@ -281,3 +284,16 @@ def test_serialized_row_round_trips_through_importer(tmp_path, keypair):
     imported = LlmProvider.objects.get(pk=store.get_target("service_providers.llmprovider", provider.id))
     assert imported.team_id == target_team.id  # assigned from the synced team, not carried in the row
     assert imported.config == {"api_key": "sk-live"}
+
+
+def test_report_points_to_reregister_webhooks_command():
```
snopoke: not a useful test

snopoke (dismissed):

PR #3749, opened 2026-07-02
Extend object-reference checks to all versions and add bulk-archive action

### LINKED ISSUE #3197: Extend object-reference checks to non-published/non-working versions and add bulk-archive action
## Background

When checking whether an object (e.g. a `CustomAction`) can be safely deleted, the current guard only considers working versions and published versions. Discussion on PR #3187 concluded that **all** versions (i.e. those that are not working and not the default) must be considered.

Backlink: https://github.com/dimagi/open-chat-studio/pull/3187#discussion_r3117628439
Requested by: @SmittieC

---

### 1. Extend reference checks to all versions
- Non-published versions that reference the object should also be detected and reported in the modal.
- Deletion must be **blocked** when any version (working, published, or otherwise) references the object.
- Archived versions can safely be ignored, as is the case currently

### 2. Bulk-archive button for non-working / non-published versions
- Add a button on the "Cannot delete" modal that allows the user to **bulk-archive** all non-working, non-published (non default) versions that reference the object.
- This reduces the manual effort when many such versions exist.

### 3. Manual cleanup for working + published versions
- Working versions and published versions that reference the object must be cleaned up **manually, one-by-one** — no bulk action should be offered for these.
- The modal should clearly indicate which references require manual intervention.

SmittieC:
### Product Description
When deleting an object (service provider, assistant, pipeline, collection), deletion is now blocked when **any** non-archived version — not just working/published — references it. The "Cannot delete" modal now offers a one-click **bulk-archive** button for non-working, non-published snapshot versions, while working and published versions must still be cleaned up manually.

Resolves #3197. Ports #3198 (branch claude/issue-3197-20260422-0552) onto current main. Closes #3198.

### Technical Description
- Added `is_blocking_object` / `is_bulk_archiveable` helpers in `apps/utils/deletion.py`.
- `get_related_experiments*` querysets (assistants, documents) now return all non-archived versions.
- Delete views categorise references into manual-only vs bulk-archiveable.
- New `experiments:bulk_archive_versions` endpoint archives non-working, non-published versions atomically (team-scoped).
- Extended `render_referenced_objects_modal` helper + `referenced_objects.html` for the bulk-archive section (threads request for CSRF).
- Added 12 tests.

### Migrations
- [x] The migrations are backwards compatible (no migrations needed — query-only changes)

### Docs and Changelog


apps/assistants/views.py
```diff
@@ -228,17 +229,34 @@ def delete(self, request, team_slug: str, pk: int):
...
+                    label=(
+                        f"{e.name} [{e.get_version_name()}]"
+                        if e.is_working_version
+                        else f"{e.name} {e.get_version_name()} [published]"
+                    ),
+                    url=e.get_absolute_url(),
                 )
-                for experiment in assistant.get_related_experiments_with_pipeline_queryset(assistant_ids=version_query)
+                for e in all_experiments
+                if not is_bulk_archiveable(e)
```
SmittieC: this is a common pattern I see. We iterate over a queryset and split it out into bulk archivable and non-bulk archivable entries. It would be great if we can call `_get_referenced_objects_context` with a queryset that returns the necessary context

SmittieC: Done in ffd5fdcc6. Extracted `get_referenced_experiment_context(experiments, team_slug)` in `apps/generics/referenced_objects.py`: it takes the referencing experiments and returns the manual/bulk-archiveable chip split, the bulk-archiveable IDs, and the bulk-archive URL in a single pass. The assistant, pipeline, collection and service-provider delete views now all call it instead of duplicating the split.

apps/assistants/views.py
```diff
@@ -228,17 +229,34 @@ def delete(self, request, team_slug: str, pk: int):
...
             ]
             return render_referenced_objects_modal(
                 "assistant",
+                request=request,
                 pipeline_nodes=pipeline_nodes,
-                experiments_with_pipeline_nodes=experiments_with_pipeline_nodes,
+                experiments_with_pipeline_nodes=manual_experiments,
+                bulk_archiveable_experiments=bulk_archiveable_experiments,
+                bulk_archiveable_ids=[e.id for e in all_experiments if is_bulk_archiveable(e)],
+                bulk_archive_url=reverse("experiments:bulk_archive_versions", args=[team_slug]),
```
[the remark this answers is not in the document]
SmittieC: Fixed in ffd5fdcc6. Extracted the manual/bulk-archiveable split into `get_referenced_experiment_context`, which classifies the experiments in a single pass (so `is_bulk_archiveable` is evaluated once per experiment) and returns the manual chips, bulk chips, bulk-archiveable IDs and the bulk-archive URL. All four delete views now call it. Placed it in `apps/generics` alongside `render_referenced_objects_modal` (rather than `apps/utils/deletion`) since it builds `Chip`s for that modal.

apps/assistants/views.py
```diff
@@ -228,17 +228,18 @@ def delete(self, request, team_slug: str, pk: int):
...
+            all_experiments = list(
+                assistant.get_related_experiments_with_pipeline_queryset(assistant_ids=version_query)
+            )
+            experiment_context = get_referenced_experiment_context(all_experiments, team_slug)
             return render_referenced_objects_modal(
                 "assistant",
+                request=request,
                 pipeline_nodes=pipeline_nodes,
-                experiments_with_pipeline_nodes=experiments_with_pipeline_nodes,
+                experiments_with_pipeline_nodes=experiment_context.manual,
```
SmittieC, line 239: rather do `**experiment_context(...)`

apps/experiments/views/experiment.py
```diff
@@ -876,6 +876,32 @@ def archive_experiment_version(request, team_slug: str, experiment_id: int, vers
...
+    """
+    Archives multiple non-working, non-published experiment versions by ID.
+    Used from the "Cannot delete" modal to clear bulk-archiveable references.
+    """
+    version_ids = request.POST.getlist("version_ids")
+    if not version_ids:
+        return HttpResponse(status=400)
+    experiments = list(
+        Experiment.objects.get_all().filter(
+            id__in=version_ids,
```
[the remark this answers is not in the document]
SmittieC: Fixed in 036efe26b. `version_ids` are now coerced with `int(...)` up front, and any non-numeric value returns `HttpResponse(status=400)` instead of letting the ORM raise `ValueError` during query construction.

apps/service_providers/views.py
```diff
@@ -118,28 +114,40 @@ def delete_service_provider(request, team_slug: str, provider_type: str, pk: int
...
+            for e in experiment_objects
+            if is_bulk_archiveable(e)
         ]
         related_assistants = [
-            Chip(label=assistant.name, url=assistant.get_absolute_url())
-            for assistant in [obj for obj in filtered_objects if isinstance(obj, OpenAiAssistant)]
+            Chip(label=assistant.name, url=assistant.get_absolute_url()) for assistant in assistant_objects
         ]
-        if related_experiments or related_assistants:
+        if manual_experiments or bulk_archiveable_experiments or related_assistants:
```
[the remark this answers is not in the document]
SmittieC: Fixed in 259093dbd. Added a catch-all guard so any non-archived related object the modal can't render still blocks deletion:

```python
handled_types = (Experiment, OpenAiAssistant, Pipeline)
other_blocking = [obj for obj in blocking_objects if not isinstance(obj, handled_types)]
...
has_blocking_references = bool(
    experiment_objects or assistant_objects or standalone_pipelines or other_blocking
)
```

Previously an unhandled blocking type (e.g. a channel's `messaging_provider`, SET_NULL, or a collection's `auth_provider`, PROTECT) fell through to `service_config.delete()` — silently deleting a provider still in use, or raising `ProtectedError` (500). Pipelines are excluded from the catch-all because they're already fully accounted for above (resolved to experiments, kept as standalone working pipelines, or intentionally dropped when a dormant version). Added a regression test, `test_delete_blocked_by_unmodelled_blocking_reference`.

Note: for these unrenderable types the modal has no dedicated section yet, so it blocks the delete without listing them by name. Surfacing them as chips can be a follow-up if we want the nicer UX.

apps/service_providers/views.py
```diff
@@ -118,28 +115,60 @@ def delete_service_provider(request, team_slug: str, provider_type: str, pk: int
...
         related_assistants = [
-            Chip(label=assistant.name, url=assistant.get_absolute_url())
-            for assistant in [obj for obj in filtered_objects if isinstance(obj, OpenAiAssistant)]
+            Chip(label=assistant.name, url=assistant.get_absolute_url()) for assistant in assistant_objects
+        ]
+        pipeline_chips = [
+            Chip(label=pipeline.name, url=pipeline.get_absolute_url()) for pipeline in standalone_pipelines
         ]
-        if related_experiments or related_assistants:
+        if experiment_objects or assistant_objects or standalone_pipelines:
```
SmittieC: Addressed in 259093dbd. The guard is now extracted into a single named predicate rather than an inline multi-branch conditional:

```python
has_blocking_references = bool(
    experiment_objects or assistant_objects or standalone_pipelines or other_blocking
)
if has_blocking_references:
    ...
```

apps/assistants/views.py
SmittieC: Done in e5f36b84d. Added `ReferencedExperimentContext.bulk_archive_kwargs()`, which returns the three bulk-archive kwargs as a dict so every call site spreads them with `**experiment_context.bulk_archive_kwargs()`. The `manual` chips stay explicit because call sites map them to different modal sections (`experiments` vs `experiments_with_pipeline_nodes`). Applied across all four delete views (assistant, pipeline, collection, service provider).

apps/service_providers/views.py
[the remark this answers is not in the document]
SmittieC: Follow-up in 8862f4d9c: the catch-all needed refining. `SyntheticVoice.voice_provider` is `CASCADE`, so a voice provider's synced voices are deleted *with* the provider — they signal ownership, not external use. The first cut of the guard treated them as blocking and would have made any VoiceProvider with team-scoped voices (ElevenLabs/Intron/OpenAIVoiceEngine) un-deletable. Fixed by excluding CASCADE-owned model classes (`get_cascade_owned_models`): only references that survive the delete (SET_NULL/PROTECT/RESTRICT) block it. Added `test_delete_allowed_with_only_cascade_owned_references` to cover it.

For the record, the full set of provider referrers beyond experiments/assistants/pipelines: `Collection`, `TranscriptAnalysis`, `ExperimentChannel`, `CustomAction`, `DocumentSource` (PROTECT), `McpServer` (all SET_NULL/PROTECT → now blocked) and `SyntheticVoice` (CASCADE → not blocked).

PR #3751, opened 2026-07-02
feat: freeze scheduled messages during data migration + sync_team improvements

SmittieC:
### Product Description
Adds a per-team **migration mode** used when a team's data is being migrated to another server. While a team is in migration mode, its outbound message firing is frozen and a banner warns users not to create or edit chatbots. The team's export API is only usable while migration mode is on.

### Technical Description
Stacked on `cs/webhook_reregistration` (its base). Introduces `Team.is_migrating` and gates behaviour off it:

- **Firing freeze** — `events/tasks.py` excludes migrating teams from static triggers, timeout triggers, and scheduled messages via the new `migrating_team_ids()` helper (`apps/teams/export_service.py`). This stops the source server from firing messages that the destination will also fire once the migration completes.
- **Export API gated on migration mode** — new `TeamIsMigrating` DRF permission on the export views, so a team can only be exported while it's explicitly in migration mode.
- **UI** — a `set_migration_lock` view + `TeamMigrationForm` toggle on the manage-team page, and a warning banner shown app-wide when `request.team.is_migrating`. `is_migrating` is added to the audited `TEAM_FIELDS`.
- **sync_team command** — reuses an already-synced team on rerun, gives friendly messages for known source failures, and reports sync duration.
- Minor: regenerate tag slugs on import; fetch only needed columns in the widget-notification commands; extract the missing-public-key message into `MISSING_PUBLIC_KEY_DETAIL`.

Note the base branch has since advanced past this branch, so the PR diff is scoped to the changes above (base-side dependency bumps / merges are not part of it).

### Migrations

`0013_team_is_migrating` adds a `BooleanField(default=False)` — additive, safe to apply while old code runs.

### Docs and Changelog


apps/teams/management/commands/sync_team.py
```diff
@@ -74,6 +111,32 @@ def _style_synced_line(message: str, count: int, style) -> str:
...
+        return
+
+    team_row = client.get_team()
+    if Team.objects.filter(slug=team_row["slug"]).exists():
+        raise CommandError(
+            f"A team with slug '{team_row['slug']}' already exists locally but isn't tracked by this "
+            "sync's state. Re-run with --force-delete to delete it and re-import from scratch."
+        )
+    importer.import_rows(TEAM_MODEL, [team_row])
+    write("synced team")
```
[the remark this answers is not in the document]
SmittieC: Fixed in 95d74e041. `load_team` now resolves the committed team via a `_committed_team` helper that raises a `CommandError` pointing at `--force-delete` when the sync state references a team that no longer exists locally, with a regression test covering the rerun-after-delete case.

*— Claude Code*

apps/teams/management/commands/sync_team.py
```diff
@@ -6,15 +6,19 @@
 translation store. Each run makes one pass over the manifest and exits; rerun to pick up new data.
 """
 
+import time
```
SmittieC, line 9: Addressed in e28ff3404. Merged the two near-identical friendly HTTP error helpers into one table-driven lookup and extracted the force-delete confirmation block out of `handle()`, lowering the module's mean cyclomatic complexity.

*— Claude Code*

apps/api/export/permissions.py
```diff
@@ -5,3 +5,11 @@ class IsTeamAdmin(BasePermission):
     def has_permission(self, request, view):
         membership = getattr(request, "team_membership", None)
         return bool(membership and membership.is_team_admin())
+
+
+class TeamIsMigrating(BasePermission):
```
snopoke: Is this really necessary? It means the API's can't be used at all except for migrations.

SmittieC: Not strictly, but only for redundancy. What we can perhaps do instead, is make a new "pre-export" endpoint to be checked by the management command that checks if the team 1. is in migrate mode and 2. has a public key configured. Or do you think we should drop the check instead?

snopoke: I think we should drop this restriction and add the migration status and public key status to the existing team endpoint.

SmittieC: https://github.com/dimagi/open-chat-studio/pull/3751/commits/52660c73b6434238959519a1c606c89ef6d940d4

apps/api/export/serializers.py
```diff
@@ -203,11 +203,28 @@ def build_resource_serializer(model):
...
-    return build_resource_serializer(entry_model(TEAM_MODEL))
+    surface and is served as one object rather than a page. It extends the importable team-row dump with
+    the two operational status fields the sync client preflights on: ``is_migrating`` (kept out of the
+    importable row, so migration mode isn't replicated to the target) and ``public_key``, which the
+    method field below collapses to a boolean -- whether a key is registered, never the key itself. Only
+    ``members`` is dropped outright."""
+    base = build_resource_serializer(entry_model(TEAM_MODEL))
+
+    class TeamExportSerializer(base):
+        public_key = serializers.SerializerMethodField(
```
snopoke: nit: 
```suggestion
        has_public_key = serializers.SerializerMethodField(
```

snopoke (dismissed):

PR #3761, opened 2026-07-06
update primary email when user changes email in profile

snopoke:
### Product Description
small change to update a user's primary email address when edited in their profile


apps/users/views.py
```diff
@@ -33,8 +33,9 @@ def profile(request):
         if form.is_valid():
             user = form.save(commit=False)
             user_before_update = CustomUser.objects.get(pk=user.pk)
+            email_changed = user_before_update.email != user.email
```
snopoke: Fixed in 56528b0ec. Extracted the email-change handling out of `profile` into `_save_profile_form`, `_send_email_confirmation_if_required`, and `_sync_primary_email_address`, reducing the view's cyclomatic complexity and nesting depth.

*— Claude Code*

snopoke: Fixed in 56528b0ec. The nested email-change blocks were extracted into dedicated helpers (`_save_profile_form`, `_send_email_confirmation_if_required`, `_sync_primary_email_address`), flattening the bumpy road in `profile`.

*— Claude Code*

snopoke: Fixed in 56528b0ec. The deeply nested email-handling logic was moved into helper functions, so `profile` no longer nests to depth 4.

*— Claude Code*

apps/users/views.py
```diff
@@ -48,6 +49,15 @@ def profile(request):
...
 
+            if email_changed and not need_to_confirm_email:
+                # The email was changed to one the user already owns (e.g. an already-verified
+                # secondary address). Keep allauth's primary EmailAddress in sync with the user's
+                # email so login codes etc. are sent to the right place. See apps/users/signals.py
+                # for the equivalent behavior when a brand-new email is confirmed.
+                email_address = EmailAddress.objects.filter(user=user, email__iexact=user.email).first()
+                if email_address:
+                    email_address.set_as_primary()
+
```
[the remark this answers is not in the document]
snopoke: Both addressed in 56528b0ec:

- **Promote only verified addresses**: the primary-promotion lookup now requires `verified=True`, so an unverified address is never made primary.
- **Brand-new emails**: we now `get_or_create` an `EmailAddress` row for the new email (created unverified) so allauth's records stay in sync with `user.email`. It is only promoted to primary once verified.

*— Claude Code*

PR #3762, opened 2026-07-06
Move Facebook channel to channels_v2

### LINKED ISSUE #3330: Move Facebook Channel to v2

SmittieC:
### Product Description
No change — internal refactor of the Facebook Messenger channel.

### Technical Description
Migrates the Facebook channel to the channels_v2 stateless pipeline architecture (ADR-0009/0010), following the same pattern as the WhatsApp and SureAdhere migrations. Resolves #3330.

Behavior-preserving details worth checking:
- Outbound files stay unsupported (`supports_files=False`) — the legacy channel never implemented file sending, so attachments continue to fall back to download links.
- Voice support and message types still come from the messaging service at runtime.
- New: `_page_id` now raises a `ValueError` when `extra_data` lacks `page_id` instead of passing `None` through to the provider (mirrors the WhatsApp channel's `phone_number_id` check).

The legacy `FacebookMessengerChannel` in `apps/chat/channels.py` is kept with a TODO-remove marker, consistent with the other migrated channels.

### Docs and Changelog


apps/chat/channels.py
```diff
@@ -1247,6 +1251,7 @@ def submit_input_to_llm(self):
             logger.exception("Failed to send typing indicator")
 
 
+# TODO: remove after channels refactor — replaced by apps.channels.channels_v2.facebook_channel.FacebookMessengerChannel
 class FacebookMessengerChannel(ChannelBase):
```
snopoke: Is this still referenced anywhere? If not we could remove it now.

SmittieC: Sure

SmittieC: https://github.com/dimagi/open-chat-studio/pull/3762/commits/063f8d305716106a9993a464c2501824e25257dd. I'll stick to FB in this PR, but there's a few other channels we can remove as well

PR #3772, opened 2026-07-07
Reflect stage failures in traces

snopoke:
### Product Description
No change.

### Technical Description
A CommCare Connect FCM delivery failure ([Sentry](https://dimagi.sentry.io/issues/7457992939/)) was reported to Sentry but the OCS/langfuse trace for the run showed no error. Root cause: in the v2 channel pipeline a stage's span only reflected failure when an exception *propagated through* the tracer's context manager. `ResponseSendingStage` catches send failures into `ctx.sending_exceptions` and recovers, so its span — and the whole trace — stayed green.

Fixing this surfaced the inverse problem too: `EarlyExitResponse`/`EarlyAbort`/`GenerationCancelled` are `Exception` subclasses, so normal consent flows, unsupported-message replies, `/reset`, and cancellations were being recorded as ERROR traces.

Span-status handling is now centralized in `ProcessingStage.__call__`, distinguishing three outcomes:
- genuine exception raised → propagates, tracer marks span + trace ERROR (unchanged);
- control-flow signal raised → caught, span closes clean, signal re-raised *after* the span exits;
- failure caught and recorded on the context → surfaced on the span via `mark_span_as_error`.

Supporting changes: the OCS tracer now fails a trace when a span is marked as error without raising (langfuse already did via `_update_span_from_context`); `new_user_message` catches `GenerationCancelled` inside the trace block; `ResponseSendingStage` records recipient/message context on its span.

Also adds input/output context to every stage span. Stages declare `span_input_fields` / `span_output_fields` as context attribute paths (dotted paths like `experiment_session.status` supported); the base renders them trace-safe — Django models collapse to `{id, model}` (never `str(model)`, which can fire queries), collections are size/depth bounded, and serialization never raises so observability can't break message processing.

Worth a reviewer's attention:
- **Behavior change:** consent/unsupported/reset/cancel traces flip from ERROR → SUCCESS, which may shift error-rate dashboards/metrics.
- The existing `message_delivery_failure_notification` still fires; no new trace-error notification fires for send failures because `ResponseSendingStage` carries no `notification_config`.
- Commits are atomic and ordered by dependency (tracer infra → stage span-status → cancellation → span I/O).

### Migrations
N/A — no migrations.

### Docs and Changelog

### DIFF
--- a/apps/channels/channels_v2/channel_base.py
+++ b/apps/channels/channels_v2/channel_base.py
@@ -76,30 +76,34 @@ class ChannelBase(ABC):
             ctx = self._create_context(message)
             pipeline = self._build_pipeline()
 
-            try:
-                with self.trace_service.trace(
-                    trace_name=self.experiment.name,
-                    session=ctx.experiment_session,
-                    inputs={"input": message.model_dump(mode="json")},
-                ) as span:
+            with self.trace_service.trace(
+                trace_name=self.experiment.name,
+                session=ctx.experiment_session,
+                inputs={"input": message.model_dump(mode="json")},
+            ) as span:
+                try:
                     ctx = pipeline.process(ctx)
+                except GenerationCancelled:
+                    # Cancellation is control flow, not a failure -- close the
+                    # trace cleanly rather than letting it propagate through the
+                    # trace context manager and mark the trace as errored.
+                    span.set_outputs({"response": "", "cancelled": True})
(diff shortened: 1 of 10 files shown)


snopoke:
Rebuilding this after the source files all moved.

snopoke:
Superseded by #3778, rebased onto the flattened `channels` layout from #3776.

PR #3776, opened 2026-07-07
refactor: flatten channels_v2 into channels

SmittieC:
### Product Description
No change — internal refactor only, no user-facing behaviour change.

### Technical Description
Removes the `apps/channels/channels_v2/` package and moves every module up into `apps/channels/` directly; the `stages/` sub-package stays a sub-package under `apps/channels/`. All imports, test mocks/patch strings, and one live developer guide reference were updated to the new paths.

Reconciliation points worth a look:
- The two `EarlyExitResponse` / `EarlyAbort` classes from `channels_v2/exceptions.py` were merged into the existing `apps/channels/exceptions.py` (which already held `ExperimentChannelException` / `InvalidTelegramChannel`).
- `stages/` moved in a single commit rather than file-by-file, because its `__init__.py` uses relative imports (`.core`, `.terminal`) that would break a partial move.
- Two references were not plain `apps.channels.channels_v2.X` module paths and needed hand-fixing: a relative import in `apps/channels/apps.py`, and a lazy package import in `apps/experiments/models.py` (`from apps.channels.channels_v2 import registry`) — the latter is only imported at call time, so `manage.py check` didn't surface it; it was caught by a repo-wide grep.

Each module was moved in its own commit (imports + tests updated together) to keep the history reviewable and `git mv` renames traceable. Dated historical planning docs under `docs/superpowers/plans|specs/` still mention the old path and were intentionally left as point-in-time records.

### Migrations
N/A — no schema changes.

### Demo
N/A

### Docs and Changelog

Verification: `apps/channels/tests/` (660 passed, 1 skipped), affected external tests in `api`/`chatbots`/`experiments`/`slack` (210 passed), `ty check apps/channels/` clean, and a repo-wide grep confirms no `channels_v2` references remain in code.


SmittieC:
Merging. We just moved a bunch of files

PR #3780, opened 2026-07-07
Null dangling scalar FK ids in Node param sync

snopoke:
### Product Description
No change.

### Technical Description
Production deploy of migration `service_providers.0059_delete_claude_sonnet_4_20250514` failed at COMMIT with a `ForeignKeyViolation` on `pipelines_node.llm_provider_id` (a deleted provider). The data migration itself succeeded — the deferred FK constraint fired at commit.

Root cause: `Node._sync_resource_fk_fields` re-derives every resource FK column from the `params` JSON but never checked that the referenced row still exists. `LlmProvider.delete()` has no guard against node references (unlike `LlmProviderModel`), so deleting one leaves `on_delete=SET_NULL` to null the FK column while the stale id lingers in `params`. When the migration touched such a node, sync resurrected the dangling id into the column, and the deferred constraint blew up at commit.

Fix: coerce a scalar FK id to null when the referenced row no longer exists, matching what `backfill_node_fks` and the `collection_indexes` M2M already do. Existence is checked via `_base_manager` so a valid reference to a soft-deleted (archived) resource isn't mistaken for dangling.

Migration 0059 rolled back cleanly (idempotency timestamp included), so it re-runs and succeeds on redeploy — no migration change needed.

### Migrations
N/A — no new migration; unblocks an existing one on redeploy.

### Docs and Changelog

### DIFF
--- a/apps/data_migrations/tests/test_remove_deprecated_models.py
+++ b/apps/data_migrations/tests/test_remove_deprecated_models.py
@@ -7,10 +7,16 @@ from apps.pipelines.models import Pipeline
 from apps.service_providers.models import LlmProviderModel
 from apps.utils.factories.experiment import ExperimentFactory
 from apps.utils.factories.pipelines import PipelineFactory
-from apps.utils.factories.service_provider_factories import LlmProviderModelFactory
+from apps.utils.factories.service_provider_factories import LlmProviderFactory, LlmProviderModelFactory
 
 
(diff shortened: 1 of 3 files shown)


snopoke:
@SmittieC FYI this came up during the recent deploy failure. I haven't verified the hypothesis - could you take this on as part of the node FK migration please.

SmittieC (approved):
Looks right to me

PR #3787, opened 2026-07-08
feat: add progress bar for chat export

SmittieC:
### Product Description
Chat exports can take a while for chatbots with lots of sessions. Previously the "Generate Chat Export" button just showed an indeterminate spinner, so users had no idea how far along the export was. This adds a live percentage progress bar (like the team-files export already has), so users can see real progress and know roughly how long is left.

### Technical Description
- `async_export_chat` now reports progress via `celery_progress.ProgressRecorder`. The total is the message count (`count_export_messages`), and progress is updated once per export chunk (1000 messages) so we don't spam Redis.
- `generate_export_rows` / `export_to_tempfile` gained an optional `progress_callback` (defaults to `None`, so the streaming export in `analysis/views.py` and all other callers are unaffected).
- Frontend: a compact inline progress bar in the sessions toolbar, driven by the `celery-progress` JS lib. `onSuccess` builds the download link client-side from the returned `file_id`. This replaces the old every-2s HTMX polling of `get_export_download_link` with the single celery-progress poll loop, and that now-unused view/URL is removed.
- Translated strings interpolated into the inline `<script>` are piped through `escapejs` so locales with apostrophes (e.g. French) can't break out of the JS string literals.

### Demo
https://www.loom.com/share/42f8d76bbfe34eda8e5df135a9380138

### Docs and Changelog

### DIFF
--- a/apps/experiments/export.py
+++ b/apps/experiments/export.py
@@ -200,6 +200,10 @@ def generate_export_rows(
     session_cache: dict[int, dict] = {}
     processed = 0
 
+    def report(count):
+        if progress_callback:
+            progress_callback(count)
+
     while True:
         chunk = list(base_qs.filter(pk__gt=last_pk)[:EXPORT_CHUNK_SIZE])
         if not chunk:
@@ -208,15 +212,17 @@ def generate_export_rows(
         for message in chunk:
             yield _yield_row_for_message(message, session_cache, experiment, translation_language)
             processed += 1
-            if progress_callback and processed % PROGRESS_UPDATE_INTERVAL == 0:
-                progress_callback(processed)
+            if processed % PROGRESS_UPDATE_INTERVAL == 0:
+                report(processed)
 
         last_pk = chunk[-1].pk
         if len(chunk) < EXPORT_CHUNK_SIZE:
             break
 
-    if progress_callback and processed and processed % PROGRESS_UPDATE_INTERVAL != 0:
-        progress_callback(processed)
+    # Report the final tally so the last partial interval (and exports smaller than one
+    # interval) still reach 100%. processed == 0 is a no-op since 0 % INTERVAL == 0.
+    if processed % PROGRESS_UPDATE_INTERVAL != 0:
+        report(processed)
 
 
 def export_rows_to_csv_stream(rows: Iterator[list]) -> Generator[str]:
--- a/apps/experiments/export.py
+++ b/apps/experiments/export.py
@@ -18,6 +18,7 @@ from apps.web.dynamic_filters.datastructures import FilterParams
 _SPOOLED_MAX_BYTES = 10 * 1024 * 1024  # 10 MB threshold before spilling to disk
 
 EXPORT_CHUNK_SIZE = 1000
+PROGRESS_UPDATE_INTERVAL = 100
 
 UTF8_BOM = "\ufeff"  # Prepended to CSV exports so Excel detects UTF-8 encoding.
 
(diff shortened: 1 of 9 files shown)


apps/experiments/tasks.py
```diff
@@ -30,11 +31,18 @@ def async_export_chat(self, experiment_id: int, query_params: dict, time_zone) -
     experiment = Experiment.objects.get(id=experiment_id)
     filtered_sessions = get_filtered_sessions(experiment, query_params, time_zone)
     filename = f"{experiment.name} Chat Export {timezone.now().strftime('%Y-%m-%d_%H-%M-%S')}.csv.gz"
+
+    progress_recorder = ProgressRecorder(self)
+    total = count_export_messages(filtered_sessions)
+
+    def report_progress(current):
+        progress_recorder.set_progress(current, total, description=f"Processing {current} of {total} messages")
```
snopoke, line 39: This should rate limit / only update after N. Updating after every message will likely slow this down a lot.


```suggestion
        if current % 100 == 0 or current >= total:
            progress_recorder.set_progress(current, total, description=f"Processing {current} of {total} messages")
```

SmittieC: https://github.com/dimagi/open-chat-studio/pull/3787/commits/19c3b829944c454d9274142d2a06dc37843ec72d

templates/experiments/components/exports.html
```diff
@@ -1,30 +1,115 @@
...
+            const bar = document.getElementById("chat-export-bar");
+            const percentEl = document.getElementById("chat-export-percent");
+            const downloadLink = document.getElementById("chat-export-download");
+            const errorEl = document.getElementById("chat-export-error");
+
+            function resetGenerateButton() {
+              if (!generateBtn) return;
+              generateBtn.disabled = false;
+              generateBtn.innerHTML = '<i class="fa-solid fa-file-export"></i> ' + GENERATE_LABEL;
+            }
```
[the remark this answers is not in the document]
SmittieC: Fixed in 13c246a8b — the label is now set via `textContent` (createTextNode) instead of `innerHTML`, so a translated label containing HTML characters stays literal text.

templates/experiments/components/exports.html
```diff
@@ -1,30 +1,115 @@
...
+
+              onSuccess: function (_bar, _msg, result) {
+                const fileId = result && result.file_id;
+                if (!fileId) {
+                  showError(MSG_NO_FILE);
+                  return;
+                }
+                if (downloadLink) {
+                  downloadLink.href = downloadBaseUrl.replace(/\/0(\/?$)/, "/" + fileId + "$1") + "?allow_s3";
+                  downloadLink.style.display = "";
```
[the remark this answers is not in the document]
SmittieC: Fixed in 13c246a8b — switched to a function replacement so a `$` in the file id cannot be interpreted as a `$`-substitution pattern.

apps/experiments/export.py
```diff
@@ -195,11 +207,17 @@ def generate_export_rows(experiment, sessions_queryset, translation_language=Non
...
             yield _yield_row_for_message(message, session_cache, experiment, translation_language)
+            processed += 1
+            if progress_callback and processed % PROGRESS_UPDATE_INTERVAL == 0:
+                progress_callback(processed)
 
         last_pk = chunk[-1].pk
         if len(chunk) < EXPORT_CHUNK_SIZE:
             break
 
+    if progress_callback and processed and processed % PROGRESS_UPDATE_INTERVAL != 0:
```
SmittieC: Fixed in 46f885e1b — extracted a small `report()` helper so the per-interval and final-flush checks are single-condition, removing the complex conditional. No behaviour change.

apps/experiments/tasks.py
SmittieC: Addressed in 19c3b829 (refactored in 46f885e1b). The throttle lives in `generate_export_rows` rather than `report_progress`: the callback now fires every `PROGRESS_UPDATE_INTERVAL` (100) messages plus a final flush, so `set_progress` runs far fewer times on large exports. The interval is intentionally separate from `EXPORT_CHUNK_SIZE`.

Issue #3384, opened 2026-05-19
Modernize evaluations DeleteViews to use form_valid() instead of delete() override


barry47products:
### Context

Spun out of the design discussion on #3203 (see https://github.com/dimagi/open-chat-studio/issues/3203#issuecomment-).

Django 4.0+ soft-deprecated overriding `delete()` directly on `DeleteView` in favour of `form_valid()`. The evaluations app has three views that still use the old pattern:

- `DeleteEvaluator`
- `DeleteDataset`
- `DeleteEvaluation` (added in #3383, kept consistent with the existing two)

### Proposal

Migrate all three to `form_valid()`. Pure refactor, no behaviour change. Doing all three together keeps the app internally consistent.

### Acceptance

- All three views use `form_valid()` instead of overriding `delete()`.
- Existing tests still pass; no new behaviour to test.

ocs-percy:
## Audit results

### Scope of the pattern

The deprecated `delete()` override only matters on subclasses of Django's `DeleteView`. I grepped for every `DeleteView` subclass in the codebase — there are exactly **three**:

| Class | File | Uses deprecated `delete()`? |
|---|---|---|
| `DeleteEvaluator` | `apps/evaluations/views/evaluator_views.py` | ✅ Yes — scope of this issue |
| `DeleteDataset` | `apps/evaluations/views/dataset_views.py` | ✅ Yes — scope of this issue |
| `TranscriptAnalysisDeleteView` | `apps/analysis/views.py` | ✅ **Already clean** — no `delete()` override |

### False positives — plain `View` subclasses

Several other Delete-style views implement a `delete(self, request, ...)` method, but they subclass `View`, not `DeleteView`. That's just a standard HTTP method handler — it's the correct pattern for AJAX/HTMX deletes and is **not affected by the Django 4.0 deprecation**:

- `DeleteSurvey`, `DeleteConsentForm`, `DeleteSourceMaterial` (experiments app)
- `DeleteMcpServer` (mcp_integrations)
- Various HTMX delete handlers in annotations, pipelines, custom_actions, assistants, files, etc.

### Conclusion

This issue already captures the full scope. No additional views need to be migrated.

barry47products:
.take

barry47products:
[...]

PR #3794, opened 2026-07-08
refactor: standardize delete() signature style across evaluations delete views

### LINKED ISSUE #3792: Standardize delete() signature style across evaluations delete views
### Context

Follow-up from #3790. That PR migrated `DeleteEvaluator`, `DeleteDataset`, `DeleteEvaluation`, and `DeleteAutoPopulationRule` from `DeleteView` to a plain `View`, modeled on the existing `DeleteFile` (`apps/files/views.py:289`) and `DeleteCustomAction` (`apps/custom_actions/views.py:160`) pattern.

Those two reference views use an explicit signature:

```python
```

The four new views instead use:

```python
def delete(self, request, *args, **kwargs):
    ...pk=kwargs["pk"]
```

Functionally identical (Django's `View.setup()` populates `self.kwargs` either way, and `team_slug` isn't referenced in either style since `request.team` already carries it), but since the PR explicitly modeled itself on `DeleteFile`/`DeleteCustomAction`, matching their signature style would be more consistent. Flagged as Minor in code review on #3790, not blocking.

### Proposal

In `apps/evaluations/views/evaluator_views.py`, `dataset_views.py`, `evaluation_config_views.py`, and `auto_population_views.py`, change each `DeleteEvaluator`/`DeleteDataset`/`DeleteEvaluation`/`DeleteAutoPopulationRule.delete()` from:

```python
def delete(self, request, *args, **kwargs):
    obj = get_object_or_404(Model, team=request.team, pk=kwargs["pk"])
```

to:

```python
    obj = get_object_or_404(Model, team=request.team, pk=pk)
```

No behaviour change, purely cosmetic consistency with the existing codebase pattern.

barry47products:
### Product Description

No user-facing change.

### Technical Description

Closes #3792.

Follow-up to #3790. `DeleteFile` (`apps/files/views.py:289`) and `DeleteCustomAction` (`apps/custom_actions/views.py:160`) use an explicit `def delete(self, request, team_slug: str, pk: int):` signature. The four views migrated in #3790 instead used `def delete(self, request, *args, **kwargs):` with `kwargs["pk"]`. Functionally identical, Django's URL resolver passes the same kwargs either way, but since #3790 explicitly modeled itself on those two views, matching their signature style is more consistent. Flagged as Minor in code review on #3790.

Considered extracting a shared base/mixin instead (the `get_object_or_404(Model, team=request.team, pk=pk)` line is now duplicated across 6 views), but the bodies diverge meaningfully beyond that one line (`DeleteFile`/`DeleteCustomAction` do reference-checking and return conditional modal responses, `DeleteCustomAction` additionally wraps in `transaction.atomic()` with row-level locking, `DeleteEvaluation` branches on a redirect query param) - a mixin would only share one line while still requiring most callers to override the rest, so kept this as a signature-only change.

No test changes needed: existing tests exercise the URL/response contract, not the view's internal parameter style. Confirmed by running the full suite unchanged.

**Verification:** full `apps/evaluations/` suite green (290 tests), `ruff check`/`ruff format --check` clean, `ty check` clean.

### Migrations

No migrations in this PR.

### Demo

Non-visual, cosmetic refactor, nothing to demo.

### Docs and Changelog

### DIFF
--- a/apps/evaluations/views/auto_population_views.py
+++ b/apps/evaluations/views/auto_population_views.py
@@ -90,8 +90,8 @@ class EditAutoPopulationRule(_RuleViewMixin, UpdateView):
 class DeleteAutoPopulationRule(LoginAndTeamRequiredMixin, PermissionRequiredMixin, View):
     permission_required = "evaluations.change_evaluationdataset"
 
-    def delete(self, request, *args, **kwargs):
-        get_object_or_404(DatasetAutoPopulationRule, team=request.team, pk=kwargs["pk"]).delete()
+    def delete(self, request, team_slug: str, pk: int):
+        get_object_or_404(DatasetAutoPopulationRule, team=request.team, pk=pk).delete()
         return HttpResponse(status=200)
 
 
--- a/apps/evaluations/views/dataset_views.py
+++ b/apps/evaluations/views/dataset_views.py
@@ -158,9 +158,9 @@ class EditDataset(LoginAndTeamRequiredMixin, PermissionRequiredMixin, UpdateView
 class DeleteDataset(LoginAndTeamRequiredMixin, PermissionRequiredMixin, View):
     permission_required = "evaluations.delete_evaluationdataset"
 
-    def delete(self, request, *args, **kwargs):
+    def delete(self, request, team_slug: str, pk: int):
         """Handle AJAX delete requests."""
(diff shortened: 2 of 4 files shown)


barry47products:
Small follow-up to #3384 (merged via #3790), included as part of that work, not a separate feature or fix.

snopoke (approved):

PR #3795, opened 2026-07-09
feat: export archived records and back-fill file content in team sync

SmittieC:
### Product Description
Makes team-to-team sync a faithful, complete snapshot. Two gaps are closed:

1. **Archived records are now exported.** Previously, archiving a chatbot, pipeline, collection, file, etc. silently dropped it (and its file bytes) from the export, so an archived resource never made it to the target server.
2. **Missing file content is back-filled on import.** When a file's bytes aren't already present in the target's storage, the importer fetches them from the source rather than failing.

### Technical Description
**Export archived records** (`feat: export archived records in the team export API`)
- Versioned models filter `is_archived=False` on their default manager. `team_scoped_queryset()` now calls `model.objects.get_all()` (the `VersionsObjectManagerMixin` escape hatch) when available to bypass that filter, so archived rows are served for every versioned model.
- `get_team_files_queryset()` (the file-blob zip) also uses `File.objects.get_all()`, so archived files' bytes ship too — otherwise imported archived `File` rows would reference blobs that were never uploaded. (Archiving a file never removes its storage blob, so the bytes are available.)

**Back-fill missing file content** (`feat: back-fill missing file content when syncing a team`)
- On import, if a file's blob is missing from target storage, fetch it from the source's file-content API and write it before saving the row, instead of failing.
- A `404` from the source means the source is missing the blob too: the row is imported without content (so references still resolve) and the file is listed in the sync report. Other fetch errors abort the sync.
- Force-delete now warns that it also clears the team's files from backend storage (they must be re-imported).

**Fix archived-collection deletion** (`fix: delete files of archived collections in delete_collection_task`)
- `delete_collection_task` runs after `Collection.archive()` has set `is_archived=True`, but it loaded the row via `Collection.objects`, whose manager hides archived rows — so the task was a silent no-op that left `CollectionFile` rows orphaned. It now loads the row with `get_all()`.

### Migrations
No migrations in this PR.

### Demo
N/A

### Docs and Changelog

### DIFF
--- a/apps/teams/export/client.py
+++ b/apps/teams/export/client.py
@@ -9,6 +9,16 @@ import requests
 _API_KEY_HEADER = "X-Api-Key"
 
 
+class FileContentNotFound(Exception):
+    """The source's file content API returned 404 -- the source is missing the blob too, so it can't
+    be backfilled. Distinct from other HTTP errors so the importer can record it and carry on rather
+    than aborting the sync."""
+
+    def __init__(self, file_id: int) -> None:
+        self.file_id = file_id
+        super().__init__(f"file {file_id} has no content on the source")
+
+
 class ResourceFetcher:
     def __init__(self, base_url, api_key, *, session=None, sleep=time.sleep):
         self.base_url = base_url.rstrip("/")
@@ -49,7 +59,24 @@ class ResourceFetcher:
                 break
             cursor = page["cursor"]
 
+    def get_file_content(self, file_id: int) -> bytes:
+        """The raw bytes of a file, from the source's file content API (``/api/files/<id>/content``).
+        Used to backfill a synced file whose blob is missing from this server's storage. A 404 means
+        the source is missing the blob too; raise ``FileContentNotFound`` so the caller can record it
+        and carry on. Other client errors surface as-is."""
+        try:
+            return self._request(f"/api/files/{file_id}/content").content
+        except requests.HTTPError as exc:
+            if exc.response is not None and exc.response.status_code == 404:
+                raise FileContentNotFound(file_id) from exc
+            raise
+
     def _get(self, path, params=None) -> dict:
+        return self._request(path, params).json()
+
+    def _request(self, path: str, params: dict | None = None) -> requests.Response:
+        """GET ``path`` with the API key header, retrying transient transport errors (timeouts, 5xx,
+        connection resets) with backoff and failing fast on 4xx. Returns the raw response."""
         url = self.base_url + path
         headers = {_API_KEY_HEADER: self.api_key}
         for attempt in range(self.max_retries):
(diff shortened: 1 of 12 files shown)


apps/teams/export/client.py
```diff
@@ -49,7 +59,24 @@ def iter_rows(self, resource, start_cursor=None, limit=100):
...
+        """The raw bytes of a file, from the source's file content API (``/api/files/<id>/content``).
+        Used to backfill a synced file whose blob is missing from this server's storage. A 404 means
+        the source is missing the blob too; raise ``FileContentNotFound`` so the caller can record it
+        and carry on. Other client errors surface as-is."""
+        try:
+            return self._request(f"/api/files/{file_id}/content").content
+        except requests.HTTPError as exc:
+            if exc.response is not None and exc.response.status_code == 404:
+                raise FileContentNotFound(file_id) from exc
+            raise
```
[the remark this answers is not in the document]
SmittieC: Archived files can be stored as empty, its fine. They have no real use apert from being present

apps/teams/export/manifest.py
```diff
@@ -233,8 +231,13 @@ def model_has_team_field(model: type[Model]) -> bool:
...
 
 def team_scoped_queryset(entry: ManifestEntry, team) -> QuerySet:
-    """The model's rows for this team, including any shared global rows."""
+    """The model's rows for this team, including any shared global rows and any archived rows.
+
+    Versioned models filter ``is_archived=False`` on their default manager; ``get_all()`` bypasses
+    that so archived rows are shared across servers along with the live ones."""
     model = entry_model(entry.model)
+    manager = model.objects
+    base = manager.get_all() if hasattr(manager, "get_all") else manager
```
snopoke: You could always use `_base_manager`: https://docs.djangoproject.com/en/6.0/topics/db/managers/#django.db.models.Model._base_manager

SmittieC: If we do that, we'll be exporting deleted channels as well. I guess it's fine

snopoke: OK, I guess it's up to you - not sure if exporting 'deleted' objects will cause issues

SmittieC: Don't think so, no. It will just be deleted at the target server as well. We'll see though

SmittieC: Done — switched `team_scoped_queryset` to `_base_manager` (f365f4ff5, ddd8d8baf).

It also surfaced a related bug: the importer's existence check in `_get_or_create` went through the default manager too, so archived/soft-deleted target rows weren't found on re-import and got recreated as duplicates — and for channels that violates the unique `external_id`. Fixed that lookup to use `_base_manager` as well.

Since `_base_manager` bypasses *all* default-manager filtering, this also means `ExperimentChannel`'s `deleted=False` filter is now bypassed, so soft-deleted channels export too — deliberate, for a faithful snapshot, and re-import of them is now idempotent (covered by a test).

snopoke (approved):

Issue #2680, opened 2026-01-21
Support incremental / partial saving for pipelines


snopoke:
Goal
- Implement incremental / partial saving for pipeline editing so we only persist the data that changed instead of saving and reloading the entire pipeline on every change.

Background / motivation
- Current behavior saves and reloads the whole pipeline for each edit. This causes unnecessary I/O, slower UI responsiveness, potential merge conflicts, and loss of transient editor state.
- Incremental saves would reduce latency, lower server/workspace load, and enable finer-grained undo/redo and conflict resolution.

Acceptance criteria / proposed approach
- Introduce a mechanism to detect which pipeline nodes/fields changed and send only those diffs to the backend.
- Backend API endpoint(s) to accept partial updates (e.g., PATCH /pipelines/:id with a JSON Patch or delta payload).
- Ensure transactional consistency so partial updates don’t leave pipeline in an invalid state; include validation and atomic apply/reject semantics.
- Preserve editor state (selection, undo stack) across partial saves when possible.
- Add tests for front-end change detection, diff generation, and backend apply/validation paths.
- Provide migration/compatibility strategy for existing clients/backends.

Risks / considerations
- Need clear conflict resolution strategy for concurrent editors.
- Validation and ordering of partial updates must be robust.
- Consider batching small rapid edits to avoid many network calls (debounce/aggregation).

PR #3797, opened 2026-07-09
Implement incremental pipeline saving (#2680)

ndegwaofficial:
### Product Description

Adds incremental pipeline saving using semantic graph diffs and optimistic concurrency.

Pipelines can now be autosaved by sending only the changed nodes and edges instead of the entire graph, while preserving existing pipeline editing behaviour. Invalid draft pipelines continue to save successfully and return validation errors to the editor as before.

### Technical Description

This PR introduces an incremental save path alongside the existing full-save API without changing the existing editing semantics.

The backend now supports PATCH /pipelines/data/<id>/, accepting a semantic graph diff (PipelineDiffPayload) together with an edit_revision concurrency token. Diffs are applied in memory to the canonical Pipeline.data, after which the existing update_nodes_from_data() reconciliation logic is reused unchanged. This preserves all existing node synchronization behaviour, including node creation, deletion, archival/version handling, and custom-action synchronization.

An independent edit_revision field was added to Pipeline for optimistic concurrency. This intentionally does not reuse version_number, which continues to represent published pipeline versions only. Successful POST and PATCH saves increment edit_revision, while stale PATCH requests return HTTP 409 Conflict.

Validation behaviour remains unchanged. Pipelines are still persisted before validation and any validation errors are returned in the response rather than preventing saves, preserving the current draft-editing workflow.

On the frontend, autosave now computes semantic node/edge diffs and submits them via PATCH. Manual saves continue using the existing full POST endpoint for backwards compatibility.

### Migrations

A single backwards-compatible migration adds the Pipeline.edit_revision field with a default value of 0.
 
#### Manual save (existing behaviour)

1. Open a pipeline.

2. Make changes.

3. Click **Save**.

4. The existing POST endpoint is used and the pipeline saves normally.

#### Incremental autosave

1. Open a pipeline.

2. Modify nodes or edges.

3. Wait for autosave.

4. Only the semantic graph diff is sent to the backend via PATCH.

#### Optimistic concurrency

1. Open the same pipeline in two browser tabs.

2. Modify and save from the first tab.

3. Modify the pipeline in the second tab.

4. The second autosave receives **409 Conflict** because its revision is stale.

#### Validation

1. Create an invalid pipeline (for example, disconnect required nodes).

2. Save or wait for autosave.

3. The pipeline is still saved and validation errors are returned, matching existing behaviour.

### Docs and Changelog


ndegwaofficial:
I have fully implemented the Minor PR for issue #2680

snopoke:
@ndegwaofficial thanks for the contribution. Just letting you know that I'll only be able to review this next week.

assets/javascript/apps/pipeline/api/api.ts
```diff
@@ -40,6 +40,32 @@ class ApiClient {
...
+          throw err;
+        }
+        throw new Error(`HTTP error! status: ${response?.status}`);
+      }
+      return response.data;
+    } catch (error) {
+      console.error(error);
+      throw error;
+    }
+  }
```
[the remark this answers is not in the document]
snopoke: @ndegwaofficial is this still relevant?

snopoke (commented):
I tested this locally with multiple tabs and when a 409 happens there is not indaction in the UI.

ndegwaofficial:
> I tested this locally with multiple tabs and when a 409 happens there is not indaction in the UI.

@snopoke I have made the requested adjustment, to show 409 err indication on UI.

ndegwaofficial:
@snopoke I have also added proper 409 detection in the  catch  block with  axios.isAxiosError(error) && error.response?.status === 409 . When a 409 is detected, it creates the custom error with  .status = 409  and  .currentRevision = ...  that the callers in  pipelineStore.ts  expect.

snopoke (commented):
The dismiss button overwrites the server state with the state from the current editor. I don't think this is clear from the UI. I'm also not sure that we want this. I think we should rather force the user to reload to get the updated state.

ndegwaofficial:
@snopoke I have made the requested fix and now the user will be forced to reload in order to see the latest version cc @davidnene

assets/javascript/apps/pipeline/Page.tsx
```diff
@@ -93,12 +93,6 @@ export default function Page() {
                 <span>This pipeline was modified in another session. Reload to see the latest version.</span>
               </div>
               <div className="flex gap-2 shrink-0">
-                <button
-                  className="btn btn-sm btn-ghost"
-                  onClick={dismissConflict}
```
snopoke, line 98: Is this function used anymore?

ndegwaofficial: @snopoke  It is still used. The standalone Dismiss button was removed, but  dismissConflict()  was rolled into the Reload button's  onClick  handler:
 
```
<button
  className="btn btn-sm btn-warning"
  onClick={() => {
    dismissConflict();
    if (currentPipelineId) {
      loadPipeline(currentPipelineId);
    }
  }}
>
  Reload
</button>
```
So clicking "Reload" first clears the conflict state, then reloads the pipeline. The destructure on line 12 and the store action are both still active. No dead code here.

snopoke (approved):

PR #3813, opened 2026-07-13
Record cost tracking data for all teams; flag gates UI only

snopoke:
### Product Description
No user-facing change. Cost/usage data is now captured for every team so it's already there when a team is later granted the dashboard, rather than starting empty at flag-flip time.

### Technical Description
Previously `flag_ai_cost_monitoring` gated both the write path (recording `UsageRecord`s) and the UI. This decouples them: `OCSTracer._record_costs` now runs for all teams, and the flag only controls the dashboard/pricing UI (already the case in `dashboard/views.py` and `service_providers/views.py`).

Worth noting: recording now runs on every trace. Pricing resolution is Redis-cached (24h TTL) and unpriced models write rows with `cost=0`, so the per-trace overhead is a cache hit plus one bulk insert. The recorder import moved to module level (aliased to avoid the name clash with the tracing `TraceContext`) since the lazy import only existed for the removed flag path.

### Migrations
N/A — no schema change.

### Docs and Changelog

### DIFF
--- a/apps/service_providers/tracing/ocs_tracer.py
+++ b/apps/service_providers/tracing/ocs_tracer.py
@@ -10,11 +10,12 @@ from django.core.cache import cache
 from langchain_core.callbacks.base import BaseCallbackHandler
 from langchain_core.outputs import LLMResult
 
+from apps.cost_tracking.services.recorder import TraceContext as CostTraceContext
+from apps.cost_tracking.services.recorder import record_usage_bulk
 from apps.experiments.models import Experiment, ExperimentSession
 from apps.ocs_notifications.notifications import trace_error_notification
 from apps.service_providers.tracing.const import OCS_TRACE_PROVIDER, SpanLevel
 from apps.service_providers.tracing.metrics import MetricsCollector
-from apps.teams.models import Flag, Team
 from apps.trace.models import Trace, TraceStatus
 
 from .base import SpanNotificationConfig, TraceContext, Tracer
@@ -39,7 +40,6 @@ class OCSTracer(Tracer):
         self.error_span_name: str = ""
         self.error_notification_config: SpanNotificationConfig | None = None
         self.metrics_collector: MetricsCollector | None = None
-        self.cost_tracking_enabled: bool = False
 
     @property
     def ready(self) -> bool:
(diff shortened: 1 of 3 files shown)


apps/service_providers/tracing/tests/test_ocs_tracer_cost.py
```diff
@@ -69,18 +55,8 @@ def _collector_with_events(self):
...
-        with patch("apps.cost_tracking.services.recorder.record_usage_bulk") as recorder:
-            tracer._record_costs()
-            recorder.assert_not_called()
-
     def test_short_circuits_when_collector_missing(self, experiment):
         tracer = OCSTracer(experiment, experiment.team_id)
-        tracer.cost_tracking_enabled = True
         tracer.metrics_collector = None
         tracer.trace_record = Mock(spec=Trace)
         with patch("apps.cost_tracking.services.recorder.record_usage_bulk") as recorder:
```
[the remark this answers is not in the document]
snopoke: Fixed in fc6fa2adc. Patched the binding used by the code under test (`apps.service_providers.tracing.ocs_tracer.record_usage_bulk`) at all three short-circuit tests, restoring the regression safety net.

*— Claude Code*

apps/service_providers/tracing/ocs_tracer.py
```diff
@@ -156,28 +155,22 @@ def _update_trace_metrics(self) -> None:
     def _record_costs(self) -> None:
         """Drain the collector's accumulated usage into UsageRecord rows.
 
-        Short-circuits when the flag is off so the cost path is zero-work
-        for teams that haven't opted in. `record_usage_bulk` swallows DB
-        errors internally; the outer `_finalize_trace` try/except catches
-        anything else (e.g. an unexpected helper failure).
+        Cost data is recorded for every team; the `flag_ai_cost_monitoring`
```
SmittieC: Super duper nit: I don't think comments like this should be in a docstring:

```
the `flag_ai_cost_monitoring` flag only gates the UI, not this write path
```

SmittieC (dismissed):

apps/service_providers/tracing/ocs_tracer.py
```diff
@@ -156,28 +155,22 @@ def _update_trace_metrics(self) -> None:
...
+        flag only gates the UI, not this write path. `record_usage_bulk`
+        swallows DB errors internally; the outer `_finalize_trace` try/except
+        catches anything else (e.g. an unexpected helper failure).
```
snopoke:
```suggestion
        Cost data is recorded for every team; `record_usage_bulk`
        swallows DB errors internally; the outer `_finalize_trace` try/except
        catches anything else (e.g. an unexpected helper failure).
```

PR #3830, opened 2026-07-15
fix: ensure Chat Viewer role works in isolation for transcript-only access

### LINKED ISSUE #3827: Investigate: ensure Chat Viewer role works in isolation for transcript-only access
## Overview

The `Chat Viewer` group exists to give team members read-only access to chat transcripts without broader admin rights. This issue is to verify it actually works in isolation — i.e. a user with *only* the Chat Viewer role can see transcripts and nothing else — and fix any gaps found.

## Current Permission Set

`Chat Viewer` currently grants:
```python
CHAT_VIEWER_PERMS = [
    AppPermSetDef("chat", [VIEW]),          # chat, chatmessage, chatattachment
    AppPermSetDef("annotations", [VIEW]),   # tag, customtaggeditem, usercomment
    ModelPermSetDef("files", "file", [VIEW]),
]
```

### 1. No path to reach session transcripts through the UI

The sessions list views both require `experiments.view_experimentsession`:

```python
# apps/chatbots/views.py
class ChatbotSessionsTableView(PermissionRequiredMixin, ...):
    permission_required = "experiments.view_experimentsession"

class AllSessionsHome(PermissionRequiredMixin, ...):
    permission_required = "experiments.view_experimentsession"
```

Chat Viewer does not have this permission, so a Chat Viewer-only user sees nothing useful in the UI — no way to list or navigate to session transcripts.

### 2. All chatbot/experiment home views are also blocked

`chatbots_home`, `single_chatbot_home`, `ChatbotExperimentTableView` etc. all require `experiments.view_experiment`, which Chat Viewer also does not have. A Chat Viewer-only user currently has no accessible landing page.

### 3. Individual session detail view is partially correct

The `chatbot_session_details_view` uses `@verify_session_access_cookie`, which already includes:
```python
elif request.team_membership and request.user.has_perm("chat.view_chat"):
    return view(request, *args, **kwargs)
```
So once a user reaches a session URL, `chat.view_chat` is sufficient. But there is no way to get there without `view_experimentsession`.

### 4. Other transcript entry points not yet audited

Similar gaps likely exist in:
- Participants view (sessions per participant)
- Pipelines transcript views
- Experiments (legacy) transcript views
- API endpoints

## Work to Do

- [ ] Audit all views/endpoints that display chat transcripts and check their permission requirements
- [ ] Determine the minimal permission set: likely add `experiments.view_experimentsession` (and possibly `experiments.view_participant`) to `CHAT_VIEWER_PERMS`
- [ ] Ensure a Chat Viewer-only user has a usable landing page/navigation in the UI (e.g. direct link to "All Sessions")
- [ ] Confirm Chat Viewer cannot access experiment *configuration* (chatbot settings, pipeline editor, etc.) — transcripts only
- [ ] Add or extend permission isolation tests in `apps/teams/tests/test_permissions.py` to cover the Chat Viewer role end-to-end
- [ ] Document what the role is intended to grant (code comment in `backends.py` would do)

## Notes

- The `verify_session_access_cookie` logic already has the right idea for session detail — just needs to be backed by reachable navigation
- The fix should not require Chat Viewer users to be granted `Chatbot Admin` or any other admin role
- `ANNOTATION_REVIEWER_GROUP` already reuses `CHAT_VIEWER_PERMS` as a base — changes should keep that consistent

ocs-percy:
### Product Description
The **Chat Viewer** role was unusable when assigned in isolation: a user with *only* the Chat Viewer role could not reach any chat transcript. The session list pages (the **All Sessions** page and the per-chatbot session tables) require `experiments.view_experimentsession`, which the group did not have — so there was no UI path to transcripts at all, and no nav entry linking there.

This PR makes the Chat Viewer role work on its own: such a user can now browse session lists and open transcripts (read-only), while remaining unable to view or edit any chatbot/experiment configuration.

Fixes #3827.

### Technical Description
- **`apps/teams/backends.py`** — Add `experiments.view_experimentsession` (view-only) to `CHAT_VIEWER_PERMS`. This is scoped to the `experimentsession` model only (via `ModelPermSetDef`), *not* the whole `experiments` app, so the role deliberately does **not** gain `view_experiment` — which would expose chatbot configuration views (`chatbots_home`, `single_chatbot_home`, the chatbot list/edit pages).
  - The transcript detail view (`chatbots:chatbot_session_view`) is already reachable: it is gated by `chat.view_chat` via the access-cookie fallback in `verify_session_access_cookie`, which Chat Viewer already had.
  - Participant-scoped session views (`apps/participants/`) require `view_participant` and are intentionally left out — they expose participant management, which is outside "transcripts only".
- **`templates/web/components/team_nav.html`** — The "All sessions" link was nested inside the `perms.experiments.view_experiment` block, so a Chat Viewer (who lacks `view_experiment`) would never see it. Added an `{% elif perms.experiments.view_experimentsession %}` branch that surfaces "All sessions" as a top-level nav entry for the Chat Viewer case, giving the role a landing path to transcripts.
- **`apps/teams/migrations/0015_sync_chat_viewer_session_perm.py`** — Data migration that re-runs `create_default_groups()` so the new group permission is applied deterministically in migration history (`create_default_groups()` also runs on every deploy, so this is idempotent).
- **`apps/teams/tests/test_permissions.py`** — Added regression tests:
  - `test_chat_viewer_can_view_sessions_regression` — group includes `view_experimentsession`.
  - `test_chat_viewer_cannot_view_experiment_config_regression` — group does **not** include `view_experiment` or any change/add/delete perms.
  - `test_chat_viewer_isolation_can_reach_transcripts_but_not_config` — end-to-end: a Chat Viewer-only user gets 200 on the session lists and a transcript, and 403 on chatbot configuration views.

### Migrations

Adds one data-only migration (`teams.0015`) that re-syncs default group permissions. No schema changes; it is idempotent and backwards compatible.

### Demo
A Chat Viewer-only user now sees an **All sessions** entry in the sidebar, can open the sessions list, and can view individual transcripts, while chatbot configuration pages return 403.

### Docs and Changelog

### Testing notes
The pure-permission-definition tests pass locally. The DB-backed tests (including the new end-to-end test and the existing `test_file_view_*` tests) could not be executed in the sandbox because its Postgres server lacks the `pgvector` extension; they are expected to run in CI.


apps/teams/migrations/0015_sync_chat_viewer_session_perm.py
snopoke, line 1: This is not necessary since we sync groups during the `migrate` command

snopoke (approved):

Issue #2490, opened 2025-11-21
Replace custom select dropdown for pipeline named history with 'react-select'


snopoke:
See `HistoryTypeWidget` in `assets/javascript/apps/pipeline/nodes/widgets.tsx`:

https://github.com/dimagi/open-chat-studio/blob/e259da6a3bab613bab0d9da1779825e76d4f60d4/assets/javascript/apps/pipeline/nodes/widgets.tsx#L967-L1031

PR #3842, opened 2026-07-17
Replace custom history-name combobox with CreatableSelect (#2490)

SRHIN:
Swap the hand-rolled history-name combobox in HistoryTypeWidget for CreatableSelect from react-select, the create-enabled variant already used elsewhere in this file. The library now owns open/close state, filtering, the create affordance, clearing, and outside-click-to-close (fixing the defect where the old dropdown stayed open until the chevron was clicked or an option chosen).

A thin adapter keeps the existing updateParamValue contract intact: onChange maps the selected/created/cleared option to a synthetic change event with name "history_name", byte-identical to what the combobox wrote. Options are mapped into react-select's { value, label } shape, and the current selection is derived from the saved historyName so a freshly created name always displays. Styling reuses the react-select-container / react-select classNamePrefix pattern from MultiSelectWidget, with nodrag preserved on the container. All combobox-only state, handlers, and markup are deleted.

No schema, API, or stored-data change.

### Docs and Changelog

### DIFF
--- a/assets/javascript/apps/pipeline/nodes/widgets.tsx
+++ b/assets/javascript/apps/pipeline/nodes/widgets.tsx
@@ -1023,7 +1023,7 @@ export function HistoryTypeWidget(props: WidgetParams) {
                 onChange={reportSelection}
                 isDisabled={props.readOnly}
                 isClearable
-                className="react-select-container nodrag"
+                className="react-select-container react-select-join nodrag"
                 classNamePrefix="react-select"
                 placeholder="Type to search or create..."
               />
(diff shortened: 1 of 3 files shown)


SmittieC (approved):

SmittieC (changes requested):
These are great improvements, thanks! Two small things:

1. You'll need to add support for dark mode:
image

2. Nit: the boundaries between the history type and history name inputs are slightly misalighted
image

snopoke:
@SRHIN just checking that you saw the PR feedback

SRHIN:
Thanks for the review, and sorry for the delay! Both addressed in d6b7c9a:

- Dark mode: react-select shipped a hard-coded light palette that ignored the theme. Added assets/styles/app/react-select.css that drives the control, menu, options, placeholder, and icons from the same daisyUI theme variables (--color-base-*, --color-primary*) the native inputs use, so it now follows light/dark automatically.

- Alignment: the control was 2px shorter than the type select (react-select's 38px default vs daisyUI's 40px) with its own rounded corners, leaving the notch at the seam. A react-select-join class matches the daisyUI select height and squares the shared left edge so they line up cleanly.

SmittieC (approved):

PR #3843, opened 2026-07-17
fix(scrub-sensitive-data-from-sentry): scrubs sensitive datat from se…

### LINKED ISSUE #3194: Scrub sensitive data from Sentry
I noticed in this sentry issue that the connect encryption key is included in the sentry data: [OPEN-CHAT-STUDIO-230](https://dimagi.sentry.io/issues/7303130488/?referrer=github_integration)

Update Sentry integration to scrub this and similar data. This may require renaming vars / args to use a convention.

See https://docs.sentry.io/platforms/python/data-management/sensitive-data/

brian-mecha:
### Product Description

Prevents the CommCare Connect encryption key from leaking into Sentry error reports.

### Technical Description

Sentry runs with attach_stacktrace=True, which ships stack-frame locals with every event, so the raw Connect encryption_key was captured whenever an exception fired in the encrypt/decrypt paths.

- Added a custom EventScrubber (new config/sentry.py, wired into sentry_sdk.init) extending DEFAULT_DENYLIST with encryption_key/encryption_key_bytes and recursive=True to also catch secrets nested in dicts/lists. Extracted to its own module so the scrubbing is unit-testable without initializing the SDK.
- Scrubbing matches variable/key names by exact, case-insensitive name, so standardized the convention: renamed the bare key arg to encryption_key in CommCareConnectClient.decrypt_messages / _encrypt_message / _decrypt_message (and call sites) so all raw key material is denylist-covered without over-scrubbing unrelated key vars.
- Added config/tests/test_sentry.py verifying sensitive locals (incl. nested) are redacted while non-sensitive vars are preserved.

### Migrations

### Docs and Changelog

Fixes #3194


brian-mecha:
@SmittieC Kindly review

apps/channels/clients/connect_client.py
```diff
@@ -74,24 +74,27 @@ def create_channel(self, connect_id: str, channel_source: str, channel_name: str
         _raise_for_status(response)
         return response.json()
 
-    def decrypt_messages(self, key: bytes, messages: list[Message]) -> list[str]:
+    def decrypt_messages(self, encryption_key: bytes, messages: list[Message]) -> list[str]:
         """
         Decrypts the `MessagePayload` list using the provided key and verifies the message authenticity
 
+        The key argument is named ``encryption_key`` so Sentry's ``EventScrubber`` denylist scrubs it
```
SmittieC, line 81: I'll rather make this a comment as apposed to a docstring.

SmittieC (approved):

PR #3846, opened 2026-07-20
refactor: base the team sync schema check on the export API schema

SmittieC:
### Product Description
No user-facing change. Internal cleanup of the team data sync (`sync_team`).

### Technical Description
Replaces the migration-state checksum that gated the team sync with a checksum of the export API's OpenAPI schema.

**Why:** migration history is a poor proxy for schema compatibility. A squash or a rename changes a server's recorded migrations without changing its schema, so the old check rejected syncs between servers whose schemas were actually identical. The export OpenAPI schema is the opposite: it is generated from the running code and describes the exact shape of the data the sync transfers (endpoints, fields, types), so the source and target agree on the checksum exactly when their export surfaces match — a direct check that both servers run compatible code.

Concretely:

- `schema_checksum()` in `apps/teams/export/manifest.py` now hashes the export API's OpenAPI document (`SchemaGenerator(api_version="export")`, no request so host placeholders stay fixed, sorted keys). Verified byte-identical across processes and `PYTHONHASHSEED` values.
- The manifest keeps its `schema_checksum` field, and `check_sync_preconditions` keeps the source/target comparison and the `--skip-schema-check` override; only the basis of the checksum changes.
- The checksum is cached per process (the schema can't change while the code is running).
- Regenerates `api-schemas/export.yml` (field description).

### Migrations

No migrations in this change.

### Demo
N/A

### Docs and Changelog

### DIFF
--- a/api-schemas/export.yml
+++ b/api-schemas/export.yml
@@ -4321,10 +4321,12 @@ components:
     Manifest:
       type: object
       description: 'The sync manifest: the call order and per-model config, plus a
-        checksum of the applied schema.'
+        checksum of the export schema.'
       properties:
         schema_checksum:
           type: string
+          description: SHA-256 of this server's export OpenAPI schema; clients compare
+            it against their own to detect an incompatible export surface.
         entries:
           type: array
           items:
--- a/apps/api/export/serializers.py
+++ b/apps/api/export/serializers.py
@@ -144,9 +144,8 @@ class ManifestEntrySerializer(serializers.Serializer):
 
 
 class ManifestSerializer(serializers.Serializer):
-    """The sync manifest: the call order and per-model config, plus a checksum of the applied schema."""
+    """The sync manifest: the call order and per-model config."""
 
-    schema_checksum = serializers.CharField()
     entries = ManifestEntrySerializer(many=True)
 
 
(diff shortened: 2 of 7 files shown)


SmittieC:
@snopoke thoughts on this?

snopoke:
I think it's OK but there's now way for anyone to know what the latest version of a specific server is right? Is there a way we can expose that info somehow?

SmittieC:
> I think it's OK but there's now way for anyone to know what the latest version of a specific server is right? Is there a way we can expose that info somehow?

I couldn't find anything robust enough, except looking at the latest commit SHA, but I don't really want to do that. Would it be really bad if we just tell people to ensure they run the latest code before running this?

snopoke:
> > I think it's OK but there's now way for anyone to know what the latest version of a specific server is right? Is there a way we can expose that info somehow?
> 
> I couldn't find anything robust enough, except looking at the latest commit SHA, but I don't really want to do that. Would it be really bad if we just tell people to ensure they run the latest code before running this?

The issue is that we deploy often so running the latest code is not necessarily straight forward. What if we used the export API schema as a proxy? The source and target schema's should match right?

SmittieC:
> The issue is that we deploy often so running the latest code is not necessarily straight forward. What if we used the export API schema as a proxy? The source and target schema's should match right?

My only concern with this is if there are migration mismatches between the servers that alters certain field constraints (as apposed to fields themselves), but I guess this doesn't happen that often and is more of an edge case. Hmm, ok thanks, let's go with this approach then

snopoke (approved):

PR #3865, opened 2026-07-21
Add durable per-channel widget auth policy (required_auth_level)

### LINKED ISSUE #3858: Chat Widget API: require authenticated sessions by default, gate legacy unauthenticated fallback per-channel
## Background

The chat widget API (`/api/chat/start/`, `/api/chat/<session>/send/`, etc.) has accumulated layered auth behaviour across three widget generations:

| Widget version | Headers sent | Current behaviour |
|---|---|---|
| **< 0.5.1** (`widget_version == "unknown"`) | none | No embed key, no session token. Access allowed via `is_public` / allowlist fallback. |
| **0.5.1 – 0.8.x** | `X-Embed-Key` | Embed key validated, domain-checked. Session started with `session_token_required=False`. |
| **0.9.0+** | `X-Embed-Key`, `X-Session-Token` | Full token-enforced flow (`use_session_token=True`). |

The problem is that the server-side logic infers which mode to use from per-request heuristics (`is_widget_request`, presence of `use_session_token` in the body) rather than from a durable per-channel policy. A brand-new channel created today is treated the same as a legacy channel running a 2024 widget — it gets the permissive fallback unless the client explicitly opts in.

### Key code paths

- **`_issue_or_opt_out_session_token`** (`apps/api/views/chat.py`): if `use_session_token` is not set by the client and the request looks like widget traffic, the session is written with `session_token_required=False` — permanently opting it out of token enforcement.
- **`SessionAccessPermission._has_legacy_access`** (`apps/api/permissions.py`): for `session_token_required=False` sessions, allows requests that have a valid embed key, or that target a public experiment.

---

### Goal

Encode the minimum required auth level as a durable **per-channel policy** rather than inferring it at request time. New channels default to the strictest level. Existing channels are migrated to the level that matches the widget version they are actually running.

### New field: `required_auth_level`

Add an integer (or char) field to `ExperimentChannel` with three levels:

```python
class WidgetAuthLevel(models.IntegerChoices):
    NONE = 0, "None (pre-0.5.1 legacy)"
    EMBED_KEY = 1, "Embed key only (0.5.1 – 0.8.x)"
    SESSION_TOKEN = 2, "Session token required (0.9.0+)"
```

`default=SESSION_TOKEN`. Only meaningful for `EMBEDDED_WIDGET` channels.

### Enforcement

| Level | What the server enforces |
|---|---|
| `NONE` | Allows the current `_has_legacy_access` path (embed key optional, `is_public` fallback permitted). |
| `EMBED_KEY` | Requires a valid `X-Embed-Key` + domain check; no session token needed. `is_public` fallback blocked. |
| `SESSION_TOKEN` | Requires valid `X-Embed-Key` + `X-Session-Token`. No legacy path reachable. |

`_issue_or_opt_out_session_token` should consult the channel's level rather than the client's `use_session_token` hint:
- `NONE` → opt out (legacy)
- `EMBED_KEY` → opt out but enforce embed key
- `SESSION_TOKEN` → always issue and enforce token

### Migration

Use the existing `widget_version` field on `ExperimentChannel` to assign the right level to every existing `EMBEDDED_WIDGET` channel:

| `widget_version` value | Assigned level | Rationale |
|---|---|---|
| `"unknown"` (pre-0.5.1, never reported a version) | `NONE` | Widget predates embed key entirely |
| parseable version `< 0.5.1` | `NONE` | Same |
| parseable version `0.5.1 ≤ v < 0.9.0` | `EMBED_KEY` | Widget sends embed key but not session token |
| parseable version `>= 0.9.0` | `SESSION_TOKEN` | Widget fully supports token flow |
| `null` (channel has never had a widget connect) | `SESSION_TOKEN` | Treat as new |

Channels that are already running a modern widget (≥ 0.9.0) are **not** grandfathered into a weaker level.

### Admin / UI

Expose `required_auth_level` in the channel edit form and admin with a human-readable explanation of what each level means. Teams can manually downgrade a specific channel if they know it's legacy.

---

## Acceptance Criteria

- [ ] New `EMBEDDED_WIDGET` channels default to `SESSION_TOKEN`; unauthenticated or token-less requests receive a clear error.
- [ ] Channels migrated to `NONE` continue to work without any client changes.
- [ ] Channels migrated to `EMBED_KEY` require a valid embed key but no session token.
- [ ] Channels already on widget ≥ 0.9.0 are migrated to `SESSION_TOKEN` — not grandfathered into a weaker level.
- [ ] Channels with `widget_version=null` (never connected) get `SESSION_TOKEN`.
- [ ] `required_auth_level` is visible and editable in admin/channel UI.
- [ ] Tests cover all three levels: correct enforcement, correct rejection, level-upgrade path.

---

## Files of Interest

- `apps/api/views/chat.py` — `_issue_or_opt_out_session_token`, `_resolve_experiment_channel`
- `apps/api/permissions.py` — `SessionAccessPermission`, `WidgetDomainPermission`
- `apps/api/authentication.py` — `EmbeddedWidgetAuthentication`
- `apps/channels/models.py` — `ExperimentChannel`, `ChannelPlatform.EMBEDDED_WIDGET`, `widget_version` field
- `apps/channels/widget_versions.py` — `is_widget_request`, `UNKNOWN_WIDGET_VERSION`
- `apps/experiments/models.py` — `ExperimentSession.session_token_required`

snopoke:
### Product Description
Chat widget channels now carry a durable per-channel authentication policy. New embedded-widget channels require a session token by default; legacy channels are migrated to the level matching the widget version they actually run. Closes #3858.

### Technical Description
Adds `WidgetAuthLevel` (NONE/EMBED_KEY/SESSION_TOKEN) and `ExperimentChannel.required_auth_level` (default SESSION_TOKEN). `chat_start_session` consults the channel level when issuing/opting out of a session token; `SessionAccessPermission` blocks the public/allowlist fallback for widget channels at EMBED_KEY+. Migration 0029 grandfathers existing channels from their reported `widget_version`. Field is editable in the channel form and visible in admin.

### Migrations

### Docs and Changelog


apps/channels/forms.py
```diff
@@ -631,9 +631,24 @@ class EmbeddedWidgetChannelForm(ExtraFormBase):
         help_text="Configuration parameters for the widget",
     )
 
+    required_auth_level = forms.TypedChoiceField(
```
snopoke: This should not be visible or editable by users.

apps/api/permissions.py
```diff
@@ -118,6 +118,15 @@ def _has_legacy_access(self, request, session) -> bool:
             # widget-authed requests rely on the embed key + domain check
             return True
```
snopoke: won't this check bypass the auth level check?

apps/api/views/chat.py
```diff
@@ -177,23 +177,40 @@ def chat_upload_file(request, session_id):
...
-    x-ocs-widget-version header (sent by 0.5.1+) or, for older widgets that send no
-    header, by `session_data.source`. Everything else (direct API consumers) defaults
-    to enforced. Returns the token, or None when opted out.
+    For embedded widget channels the decision comes from the channel's durable
+    `required_auth_level` policy, not from per-request heuristics:
+      - SESSION_TOKEN → always issue and enforce a token
+      - EMBED_KEY / NONE → opt out (the embed key, if required, is enforced by the
+        permission classes)
+
+    For every other caller (direct API consumers, pre-0.5.1 widgets that never resolve
```
snopoke: we do not need support direct API consumers for this API

apps/channels/forms.py
snopoke: Fixed in e35e7adcc. Removed `required_auth_level` from the embedded widget channel form entirely — the level is now system-managed (model default + grandfathering migration 0029) and no longer user-editable. It stays visible read-only in the Django admin.

*— Claude Code*

apps/api/permissions.py
snopoke: Good catch — it did. Fixed in a6d312e73. `_has_legacy_access` now gates the embed-key branch on the channel level: a valid embed key satisfies EMBED_KEY/NONE channels but never a SESSION_TOKEN channel, even on a session left `session_token_required=False`. Added a defense-in-depth test.

*— Claude Code*

apps/api/views/chat.py
snopoke: Fixed in 1bf91d23c. Dropped the direct-API-consumer / version-sniff fallback and removed the `use_session_token` request field; token issuance is now driven solely by the channel's `required_auth_level` (non-widget channels default to issuing a token). This supersedes ADR-0041's opt-out mechanism, documented in new ADR-0044.

*— Claude Code*

apps/api/permissions.py
```diff
@@ -114,9 +114,23 @@ def _token_grants_access(self, request, session) -> bool:
...
+            # WidgetDomainPermission) satisfies EMBED_KEY and NONE channels. It never
+            # satisfies a SESSION_TOKEN channel — that always requires the token, even
+            # if the session was (mis)configured with session_token_required=False.
+            return level != WidgetAuthLevel.SESSION_TOKEN
+
+        # No embed key. At EMBED_KEY and above a valid embed key is mandatory, so the
+        # public / allowlist fallback is only reachable for NONE-level widget channels
+        # (and non-widget sessions, where level is None).
+        if level is not None and level != WidgetAuthLevel.NONE:
+            return False
```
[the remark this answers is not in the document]
snopoke: Fixed in ab2439611. Confirmed real: `EmbeddedWidgetAuthentication` resolves `request.auth` to any EMBEDDED_WIDGET channel of the *session's experiment* whose token matches, and `WidgetDomainPermission` validates the origin against `request.auth`'s own domains — so a second widget channel on the same experiment (EMBED_KEY/NONE) could be reached with an unrelated channel's key + domain. (Cross-*experiment* access was already blocked since the key lookup is scoped to the session's experiment; the SESSION_TOKEN case was closed earlier.)

`_has_legacy_access` now requires `request.auth == session.experiment_channel` before granting the widget-authed branch. Added `test_embed_key_for_other_channel_denied` covering the two-channels-one-experiment case.

*— Claude Code*

Issue #3861, opened 2026-07-21
Deleting an Evaluator destroys results and aggregates from completed historical runs


barry47products:
### Context

Split out of #3385, which guards evaluation deletes against runs that are still in flight. This one is different: it fires when nothing is running.

### Problem

Delete an evaluator and its history goes with it, from every completed run that ever used it. The confirm dialog says "This will permanently delete the evaluator. Are you sure?" (`apps/evaluations/tables.py:200`) and the delete returns 200. Nothing tells you that results are about to go.

`Evaluator.delete()` (`apps/evaluations/models.py:135-138`) only refuses while a related run is `PENDING` or `PROCESSING`. Once every run is terminal it hands straight off to the database cascades, and those don't care about run status.

### What is lost

I originally had this down as results and aggregates. It's wider than that. Everything below is `CASCADE`:

| Row | Path from `Evaluator` |
| --- | --- |
| `EvaluationResult` | `EvaluationResult.evaluator` (`models.py:777`) |
| `EvaluationRunAggregate` | `EvaluationRunAggregate.evaluator` (`models.py:820`) |
| `EvaluatorTagRule` | `EvaluatorTagRule.evaluator` (`models.py:831`) |
| `AppliedTag` | `AppliedTag.rule` (`models.py:859`) |
| `Score` (assessments) | `Score.automated_result` (`apps/assessments/models.py:43-49`) |

So the applied-tag history and the judge-side concordance scores go too, not just the results table.

The `EvaluationRun` row survives, which is almost worse. Its `evaluator_ids` (`models.py:717`) is a plain integer array, not a foreign key, so the run keeps pointing at an id that no longer exists. The results page builds its columns from `Evaluator.objects.filter(id__in=run.evaluator_ids)` (`apps/evaluations/views/evaluation_config_views.py:437`), which is why the column just disappears instead of erroring. You'd have to already know the evaluator was there to notice it's gone.

### Steps to reproduce

1. Create an evaluator and add it to an evaluation config.
2. Run the config and wait for `COMPLETED`.
3. Open the run results page. Note the evaluator's column and aggregate.
4. Delete the evaluator from the evaluators list. The dialog mentions only the evaluator, and the delete succeeds.
5. Reopen the run results page.

Expected: the run still shows the evaluator's results and aggregate, or the delete is refused and says why.

Actual: the column, the aggregate, the applied tags and any scores derived from those results are gone. The run still says `COMPLETED`.

[...]

PR #3867, opened 2026-07-21
Guard destructive evaluation deletes against in-flight runs

### LINKED ISSUE #3385: Guard destructive evaluation operations against in-flight runs
### Context

Spun out of the design discussion on #3203 (see https://github.com/dimagi/open-chat-studio/issues/3203#issuecomment-).

If an `EvaluationRun` is `PROCESSING` when its parent `EvaluationConfig` is deleted, the `CASCADE` wipes the run row and the Celery worker crashes on the next ORM access. The same gap exists for `DeleteEvaluator` and `DeleteDataset` — any destructive operation on an evaluation resource while a run is in flight can corrupt the run.

### Possible approaches

- UI guard (hide / disable delete while runs are processing)
- Server-side 409 if any related run is in a non-terminal state
- Soft-delete pattern (mark deleted, defer cascade until runs finish)

Probably worth scoping across all three delete views together rather than fixing piecemeal.

### Acceptance

- A user cannot delete an `EvaluationConfig`, `Evaluator`, or `Dataset` in a way that strands a `PROCESSING` `EvaluationRun`.
- Behaviour is consistent across the three resources.

barry47products:
### Product Description

Deleting an evaluation, evaluator, dataset, or dataset message, or clearing an evaluation's run history, while an evaluation run is still in progress could strand or crash that background run. This PR blocks those destructive actions while a related run is `PENDING` or `PROCESSING`: the action is rejected with a clear message ("Cannot delete this evaluation while evaluation runs are in progress. Wait for the runs to finish, then try again.") shown in the standard error modal, and nothing is deleted. Once the runs finish (or fail), the same action succeeds normally.

Fixes #3385.

### Technical Description

A shared guard lives in `apps/evaluations/models.py`: `NON_TERMINAL_RUN_STATUSES` plus `raise_if_runs_in_flight(runs, resource_label)`, which raises a new `InFlightRunsError` (in `apps/evaluations/exceptions.py`, subclassing `ValidationError`) when any run in the given queryset is non-terminal.

- **Model `delete()` overrides** on `EvaluationConfig`, `Evaluator`, `EvaluationDataset`, and `EvaluationMessage` call the guard before `super().delete()`, each with the query that is exactly the set of runs the delete would cascade-destroy:
  - config: `config=self`
  - evaluator (M2M): `config__evaluators=self`
  - dataset (FK cascade): `config__dataset=self`
  - message: `Q(config__dataset__messages=self) | Q(scoped_messages=self)` (dataset membership for FULL/PREVIEW runs, and DELTA `scoped_messages`)
- **Views** (`DeleteEvaluation`, `DeleteEvaluator`, `DeleteDataset`, `delete_message`) catch `InFlightRunsError` and return `HttpResponse(", ".join(e.messages), status=409)`. 409 is required: the global htmx handler surfaces a 4xx body in a modal, and because htmx only swaps on 2xx, a table-row delete does not replace the row with error text.
- **`ClearEvaluationRuns`** deletes via `queryset.delete()` (which bypasses `Model.delete()`), so it calls the guard directly over a **pinned id set** (`run_ids = list(...values_list("id"))`, then `filter(id__in=run_ids)` for the guard, tag-removal, and delete) so all three operate on the identical frozen set.
- **Celery task hardening** (`apps/evaluations/tasks.py`): `evaluate_single_message_task` guards its run and message fetches, and `run_evaluation_task`'s failure handler re-fetches with `.filter().first()` instead of `.get()`, so the residual race/cascade paths the model guard cannot cover degrade to a logged no-op instead of an unhandled task failure (previously the failure handler could raise a second `DoesNotExist`, masking the original error).

**Divergence from the `LlmProviderModel` guarded-delete precedent** (`apps/service_providers/`): a dedicated `InFlightRunsError` subclass (vs plain `ValidationError`) so views catch precisely, and HTTP **409** (vs the precedent's 400) per @snopoke's decision on the issue.

Tests are behaviour tests through the view (`client.delete`/`client.post`, asserting status code + DB state), matching the existing delete-test files; task tests call the task directly and assert observable DB state. The two defensive task branches and the message `scoped_messages` guard clause are each mutation-verified (removing the guard fails the corresponding test).

### Migrations

No migrations. The change is model methods + a module-level constant + an exception class; no schema change (`makemigrations --check` reports "No changes detected").

### Known Limitations

1. **TOCTOU race (residual, mitigated).** The guard is check-then-delete with no lock, and run creation also takes no lock, so a run entering `PENDING` between the guard SELECT and the delete is not caught (millisecond window; most likely via `auto_run_on_append` or two users acting at once). The task hardening turns the worst case (a run cascade-deleted mid-flight) into a logged no-op. Fully closing it would need `select_for_update` on both the delete and run-creation paths; the cited `LlmProviderModel` precedent has the same gap.
2. **Admin single-object delete now 500s.** `ModelAdmin.delete_model` calls `obj.delete()`, which raises `InFlightRunsError` -> unhandled 500 in the Django admin. Blocking is safer than silently stranding a run, so this is accepted.
3. **Message/PREVIEW/DELTA over-block.** A message is blocked if any in-flight run's dataset contains it, even for a PREVIEW/DELTA run that will not actually process it. Conservative and safe (never strands a run); in-flight runs are short-lived.
4. **Completed-run history still cascades (out of scope).** Deleting a message or evaluator still removes `EvaluationResult`/aggregate rows belonging to *terminal* runs (those FKs are CASCADE). This guard only protects in-flight runs; the historical-data-loss behaviour is unchanged and tracked separately in #3861.

### Demo

Verified end-to-end in the running app: with a `PROCESSING` run attached, deleting the config, deleting the dataset, and clearing runs each return 409 and show the error modal with the object untouched; flipping the run to a terminal status lets the same actions succeed. Backend paths additionally verified via authenticated requests and a runtime demonstration of the task hardening degrading to a logged no-op.

### Docs and Changelog

### DIFF
--- a/apps/evaluations/exceptions.py
+++ b/apps/evaluations/exceptions.py
@@ -1,6 +1,13 @@
+from django.core.exceptions import ValidationError
+
+
 class EvaluationRunException(Exception):
     pass
 
 
 class HistoryParseException(Exception):
     pass
+
+
+class InFlightRunsError(ValidationError):
+    """Raised when a delete is blocked because related EvaluationRuns are still in progress."""
--- a/apps/evaluations/models.py
+++ b/apps/evaluations/models.py
@@ -152,6 +152,13 @@ class EvaluationMessage(BaseModel):
         output_content = self.output.get("content", "no content")
         return f"{input_role}: {input_content}, {output_role}: {output_content}"
 
+    def delete(self, *args, **kwargs):
+        related_runs = EvaluationRun.objects.filter(
+            models.Q(config__dataset__messages=self) | models.Q(scoped_messages=self)
+        )
+        raise_if_runs_in_flight(related_runs, "message")
+        return super().delete(*args, **kwargs)
+
     @classmethod
     def create_from_sessions(
         cls, team: Team, external_session_ids, filtered_session_ids=None, filter_params=None, timezone=None
--- a/apps/evaluations/models.py
+++ b/apps/evaluations/models.py
@@ -15,6 +15,7 @@ from pydantic import BaseModel as PydanticBaseModel
 
 from apps.chat.models import ChatMessage, ChatMessageType
 from apps.chatbots.version_resolver import VersionSelectionRule, resolve_chatbot_version
+from apps.evaluations.exceptions import InFlightRunsError
 from apps.evaluations.export import build_evaluation_table_data
 from apps.evaluations.rule_validation import (
     ConditionType,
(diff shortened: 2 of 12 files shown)


barry47products:
@snopoke kindly review 👀

snopoke (approved):

Issue #2681, opened 2026-01-21
Extend RAG pipeline: add contextual retrieval, hybrid search, and reranking


snopoke:
We currently have a basic RAG pipeline. This issue proposes extending it with the techniques described in Anthropic’s “Contextual Retrieval”:

https://www.anthropic.com/engineering/contextual-retrieval

Why
Anthropic shows that prepending concise, chunk-specific explanatory context before embedding and indexing (Contextual Embeddings + Contextual BM25) reduces retrieval failures significantly, and that adding a reranker further improves results. These techniques should reduce hallucination risk and improve answer relevance for our knowledge-heavy workflows.

Proposed work
- Add contextualization preprocessing:
  - Generate concise, chunk-specific context (50–100 tokens) that situates each chunk within its source document and prepend it before embedding and BM25 indexing.
  - Provide a configurable contextualizer (use model-backed generation with prompt caching or a local summarizer).
- Implement hybrid search (dense + lexical):
  - Keep our dense embeddings search and add lexical search (Elasticsearch/OpenSearch/BM25).
  - Merge results with rank-fusion and dedupe.
- Add a reranking stage:
  - Optionally rerank top-N candidates (e.g., 150 → rerank → top-20) using a cross-encoder or lightweight reranker (Cohere/Voyage/custom).
  - Make reranking configurable to trade off cost/latency vs accuracy.
- Contextual retrieval API surface:
  - Extend retrieval API to accept a context object (recent conversation turns, intent, user metadata) and use it to condition retrieval.
  - Provide examples showing how to pass conversation state and system prompts.
- Evaluation + metrics:
  - Add end-to-end evals that measure retrieval recall@K, failure rate (1 - recall@K), and downstream answer quality on representative datasets.
  - Collect telemetry: retrieval latency, reranker cost, hit rates, and changes in hallucination rate.
- Docs + examples:
  - Architecture doc explaining preprocessing, indexing, and runtime flow.
  - Cookbook-style example using prompt-caching or our chosen contextualizer.
- Backwards compatibility & rollout:
  - Provide a feature flag to enable/disable contextualization and reranking.
  - Migration plan for reindexing existing corpora.

Acceptance criteria
- Contextualized chunks stored in vector + BM25 indexes (configurable contextualizer).
- Hybrid retrieval returns combined, deduplicated top-K candidates.
- Reranker integrates and is configurable (on/off + top-N).
[...]

PR #3875, opened 2026-07-21
Phase 1: Contextual retrieval — storage, contextualizer, and ingestion wiring (#2681)

uthy4r:
Implements Phase 1 of the RAG pipeline extension (issue #2681), following the approved design doc. Ships the storage, generation, and wiring layers for contextual retrieval end to end, all behind a feature flag and a per-collection setting, both defaulting off.

## What this PR does

1. **Storage.** New `FileChunkEmbedding.context` field (blank TextField) and `contextualized_text` property that prefixes the context to the chunk when present, otherwise returns the raw text unchanged.
2. **Generation.** New `Contextualizer` interface in `apps/service_providers/llm_service/contextualizer.py` with two implementations:
   - `StaticContextualizer`: zero-cost, builds a header from the file name and page number.
   - `LLMContextualizer`: calls the collection's chat model with an Anthropic-style prompt; falls back to static on any error so indexing never dies.
3. **Wiring.** `LocalIndexManager.add_files` generates a per-chunk context, stores it, and embeds the contextualized text (context + chunk) instead of the raw chunk. Collections gain `enable_contextual_retrieval` and `contextualizer_strategy` fields. A waffle flag `flag_contextual_retrieval` gates the whole feature.

## Backwards compatibility

Everything defaults off. With the flag off or `enable_contextual_retrieval` unset, `add_files` produces byte-identical results to the current code path — proven by the existing 37 index-manager tests still passing unchanged, plus 3 new tests covering the wiring and the "no context when strategy is None" fallback.

The schema changes are additive only: one new nullable column on `FileChunkEmbedding`, two new columns on `Collection` with safe defaults. No existing data is touched.

## Testing

- 8 unit tests for the contextualizer implementations (fallback paths, truncation, prompt content).
- 3 new integration tests in `test_index_managers.py` (context stored when enabled, no context when disabled, embedded input includes context).
- End-to-end verification against a real database confirmed context is generated and stored during the full ingestion path.

## Deliberate scope decisions — reviewer input welcome

1. **Only the static strategy is wired.** The `llm` choice is defined in the enum and the class exists, but `_contextualizer_strategy` currently returns "static" for both settings. Fully wiring the LLM path requires deciding how to resolve a chat model name for the collection's `llm_provider` (the codebase has no default-model concept per provider). Options: add a `contextualizer_llm_model` FK to `Collection`; hardcode a per-provider default; require the user to pick a model explicitly. Happy to add whichever you prefer as a follow-up.
2. **Versioning and audit.** The new settings fields are not part of `_get_version_details` (version snapshots) or `@audit_fields` (change audit) on `Collection`. Reasonable arguments both ways; happy to add if you'd like these tracked.
3. **Flag check outside a request.** Indexing runs in a Celery task, so `_contextualizer_strategy` uses `Flag.is_active_for_team(self.team)` plus explicit `everyone` precedence rather than `flag_is_active(request, ...)`. Discovered during end-to-end testing that `flag_is_active(None, ...)` doesn't reliably fire team-scoped flags in a no-request context.

## What's next (Phase 2 / 3)

Hybrid search (RRF over dense + Postgres FTS) and optional reranking follow in separate PRs per the phased delivery model. Each behind its own flag, each additive.

## Design doc

The approved design doc is the source of truth for the overall architecture; this PR is Phase 1 of that plan.


apps/documents/models.py
```diff
@@ -75,6 +75,11 @@ def status_enum(self):
         return FileStatus(self.status)
 
 
+class ContextualizerStrategy(models.TextChoices):
+    STATIC = "static", _("Static (no LLM, zero cost)")
```
snopoke: I don't think this mode is useful - we could probably drop the `contextualizer_strategy` field entirely

apps/service_providers/llm_service/contextualizer.py
```diff
@@ -0,0 +1,96 @@
...
+    ):
+        self._chat_model = chat_model
+        self._fallback = fallback or StaticContextualizer()
+        self._max_document_chars = max_document_chars
+
+    def get_context(self, *, document: str, chunk: str) -> str:
+        truncated = document[: self._max_document_chars]
+        messages = [
+            ("system", CONTEXT_SYSTEM_PROMPT),
+            ("human", CONTEXT_USER_PROMPT.format(document=truncated, chunk=chunk)),
```
snopoke: you might get better caching if you put the truncated doc into the system prompt

snopoke (commented):
@uthy4r thanks for the contribution. I've made some initial comments. Please also review the automated review comments.

apps/service_providers/llm_service/contextualizer.py
uthy4r: Done, the document now goes in the system prompt and only the chunk varies in the human message, so the document prefix can be cached across all chunks of the same file. Matches the pattern from Anthropic's contextual retrieval post.

apps/documents/models.py
uthy4r: Agreed, dropped it. Removed the ContextualizerStrategy enum and the contextualizer_strategy field entirely. Contextual retrieval now just means LLM contextualization; StaticContextualizer stays only as the internal fallback when the LLM call fails. To resolve the chat model I added a contextualizer_llm_model FK to Collection (nullable), so the model is configured explicitly per collection rather than guessed. Open to a different approach here if you'd prefer resolving the model some other way.

apps/documents/models.py
```diff
@@ -321,7 +334,49 @@ def get_index_manager(self):
...
+            active = True
+        elif flag.everyone is False:
+            active = False
+        else:
+            active = flag.is_active_for_team(self.team)
+        if not active:
+            return None
+        try:
+            service = self.llm_provider.get_llm_service()
+            chat_model = service.get_chat_model(self.contextualizer_llm_model.name)
```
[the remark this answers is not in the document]
uthy4r: The silent fallback here is intentional: contextualization must never abort indexing, so if the provider can't produce a chat model we skip it rather than fail the file. I've added a warning log so the misconfiguration is diagnosable rather than invisible. I'd avoid raising a user-facing error at this point since it runs in a background indexing task, but I'm happy to add form-level validation on the collection config (rejecting an embedding-only provider when contextual retrieval is enabled) as a follow-up if you'd prefer catching it at configuration time.

apps/service_providers/llm_service/index_managers.py
```diff
@@ -318,12 +327,16 @@ def add_files(
                         "Failed to update collection file status", extra={"collection_file_id": collection_file_id}
                     )
 
-    def chunk_file(self, file: File, chunk_size: int, chunk_overlap: int) -> list[str]:
+    def chunk_file(
+        self, file: File, chunk_size: int, chunk_overlap: int, text: str | None = None
+    ) -> list[str]:
```
snopoke: This is only called in one place (outside tests) so I think it makes more sense to drop the file argument entirely and always pass in text directly

apps/service_providers/llm_service/contextualizer.py
```diff
@@ -0,0 +1,100 @@
...
+        """Return a short context string for the chunk, or "" if none could be produced."""
+
+
+class StaticContextualizer(Contextualizer):
+    """Zero-cost contextualizer that builds a header from document structure.
+
+    Used as the fallback when the LLM contextualizer fails. Takes no external calls.
+    """
+
+    def __init__(self, *, file_name: str = "", page_number: int | None = None):
```
snopoke: The arguments are never passed in any of the usages. I would suggest dropping this class and just returning a blank string when there is an error.

apps/service_providers/tests/test_index_managers.py
```diff
@@ -63,128 +64,70 @@ def delete_files(self, *args, **kwargs): ...
...
             iterator = CollectionFile.objects.filter(id=collection_file.id).iterator(1)
-            index_manager.add_files(iterator)
+            contextualizing_index_manager.add_files(iterator)
 
-        assert spy.call_count == 2
+        assert spy.call_count > 0
         for call in spy.call_args_list:
-            assert call.kwargs == {"input_type": "document"} or call.args[1:] == ("document",)
-
-    def test_add_files_skips_chunks_that_are_empty_after_nul_stripping(self, local_index_instance, index_manager):
```
snopoke, line 162: why did these tests get deleted - are they no longer relevant?

snopoke (commented):
Few small comments but otherwise looks good.

apps/service_providers/tests/test_index_managers.py
uthy4r: Good catch, that deletion was accidental, not intentional. Restored the full TestLocalIndexManager class unchanged.

apps/service_providers/llm_service/contextualizer.py
uthy4r: Dropped the class. LLMContextualizer now returns an empty string on error, so indexing continues without a context header.

apps/service_providers/llm_service/index_managers.py
uthy4r: Done, chunk_file now takes the text directly and the file argument is gone.

apps/documents/migrations/0017_collection_contextualizer_llm_provider.py
snopoke, line 1: @uthy4r please merge this migration into the previous one

apps/documents/migrations/0018_merge_20260727_1530.py
snopoke, line 1: This migration can be removed and you can resolve the merge issue by renaming the migration in your branch.

apps/service_providers/llm_service/index_managers.py
```diff
@@ -318,12 +325,14 @@ def add_files(
...
-    def chunk_file(self, file: File, chunk_size: int, chunk_overlap: int) -> list[str]:
+    def chunk_file(self, text: str, chunk_size: int, chunk_overlap: int) -> list[str]:
         """
         Split text content into overlapping chunks for processing.
 
         Args:
-            text: The text content to be chunked.
+            text: Optional pre-read file content. Passed by callers that already
+                have the content in memory (e.g. contextual indexing) to avoid a
+                second read.
```
snopoke:
```suggestion
            text: File content as text.
```

apps/documents/migrations/0018_merge_20260727_1530.py
uthy4r: Removed. Renumbered ours to 0017 so it depends on documents.0016_documentsourcesynclog_files_failed_and_more, giving a linear history with no merge migration.

apps/documents/migrations/0017_collection_contextualizer_llm_provider.py
uthy4r: Done, the two are now a single migration.

snopoke (dismissed):

snopoke:
@uthy4r this looks good to go once you resolve the lint issues

snopoke:
@uthy4r the type check is still failing:

```
error[invalid-method-override]: Invalid override of method `chunk_file`
   --> apps/service_providers/tests/test_failed_file_retrieval.py:34:9
    |
 34 |     def chunk_file(self, file, chunk_size=None, chunk_overlap=None):
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Definition is incompatible with `LocalIndexManager.chunk_file`
    |
   ::: apps/service_providers/llm_service/index_managers.py:340:9
    |
340 |     def chunk_file(self, text: str, chunk_size: int, chunk_overlap: int) -> list[str]:
    |         ----------------------------------------------------------------------------- `LocalIndexManager.chunk_file` defined here
    |
info: the parameter named `file` does not match `text` (and can be used as a keyword parameter)
info: This violates the Liskov Substitution Principle
```

uthy4r:
> @uthy4r the type check is still failing:
> 
> ```
> error[invalid-method-override]: Invalid override of method `chunk_file`
>    --> apps/service_providers/tests/test_failed_file_retrieval.py:34:9
>     |
>  34 |     def chunk_file(self, file, chunk_size=None, chunk_overlap=None):
>     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Definition is incompatible with `LocalIndexManager.chunk_file`
>     |
>    ::: apps/service_providers/llm_service/index_managers.py:340:9
>     |
> 340 |     def chunk_file(self, text: str, chunk_size: int, chunk_overlap: int) -> list[str]:
>     |         ----------------------------------------------------------------------------- `LocalIndexManager.chunk_file` defined here
>     |
> info: the parameter named `file` does not match `text` (and can be used as a keyword parameter)
> info: This violates the Liskov Substitution Principle
> ```



Fixed, that mock in test_failed_file_retrieval.py still had the old parameter name. Pushed in 6cbc771.

snopoke (dismissed):

snopoke:
@uthy4r there are still other failures:

### API schema needs to be updated: 

```shell
inv schema
# OR
python manage.py spectacular --api-version export --file api-schemas/export.yml --validate
```

### Hoist inline imports that don't need to be inline

```
FAIL apps/documents/models.py:415: `from apps.service_providers.exceptions import ServiceProviderConfigError` hoists cleanly
     Move it to module level
FAIL apps/documents/models.py:457: `from apps.teams.flags import Flags` hoists cleanly
     Move it to module level
FAIL apps/documents/models.py:458: `from apps.teams.models import Flag` hoists cleanly
     Move it to module level

uthy4r:
> @uthy4r there are still other failures:
> 
> ### API schema needs to be updated:
> ```shell
> inv schema
> # OR
> python manage.py spectacular --api-version export --file api-schemas/export.yml --validate
> ```
> 
> ### Hoist inline imports that don't need to be inline
> ```
> FAIL apps/documents/models.py:415: `from apps.service_providers.exceptions import ServiceProviderConfigError` hoists cleanly
>      Move it to module level
> FAIL apps/documents/models.py:457: `from apps.teams.flags import Flags` hoists cleanly
>      Move it to module level
> FAIL apps/documents/models.py:458: `from apps.teams.models import Flag` hoists cleanly
>      Move it to module level
> ```

Both fixed: API schema regenerated and the three inline imports hoisted to module level. I've since installed the full pre-commit suite locally, ruff, ruff-format, djade, djlint, eslint, prettier, actionlint and the local hooks all pass on the changed files, plus 898 tests. The only check I can't verify locally is ty, since its configured .venv path doesn't exist in my Docker setup.

snopoke (approved):

Issue #3851, opened 2026-07-20
Usage API: explicit start/end window and granularity bucketing


snopoke:
## What to build

Add explicit time-window and bucketing support to `GET /api/v2/usage/` (design: `docs/design/usage-api.md`):

- `start` / `end` (ISO date/datetime), half-open `[start, end)`, mutually exclusive with `period`.
- `granularity` = `total` (default) / `daily` / `weekly` / `monthly`; when not `total` and ungrouped, `results` is one row per bucket carrying `bucket_start`, computed in `tz`.

Guardrails: a max-window guard relative to granularity (reject e.g. `daily` over a multi-year range) and a scoped DRF throttle class for the endpoint.

## Acceptance criteria

- [ ] `start`/`end` define a half-open window; passing both `period` and `start`/`end` 400s
- [ ] `granularity=daily|weekly|monthly` returns one row per bucket with `bucket_start`, `tz`-aware
- [ ] `granularity=total` (default) preserves the single-object behaviour
- [ ] Over-large window + fine granularity is rejected with a clear 400
- [ ] A scoped throttle is applied to the endpoint
- [ ] Tests cover window math, bucket boundaries under a non-UTC `tz`, and the guards

## Blocked by

- #3848

Issue #3848, opened 2026-07-20
Usage API: participant message count for current month (GET /api/v2/usage/)


snopoke:
## What to build

The tracer-bullet slice of the Usage API (design: `docs/design/usage-api.md`). A single team-scoped endpoint `GET /api/v2/usage/` that answers the originating client request end-to-end: how many messages a participant has sent in the current calendar month.

Establishes the whole vertical: the `apps/api/v2/usage/` package (view, response serializer, query-param serializer, service, tests), URL registration in `apps/api/v2/urls.py`, a new `usage` OAuth scope, and a `"Usage"` tag in `SPECTACULAR_SETTINGS`.

Scope for this slice:
- `metric=messages` only, ungrouped (a single totals object).
- Time selection via `period=current_month` and `period=previous_month`.
- `tz` query param (IANA name, default `UTC`) defining calendar-month boundaries.
- Participant filtering by `participant` (public_id UUID) and `participant_identifier` (raw email/phone).
- Messages counted from `ChatMessage`, scoped `chat__team`, participant path `chat__experiment_session__participant`, returned as `{human, ai, total}`.

**HITL: this slice sets the public API contract. It needs a human contract review of the endpoint/response shape before merge.**

## Acceptance criteria

- [ ] `GET /api/v2/usage/?metric=messages&period=current_month&participant=<public_id>` returns `{human, ai, total}` counts for that participant, team-scoped
- [ ] `participant_identifier` works as an alternative to `participant`
- [ ] `period=previous_month` and `tz` shift the calendar-month boundaries correctly (message at 23:30 UTC on month-end lands in the right month under a non-UTC `tz`)
- [ ] Endpoint requires auth and the new `usage` OAuth scope; queries are scoped to `request.team`
- [ ] Response envelope includes `period` (`start`, `end`, `timezone`)
- [ ] Appears in the v2 OpenAPI schema under the `Usage` tag; `api-schemas/*.yml` regenerated
- [ ] Tests cover count correctness, team isolation, `tz` boundaries, and both participant handles

## Blocked by

None - can start immediately.

PR #3886, opened 2026-07-22
Usage API: explicit start/end window and granularity bucketing

snopoke:
Extends `GET /api/v2/usage/` (issue #3851, building on #3848). Design: `docs/design/usage-api.md`.

### Product Description
API clients can now request usage over an explicit date range and broken down into daily/weekly/monthly buckets, not just a whole calendar month.

### Technical Description
- `start`/`end` give a half-open `[start, end)` window, mutually exclusive with `period`. `period` lost its serializer-level default so `validate()` can tell "not supplied" from an explicit window; the current-month fallback now lives in `_resolve_window`.
- Naive/date-only `start`/`end` are interpreted in the request `tz` (a small `_WindowDateTimeField` defers DRF's default UTC coercion), so window and bucket boundaries honour `tz`.
- `granularity=total` keeps the single-object `results`; `daily`/`weekly`/`monthly` return one zero-filled row per bucket with `bucket_start`. Each metric runs a single tz-aware `Trunc` grouped query, mapped back onto Python-generated bucket boundaries by local date (weeks start Monday, matching Django).
- Max-window guard (`MAX_BUCKETS=366` per granularity) rejects e.g. `daily` over a multi-year range; `total` is unconstrained.

Worth a reviewer's attention — **this slice sets public API contract and needs the HITL contract review flagged on #3848**:
- `results` is polymorphic (object for `total`, array of bucket rows otherwise). The OpenAPI schema documents both shapes as a `oneOf` of object serializers, but drf-spectacular can't express "object *or* array" in one field, so the array-wrapping is described in the field text rather than the type. If the contract review wants a strict schema here, that's the place to decide.
- `bucket_start`/`period` serialize in UTC (DRF default), with `period.timezone` echoing the requested zone — consistent across the envelope.

Throttling was intentionally left out; it will be added in a later slice.

### Migrations
N/A — no schema changes.

### Docs and Changelog


apps/api/v2/usage/services.py
```diff
@@ -1,18 +1,20 @@
 """Query orchestration for the usage API (``GET /api/v2/usage/``).
 
 This is the single place that turns validated query params into team-scoped aggregates. It supports
-the ``messages``, ``sessions``, and ``participants`` metrics over a calendar month; later slices add
-cost/tokens, explicit windows, granularity, and grouping without changing this contract. See
-``docs/design/usage-api.md``.
+the ``messages``, ``sessions``, and ``participants`` metrics over an explicit ``[start, end)`` window
```
snopoke: Fixed in 0f331e561. Introduced a `UsageQuery` parameter object so the aggregation helpers take one argument instead of the repeated (team, start, end, participant, participant_identifier) cluster, bringing the module's average argument count back down.

*— Claude Code*

apps/api/v2/usage/services.py
```diff
@@ -40,43 +63,96 @@ class MessageCounts(TypedDict):
...
 
 
 def usage_query(
     team: Team,
     *,
     metrics: set[str],
-    period: str,
+    start: datetime,
+    end: datetime,
+    granularity: str = GRANULARITY_TOTAL,
```
snopoke: Fixed in 0f331e561. `usage_query` now takes a single `UsageQuery` argument.

*— Claude Code*

apps/api/v2/usage/services.py
```diff
@@ -40,43 +63,96 @@ class MessageCounts(TypedDict):
...
     if METRIC_SESSIONS in metrics:
         results[METRIC_SESSIONS] = _session_count(team, start, end, participant, participant_identifier)
     if METRIC_PARTICIPANTS in metrics:
         results[METRIC_PARTICIPANTS] = _active_participant_count(team, start, end, participant, participant_identifier)
-    return UsageResult(period_start=start, period_end=end, timezone=str(tz), results=results)
+    return results
 
 
-def _session_count(
+def _bucketed(
```
snopoke: Fixed in 0f331e561. `_bucketed` now takes a single `UsageQuery` argument.

*— Claude Code*

apps/api/v2/usage/services.py
```diff
@@ -147,6 +232,106 @@ def _message_queryset(
...
+        .values("bucket")
+        .annotate(
+            human=Count("id", filter=Q(message_type=ChatMessageType.HUMAN)),
+            ai=Count("id", filter=Q(message_type=ChatMessageType.AI)),
+        )
+    )
+    return {
+        _bucket_date(row["bucket"], tz): MessageCounts(human=row["human"], ai=row["ai"], total=row["human"] + row["ai"])
+        for row in rows
+    }
```
snopoke: Fixed in 0f331e561. `_message_counts_by_bucket` now takes `(query, trunc)`.

*— Claude Code*

apps/api/v2/usage/services.py
```diff
@@ -147,6 +232,106 @@ def _message_queryset(
...
+    participant: str | None,
+    participant_identifier: str | None,
+) -> dict:
+    rows = (
+        _session_queryset(team, start, end, participant, participant_identifier)
+        .annotate(bucket=trunc("created_at", tzinfo=tz))
+        .values("bucket")
+        .annotate(n=Count("id"))
+    )
+    return {_bucket_date(row["bucket"], tz): row["n"] for row in rows}
```
snopoke: Fixed in 0f331e561. `_session_counts_by_bucket` now takes `(query, trunc)`.

*— Claude Code*

apps/api/v2/usage/services.py
```diff
@@ -147,6 +232,106 @@ def _message_queryset(
...
+    participant: str | None,
+    participant_identifier: str | None,
+) -> dict:
+    rows = (
+        _active_participant_queryset(team, start, end, participant, participant_identifier)
+        .annotate(bucket=trunc("created_at", tzinfo=tz))
+        .values("bucket")
+        .annotate(n=Count("chat__experiment_session__participant", distinct=True))
+    )
+    return {_bucket_date(row["bucket"], tz): row["n"] for row in rows}
```
snopoke: Fixed in 0f331e561. `_participant_counts_by_bucket` now takes `(query, trunc)`.

*— Claude Code*

apps/api/v2/usage/services.py
snopoke: Fixed in 0f331e561. Extracted the shared `_scalar_by_bucket` helper; the session and participant bucket functions now delegate to it instead of duplicating the annotate/group/reduce structure.

*— Claude Code*

apps/api/v2/usage/param_serializers.py
```diff
@@ -51,4 +103,46 @@ def validate(self, attrs: dict) -> dict:
...
+        has_window = start is not None or end is not None
+        if period and has_window:
+            raise serializers.ValidationError("Provide either 'period' or an explicit 'start'/'end' window, not both.")
+        if not has_window:
+            return _month_bounds(period or PERIOD_CURRENT_MONTH, tz)
+        if start is None or end is None:
+            raise serializers.ValidationError("Both 'start' and 'end' are required for an explicit window.")
+        if end <= start:
+            raise serializers.ValidationError("'end' must be after 'start'.")
+        return start, end
```
snopoke: Fixed in 2190f0817. Split the explicit-window validation out of `_resolve_window` into `_explicit_window`, reducing the method's branching.

*— Claude Code*

apps/api/v2/usage/serializers.py
```diff
@@ -29,14 +31,40 @@ class UsageResultsSerializer(serializers.Serializer):
...
+        PolymorphicProxySerializer(
+            component_name="UsageResultsOrBuckets",
+            serializers=[UsageResultsSerializer, UsageBucketResultSerializer],
+            resource_type_field_name=None,
+        )
+    )
+    def get_results(self, obj):
+        if obj.granularity == GRANULARITY_TOTAL:
+            return UsageResultsSerializer(obj.results).data
+        return UsageBucketResultSerializer(obj.results, many=True).data
```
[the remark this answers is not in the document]
snopoke: Fixed in 719231e05. Put the `PolymorphicProxySerializer` in manual mode (`many=False` on the proxy) so the `UsageBucketResult` sub-serializer's `many=True` is honored — thanks for the issue #692 pointer, that was the missing piece. `UsageResultsOrBuckets` is now a `oneOf` of the `UsageResults` object and an `array` of `UsageBucketResult`, matching what `get_results` returns. Schema regenerated.

*— Claude Code*

PR #3894, opened 2026-07-23
Make Node rows the sole source of pipeline node content

### LINKED ISSUE #1106: Store related object FK's on the node itself instead of node params

SmittieC:
### Product Description
No change — internal storage refactor.

### Technical Description
`Pipeline.data` duplicated every node's content (type/label/params) inside `data.nodes[*].data`, and the copies drifted: version publishing rewrites `Node.params` but not the blob, `set_params` writes the row only. `flow_data` already compensated by overwriting blob content with row values on every read. This makes the `Node` rows the single source of truth: `Pipeline.data` keeps layout only (node `id`, react-flow `type`, `position`, edges, viewport), and saves are inverted — persist the stripped layout, then reconcile rows via `update_nodes_from_data(node_data)` with an explicit content mapping. The wire format is unchanged in both directions.

The decision is recorded in ADR-0046; `docs/design/pipeline-data-layout-only.md` carries the contract, the per-caller table, and the migration rules — reviewing those two files first will make the diff read faster.

Worth a closer look:
- `apply_pipeline_patch` builds the content mapping strictly from the diff's add/update nodes, so stale blobs still embedded in pre-migration rows are never pushed back into the rows (`test_patch_does_not_rewrite_rows_from_stale_stored_blob` covers the exact drift scenario this refactor exists to kill).
- `FlowNode.data` became optional at the wire level; a payload node without content and without an existing row is rejected as a 400 by `update_nodes_from_data`'s raise, not by pydantic.
- `duplicate_pipeline_with_new_ids` now takes a `{flow_id: node_type}` mapping from the rows to keep copied ids human-readable (`LLMResponseWithPrompt-a1b2c`), since the blob it used to read is gone.
- `teams/export/importer` no longer remaps resource ids inside `pipeline.data` (`remap_pipeline_data` is deleted): node content — including the remapped params — is imported via the `Node` rows, and the pipeline blob imports as-is. An old export file therefore lands with its stale embedded blob still in `data`; no read path consumes it, and the next save or a `strip_node_data` run removes it.

Closes #1106

### Migrations

No Django migration ships in this PR (an earlier revision had a data migration, 0029; it was replaced). Nothing rewrites existing rows at deploy time — the new code reads both the old format (blob embedded) and the new layout-only format, so unmigrated rows are safe indefinitely.

Rollout:
- Deploy the new code; no migration ordering to coordinate.
- Blobs are stripped lazily: every save path (POST/PATCH, revert, import) persists layout-only data, so pipelines converge to the new format as they are edited.
- To strip the remainder eagerly, run the `strip_node_data` management command any time after deploy. It is idempotent and batch-based, and it skips (and logs) any pipeline whose blob has no backing `Node` row rather than destroy the only copy of its content. A data migration reusing the same helpers is deferred to a follow-up PR.
- Rollback: pre-ADR-0046 code requires the blob (`FlowNode.data` was a mandatory pydantic field), so rolling the code back after any pipelines have been saved in the new format requires `strip_node_data --rebuild` first, which rebuilds each node's blob from its `Node` row.

### Docs and Changelog

ADR + design doc are included in this PR; no user-facing changelog entry needed.


docs/adr/0046-layout-only-pipeline-data.md
```diff
@@ -0,0 +1,62 @@
...
+Saves are inverted: the layout-only graph is persisted first, then
+`update_nodes_from_data(node_data)` reconciles the rows, taking content from an explicit
+`{flow_id: {type, label, params}}` mapping instead of reading it out of `self.data`.
+Only paths where content genuinely originates outside the database supply the mapping —
+UI saves, pipeline creation, import files. Paths that already have rows reconstruct from
+them: revert builds the mapping from the version's rows; publishing versions and copies
+never calls `update_nodes_from_data` at all (rows are copied directly). A node id present
+in the graph but absent from both the mapping and the rows is an error.
+
+An idempotent data migration strips the blob from existing rows, skipping (and logging)
```
SmittieC: remove this part. It's not really useful in here

apps/pipelines/views.py
```diff
@@ -410,10 +410,15 @@
...
+        pipeline.data, node_data = split_flow_data(data.data.model_dump())
         pipeline.edit_revision += 1
         pipeline.save(update_fields=["name", "data", "edit_revision"])
-        pipeline.update_nodes_from_data()
+        try:
+            pipeline.update_nodes_from_data(node_data)
+        except ValueError as e:
+            # A payload node without content for which no row exists — client error
+            transaction.set_rollback(True)
+            return JsonResponse({"error": str(e)}, status=400)
```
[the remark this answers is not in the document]
SmittieC: Fixed in e13c3062a. The view now catches a dedicated `MissingNodeDataError` (instead of bare `ValueError`) and builds the 400 message from the exception's structured `node_ids` attribute — client-supplied flow ids only — rather than echoing the exception message, so no internal exception detail can flow into the response.

apps/pipelines/views.py
```diff
@@ -450,11 +455,15 @@
...
+        pipeline.data, node_data = apply_pipeline_patch(pipeline.data, patch)
         pipeline.edit_revision += 1
         pipeline.save(update_fields=["name", "data", "edit_revision"])
-        pipeline.update_nodes_from_data()
+        try:
+            pipeline.update_nodes_from_data(node_data)
+        except ValueError as e:
+            # A diff node without content for which no row exists — client error
+            transaction.set_rollback(True)
+            return JsonResponse({"error": str(e)}, status=400)
```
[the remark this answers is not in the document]
SmittieC: Fixed in e13c3062a. Same change as the POST handler: a dedicated `MissingNodeDataError` is caught and the 400 message is built from the exception's structured `node_ids` attribute (client-supplied flow ids only), not from the exception message.

apps/pipelines/migrations/0029_strip_node_data_from_pipeline_data.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("pipelines", "0028_pipeline_edit_revision"),
+    ]
+
+    operations = [
+        migrations.RunPython(_strip_node_data, _rebuild_node_data, elidable=True),
+    ]
```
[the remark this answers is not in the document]
SmittieC: Fixed in e13c3062a — added `atomic = False` with a comment explaining why. The strip/rebuild utils are idempotent and rerunnable by design, so incremental commits are safe and avoid holding row locks on the whole Pipeline table.

apps/pipelines/patching.py
```diff
@@ -29,7 +32,13 @@ def apply_pipeline_patch(current_data: dict, patch: PipelineDiffPayload) -> dict
     for key in current_data:
         if key not in merged:
             merged[key] = current_data[key]
-    return merged
+    layout_data, _ = split_flow_data(merged)
+    node_data = {
+        node.id: {"type": node.data.type, "label": node.data.label, "params": node.data.params}
+        for node in [*patch.nodes.add, *patch.nodes.update]
+        if node.data
+    }
```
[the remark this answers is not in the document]
SmittieC: Fixed in e13c3062a. `apply_pipeline_patch` now emits content only for update nodes and adds whose id was not already in the graph — with the exception that delete+re-add of the same id in one patch is a genuine replacement and still emits content. Added regression tests for both cases (`test_duplicate_add_emits_no_node_content`, `test_delete_then_add_same_id_emits_node_content`). Worth noting the practical exposure was narrow: a true retry after a successful PATCH is rejected by the `base_revision` check with a 409 before reaching this code, so this mainly guards against a buggy client adding an existing id within a current revision.

docs/adr/0046-layout-only-pipeline-data.md
SmittieC: Fixed in 044e7f08f — removed the migration/backward-compat paragraph from the ADR. The design doc companion keeps those implementation details.

apps/pipelines/models.py
```diff
@@ -126,12 +126,19 @@ def create_default(cls, team, name=None, llm_provider_id=None, llm_provider_mode
...
-    def update_nodes_from_data(self) -> None:
-        """Set the nodes on the pipeline from data coming from the frontend"""
-        nodes = [FlowNode(**node) for node in self.data["nodes"]]
-        # Delete old nodes
+    def update_nodes_from_data(self, node_data: dict[str, dict]) -> None:
+        """Reconcile this pipeline's ``Node`` rows with the node list in ``self.data``.
+
+        ``self.data`` (already persisted) decides graph membership only: rows whose
+        flow_id disappeared from it are deleted, or archived when they have versions.
+        Node content never comes from ``self.data`` — the rows own it (ADR-0046).
```
SmittieC, line 134: this should be "Node content never comes from ``node_data``"?

apps/teams/export/importer.py
```diff
@@ -121,14 +122,16 @@ def remap_node_params(params: dict, store: FKTranslationStore) -> dict:
 
 
 def remap_pipeline_data(data: dict | None, store: FKTranslationStore) -> dict | None:
```
SmittieC: This shouldn't be necessary if the soiurce pipeline.data is correct in the first place. I think we should trust that the source pipeline.data is correct

apps/pipelines/migrations/0029_strip_node_data_from_pipeline_data.py
```diff
@@ -0,0 +1,30 @@
...
+
+def _rebuild_node_data(apps, schema_editor):
+    # Pre-ADR-0046 code requires the embedded blob, so reversing rebuilds it from the
+    # Node rows (which own the content and were untouched by the forward migration).
+    Pipeline = apps.get_model("pipelines", "Pipeline")
+    Node = apps.get_model("pipelines", "Node")
+    rebuild_node_data_in_pipelines(Pipeline, Node)
+
+
+class Migration(migrations.Migration):
```
SmittieC: this should be an idempotent command, not a migration. the migration will follow in a separate PR

apps/pipelines/models.py
SmittieC: No code change — the suggested wording would invert the meaning. Node content *does* come from `node_data` (the next sentence: rows are created or updated from the mapping). This sentence contrasts with `self.data`, which is layout-only per ADR-0046 and only decides graph membership, so `self.data` is correct here.

apps/teams/export/importer.py
SmittieC: Fixed in 49d673eab. Removed `remap_pipeline_data` entirely — imported pipeline data is trusted to be layout-only, so there is nothing to strip or remap. Node params are still remapped via `remap_node_params`.

apps/pipelines/migrations/0029_strip_node_data_from_pipeline_data.py
SmittieC: Fixed in fae1d7a21. Migration 0029 removed and replaced with an idempotent `strip_node_data` management command (`--rebuild` runs the reverse). The strip/rebuild helpers stay in `migrations/utils` for the follow-up migration PR to reuse.

snopoke (commented):
I haven't reviewed this closely but I'm interested to see that we're still storing the node layout data on the pipeline model. Is there a reason we couldn't store that at the node to avoid having to merge it back in?

SmittieC:
> I haven't reviewed this closely but I'm interested to see that we're still storing the node layout data on the pipeline model. Is there a reason we couldn't store that at the node to avoid having to merge it back in?

I suppose we can. My reasoning for not doing it - which I realize is not clear - was that the pipeline knows where nodes are / should be and not the nodes themselves. I guess it doesn't really matter. I'll make an update :+1:

SmittieC:
> > I haven't reviewed this closely but I'm interested to see that we're still storing the node layout data on the pipeline model. Is there a reason we couldn't store that at the node to avoid having to merge it back in?
> 
> I suppose we can. My reasoning for not doing it - which I realize is not clear - was that the pipeline knows where nodes are / should be and not the nodes themselves. I guess it doesn't really matter. I'll make an update 👍

See https://github.com/dimagi/open-chat-studio/pull/3894/commits/f9101418ec7d00f550a1715524e102c5dedcb07e. For this PR, I'm going to only mirror the position in `pipeline.data` to the `Node` records, and in a next PR - after running the data migration on prod - I'll make some logic changes that can safely assume that all node data can be found on the `Node` records.

FYI: This branch is safe to deploy without having to run the data migration. As people edit pipelines, it will start to converge to the "new form". The data migration is just to tackle the rest.

apps/pipelines/migrations/utils/strip_node_data.py
```diff
@@ -32,6 +35,37 @@ def strip_node_data_from_pipelines(Pipeline, Node):
...
+    The blob is authoritative, so a differing row value is overwritten. Blob nodes
+    without a usable ``{"x", "y"}`` position or without a backing row are skipped.
+    """
+    nodes = (pipeline.data or {}).get("nodes") or []
+    if not nodes:
+        return
+    rows = {}
+    # "-is_archived" puts archived rows first so a non-archived row wins on flow_id collision
+    for row in Node._base_manager.filter(pipeline_id=pipeline.id).order_by("-is_archived"):
+        rows[row.flow_id] = row
```
snopoke: This could be done in outside the function and used in both `_backfill_positions` and `_strip_nodes` to avoid a 2nd query.

snopoke (dismissed):
One suggestion but otherwise LGTM

apps/pipelines/migrations/utils/strip_node_data.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/3894/commits/3ffdff22319eaf2fb69eac4c59200941a58d2166

PR #3908, opened 2026-07-23
feat: pipeline build-state reporting + additive inspect fields (write API phase 1, PRs 1–2)

SmittieC:
This is **PRs 1 and 2** of the sequence listed in the [write API plan](https://docs.google.com/document/d/1oD25fmxCjdam728UOinAj7bry15ih1pxdhf5paokalw/edit?tab=t.xzd2t2em41gc#heading=h.rvrvnj6iypo8).

### Product Description

`GET /api/v2/chatbots/{id}/inspect/` now reports whether a chatbot's pipeline is usable, plus the identifiers a caller needs to reference things by id:

- top level: `pipeline_valid`, `pipeline_errors` (three buckets — `node` / `edge` / `pipeline`), and `unwired_handles` (an advisory "what still needs wiring" map)
- `output_handles` per node, `id` per edge, `model_id` on a node's `llm` and on an indexed collection's `embedding`, `synthetic_voice_id` on `voice`

Additive except one narrowing: `pipeline` loses `nullable: true`, because the endpoint now requires a pipeline to be there — see known characteristic 2. The pipeline editor also now shows **all** graph-level errors at once instead of only the first.

### Technical Description

**No write endpoints yet** — this is the read side plus the reporting core both sides will share.

- **`Pipeline.validate()` always returns a fully-populated `ErrorReport` now**, so an error-free report is a *truthy* dict of three empty values and callers must use `has_errors()`. `errors["pipeline"]` also changed from a string to a list (editor front-end updated to match).
- **`PipelineGraph.build_errors` collects structural problems instead of raising on the first.** Six cases that previously crashed now land in the report: both missing terminals together, an edge endpoint with no node row, a conditional edge stranded on a handle its source doesn't offer, a named handle on a non-router source, an unknown/removed node type, and a `PipelineNodeBuildError` escaping a validator. Not API-only — the editor, the publish gate and pipeline execution all previously 500'd or crashed on this data. At runtime the failure mode moves from a bare `KeyError` to `PipelineBuildError`, which `apps/channels/pipeline.py:216` already catches.
- **Node validation no longer crashes on an unlabelled model-level error.** For a pydantic error with no `loc`, `_node_validation_errors` read `error["ctx"]["field"]` — a key only the codebase's own `PydanticCustomError` raises supply, so a validator raising a plain `ValueError` or failing an `assert` would `KeyError` straight out of `validate()`. It now falls back to the node's `"root"` bucket, which the editor already renders (`PipelineNode.tsx:33`). Pre-existing on `main`, but squarely in this PR's "report it, don't crash" remit.
- **New `apps/pipelines/build_state.py`** folds the above into `{pipeline_valid, errors, unwired_handles}`.
- **`errors` → `pipeline_errors`.** At a REST response root, `errors` conventionally means "what went wrong with your request" — v1 uses it exactly that way (`apps/api/views/participants.py:257`) — whereas this describes the subject. Renamed before phase 2 builds against it.

### Known characteristics

1. **The render runs `validate()` on every GET**, so queries grow per LLM-bearing node: over `pipeline_build_state`, 3 + 4n (6n for LLM-backed routers). Each pass builds a fresh `ORMRepository`, so `@instance_cache` never engages. `EXPECTED_RENDER_QUERIES` rose 13 → 17 and guards one fixed bot — **no slope guard yet.**
2. **The API requires a pipeline and fails hard without one — deliberately.** Every v2 chatbot endpoint assumes `experiment.pipeline` is not null and that its `pipeline.data` is populated. There is no null-pipeline branch and no empty-graph fallback: `validate()` subscripts `data["edges"]`, so both assumptions are load-bearing. Nothing enforces them at the model level — `Experiment.pipeline` is still `null=True, on_delete=SET_NULL` and `ChatbotViewSet.get_queryset()` doesn't filter on it — so a pipeline-less Experiment (deleting a `Pipeline` row silently produces one) is listed by `GET /api/v2/chatbots/` and then **500s when inspected**, where `main` returned `pipeline: null`. **That failure is accepted.** A chatbot with no pipeline is a corrupt row rather than a state the product supports, and a loud 500 is the wanted outcome — better than a `pipeline__isnull=False` filter that hides the row behind a 404, or a null-safe response that reports a pipeline-less chatbot as something you can inspect. This is also why the schema drops `nullable: true` on `pipeline`. (The `pipeline.data or {}` default inside `unwired_handles` covers a graph with no edges stored yet, not null `data`.)
3. **`unwired_handles` judges wiredness from the stored edges only.** An edge pointing at a handle its source no longer offers still marks that pair wired; the stranded edge is `pipeline_errors.edge`'s concern.

### Tests

`uv run pytest apps/pipelines apps/api/v2` — 723 passed. New: `test_build_state.py` (24 tests, one per case above) and `test_tasks.py`, covering the `has_errors` guard on the editor's test-message task — the one `validate()` caller with no coverage, which the truthiness change had silently broken for *every* pipeline.

### Demo

No screencast; the surface is a JSON response, its shape pinned by the golden-master `test_full_response_body` and the regenerated `api-schemas/v2.yml`.

### Docs and Changelog

Schema regenerated in-repo, so the reference docs follow. Reviewer notes: `pipeline` loses `nullable: true` and a null pipeline is now a hard failure; the field is `pipeline_errors`, not `errors`; and `Pipeline.validate()`'s return contract changed for **every** caller, so anything outside this diff reading it needs checking.


apps/pipelines/build_state.py
```diff
@@ -0,0 +1,111 @@
...
+    for node in pipeline.node_set.all():
+        dangling = []
+        if node.type != StartNode.__name__ and node.flow_id not in wired_inputs:
+            dangling.append({"handle": STANDARD_INPUT_NAME, "label": None})
+        for handle in node_output_handles(node):
+            if (node.flow_id, handle["handle"]) not in wired_outputs:
+                dangling.append(handle)
+        if dangling:
+            unwired[node.flow_id] = dangling
+    return unwired
```
SmittieC: Fixed in 1cff04aae: extracted the per-node scan into a `_dangling_handles` helper, bringing `unwired_handles` below the complexity threshold.

apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -584,6 +660,7 @@ def test_full_response_body():
             "identifier_type": "email",
         },
         "voice": {
+            "id": bot.synthetic_voice.id,
```
SmittieC: This is a deliberate golden-master test: it asserts the complete inspect response body as a single literal so any contract change shows up as a diff. The growth is the new `pipeline_valid`/`errors`/`unwired_handles` fields this PR adds; splitting the assertion would obscure the full contract (no code change needed).

apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -459,14 +532,17 @@ def _full_bot():
                         "label": "Answer",
                         # the react FE persists ids as strings
                         "params": {
+                            "name": "Answer",
```
SmittieC, line 531: `_full_bot` is the fixture builder for the fully-populated bot behind the golden-master test; the delta here is one line. Splitting it to duck the threshold would spread one coherent setup across helpers and hurt readability (no code change needed).

apps/pipelines/models.py
```diff
@@ -165,13 +165,21 @@ def validate(self, full=True) -> dict:
         errors = defaultdict(dict)
         nodes = self.node_set.all()
         for node in nodes:
-            node_class = getattr(pipeline_nodes, node.type)
+            node_class = getattr(pipeline_nodes, node.type, None)
+            if node_class is None:
+                # A node whose class has since been removed must be reported, not crash validation.
+                errors[node.flow_id]["root"] = f"Unknown node type: {node.type}"
+                continue
```
SmittieC: Fixed in 1cff04aae: per-node validation moved to `Pipeline._node_validation_errors`, bringing `validate` back under the complexity threshold.

apps/api/v2/tests/test_inspect_schema.py
```diff
@@ -98,6 +101,12 @@ def test_node_params_fields_carry_descriptions(api_schema):
     assert described.get("history_type")
 
 
+def test_graph_edge_declares_id(api_schema):
+    """Each graph edge carries its ``id`` — the future edge-delete address for the write API."""
```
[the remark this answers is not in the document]
SmittieC: Fixed in 779296a3d: the docstring now states only the current contract ("Each graph edge exposes its `id`.").

apps/pipelines/graph.py
```diff
@@ -94,12 +94,23 @@ def end_node(self) -> Node:
...
+                continue
             # this creates a map of the form:
             # {source_node: {'source_handle_1': value_to_follow_edge_1, 'source_handle_2': value_to_follow_edge_2}}
             conditional_edge_map[edge.source][output_map[edge.sourceHandle]] = edge.target
+        if stranded_edges:
+            raise PipelineBuildError(
+                "One or more edges reference a router output that no longer exists",
+                edge_ids=[edge.id for edge in stranded_edges],
+            )
         return conditional_edge_map
```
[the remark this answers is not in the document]
SmittieC: Fixed in c31af7773: stranded-edge detection is now scoped to edges whose source is reachable from Start, matching the build's reachable-node contract. Added a regression test (`test_stranded_edge_on_unreachable_router_does_not_block_the_build`) covering a reachable router alongside an unreachable island with a stranded edge — it failed before the fix and passes after.

apps/api/v2/inspect/serializers.py
```diff
@@ -262,6 +263,12 @@ class FlattenedModelProviderSerializer(FlattenedProviderSerializer):
     Expects a :class:`ProviderModelPair`; a missing provider or model renders its fields as null.
     """
 
+    id = serializers.IntegerField(
```
SmittieC: rename to "model_id"

apps/api/v2/inspect/serializers.py
```diff
@@ -290,6 +297,12 @@ class FlattenedVoiceSerializer(FlattenedProviderSerializer):
     Expects a :class:`VoicePair`.
     """
 
+    id = serializers.IntegerField(
```
SmittieC: rename to synthetic_voice_id

api-schemas/v2.yml
```diff
@@ -723,20 +723,44 @@ components:
           - $ref: '#/components/schemas/InspectPipeline'
           readOnly: true
           nullable: true
+        pipeline_valid:
+          type: boolean
+          nullable: true
+          readOnly: true
+          description: Whether this chatbot's pipeline validates cleanly (all three
+            ``errors`` buckets empty); null for a chatbot with no pipeline. Unwired
```
SmittieC: all chatbots will have pipelines

apps/api/v2/inspect/serializers.py
SmittieC: Fixed in eef42ee93. `FlattenedModelProviderSerializer.id` is now `model_id` (inherited by `FlattenedLlmSerializer`, so both the node `llm` object and the collection `embedding` object rename), with the tests and `api-schemas/v2.yml` updated to match.

apps/api/v2/inspect/serializers.py
SmittieC: Fixed in eef42ee93. `FlattenedVoiceSerializer.id` is now `synthetic_voice_id`. Its help text no longer needs to spell out the write-back name, so it reads "The synthetic voice's id — the value to write back to set the voice."

api-schemas/v2.yml
SmittieC: Fixed in eef42ee93 — and made it true at the query level, since `Experiment.pipeline` is still `null=True, on_delete=SET_NULL` in the DB.

- `pipeline` and `pipeline_valid` are no longer nullable, and `_build_state` reads `pipeline_build_state(experiment.pipeline)` directly instead of branching on `pipeline is None`.
- `ChatbotViewSet.get_queryset()` and `_inspect_target_queryset()` now filter `pipeline__isnull=False`, matching `apps/chatbots/views.py:200` — having a pipeline is what makes an Experiment a chatbot. A pipeline-less legacy experiment therefore 404s rather than 500ing on the now-undefended `pipeline.validate()`.
- The old `pipeline_valid is None` test is replaced by `test_experiment_without_pipeline_is_not_a_chatbot`, which asserts the 404.

apps/api/v2/inspect/versioning.py
```diff
@@ -19,18 +19,24 @@ class InspectVersionError(ValueError):
...
-        "synthetic_voice__voice_provider",
-        "pipeline",
-    ).prefetch_related(
-        Prefetch("pipeline__node_set", queryset=inspect_node_queryset()),
-        Prefetch("static_triggers", queryset=StaticTrigger.objects.select_related("action")),
-        Prefetch("timeout_triggers", queryset=TimeoutTrigger.objects.select_related("action")),
+    # Only pipeline-backed experiments are chatbots (matching the chatbots UI list), so a
+    # pipeline-less legacy experiment resolves to nothing and the view returns a 404.
+    return (
+        Experiment.objects.filter(pipeline__isnull=False)
```
SmittieC: assume that all experiments have pipelines

apps/api/v2/tests/test_inspect_schema.py
```diff
@@ -98,6 +101,12 @@ def test_node_params_fields_carry_descriptions(api_schema):
     assert described.get("history_type")
 
 
+def test_graph_edge_declares_id(api_schema):
```
SmittieC: low value test

apps/api/v2/inspect/versioning.py
SmittieC: Fixed in 3d79fe46d: dropped `pipeline__isnull=False` from `_inspect_target_queryset()` and the matching filter on `ChatbotViewSet.get_queryset()` (both added by eef42ee93 for the same defensive reason).

`pipeline`/`pipeline_valid` stay non-nullable and `_build_state` still reads `experiment.pipeline` undefended — the assumption now lives in the code rather than being enforced at the query level. Also removed `test_experiment_without_pipeline_is_not_a_chatbot` and the `_build_state_bot` kwargs passthrough it needed.

One thing to flag: `Experiment.pipeline` is still `null=True, on_delete=SET_NULL` in the DB, so a pipeline-less legacy row now 500s on `pipeline_build_state(None)` instead of 404ing.

apps/api/v2/tests/test_inspect_schema.py
SmittieC: Fixed in 4348f08f0: removed `test_graph_edge_declares_id`. The golden-master assertion in `test_chatbots_inspect.py` already covers the edge `id`.

apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -591,6 +666,7 @@ def test_full_response_body():
             "identifier_type": "email",
         },
         "voice": {
+            "synthetic_voice_id": bot.synthetic_voice.id,
```
SmittieC, line 799: Fixed in b947747af. `test_full_response_body` is now five lines: the expected payload moved into `_expected_full_response`, with the two largest nested chunks — the pipeline's node objects and the events block — in `_expected_pipeline_nodes` and `_expected_events`. Longest of the three is 104 lines, well under the 150 threshold.

The golden-master property is intact — the test still asserts the entire payload in one equality, so a contract change still surfaces as a single diff. That was the reason for declining the earlier instance of this biomarker; splitting the *expectation* rather than the *assertion* keeps that and satisfies the threshold.

*— Claude Code*

apps/api/v2/inspect/serializers.py
```diff
@@ -749,6 +811,36 @@ def get_voice(self, experiment):
...
         return ChannelSerializer(get_channels(experiment), many=True).data
 
+    def _build_state(self, experiment) -> dict:
+        # Computed once per instance (validate() builds the whole graph) and shared by the three
+        # build-state fields below. Keyed by pk because DRF reuses one serializer instance across
+        # items under many=True.
+        cached = getattr(self, "_build_state_cache", None)
+        if cached is None or cached[0] != experiment.pk:
+            self._build_state_cache = (experiment.pk, pipeline_build_state(experiment.pipeline))
+        return self._build_state_cache[1]
```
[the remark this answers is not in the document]
SmittieC: This is fine, since this is not expected and should fail hard when detected

apps/pipelines/graph.py
```diff
@@ -119,23 +193,59 @@ def build_from_pipeline(cls, pipeline: Pipeline) -> Self:
         edge_data = [Edge(**edge) for edge in pipeline.data["edges"]]
         return cls(nodes=node_data, edges=edge_data)
```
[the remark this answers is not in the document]
SmittieC: This is fine. Failing hard is fine, since this is not a normal case

apps/pipelines/models.py
```diff
@@ -223,15 +229,43 @@ def validate(self, full=True) -> dict:
...
+        from apps.pipelines.nodes import nodes as pipeline_nodes  # noqa: PLC0415 - circular: nodes.nodes→models
 
+        node_class = getattr(pipeline_nodes, node.type, None)
+        if node_class is None:
+            # A node whose class has since been removed must be reported, not crash validation.
+            return {"root": f"Unknown node type: {node.type}"}
+        try:
+            node_class.model_validate({**node.params, "node_id": node.flow_id, "django_node": node})
+        except pydantic.ValidationError as e:
+            return {(error["loc"][0] if error["loc"] else error["ctx"]["field"]): error["msg"] for error in e.errors()}
```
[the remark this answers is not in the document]
SmittieC: Real, and verified — though not reachable with today's node classes.

`ctx["field"]` only exists because *we* put it there: every empty-`loc` error the pipeline nodes raise is a `PydanticCustomError` passing `{"field": ...}` (`nodes.py:358`, `mixins.py:102`, …). Any other empty-`loc` error has a different `ctx`:

```
ValueError from a model_validator  -> {'type': 'value_error',     'loc': (), 'ctx': {'error': ValueError('boom')}}
failed assert in a model_validator -> {'type': 'assertion_error', 'loc': (), 'ctx': {'error': AssertionError(...)}}
```

Both `KeyError` on `ctx["field"]`, and since `Pipeline.validate()` is what the new inspect fields call, that lands as a 500 instead of an error report — the exact failure mode this PR exists to remove.

No node validator raises a bare `ValueError` today, so it is latent rather than live. Fixed regardless: the lookup falls back to the node's `"root"` bucket, which the docstring already promised and which the editor already renders (`PipelineNode.tsx:33`).

```python
return {
    (error["loc"][0] if error["loc"] else error.get("ctx", {}).get("field", "root")): error["msg"]
    for error in e.errors()
}
```

Covered by `TestNodeValidationErrors::test_model_level_error_naming_no_field_lands_on_the_node`, which registers a node class whose model validator raises a plain `ValueError`; against the old expression it fails with `KeyError: 'field'`.

SmittieC:
Superseded by https://github.com/dimagi/open-chat-studio/pull/4053 which has a cleaner commit history

PR #3912, opened 2026-07-24
feat: deploy-safe evaluation runs

SmittieC:
### Product Description

Evaluation runs no longer brick when a deploy kills a worker mid-run. Previously a deploy (SIGTERM → SIGKILL) could silently drop a chunk's remaining messages: the chord callback never fired, the run stayed `PROCESSING` forever, and that in turn blocked dataset/evaluator/config edits via `InFlightRunsError`, with no recovery path. Now a deploy delays a run by minutes instead of stranding it — no manual fixes, no duplicated results, and queue depth + LLM load stay bounded (≈ the old cap of 10).

### Technical Description

Replaces the Celery-chord dispatch (`run_evaluation_task` → `.chunks()` → `chord` → `mark_evaluation_complete`) with a stateless beat coordinator. Completion is derived from the database each tick rather than from Celery chord state (which lives in Redis and is unrecoverable when lost).

The design decision is captured in [ADR-0046](docs/adr/0046-deploy-safe-evaluation-runs.md) (extends ADR-0020).

**How it works**
- **Frozen plan at creation** — `EvaluationConfig.run()` snapshots the message plan into `scoped_messages` (now for *all* run types: FULL = all dataset ids, PREVIEW = the sample, DELTA = the explicit list) and the evaluator ids into a new `evaluator_ids` field, and assigns a uuid `job_id` used as the progress key. The coordinator only ever reads this frozen plan, never the live dataset (which `auto_populate_eval_datasets` grows every 5 min).
- **Coordinator** — `coordinate_evaluation_runs` (beat, every 30s) drives each active run under `select_for_update(skip_locked=True)` so overlapping sweeps partition runs. Per tick it recomputes remaining work from `EvaluationResult`, then either dispatches the next batch (≤ `BATCHES_PER_TICK`×`BATCH_SIZE` = 30 messages, as batch tasks of 3), re-dispatches only the unfinished messages of a stalled dispatch, or completes the run (aggregates + stale-tag reversal). Coordination state (`in_flight`, `batch_dispatched_at`, `stall_count`) is saved *before* any `.delay()`, and dispatch happens *after* the transaction commits — so a crash mid-dispatch under-dispatches (repaired by the stall branch) rather than dispatching work the saved state doesn't reflect.
- **Dumb batch task** — `evaluate_message_batch` (`acks_late=True`, `soft_time_limit=240`) evaluates a small batch in-process and exits. No chords, callbacks, or self-rescheduling; any lost piece is repaired by the next tick. `acks_late` lets Redis redeliver a SIGKILLed batch after the visibility timeout.
- **Idempotent message evaluation** — `evaluate_single_message_task` drops evaluators that already have a result for `(run, message)` and returns before bot generation if none remain, so broker redelivery and stall re-dispatch are cheap no-ops for completed work.
- **Duplicate-proof** — a unique constraint `unique_result_per_run_message_evaluator` on `(run, message, evaluator)`, with `IntegrityError` handling at both create sites, makes duplicate result rows impossible even when a redelivery races a stall re-dispatch.
- **Fast-path** — `run_evaluation_task` shrinks to a single immediate coordination tick so users don't wait for the first beat; the sweep's PENDING branch is its backstop.
- **Failure handling** — 3 consecutive no-progress stalls flip a run to `FAILED` with an error message, which clears the `InFlightRunsError` edit-blockage without DB surgery.
- **Progress UI** — the coordinator publishes the done/total it already computes to the Celery result backend under `job_id` (celery_progress meta shape); the run page and runs-list status column poll `celery_progress:task_status`. Each poll is an O(1) Redis GET regardless of viewer count.
- **Observability** — one Taskbadger task per run (created *after* the coordination tick commits, never under the row lock, and updated each tick), with `coordinate_evaluation_runs` and `evaluate_message_batch` excluded from `CelerySystemIntegration` auto-tracking so a run is one TB task, not dozens.

**Tuning:** `BATCH_SIZE=3`, `BATCHES_PER_TICK=10`, `STALL_TIMEOUT=12min`, `MAX_STALLS=3`, `soft_time_limit=240s`, beat tick `30s`.

**Tests:** full `apps/evaluations/` suite passes (343). New `test_evaluation_coordination.py` covers plan freezing, message idempotency, the batch task, every sweep branch (first-batch dispatch, dispatch cap, batch-done → next batch, DB-derived completion, stall re-dispatch, 3-stall → FAILED, fresh-batch no-op, `_is_stalled` floor, multi-evaluator completion counting), the progress/Taskbadger publish path, and an end-to-end multi-batch convergence run that asserts exact result count and zero duplicates.

### Migrations

`0016_evaluationrun_batch_dispatched_at_and_more` is **schema-only**: adds `EvaluationRun.{evaluator_ids, in_flight, batch_dispatched_at, stall_count, taskbadger_task_id}` (all with safe defaults) and the `EvaluationResult` unique constraint. No data migration. `evaluator_ids` and `in_flight` are `ArrayField(IntegerField())`; the rest use scalar defaults, so every added column is a constant-default `ADD COLUMN` (metadata-only on PostgreSQL 11+, no table rewrite).

**⚠️ Pre-deploy gates (must be handled by whoever deploys):**

1. **Duplicate check before applying the constraint.** The new unique constraint on `(run, message, evaluator)` will *fail to apply* if production already contains duplicate rows (plausible, since duplicate-drop is one of the bugs this PR fixes). Run this first and clean up (keeps the newest row per group):
   ```sql
   -- 1. Inspect
   SELECT run_id, message_id, evaluator_id, COUNT(*)
   FROM evaluations_evaluationresult
   GROUP BY run_id, message_id, evaluator_id
   HAVING COUNT(*) > 1;

   -- 2. Delete all but the highest id per group (run inside a transaction, verify counts first)
   DELETE FROM evaluations_evaluationresult a
   USING evaluations_evaluationresult b
   WHERE a.run_id = b.run_id
     AND a.message_id = b.message_id
     AND a.evaluator_id = b.evaluator_id
     AND a.id < b.id;
   ```

2. **In-flight runs at cutover.** Runs that are `PENDING`/`PROCESSING` at deploy time were created by the old code without `evaluator_ids`, so the first post-deploy beat tick marks them `COMPLETED` with whatever partial results exist (and computes aggregates over that partial set). This is intentional — it clears the old "stuck forever" runs — but it is a one-time behavior. Also expect transient worker-log noise: any `mark_evaluation_complete` chord callback still queued from before the deploy will error as an unregistered task. Both are harmless; flag on-call so it isn't mistaken for a regression.

### Demo

Backend reliability change (no UI surface beyond the existing run-progress bar, which now advances in ~30s steps and is fed from Redis). To exercise: trigger an evaluation run and watch it progress and complete on the run page; kill a worker mid-run and confirm the run recovers on a later tick instead of stranding in `PROCESSING`.

### Docs and Changelog

Internal execution/reliability change; no user-facing docs update needed. Design recorded in ADR-0046.


apps/evaluations/models.py
```diff
@@ -541,8 +559,17 @@ class EvaluationRun(BaseTeamModel):
...
-        help_text="Subset of dataset messages this run evaluated. Empty for FULL/PREVIEW.",
+        help_text=(
+            "The frozen message plan for this run (all dataset ids for FULL, the sample for PREVIEW, "
+            "the explicit list for DELTA)."
+        ),
     )
+    # Coordination state, written only by the beat coordinator under a row lock.
+    evaluator_ids = SanitizedJSONField(default=list)  # evaluator ids frozen at creation
+    in_flight = SanitizedJSONField(default=list)  # message ids of the current wave
+    wave_dispatched_at = models.DateTimeField(null=True, blank=True)
```
SmittieC: rename to `batch_dispatched_at`. Use "batch" instead of "wave" in all places

apps/evaluations/models.py
```diff
@@ -541,8 +559,17 @@ class EvaluationRun(BaseTeamModel):
...
         blank=True,
         related_name="scoping_runs",
-        help_text="Subset of dataset messages this run evaluated. Empty for FULL/PREVIEW.",
+        help_text=(
+            "The frozen message plan for this run (all dataset ids for FULL, the sample for PREVIEW, "
+            "the explicit list for DELTA)."
+        ),
     )
+    # Coordination state, written only by the beat coordinator under a row lock.
+    evaluator_ids = SanitizedJSONField(default=list)  # evaluator ids frozen at creation
```
SmittieC: Why not use ArrayField instead?

apps/evaluations/models.py
```diff
@@ -491,19 +493,35 @@ def run(
...
+
+            if run_type == EvaluationRunType.PREVIEW:
+                message_ids = list(self.dataset.messages.values_list("id", flat=True)[:PREVIEW_SAMPLE_SIZE])
+            elif run_type == EvaluationRunType.DELTA:
+                message_ids = [message.id for message in (scoped_messages or [])]
+            else:  # FULL
+                message_ids = list(self.dataset.messages.values_list("id", flat=True))
+
+            if message_ids:
+                run.scoped_messages.add(*message_ids)
```
SmittieC, line 525: we need an ADR for this whole change, especially highlighting why scoped_messages is used for full runs now also

apps/evaluations/tasks.py
```diff
@@ -50,6 +50,14 @@
 
 EVAL_SESSIONS_TTL_DAYS = 30
 
+# --- Beat-coordinated evaluation waves (see docs/superpowers/plans/2026-07-23-deploy-safe-eval-runs.md) ---
+BATCH_SIZE = 3  # messages per batch task
+WAVE_SIZE = 10  # batches per wave; WAVE_SIZE * BATCH_SIZE = messages dispatched per wave
```
SmittieC: again, rename to "batch_size"

apps/evaluations/tasks.py
```diff
@@ -61,12 +69,31 @@ def _save_dataset_error(dataset: EvaluationDataset, error_message: str):
     dataset.save(update_fields=["status", "error_message", "job_id"])
 
 
-@shared_task(base=TaskbadgerTask)
+def _pending_evaluator_ids(evaluation_run, message_id, evaluator_ids):
```
SmittieC: type hints

apps/evaluations/migrations/0016_evaluationrun_evaluator_ids_evaluationrun_in_flight_and_more.py
```diff
@@ -0,0 +1,50 @@
...
+        migrations.AddField(
+            model_name='evaluationrun',
+            name='taskbadger_task_id',
+            field=models.CharField(blank=True, max_length=255),
+        ),
+        migrations.AddField(
+            model_name='evaluationrun',
+            name='wave_dispatched_at',
+            field=models.DateTimeField(blank=True, null=True),
+        ),
```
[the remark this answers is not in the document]
SmittieC: No code change needed. All five added columns use **constant** defaults (`ArrayField default=list` → `[]`, `stall_count` → `0`, `taskbadger_task_id` → `''`, `batch_dispatched_at` → `NULL`), so on PostgreSQL 11+ each is a metadata-only `ADD COLUMN` — Django does not rewrite or backfill the table. The only locking operation in this migration is the unique-index build, which is addressed in the sibling thread on the constraint.

apps/evaluations/migrations/0016_evaluationrun_batch_dispatched_at_and_more.py
```diff
@@ -0,0 +1,50 @@
...
+        ),
+        migrations.AlterField(
+            model_name='evaluationrun',
+            name='scoped_messages',
+            field=models.ManyToManyField(blank=True, help_text='The frozen message plan for this run (all dataset ids for FULL, the sample for PREVIEW, the explicit list for DELTA).', related_name='scoping_runs', to='evaluations.evaluationmessage'),
+        ),
+        migrations.AddConstraint(
+            model_name='evaluationresult',
+            constraint=models.UniqueConstraint(fields=('run', 'message', 'evaluator'), name='unique_result_per_run_message_evaluator'),
+        ),
```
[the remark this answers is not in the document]
SmittieC: No code change needed — this is already handled as a documented pre-deploy gate in the PR's **Migrations** section: duplicate `(run, message, evaluator)` rows must be cleaned before the constraint applies (cleanup SQL is included), and runs left in-flight by the old code complete on their partial results on the first post-deploy tick. The rollout `IntegrityError` risk is bounded — the old chord path only produced a duplicate on a retry/race, and both new create sites swallow `IntegrityError`.

apps/evaluations/tasks.py
```diff
@@ -101,30 +133,319 @@ def evaluate_single_message_task(evaluation_run_id, evaluator_ids, message_id):
...
+    """
+    plan_ids = list(run.scoped_messages.values_list("id", flat=True))
+    total = len(plan_ids)
+    evaluator_ids = run.evaluator_ids or []
+
+    _ensure_taskbadger_task(run, total)
+
+    if total == 0 or not evaluator_ids:
+        _finalize_complete(run)
+        return _TickResult(batches=[], done=0, total=total, terminal="success")
```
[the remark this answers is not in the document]
SmittieC: Fixed in 9476980. `_ensure_taskbadger_task` moved out of `_coordinate_locked_run` (which ran inside the row lock + open transaction) to after the transaction commits in `_drive_run`, just before `_publish_tick`. The synchronous `create_task_safe` HTTP call no longer holds the lock while other runs wait in the sweep, and creating the task only post-commit removes the rollback-orphan window. The `taskbadger_task_id` guard keeps it effectively once-per-run; a rare duplicate under overlapping sweeps is harmless for monitoring.

apps/evaluations/models.py
SmittieC: Done in 9732f53. Renamed `wave_dispatched_at` → `batch_dispatched_at`, and every other "wave" in code/comments/tests is now "batch". The 3-message sub-task keeps its existing `batch`/`BATCH_SIZE` name; the per-tick group's `WAVE_SIZE` became `BATCHES_PER_TICK` (see the other thread) to avoid a name collision.

apps/evaluations/models.py
SmittieC: Done in 9732f53. `evaluator_ids` and `in_flight` are now `ArrayField(models.IntegerField(), default=list)` instead of `SanitizedJSONField` — they are always `list[int]`, so ArrayField is the better-typed fit. Migration 0016 was regenerated accordingly.

apps/evaluations/models.py
SmittieC: Done in c35c215. Added `docs/adr/0046-deploy-safe-evaluation-runs.md` (extends ADR-0020) and wired it into the ADR index + mkdocs nav. It calls out specifically why `scoped_messages` is now frozen for FULL runs too: the coordinator reads only the frozen plan, never the live dataset, which `auto_populate_eval_datasets` grows every ~5 min — so a run stays completable even as its dataset grows.

apps/evaluations/tasks.py
SmittieC: Done in 9732f53. `WAVE_SIZE` → `BATCHES_PER_TICK`, kept distinct from the existing `BATCH_SIZE = 3` (messages per batch task) to avoid a name collision. `BATCHES_PER_TICK * BATCH_SIZE` = 30 messages dispatched per tick, unchanged.

apps/evaluations/tasks.py
SmittieC: Done in 5af5ae4. Added type hints to `_pending_evaluator_ids(evaluation_run: EvaluationRun, message_id: int, evaluator_ids: list[int]) -> list[int]` and to the two helpers extracted in the complexity fix (`_create_evaluation_result`, `_run_evaluator_on_message`). I left the coordination helpers' `run` params unannotated for now: typing them as `EvaluationRun` makes `ty` reject the `run.in_flight = ...` ArrayField assignments (a known limitation with Django ArrayField descriptors), and those functions already carry return-type hints. Happy to annotate them too with a targeted `# ty: ignore[invalid-assignment]` if you'd prefer full coverage.

docs/adr/0046-deploy-safe-evaluation-runs.md
```diff
@@ -0,0 +1,39 @@
...
+
+<span class="adr-status adr-status-proposed">PROPOSED</span>
+
+<p class="adr-meta">Author: Open Chat Studio · Created: 2026-07-24</p>
+
+<p class="adr-meta">Extends: <a href="0020-delta-evaluation-runs-scoped-to-appended-messages.md">ADR-0020</a></p>
+
+## Context
+
+Evaluation dispatch used a Celery chord: `run_evaluation_task` split the dataset with `.chunks()`, fanned them out, and a `mark_evaluation_complete` callback fired when all chunks finished. Chord state lives only in Redis and is unrecoverable once lost. A deploy (SIGTERM → SIGKILL) mid-run dropped a chunk's remaining messages: the callback never fired, the run stayed `PROCESSING` forever, and that blocked dataset/evaluator/config edits via `InFlightRunsError` with no recovery path. Auto-population ([ADR-0019](0019-poll-source-experiments-to-auto-populate-eval-datasets.md)) also grows datasets every ~5 min, so a `FULL` run that reads the live dataset at completion time races rows appended after it started.
```
SmittieC: >  run that reads the live dataset at completion time races rows appended after it started.

this is not entirely true

docs/adr/0046-deploy-safe-evaluation-runs.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Decision
+
+We will replace chord dispatch with a stateless beat coordinator that derives all run state from the database each tick.
+
+- **Frozen plan at creation.** `EvaluationConfig.run()` snapshots the message plan into `scoped_messages` and the evaluator ids into a new `evaluator_ids` field for *every* run type — FULL = all current dataset ids, PREVIEW = the sample, DELTA = the explicit list ([ADR-0020](0020-delta-evaluation-runs-scoped-to-appended-messages.md)). This extends `scoped_messages`, previously populated only for DELTA, to all types. The coordinator reads only this frozen plan, never the live dataset, so mid-run auto-population cannot change or stall what a run evaluates.
+- **Beat coordinator.** `coordinate_evaluation_runs` (every 30s) drives each non-terminal run under `select_for_update(skip_locked=True)` so overlapping sweeps partition runs. Each tick recomputes remaining work from `EvaluationResult` rows, then dispatches the next batch (≤ `BATCHES_PER_TICK`×`BATCH_SIZE` messages), re-dispatches a stalled batch's unfinished messages, or completes the run.
```
SmittieC: >  (every 30s)
let's not mention the period as it may change

docs/adr/0046-deploy-safe-evaluation-runs.md
SmittieC: Fixed in 94c48ee8d. You're right — the old `run_evaluation_task` read `dataset.messages.all()` into an in-memory list *at dispatch*, and the `mark_evaluation_complete` chord callback never re-read the dataset, so there was no completion-time race. Reworded the Context to describe what actually happened: the FULL run's scope lived only in that transient list plus the Redis chord state, so a crashed run couldn't be recovered by re-reading the (since-grown) dataset.

docs/adr/0046-deploy-safe-evaluation-runs.md
SmittieC: Fixed in 8859b540f. Dropped the concrete `(every 30s)` from the Beat coordinator bullet, and changed the Consequences "~30s floor" to "a floor of one beat interval" so neither hardcodes the period.

SmittieC:
@snopoke I'm yet to test this locally, so I'm keeping this in draft, but feel free to have a look in the meantime

SmittieC:
> @snopoke I'm yet to test this locally, so I'm keeping this in draft, but feel free to have a look in the meantime

FYI I have tested it locally now and it seems fine

SmittieC:
I see there a few duplicate evaluation results that will cause the unique constraint migration to fail. I'll investigate and see if we can resolve them

snopoke (approved):
LGTM

SmittieC:
I removed duplicate `EvaluationResult` instances (there were 23), so this PR should be safe to deploy. It shouldn't have any real effect, since 1. there were no tags associated and 2. the scores for each one is superceded by the newer one in `_latest_score_per_target`. Even if it had a negative impact, evals can be rerun. cc @snopoke

PR #3947, opened 2026-07-25
Stop chunks of unindexed files reaching collection retrieval

### LINKED ISSUE #3433: Local-index retrieval surfaces chunks from FAILED files
### Bug

`LocalIndexManager.query` in `apps/service_providers/llm_service/index_managers.py` (around line 350) filters `FileChunkEmbedding` rows by `collection_id` only, not by `CollectionFile.status`. When indexing a file fails partway through, the chunks that were already persisted before the failure remain in the database. The `CollectionFile` row is marked `FAILED`, but `query()` still returns those orphaned chunks in retrieval results.

### Reproduction

1. Index a file in a local Indexed Collection where embedding succeeds for chunk 1 but fails for chunk 2 (e.g. a transient provider error on the second chunk).
2. The ``CollectionFile.status`` is set to ``FileStatus.FAILED``.
3. The ``FileChunkEmbedding`` row for chunk 1 is left in the DB.
4. A subsequent ``manager.query(index_id=..., query=...)`` returns that orphan row in its top-k results.

Reproduced locally with a mocked provider during the investigation for #3237.

### Why it matters

User-facing: a chatbot can cite or quote from a partial / corrupted document fragment that should never have been considered indexed. The ``FAILED`` status is the application's signal that the file isn't trustworthy for retrieval, and that signal is currently ignored on the read path.

### Suggested fix sketch

Either:
- Filter ``FileChunkEmbedding`` by a joined ``CollectionFile.status == FileStatus.COMPLETED`` in ``LocalIndexManager.query``, or
- Delete already-persisted chunks when a file fails partway (in the ``except`` branch of ``add_files``).

The filter approach is safer (read-side, idempotent, works retroactively for any already-corrupted collections).

### Context

Surfaced while investigating #3237. Not bundled into that PR because it's an independent code path with an independent fix.

barry47products:
### Product Description

A file that fails partway through indexing into a local collection leaves the chunks it already embedded in the database, and nothing filtered them out on the way back. Chatbots retrieved fragments of documents that never finished indexing and cited them as sourced from that file. Retrying the upload stacked duplicates on top, publishing copied the orphans into the published version permanently, and a worker killed mid-index left partial chunks served as though complete.

### Technical Description

`chunk_from_indexed_file()` (`documents/models.py`) excludes chunks whose `CollectionFile` has not indexed cleanly, and both retrieval queries now use it. On the write side, `add_files` discards its own chunks when indexing fails, `create_new_version` no longer copies chunks of unindexed files, and `retry_failed_uploads` clears stale chunks before re-indexing.

Two decisions where the obvious alternative is wrong:

* **Excluding `PENDING`/`IN_PROGRESS`/`FAILED` rather than requiring `COMPLETED`.** `create_new_version` adds files via `files.add(...)`, leaving `status` blank, so requiring `COMPLETED` returns zero chunks for every published local collection.
* **The retry discard sits in the view, scoped to failed files, not in `add_files`.** `FileChunkEmbedding.working_version` is a CASCADE self-FK, so deleting a working chunk destroys the copies held by published versions.

Both are pinned by tests. Each of the four changes is mutation tested: reverting any one turns exactly one group red.

The two search tool fixtures in `test_tools.py` create chunks with no `CollectionFile` at all, a state production cannot reach. Left alone here, since making them realistic changes no outcome under this filter and the file is already over its length threshold.

Scope note: the issue names `LocalIndexManager.query`, which serves only the Query Collection preview. Chat retrieval uses a near-identical query in `_perform_collection_search`, so both are fixed.

Not addressed here: `CollectionFile` has no uniqueness constraint on `(file, collection)`. Duplicates are not reachable through any current path, but a stale `FAILED` row beside a good one would hide that file's chunks. Fixing it needs a constraint plus deduplicating existing rows.

### Migrations

No migrations. `makemigrations --check` reports no changes.

### Demo

Query Collection on a collection with a failed file: chunks returned before, none after.

### Docs and Changelog

Closes #3433

### DIFF
--- a/apps/chat/agent/tools.py
+++ b/apps/chat/agent/tools.py
@@ -23,7 +23,7 @@ from apps.chat.agent import schemas
 from apps.chat.agent.calculator import calculate
 from apps.chat.agent.openapi_tool import openapi_spec_op_to_function_def
 from apps.chat.models import ChatAttachment
-from apps.documents.models import Collection
+from apps.documents.models import Collection, chunk_from_indexed_file
 from apps.events.forms import ScheduledMessageConfigForm
 from apps.events.models import ScheduledMessage, TimePeriod
 from apps.experiments.models import AgentTools, Experiment, ExperimentSession
@@ -172,6 +172,7 @@ def _perform_collection_search(
     embeddings = list(
         FileChunkEmbedding.objects.annotate(distance=CosineDistance("embedding", query_vector))
         .filter(collection_id=collection.id)
+        .filter(chunk_from_indexed_file())
         .order_by("distance")
         .select_related("file")
         .only("text", "file__name", "file__metadata")[:max_results]
--- a/apps/documents/models.py
+++ b/apps/documents/models.py
@@ -256,9 +256,20 @@ class Collection(BaseTeamModel, VersionsMixin):
                     index_collection_files(collection_files)
             else:
                 # Create versions of file chunk embeddings and add them to the new collection
+                unindexed_file_ids = set(
+                    CollectionFile.objects.filter(collection=self, status__in=UNINDEXED_FILE_STATUSES).values_list(
+                        "file_id", flat=True
+                    )
+                )
                 for embedding in self.filechunkembedding_set.iterator(chunk_size=50):
                     # Skip embeddings for files that are no longer in the collection
                     if embedding.file_id not in file_versions:
+                        continue
+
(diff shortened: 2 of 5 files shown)


barry47products:
@snopoke or @SmittieC kindly review?

snopoke (approved):

PR #3955, opened 2026-07-27
Fix: transaction-atomic-db-exceptions): Fixes transaction.atomic excep…

### LINKED ISSUE #3185: Bug: DB exceptions caught inside `transaction.atomic` blocks causing "can't execute queries until the end of the 'atomic' block" errors
## Summary

We've recently seen a spike in Django errors:

> `An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.`

Root cause: several places in the codebase catch database exceptions (or bare `Exception`) **inside** a `transaction.atomic()` block. When a DB error fires inside an atomic block, PostgreSQL marks the transaction as broken — any subsequent query (including Django's implicit `COMMIT` on block exit) will raise `TransactionManagementError`.

The correct pattern is to wrap the atomic block from the outside, or use a nested `atomic()` as a savepoint.

```python
# Broken
with transaction.atomic():
    try: do_stuff()
    except IntegrityError: handle()   # atomic exits "cleanly" → COMMIT on broken tx

# Fixed (option A — try outside)
try:
    with transaction.atomic():
        do_stuff()
except IntegrityError:
    handle()

# Fixed (option B — nested savepoint, when the outer atomic must continue)
with transaction.atomic():
    try:
        with transaction.atomic():  # savepoint
            do_stuff()
    except IntegrityError:
        handle()
```

## High-severity occurrences (bug confirmed)

| # | Location | Function | Caught | Notes |
|---|---|---|---|---|
| 1 | `apps/filters/views.py:44-49` | `create_filter_set` | `IntegrityError` | `try` inside `with transaction.atomic():`; returns `JsonResponse` from the except branch. |
| 2 | `apps/experiments/models.py:1569-1613` | `ExperimentSession.ad_hoc_bot_message` | bare `Exception` | `@transaction.atomic()` on method; swallowed when `fail_silently=True` (default). |
| 3 | `apps/service_providers/views.py:181-196` | `CreateServiceProvider.form_valid` | bare `Exception` | `@transaction.atomic()`; `sync_voices()` error swallowed, method returns normally. |
| 4 | `apps/assistants/views.py:124-136` | `CreateOpenAiAssistant.form_valid` | `OpenAiSyncError`, `Exception` | `@transaction.atomic()`; catches errors from `push_assistant_to_openai`, still returns `HttpResponseRedirect`. |
| 5 | `apps/assistants/views.py:144-160` | `EditOpenAiAssistant.form_valid` | `OpenAiSyncError`, `Exception` | Same pattern as #4. |
| 6 | `apps/experiments/management/commands/fix_vector_store_duplication.py:35-66` | `Command.handle` | bare `Exception` | `with transaction.atomic():` then `try/except Exception` swallowed (prints traceback only). |
| 7 | `apps/chat/agent/tools.py:427-466` | `AttachMediaTool.action` | `DoesNotExist`, `IntegrityError` | `@transaction.atomic` on method; loop continues after `IntegrityError`, running more queries inside the same broken atomic — cascading failures. **Most likely live culprit** given it runs on every file attach in the chat agent. |

## Lower risk — try/except inside atomic but only catches non-DB exceptions

These don't break the transaction today (the caught exceptions never mark it broken), but should be reviewed to ensure no DB calls in the `try` body can raise anything else:

- `apps/chat/agent/tools.py:521` — catches `Collection.DoesNotExist` only
- `apps/pipelines/management/commands/import_pipeline.py:24,32` — `FileNotFoundError`, `JSONDecodeError`, `DoesNotExist`
- `apps/annotations/models.py:180` — `FieldDoesNotExist`
- `apps/api/views/channels.py:185` — `HTTPStatusError`, `HTTPError`

## Suggested approach

1. Fix the 7 high-severity cases first, starting with `AttachMediaTool.action` (#7) as it's the most likely production culprit.
2. Add a lightweight lint/test — e.g. an AST check in CI — to prevent new occurrences of the broken pattern.
3. Audit lower-risk cases and tighten where easy.

## How the list was produced

AST scan across `apps/**/*.py` for functions decorated with `@transaction.atomic` or containing `with transaction.atomic():`, then flagging any `try/except` nested inside the atomic scope. Each hit was verified by reading the code to distinguish true inside-atomic try/except from the correct outside-atomic pattern.

brian-mecha:
### Product Description

Fixes a class of production errors where users would see a failed request or a silently discarded save instead of the intended outcome. The visible symptom was a spike in TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.

### Technical Description

Root cause. A DB error inside transaction.atomic() leaves the transaction aborted. Catching it without leaving the block means the next query or the block's own exit raises. Verified against a real database.

What to look at:

1. Pattern choice per site. Nested atomic() savepoint where the outer transaction must still commit (AttachMediaTool, both assistant form_valids, run_post_save_hook); try hoisted outside where the work should be discarded (create_filter_set, ad_hoc_bot_message, the vector-store command). The savepoint form preserves current behaviour. Hoisting in the assistant views would also roll back the local save on a sync failure, a product change I avoided.

2. Tests use real aborted transactions. A mocked IntegrityError doesn't reproduce this bug; the three regression tests use duplicate-key/raw-SQL failures and were verified to fail before the fix.

3. New AST check (scripts/check_atomic_exception_handling.py) runs as a pre-commit hook plus a repo-wide test. It's intra-file, so it can't catch the cross-function variant (run_post_save_hook).

4. Also included: four other narrow catches reviewed and left alone (can't break the transaction), plus two adjacent fixes:
    -  a @transaction.atomic()/@staticmethod ordering bug
    - two redundant atomic blocks on read-only paths.

### Docs and Changelog

Fixes #3185

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -73,3 +73,12 @@ repos:
         pass_filenames: false
         always_run: true
         fail_fast: true
+      - id: atomic-exception-handling
+        name: database exceptions must not be caught inside transaction.atomic()
+        # Stdlib-only AST check, so it needs no virtualenv of its own. The repo-wide
+        # gate lives in scripts/test_check_atomic_exception_handling.py; this hook is
+        # the fast signal on the files a commit or PR actually touches.
+        entry: python3 scripts/check_atomic_exception_handling.py apps --files
+        language: system
+        files: ^apps/.*\.py$
+        exclude: /migrations/
--- a/apps/chat/agent/tools.py
+++ b/apps/chat/agent/tools.py
@@ -484,11 +484,12 @@ class AttachMediaTool(CustomBaseTool):
 
         response = []
         include_links = self.experiment_session.experiment_channel.platform == ChannelPlatform.WEB
+        chat_attachment = self.chat_attachment
         for file_id in file_ids:
             try:
                 with transaction.atomic():
                     file = File.objects.get(id=file_id)
-                    self.chat_attachment.files.add(file_id)
+                    chat_attachment.files.add(file_id)
                     self.tool_callbacks.attach_file(file_id)
                     file_response = SUCCESSFUL_ATTACHMENT_MESSAGE.format(file_id=file_id, name=file.name)
 
--- a/apps/chat/agent/tools.py
+++ b/apps/chat/agent/tools.py
@@ -477,16 +477,23 @@ class AttachMediaTool(CustomBaseTool):
         )
         return chat_attachment
 
-    @transaction.atomic
     def action(self, file_ids: list[int]) -> str:
         if len(file_ids) > 5:
             return "A maximum of 5 files can be attached."
 
         response = []
         include_links = self.experiment_session.experiment_channel.platform == ChannelPlatform.WEB
+        # Resolve the parent row here, outside the per-file blocks below. `chat_attachment` is a
+        # cached_property, so leaving it to be evaluated on the first loop iteration puts its
+        # get_or_create inside that file's transaction: if the file fails, the rollback drops the row
+        # while the cached object keeps its id, and the m2m table's deferred foreign key then fails
+        # at COMMIT — after every later file has been reported attached.
         chat_attachment = self.chat_attachment
         for file_id in file_ids:
             try:
+                # One transaction per file, so a failed attachment rolls back on its own and the
+                # loop can keep going. Catching a DB error without leaving the block would abort
+                # the transaction and break every remaining iteration.
                 with transaction.atomic():
                     file = File.objects.get(id=file_id)
                     chat_attachment.files.add(file_id)
--- a/apps/chat/agent/tools.py
+++ b/apps/chat/agent/tools.py
@@ -486,28 +486,29 @@ class AttachMediaTool(CustomBaseTool):
         include_links = self.experiment_session.experiment_channel.platform == ChannelPlatform.WEB
         for file_id in file_ids:
             try:
-                file = File.objects.get(id=file_id)
-                self.chat_attachment.files.add(file_id)
-                self.tool_callbacks.attach_file(file_id)
-                file_response = SUCCESSFUL_ATTACHMENT_MESSAGE.format(file_id=file_id, name=file.name)
+                with transaction.atomic():
+                    file = File.objects.get(id=file_id)
+                    self.chat_attachment.files.add(file_id)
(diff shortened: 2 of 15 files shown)


apps/chat/agent/tools.py
SmittieC: is this transaction still needed if we have one in each loop?

brian-mecha: @SmittieC Valid question. I have fixed this by hoisting `chat_attachment` above the loop, so the parent row is created in the outer transaction and survives a per-file rollback. 
By keeping the decorator, it now has a real job (_grouping the parent-row creation with the attachments_), and it gives all-or-nothing semantics across files.

.pre-commit-config.yaml
```diff
@@ -73,3 +73,12 @@ repos:
         pass_filenames: false
         always_run: true
         fail_fast: true
+      - id: atomic-exception-handling
+        name: database exceptions must not be caught inside transaction.atomic()
```
SmittieC, line 77: This is a good idea. How long does it take to run this script generally? I'm just worried it will make all commits slower. We could run this as a test instead if it takes too long?

SmittieC:
Would you mind checking the code rabbit comments and commenting + resolving them as well please? Sometimes its off, but other times it has legit conerns. Also resolving clutters the PR less. Thanks!

apps/chat/agent/tools.py
SmittieC: I don't follow how this different from what was there previously?

.pre-commit-config.yaml
brian-mecha: The hook only gets the files the commit touches. One changed file is `0.02s`.  A full apps/ scan is `~0.7s` across 1,184 files, but that only happens if you run it manually.

It's fast because it's a pure-stdlib AST parse: no imports of the project, no Django setup, no DB. That's also why it's language: system with python3 rather than uv run: no virtualenv to build or activate.

It's also already wired as a test, so we get both test_the_repo_is_clean in scripts/test_check_atomic_exception_handling.py scans the whole tree on every CI run. That's the actual gate (_it catches violations arriving via a merge or a file rename, which a changed-files hook can't see_) the pre-commit hook is just the fast local signal so you find out before pushing.

apps/chat/agent/tools.py
brian-mecha: I see your point. I have removed the transaction.

SmittieC (approved):
LGTM. Thanks!

PR #3964, opened 2026-07-27
Clean up gevent references in celery config

snopoke:
### Product Description
No change.

### Technical Description
Production workers run with the thread pool, but `heroku.yml`, `docker-compose.prod.yml`, the hosting docs and the `inv celery` task all still specified `--pool gevent`. Swapped them to `--pool threads` and removed the now-unused `gevent` prod dependency.

Worth a look: concurrency in the prod commands drops from 100 to 20. 100 OS threads each holding a DB connection would exhaust the pool, but if the deployed value is actually something else, that's the number to use here.

The `inv celery --gevent` dev flag is renamed to `--threads` (concurrency stays at 10).

### Docs and Changelog

Hosting docs updated in-repo; no changelog entry needed.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -72,7 +72,7 @@ services:
 
   celery_worker:
     <<: *app
-    command: celery -A config worker -l INFO --pool gevent --concurrency 100
+    command: celery -A config worker -l INFO --pool threads --concurrency 20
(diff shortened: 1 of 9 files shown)


SmittieC (approved):
Goodbye gevent, and thank you for your service 🫡

PR #3965, opened 2026-07-28
fix: use static names for eval tasks and include the run id in the data

SmittieC:
### Product Description

No user-facing change. Newly created evaluation runs now start on the next coordinator tick (up to ~30s) instead of immediately.

### Technical Description

Improves the Taskbadger integration for evaluation runs and fixes duplicate Taskbadger tasks:

* Use a static name ("Evaluation run") for the per-run Taskbadger task and include the run id, run type, config and dataset in the task data, so Taskbadger groups these tasks by name instead of creating a distinct name per run.

### Migrations

No migrations.

### Demo

N/A

### Docs and Changelog

### DIFF
--- a/apps/evaluations/tasks.py
+++ b/apps/evaluations/tasks.py
@@ -435,13 +435,22 @@ def _ensure_taskbadger_task(run: EvaluationRun, total: int) -> None:
     """
     if run.taskbadger_task_id:
         return
+    # Below the guard on purpose: one query per run, not one per tick per active run.
+    config = EvaluationConfig.objects.select_related("dataset").get(id=run.config_id)
     task = taskbadger.create_task_safe(
         name="Evaluation run",
         status=StatusEnum.PROCESSING,
         value=0,
         value_max=total,
         stale_timeout=TASKBADGER_STALE_TIMEOUT,
-        data={"run_id": run.id},
+        data={
+            "run_id": run.id,
+            "run_type": run.type,
+            "config_id": config.id,
+            "config_name": config.name,
+            "dataset_id": config.dataset_id,
+            "dataset_name": config.dataset.name,
+        },
     )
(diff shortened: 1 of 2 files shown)


apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -364,6 +364,13 @@ def test_shared_provider_inlined_identically(inspect_bot):
     assert _node(payload, "Route")["llm"] == _node(payload, "Answer")["llm"] == _expected_llm(inspect_bot)
 
 
+@pytest.mark.django_db()
+def test_graph_and_nodes_agree_on_order(inspect_bot):
```
snopoke: @SmittieC These other changes seem unrelated to the static eval task name - do you want to separate them out or bundle them?

apps/evaluations/tasks.py
```diff
@@ -436,11 +436,12 @@ def _ensure_taskbadger_task(run: EvaluationRun, total: int) -> None:
     if run.taskbadger_task_id:
         return
     task = taskbadger.create_task_safe(
-        name=f"Evaluation run {run.id}",
+        name="Evaluation run",
         status=StatusEnum.PROCESSING,
         value=0,
         value_max=total,
         stale_timeout=TASKBADGER_STALE_TIMEOUT,
+        data={"run_id": run.id},
```
snopoke: How about adding some more info in there to make debugging easier: evaluation config id and name, dataset id and name, run type (full, preview, delta)

SmittieC: https://github.com/dimagi/open-chat-studio/pull/3965/commits/9273f30df451e8d86e8f15793fb9c898a98656ec

snopoke (dismissed):

Issue #3963, opened 2026-07-27
Large evaluation datasets (10k+ sessions) fail silently: worker OOM leaves dataset stuck in PROCESSING


snopoke:
> **Revised after #3912 ("deploy-safe evaluation runs", ADR-0047) landed.** That PR replaced the chord-based eval dispatch with a stateless beat coordinator and resolves several items this issue originally listed. The revised assessment is in "Impact of #3912" below. **The root cause — dataset *creation* OOM — is untouched by #3912 and stands as written.**

## Problem

Creating a **session-mode evaluation dataset** from a large number of sessions (reported at >10,000) fails. The user-visible symptom is not an error: the dataset sits at `PROCESSING` indefinitely with the progress bar frozen at 40%, and no error message is ever shown.

Investigation found that dataset *creation* is roughly the fifth blocker in the lifecycle. A 10k-session dataset currently cannot be selected in production, cannot be evaluated to completion, cannot have its results viewed, cannot be exported, and cannot be deleted cleanly.

## Root cause: worker OOM, failing silently

`make_session_evaluation_messages` (`apps/evaluations/utils.py:208`) materialises **every ChatMessage across all selected sessions** into one Python list:

```python
all_messages = list(
    ChatMessage.objects.filter(chat__experiment_session__external_id__in=session_external_ids)
    .annotate(...)
    .order_by("created_at")
)
```

`list(qs)` uses `chunked_fetch=False`, so libpq holds the entire raw result set at the same time as the growing instance list. `all_messages`, the `sessions` grouping dict, `session_map`, and the result list are all simultaneously alive; nothing is freed until the function returns.

Per-object costs below were **measured** with `tracemalloc` against the real model classes (Python 3.13, Django 5.2.15). Baseline Celery worker RSS after `django.setup()` is 274 MB. The totals are extrapolated from those measurements, not from an end-to-end production run.

| 10k sessions at | ChatMessage list | Estimated peak RSS |
|---|---|---|
| 6 msgs/session, 200 chars | 127 MB | ~480 MB |
| 20 msgs/session, 500 chars | 797 MB | ~1.4 GB |
| 60 msgs/session, 1500 chars | 5.2 GB | ~8.1 GB |

A typical production chatbot (20 messages/session, ~500-char turns) already exceeds a 1 GB worker. This is not a pathological case.

### Why the failure is invisible

[...]

PR #3972, opened 2026-07-28
fix: stream session-mode dataset population, correct trace sourcing

snopoke:
Creating a session-mode evaluation dataset from more than a few thousand sessions OOM-killed the Celery worker (#3963). Because that is a SIGKILL, and `task_acks_late` is globally inert, the dataset was left stuck in `PROCESSING` forever with no error and the progress bar frozen at 40%.

### Product Description

Large session-mode datasets can now be created, and a failure reports what actually went wrong plus how far it got. Re-running resumes rather than restarting.

Session-mode datasets also now carry real `participant_data`/`session_state`; previously both were always empty.

### Technical Description

`make_session_evaluation_messages` materialised every `ChatMessage` across every selected session into one list. Measured per-object costs put peak RSS around 1.4 GB at a typical 20 messages/session — over a 1 GB worker. It is now a generator that loads one chunk of sessions at a time, and `add_messages_stream` commits per batch instead of holding one transaction (and one of only 35 pool connections) for the whole job.

Worth review attention:

- **`participant_data` was silently always `{}`.** The old code read it from the last AI message's `input_message_trace`, but production sets `Trace.input_message` to the *human* message (`apps/channels/stages/core.py`) and the AI message to `output_message`. The pre-existing test passed only because it wired traces the other way round; that test was corrected rather than kept. This changes what session-mode evaluators see.
- **`session_state` now comes from `ExperimentSession.state`, not the trace.** `Trace.session_state` is only the snapshot from when the trace *opened*; post-turn state is written back to the session. Sourcing it from the trace would have left it a turn behind the now-end-of-turn `participant_data`. This matches how exports report session state.
- **Per-batch commits replace one long lock.** Each batch still takes the same dataset row lock, which is all the dedup invariant needs — mutual exclusion only has to cover one read-check-insert cycle. Two concurrent adds interleave between batches without duplicating. `_existing_dedup_keys` gained a `candidates` argument so each batch does one index-assisted lookup instead of rescanning the dataset.
- **External ids are resolved to pks once**, then chunked. Keyset-paginating the `external_id` queryset would have re-sent the whole IN-list on every page (10k bind parameters per chunk, and a hard ceiling at PostgreSQL's 65535-parameter limit).
- `EvaluationConfig.run()` takes `scoped_message_ids` instead of instances, since it only ever needed ids.

Deliberately out of scope, tracked in #3963: users still cannot *select* 10k sessions in production (the WAF caps the POST body at ~215 ids — that needs a matching `export_waf_allow_list` regeneration in ocs-deploy), and message-mode has the same reverse-FK row multiplication in a different code path.

Chunking is the fix here, not a timeout — `soft_time_limit` is silently discarded by the gevent pool, which is worth its own issue.

### Migrations

None.

### Docs and Changelog

### DIFF
--- a/apps/evaluations/models.py
+++ b/apps/evaluations/models.py
@@ -412,17 +412,8 @@ class EvaluationDataset(BaseTeamModel):
             keyed_batch = [(message, self._dedup_key(message)) for message in batch]
             with transaction.atomic():
                 EvaluationDataset.objects.select_for_update().get(pk=self.pk)
-                candidates = {key for _, key in keyed_batch if key is not None}
-                seen = self._existing_dedup_keys(candidates)
-                to_create = []
-                for message, key in keyed_batch:
-                    if key is not None:
-                        if key in seen:
-                            skipped += 1
-                            continue
-                        seen.add(key)
-                    to_create.append(message)
-
+                to_create, batch_skipped = self._partition_new_in_batch(keyed_batch)
+                skipped += batch_skipped
                 if to_create:
                     created = EvaluationMessage.objects.bulk_create(to_create, batch_size=batch_size)
                     self.messages.add(*created)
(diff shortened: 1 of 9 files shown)


apps/evaluations/models.py
```diff
@@ -352,16 +358,78 @@ def _dedup_key(self, message: EvaluationMessage) -> tuple | None:
...
+                        seen.add(key)
+                    to_create.append(message)
+
+                if to_create:
+                    created = EvaluationMessage.objects.bulk_create(to_create, batch_size=batch_size)
+                    self.messages.add(*created)
+                    created_ids.extend(message.id for message in created)
+            if progress_callback:
+                progress_callback(len(created_ids), skipped)
+        return created_ids, skipped
```
snopoke, line 423: Fixed in d49c05d — extracted the per-batch dedup loop into `_partition_new_in_batch`, dropping the nested depth from 5 to 2.

*— Claude Code*

SmittieC (approved):
Generally looks good to me

Issue #3971, opened 2026-07-28
Surface index failure reasons for remote-index collections


barry47products:
Follow-up to #3164, split out per @snopoke's feedback. That PR surfaces `failure_reason` for local-index collections only; remote-index collections still show a bare "Failed" badge with no explanation.

The fix is not symmetric with the local path because both remote raise sites discard the provider detail before the handler in `RemoteIndexManager.add_files` sees it:

- `_ensure_remote_file_exists` raises `FileUploadError() from None`, dropping the cause entirely.
- `link_files_to_remote_index` raises `UnableToLinkFileException("Failed to link files to OpenAI vector store") from e`, replacing the provider message with a static string.

So the reason needs to be captured at those two raise sites (e.g. carried on the exception, formatted with the existing `format_failure_reason`) rather than in the handlers. The `add_files` handlers then write it: per file for `FileUploadError`, and to the whole batch for `UnableToLinkFileException`, since a link failure applies to every file in the call.

Everything downstream already works for remote once the field is populated: clearing on re-index, the tooltip, and the aria-label all key off `CollectionFile.failure_reason` with no local/remote distinction.

Context: #3161 (the bad-API-key report that motivated #3164) was a remote-path failure, so this half is where one of the two original reports actually originated.

barry47products:
@SmittieC design for #3971 below. Two calls I'd like from you before I start.

1. The provider message can go onto the exception at the raise site, or `format_failure_reason` can walk the chain to reach it. `from None` leaves `__context__` bound, so both work. I've written it up the first way to keep that function a formatter, but the second touches neither raise site.

2. Chasing this down turned up five paths, and only two of them are the ones this issue was split out for. The write-up covers all five, on the grounds that "every remote failure says why" is a more useful thing to ship than two better strings. Say if you'd rather I cut it back to the two and file the rest.

## What gets written today

| Site | Granularity | Reason written today |
| --- | --- | --- |
| `index_managers.py:150` no extractable text | per file | `FileReadException: No text could be extracted from this file` |
| `index_managers.py:156` upload fails | per file | `FileUploadError` |
| `index_managers.py:172` linking fails | whole batch | `UnableToLinkFileException: Failed to link files to OpenAI vector store` |
| `documents/tasks.py:171` linking fails during collection-from-assistant | whole batch | nothing written |
| `documents/tasks.py:59` anything `add_files` doesn't catch | whole batch | nothing written, rows stay `IN_PROGRESS` |

[...]

Issue #3162, opened 2026-04-11
Local-index file indexing fails silently in LocalIndexManager.add_files (status=Failed, 0 chunks)


jjackson:
## Summary

Files uploaded to a local-index collection (`is_index=True`, `is_remote_index=False`, `llm_provider=<OpenAI>`, `embedding_provider_model=text-embedding-3-small`) transition from `In Progress` → `Failed` within ~6 seconds, with `0 chunks` stored. The exception is caught and logged inside `LocalIndexManager.add_files` but never surfaced to the UI beyond the "Failed" badge.

Because the broad `except Exception as e: logger.exception(...)` in `add_files` swallows the actual error, users see a failing upload with no actionable information. Any real diagnosis requires server log access.

> **Note:** I don't have a local OCS dev environment set up, so I couldn't reproduce this in-tree to produce a verified PR fix. Filing this as an issue with the evidence I have.

## Repro against `https://chatbots.dimagi.com`

Team: `connect-ace`. Super Admin user. Service provider 378 = OpenAI with valid funded key. Embedding model 1 = `text-embedding-3-small`.

```bash
# 1. Create a local-index collection
curl -X POST https://chatbots.dimagi.com/a/connect-ace/documents/collection/new/ \
  -H 'X-CSRFToken: <token>' -b '<session>' -H 'Referer: https://chatbots.dimagi.com' \
  -d 'name=probe&summary=probe&is_index=True' \
  -d 'llm_provider=378&embedding_provider_model=1' \
  -d 'csrfmiddlewaretoken=<token>'
# → 302 → /a/connect-ace/documents/collections/<cid>

# 2. Upload a small markdown file
curl -X POST https://chatbots.dimagi.com/a/connect-ace/documents/collections/<cid>/add_files \
  -H 'X-CSRFToken: <token>' -b '<session>' -H 'Referer: https://chatbots.dimagi.com' \
  -F 'csrfmiddlewaretoken=<token>' \
  -F 'files=@idd.md'
# → 200, File created, CollectionFile created, Celery task scheduled

# 3. Poll status
curl https://chatbots.dimagi.com/a/connect-ace/documents/collections/<cid>/files/<cf_pk>/status
# → t=0s: data-tip="In Progress", 0 chunks
# → t=6s: data-tip="Failed", 0 chunks  ← stuck here
```

[...]

Issue #3161, opened 2026-04-11
Collection create with is_remote_index=True crashes with 500 (OpenAI provider, valid key)


jjackson:
## Summary

Creating an indexed collection with `is_remote_index=True` on an OpenAI-typed LLM provider crashes inside `ensure_remote_index_created()` → `LlmProvider.create_remote_index()` → `client.vector_stores.create(...)`, returning an opaque 500 to the user. The same provider, with the same API key, successfully creates chat completions, embeddings, and vector stores when called directly via curl.

> **Note:** I don't have a local OCS dev environment set up, so I couldn't reproduce this in-tree to produce a verified PR fix. Filing this as an issue with the evidence I have.

## Repro against `https://chatbots.dimagi.com`

Team: `connect-ace`. User account has Super Admin on the team. Service provider 378 configured as Type=OpenAI with a valid `sk-svcacct-*` key and a funded OpenAI organization.

```bash
# Submit the collection create form
curl -X POST https://chatbots.dimagi.com/a/connect-ace/documents/collection/new/ \
  -H 'X-CSRFToken: <token>' \
  -b '<session-cookie>' \
  -H 'Referer: https://chatbots.dimagi.com' \
  -d 'name=probe-remote&summary=probe' \
  -d 'is_index=True' \
  -d 'is_remote_index=on' \
  -d 'llm_provider=378' \
  -d 'embedding_provider_model=1' \
  -d 'csrfmiddlewaretoken=<token>'
```

**Observed:** `HTTP 500` with the generic `<h1>Shucks. That's an error.</h1>` page. No error details surfaced to the client.

**Expected:** `HTTP 302 Location: /a/connect-ace/documents/collections/<new_id>`, like `is_remote_index=False` does.

## What I ruled out

| Hypothesis | Tested | Result |
|---|---|---|
| Permissions issue | User is Super Admin on team | ✗ not it |
| API key invalid | `POST https://api.openai.com/v1/embeddings` with same key | ✓ HTTP 200, valid embedding returned |
| No billing / quota | Direct `POST /v1/embeddings` | ✓ charges go through |
| OpenAI vector stores API broken for this key | `POST https://api.openai.com/v1/vector_stores -H 'OpenAI-Beta: assistants=v2' -d '{\"name\":\"test\"}'` with same key | ✓ HTTP 200, returned `vs_69d9b8028cd08191bbf3502097fa08ac` |
[...]

PR #3973, opened 2026-07-28
Surface index failure reasons for collection files

### LINKED ISSUE #3164: Improve error reporting for index collection files
If a file uploaded to an indexed collection fails to index there is no indication of what the error was.

* Add a failure_reason field (TextField, blank=True) to CollectionFile
* Store the exception message when indexing fails in LocalIndexManager.add_files()
* Show it in the UI, even just a tooltip on the failed badge

barry47products:
### Product Description

When a file fails to index into a collection, the only signal today is a red "Failed" dot with no explanation - diagnosing even a bad API key means asking someone with server log access (#3162 / #3161 were filed over exactly this). This PR stores the reason for the failure and shows it in the tooltip on the failed badge, so users can self-diagnose wrong file formats, provider misconfiguration, and auth errors directly from the collection page.

The reason always describes the most recent attempt: it is cleared whenever a file re-enters indexing (retry, provider change, document source re-sync) and on success.

Fixes #3164

### Technical Description

- `CollectionFile.failure_reason` (`TextField`, `blank=True`), written in `LocalIndexManager.add_files()` from the exception it already catches, formatted as `"{ExceptionType}: {message}"` - whitespace collapsed to one line, NUL bytes stripped, lone surrogates replaced (either would otherwise fail the Postgres save; a surrogate raises `UnicodeEncodeError`, which is not a `DatabaseError`, so it would escape the recovery block and strand the rest of the batch at `IN_PROGRESS`), truncated to 500 characters.
- Cleared at the `IN_PROGRESS` transition in `index_collection_files` (the choke point every re-index route passes through), in `retry_failed_uploads`, in `_update_file` on document source re-sync, and on the success branch of `add_files`.
- Rendered in `templates/documents/collection_file_status.html` as the `data-tip` on the status badge (with `tooltip-error` styling and `before:wrap-break-word`, since daisyUI sets no word-breaking and a masked API key is an unbroken ~60-char token that otherwise overflows the bubble) and as the dot's `aria-label`, truncated to 100 characters so screen readers get a short accessible name rather than the full provider blob.
- Scope is the local index path only, per the issue. The remote index path needs the reason captured at its raise sites (`FileUploadError() from None` discards the cause) and is split out to #3971 as agreed on the ticket.

Notes for reviewers:

- Provider error text is only masked by the provider's own convention (OpenAI returns `sk-svcac****...im8A`). Truncation caps exposure; a redaction pass was considered and deferred - flagging in case maintainers want it now.
- XSS reviewed: no `|safe`/`mark_safe` in the chain, Django attribute-escaping applies, and `data-tip` is rendered via CSS `content: attr()`, never HTML-parsed.
- `failure_reason` flows into the team export payload via `build_resource_serializer` (`fields="__all__"`). `collectionfile` is not a sealed manifest entry; sealing the whole entry for one incidental field seemed disproportionate, but it is a maintainer call.

### Migrations

Single additive `AddField` (`0018`); `TextField(blank=True)` with implicit `""` default.

### Demo

Failed file with the reason in the tooltip (real OpenAI 401 body, wrapping inside the bubble):
tooltip-long-401-wrapped

Short reason and the completed fallback: 
tooltip-short-reason
tooltip-completed-fallback

### Docs and Changelog


apps/documents/models.py
```diff
@@ -42,6 +42,26 @@ class FileStatus(models.TextChoices):
...
+    whether the fault is auth, quota, or input.
+    """
+    # NUL bytes and lone surrogates (surrogateescape-decoded filenames in OSError messages) both
+    # fail the Postgres save -- and a surrogate raises UnicodeEncodeError, which is not a
+    # DatabaseError, so it would escape add_files' recovery block and strand the whole batch.
+    # Same precedent as the NUL-stripping in add_files' chunk handling.
+    message = " ".join(str(exc).split()).replace("\x00", "")
+    message = message.encode("utf-8", errors="replace").decode("utf-8")
+    reason = f"{type(exc).__name__}: {message}" if message else type(exc).__name__
+    return reason[:FAILURE_REASON_MAX_LENGTH]
```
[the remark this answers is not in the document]
barry47products: @coderabbitai, fixed in 832710bcb - NULs are now stripped before whitespace collapsing, with a regression case (`nul_between_spaces_leaves_single_space`) covering a NUL flanked by spaces.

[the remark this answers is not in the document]
barry47products: @coderabbitai Fixed in 832710bcb - NULs are now stripped before whitespace collapsing, with a regression case covering a NUL flanked by spaces.

barry47products:
@snopoke kindly review ?

templates/documents/collection_file_status.html
```diff
@@ -1,6 +1,10 @@
 <div id="file-status-{{ collection_file.id }}"
-     class="tooltip"
-     data-tip="{{ collection_file.status_enum.label }}"
+{# `before:wrap-break-word` targets the tooltip body, which daisyUI renders as a ::before pseudo-element.
+   It already wraps at 20rem but sets no word-breaking, so an unbroken token longer than that
+   overflows the bubble -- and a provider error carrying a masked API key (sk-svcac****...im8A) is
+   exactly such a token. #}
```
snopoke: django templates don't support multi-line comments. You should use the {% comment %}{% endcomment %} tags.

barry47products: Thanks for picking that up, I'll push up a small change.

barry47products: Good catch, thanks - the multi-line `{# #}` was rendering literally into the output. Switched to `{% comment %}{% endcomment %}` tags in 94bfe27 and verified the rendered page is clean.

snopoke (approved):

Issue #3989, opened 2026-07-28
Message-level dataset clones still materialise every session id for "all matching filters"


snopoke:
## Context

Part 3 of #3963 ("store the filter, not the IDs") landed for **session-level** datasets: `create_dataset_from_sessions_task` now accepts a serialized session filter, re-resolves it inside the worker, and streams sessions through `add_messages_stream`, so "all sessions matching the current filters" never materialises a list of UUIDs.

**Message-level** datasets still resolve that scope to an explicit id list. To stop the unbounded case reaching production in the meantime, both entry points now cap it at `MESSAGE_MODE_ALL_MATCHING_LIMIT = 1000` sessions (`apps/evaluations/dataset_clone.py`):

- `EvaluationDatasetBaseForm._clean_clone_all_matching` (create form) — form error
- `resolve_sessions_to_clone` → `EvalDatasetAddSessionsView.post` (Add Sessions page) — `messages.error`

Session-level is exempt: it ships the filter, so it has no ceiling.

## Why message mode can't just be handed the filter

Two separate ceilings, not one:

1. The ids travel in the Celery message (~37 bytes each).
2. `EvaluationMessage.create_from_sessions` (`apps/evaluations/models.py:184`) returns a `list` of every message across every selected session, and `dataset.add_messages(...)` takes the whole list. This is the same materialisation #3963 identifies as the root cause, sitting in the message-mode task rather than the session-mode one.

So lifting the cap means doing the streaming rewrite too — a filter parameter alone would just move the OOM from the broker to the worker.

## Work

- Give `create_from_sessions` a generator form that yields per chunk of sessions (mirror `iter_session_evaluation_messages_for_sessions` in `apps/evaluations/utils.py`, which pages over session pks at `SESSION_CHUNK_SIZE`).
- Switch `create_dataset_from_session_messages_task` to `add_messages_stream` with a per-batch progress callback, as `create_dataset_from_sessions_task` does.
- Add a **separate** parameter for the session filter. The task's existing `filter_query` is a `ChatMessageFilter` applied to `filtered_session_ids` (the "filtered messages only" clone scope) — a different filter over a different model, so reusing the name would conflict.
- Route `all_matching` through it in `_clean_clone_all_matching` and `resolve_sessions_to_clone`, and delete `MESSAGE_MODE_ALL_MATCHING_LIMIT` plus the two client-side warnings that reference it.
[...]

PR #3990, opened 2026-07-28
Clone eval datasets from a stored session filter instead of a list of ids

snopoke:
Part 3 of #3963 — "store the filter, not the IDs". Selecting 10k sessions on the Create Dataset / Add Sessions pages currently means the browser fetching every session UUID as JSON, posting them back in a hidden input (~390 KB, over the WAF body limit at a few hundred), and the task holding a 10k-parameter `IN` clause. Session-mode clones now carry the *filter* instead: the browser only ever asks for a count, and the task re-resolves the filter and streams sessions through `add_messages_stream`.

### Product Description
"All matching filters" is now a scope on the session selector rather than a bulk pre-selection of rows, so a filtered selection of any size can be cloned into a session-level dataset. Datasets of either level can also be created empty and filled later.

### Technical Description
Worth the reviewer's attention:

- **`""` vs `None` for `filter_query`** is load-bearing. `""` means "no active filters — every session in the team"; `None` means "use the id list instead". `SessionSelection` carries one or the other, never both.
- **The pk ceiling in `_sessions_for_session_mode_clone`.** The task counts sessions and then pages over them in separate queries, so a relative filter ("last 7 days") resolving against a moving `now()` would let the count and the walk disagree. `pk__lte=max(pk)` snapshots the set at job start.
- **Filters are read from the POST body, not the query string** (`CreateDataset.get_form_kwargs`, and the existing behaviour in `EvalDatasetAddSessionsView`). The form mirrors them as hidden inputs; falling back to the URL would silently widen the resolution to the whole team if the query string were ever lost.
- **Message mode is capped at 1000 sessions** (`MESSAGE_MODE_ALL_MATCHING_LIMIT`) rather than being given the filter. It has a second ceiling the session path doesn't: `EvaluationMessage.create_from_sessions` builds every message for every session in one list before anything is written, so a filter param alone would move the OOM from the broker to the worker. Lifting it needs the streaming rewrite — #3989. Enforced at both entry points and warned about client-side.
- **Both levels may now be created empty.** Previously session-mode only, justified by auto-population being session-only; either level can be filled from the Add Sessions page, so the asymmetry was arbitrary. No job is dispatched, so the dataset is closed out at `COMPLETED` at save time rather than sitting at `PENDING`.
- `dataset_sessions_selection_json` → `dataset_sessions_selection_count`. Nothing else consumed the id list.

### Deploy note
`EvalDatasetAddSessionsView` gained `@waf_allow(SizeRestrictions_BODY)` for the hand-picked path, which still posts one UUID per row. The decorator alone changes nothing in production — the regexes from `export_waf_allow_list` have to be pasted into the ocs-deploy `waf` module.

### Migrations
N/A — no schema change.

### Demo
On the session list, filter to something large, then Create dataset → the scope toggle arrives on "All matching filters" with a live count, and no session ids appear anywhere in the page source.

### Docs and Changelog

The user-visible change is the "Selected only / All matching filters" scope toggle on dataset creation and Add Sessions, and that large filtered selections now work.


SmittieC (approved):
LGTM (I haven't reviewed it super closely, but it seems like good changes)

PR #3996, opened 2026-07-29
Feat: Adds notifications for all teams

### LINKED ISSUE #3430: The user notifications page should show notifications from all teams
The Notifications list currently only shows notifications from the current team which makes sense since it is a team scoped view. However the nav link is under the "My Account" section which suggests it is global.

I think it would be useful to have the notifications page show the current teams notifications by default but allow you to see notifications from other teams without having to switch to that team and re-navigate to the notifications page.

brian-mecha:
### Product Description

The Notifications page is no longer scoped to a single team. It now lives at `/notifications/` and shows notifications from every team you belong to by default, with a `Team filter` to narrow to one. 
The notification badge in the nav counts unread notifications across all your teams. `"Mark all as read"` now respects whatever filters (_including team_) are currently applied to the list. 
`Do Not Disturb` and `muting` remain per-team, unchanged.

### Technical Description

- Moved notifications out of `team_urlpatterns` into a global `/notifications/` route; views switched to plain `LoginRequiredMixin`.
- `UserNotificationTableView` now queries `team__in=request.user.teams.all()` instead of `team=request.team`, with a new `TeamFilter` (_reusing the existing filter-bar machinery_) to narrow to one team. Added a team badge per row for readability.
- Read/mute/unmute/event-detail views updated to scope by the row's own team (or `request.user.teams.all()`) instead of request.team, since rows can now belong to any of the user's teams.
- New `unread_notifications_count_all_teams()` sums the per-team counts and caches the aggregate under `{user_id}-all-teams-...`; `bust_unread_notification_cache` now clears both the team and all-teams keys in one call.
- Extracted `get_filtered_user_notifications()`, shared by the table view and "mark all read", so mark-all-read always acts on exactly the filtered/team-scoped set currently displayed.
- DND/mute intentionally left per-team, unchanged.

### Demo

Screenshot 2026-07-29 at 2 59 40 PM

### Docs and Changelog

Fixes #3430


apps/ocs_notifications/views.py
```diff
@@ -21,16 +22,15 @@
...
-class NotificationHome(LoginAndTeamRequiredMixin, TemplateView):
+class NotificationHome(LoginRequiredMixin, TemplateView):
     template_name = "generic/object_home.html"
 
     def get_context_data(self, **kwargs):
-        table_url = reverse("ocs_notifications:notifications_table", args=[self.request.team.slug])
+        table_url = reverse("ocs_notifications:notifications_table")
         user_preferences, _created = UserNotificationPreferences.objects.get_or_create(
             user=self.request.user, team=self.request.team
         )
```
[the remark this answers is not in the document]
snopoke: @brian-mecha this is a valid comment and points to an issue with the current design. UserPreferences are per team but this UI is now a global view so we can't use a single button to manage 'do not disturb' status.

My suggestion:

1. list teams for which the user has silenced notifications and display them as buttons which the user can press on to cancel the DND state.
2. to set DND state, add a modal which gives the user the option to select which teams or 'All Teams' and the duration.

If you want to push that change to a new PR then I think you can just remove the button and state from this view for this PR with a note.

apps/web/context_processors.py
```diff
@@ -52,9 +52,8 @@ def unread_notifications_count(request):
     Adds unread notification count to context
     """
     count = 0
-    team = getattr(request, "team", None)
-    if hasattr(request, "user") and request.user.is_authenticated and team:
-        count = request.user.unread_notifications_count(team=team)
```
snopoke, line 57: Is the `unread_notifications_count` function (and the related cache) still used anywhere?

apps/ocs_notifications/views.py
```diff
@@ -184,8 +188,8 @@ def post(self, request, team_slug: str, notification_id: int, *args, **kwargs):
...
 
 
-class ToggleDoNotDisturbView(LoginAndTeamRequiredMixin, View):
-    def post(self, request, team_slug: str, *args, **kwargs):
+class ToggleDoNotDisturbView(LoginRequiredMixin, View):
+    def post(self, request, *args, **kwargs):
         duration_param = request.POST.get("duration", None)
         user_preferences, _created = UserNotificationPreferences.objects.get_or_create(
             user=request.user, team=request.team
```
snopoke: `request.team` may not exist here now that this isn't a team view. This view will need to be modified along with the UI as mentioned.

snopoke (changes requested):

apps/web/context_processors.py
brian-mecha: Yes, the  `unread_notifications_count(team=...)` and its per-team cache are still actively used, just no longer called directly from the context processor:

- `apps/users/models.py:107`: `unread_notifications_count_all_teams()` calls `self.unread_notifications_count(team)` for every team the user belongs to and sums the results. This is by design (from proposal No. 3 in issue [#3430]): the aggregate reuses the existing per-team cached counts rather than recomputing from scratch, so the per-team method/cache is the building block, not dead code.
- Per-team cache (_get_user_notification_cache_value/set_user_notification_cache keyed by team_slug_) is written by that method and read/busted throughout `apps/ocs_notifications/views.py`, `utils.py`, and `apps/users/views.py` (_mute, toggle-read, mark-all-read, notification creation, preference updates_). All still call `bust_unread_notification_cache(team_slug=...)`, which clears both the per-team key and the all-teams key.

The per-team method/cache and the all-teams method/cache are two layers of the same caching design, not redundant code. No breaking changes.

apps/ocs_notifications/views.py
brian-mecha: Updated to reflect the later redesign of DND

brian-mecha:
@snopoke , As regards the ealirer comment on DND redesign, I have undone/refactored the affected files and added a note on the same as well for future reference when that design will be ready.

Screenshot 2026-07-30 at 7 46 48 PM

snopoke (approved):

snopoke:
@brian-mecha please create a followup issue to track the DND work.

Issue #3995, opened 2026-07-29
Evaluations: remove llm_provider_id / llm_provider_model_id from Evaluator.params


snopoke:
Part 2 of #3977. Part 1 (#3992) added the FK columns and made them authoritative; this issue removes the now-redundant copy of the ids from `Evaluator.params`.

## Background

`Evaluator.llm_provider` and `Evaluator.llm_provider_model` are real FKs and are what the runtime resolves. But `params` still carries `llm_provider_id` / `llm_provider_model_id`, because the evaluator edit form is generated from the pydantic schema on `LLMResponseMixin` — the fields have to be in that schema for the picker widget to exist at all.

So the duplication #3977 set out to remove is still there, now with a defined winner:

- `Evaluator.save()` → `sync_llm_provider_fks()` derives the FK columns from `params` on every write
- `Evaluator.get_evaluator_params()` overlays the FK values back over `params` at run time, so a stale params id is discarded

That works, but it leaves one invariant enforced only by convention: **`params` is written, the FK is derived.** Any writer that sets the FK directly, or uses `QuerySet.update()`, drifts — and `get_evaluator_params()` will prefer the drifted FK. There is no constraint, no `full_clean` hook, and no divergence-detection command (the pipeline side at least has `backfill_node_fks`).

## Proposed Change

1. Stop writing the ids into `params`. The form needs to read and write `Evaluator.llm_provider_id` / `llm_provider_model_id` instead — either by moving them out of the pydantic schema into real form fields, or by keeping them in the schema for rendering only and injecting the values from the FKs.
2. Data migration to strip both keys from existing `Evaluator.params`.
3. Delete `sync_llm_provider_fks()` and the `Evaluator.save()` override.
4. Simplify the repointing paths, which currently exist to move `params` and the FK together: `Evaluator.set_llm_provider_model_id()`, `_repoint_evaluators()` in `default_models.py`, the evaluator loop in `remove_deprecated_models`, and the params remapping in `clone_team`.
5. `EvaluatorForm._validate_llm_provider_selection` reads the ids out of the params blob today; once they arrive as real form fields, the team-scoping check can be a `ModelChoiceField` queryset and the type-compatibility check can move with it.
6. `EvaluatorFactory` can then build a valid `LlmEvaluator` directly and `configure_evaluator_llm_provider` can go.

[...]

Issue #3974, opened 2026-07-28
Evaluator params are invisible to provider delete and replace paths


snopoke:
`Evaluator.params` stores `llm_provider_id` / `llm_provider_model_id` as JSON with no FK. #3970 taught the usages page to find those references, but deliberately kept the lookup out of `_get_related_objects_querysets` so deletion behaviour stayed unchanged. That leaves evaluator references invisible to every delete and replace path.

### Consequences today

- `LlmProviderModel.delete()` and `has_related_objects()` don't see evaluators, so a provider model still used by an evaluator can be deleted. The evaluator then fails at run time in `LLMResponseMixin.get_llm_provider_model()` with `EvaluationRunException("LLM Provider Model does not exist")`.
- `_replace_custom_model_with_global()` (`apps/service_providers/llm_service/default_models.py`) repoints pipeline-node params, then deletes the custom model. Evaluator params are never repointed, so evaluators are silently left pointing at a deleted row.
- `remove_deprecated_models` has the same shape: it rewrites node params and FK fields, not evaluator params.
- `delete_service_provider` (`apps/service_providers/views.py`) only warns about experiments and assistants, so deleting an `LlmProvider` referenced by an evaluator gives no warning either.

### Why it wasn't just wired in

Order matters. If `get_related_evaluators_queryset` is added to `_get_related_objects_querysets` on its own, `LlmProviderModel.delete()` starts raising `ValidationError` inside `_replace_custom_model_with_global` — turning a silent orphan into a hard failure of the default-models sync. The repointing has to land first, or in the same change.

### Suggested shape

1. Repoint evaluator params wherever node params are already repointed — `_replace_custom_model_with_global` and `remove_deprecated_models` — mirroring `_update_pipeline_node_param`.
2. Then add `get_related_evaluators_queryset` to `_get_related_objects_querysets`, and drop the explicit call in `get_provider_usages` plus the "deliberately not part of" note on the helper.
3. Rename the `pipeline_param_key` argument on `get_related_objects` / `has_related_objects` / `_get_related_objects_querysets` to something model-agnostic (`params_key`), since it would no longer be pipeline-specific.
4. Replace `test_evaluator_params_do_not_block_provider_model_deletion` (`apps/service_providers/tests/test_models.py`) with its inverse, and add coverage for the custom-to-global replacement repointing evaluator params.

Both are row-mutating paths, so worth checking whether existing rows already hold stale evaluator params before shipping.

### Context

[...]

PR #3999, opened 2026-07-29
Extract evaluator LLM provider references into FK fields

### LINKED ISSUE #3977: Evaluations: extract llm_provider_id / llm_provider_model_id from Evaluator.params into FK fields
## Background

`Evaluator.params` is a generic `SanitizedJSONField` that stores type-specific configuration for each evaluator. For LLM-based evaluators, the `LLMResponseMixin` in `evaluators.py` defines `llm_provider_id` and `llm_provider_model_id` as Pydantic integer fields that end up serialised into this JSON blob:

```python
class LLMResponseMixin(BaseModel):
    llm_provider_id: int = Field(..., title="LLM Model", ...)
    llm_provider_model_id: int = Field(..., json_schema_extra=UiSchema(widget=Widgets.none))
    ...

    def get_llm_service(self) -> LlmService:
        provider = LlmProvider.objects.get(id=self.llm_provider_id)  # runtime lookup from JSON
        ...
```

Because these IDs live in a JSON column there is:

- **No referential integrity** — if an `LlmProvider` or `LlmProviderModel` is deleted, the evaluator silently holds a dangling integer
- **No cascading** — on-delete behaviour has to be handled manually at runtime
- **Difficult querying** — filtering evaluators by provider requires JSON field lookups
- **Inconsistency** — the rest of the codebase uses proper FK fields for provider relationships

## Proposed Change

1. Add two nullable FK fields directly on the `Evaluator` model:
   ```python
   llm_provider = models.ForeignKey(
       "service_providers.LlmProvider",
       on_delete=models.SET_NULL,
       null=True, blank=True,
       related_name="evaluators",
   )
   llm_provider_model = models.ForeignKey(
       "service_providers.LlmProviderModel",
       on_delete=models.SET_NULL,
       null=True, blank=True,
       related_name="evaluators",
   )
   ```
   Nullable so `PythonEvaluator` (and any future non-LLM evaluators) are unaffected.

2. Write a **data migration** that reads `params["llm_provider_id"]` / `params["llm_provider_model_id"]` from existing rows and populates the new FK columns.

3. Update `LLMResponseMixin.get_llm_service()` and `get_llm_provider_model()` to use the FK values passed in (or remove them from the mixin and resolve from the model directly before calling `run()`).

4. Update `Evaluator.run()` to inject the FK-resolved values when constructing the evaluator Pydantic instance rather than relying on the IDs in `params`.

5. Update the evaluator views (`_evaluator_parameter_values`, `_evaluator_default_values`, form save logic) to write to the new FK fields instead of (or in addition to) `params`.

6. Remove `llm_provider_id` / `llm_provider_model_id` from `params` once the migration and view changes are in place.

## Considerations

- `PythonEvaluator` has no LLM dependency — nullable FKs keep it clean
- `llm_provider_model_id` should be extracted in the same pass since it has exactly the same problem
- Tests in `test_llm_evaluator.py` and `test_evaluator_views.py` will need updating
- Worth checking if `llm_temperature` should stay in `params` (it is evaluator-specific config with no FK equivalent, so probably yes)

snopoke:
Closes #3977. Part 2 is #3995. Supersedes #3992, which carried this plus four follow-ons.

#3998 has merged; this now targets `main`.

### Product Description
An evaluator whose LLM provider or model has been deleted now says so — "has no LLM provider configured, edit the evaluator and select a provider and model" — instead of failing mid-run with a bare lookup error.

### Technical Description
`Evaluator.params` held `llm_provider_id`/`llm_provider_model_id` as plain integers, so deleting a provider left a dangling id (see #3974, and one such row in prod today). These become real FKs with `SET_NULL`.

`params` keeps its copy of both ids: the evaluator form is generated from the pydantic schema in `evaluators.py`, so the ids have to stay in the schema for the UI to edit them. The FK is the authoritative reference — `Evaluator.save` derives the columns from `params` on every write (dangling ids resolve to null rather than tripping the constraint) and `get_evaluator_params()` reads them back off the FKs at run time. Deriving in `save` rather than in the form is what keeps cloning, bootstrap data, and factories consistent without each remembering. Same shape as `Node._sync_resource_fk_fields`.

`get_related_evaluators_queryset` is gone — the usages page reaches evaluators through the reverse FK now.

Two consequences worth calling out, because they have no corresponding line of code:

- **Deleting an `LlmProviderModel` an evaluator uses is now blocked**, the moment the reverse FK exists — `get_candidate_relations_to_delete` picks it up and the existing guard does the rest. The flows that legitimately need the model gone (`_replace_custom_model_with_global`, `remove_deprecated_models`) repoint evaluators first.
- The delete button disappears for such a model, via `has_related_objects`.

`_replace_custom_model_with_global` also runs from migrations, where `Evaluator` may be absent from the app state entirely. It raises unless the FK column is genuinely not in the database yet, so a future migration missing the `evaluations.0018` dependency fails with instructions rather than a deferred FK violation at commit.

### Migrations

Old code ignores the new columns, so the schema change and backfill are safe to apply ahead of cutover. But an old web worker that creates or edits an evaluator after the backfill writes `params` only, leaving both FKs null — and after cutover those evaluators hit the "no LLM provider configured" error despite valid `params`, recoverable only by re-saving each one. Nothing re-syncs them automatically. Small window, small blast radius, but worth deciding deliberately: re-run the backfill after cutover, or hold this and make `params` a read-through fallback for one release.

The backfill deliberately leaves dangling ids null. Prod has exactly one such evaluator (`KMC - Coverage Eval`, team Vaccine_Coach); it was already broken at run time and still needs a human to re-pick the model — the original id can't be recovered.

### Docs and Changelog


apps/data_migrations/management/commands/remove_deprecated_models.py
```diff
@@ -89,6 +89,13 @@ def perform_migration(self, dry_run=False):
         for db_model, replacement_name, replacement_model in models_to_delete:
             db_model_id = db_model.id  # Capture before delete sets pk to None
 
+            # Evaluators keep a copy of the model id in params for the form UI, so they are
+            # repointed through the model method that moves params and FK together. Has to
+            # happen before the generic FK pass, which would otherwise claim them first.
+            new_value = replacement_model.id if replacement_model else None
+            for evaluator in db_model.evaluators.all():
+                evaluator.set_llm_provider_model_id(new_value)
+
```
snopoke: Fixed in 2d9b1002c. Extracted the reference-repointing block into `_repoint_references`, dropping `perform_migration` back below the complexity threshold; behaviour is unchanged.

*— Claude Code*

snopoke: Fixed in 2d9b1002c. The nested evaluator/FK/pipeline loops now live in `_repoint_references`, so `perform_migration` is back under the nested-blocks threshold; behaviour is unchanged.

*— Claude Code*

apps/evaluations/tests/test_evaluator_llm_provider_fks.py
```diff
@@ -0,0 +1,149 @@
...
+        evaluator = EvaluatorFactory.create(team=provider.team, params=_params(provider.id, provider_model.id))
+        other_provider = LlmProviderFactory.create(team=provider.team)
+
+        evaluator.params["llm_provider_id"] = other_provider.id
+        evaluator.save(update_fields=["params"])
+
+        evaluator.refresh_from_db()
+        assert evaluator.llm_provider_id == other_provider.id
+
+    def test_dangling_ids_become_null(self, provider, provider_model):
```
SmittieC, line 52: The reasoning behind keeping this as-is for now is that these will be removed in the followup anyways?

apps/evaluations/tests/test_evaluator_llm_provider_fks.py
```diff
@@ -0,0 +1,149 @@
...
+        """A deleted provider leaves a stale id in params; the FK must not resurrect it."""
+        evaluator = EvaluatorFactory.create(
+            team=provider.team, params=_params(provider.id + 1000, provider_model.id + 1000)
+        )
+
+        evaluator.refresh_from_db()
+        assert evaluator.llm_provider_id is None
+        assert evaluator.llm_provider_model_id is None
+
+    def test_python_evaluator_has_no_provider(self):
```
SmittieC, line 66: Might be a good idea to add a DB constraint to ensure that LLM providers have `provider_ids`?

apps/service_providers/llm_service/default_models.py
```diff
@@ -317,8 +318,54 @@ def _update_existing_global_model(existing_global_model, model):
         existing_global_model.save()
 
 
+def _evaluator_provider_model_fk_in_db() -> bool:
```
SmittieC, line 321: If my understanding is correct, this method only exists for this PR and will be removed in a followup?

apps/service_providers/tests/test_default_models.py
```diff
@@ -79,6 +85,74 @@ def test_converts_custom_models_to_global_models_pipelines():
...
+    evaluator.refresh_from_db()
+    assert evaluator.llm_provider_model_id == global_model.id
+    assert evaluator.params["llm_provider_model_id"] == global_model.id
+
+
+@pytest.mark.django_db()
+def test_repointing_raises_when_the_evaluator_relation_is_missing_but_its_fk_is_live():
+    """A migration state predating evaluations.0018 cannot see a constraint that still binds.
+
+    ``_replace_custom_model_with_global`` deletes the custom model, so skipping the repoint
```
SmittieC: I'm not following here?

SmittieC (dismissed):
LGTM. A few clarifying questions only

apps/evaluations/tests/test_evaluator_llm_provider_fks.py
snopoke: Yes. `params` has to keep the ids while the evaluator form is generated from the pydantic schema on `LLMResponseMixin` — drop them from the schema and the provider picker stops existing. So the stale id stays in `params`, the FK is what the runtime resolves, and `get_evaluator_params()` overlays the FK values back over `params` so the stale one is discarded. #3995 removes the copy from `params` altogether.

Made that explicit in the test docstring in 4d04cdf37.

*— Claude Code*

apps/evaluations/tests/test_evaluator_llm_provider_fks.py
snopoke: Deliberately not — a constraint would fight the design rather than protect it:

- The FKs are `SET_NULL`. Deleting an `LlmProvider` nulls `llm_provider_id` on every evaluator using it, so a "LLM evaluators must have a provider" check constraint would reject that write and make provider deletion fail. Letting the FK go null and surfacing it as *"has no LLM provider configured"* at run time is the point of the PR.
- Whether a provider is required depends on the evaluator type — `requires_llm_provider()` asks `issubclass(self.evaluator, LLMResponseMixin)`. A check constraint cannot ask that, so it would have to hardcode the type list and drift as evaluator types are added.
- Prod already has one row that would violate it (`KMC - Coverage Eval`, team Vaccine_Coach — see the PR description). The constraint would not apply until a human re-picks that model.

The invariant that *is* worth enforcing is the other one — `params` is written, the FK is derived, and nothing stops a writer setting the FK directly or going through `QuerySet.update()`. That is tracked in #3995, which weighs a `full_clean` hook vs. a divergence-detection command like the pipeline side has.

*— Claude Code*

apps/service_providers/llm_service/default_models.py
snopoke: No — this one outlives the PR, and it is not tied to the `params` duplication #3995 removes.

It guards a migration-graph hazard that is permanent: `_update_llm_provider_models` also runs from migrations via `migration_utils.llm_model_migration()`, and a *future* migration that calls it without declaring a dependency on `evaluations.0018` gets an app state with no `Evaluator` model — so no `evaluators` reverse accessor to repoint through — while the FK column is live in the database and its constraint still binds. Introspecting the column is how `_repoint_evaluators` tells that case apart from the genuinely safe one (column not installed yet, nothing to repoint), so it can raise naming the dependency to add instead of letting the subsequent delete fail on a deferred FK violation at commit.

Happy to drop it and rely on convention for that dependency if you would rather carry less machinery.

*— Claude Code*

apps/service_providers/tests/test_default_models.py
snopoke: Fair — that docstring was too compressed. Rewritten in 4d04cdf37. What it is testing:

A migration whose dependencies do not reach `evaluations.0018` gets an app state with no `Evaluator` model, so `_repoint_evaluators` has no `evaluators` accessor to repoint through. The FK column is in the database regardless, and its constraint still binds. Carrying on would mean `_replace_custom_model_with_global` deletes the custom model out from under those evaluators, and Postgres only notices the orphans when the deferred constraint is checked at commit — an opaque failure a long way from its cause. So it raises here instead, naming the dependency to add.

*— Claude Code*

SmittieC (approved):

apps/service_providers/tests/test_default_models.py
snopoke: This (and the weirdness in `default_models.py`) is because of how Django does migrations. The model class that get's presented to the migration command is a generated version based on the schema up to this point in the migrations. Django iteratively build the model as the migrations are processed.

So this code is supposed to make migrations fail if we're trying to operate on old versions of the model.

This should be temporary.

PR #4001, opened 2026-07-29
Keep evaluators visible when an LLM provider or model goes away

snopoke:
### Product Description
Deleting an LLM provider an evaluator still uses is now blocked, with the evaluator listed in the confirmation modal. Teams are also told when a model deprecation or removal affects an evaluator, which previously went unmentioned.

### Technical Description
Two gaps that only became visible once evaluators reference providers by FK:

- `delete_service_provider` already collected evaluators via `get_related_objects`, but only turned `Experiment` and `OpenAiAssistant` into blocking chips and dropped the rest — so the delete went through, leaving the evaluator with a nulled FK and a stale id in `params`. `render_referenced_objects_modal` already accepted `evaluators`; it just wasn't passed.
- `get_affected_teams_data` reported chatbots, pipelines and assistants only. It gains an `evaluators` bucket, and both `deprecated_model_notification` and `deleted_model_notification` now take the four buckets as one required `AffectedResources`, so a new kind of resource has to be threaded through deliberately. The deprecation warning is the more useful of the two — it arrives while the model still works.

The collector must run before anything repoints evaluators, or the FK has already moved; that ordering is documented and covered by a test in `remove_deprecated_models`.

### Migrations
None.

### Docs and Changelog


apps/ocs_notifications/notifications.py
```diff
@@ -216,16 +216,15 @@ def deprecated_model_notification(
     affected_chatbots: dict[str, str],
     affected_pipelines: dict[str, str],
     affected_assistants: dict[str, str],
+    affected_evaluators: dict[str, str],
```
snopoke: Fixed in e3cbb6434. The four `affected_*` dicts are now one `AffectedResources` value object, so the signature is back to 4 arguments. It also removes the duplication that came with them — the four-tuple summary list and the four-way `links` merge were each written out twice.

*— Claude Code*

apps/ocs_notifications/notifications.py
```diff
@@ -343,16 +348,15 @@ def deleted_model_notification(
...
```
snopoke: Fixed in e3cbb6434, same change as the sibling `deprecated_model_notification` — both now take a single `AffectedResources`.

*— Claude Code*

apps/ocs_notifications/tests/test_notifications.py
```diff
@@ -402,6 +402,7 @@ def test_with_replacement(self, mock_create_notification):
             affected_chatbots={"My Bot": "/chatbots/my-bot/"},
             affected_pipelines={},
             affected_assistants={},
+            affected_evaluators={},
```
snopoke: Partly addressed in e3cbb6434. The four flagged tests each ended with four `affected_*={}` lines; they now build an `AffectedResources` through a local `_affected()` helper that defaults the kinds a test does not care about, which removes most of the shared bulk.

Not parametrizing them any further, though. They cover two different functions with different titles, slugs and message wording — folding all four into one table would hide what each one actually asserts, which is the part worth reading.

*— Claude Code*

apps/service_providers/views.py
```diff
@@ -159,11 +160,18 @@ def delete_service_provider(request, team_slug: str, provider_type: str, pk: int
...
             for assistant in [obj for obj in filtered_objects if isinstance(obj, OpenAiAssistant)]
         ]
-        if related_experiments or related_assistants:
+        # Evaluators reference the provider by FK. Deleting underneath one leaves it
+        # unrunnable — nulled FK, stale id in params — so it blocks like the rest.
+        related_evaluators = [
+            Chip(label=evaluator.name, url=evaluator.get_absolute_url())
+            for evaluator in [obj for obj in filtered_objects if isinstance(obj, Evaluator)]
+        ]
+        if related_experiments or related_assistants or related_evaluators:
```
snopoke: Fixed in 763656a1e. The three chip lists are now a `chips_by_kind` dict guarded by `any(...)` and splatted into `render_referenced_objects_modal`, so the three-branch conditional is gone and the repeated filter-and-map is stated once per kind. The experiment label logic moved into `_experiment_chip_label`.

*— Claude Code*

apps/ocs_notifications/tests/test_notifications.py
```diff
@@ -426,9 +435,10 @@ def test_without_replacement(self, mock_create_notification):
             team=team,
             model_name="gpt-4",
             replacement_model_name=None,
-            affected_chatbots={},
-            affected_pipelines={"My Pipeline": "/pipelines/1/"},
-            affected_assistants={"My Assistant": "/assistants/1/"},
+            affected=_affected(
+                pipelines={"My Pipeline": "/pipelines/1/"},
+                assistants={"My Assistant": "/assistants/1/"},
+            ),
```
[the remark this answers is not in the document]
snopoke: Declined — down from 4 similar functions to 3 after 76f04baae, and the remainder is deliberate.

These cover two different functions (`deprecated_model_notification` and `deleted_model_notification`) whose titles, slugs and message wording all differ. What they share is the shape every notification test here has: build a team, call the function, assert on the `create_notification` kwargs. Collapsing them into one parametrized table would move the expected title and message out of the test body and into a fixture argument, which hides the thing each test exists to pin down.

The incidental duplication — the four `affected_*={}` arguments — is already gone via the `_affected()` helper.

*— Claude Code*

apps/ocs_notifications/notifications.py
```diff
@@ -208,27 +209,50 @@ def tool_error_notification(team, tool_name: str, error_message: str, session=No
...
+    evaluators: dict[str, str]
+
+    def summary_text(self) -> str:
+        """Summarise the non-empty kinds, e.g. `2 chatbot(s), 1 evaluator(s)`."""
+        parts = [f"{len(items)} {noun}(s)" for noun, items in self._by_noun() if items]
+        return ", ".join(parts) if parts else "some resources"
+
+    def links(self) -> dict[str, str]:
+        """Every affected resource as a single `{name: url}` map for the notification links."""
+        return {**self.chatbots, **self.pipelines, **self.assistants, **self.evaluators}
```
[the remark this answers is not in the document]
snopoke: Fixed in a0622b095. Confirmed real: reverting the fix leaves `links()` returning `{'Support': '/evaluators/7/'}` for a chatbot and an evaluator both named "Support" — the chatbot URL is dropped while `summary_text()` still counts both.

Worth noting the merge itself is unchanged from `main`; what this PR did was add the fourth (`evaluators`) bucket, which is what makes the collision reachable.

Qualified only the colliding names (`Support (chatbot)` / `Support (evaluator)`) rather than labelling every link by kind as suggested — names are unique within a kind, so unconditional prefixing would churn every existing notification's link labels to fix a case that usually is not present. Collision coverage added in `TestAffectedResourcesLinks`.

*— Claude Code*

apps/data_migrations/tests/test_notify_deprecated_models.py
```diff
@@ -58,7 +58,7 @@ def test_notifies_affected_teams(self, mock_notify):
         assert call_kwargs["team"] == experiment.team
         assert call_kwargs["model_name"] == "openai/test-deprecated-model"
         assert call_kwargs["replacement_model_name"] == "test-replacement-model"
-        assert experiment.name in call_kwargs["affected_chatbots"]
+        assert experiment.name in call_kwargs["affected"].chatbots
```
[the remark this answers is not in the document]
snopoke: Fixed in 4460ceae1. Added `test_notifies_the_team_about_affected_evaluators` covering the evaluator bucket at the command level.

Added it as its own test rather than extending the two existing ones: `EvaluatorFactory` builds its own team, so adding an evaluator alongside the pipeline-owning experiment would make the command notify two teams and break their `assert_called_once`.

*— Claude Code*

PR #4014, opened 2026-07-30
Fix OOM restart loop in evaluation coordination

snopoke:
### Product Description

Evaluation runs stopped progressing. Fixes that, and the celery worker restart loop behind it.

### Technical Description

Found while debugging prod: celery workers were being OOM-killed (exit 137) every ~30s. `coordinate_evaluation_runs` was in flight in 12/12 sampled kills, and 592 of 1008 invocations that day never completed.

One FULL run over a 3548-message dataset had a result for every planned message, so each tick computed `remaining` as empty and re-entered the completion path. `_finalize_complete` did `mark_complete()` **and** both completion sweeps inside the tick's `transaction.atomic()`, so the OOM kill those sweeps caused rolled the COMPLETED status back — leaving a finished run at PROCESSING for the next tick to redo, forever.

Because the sweep drove runs inline in `created_at` order and this was the oldest, nothing behind it was ever driven: 25 runs sat PENDING for hours with stale Taskbadger tasks.

Three changes:

- **A terminal transition commits on its own.** The tick now commits only the status; `finalize_evaluation_run` carries aggregates and tag reversal after commit. Idempotent, so a retry after a partial pass converges. A failure there now costs the run its aggregates instead of costing the worker its life.
- **Per-run ticks.** `coordinate_evaluation_runs` fans out `drive_evaluation_run` per run. The old inline loop's `try`/`except` could not catch a SIGKILL, so one run's death was every later run's death. `skip_locked` makes an overlapping tick a cheap no-op.
- **Both sweeps stream** (`.iterator()`) and retain only what they aggregate, instead of materialising every result dict and the whole dataset with its `session__chat` joins.

Worth a look from a reviewer:

- Both new tasks are added to the Taskbadger `excludes`. Without that, `drive_evaluation_run` would create a TB task per run per 30s.
- Known gap, documented in the ADR rather than silently accepted: a crash between the tick's commit and `finalize_evaluation_run.delay` leaves a COMPLETED run unfinalized with nothing to repair it. The dispatch is deliberately last after commit, behind the progress publish and the Taskbadger update, so the window holds a blocking HTTP call rather than being a bare `.delay()`. Closing it needs a `finalized_at` marker — "COMPLETED with no aggregates" is legitimate for an all-errors run, so it can't be the repair signal. That marker is added in #4020, stacked on this branch, which also fixes a related regression review caught here: the results page could render a COMPLETED run as having no aggregates before finalization ran.
- ADR-0047 amended, since it documented the inline sweep and `_finalize_complete`.
- Unrelated to this bug but still true: the worker runs `--pool=threads --concurrency 10` in a 2 GB container, i.e. ten tasks sharing 2 GB. Thin for LLM-adjacent work; suggest bumping it in ocs-deploy separately.

### Docs and Changelog


apps/evaluations/tasks.py
```diff
@@ -367,11 +368,17 @@ def _fail_run_for_unconfigured_evaluators(run: EvaluationRun, names: list[str])
...
+def _complete_run(run: EvaluationRun) -> None:
+    """Flip the run to COMPLETED. The completion side effects run after the tick commits.
+
+    Only the status transition belongs under the coordination lock. Aggregates and tag
+    reversal both sweep the whole run, so running them here held the row lock for the
+    length of that sweep and — because the transition shares this transaction — a crash
+    part-way through rolled the run back to PROCESSING. The next tick then saw the same
+    finished run and re-ran the same doomed sweep, once per beat interval, forever.
+    `finalize_evaluation_run` carries them instead (dispatched from `_drive_run`).
+    """
```
SmittieC, line 380: I'm only noticing now that we should perhaps only call `run.mark_complete()` when the status is in a non terminal state

apps/evaluations/tasks.py
```diff
@@ -440,31 +447,61 @@ def _drive_run(run_id: int) -> None:
     with current_team(run.team):
         for batch in result.batches:
             evaluate_message_batch.delay(run.id, batch)
+        if result.terminal == "success":
```
SmittieC: if we extend this check to `result.terminal == "success" and result.done > 0` then we don't queue a task that is doomed to waste resources by trying to compute aggregates for non-existing results

apps/evaluations/tasks.py
```diff
@@ -440,31 +447,61 @@ def _drive_run(run_id: int) -> None:
...
+def finalize_evaluation_run(run_id: int) -> None:
+    """Completion side effects for a run already committed as COMPLETED.
+
+    Split out of the coordination tick because both steps sweep the whole run and the
+    tick holds a row lock (see `_complete_run`). Safe to re-run: aggregates are
+    update_or_create'd per evaluator and stale-tag removal is a delete, so a retry after
+    a partial pass converges. A permanent failure here costs the run its aggregates and
+    leaves stale tags in place — it no longer costs the worker its life.
+    """
+    run = EvaluationRun.objects.filter(id=run_id).select_related("team", "config__dataset").first()
```
SmittieC, line 470: thoughts on adding an extra (safety) clause to filter on status as well so that we don't accidentally compute aggregates for incomplete runs?

SmittieC (dismissed):
Nothing blocking (although coderabbit's comment seems legit). Thanks for fixing these

apps/evaluations/tasks.py
SmittieC: Done in ee38c5fd1 — the dispatch is now gated on `result.terminal == "success" and result.done > 0`.

Worth noting it saves more than the wasted aggregate query. The only success path with `done == 0` is the `total == 0 or not evaluator_ids` branch, and `reverse_stale_tags` walks `run.config.dataset.messages` off the **live** config rather than the run’s frozen `evaluator_ids` snapshot. So in the no-evaluators case it would sweep the whole dataset with an empty `applied_by_message` and treat every managed tag as stale. Reaching that needs evaluators removed from the config and re-added inside one 30s beat, so it was near-unreachable rather than a live bug, but the gate closes it for free.

`done > 0` is also the right discriminator rather than `total > 0` — `total` can be non-zero in that branch. Covered by `test_completion_with_nothing_evaluated_skips_finalization`.

apps/evaluations/tasks.py
SmittieC: Added in ee38c5fd1. Kept it as a separate branch from the `run is None` check rather than folding it into the filter, since "run gone" and "run in an unexpected state" are different situations and the second should log louder:

```python
if run.status != EvaluationRunStatus.COMPLETED:
    logger.warning("Evaluation run %s is %s, not completed; skipping finalization", run_id, run.status)
    return
```

Nothing is lost by gating on COMPLETED specifically — the results UI already filters aggregates on `run__status=COMPLETED`, so a FAILED run’s partial aggregates would never surface anyway. Covered by `test_finalization_is_a_noop_for_a_run_that_is_not_completed`, parametrized over pending/processing/failed.

apps/evaluations/tasks.py
SmittieC: I looked at this and I think it’s unreachable as written, so I have left the code alone — happy to add it if you disagree.

`_complete_run` has exactly two callers, both inside `_coordinate_locked_run`, which is only reachable from `_drive_run`. That entry query already does the filtering:

```python
EvaluationRun.objects.select_for_update(skip_locked=True)
    .filter(id=run_id, status__in=NON_TERMINAL_RUN_STATUSES)
```

and the row lock is held for the whole tick, so nothing can flip the status underneath it. A guard here would be unreachable, and it would also quietly swallow the bug if that invariant ever broke rather than letting it surface.

If we do want the protection, I would put it in `mark_complete`/`mark_failed` themselves instead. That covers every caller — including the views and management commands — and it guards the thing an accidental re-terminate actually damages, which is `finished_at` getting overwritten. Say the word and I will move it there.

apps/evaluations/tasks.py
```diff
@@ -440,31 +447,61 @@ def _drive_run(run_id: int) -> None:
     with current_team(run.team):
         for batch in result.batches:
             evaluate_message_batch.delay(run.id, batch)
+        if result.terminal == "success":
+            finalize_evaluation_run.delay(run.id)
         _ensure_taskbadger_task(run, result.total)
         _publish_tick(run, result)
```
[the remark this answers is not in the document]
SmittieC: Valid — this was the one genuinely non-self-healing call in that block. Everything else after the transaction is already internally guarded (`_publish_progress` has its own try/except, and the Taskbadger calls are the `*_safe` variants), so the bare `.delay()` was the only unprotected one, and because the run is terminal by then no later tick retries the publish.

Fixed in ee38c5fd1, but by moving the dispatch below `_publish_tick` rather than wrapping it. Same isolation, no extra nesting, and the exception still surfaces via `drive_evaluation_run`’s `logger.exception` instead of being logged and dropped locally. `_ensure_taskbadger_task` has to stay ahead of the publish since `_update_taskbadger` needs the task id.

For the record, `evaluate_message_batch.delay` in the loop above has the same exposure but does not need the same treatment: the run stays non-terminal there, so the next tick re-dispatches. Covered by `test_failed_finalize_dispatch_still_publishes_completion`.

docs/adr/0047-deploy-safe-evaluation-runs.md
```diff
@@ -30,6 +31,21 @@ We will replace chord dispatch with a stateless beat coordinator that derives al
...
+- **One run's tick cannot be another run's fate.** Per-run `drive_evaluation_run` tasks replace the inline loop, because the inline loop's `try`/`except` could not catch an OOM kill — the signal takes the whole sweep with it.
+
+Both completion sweeps also now stream (`.iterator()`) and retain only what they aggregate, so their memory is bounded by distinct result fields rather than by run size.
+
+The trade this accepts: a crash in the window between the tick's commit and `finalize_evaluation_run.delay` leaves a COMPLETED run unfinalized for good, where the old coupling would have retried it. The dispatch is the first thing after commit and a completing tick has no batches to dispatch ahead of it, so the window is small — but it is not zero, and no sweep repairs it. Runs with only error results legitimately produce no aggregates, so "COMPLETED with no aggregates" cannot be used as the repair signal; a `finalized_at` marker would be needed to close this properly.
```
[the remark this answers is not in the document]
snopoke: Correct on all three counts — fixed in 2ff9c8220 (ADR prose only; the code ordering is intentional and stays).

Verified against the code: after the `transaction.atomic()` block `_drive_run` runs the batch loop, then `_ensure_taskbadger_task`, then `_publish_tick`, and only then `finalize_evaluation_run.delay`. On a completing tick `_publish_tick` takes the `terminal == "success"` branch, so the window does contain a result-backend `store_result` plus `_update_taskbadger(..., status=SUCCESS)` → `taskbadger.update_task_safe`, which is a blocking HTTP PATCH (it swallows errors, but not the round trip). And line 19 of the same file already stated the opposite ordering.

I took your suggestion with light trimming, since the Decision bullet already carries the "why last" reasoning and this paragraph only needs to point at it:

> The dispatch is deliberately *last* after commit, behind the progress publish and the Taskbadger update (see the decision above), so the window holds a result-backend write and a blocking Taskbadger HTTP call — wider than a bare `.delay()`, and no sweep repairs a run stranded in it.

Dropped "a completing tick has no batches to dispatch ahead of it" — true, but it only existed to prop up the "window is small" claim. The `finalized_at` sentence is unchanged.

apps/evaluations/aggregation.py
```diff
@@ -3,42 +3,57 @@
...
+    for field_name, value in result.items():
+        if value is not None and get_aggregators_for_value(value):
+            field_values[field_name].append(value)
+
+
+def _aggregate_fields(field_values: defaultdict[str, list]) -> dict:
+    return {field_name: aggregate_field(values) for field_name, values in field_values.items()}
+
+
 def compute_evaluator_aggregates(results: list[dict]) -> dict:
```
snopoke: I think this function is now unused

apps/evaluations/tagging.py
```diff
@@ -146,8 +149,9 @@ def _get_possible_tags(evaluators: list[Evaluator]) -> frozenset[int]:
 def _build_applied_by_message(run: EvaluationRun) -> defaultdict[int, set[int]]:
     """Batch all AppliedTag lookups for this run to avoid an O(N) query per message."""
     applied: defaultdict[int, set[int]] = defaultdict(set)
-    for row in AppliedTag.objects.filter(evaluation_result__run=run).values("evaluation_result__message_id", "tag_id"):
-        applied[row["evaluation_result__message_id"]].add(row["tag_id"])
+    rows = AppliedTag.objects.filter(evaluation_result__run=run).values_list("evaluation_result__message_id", "tag_id")
+    for message_id, tag_id in rows.iterator(chunk_size=TAG_CHUNK_SIZE):
+        applied[message_id].add(tag_id)
```
snopoke, line 154: Note for self: this changes two things:

1. `values` -> `values_list`: reduces memory by using a tuple instead of a dict
2. `iterator`: using the iterator streams data from PG (when `DISABLE_SERVER_SIDE_CURSORS = False` which it is) and doesn't cache the results on python so although we still build the full dataset in memory in the loop we don't pay the penalty of also having the full result set in the query cache.

apps/evaluations/tests/coordination.py
```diff
@@ -0,0 +1,19 @@
+"""Test helper for driving evaluation coordination synchronously."""
+
+from unittest.mock import patch
+
+from apps.evaluations.models import NON_TERMINAL_RUN_STATUSES, EvaluationRun
+from apps.evaluations.tasks import drive_evaluation_run, finalize_evaluation_run
+
+
+def sweep():
+    """Tick every active run synchronously, as a worker would.
```
snopoke, line 10: why not just use celery `always_eager`?

apps/evaluations/tests/coordination.py
```diff
@@ -0,0 +1,19 @@
...
+
+    `coordinate_evaluation_runs` only fans out; the per-run tick and the completion side
+    effects are separate tasks (`drive_evaluation_run`, `finalize_evaluation_run`), so both
+    are run inline here to leave the same end state a worker would reach.
+    """
+    run_ids = list(EvaluationRun.objects.filter(status__in=NON_TERMINAL_RUN_STATUSES).values_list("id", flat=True))
+    with patch("apps.evaluations.tasks.finalize_evaluation_run.delay", side_effect=finalize_evaluation_run):
```
snopoke, line 17: should this also patch `evaluate_message_batch.delay`?

apps/evaluations/tasks.py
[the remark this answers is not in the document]
snopoke: Fixed in ee38c5fd1, by reordering rather than by catching. The dispatch now runs *last*, after `_ensure_taskbadger_task` and `_publish_tick`, so a broker error there can no longer skip either. Same outcome as the suggested try/except, without a second error path to reason about — and `test_failed_finalize_dispatch_still_publishes_completion` pins it.

*— Claude Code*

apps/evaluations/tasks.py
snopoke: Not applicable — it already is. `_drive_run` takes the row with `select_for_update(skip_locked=True).filter(id=run_id, status__in=NON_TERMINAL_RUN_STATUSES)`, so `_coordinate_locked_run` (the only caller of `_complete_run`) is unreachable for a terminal run. An extra check here would be dead code.

*— Claude Code*

apps/evaluations/tasks.py
snopoke: Fixed in ee38c5fd1 — the guard is now `result.terminal == "success" and result.done > 0`. Worth more than the wasted sweep: `reverse_stale_tags` reads the live config rather than the run's frozen evaluator snapshot, so finalizing the no-evaluators branch would have walked the whole dataset with an empty applied-tag map and treated every managed tag as stale. Covered by `test_completion_with_nothing_evaluated_skips_finalization`.

*— Claude Code*

apps/evaluations/tasks.py
snopoke: Fixed in ee38c5fd1 — `finalize_evaluation_run` now re-checks `status == COMPLETED` and logs a warning otherwise, so a stray dispatch cannot aggregate a partial result set. Parametrized over PENDING/PROCESSING/FAILED in `test_finalization_is_a_noop_for_a_run_that_is_not_completed`.

*— Claude Code*

docs/adr/0047-deploy-safe-evaluation-runs.md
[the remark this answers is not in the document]
snopoke: Fixed in 2ff9c8220. The paragraph now states the dispatch is deliberately last, names what sits in the window (a result-backend write plus a blocking Taskbadger call), and gives the dispatch-error trade-off as the justification instead of "the window is small".

*— Claude Code*

apps/evaluations/aggregation.py
snopoke: Fixed in 4a8ec1e97 — deleted. Zero callers repo-wide once `compute_aggregates_for_run` was rewritten to stream and accumulate directly.

*— Claude Code*

apps/evaluations/tagging.py
snopoke: Both correct. One caveat on the second point: `applied` still retains every `(message_id, tag_id)` pair, so this function is not memory-bounded the way `_compute_stale_by_target` now is — what the iterator removes is the redundant second copy of data that is about to be aggregated anyway. If this path ever OOMs again, that dict is the thing to attack.

*— Claude Code*

apps/evaluations/tasks.py
```diff
@@ -442,29 +449,68 @@ def _drive_run(run_id: int) -> None:
             evaluate_message_batch.delay(run.id, batch)
         _ensure_taskbadger_task(run, result.total)
         _publish_tick(run, result)
+        # Dispatched last on purpose. The run is terminal by now, so no later tick repeats
+        # this block — a broker error here would otherwise cost the run its completion
+        # signal (stopping the UI poll) as well as its aggregates. `done == 0` means there
+        # is nothing to aggregate and no applied tags to reverse against.
+        if result.terminal == "success" and result.done > 0:
+            finalize_evaluation_run.delay(run.id)
```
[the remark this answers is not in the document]
snopoke: Confirmed and fixed in #4020 (stacked on this branch). Verified the chain against the dev server before fixing: the reloaded page really did render the completed run with no Aggregates section and nothing re-fetching it.

Took the `finalized_at` route rather than re-publishing or holding the `stop=True` publish. Re-publishing does not help — the reloaded page has no poller left to hear it. Holding the publish narrows the window but does not close it (a manual refresh during finalization lands in the same state) and trades it for a poller stuck on a lost finalization. `finalized_at` also resolves the all-errors ambiguity you flagged: the aggregates block polls itself while the run is unfinalized, bounded by a grace window since nothing retries a lost finalization.

*— Claude Code*

apps/evaluations/aggregation.py
```diff
@@ -3,42 +3,57 @@
...
-    field_values = defaultdict(list)
+    field_values: defaultdict[str, list] = defaultdict(list)
     for result in results:
-        for field_name, value in result.items():
-            if value is not None and get_aggregators_for_value(value):
-                field_values[field_name].append(value)
+        _collect_aggregatable_values(result, field_values)
 
-    return {field_name: aggregate_field(values) for field_name, values in field_values.items()}
+    return _aggregate_fields(field_values)
```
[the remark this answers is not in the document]
snopoke: Fixed in 4a8ec1e97 — deleted. Confirmed the same zero-caller result independently.

*— Claude Code*

apps/evaluations/aggregation.py
snopoke: Fixed in 4a8ec1e97 — deleted. Zero callers repo-wide once `compute_aggregates_for_run` was rewritten to stream and accumulate directly.

*— Claude Code*

apps/evaluations/tasks.py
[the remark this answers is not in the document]
snopoke: Fixed in ee38c5fd1, by reordering rather than by catching. The dispatch now runs *last*, after `_ensure_taskbadger_task` and `_publish_tick`, so a broker error there can no longer skip either. Same outcome as the suggested try/except, without a second error path to reason about — and `test_failed_finalize_dispatch_still_publishes_completion` pins it.

*— Claude Code*

apps/evaluations/tasks.py
snopoke: Not applicable — it already is. `_drive_run` takes the row with `select_for_update(skip_locked=True).filter(id=run_id, status__in=NON_TERMINAL_RUN_STATUSES)`, so `_coordinate_locked_run` (the only caller of `_complete_run`) is unreachable for a terminal run. An extra check here would be dead code.

*— Claude Code*

apps/evaluations/tasks.py
snopoke: Fixed in ee38c5fd1 — the guard is now `result.terminal == "success" and result.done > 0`. Worth more than the wasted sweep: `reverse_stale_tags` reads the live config rather than the frozen evaluator snapshot, so finalizing the no-evaluators branch would have walked the whole dataset with an empty applied-tag map and treated every managed tag as stale. Covered by `test_completion_with_nothing_evaluated_skips_finalization`.

*— Claude Code*

apps/evaluations/tasks.py
snopoke: Fixed in ee38c5fd1 — `finalize_evaluation_run` now re-checks `status == COMPLETED` and logs a warning otherwise, so a stray dispatch cannot aggregate a partial result set. Parametrized over PENDING/PROCESSING/FAILED in `test_finalization_is_a_noop_for_a_run_that_is_not_completed`.

*— Claude Code*

docs/adr/0047-deploy-safe-evaluation-runs.md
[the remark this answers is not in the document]
snopoke: Fixed in 2ff9c8220. The paragraph now states the dispatch is deliberately last, names what sits in the window (a result-backend write plus a blocking Taskbadger call), and gives the dispatch-error trade-off as the justification instead of "the window is small".

*— Claude Code*

apps/evaluations/tagging.py
snopoke: Both correct. One caveat on the second point: `applied` still retains every `(message_id, tag_id)` pair, so this function is not memory-bounded the way `_compute_stale_by_target` now is — what the iterator removes is the redundant second copy of data that is about to be aggregated anyway. If this path ever OOMs again, that dict is the thing to attack.

*— Claude Code*

apps/evaluations/tasks.py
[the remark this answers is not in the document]
snopoke: Confirmed and fixed in #4020 (stacked on this branch). Verified the chain against the dev server before fixing: the reloaded page really did render the completed run with no Aggregates section and nothing re-fetching it.

Took the `finalized_at` route rather than re-publishing or holding the `stop=True` publish. Re-publishing does not help — the reloaded page has no poller left to hear it. Holding the publish narrows the window but does not close it (a manual refresh during finalization lands in the same state) and trades it for a poller stuck on a lost finalization. `finalized_at` also resolves the all-errors ambiguity you flagged: the aggregates block polls itself while the run is unfinalized, bounded by a grace window since nothing retries a lost finalization.

*— Claude Code*

apps/evaluations/aggregation.py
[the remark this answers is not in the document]
snopoke: Fixed in 4a8ec1e97 — deleted. Confirmed the same zero-caller result independently.

*— Claude Code*

apps/evaluations/tests/coordination.py
snopoke: `task_always_eager` is not enabled anywhere in this project's test settings, so it would have to be switched on inside the helper — and it is global: every `.delay()` in the call tree goes inline, including `evaluate_message_batch`, which would run real evaluator code in any test that did not patch it. It also swallows exceptions unless `task_eager_propagates` is set too, which is the wrong default for a helper meant to leave the same end state a worker would. The targeted patch makes it explicit that exactly one task runs inline.

*— Claude Code*

apps/evaluations/tests/coordination.py
snopoke: Deliberately not: all 15 `sweep()` call sites already patch `evaluate_message_batch.delay` themselves because they assert on it (dispatch counts, batch contents), and a mock installed inside `sweep()` would not be reachable for those assertions. The split is "the test owns the mock it asserts on; `sweep()` owns the one nothing asserts on". Worth knowing the hazard, though: a future test that calls `sweep()` down a dispatching path without patching would queue to the real broker and silently pass.

*— Claude Code*

SmittieC (approved):

Issue #3986, opened 2026-07-28
Trace model: remove n_total_tokens / n_prompt_tokens / n_completion_tokens fields


ocs-percy:
Follow-up to #3984.

Once #3984 lands, the only remaining readers of `Trace.n_total_tokens`, `n_prompt_tokens`, and `n_completion_tokens` are:

- `templates/trace/trace_detail.html` — displays per-trace token counts and a prompt/completion bar
- `apps/service_providers/tracing/ocs_tracer.py` — writes the fields after each pipeline run
- `apps/web/management/commands/bootstrap_data.py` — seeds them for local dev

`UsageRecord` already carries a `trace` FK (`null=True, on_delete=SET_NULL`) and stores input/output tokens per `(provider, model, service_kind)` bucket, so the same data is available there — with more granularity — and is already the canonical source after #3984.

`n_turns` and `n_toolcalls` are **out of scope**: they are trace-level concepts with no UsageRecord equivalent (#3984 explicitly keeps them).

## Scope

**`apps/trace/models.py`**
- Remove `n_total_tokens`, `n_prompt_tokens`, `n_completion_tokens` fields
- Add migration to drop the three columns

**`apps/service_providers/tracing/ocs_tracer.py`**
- Stop writing the three fields in the post-run save
- `TraceMetrics` / `MetricsCollector.get_metrics()` can keep computing these values for intermediate use (e.g. the `_sum_token_counts` path), but they no longer need to persist on the Trace row; consider removing them from `TraceMetrics` entirely if nothing else consumes them

**`templates/trace/trace_detail.html`**
- Replace `trace.n_total_tokens / n_prompt_tokens / n_completion_tokens` with an aggregation over `UsageRecord` rows for the same trace
- Suggested query in the view: `UsageRecord.objects.filter(trace=trace).values("service_kind").annotate(tokens=Sum("quantity"))` — split by `ServiceKind.LLM_INPUT` / `LLM_OUTPUT`; include `LLM_CACHED_INPUT` and `LLM_CACHE_WRITE` in the input total to match the old headline figure
- The breakdown will now be richer (per-model rows) so consider exposing that rather than just replicating the old bar

**`apps/web/management/commands/bootstrap_data.py`**
- Remove the three kwargs from the `Trace(...)` constructor call

**Tests**
- `apps/service_providers/tracing/tests/test_ocs_tracer_metrics.py` — remove assertions on `trace.n_total_tokens`
[...]

Issue #3984, opened 2026-07-28
Consolidate token reporting on UsageRecord instead of Trace


snopoke:
Follow-up from #3978 (ADR-0048).

The cross-team usage report reads tokens and cost from two different tables, so the two halves of one report cannot be reconciled:

- `get_usage_data` (`apps/admin/queries.py:77`) and `get_token_usage_by_team` (`:95`) sum `Trace.n_total_tokens`.
- `get_cost_usage_by_team` sums `UsageRecord.quantity`.

Judge (evaluator) LLM calls bypass the tracer entirely, so they produce a `UsageRecord` and no `Trace` at all — their tokens land in the cost half and nowhere in the token half. #3978 closed the opposite-direction gap (the token half used to exclude the evaluations platform while the cost half counted it), but that one remains.

`UsageRecord` is strictly the richer source: both are populated from the same `MetricsCollector`, but `UsageRecord` breaks down by model, provider, service kind, confidence and now `source`, and it covers writers that have no trace.

### Scope

Only those two queries aggregate trace tokens for reporting. `templates/trace/trace_detail.html` shows one trace's own token counts, which is legitimately per-trace and should stay as-is.

### Things to work out

- **`run_count` cannot move.** It is `Count("id")` over traces — a "run" is a trace concept. `UsageRecord.trace_id` is null for judge rows, so `Count("trace_id", distinct=True)` would undercount activity. Either keep the run count trace-sourced (and document that the two columns have different denominators) or redefine what a run is.
  - Drop `run_count` entirely.
- **The totals will not match exactly.** `Trace.n_total_tokens` is the summed `usage_metadata.total_tokens`; the `UsageRecord` equivalent is a sum over service-kind buckets, where the headline input has cache-read/cache-write subtracted out into their own kinds (`_split_buckets`). Verify how far apart they actually land before switching.
- **Historical discontinuity.** `UsageRecord` rows only exist from the cost-tracking rollout, while `Trace` goes back further, so pre-rollout periods would report zero tokens. There is precedent for this kind of step in the same docstring (the earlier move off the character-based proxy), but it needs a deliberate call.
  - no special action required
- Whether the reporting API's response shape can change, given `PROVIDER_REPORTING_API_TOKEN` implies an external consumer.
  - this can change shape

PR #4015, opened 2026-07-30
Read trace token counts from UsageRecord

snopoke:
Phase 1 of #3986. `UsageRecord` has been the canonical token source since #3984, so the three `Trace` counters are now redundant — and worse than the replacement: they only ever summed usage the provider reported, while `UsageRecord` also carries estimated calls and splits by (provider, model).

Split into two releases on purpose. This PR removes every reader and writer but leaves the columns in place; the `RemoveField` migration lands in a following release (#4016). Dropping them together would mean old workers hitting `Trace.objects.create()` on missing columns during the rolling deploy — that call isn't wrapped, so chat requests would error for the deploy window.

Worth a closer look:

- **The in/out definition on the card deliberately differs from the usage API's `TokenCounts`.** `trace_token_usage` folds `LLM_CACHE_WRITE` into the input side so `in + out == total` and the headline reproduces what the provider reported for the call. `TokenCounts` keeps cache-write in `total` only, which is right for that API but would make the card's bar not add up.
- **`MetricsCollector` no longer subclasses `UsageMetadataCallbackHandler`.** Its `usage_metadata` dict existed solely to feed the trace counters; the cost path keeps its own `(provider, model)` buckets and is unaffected.
- **Phase 2 also needs `api-schemas/export.yml` regenerated** — the fields are exposed through the auto-generated `TraceDetail` export serializer. The importer reads `model._meta.concrete_fields` and ignores extra keys, so older export payloads still import after the drop.

### Migrations
N/A — no schema change in this PR.

### Demo
Verified against seeded dev data (`bootstrap_data` now links usage rows to their trace, so the card isn't blank locally):

The card keeps its previous shape and gains a per-model row underneath, with cached / cache-write tokens called out where present. Traces with no recorded usage render the em dash rather than a zero.

tokens-card

### Docs and Changelog

### DIFF
--- a/apps/web/management/commands/bootstrap_data.py
+++ b/apps/web/management/commands/bootstrap_data.py
@@ -408,33 +408,7 @@ class Command(BaseCommand):
         priced_model = priced_rule.model_name
         now = timezone.now()
 
-        created = 0
-        # Priced, exact spend: each session bills a little every day, with a
-        # per-session multiplier so bots differ in the Bot Performance table.
-        # Input and output are separate rows, as the recorder writes them.
-        for day in range(_USAGE_DAYS):
-            timestamp = now - timedelta(days=day, hours=3)
-            for idx, session in enumerate(sessions):
-                cost = Decimal("0.02") * (idx + 1) * (1 + day % 3)
-                # Only the latest day belongs to the seeded trace — a trace is one request.
-                trace = None if day else traces.get(session.id)
-                for service_kind, quantity, share in (
-                    (ServiceKind.LLM_INPUT, 400 + 50 * idx, Decimal("0.8")),
-                    (ServiceKind.LLM_OUTPUT, 100 + 20 * idx, Decimal("0.2")),
-                ):
-                    created += self._create_usage_record(
-                        team,
-                        session,
-                        provider_type,
-                        priced_model,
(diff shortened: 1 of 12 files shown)


apps/web/management/commands/bootstrap_data.py
```diff
@@ -366,17 +370,20 @@ def _seed_traces(self, team, sessions: list[ExperimentSession]) -> None:
...
-                n_prompt_tokens=200,
-                n_completion_tokens=50,
             )
         self.stdout.write(self.style.SUCCESS(f"  Created {len(sessions)} trace(s)"))
+        return traces
 
-    def _seed_usage_records(self, team, sessions: list[ExperimentSession], llm_provider) -> None:
+    def _seed_usage_records(
+        self, team, sessions: list[ExperimentSession], llm_provider, traces: dict[int, Trace]
+    ) -> None:
```
snopoke, line 379: Fixed in 249b3561d65651e95a2544eb57db8d5fe52e525e — extracted the daily priced-spend loop into `_seed_daily_priced_usage`. `_seed_usage_records` is now 63 lines and the new helper 38, both under the 80 threshold.

The extraction is a pure move (verified: seeding produces an identical set of `UsageRecord` rows before and after).

*— Claude Code*

PR #4025, opened 2026-07-31
Add hybrid search: fuse lexical retrieval with dense retrieval via RRF

VicvekSr9485:
Phase 2 of the approved RAG extension design. Adds lexical retrieval alongside the existing dense retrieval and fuses the two ranked lists with Reciprocal Rank Fusion, introducing no new infrastructure: Postgres full-text search runs on the same database that already stores the chunks.

Lexical index. FileChunkEmbedding gains a stored generated `search_vector` column, to_tsvector(config, coalesce(context,'') || ' ' || text), with a GIN index. Being generated, existing rows are backfilled by the migration and future writes stay consistent with no trigger or application code. The FTS config defaults to `simple` -- language-neutral tokenization, which suits the multilingual and code-mixed corpora common in OCS deployments. The index is built with AddIndexConcurrently in a non-atomic migration so it does not block writes; the migration documents the one-time table rewrite for operators.

Retrieval service. New apps/documents/retrieval.py exposes a single `search_collection` entry point that both `_perform_collection_search` and the collection query preview now use, removing the duplicated dense query. It runs a CosineDistance search and a websearch_to_tsquery search ranked by ts_rank_cd, each fetching fetch_k candidates, then fuses them with weighted RRF (k = 60, dense weight 0.7 by default), deduplicating by chunk id and breaking ties on id so ordering is deterministic.

RRF fuses ranks rather than scores deliberately: cosine distances and ts_rank values are on incomparable scales, so score-level fusion would need brittle per-query normalization. Postgres ts_rank is not true BM25; if evaluation shows lexical quality is the limiting factor, a BM25 extension can be swapped in behind this same interface later.

Configuration. The `flag_hybrid_search` waffle flag gates the behaviour; when inactive, `search_collection` is dense-only and returns the same results as before. dense_weight and fetch_k are settings with per-collection override fields, kept off the pipeline node UI for now; the node's existing max_results still controls the final top_k. Remote indexes never use hybrid search, since their chunks live at the provider and have no local search_vector.

Edge cases covered: queries that tokenize to nothing (stopwords, punctuation, emoji, unbalanced quotes) fall back to dense-only ordering; chunks appearing in only one list are handled natively by RRF; chunks of files that failed indexing and chunks of other collections stay excluded from both branches.

search_vector is excluded from the team export payload -- Postgres recomputes it on insert, and unlike `embedding` it cannot be written back.

Verified the GIN index is used, on PostgreSQL 16 to match the pgvector/pgvector:pg16 image CI and the reference deployment run: at 60k chunks the planner picks `Bitmap Index Scan on file_chunk_search_vector_idx` with `Index Cond: (search_vector @@ 'quokka'::tsquery)` in 1.1ms. At small table sizes it correctly prefers a sequential scan, since the chunk heap is only ~1MB at 5k rows with the embedding column TOASTed out of line.

### Migrations


VicvekSr9485:
@snopoke ready for review. Merged `main` and addressed the bot reviews in 28faa80 and e3b45e4.

  Took three findings: DB check constraints on the two tuning knobs (they're in no form, so `full_clean()` never runs
  and the field validators never actually fired), clamping `fetch_k` to `top_k` (a per-collection `fetch_k` below the
  node's `max_results` was silently truncating results), and trimming the redundant cross-app deps from the migration.

  Pushed back on two, happy to be overruled on either:

  - *Missing waffle flags shouldn't return False.* `WAFFLE_FLAG_DEFAULT` is unset so it's `False`, and waffle's own
  `is_active()` returns `FLAG_DEFAULT` for a missing flag, so this already matches. Treating missing as enabled would
  also switch hybrid search on everywhere before anyone creates the flag, and would change Phase 1's
  contextual-retrieval gating too.
  - *Dense branch does unnecessary work.* `values_list("id")` discards the `select_related`/`only`, so the emitted SQL is `SELECT id` with no join.

  **One thing I'd like your call on:** the migration adds a *stored generated* column, which rewrites
  `files_filechunkembedding` once under an ACCESS EXCLUSIVE lock. That rewrite is what backfills `search_vector`, so
  there's no data migration, but on a large chunk table it wants a maintenance window. If that's not acceptable
  against production volume, I'll switch to a plain `tsvector` column with a trigger and a batched backfill instead:
  no rewrite, at the cost of trigger code. The design doc chose the generated column specifically to avoid triggers,
  so I followed it, but you know the table size and I don't.

  Everything else is in the PR description.

apps/documents/models.py
```diff
@@ -196,6 +198,26 @@ class Collection(BaseTeamModel, VersionsMixin):
         related_name="+",
         help_text="The LLM provider used with contextualizer_llm_model to generate context headers.",
     )
+    # Hybrid search tuning. Null means "use the DOCUMENT_SEARCH_* setting default"; these are
+    # deliberately kept off the pipeline node UI for now to avoid overwhelming bot builders.
+    search_dense_weight = models.FloatField(
```
SmittieC, line 257: How about making `setting.DOCUMENT_SEARCH_DENSE_WEIGHT` the default on this field so we don't have to use `search_dense_weight_or_default`? Same with `DOCUMENT_SEARCH_FETCH_K`

snopoke:
> **One thing I'd like your call on:** the migration adds a _stored generated_ column, which rewrites
> `files_filechunkembedding` once under an ACCESS EXCLUSIVE lock. That rewrite is what backfills `search_vector`, so
> there's no data migration, but on a large chunk table it wants a maintenance window. If that's not acceptable
> against production volume, I'll switch to a plain `tsvector` column with a trigger and a batched backfill instead:
> no rewrite, at the cost of trigger code. The design doc chose the generated column specifically to avoid triggers,
> so I followed it, but you know the table size and I don't.

I'm OK to roll this out in a maintenance window rather than switch to using triggers.

VicvekSr9485:
> > **One thing I'd like your call on:** the migration adds a _stored generated_ column, which rewrites
> > `files_filechunkembedding` once under an ACCESS EXCLUSIVE lock. That rewrite is what backfills `search_vector`, so
> > there's no data migration, but on a large chunk table it wants a maintenance window. If that's not acceptable
> > against production volume, I'll switch to a plain `tsvector` column with a trigger and a batched backfill instead:
> > no rewrite, at the cost of trigger code. The design doc chose the generated column specifically to avoid triggers,
> > so I followed it, but you know the table size and I don't.
> 
> I'm OK to roll this out in a maintenance window rather than switch to using triggers.

Thanks @snopoke. Leaving the generated column as is then.
Both operator notes are in the migration docstring for whoever runs it.

apps/documents/models.py
VicvekSr9485: Worth weighing one consequence first.

A field default is written into each row as the row is created, and the migration would backfill 0.7 into every collection that already exists. After that, DOCUMENT_SEARCH_DENSE_WEIGHT only governs newly created collections. Retuning during rollout would mean a data migration over existing rows instead of a settings change and a deploy.

The nullable version also keeps a distinction that seems worth having: null means nobody has configured this collection, which is not the same as someone deliberately choosing 0.7.

Referencing a setting from a field definition freezes the value rather than linking to it.
There is precedent in the repo: HalfVectorField(dimensions=settings.EMBEDDING_VECTOR_SIZE) is recorded in files/0008 as dimensions=1024.

Given these are tuning knobs for a flag gated rollout, I lean towards keeping the nullable field, but happy to switch if you would rather have the simpler one.

If the two _or_default properties are the part that grates, a third option is to collapse them into a single resolve_search_knobs() returning both values, so retrieval makes one call instead of two.

SmittieC: > Retuning during rollout would mean a data migration over existing rows instead of a settings change and a deploy.

I'm not sure what "during rollout" refers to exactly, but this will be true for collections created with this code as-is as well, and it isn't a big deal to run a data migration in my opinio. I'd rather do a data migration when the default changes than run the risk of someone forgetting to call `search_fetch_k_or_default`.

VicvekSr9485: Done in 2d4ccb90c. Both fields now default to the corresponding DOCUMENT_SEARCH_* setting and are NOT NULL, so the accessors are gone and callers read the fields directly. Constraints and tests updated to match, and the merge conflict with main is resolved.

Good to merge from my side if there is nothing else, so we can move to the 3rd phase.

snopoke:
@VicvekSr9485 I had Claude do a deep dive on this change it found a few issues (one major one):

## 1. High — lexical branch is a no-op for real queries

`apps/documents/retrieval.py:138`

`websearch_to_tsquery` ANDs every lexeme, and the `simple` FTS config removes no
stopwords and does no stemming. So a multi-word natural-language query requires
*every* token — including `what`, `is`, `the`, `of` — to appear in the same chunk.

Verified against Postgres 16:

```
websearch_to_tsquery('simple','what is the capital of France')
  -> 'what' & 'is' & 'the' & 'capital' & 'of' & 'france'

to_tsvector('simple','Paris is the capital of France')
  @@ websearch_to_tsquery('simple','what is the capital of France')  -> false
```

The same query under `english` reduces to `'capit' & 'franc'` and matches.

**Failure scenario.** The LLM search tool calls `_perform_collection_search` with a
natural-language question, `_lexical_candidate_ids` returns `[]`, and
`search_collection` silently falls back to dense-only — paying an extra DB round
trip per search for nothing. The feature is effectively off in production while
appearing to work.

**Why tests don't catch it.** Every query in `test_retrieval.py` is a single token
(`quokka`, `nonexistentterm`), which trivially satisfies the AND.

**Recommended action.** The config must become per-collection data. Full design and
staged plan in [the appendix](#appendix-change-plan-for-finding-1); in brief:

1. Add a `language` field to `Collection` (default `simple`) and use it for both the
   stored vector and the query — they must agree or nothing matches.
2. OR-combine the query lexemes, but only for configs that strip stopwords.
3. Add a multi-word-query test. This is the single highest-value test to add; without
   it the same regression returns silently.

----

The full write-up including the other findings is here: https://docs.google.com/document/d/1BVkd46zfB2OtX2XdXdJjUkCNu75gVCHI6fZqst-YZl8/edit?tab=t.0

VicvekSr9485:
> @VicvekSr9485 I had Claude do a deep dive on this change it found a few issues (one major one):
> 
> ## 1. High — lexical branch is a no-op for real queries
> `apps/documents/retrieval.py:138`
> 
> `websearch_to_tsquery` ANDs every lexeme, and the `simple` FTS config removes no stopwords and does no stemming. So a multi-word natural-language query requires _every_ token — including `what`, `is`, `the`, `of` — to appear in the same chunk.
> 
> Verified against Postgres 16:
> 
> ```
> websearch_to_tsquery('simple','what is the capital of France')
>   -> 'what' & 'is' & 'the' & 'capital' & 'of' & 'france'
> 
> to_tsvector('simple','Paris is the capital of France')
>   @@ websearch_to_tsquery('simple','what is the capital of France')  -> false
> ```
> 
> The same query under `english` reduces to `'capit' & 'franc'` and matches.
> 
> **Failure scenario.** The LLM search tool calls `_perform_collection_search` with a natural-language question, `_lexical_candidate_ids` returns `[]`, and `search_collection` silently falls back to dense-only — paying an extra DB round trip per search for nothing. The feature is effectively off in production while appearing to work.
> 
> **Why tests don't catch it.** Every query in `test_retrieval.py` is a single token (`quokka`, `nonexistentterm`), which trivially satisfies the AND.
> 
> **Recommended action.** The config must become per-collection data. Full design and staged plan in [the appendix](#appendix-change-plan-for-finding-1); in brief:
> 
> 1. Add a `language` field to `Collection` (default `simple`) and use it for both the
>    stored vector and the query — they must agree or nothing matches.
> 2. OR-combine the query lexemes, but only for configs that strip stopwords.
> 3. Add a multi-word-query test. This is the single highest-value test to add; without
>    it the same regression returns silently.
> 
> The full write-up including the other findings is here: https://docs.google.com/document/d/1BVkd46zfB2OtX2XdXdJjUkCNu75gVCHI6fZqst-YZl8/edit?tab=t.0

Thanks for the deep dive @snopoke. All five confirmed against Postgres 16 before changing anything. Fixed in c1a6bcf.

**1 (lexical no-op).** Fixed as the appendix recommends. `Collection.search_language` (29 configs, default `simple`) now drives both sides; `search_vector` stops being generated and is written by the indexing pipeline, since `to_tsvector(<column>::regconfig, ...)` is rejected as not immutable. Terms are OR-combined under a language config and left as websearch AND under `simple`. Added `rebuild_search_vectors()` for the language-change drift case.

Worth flagging on the test: your suggested query reduces to `'capit' & 'franc'` under english, which that chunk contains, so it passes even under AND and would not have caught the OR half. I used your partial-match example instead ("capital of France population 2024 census") and verified it fails when I force AND back on.

**2 (distance annotation).** Fused score carried through and rendered. I also reordered so lexical runs first: when it misses, the dense queryset returns directly, which keeps its `distance` and costs one query rather than two.

**3 (migration).** Split into an atomic AddField and a non-atomic AddIndexConcurrently. Verified the retry path by dropping the index and its migration row. The column is no longer generated, so the add is metadata-only and the rewrite warning is gone from the docstring.

**4.** Fixed the archived-collection raise via `get_all()` since that regression was mine.

**5.** Not changed, because it directly contradicts @SmittieC's review on this PR, which asked for settings-derived field defaults specifically to remove the accessors. I have not flipped it back unilaterally. I did take the fallback and documented the AlterField requirement and the valid range next to the settings. Could you two settle it and I will implement whichever you land on?

Retrieval tests: 39 -> 52, covering all six you named plus the OR case. 2175 passing overall.

apps/documents/retrieval.py
```diff
@@ -0,0 +1,214 @@
...
+    # `values_list` drops the select_related/only from the shared dense queryset, so this
+    # fetches ids alone -- no join -- while keeping one definition of the dense ranking.
+    dense_ids = list(_dense_queryset(collection, query_vector, fetch_k).values_list("id", flat=True))
+    dense_weight = collection.search_dense_weight
+    scores = _rrf_scores([dense_ids, lexical_ids], weights=[dense_weight, 1 - dense_weight])
+    fused_ids = _rank_by_score(scores)[:top_k]
+    return _load_chunks_in_order(fused_ids, scores)
+
+
+def reciprocal_rank_fusion(
```
snopoke: `reciprocal_rank_fusion` doesn't appear to be used anywhere. Can we remove it?

apps/documents/views.py
```diff
@@ -201,10 +201,13 @@ def get_context_data(  # ty: ignore[invalid-method-override]
...
+    chunks = index_manager.query(index_id=pk, query=request.GET.get("query"), top_k=int(request.GET.get("top_k", 5)))
     context = {
-        "chunks": index_manager.query(
-            index_id=pk, query=request.GET.get("query"), top_k=int(request.GET.get("top_k", 5))
-        ),
+        "chunks": chunks,
+        # Derived from the results rather than from the flag: hybrid search falls back to a
+        # dense-only ranking when the query has no lexical matches, and those chunks carry a
+        # distance rather than a fused score.
+        "hybrid_search": any(getattr(chunk, "fused_score", None) is not None for chunk in chunks),
```
snopoke: In the case where only dense search happens this will loop over the full list where it could just check the first one since either all will have or none will have the attribute.

config/settings.py
```diff
@@ -971,6 +971,23 @@ def _before_create(task: dict):
...
+# build a chunk's vector must be the one used to parse the query.
+#
+# These two seed the matching Collection fields when a collection is created. Django bakes the
+# value into the migration, so changing one here needs an AlterField migration, and retuning
+# existing collections needs a data migration. Both must stay within the range enforced by the
+# check constraints on Collection, or every insert fails.
+# Weight given to the dense ranking in fusion; the lexical ranking gets (1 - weight). Range 0..1.
+DOCUMENT_SEARCH_DENSE_WEIGHT = 0.7
+# Candidates pulled from each branch before fusion. Larger values trade latency for recall. Min 1.
+DOCUMENT_SEARCH_FETCH_K = 40
```
snopoke: I think we should drop these two and move the values directly to the model since that is the only place they are used and once the models are created, changing the values here has not impact without a database migration.

snopoke (commented):
@VicvekSr9485 thanks for the updates. There are a few small comments.

config/settings.py
VicvekSr9485: Thank you @snopoke , all three resolved in 2ea6b203.

`reciprocal_rank_fusion` removed. It lost its last caller when `search_collection` started needing the scores as well as the order, so it was only reachable from tests. Preview flag now checks `chunks[0]` instead of scanning. Settings dropped, values moved to literal field defaults.

If there's anything left, kindly let me know and if none, you can merge so we can procced.

apps/documents/models.py
```diff
@@ -448,21 +532,47 @@ def _build_contextualizer(self):
...
+        """Whether retrieval should fuse lexical results with dense results for this collection.
+
+        Remote indexes are excluded: their chunks live at the provider, so there is no local
+        `search_vector` to search lexically.
+        """
+        if self.is_remote_index:
+            return False
+        return self._flag_active_for_team(Flags.HYBRID_SEARCH)
+
+    def rebuild_search_vectors(self) -> int:
```
snopoke: This is only called from tests.

apps/documents/retrieval.py
```diff
@@ -0,0 +1,204 @@
...
+    return sorted(scores, key=lambda chunk_id: (-scores[chunk_id], chunk_id))
+
+
+def _dense_queryset(collection: Collection, query_vector: list[float], limit: int):
+    """Chunks ranked by cosine distance between their embedding and the query embedding."""
+    return (
+        FileChunkEmbedding.objects.annotate(distance=CosineDistance("embedding", query_vector))
+        .filter(collection_id=collection.id)
+        .filter(chunk_from_indexed_file())
+        .order_by("distance")
```
snopoke: This should also order by `id` (same as lexical search) to keep it deterministic

templates/documents/collection_query_results.html
```diff
@@ -2,17 +2,27 @@
...
     </div>
 {% endif %}
 
 {% for chunk in chunks %}
     <div class="flex justify-between items-start mb-1">
         <p class="text-sm text-base-content/70">Source: {{ chunk.file.name }}</p>
-        <div class="badge badge-outline badge-sm {% if chunk.distance <= 0.3 %}badge-success{% elif chunk.distance <= 0.7 %}badge-warning{% else %}badge-error{% endif %}">
-            Distance: {{ chunk.distance|floatformat:3 }}
-        </div>
+        {% if chunk.fused_score %}
```
snopoke: This should check for `fuse_score != None` since it can be 0

apps/service_providers/llm_service/index_managers.py
```diff
@@ -399,6 +414,7 @@ def _embed_file(
                 # Content that is entirely NUL bytes clears the check above but sanitizes away
                 # chunk by chunk. Nothing was indexed, so this is a failure by the same reasoning.
                 raise FileReadException(NO_EXTRACTABLE_TEXT)
+            self._build_search_vectors(embeddings, collection_file.collection)
```
snopoke: This should have it's own exception handling otherwise an error here deletes all the embeddings that were just computed.

apps/documents/tests/test_retrieval.py
```diff
@@ -0,0 +1,494 @@
...
+
+    def test_hybrid_results_carry_a_fused_score(self):
+        collection, file = _make_indexed_collection(search_language=SearchLanguage.ENGLISH)
+        _add_chunk(collection, file, "Paris is the capital of France.", _unit_vector(0))
+
+        with mock.patch.object(type(collection), "get_query_vector", return_value=_unit_vector(0)):
+            with override_flag(HYBRID_FLAG, active=True):
+                results = search_collection(collection, "capital of France", top_k=5)
+
+        assert [result.fused_score for result in results] == [pytest.approx(result.fused_score) for result in results]
```
snopoke: This is comparing `result` with `result` so it will always match. The line below already asserts what this test is testing for.

```suggestion
```

snopoke (commented):
A few small comments but mostly looking good.

apps/documents/tests/test_retrieval.py
VicvekSr9485: All five done in b1b974a.

templates/documents/collection_query_results.html
```diff
@@ -14,7 +14,11 @@
 {% for chunk in chunks %}
     <div class="flex justify-between items-start mb-1">
         <p class="text-sm text-base-content/70">Source: {{ chunk.file.name }}</p>
-        {% if chunk.fused_score %}
+        {# Keyed off the context flag, not the score: a fused score of 0 is legitimate (it happens
+           when the dense weight is 0 and a chunk was found by dense search alone), so testing the
+           value for truthiness would send those rows down the distance branch, which has no
+           distance to show. #}
```
snopoke: Django doesn't support multi-line comments

snopoke (changes requested):
Just one blocking change to the `templates/documents/collection_query_results.html` template

templates/documents/collection_query_results.html
VicvekSr9485: Thanks for catching that @snopoke. Should have simply used the proper {% comment %}...{% endcomment %} tag due to how Django's lexer works. Already collapsed it to a single line.

snopoke (approved):

PR #4076, opened 2026-08-04
Fix collection polling targets and extract PDF text with PDFium

SmittieC:
### Product Description

Collection file lists no longer disappear while a sync is running, and PDF text extraction is faster and more accurate, especially on table-heavy documents.

### Technical Description

Two independent changes in the same feature area; review separately.

**htmx polling targets** (`970f449c`) — htmx resolves `hx-target="this"` to the closest ancestor that *declares* the attribute, not the element making the request. Three polling fragments declared none, so their 5s polls replaced the file-list container with a single status badge, taking its `search-files` subscription with it. Each is now pinned to itself. The manual-sync view hit the same symptom by returning an empty body mid-sync against an `outerHTML` swap.

**PDF extraction via PDFium** (`513257f1`, `bf6d8ea9`) — PDFs route to a pypdfium2 reader, one `DocumentPart` per page; sync and index ingestion both funnel through `read_file_content`.

Decisions worth attention:

- **No pdfminer fallback.** An unopenable PDF raises `FileReadException` rather than plaintext-decoding the raw bytes, which used to index megabytes of noise as document text. pdfminer isn't kept as a backstop because its slow path is unbounded CPU-bound pure Python — on a damaged 133-page WHO document it ran 10 minutes without finishing, where PDFium fails in milliseconds.
- **A module lock serialises parsing.** PDFium isn't thread-safe and Celery runs `--pool threads --concurrency 20`; unsynchronised, 386 of 400 concurrent parses of a *valid* PDF failed, which would mark good files permanently FAILED under load. Pinned by `test_concurrent_reads_do_not_corrupt_pdfium`.
- **Header sniffing scans 1024 bytes**, not just offset 0 — the spec allows `%PDF-` at a non-zero offset, and such files previously fell through to pdfminer. Trade-off: a text file mentioning `%PDF-` early now fails instead of being read as text.

### Docs and Changelog

Extracted text differs from pdfminer's (per-page parts, collapsed justification whitespace), so re-syncing an existing collection regenerates different chunks and embeddings. Adds `pypdfium2` (BSD-3-Clause / Apache-2.0).

### DIFF
--- a/apps/documents/readers.py
+++ b/apps/documents/readers.py
@@ -103,9 +103,9 @@ def pdf_read(file_obj) -> Document:
     so a thread cannot be cancelled, and a SIGALRM budget fires only on the main thread,
     which is not where our gevent or threaded workers run requests.
 
-    PDFium itself is not thread-safe, and both the Celery workers (--pool threads) and
-    gunicorn (--threads) parse PDFs concurrently in one process, so a module lock serialises
-    every parse: unsynchronised concurrent use corrupts library state and fails valid PDFs.
+    PDFium itself is not thread-safe and we parse PDFs from threaded workers, so a module lock
+    serialises every parse: unsynchronised concurrent use corrupts library state and fails
+    valid PDFs.
     """
     import pypdfium2  # noqa: PLC0415 - TID253: loads a shared library, keep off the import path
 
--- a/apps/documents/readers.py
+++ b/apps/documents/readers.py
@@ -1,4 +1,5 @@
 import logging
+import threading
 from io import BytesIO
 
 from bs4 import UnicodeDammit
@@ -7,6 +8,17 @@ from pydantic import BaseModel, Field
 from apps.files.models import File
 
 logger = logging.getLogger("ocs.documents")
+
+PDF_MAGIC = b"%PDF-"
+PDF_HEADER_SCAN_BYTES = 1024
+
+_pdfium_lock = threading.Lock()
+
+
+def _looks_like_pdf(head: bytes) -> bool:
+    """The PDF header may sit anywhere in the first 1024 bytes, not just at offset zero;
+    mail gateways and misbehaving HTTP servers produce such files, and PDFium reads them."""
+    return PDF_MAGIC in head[:PDF_HEADER_SCAN_BYTES]
 
 
 class FileReadException(Exception):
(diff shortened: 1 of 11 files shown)


apps/documents/readers.py
```diff
@@ -43,15 +54,87 @@ def get_contents_as_string(self):
...
+    scatters table columns away from the rows they belong to.
+
+    There is deliberately no pdfminer fallback for a file PDFium cannot open. PDFium already
+    recovers the damage worth recovering -- it rebuilds a clobbered xref table in under a
+    millisecond, the same damage that took pdfminer ~13 minutes to return a fragment of --
+    and pdfminer's slow path cannot be bounded once it starts: it is CPU-bound pure Python,
+    so a thread cannot be cancelled, and a SIGALRM budget fires only on the main thread,
+    which is not where our gevent or threaded workers run requests.
+
+    PDFium itself is not thread-safe, and both the Celery workers (--pool threads) and
```
SmittieC: remove the mention  of celer yand gunicorn. it's too specific

apps/documents/test_readers.py
```diff
@@ -52,3 +64,124 @@ def test_plaintext_reader_unicode_dammit_raises_exception(self):
...
+            with pytest.raises(FileReadException, match="Could not extract text"):
+                markitdown_read(BytesIO(b"%PDF-1.4 binary junk"))
+
+    def test_unreadable_pdf_with_offset_header_fails_instead_of_being_decoded(self):
+        with mock.patch("markitdown.MarkItDown") as markitdown:
+            markitdown.return_value.convert.side_effect = FileConversionException(attempts=[])
+
+            with pytest.raises(FileReadException, match="Could not extract text"):
+                markitdown_read(BytesIO(b"junk prefix %PDF-1.4 then binary junk"))
+
```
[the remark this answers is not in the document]
SmittieC: Fixed in dfddae7a1. Merged the two methods into one `pytest.mark.parametrize` test with `header-at-start` / `header-with-prefix` case IDs; the shared MarkItDown mock and `FileReadException` assertion are unchanged.

apps/documents/readers.py
SmittieC: Fixed in 0f647d704. The rationale now reads "PDFium itself is not thread-safe and we parse PDFs from threaded workers" — the reason for the lock without pinning it to one deployment setup. I applied the same edit to the duplicated sentence in `test_readers.py::test_concurrent_reads_do_not_corrupt_pdfium` so the two don't diverge.

snopoke (dismissed):
do we still need `markitdown[pdf]`?

Issue #4008, opened 2026-07-30
Deprecate Gemini 2.5 models on Vertex AI (Enterprise Agent Platform) — Oct 2026


ocs-percy:
## Overview

Google has issued a deprecation notice for Gemini 2.5 models on the **Gemini Enterprise Agent Platform (Vertex AI)**. OCS currently lists all three deprecated models in the `google_vertex_ai` provider, including the default model.

> **Note:** This notice applies only to `google_vertex_ai`. The Google AI (AI Studio) provider has separate deprecation dates — a follow-up notice is expected.

---

## Timeline

| Date | What happens |
|------|-------------|
| **October 20, 2026** | Models enter Extended Lifecycle Access (ELA). Same GA terms and pricing. |
| **January 28, 2027** | ELA pricing ends. Costs increase significantly. Availability in some regions may change. |

---

## Affected models in OCS

All three are in `apps/service_providers/llm_service/default_models.py` under `google_vertex_ai`:

| Model | Current OCS status | Recommended migration target |
|---|---|---|
| `gemini-2.5-flash` | **`is_default=True`** | `gemini-3.5-flash` / `gemini-3.5-flash-lite` / `gemini-3.1-flash-lite` |
| `gemini-2.5-pro` | **`is_translation_default=True`** | `gemini-3.5-flash` / `gemini-3.6-flash` |
| `gemini-2.5-flash-lite` | Listed (no default flag) | `gemini-3.1-flash-lite` / Gemma 4 |

The default and translation-default flags are the most urgent — any team that hasn't explicitly chosen a model will be on deprecated models after October 20.

---

### 1. Add missing migration target model

`gemini-3.1-flash-lite` is the primary recommended replacement for `gemini-2.5-flash-lite` and one of the targets for `gemini-2.5-flash`, but it's not yet in the `google_vertex_ai` model list. Add it.

### 2. Update defaults in `google_vertex_ai`

- Change `is_default` from `gemini-2.5-flash` to `gemini-3.5-flash` (or `gemini-3.5-flash-lite`)
- Change `is_translation_default` from `gemini-2.5-pro` to `gemini-3.5-flash` or `gemini-3.6-flash`

A migration will be needed to repoint any DB records using the old defaults.

### 3. Mark 2.5 models as deprecated

Set `deprecated=True` (and a `replacement`) on all three models in `google_vertex_ai`:

```python
[...]

PR #4080, opened 2026-08-04
Deprecate Gemini 2.5 models on Vertex AI

snopoke:
### Product Description
Teams on the Google Vertex AI provider are moved off the Gemini 2.5 models ahead of Google's retirement dates (Extended Lifecycle Access 2026-10-20, ELA pricing ends 2027-01-28). New chatbots default to `gemini-3.5-flash` instead of the deprecated `gemini-2.5-flash`, and teams still referencing a 2.5 model are notified with the recommended replacement.

### Technical Description
Closes part of #4008 (items 2 and 3).

- Adds `gemini-3.1-flash-lite` to `google_vertex_ai` (1M context, GA) with pricing.
- Moves `is_default` to `gemini-3.5-flash` and `is_translation_default` to `gemini-3.6-flash`. These flags are runtime-only (form preselection and `get_default_model`) and are not persisted, so no DB repointing is needed — existing explicit references stay put and are covered by the deprecation notification instead.
- Marks all three 2.5 models `deprecated=True` with replacements.
- `llm_model_migration()` and `notify_deprecated_models` move from `0070` to the new `0071` per the once-per-deployment rule in `docs/developer_guides/managing_models.md`. `0070`'s `load_pricing_data()` stays (safe to re-run).

The `google` (AI Studio) provider is deliberately untouched — Google's notice excludes it and a separate notice is expected (#4008 item 5).

Issue #4008 item 4 (thought signature circulation) was investigated and needs no code change; findings are in a comment on the issue. Short version: the tool-calling loop runs in a single in-process `agent.invoke()` with no checkpointer, so `langchain-google-*` keeps signatures in `additional_kwargs` for the whole loop, and OCS never persists tool messages. One risk worth knowing: `langchain-google-vertexai` has no dummy-signature fallback, so anything that splits a turn across process boundaries would 400 on Gemini 3 here where the AI Studio provider degrades silently.

### Migrations

### Demo
```
vertex default: gemini-3.5-flash
translation defaults: {... 'Google Vertex AI': 'gemini-3.6-flash'}
```

### Docs and Changelog

### DIFF
--- a/apps/service_providers/llm_service/default_models.py
+++ b/apps/service_providers/llm_service/default_models.py
@@ -146,13 +146,16 @@ DEFAULT_LLM_PROVIDER_MODELS = {
         Model("gemini-2.0-flash", 1048576, deprecated=True),
     ],
     "google_vertex_ai": [
-        Model("gemini-3.6-flash", 1048576),
-        Model("gemini-3.5-flash", 1048576),
+        Model("gemini-3.6-flash", 1048576, is_translation_default=True),
+        Model("gemini-3.5-flash", 1048576, is_default=True),
         Model("gemini-3.5-flash-lite", 1048576),
         Model("gemini-3.1-pro-preview", 1048576),
-        Model("gemini-2.5-pro", 1048576, is_translation_default=True),
-        Model("gemini-2.5-flash", 1048576, is_default=True),
-        Model("gemini-2.5-flash-lite", 1048576),
+        Model("gemini-3.1-flash-lite", 1048576),
+        # Google is retiring the 2.5 models on the Gemini Enterprise Agent Platform: they enter
+        # Extended Lifecycle Access on 2026-10-20 and lose ELA pricing on 2027-01-28.
+        Model("gemini-2.5-pro", 1048576, deprecated=True, replacement="gemini-3.6-flash"),
+        Model("gemini-2.5-flash", 1048576, deprecated=True, replacement="gemini-3.5-flash"),
+        Model("gemini-2.5-flash-lite", 1048576, deprecated=True, replacement="gemini-3.1-flash-lite"),
     ],
 }
 
--- a/apps/service_providers/migrations/0070_deepseek_model_updates.py
+++ b/apps/service_providers/migrations/0070_deepseek_model_updates.py
@@ -1,24 +1,15 @@
 from django.db import migrations
 
-from apps.cost_tracking.migration_utils import load_pricing_data
-
 
 class Migration(migrations.Migration):
     dependencies = [
         ("service_providers", "0069_add_deepseek_v4_flash"),
+        # Retained so the graph stays stable for environments that already applied this.
         ("cost_tracking", "0001_initial"),
-        # Retained from when this migration ran llm_model_migration() /
-        # notify_deprecated_models, so the graph stays stable for environments that
-        # already applied it.
         ("evaluations", "0018_evaluator_llm_provider_fks"),
         ("teams", "0013_team_files_export_team_files_export_task_id"),
     ]
 
     operations = [
-        # Add deepseek-v4-pro and mark deepseek-chat / deepseek-reasoner deprecated
-        # (replacement: deepseek-v4-flash).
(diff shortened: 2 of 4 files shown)


apps/service_providers/llm_service/default_models.py
```diff
@@ -146,13 +146,16 @@ def k(n: int) -> int:
...
         Model("gemini-3.5-flash-lite", 1048576),
         Model("gemini-3.1-pro-preview", 1048576),
-        Model("gemini-2.5-pro", 1048576, is_translation_default=True),
-        Model("gemini-2.5-flash", 1048576, is_default=True),
-        Model("gemini-2.5-flash-lite", 1048576),
+        Model("gemini-3.1-flash-lite", 1048576),
+        # Google is retiring the 2.5 models on the Gemini Enterprise Agent Platform: they enter
+        # Extended Lifecycle Access on 2026-10-20 and lose ELA pricing on 2027-01-28.
+        Model("gemini-2.5-pro", 1048576, deprecated=True, replacement="gemini-3.6-flash"),
+        Model("gemini-2.5-flash", 1048576, deprecated=True, replacement="gemini-3.5-flash"),
```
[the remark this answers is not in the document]
snopoke: Recommended replacements:
* Gemini 2.5 Flash Lite --> Gemini 3.1 Flash Lite / Gemma 4
* Gemini 2.5 Flash --> Gemini 3.5 Flash Lite / 3.1 Flash Lite / 3.5 Flash
* Gemini 2.5 Pro --> Gemini 3.5 Flash / 3.6 Flash

apps/service_providers/migrations/0070_deepseek_model_updates.py
```diff
@@ -1,27 +1,24 @@
...
     ]
 
     operations = [
         # Add deepseek-v4-pro and mark deepseek-chat / deepseek-reasoner deprecated
         # (replacement: deepseek-v4-flash).
-        llm_model_migration(),
+        # llm_model_migration() and notify_deprecated_models moved to 0071_deprecate_vertex_gemini_2_5
+        #
         # Seed pricing for deepseek-v4-flash and deepseek-v4-pro (off-peak rates).
         load_pricing_data(),
```
snopoke: drop this as well since we are running it again in a later migration

apps/service_providers/migrations/0070_deepseek_model_updates.py
```diff
@@ -1,27 +1,24 @@
...
+        # already applied it.
         ("evaluations", "0018_evaluator_llm_provider_fks"),
-        # notify_deprecated_models queries Team with live models, so all Team
-        # schema changes must be applied first.
         ("teams", "0013_team_files_export_team_files_export_task_id"),
     ]
 
     operations = [
         # Add deepseek-v4-pro and mark deepseek-chat / deepseek-reasoner deprecated
         # (replacement: deepseek-v4-flash).
```
snopoke: remove comment

apps/service_providers/migrations/0070_deepseek_model_updates.py
snopoke: Fixed in 25a49ed4e. Dropped load_pricing_data() from 0070 — it loads the whole seed file, so 0071 covers the DeepSeek entries too. 0070 is now a no-op with its dependencies retained for graph stability.

*— Claude Code*

apps/service_providers/migrations/0070_deepseek_model_updates.py
snopoke: Fixed in 25a49ed4e. Removed the stale operations comments (and the now-unused load_pricing_data import).

*— Claude Code*

apps/service_providers/llm_service/default_models.py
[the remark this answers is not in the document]
snopoke: Not applicable — keeping the current mapping, confirmed with the author.

The replacement field drives what OCS steers teams onto, so it points at the models this PR makes is_default / is_translation_default (gemini-3.5-flash and gemini-3.6-flash). Following the upstream lifecycle table would send gemini-2.5-pro users to gemini-3.1-pro-preview, i.e. onto a preview model, which we do not want as a migration target.

*— Claude Code*

PR #4103, opened 2026-08-04
Read node resource ids from the FK columns when building flow_data

### LINKED ISSUE #4100: Read FK params from FK fields when constructing flow_data
The Node resources that have FKs on a Node db record as well (llm provider id, model provider id etc) are still stored / duplicated in `Node.params` as well and read from there when we call `to_flow_node()`. We need to pluck it out from there and enrich params with the actual FKs

SmittieC:
### Product Description

A pipeline node that references a resource which has since been deleted now shows that field as unset in the editor, instead of showing an id that no longer resolves.

### Technical Description

`Node` mirrors the resource ids in its `params` (`llm_provider_id`, `collection_id`, `collection_index_ids`, ...) onto FK/M2M columns, but `to_flow_node()` still served the copies in `params`. Those copies drift: `SET_NULL` nulls the column when a referenced resource is deleted while the stale id lingers in `params`.

- ``Node.resource_params()`` (new) returns the resource ids read off the FK columns, plus ``collection_index_ids`` from the ``collection_indexes`` M2M. Only params the row already carries are returned, so a node type that references no resource does not grow a null one.
- ``to_flow_node()`` enriches the params it serves from it, so the columns win over the duplicated ids.
- ``Pipeline.flow_data`` prefetches the M2M in one query via ``prefetch_related_objects`` (reusing a prefetched ``node_set``); ``revert_to_version`` prefetches the same way.
- ``SelectWidget`` compares its option value as a string, since ids now arrive as ints (otherwise the edit link goes missing).

`params` stays the write side — it is what the editor posts, what the node classes validate and what the runtime builds from. Only the read is corrected. Removing the ids from stored `params` altogether is a follow-up: it needs the runtime build, validation, versioning specs and deletion helpers migrated first, plus a data migration.

Closes #4100

In a followup PR we can remove the resource fields that are duplicated by the FKs from the params, or not. (I'm leaning towards leaving it)

### Demo

`TestEditorEndpointServesResourceFKs` covers it end-to-end: PATCH a node with a provider → delete the provider → GET serves `llm_provider_id: null` while the row's `params` still carry the stale id.

### Docs and Changelog

### DIFF
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -296,7 +296,12 @@ class Pipeline(BaseTeamModel, VersionsMixin):
         # rows still describe a graph. Same trigger as ``data_without_positions``' guard but a
         # different fallback: that one returns the empty data as-is rather than rebuilding.
         flow = Flow(**(self.data or {"edges": []}))
-        flow.nodes = [node.to_flow_node() for node in self.node_set.all()]
+        nodes = list(self.node_set.all())
+        # Each node reads its collection_indexes M2M (Node.resource_params); prefetch them in one
+        # query. prefetch_related_objects works off the list, so a prefetched node_set is still
+        # reused, and it no-ops when the M2M was prefetched along with it.
+        models.prefetch_related_objects(nodes, "collection_indexes")
+        flow.nodes = [node.to_flow_node() for node in nodes]
         return flow.model_dump()
 
     @property
@@ -338,7 +343,8 @@ class Pipeline(BaseTeamModel, VersionsMixin):
         The version's stored ``data`` supplies the edges only.
         """
         node_data: dict[str, FlowNode | None] = {}
-        for version_node in version.node_set.all():
+        # to_flow_node reads each node's collection_indexes M2M — prefetch rather than query per row.
+        for version_node in version.node_set.prefetch_related("collection_indexes"):
             flow_node = version_node.to_flow_node()
             for spec in get_versioned_param_specs(version_node.type):
                 spec.revert_referenced_record(version_node, flow_node.data.params)
@@ -503,8 +509,11 @@ class Node(BaseModel, VersionsMixin, CustomActionOperationMixin):
         """This row as a react-flow node, content included.
 
         ``params`` is deep-copied so a caller that rewrites the returned node's params
-        (``Pipeline.revert_to_version``) cannot reach back into this row's stored dict.
+        (``Pipeline.revert_to_version``) cannot reach back into this row's stored dict, and the
+        resource ids it duplicates are re-read from the FK columns (see ``resource_params``).
         """
+        params = copy.deepcopy(self.params or {})
+        params.update(self.resource_params())
         return FlowNode(
             id=self.flow_id,
             position=self.position or {"x": 0, "y": 0},
@@ -512,7 +521,7 @@ class Node(BaseModel, VersionsMixin, CustomActionOperationMixin):
             data=FlowNodeData(
                 id=self.flow_id,
                 type=self.type,
-                params=copy.deepcopy(self.params),
+                params=params,
                 label=self.label,
             ),
         )
@@ -588,6 +597,33 @@ class Node(BaseModel, VersionsMixin, CustomActionOperationMixin):
             for field in cls._meta.get_fields()
             if isinstance(field, models.ForeignKey) and field.remote_field.on_delete is models.SET_NULL
         ]
+
+    def resource_params(self) -> dict:
+        """The ids of the resources this node references, read off its FK columns.
+
+        ``params`` duplicates these ids — it is what the editor posts and what the node classes
+        validate — but the FK columns are the constraint-backed mirror of them (see
+        ``_sync_resource_fk_fields``): when a referenced resource is deleted the column is nulled
+        while the stale id lingers in params. Reading them back from the columns is what
+        ``to_flow_node`` serves, so a dangling reference reads as unset rather than as an id that
+        no longer resolves, and the ids come out as ints — the form the columns store them in.
+
+        Only params the row already carries are returned: a node type that references no resource
+        must not grow a null param for one.
+        """
+        params = self.params or {}
+        resource_params = {
+            f"{field_name}_id": getattr(self, f"{field_name}_id")
+            for field_name in self.resource_fk_fields()
+            if f"{field_name}_id" in params
+        }
+        if "collection_index_ids" in params:
+            # ``all()`` rather than values_list so a prefetched M2M is reused; sorted because the
+            # through rows have no ordering of their own and the read must be stable.
+            resource_params["collection_index_ids"] = sorted(
+                collection.id for collection in self.collection_indexes.all()
+            )
+        return resource_params
 
     def _sync_resource_fk_fields(self):
         """Populate FK/M2M fields from the params JSON.
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -610,7 +610,10 @@ class Node(BaseModel, VersionsMixin, CustomActionOperationMixin):
 
         Scalar ids are only returned for params the row already carries: a node type that
         references no resource must not grow a null param for one. ``collection_index_ids`` is
-        returned whenever the M2M holds anything, whether or not params carries it.
+        returned whenever the M2M holds anything — the M2M is the reference, so what it holds is
+        served whether or not the params copy was ever written — and also whenever params carries
+        the key, so that a params copy left behind by a deleted index is corrected to ``[]``
+        rather than served as-is. A node with neither grows no empty list.
         """
         params = self.params or {}
         resource_params = {
@@ -621,7 +624,7 @@ class Node(BaseModel, VersionsMixin, CustomActionOperationMixin):
         # ``all()`` rather than values_list so a prefetched M2M is reused; sorted because the
         # through rows have no ordering of their own and the read must be stable.
         collection_index_ids = sorted(collection.id for collection in self.collection_indexes.all())
-        if collection_index_ids:
+        if collection_index_ids or "collection_index_ids" in params:
             resource_params["collection_index_ids"] = collection_index_ids
         return resource_params
 
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -608,8 +608,9 @@ class Node(BaseModel, VersionsMixin, CustomActionOperationMixin):
         ``to_flow_node`` serves, so a dangling reference reads as unset rather than as an id that
         no longer resolves, and the ids come out as ints — the form the columns store them in.
 
-        Only params the row already carries are returned: a node type that references no resource
-        must not grow a null param for one.
+        Scalar ids are only returned for params the row already carries: a node type that
+        references no resource must not grow a null param for one. ``collection_index_ids`` is
+        returned whenever the M2M holds anything, whether or not params carries it.
         """
         params = self.params or {}
         resource_params = {
@@ -617,12 +618,11 @@ class Node(BaseModel, VersionsMixin, CustomActionOperationMixin):
             for field_name in self.resource_fk_fields()
             if f"{field_name}_id" in params
         }
-        if "collection_index_ids" in params:
-            # ``all()`` rather than values_list so a prefetched M2M is reused; sorted because the
-            # through rows have no ordering of their own and the read must be stable.
-            resource_params["collection_index_ids"] = sorted(
-                collection.id for collection in self.collection_indexes.all()
-            )
+        # ``all()`` rather than values_list so a prefetched M2M is reused; sorted because the
+        # through rows have no ordering of their own and the read must be stable.
+        collection_index_ids = sorted(collection.id for collection in self.collection_indexes.all())
+        if collection_index_ids:
+            resource_params["collection_index_ids"] = collection_index_ids
         return resource_params
 
     def _sync_resource_fk_fields(self):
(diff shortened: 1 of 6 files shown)


apps/pipelines/models.py
```diff
@@ -589,6 +598,33 @@ def resource_fk_fields(cls):
...
+        Only params the row already carries are returned: a node type that references no resource
+        must not grow a null param for one.
+        """
+        params = self.params or {}
+        resource_params = {
+            f"{field_name}_id": getattr(self, f"{field_name}_id")
+            for field_name in self.resource_fk_fields()
+            if f"{field_name}_id" in params
+        }
+        if "collection_index_ids" in params:
```
SmittieC: @claude remove this check. If there are collection indexes (self.ecollection_indexes.all() returns anything), it should be included in resource_params

apps/pipelines/models.py
```diff
@@ -296,7 +296,12 @@ def flow_data(self) -> dict:
         # rows still describe a graph. Same trigger as ``data_without_positions``' guard but a
         # different fallback: that one returns the empty data as-is rather than rebuilding.
         flow = Flow(**(self.data or {"edges": []}))
-        flow.nodes = [node.to_flow_node() for node in self.node_set.all()]
+        nodes = list(self.node_set.all())
+        # Each node reads its collection_indexes M2M (Node.resource_params); prefetch them in one
+        # query. prefetch_related_objects works off the list, so a prefetched node_set is still
+        # reused, and it no-ops when the M2M was prefetched along with it.
+        models.prefetch_related_objects(nodes, "collection_indexes")
```
SmittieC: rather prefetch at the .flow_data call sites

apps/pipelines/models.py
```diff
@@ -589,6 +598,36 @@ def resource_fk_fields(cls):
             if isinstance(field, models.ForeignKey) and field.remote_field.on_delete is models.SET_NULL
         ]
 
+    def resource_params(self) -> dict:
```
SmittieC, line 602: Shorten this docstring and keep it succinct

apps/pipelines/models.py
```diff
@@ -589,6 +598,36 @@ def resource_fk_fields(cls):
...
+        returned whenever the M2M holds anything — the M2M is the reference, so what it holds is
+        served whether or not the params copy was ever written — and also whenever params carries
+        the key, so that a params copy left behind by a deleted index is corrected to ``[]``
+        rather than served as-is. A node with neither grows no empty list.
+        """
+        params = self.params or {}
+        resource_params = {
+            f"{field_name}_id": getattr(self, f"{field_name}_id")
+            for field_name in self.resource_fk_fields()
+            if f"{field_name}_id" in params
```
SmittieC: we don't want the resource ids in the params (not for this PR), so we can just build the ids from resource_fk_fields

SmittieC: Done in `d5539c5b7`. `resource_params` no longer looks at `params` at all — every scalar id comes off `resource_fk_fields()` and `collection_index_ids` off the M2M, unconditionally:

```python
resource_params = {
    f"{field_name}_id": getattr(self, f"{field_name}_id") for field_name in self.resource_fk_fields()
}
resource_params["collection_index_ids"] = sorted(index.id for index in self.collection_indexes.all())
```

That also retires the `or "collection_index_ids" in params` clause from the thread above: the key is now always present, so a stale copy left by a deleted index is corrected to `[]` without needing to check params for it.

Two consequences worth naming, both intended by "we don't want the resource ids in the params" but visible before that lands:

1. Every node now serves all seven keys, so a `StartNode` reads as `{"name": "start", "llm_provider_id": None, ..., "collection_index_ids": []}`. The editor round-trips params, so a save persists those nulls into `Node.params`. Harmless for `_sync_resource_fk_fields` (`as_int(None)` → `None`) and the node classes ignore unknown params, but it does mean the JSON grows before it shrinks.
2. `Node._get_version_details` iterates `self.params.items()`, so once a node has been saved through the editor its version-history entry will list the null resource params as rows. Worth a follow-up when the ids come out of params for good — flagging rather than fixing here since that's the same "not for this PR" boundary.

Tests updated: `test_params_the_node_does_not_carry_are_not_added` is now `test_every_resource_id_is_served_from_the_columns` (asserts the full dict), plus `test_scalar_ids_are_served_even_when_params_omit_them`. `test_revert_from_version_with_old_format_data` and `test_post_persists_layout_only_data` pinned the old exact-dict shape and were updated to the new one.

apps/pipelines/models.py
SmittieC: Done in `d5539c5b7`. `flow_data` no longer prefetches; it just reads `self.node_set.all()` and carries a comment saying callers must have `node_set__collection_indexes` prefetched. The prefetch moved to the call sites:

- `pipeline_data` GET — `Pipeline.objects.prefetch_related("node_set__collection_indexes").get(pk=pk)`
- `_handle_pipeline_post` / `_handle_pipeline_patch` — the existing `prefetch_related("node_set")` became `prefetch_related("node_set__collection_indexes")`
- `get_widget_page_context` — `prefetch_related_objects([pipeline], ...)`, since it reaches `flow_data` via `data_without_positions` and the pipeline arrives already fetched from two different view hierarchies
- `revert_to_version` already did this

One case doesn't have a queryset to hang it off: the POST/PATCH responses read `flow_data` *after* `clear_node_caches()`, and that calls `refresh_from_db()` with no fields, which resets `_prefetched_objects_cache` — so the response read would have gone back to being an N+1. Rather than duplicate a re-fetch in both handlers I put the re-prefetch in `clear_node_caches` itself, which is the one place that knows the cache was just dropped and whose every caller reads `flow_data` next. Happy to move it into the two views instead if you'd rather it be literal.

New test `test_editor_load_reads_the_indexes_once` pins this at the endpoint rather than the model: it counts queries touching `pipelines_node_collection_indexes` during a real GET, parametrized over 1 and 6 nodes. Verified it fails without the view prefetch (3 reads for 1 node, 8 for 6). `test_collection_indexes_are_prefetched` now fetches the way a call site does and pins 3 queries across 1/3/8 nodes, and `test_clear_node_caches_re_primes_the_prefetch` covers the save-path read.

apps/pipelines/models.py
SmittieC: Done in `d5539c5b7` — cut from 16 lines to 5:

```
The ids of the resources this node references, read off its FK columns and M2M.

Params carries copies of these ids, but the columns are the constraint-backed mirror of
them (see ``_sync_resource_fk_fields``): deleting a resource nulls the column while the
stale id lingers in params. ``to_flow_node`` serves what this returns, so a dangling
reference reads as unset rather than as an id that no longer resolves.
```

Most of the length was the two paragraphs justifying the `in params` gates, which are gone with the change in the thread below.

apps/pipelines/models.py
SmittieC: Superseded by `d5539c5b7`. The `or "collection_index_ids" in params` clause I added in `0a9a695` is gone too — `resource_params` no longer reads `params` at all, so `collection_index_ids` is always served from the M2M. That gives you what you originally asked for (the M2M is the reference) *and* keeps the deleted-index correction, since an always-present key means a stale params copy is overwritten with `[]` rather than falling through. See the thread on line 622 for the full change.

apps/pipelines/models.py
```diff
@@ -296,7 +296,12 @@ def flow_data(self) -> dict:
         # rows still describe a graph. Same trigger as ``data_without_positions``' guard but a
         # different fallback: that one returns the empty data as-is rather than rebuilding.
         flow = Flow(**(self.data or {"edges": []}))
-        flow.nodes = [node.to_flow_node() for node in self.node_set.all()]
+        nodes = list(self.node_set.all())
```
SmittieC: Not addressed. This is a module-level metric — 6 responsibilities across the 44 functions already in `apps/pipelines/models.py` — flagged on a file this PR adds one method to. Splitting `models.py` is well outside this change. This pass actually shrinks the method it landed on (`resource_params` went from 10 lines to 5).

apps/pipelines/tests/test_node_resource_fks.py
```diff
@@ -153,6 +158,292 @@ def test_update_nodes_from_data_populates_fks(self):
...
+    """to_flow_node() serves the resource ids from the FK columns, not from the copies in params."""
+
+    def test_scalar_ids_come_from_the_fk_columns(self):
+        """The editor posts (and params therefore store) the ids as strings; the flow node serves
+        the ints held by the columns."""
+        provider = LlmProviderFactory.create()
+        model = LlmProviderModelFactory.create()
+        node = NodeFactory.create(
+            type="LLMResponseWithPrompt",
+            params={"llm_provider_id": str(provider.id), "llm_provider_model_id": str(model.id)},
```
SmittieC: remove the resource ids from the params here so we can be 100% sure it's extracted from the FK resources

SmittieC: Fixed in `547261a54`. `test_scalar_ids_come_from_the_fk_columns` now creates the node with `params={"name": "llm"}` and the FKs set on the row, so the served ints can only have come off the columns — there is no copy in params to fall back to. Did the same to `test_collection_index_ids_come_from_the_m2m`, which had the equivalent `{"collection_index_ids": [str(second.id)]}` copy.

That made three tests redundant, so they went in the same commit:
- `test_scalar_ids_are_served_even_when_params_omit_them` and `test_collection_index_ids_are_served_even_when_params_omit_them` — that is now exactly what the two tests above do.
- `test_fk_column_wins_over_a_drifted_params_copy` — with `resource_params` no longer reading params at all, a *different* id in params is the same code path as no id; the drift cases that still pin something (`test_reference_to_deleted_resource_reads_as_unset`, `test_deleted_collection_index_reads_as_empty`, and the endpoint test that asserts the row keeps the stale id while the read is corrected) are all kept.

Also removed `test_prefetched_node_set_is_reused` per your other note, and `test_collection_indexes_are_prefetched` with it — the prefetch now lives at the call sites, so hand-writing `prefetch_related(...)` in a test only asserted that Django's prefetch works. `test_editor_load_reads_the_indexes_once` pins the N+1 at the real call site, and `test_clear_node_caches_re_primes_the_prefetch` covers the save path.

Note the flow_data-level `test_resource_ids_served_from_fk_columns` still posts string ids through `update_nodes_from_data` — there params carrying them is the point, since that is what populates the columns.

Verification: `apps/pipelines` suite green (569 passed), ruff check/format and ty clean.

apps/pipelines/models.py
```diff
@@ -194,16 +194,20 @@ def update_nodes_from_data(self, node_data: dict[str, FlowNode | None]) -> None:
             created_node.update_from_params()
 
     def clear_node_caches(self) -> None:
-        """Forget the cached ``Node`` rows and the ``flow_data`` rebuilt from them.
+        """Re-read the ``Node`` rows and drop the ``flow_data`` built from the stale ones.
```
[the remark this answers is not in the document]
SmittieC: Not addressed — no code change. This is a module-level metric (6 responsibilities across the 44 functions already in `apps/pipelines/models.py`) flagged on a file this PR adds one method to. Splitting `models.py` is well outside this change; `resource_params` itself is 5 lines. Same call as the earlier Low Cohesion thread on this file.

apps/pipelines/tests/test_node_resource_fks.py
```diff
@@ -1,10 +1,15 @@
+import json
```
SmittieC, line 1: Partly addressed. The metric counts one "responsibility" per test function, which is what a test module is supposed to look like — a low-cohesion score on a test file isn't a defect signal. That said, `547261a54` does cut it: five tests removed (seven cases, counting the parametrized one) as redundant or no longer pinning anything. Not restructuring the file further for the metric's sake.

apps/pipelines/views.py
```diff
@@ -387,7 +389,8 @@ def pipeline_data(request, team_slug: str, pk: int):
         return _handle_pipeline_patch(request, pk, team_slug)
 
     try:
-        pipeline = Pipeline.objects.get(pk=pk)
+        # flow_data below rebuilds every node from its row, reading the collection_indexes M2M.
+        pipeline = Pipeline.objects.prefetch_related("node_set__collection_indexes").get(pk=pk)
```
snopoke: this should also filter on team

SmittieC: https://github.com/dimagi/open-chat-studio/pull/4103/changes/a790f4a920cebc05b21decd7610d3d7069b58ea4

apps/pipelines/views.py
```diff
@@ -387,7 +389,8 @@ def pipeline_data(request, team_slug: str, pk: int):
         return _handle_pipeline_patch(request, pk, team_slug)
 
     try:
-        pipeline = Pipeline.objects.get(pk=pk)
+        # flow_data below rebuilds every node from its row, reading the collection_indexes M2M.
+        pipeline = Pipeline.objects.prefetch_related("node_set__collection_indexes").get(pk=pk, team=request.team)
     except Pipeline.DoesNotExist:
         pipeline = Pipeline.objects.create(id=pk, team=request.team, data={"edges": []}, name="New Pipeline")
```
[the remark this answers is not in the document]
snopoke: @SmittieC this sees like a genuine issue. I'm also not sure why we would create one here at all.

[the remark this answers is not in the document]
SmittieC: Interesting. See https://github.com/dimagi/open-chat-studio/pull/4103/changes/d24d87cde0a8b0828bee837ec5f9d3ea29329804.

snopoke (approved):

PR #4106, opened 2026-08-05
fix: stop empty messages poisoning a session's history

snopoke:
### Product Description
A participant who sends an empty message (a blank Connect message, a transcript that came back silent) permanently breaks that conversation — the bot answers the empty message fine, then every subsequent message in the session fails.

### Technical Description
Sentry shows `anthropic.AnthropicError: 400 messages: text content blocks must contain non-whitespace text` from `handle_commcare_connect_message`.

The placeholder for an empty turn was applied only at call time, in `format_multimodal_input`, and never persisted. So the `ChatMessage` row kept the raw empty string: the first turn worked, and from then on history replayed that row as `{"type": "text", "text": ""}`.

Fixed on both sides — write, so no new bad rows appear; read, so the sessions already broken in production recover without a backfill. Attachment-only messages deliberately keep their empty text on the write side, since their content is in the attachments and the UI renders those; the read side still substitutes for them.

Worth a look:

- **The read path is human-messages-only.** An empty *AI* row can still poison a session: `_get_final_ai_message` falls back to `messages[-1]` on a pure tool-call chain, whose `.text` is empty, and that gets persisted. Left out on purpose — putting placeholder text into an assistant turn is a product call, not a bug fix. Happy to widen it if you'd rather.
- **The `EvaluationMessage` hunk guards dead code.** `as_langchain_messages` has no production callers (eval generation goes through `ChatMessageCreationStage` instead). Included because it's the same defect one line away from being live; say the word and I'll drop it to keep the diff focused.
- `apps/experiments/tasks.py:219` has the same unguarded pattern in prompt-builder history. Different feature, different data source, left alone.

### Migrations
N/A — no schema or data changes. Existing empty rows are handled on read.

### Demo
Reproduces without a live provider: send an empty message to a Connect-channel bot, then send a second message. Every new assertion was confirmed to fail against unmodified source and pass after, and the "normal content untouched" cases pass either way so the substitution isn't over-reaching.

### Docs and Changelog

Changelog note: fixes empty participant messages breaking the rest of a conversation.

### DIFF
--- a/apps/channels/stages/core.py
+++ b/apps/channels/stages/core.py
@@ -36,6 +36,7 @@ from apps.ocs_notifications.notifications import (
 from apps.service_providers.llm_service.history_managers import ExperimentHistoryManager
 from apps.service_providers.tracing import TraceInfo
 from apps.service_providers.tracing.base import SpanNotificationConfig
+from apps.utils.llm_messages import EMPTY_MESSAGE_PLACEHOLDER
 
 if TYPE_CHECKING:
     from apps.users.models import CustomUser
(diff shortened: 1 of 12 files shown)


apps/channels/stages/core.py
```diff
@@ -579,6 +580,14 @@ def process(self, ctx: MessageProcessingContext) -> None:
...
             metadata.update(ctx.trace_service.get_trace_metadata())
 
+        # A message with nothing in it at all is sent to the LLM as a placeholder
+        # (see `format_multimodal_input`), so store that same placeholder -- persisting the
+        # raw empty string makes every later turn replay an empty text content block, which
+        # Anthropic rejects. Attachment-only messages keep their empty text: the content is
+        # in the attachments, which the UI renders.
+        if not ctx.user_query.strip() and not metadata[attachments_key]:
+            ctx.user_query = EMPTY_MESSAGE_PLACEHOLDER
+
```
snopoke, line 568: Fixed in bd7c301d2 — extracted the voice-note attachment save out of `ChatMessageCreationStage.process` into a `_save_voice_note` helper, which brings the method back under the complexity threshold.

*— Claude Code*

Issue #4107, opened 2026-08-05
sync_all_document_sources_task uses Task.map(), which runs sequentially in-process


snopoke:
## Problem

\`sync_all_document_sources_task\` (\`apps/documents/tasks.py\`) dispatches per-source syncs via:

```python
sync_document_source_task.map(auto_sources).apply_async(queue=Queues.BACKGROUND)
```

Celery's \`Task.map()\` is not a parallel fan-out. Its builtin implementation (\`celery/app/builtins.py\`):

```python
@app.task(name='celery.map', shared=False, lazy=False)
def xmap(task, it):
    task = signature(task, app=app).type
    return [task(item) for item in it]
```

calls the task directly (\`task(item)\`, not \`.delay()\`/\`.apply_async()\`). One \`celery.map\` message is sent, and a single worker thread then loops through every stale document source **serially, in-process**, inside that one task execution. Celery's own docs confirm this: "Tasks executed sequentially in process, this is not a parallel operation like group."

Consequences:

* **No parallelism** across the worker pool for this job, regardless of \`--concurrency\`/\`--pool threads\`.
* **No isolation between sources** — it's a plain list comprehension, so one source raising an exception aborts the sync for every source after it in that week's run.
* **No real per-item task** — \`sync_document_source_task\`'s own retry, timeout, and \`self.request.id\`-based lock (used for the concurrent-sync guard) only work by accident, since there's no dispatched message backing each item.

\`apps/documents/views.py\` already dispatches the same task correctly for the manual "sync now" action:

```python
sync_document_source_task.apply_async(args=[document_source.id], task_id=task_id)
```

## Proposed fix

Replace the \`.map()\` call in \`sync_all_document_sources_task\` with individual dispatches per source (a loop of \`.apply_async()\` calls, or a \`group(...)\` of signatures), mirroring the existing manual-sync call site, so the weekly batch gets real parallelism, per-source isolation, and working retry/lock semantics.

## Context

Found during code review of #4022 (Celery queue split). The queue-routing symptom (this call silently falling through to the chat queue instead of \`background\`) was already fixed in that PR by passing \`queue=Queues.BACKGROUND\` explicitly — but that only fixes where the single wrapping message goes, not the underlying execution model.

snopoke:
@claude please implement this

PR #4108, opened 2026-08-05
Split Celery into chat, background, and evaluations queues

### LINKED ISSUE #4022: Split Celery into separate queues: chatbot execution vs background/long-running tasks
## Problem

All Celery tasks run in a single default queue, meaning heavyweight background work (evaluations, document indexing, exports) competes directly with user-facing chatbot message handling. A slow or backed-up evaluation run can starve inbound messages of workers, increasing response latency for end users.

## Proposed Solution

Split into **2 queues** (with a possible 3rd — see below):

### Queue 1: `default` (existing name, no migration pain)
Real-time, user-facing, latency-sensitive tasks. Keep these on the existing `celery` default queue so no code changes are needed for the happy path.

| App | Tasks |
|-----|-------|
| `channels` | `handle_telegram_message`, `handle_twilio_message`, `handle_turn_message`, `handle_email_message`, `handle_commcare_connect_message`, `handle_meta_cloud_api_message`, `handle_sureadhere_message`, `ratchet_widget_auth_levels` |
| `chatbots` | `send_bot_message` |
| `events` | `fire_static_trigger`, `fire_trigger`, `retry_scheduled_message`, `poll_scheduled_messages`, `enqueue_static_triggers`, `enqueue_timed_out_events` |
| `api` | `trigger_bot_message_task`, `setup_connect_channels_for_bots` |
| `pipelines` | `get_response_for_pipeline_test_message`, `send_email_from_pipeline` |

### Queue 2: `background`
Long-running, resource-intensive, or user-initiated async work. Annotate these tasks with `queue='background'`.

| App | Tasks |
|-----|-------|
| `evaluations` | All evaluation tasks — `evaluate_message`, `evaluate_message_batch`, `coordinate_evaluation_runs`, `run_bot_generation`, `update_dataset_from_csv_task`, `create_dataset_from_csv_task`, `upload_evaluation_run_results_task`, `create_dataset_from_session_messages_task`, cleanup tasks |
| `documents` | `index_collection_files_task`, `sync_document_source_task`, `sync_all_document_sources_task`, `async_create_collection_version`, `create_collection_zip_task`, `create_collection_from_assistant_task`, `migrate_vector_stores` |
| `experiments` | `async_export_chat`, `async_create_experiment_version` |
| `analysis` | `process_transcript_analysis` |
| `teams` | `delete_team_async`, `create_team_files_zip_task` |
| `assistants` | `delete_openai_assistant_task` |
| `mcp_integrations` | `sync_tools_task` |
| `custom_actions` | `check_all_custom_actions_health`, `check_single_custom_action_health` |
| `files` / `web` / `dashboard` / `ocs_notifications` | Periodic cleanup tasks |

### Optional Queue 3: `evaluations`
Evals are by far the largest workload — they can run hundreds of LLM calls per run with dedicated concurrency. A dedicated `evaluations` queue would let us tune worker concurrency independently (e.g. more threads for evals, fewer but more reliable workers for general background work). Start with 2 queues and promote to 3 if evals still saturate the `background` queue.

### Task annotation
Add `queue='background'` to all tasks in the second group:
```python
@shared_task(bind=True, base=TaskbadgerTask, queue='background')
def evaluate_message_batch(...):
    ...
```

### Worker startup
Run two workers pointing at different queues (`Procfile.dev` + production config):
```
worker: celery -A config worker -Q default -l INFO
background-worker: celery -A config worker -Q background -c 4 -l INFO
```

The concurrency on `background` can be tuned separately — eval tasks are I/O-bound (LLM calls), so higher concurrency with `--pool threads` may be worth exploring there.

### Healthcheck
`HEALTHCHECK_CELERY_QUEUE_TIMEOUT` in `settings.py` currently only checks a single queue. The healthcheck logic should be updated to probe both queues.

### Periodic/beat tasks
Beat-scheduled tasks go on whichever queue makes sense. Most cleanup tasks are fine on `background`; event polling (`poll_scheduled_messages`, `enqueue_timed_out_events`) should stay on `default` since they fan out latency-sensitive work.

## Open Questions

- Should evals get their own queue (Option B above)?
- What concurrency/pool settings should the background worker use in production? The eval tasks already have a `BATCH_SOFT_TIME_LIMIT = 240s` guard — we should also consider a `time_limit` on the background worker.
- Does `ocs-deploy` need changes to spin up a second ECS Fargate task for the background worker?
- Should we add the `background` queue to Taskbadger/Sentry monitoring?

snopoke:
### Product Description
No user-facing change. Background/evaluation load (indexing, exports, eval fan-out) can no longer delay inbound chat message handling.

### Technical Description
Closes #4022. Every task now declares an explicit `queue=Queues.{CHAT,BACKGROUND,EVALUATIONS}`, enforced by a new guard test (`apps/utils/tests/test_celery_queues.py`) that fails if any task under `apps/` omits it — an unannotated task defaults to the chat queue, which is exactly the regression this split exists to prevent.

Rollout is backwards compatible by construction: `CELERY_TASK_QUEUES` declares all three queues, and a worker started without `-Q` consumes all of them (verified live — see commit `61d11f91f`). So this PR alone changes nothing operationally; local dev, Docker, Heroku, and Kamal keep running a single worker. Splitting into dedicated workers is a separate ocs-deploy change with no ordering dependency on this one.

Worth a closer look:
- `evaluate_message_batch` (the fan-out unit) is the only task on `Queues.EVALUATIONS`. `coordinate_evaluation_runs`/`drive_evaluation_run`/`finalize_evaluation_run` stay on `Queues.BACKGROUND` deliberately — they're the control plane that dispatches and drains the eval queue, so putting them on the queue they control would let a saturated queue block its own drain.
- `sync_document_source_task.map(...)` was silently falling through to the chat queue — `Task.map()` wraps calls in Celery's builtin `celery.map` task, which doesn't inherit the mapped task's `queue=` decorator. Fixed by passing `queue=` explicitly on the `.map(...).apply_async()` call. `.map()`'s sequential-in-process execution model is a separate, pre-existing issue tracked at #4107.
- Also dropped `bind=True`/`base=TaskbadgerTask` from several tasks that never referenced `self` — Taskbadger auto-tracks every task by name regardless of base class, so the annotation was dead weight, not a correctness fix.

### Migrations
N/A — no migrations.

### Demo
N/A — no user-facing behavior change; verified via live worker/beat/healthcheck testing described in commit `61d11f91f`.

### Docs and Changelog

Hosting docs (`docs/hosting/index.md`, `heroku.md`, `kamal.md`, `docs/getting-started/docker-setup.md`, `local-setup.md`) updated in this PR to describe the queues and the optional worker-split scaling step.

### DIFF
--- a/apps/utils/celery.py
+++ b/apps/utils/celery.py
@@ -1,3 +1,34 @@
+from enum import StrEnum
+
+
+class Queues(StrEnum):
+    """The Celery queues tasks are routed to.
+
+    Every task defined under ``apps/`` must declare one explicitly via ``@shared_task(queue=...)``;
+    ``apps/utils/tests/test_celery_queues.py`` fails the build if one doesn't. The point of the
+    split is that a backed-up evaluation run can never starve inbound chat messages of workers.
+
+    A worker started without ``-Q`` consumes every queue in ``settings.CELERY_TASK_QUEUES``, so the
+    single-worker setups (local dev, docker compose, Heroku, Kamal, self-hosters) keep working
+    unchanged. Only production runs a dedicated worker per queue.
+    """
+
+    #: Latency-sensitive chat path: inbound message handlers, event triggers, outbound bot
+    #: messages. Named "celery" because it is also ``task_default_queue`` — third-party and
+    #: built-in tasks (e.g. ``celery.backend_cleanup``) fall through to it, and keeping the
+    #: default name means nothing enqueued before this split was introduced was stranded.
+    CHAT = "celery"
+
+    #: Long-running or resource-intensive work: indexing, exports, imports, cleanup crons, and
+    #: the evaluation control plane.
+    BACKGROUND = "background"
+
+    #: Evaluation fan-out only. Deliberately excludes the tasks that dispatch and drain it
+    #: (``coordinate_evaluation_runs``, ``drive_evaluation_run``, ``finalize_evaluation_run``) so
+    #: a saturated queue can never block its own drain.
+    EVALUATIONS = "evaluations"
+
+
 class TaskbadgerTaskWrapper:
     """Wrapper for Celery tasks to provide progress reporting via taskbadger.
 
--- a/apps/utils/tests/test_celery_queues.py
+++ b/apps/utils/tests/test_celery_queues.py
@@ -49,6 +49,20 @@ def _declared_queue(decorator: ast.expr) -> str | None:
     return None
 
 
+def _tasks_in_module(path: Path) -> list[tuple[str, str, str | None]]:
+    """Every task-decorated function defined in a single module."""
+    found = []
+    tree = ast.parse(path.read_text(encoding="utf-8"))
+    for node in ast.walk(tree):
+        if not isinstance(node, ast.FunctionDef | ast.AsyncFunctionDef):
+            continue
+        for decorator in node.decorator_list:
+            if _is_task_decorator(decorator):
+                location = f"{path.relative_to(APPS_DIR.parent)}:{node.lineno}"
+                found.append((location, node.name, _declared_queue(decorator)))
+    return found
+
+
 @cache
 def _source_tasks() -> list[tuple[str, str, str | None]]:
     """Every task defined under ``apps/`` as ``(location, function name, declared queue)``.
@@ -60,14 +74,7 @@ def _source_tasks() -> list[tuple[str, str, str | None]]:
     for path in sorted(APPS_DIR.rglob("*.py")):
         if "migrations" in path.parts:
             continue
-        tree = ast.parse(path.read_text(encoding="utf-8"))
-        for node in ast.walk(tree):
-            if not isinstance(node, ast.FunctionDef | ast.AsyncFunctionDef):
-                continue
-            for decorator in node.decorator_list:
-                if _is_task_decorator(decorator):
-                    location = f"{path.relative_to(APPS_DIR.parent)}:{node.lineno}"
-                    found.append((location, node.name, _declared_queue(decorator)))
+        found.extend(_tasks_in_module(path))
     return found
 
 
--- a/apps/utils/tests/test_celery_queues.py
+++ b/apps/utils/tests/test_celery_queues.py
@@ -49,6 +49,11 @@ def _declared_queue(decorator: ast.expr) -> str | None:
     return None
 
 
+def _task_decorator(node: ast.FunctionDef | ast.AsyncFunctionDef) -> ast.expr | None:
+    """The task decorator on this function definition, if it has one."""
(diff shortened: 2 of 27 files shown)


apps/utils/tests/test_celery_queues.py
```diff
@@ -0,0 +1,153 @@
...
+            continue
+        tree = ast.parse(path.read_text(encoding="utf-8"))
+        for node in ast.walk(tree):
+            if not isinstance(node, ast.FunctionDef | ast.AsyncFunctionDef):
+                continue
+            for decorator in node.decorator_list:
+                if _is_task_decorator(decorator):
+                    location = f"{path.relative_to(APPS_DIR.parent)}:{node.lineno}"
+                    found.append((location, node.name, _declared_queue(decorator)))
+    return found
```
snopoke, line 84: Fixed in 43bd25dea — extracted the per-file AST parsing into `_tasks_in_module`, so `_source_tasks` itself only has the single `if "migrations" in path.parts` guard.

*— Claude Code*

snopoke, line 84: Fixed in 43bd25dea — same extraction; splitting the per-file parsing into `_tasks_in_module` brings nesting depth down to 3 in each function.

*— Claude Code*

config/settings.py
```diff
@@ -19,8 +19,11 @@
 import environ
 from celery.schedules import crontab
 from django.utils.translation import gettext_lazy
+from kombu import Queue
```
snopoke, line 22: Fixed in 140dd32bb — moved the `Queue(...)` tuple-comprehension into a `_celery_task_queues()` helper (matching the existing `_public_media_url` pattern in this file), so the branch-carrying construct moves out of global scope.

*— Claude Code*

apps/utils/celery.py
```diff
@@ -1,3 +1,34 @@
...
+class Queues(StrEnum):
+    """The Celery queues tasks are routed to.
+
+    Every task defined under ``apps/`` must declare one explicitly via ``@shared_task(queue=...)``;
+    ``apps/utils/tests/test_celery_queues.py`` fails the build if one doesn't. The point of the
+    split is that a backed-up evaluation run can never starve inbound chat messages of workers.
+
+    A worker started without ``-Q`` consumes every queue in ``settings.CELERY_TASK_QUEUES``, so the
+    single-worker setups (local dev, docker compose, Heroku, Kamal, self-hosters) keep working
+    unchanged. Only production runs a dedicated worker per queue.
```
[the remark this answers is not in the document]
snopoke: Noted — declined by author (unattended run treats low-priority suggestions as declined). The docstring already says "Only production runs a dedicated worker per queue" and the PR description calls out the worker-split as a separate, unordered ocs-deploy change, so the current wording is intentional rather than an oversight.

*— Claude Code*

SmittieC (approved):

PR #4113, opened 2026-08-05
Store synced documents as source bytes and show indexing progress

SmittieC:
### Product Description

Two changes to collections backed by a document source.

**A synced file is now the file the source served.** Downloading a synced PDF gives you a PDF that opens. Previously the sync extracted text at fetch time and stored *that* under the source's filename, so a document called `report.pdf` was really a text file.

**The collection page shows aggregate indexing progress.** An "X of Y files indexed" line above the file list, with failures counted separately, so a large sync is visible without reading every per-file badge. It polls while work is outstanding and goes quiet once everything settles.

### Technical Description

**Source bytes, sniffed content type** (ADR-0051)

Loaders yield `SourceDocument(content: bytes, metadata: dict)` instead of a LangChain `Document`. The bytes are written verbatim and `File.create` sniffs the content type from them, ignoring the type the source claimed. `Document.from_file` already picks a reader off `File.content_type`, so extraction moves to index time. Where a source has no file to serve (Confluence pages, GitHub's decoded contents) the loader encodes the text it was given — the invariant is "the rawest representation the source offers", not "always binary".

A file that yields no text at index time is recorded FAILED rather than COMPLETED, since a file with no embeddings otherwise looks identical to one that indexed cleanly. The remote manager only does this when the local read *succeeded and returned nothing* — a read that raises counts as unknown, so a format our readers can't parse but the provider can is still handed to the provider.

Tradeoffs, all recorded in the ADR:

- Attachments are persisted in full, up to the existing 50 MB fetch cap. Storage for a PDF-heavy feed grows from extracted text to whole documents.
- Parse cost moves from the sync to the indexing worker, and re-indexing re-parses.
- The remote path reads each file locally before upload — work the provider would otherwise do, accepted because a provider that indexes a file as empty reports nothing wrong.
- Rows synced before this change are **not** backfilled; they are only rewritten when their `sha`/`date` changes. Repairing them needs a forced re-sync, not in this PR.

**Indexing progress**

A partial polled every 5s while work is outstanding, dropping its own trigger once nothing is pending. Blank-status `CollectionFile` rows are excluded from every count — they never went through indexing, and a local index version is full of them since versioning copies chunks across wholesale. Outstanding work is counted positively from `PENDING`/`IN_PROGRESS` rather than derived from the total. The line uses `role="progressbar"` rather than a live region, because htmx replaces the element itself on each poll.

**Also in this branch**

- `ensure_extension` extracted from `File.create` and reused, so a synced file with no extension no longer gains one on create and loses it on the next update.
- `_update_file` clears `File.external_id`, so an update re-uploads instead of leaving the previous bytes indexed remotely under a new content type.

### Migrations

No migrations in this PR.

### Demo
We show the indexed progress at the top also

image

### Docs and Changelog

ADR-0051 added at `docs/adr/0051-store-source-bytes-extract-at-index-time.md`, wired into `docs/adr/index.md` and the mkdocs nav.


apps/service_providers/llm_service/index_managers.py
```diff
@@ -283,6 +317,12 @@ def add_files(
             embeddings = []
             try:
                 document_text = file.read_content()
+                if not document_text.strip():
+                    # Source files are stored unparsed (ADR-0051), so a scanned or image-only
+                    # document reaches indexing and yields nothing. Without this, it would be
+                    # marked COMPLETED with no embeddings -- indistinguishable from a file that
+                    # actually indexed.
+                    raise FileReadException(NO_EXTRACTABLE_TEXT)
```
SmittieC: Fixed in d707f3e23. Extracted the per-file body of `LocalIndexManager.add_files` into `_embed_file`, which chunks, embeds and persists one file and cleans up its own partial writes on failure. `add_files` is now just the loop, the status transition and the save, which brings it back under the threshold.

apps/documents/document_source_service.py
```diff
@@ -218,18 +218,15 @@ def _map_existing_files(self) -> dict[str, CollectionFile]:
...
         A single bad document must not abort the whole sync: log it, record it, and
         carry on so the remaining files are still processed and indexed.
         """
-        if not document.page_content.strip():
-            msg = (
-                "Skipping document with empty content "
-                "(file may be a scanned/image-based document with no extractable text)"
-            )
+        if not document.content.strip():
+            msg = "Skipping document with empty content (the source served no bytes)"
```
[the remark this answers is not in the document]
SmittieC: Fixed in d707f3e23. Changed to `if not document.content:` — the skip now applies only when the source served nothing at all. Whitespace-only bytes are stored (per ADR-0051) and indexing records them as failed via `NO_EXTRACTABLE_TEXT`. Added `test_whitespace_only_payload_is_stored_not_skipped` to cover the stale-content case you described.

apps/service_providers/llm_service/index_managers.py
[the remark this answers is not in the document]
SmittieC: Fixed in d707f3e23. Rather than normalizing NUL bytes before the check, `_embed_file` now raises `FileReadException(NO_EXTRACTABLE_TEXT)` when the chunk loop produces no embeddings at all. That covers NUL-only content and any other case where every chunk sanitizes away, and keeps the existing per-chunk skip (a file with a mix of NUL-only and real chunks still indexes the real ones). Regression test: `test_add_files_fails_when_every_chunk_sanitizes_away`.

templates/documents/partials/indexing_progress.html
```diff
@@ -0,0 +1,38 @@
...
+      announced at all -- so announcements would be unreliable, and the ones that did fire
+      would repeat the same unchanged count every 5s. A progressbar carries its value in
+      aria-valuenow/aria-valuetext, which assistive tech reports on demand instead.
+    {% endcomment %}
+    <div class="flex items-center gap-2 text-sm text-base-content/70"
+         role="progressbar"
+         aria-valuemin="0"
+         aria-valuemax="{{ progress_total }}"
+         aria-valuenow="{{ progress_indexed }}"
+         aria-valuetext="{{ progress_indexed }} of {{ progress_total }} files indexed{% if progress_failed %}, {{ progress_failed }} failed{% endif %}">
```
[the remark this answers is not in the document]
SmittieC: Fixed in d707f3e23. Added `aria-label="File indexing progress"` to the progressbar.

apps/documents/tests/test_views.py
```diff
@@ -760,3 +760,104 @@ def test_aria_label_is_bounded_while_data_tip_carries_the_full_reason(self, coll
         assert f'data-tip="{long_reason}"' in content
         assert f'aria-label="{long_reason}"' not in content
         assert 'aria-label="AuthenticationError: ' in content
+
```
SmittieC, line 763: Not acting on this one. `test_views.py` is a test module organised by view under test, so its "responsibilities" count tracks the number of views the app exposes rather than any coupling in the code. This PR adds tests for the new indexing-progress endpoint alongside the existing ones; splitting the module to satisfy the metric would scatter shared fixtures without making anything easier to change.

docs/adr/0051-store-source-bytes-extract-at-index-time.md
```diff
@@ -0,0 +1,75 @@
+# ADR-0051: Store the bytes a source served; extract text at index time
+
+<span class="adr-status adr-status-accepted">ACCEPTED</span>
+
+<p class="adr-meta">Author: Chris Smit · Created: 2026-08-05</p>
```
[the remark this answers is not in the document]
SmittieC: Not acting on this one — the guideline is being misapplied. The `status: active|stable|extracted` frontmatter in AGENTS.md governs *design and spec* docs, marking when their decisions are settled enough to extract into ADRs. ADRs are the extraction target, not a source: they carry their own state in the `adr-status` badge (ACCEPTED here) and in the `docs/adr/index.md` table. Neither `docs/adr/_template.md` nor any of the 51 existing ADRs has frontmatter, so adding it here would make this file the odd one out.

templates/documents/partials/indexing_progress.html
```diff
@@ -0,0 +1,39 @@
...
+{% endcomment %}
+<div id="collection-indexing-progress"
+     {% if progress_outstanding %}
+       hx-get="{% url 'documents:collection_indexing_progress' request.team.slug collection.id %}"
+       hx-trigger="every 5s"
+       hx-target="this"
+       hx-swap="outerHTML"
+     {% endif %}>
+  {% if progress_total %}
+    {% comment %}
```
SmittieC: remove these comments

apps/documents/document_source_service.py
```diff
@@ -294,15 +299,30 @@ def _create_file(self, document: Document, identifier: str):
...
         existing_file.file = content_file
         existing_file.content_size = content_file.size
+        existing_file.content_type = content_type
         existing_file.metadata = document.metadata
+        # Whatever was uploaded to a remote index describes the old bytes -- a different
+        # format entirely, now that the source's own file is what gets stored. Clearing the
+        # id makes the next index run re-upload rather than trust the stale copy.
+        existing_file.external_id = ""
+        existing_file.external_source = ""
         existing_file.save()
```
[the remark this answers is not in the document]
SmittieC: Agreed, and fixed: `_update_file` no longer touches `external_id`/`external_source`. The uploaded copy keeps its only handle, so `delete_files_from_index`/`delete_files` can still reach it.

I didn't add the delete-then-reupload step. The gap it would close — a remote index keeps serving the old bytes until something re-uploads — predates this PR (`_update_file` preserved the id before too), so it's recorded as a known consequence in ADR-0051 rather than fixed here.

`test_update_discards_the_remote_copy_of_the_old_bytes` is now `test_update_keeps_the_remote_file_handle` and asserts the id and source survive an update.

apps/documents/source_loaders/github.py
```diff
@@ -72,15 +102,33 @@ def _load_blob(self, loader: GithubFileLoader, entry: dict) -> SourceDocument |
...
+            # abort the whole sync over a single file, which is what this loader avoids.
+            logger.warning(
+                "Skipping %s: it outgrew the %s byte limit for a collection file mid-download", path, max_bytes
+            )
+            return None
+
+        # An empty file is ordinary in a repo -- a .gitkeep, an empty __init__.py -- but the
+        # sync counts empty content as a per-file failure, which would leave the source
+        # reporting a sync error on every run. There is nothing to index either way.
+        if not content:
```
snopoke, line 128: Does this not have the same symptom as a previous commit fixed: a file that had content and then became empty remains stale?

apps/documents/views.py
```diff
@@ -100,10 +100,14 @@ def single_collection_home(request, team_slug: str, pk: int):
     document_sources = (
         DocumentSource.objects.working_versions_queryset().filter(collection=collection).prefetch_related("sync_logs")
     )
-    file_stats = CollectionFile.objects.filter(collection=collection).aggregate(
-        total=Count("id"),
-        manual=Count("id", filter=Q(document_source__isnull=True)),
-        unsendable=Count("id", filter=~Q(file__unsupported_channels={})),
+    file_stats = (
+        CollectionFile.objects.filter(collection=collection)
+        .exclude(document_source__is_archived=True)
```
snopoke, line 105: should this also exclude where `status=""` as is done elsewhere?

snopoke (commented):
I see you dropped the ADR, was it not useful?

SmittieC:
> I see you dropped the ADR, was it not useful?

I didn't think it was. I'd be happy to restore it if you think it is?

apps/documents/views.py
SmittieC: > as is done elsewhere?

You're referring to what `_indexing_progress` does? In which, no. `file_stats` shows stats about the files being synced whereas the `status=""` is needed for indexing stats.

apps/documents/source_loaders/github.py
SmittieC: No — this skip happens before the document is yielded, so the identifier never reaches `seen_identifiers` and `_remove_stale_files` deletes the already-synced copy rather than leaving it stale. Same for the two size skips above it: the file is dropped from the collection (counted as removed) while it's still in the repo.

snopoke:
> > I see you dropped the ADR, was it not useful?
> 
> I didn't think it was. I'd be happy to restore it if you think it is?

Probably fine to leave it.

snopoke (approved):

PR #4129, opened 2026-08-06
Remove unused postcss devDependencies

snopoke:
### Product Description
No change.

### Technical Description
`postcss.config.js` lists only `@tailwindcss/postcss`, so none of these three plugins were running. `autoprefixer` and `postcss-import` were orphaned by the Tailwind v4 migration in b0d222fb, which replaced the whole v3 plugin chain; `postcss-nested` was never referenced at all, going back to the initial repo migration. Tailwind v4 handles nesting, imports and vendor prefixing internally.

Mainly worth doing to stop the recurring Dependabot stream — #4126, #4074, #817 were all bumps of a package nothing loaded. Closing #4126 as obsoleted rather than merging it.

One thing worth a look: `components/chat_widget/postcss.config.js` does `require('postcss-import')` and `require('autoprefixer')`, and Node resolution would walk up to the root `node_modules`. It doesn't here — the widget declares both in its own `package.json` and `require.resolve` from that directory returns `components/chat_widget/node_modules/...` both before and after this change. Not a pnpm workspace either (`pnpm-workspace.yaml` only sets `minimumReleaseAge`), so the widget install is independent.

`postcss` and `postcss-loader` stay — `postcss-loader` is in the webpack rule chain and `postcss` is its peer.

### Migrations
N/A

### Verification
Built with and without the three deps and compared emitted CSS: all six files (`site-tailwind`, `site-base`, `pipeline`, `editors`, `dashboard`, `jsonEditor`) byte-identical, in both `pnpm run dev` and `NODE_ENV=production pnpm run build`. `vitest run` green.

### Docs and Changelog


SmittieC (approved):
Sounds good to me

PR #4131, opened 2026-08-06
docs: design and ADRs for deploy-safe document source syncs

SmittieC:
### Product Description

Docs only — no product change. This is the design for making document source syncs survive a deploy, an OOM kill, or a worker recycle: today a sync killed mid-run strands files un-indexed, leaves its sync log reporting "in progress" forever with no error in the UI, and re-downloads the whole source on the next run.

### Technical Description

Adds the design spec and the four ADRs extracted from it:

- `docs/superpowers/specs/2026-08-05-deploy-safe-document-syncs-design.md` — the design. Also available as a [Google Doc](https://docs.google.com/document/d/1h5zPMda7f3pNyo5LG2DXdEq7nL7iqEj4k0zDb3zAE8g/edit?tab=t.0) (identical content, easier to comment on).
- ADR-0051 — a beat coordinator owns all index dispatch
- ADR-0052 — manifest-first change detection
- ADR-0053 — detect an interrupted sync from its log's heartbeat
- ADR-0054 — a stale in-progress log is the restart ticket

All four ADRs are `PROPOSED`; the implementation ships in three independent phases described in the design. `docs/adr/index.md` and the mkdocs nav are updated.

### Migrations

No migrations in this PR. The design records the ones the implementation will need, including a partial unique constraint on `CollectionFile(document_source, external_id)` that needs a duplicate cleanup before it can apply.

### Demo

n/a

### Docs and Changelog

This PR *is* the docs update.

### DIFF
--- a/docs/adr/0054-stale-in-progress-log-is-the-restart-ticket.md
+++ b/docs/adr/0054-stale-in-progress-log-is-the-restart-ticket.md
@@ -1,0 +1,49 @@
+# ADR-0054: A stale in-progress log is the restart ticket
+
+<span class="adr-status adr-status-proposed">PROPOSED</span>
+
+<p class="adr-meta">Author: Open Chat Studio · Created: 2026-08-05</p>
+
+<p class="adr-meta">Extends: <a href="0053-detect-interrupted-sync-with-heartbeat.md">ADR-0053</a></p>
+
+## Context
+
+A sync killed mid-run leaves its `DocumentSourceSyncLog` at `IN_PROGRESS` forever, because the error handler never runs, and nothing re-dispatches it. A source with auto-sync waits up to a week for the next scheduled run; a manual-only source waits indefinitely.
+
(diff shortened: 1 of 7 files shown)


docs/adr/0054-stale-in-progress-log-is-the-restart-ticket.md
snopoke, line 1: one thing I'd add here is that `interrupted_sync_attempts` should be reset as soon as the task makes any real progress i.e. after the first successful index. 

The reason for this is to allow tasks that are making real progress from continuing indefinitely rather than ending them early just because they have been interrupted a few times.

SmittieC:
Closing this, since this was only a placeholder for the [Google Doc spec](https://docs.google.com/document/d/1h5zPMda7f3pNyo5LG2DXdEq7nL7iqEj4k0zDb3zAE8g/edit?tab=t.0)

PR #4149, opened 2026-08-07
Write API Phase 1 — 3: Discovery endpoints

### LINKED ISSUE #4138: Write API Phase 1 — 3: Discovery endpoints

## Scope

`GET /api/v2/pipeline/nodes/` and `GET /api/v2/pipeline/options/` (spec §5.3). Read-only, team-level; independent of the other tickets.

## Deliverables

- Extract the builder's private helpers — `_pipeline_node_schemas`, `_pipeline_node_parameter_values`, `_pipeline_node_default_values` (`apps/pipelines/views.py:218–358`) — into a shared module both the builder view and the API import (spec impl note: "a production build should extract to a shared module").
- `/pipeline/nodes/`: `[{type, description, schema}]` from `model_json_schema()`; `?type=` filter, unknown type → 404. Global/static/cacheable.
- `/pipeline/options/`: team-scoped valid ids per param + `default_values`; decouple the per-experiment `synthetic_voices` arg via `SyntheticVoice.get_for_team(team, [])`.

## Files

New shared module in `apps/pipelines/`; `apps/pipelines/views.py` (delegate to it); new views + routes in `apps/api/v2/`; tests.

## Depends on

Nothing.

## Test focus

Options are team-scoped (another team's providers absent); `?type=` filter and 404; builder view behaviour unchanged after the extraction (regression test on the existing builder context).

SmittieC:
Closes #4138. Part of the Phase 1 write API.

### Product Description

Two read-only endpoints so a client can discover what it may build:

* `GET /api/v2/pipeline/nodes/` — the 8 node types a pipeline may contain, each with the JSON Schema of its `params` and its output topology. `?type=` fetches one.
* `GET /api/v2/pipeline/options/` — the values each param accepts, scoped to the API key's team. `?node_type=` narrows it to one type.

Both are team-scoped, `chatbots:read`, and work with a read-only API key.

### Technical Description

The consumer is a model with no UI and no changelog, so: anything the OpenAPI description would have to explain, the payload carries instead.

* `get_node_parameter_values` (moved to `apps/pipelines/nodes/node_metadata.py`) is now the source of truth for both the builder and the API, uniformly snake_case, with each option key named after the param that reads it.
* The payload states its own rules: `must_match`, `options_keyed_by`, `applies_when`, `requires_feature_flag`, output topology (`per_keyword` with `output_{index}` handles for routers, `single` otherwise), and a `description` on every param (was 40% coverage).
* `UiSchema(api_exclude=True)` withholds a param and its option list; `mcp_tools` is the only user. Options are a whitelist of what the listed types reference, so deprecating a type drops its option lists automatically.
* Internal `ui:`/`api:` keys are stripped; the two that carry meaning become `applies_when` and `requires_feature_flag`.
* Deprecated and server-managed types are absent, not flagged. A direct lookup 404s with `valid_types` and the reason.
* Prompt variables are keyed by prompt flavour (`template_variables`, `llm_prompt_variables`, `router_prompt_variables`) — LLM and router nodes both have a `prompt` param but accept different variable sets. This is the one documented exception to the same-name rule.
* Response schema is typed (was `additionalProperties: {}`); `/nodes/` carries an `ETag` and honours `If-None-Match`.

Every option list is scoped to what the team can actually reach:

* Synthetic voices are filtered to services the team has a voice provider for — previously 564 entries (~85% of the response) for a team with no voice provider.
* `llm_provider_model_id` is filtered to the team's LLM provider types, and deprecated models are dropped the same way deprecated node types are. Global models (`team=None`) cover all nine provider types, so a team holding only an OpenAI key was offered 88 models: 58 for providers it has no credentials to call and 5 already deprecated.
* `tool_config` is keyed by provider type, like the `built_in_tools` beside it, but was a hardcoded dict covering every type. It now carries only the team's.

The LLM scoping alone takes a single-OpenAI-provider team's full response from 15,054 to 7,945 bytes; `?node_type=` narrows it further to 5,457 (LLM node) or 1,633 (template node).

Out of scope: `_evaluator_parameter_values` in `apps/evaluations/` keeps its PascalCase keys — it's a separate hand-rolled copy, not part of the shared helper.

Sharp edge: `MUST_MATCH` and `OPTIONS_KEYED_BY` in `contract.py` are hand-maintained and keyed by param name, so renaming a param silently drops the link.

### Migrations

No migrations.

- [ ] The migrations are backwards compatible *(N/A)*

### Demo

`GET /api/v2/pipeline/options/` — 15 keys, abridged:

```json
{
  "llm_provider_id": [{ "value": 1, "label": "Prod OpenAI", "type": "openai" }],
  "source_material": [{ "value": 3, "label": "Returns policy" }],
  "voice_provider_id": [{ "value": 2, "label": "Prod Polly", "type": "aws" }],
  "llm_prompt_variables": [
    { "label": "source_material", "description": "The full text of the source material chosen in `source_material_id`. ..." }
  ],
  "router_prompt_variables": [
    { "label": "session_state", "description": "State that survives for the whole session. ..." }
  ],
  "default_llm_provider": { "llm_provider_id": 1, "llm_provider_model_id": 5 }
}
```

`GET /api/v2/pipeline/nodes/?type=StartNode` → 404

```json
{
  "detail": "Node type 'StartNode' is managed by the server and cannot be created or configured. It may appear as a node's `type` in /inspect/ responses.",
  "valid_types": ["CodeNode", "ExtractParticipantData", "ExtractStructuredData", "LLMResponseWithPrompt",
                  "RenderTemplate", "RouterNode", "SendEmail", "StaticRouterNode"]
}
```

Verified against a running server with a read-only API key: 8 types listed, no namespaced keys, `?type=` returns one element, unknown/deprecated/server-managed names each 404 with their own reason, `If-None-Match` gives a 304, unauthenticated 401. The options payload was checked against a real team's data: only the configured provider's models, no deprecated ones, no `tool_config` entry for a provider the team lacks. The pipeline builder was also driven in a browser — dropdowns, built-in tools, and resource pickers still populate.

### Docs and Changelog

`api-schemas/v2.yml` regenerated and `Pipelines` tag declared in `SPECTACULAR_SETTINGS`.


apps/pipelines/nodes/node_metadata.py
```diff
@@ -0,0 +1,172 @@
...
+        OptionsSource.text_editor_autocomplete_vars_router_node: PromptVars.get_router_prompt_vars(),
+        OptionsSource.jinja_node: PromptVars.get_jinja_vars(),
+        OptionsSource.synthetic_voice_id: sorted(
+            [
+                _option(voice.id, str(voice), voice.service.lower()) | {"provider_id": voice.voice_provider_id}
+                for voice in synthetic_voices
+            ],
+            key=lambda v: v["label"],
+        ),
+    }
```
SmittieC: Fixed in f46ec1104.

The complexity was almost entirely the fifteen comprehensions inlined into the returned dict literal (this function was moved here from `apps/pipelines/views.py` in this PR, which is why it reads as new). Split it into six helpers grouped by concern — LLM providers, team resources, tools, built-in tools, prompt variables, synthetic voices — leaving the public function as a merge of the six. `_option` moved to module level.

`get_node_parameter_values` is now complexity 1; the highest in the file is 7, against the threshold of 11.

Key order and every value are unchanged. Verified by diffing old vs. new output over a team with providers, source material, an assistant, both collection kinds and synthetic voices, across both `include_versions` settings.

*— Claude Code*

apps/api/v2/discovery.py
```diff
@@ -0,0 +1,252 @@
...
+            OpenApiParameter(
+                name="type",
+                type=OpenApiTypes.STR,
+                location=OpenApiParameter.QUERY,
+                required=False,
+                description="Return only this node type. An unknown or deprecated name returns 404.",
+            )
+        ],
+        responses={200: NodeTypeSerializer(many=True)},
+    )
```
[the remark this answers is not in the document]
SmittieC: Good catch — valid, the schema contradicted its own parameter description. Fixed in e68546c35.

Added `404: OpenApiResponse(...)` with a `NodeTypeNotFoundSerializer` for the `{"detail": ...}` body DRF renders for `NotFound`, and regenerated `api-schemas/v2.yml` via `inv schema`.

The 404 behaviour was already covered by `test_type_filter_404s`; ac0434932 extends it to assert the body carries exactly the documented `detail` key, so the schema and the response cannot drift.

*— Claude Code*

api-schemas/v2.yml
```diff
@@ -115,6 +115,104 @@ paths:
               schema:
                 $ref: '#/components/schemas/Me'
           description: ''
+  /api/v2/pipeline/nodes/:
+    get:
+      operationId: pipeline_node_list
+      description: |-
```
SmittieC, line 121: there's too much useless info in this description. Keep it to useful info only

api-schemas/v2.yml
```diff
@@ -115,6 +115,104 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NodeTypeNotFound'
+          description: No such node type, or the type is deprecated and therefore
+            not listed.
+  /api/v2/pipeline/options/:
+    get:
+      operationId: pipeline_options
+      description: |-
```
SmittieC, line 246: this is too verbost. trim down the description to useful info only

api-schemas/v2.yml
```diff
@@ -1644,6 +1742,36 @@ components:
...
+      properties:
+        type:
+          type: string
+          description: Write this to a node's `type` field.
+        description:
+          type: string
+        can_add:
+          type: boolean
+          description: False for structural nodes the server manages itself -- Start,
+            End and Passthrough. They are listed so the agent can resolve types it
```
SmittieC: Instead of shipping `can_add`, we should just return nodes that can be added in the first place

api-schemas/v2.yml
```diff
@@ -115,6 +115,104 @@ paths:
...
+      - OAuth2:
+        - chatbots:read
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                type: object
```
SmittieC: we're missing collections from the schema

api-schemas/v2.yml
SmittieC: Trimmed in 2555aa351 — 128 words to 38.

Nothing was dropped, though. A typed OpenAPI schema landed in the same commit (previously the `/options/` response was `additionalProperties: {}`), so most of this description was duplicating a field description. Each fact moved to the field it belongs on:

- the `options_source` / `must_match` / `options_keyed_by` / `applies_when` / `requires_feature_flag` glossary → `NodeType.schema`, which is the field whose contents it describes

What's left is the part with nowhere else to live: what the endpoint returns, and that it's cacheable by ETag.

*— Claude Code*

api-schemas/v2.yml
SmittieC: Trimmed in 2555aa351 — 162 words to 35.

Same dedup as the sibling thread: the typed response schema in that commit means these facts now have a home on the field itself, so the prose was saying each one twice.

- opaque composite ids → `Option.value`
- `voice_provider_id` isn't any param's option list → that property's own description
- "write the `label` into the prompt" → `PromptVariable.label`, `$ref`'d by all three prompt-var keys so it renders once instead of three times
- `?node_type=` usage → the parameter's description

Left behind: what the endpoint returns, and the `options_source` rule for resolving a param.

*— Claude Code*

api-schemas/v2.yml
SmittieC: Done in 2555aa351. `/pipeline/nodes/` now lists only the 8 addable types and `NodeType` has no `can_add` field.

I checked the design doc's counter-argument first, since it explicitly argued the other way — that filtering these would leave `"type": "StartNode"` from an `/inspect/` response unresolvable. It doesn't hold up: all three have a single fixed `name` param and nothing else, and `StartNode`/`EndNode` are `can_delete: false` and excluded from versioning via `reserved_types`. An agent that resolved one would learn nothing it could act on.

One residual worth flagging, since it's the half of that argument that *is* real: `/inspect/` does still report these as a node's `type`, so looking one up is a reasonable thing for an agent to do. `?type=StartNode` now 404s with

> Node type 'StartNode' is managed by the server and cannot be created or configured. It may appear as a node's `type` in /inspect/ responses.

rather than "Unknown node type" — the same distinction already drawn for deprecated names. `/options/?node_type=` agrees.

Fallout: every listed type is now a `pipelineNode`, so the zero-output branch of the output topology became unreachable and was deleted rather than left as dead code. That cost the `EndNode` topology test — genuinely unreachable now, not skipped.

Decision and the rejected alternative are recorded in ADR-0051 (2d46d513a).

*— Claude Code*

api-schemas/v2.yml
SmittieC: Fixed in 2555aa351.

The line you flagged is `additionalProperties: {}` — the untyped `/options/` 200 response, which is exactly why collections had nowhere to be described. They were always *in* the payload (`collection` and `collection_index`, from `_team_resource_options`); the schema just couldn't say so.

That commit replaces the untyped response with `PipelineOptionsSerializer`, so both keys are now declared and described:

- `collection` — "Media collections -- files a node can talk about."
- `collection_index` — "Searchable indexes a node can retrieve from."

Both also added to the response example, which had omitted them too.

*— Claude Code*

apps/pipelines/nodes/node_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+        and cls not in (pipeline_nodes.PipelineNode, pipeline_nodes.PipelineRouterNode)
+    ]
+    for node_class in node_classes:
+        schemas.append(_get_node_schema(node_class))
+
+    return schemas
+
+
+def _get_node_schema(node_class):
+    schema = resolve_references(node_class.model_json_schema())
```
SmittieC, line 272: since this is also now using `resolve_references`, add a schema_utils.py file in the shared app and move all methods used by multiple apps in there

apps/api/v2/discovery/views.py
```diff
@@ -0,0 +1,233 @@
...
+
+from apps.api.permissions import DjangoModelPermissionsWithView
+from apps.experiments.models import SyntheticVoice
+from apps.oauth.permissions import TokenHasOAuthResourceScope
+from apps.pipelines.models import Pipeline
+from apps.pipelines.node_options import get_node_default_values, get_node_parameter_values
+from apps.service_providers.models import LlmProvider, LlmProviderModel, VoiceProvider
+
+from .node_types import _etag, _node_types, _unknown_node_type
+from .options import _clean_options, _describe_prompt_vars, _keys_for_node_type, _rename_option_keys
```
SmittieC: Don't import private methods

apps/api/v2/discovery/options.py
```diff
@@ -0,0 +1,75 @@
+"""Reshaping the builder's option lists into what an agent reads."""
```
SmittieC: the methods in this file are all 1. private and 2. used only by PipelineOptionsView, so why not just move it to that class as private methods?

api-schemas/v2.yml
```diff
@@ -1715,6 +1984,90 @@ components:
...
+            type: array
+            items:
+              $ref: '#/components/schemas/Option'
+          description: Keyed by LLM provider type.
+        built_in_tools_config:
+          type: object
+          additionalProperties: {}
+          description: Per-provider, per-tool config field descriptors. Keyed by LLM
+            provider type.
+        text_editor_autocomplete_vars_llm_node:
```
SmittieC: This means nothing for a client. Remove this and text_editor_autocomplete_vars_router_node.

api-schemas/v2.yml
```diff
@@ -1715,6 +1984,90 @@ components:
...
+          type: array
+          items:
+            $ref: '#/components/schemas/PromptVariable'
+        text_editor_autocomplete_vars_router_node:
+          type: array
+          items:
+            $ref: '#/components/schemas/PromptVariable'
+        jinja_node:
```
SmittieC: call this "prompt_variables"

apps/api/v2/discovery/contract.py
```diff
@@ -0,0 +1,60 @@
...
+    "VoiceProviderId": "voice_provider_id",
+}
+
+# Params whose values come from an options key the builder never declared, because it renders them
+# with a bespoke widget rather than the generic `select`. Without these the join has four exceptions
+# and the agent has to guess them from naming.
+IMPLIED_OPTIONS_SOURCE = {
+    "llm_provider_id": "llm_provider_id",
+    "llm_provider_model_id": "llm_provider_model_id",
+    "tool_config": "built_in_tools_config",
```
SmittieC: Can we not just rename `built_in_tools_config` to `tool_config` and update all places where it is currently named as `built_in_tools_config`?

apps/api/v2/discovery/node_types.py
```diff
@@ -0,0 +1,157 @@
...
+    }
+    plain = {key: value for key, value in prop.items() if not key.startswith("ui:")}
+    return plain | translated | _param_links(name, prop)
+
+
+def _param_links(name: str, prop: dict) -> dict:
+    """The cross-param rules the builder enforces in JS and the schema never stated."""
+    links = {}
+    if "ui:optionsSource" not in prop and name in IMPLIED_OPTIONS_SOURCE:
+        links["options_source"] = IMPLIED_OPTIONS_SOURCE[name]
```
SmittieC: I don't think `options_source` provides much value. The option source and noed parameter keys typically look the same, so it's pretty clear which belongs to which

apps/pipelines/nodes/node_metadata.py
SmittieC: Fixed in 9df3111c7. `resolve_references` now lives in `apps/utils/schema_utils.py`, with `apps/custom_actions/tests/test_resolve_refs.py` moved alongside it as `apps/utils/tests/test_schema_utils.py`. Four apps read it — `custom_actions`, `pipelines`, `evaluations` and `service_providers`.

I left the rest of `apps/custom_actions/schema_utils.py` where it is: `APIOperationDetails`/`ParameterDetail`/`get_operations_from_spec*` are about custom-action OpenAPI operations specifically, and the only out-of-app reader is a `TYPE_CHECKING`-style local import in `apps/api/v2/inspect/serializers.py`. Say the word if you want those moved too.

apps/api/v2/discovery/views.py
SmittieC: Fixed in 15cf162dd. `node_types.py` now exposes `get_node_types()`, `etag()`, `unknown_node_type()` and `option_keys_for_node_type()` as public names; everything still private stays inside the module. `options.py` is gone entirely (see the thread below), so `views.py` no longer imports a private name from anywhere.

apps/api/v2/discovery/options.py
SmittieC: Fixed in 15cf162dd. `apps/api/v2/discovery/options.py` is deleted — `_clean_options`, `_describe_prompt_vars` and the key renaming are now private methods on `PipelineOptionsView`.

One piece did not move there: `_keys_for_node_type` is now `option_keys_for_node_type()` in `node_types.py`. It reads the builder's raw node schemas (`ui:optionsSource`) rather than the API payload, which is a node-types concern, and with `options_source` no longer emitted it is the only remaining consumer of that pairing.

api-schemas/v2.yml
SmittieC: Fixed in 15cf162dd. Both are gone from the payload (`HIDDEN_OPTION_KEYS` in `contract.py`), along with their serializer fields and the `?node_type=` scoping entries.

Worth flagging one consequence, since it is not reversible by a client: those two were the only lists naming the variables an `LLMResponseWithPrompt` or `RouterNode` prompt may reference, and the three sets genuinely differ — `{source_material}` is valid in an LLM prompt and a 400 in a router prompt. After this change those two params have no listed variables at all; what a prompt may use has to come from the param's own description. I also dropped the four `PROMPT_VAR_DESCRIPTIONS` entries that only those lists served (`source_material`, `current_datetime`, `media`, `collection_index_summaries`) and narrowed the guard test to the jinja vars, rather than leaving descriptions nothing serves. Easy to put back if you would rather keep all three lists under clearer names.

api-schemas/v2.yml
SmittieC: Fixed in 15cf162dd. `jinja_node` is renamed to `prompt_variables` on the API surface via `OPTIONS_KEY_RENAMES`; the builder still sees `jinja_node`, since its JS reads the key verbatim.

apps/api/v2/discovery/contract.py
SmittieC: Fixed in 2f370dba7. `OptionsSource.built_in_tools_config` is now `OptionsSource.tool_config`, matching the node param it feeds. Updated in `apps/pipelines/nodes/base.py`, `apps/pipelines/node_options.py` and the builder widget (`assets/javascript/apps/pipeline/nodes/widgets.tsx:1132`), plus the serializer, the regenerated `api-schemas/v2.yml` and the tests. `IMPLIED_OPTION_KEYS` keeps an entry for it — the builder still declares no `ui:optionsSource` because it hard-codes the widget — but it is now an identity pairing like the other three.

apps/api/v2/discovery/node_types.py
SmittieC: Fixed in 15cf162dd. `options_source` is no longer emitted on any param, and `ui:optionsSource` joins the `ui:` keys that get dropped rather than translated.

Three follow-on changes this needed:

- **ADR-0051 rewritten**, since "the `options_source` join is total" was its lead decision. The rule is now "a param's options live under the `/pipeline/options/` key of the same name", with the renames (`LlmProviderId` → `llm_provider_id`, `jinja_node` → `prompt_variables`, `built_in_tools_config` → `tool_config`) doing the work that made that true. I edited it in place rather than writing a superseding ADR — it is unmerged and only exists on this branch, so there is no published decision to reverse. Shout if you would rather have the supersession trail.
- **`?node_type=` scoping** still needs the param → option-key pairing, so `option_keys_for_node_type()` reads `ui:optionsSource` internally. That is now the only consumer.
- **`test_every_options_source_resolves_to_an_options_key`** is replaced by `test_every_key_a_node_type_scopes_to_is_actually_served`, which keeps the same no-dangling-pointer guarantee at the scoping level, plus `test_no_param_carries_an_options_source` as the guard on the removal.

One case where the names do not match after this: `RenderTemplate.template_string` and the three `SendEmail` fields draw from `prompt_variables`. `?node_type=` still returns exactly that key for those nodes, so it is discoverable, just not by name alone.

apps/api/v2/tests/test_pipeline_discovery.py
```diff
@@ -0,0 +1,703 @@
...
+
+@pytest.mark.django_db()
+@pytest.mark.parametrize("auth_method", ["api_key", "oauth"])
+def test_list_node_types(auth_method, team):
+    client = ApiTestClient(team.members.first(), team, auth_method=auth_method)
+    response = client.get(reverse("api:v2:pipeline-nodes"))
+
+    assert response.status_code == 200
+    by_type = {entry["type"]: entry for entry in response.json()}
+    assert by_type["LLMResponseWithPrompt"]["description"]
```
SmittieC: Do these checks for all expected node types

apps/api/v2/tests/test_pipeline_discovery.py
```diff
@@ -0,0 +1,703 @@
...
+
+    types = {entry["type"] for entry in entries}
+    assert {"AssistantNode", "BooleanNode", "LLMResponse"}.isdisjoint(types)
+    assert {"StartNode", "EndNode", "Passthrough"}.isdisjoint(types)
+    assert "LLMResponseWithPrompt" in types
+    assert not [entry for entry in entries if "can_add" in entry]
+
+
+@pytest.mark.django_db()
+def test_no_namespaced_schema_key_survives_anywhere(team):
```
SmittieC, line 96: What is this testing? Add a docstring

apps/api/v2/tests/test_pipeline_discovery.py
```diff
@@ -0,0 +1,703 @@
...
+        pytest.param(
+            "LLMResponseWithPrompt",
+            {"source_material", "collection", "collection_index", "agent_tools", "custom_actions"},
+            id="params-the-builder-declares-a-source-for",
+        ),
+        pytest.param("RenderTemplate", {"template_variables"}, id="template-vars"),
+        pytest.param("SendEmail", {"template_variables"}, id="template-vars-outside-the-template-node"),
+    ],
+)
+def test_scoping_covers_every_param_that_reads_an_option_list(team_with_resources, node_type, expected_keys):
```
SmittieC, line 125: add docstrings to these tests

apps/utils/schema_utils.py
```diff
@@ -0,0 +1,50 @@
...
+                    data[k] = v
+        elif isinstance(data, list):
+            for i, item in enumerate(data):
+                if isinstance(item, dict | list):
+                    data[i] = resolve_ref(item, f"{path}/{i}")
+                else:
+                    data[i] = item
+        return data
+
+    return resolve_ref(deepcopy(openapi_spec), "")
```
SmittieC: Fixed in 53e299e. `resolve_references` was a closure doing three jobs; it is now three module-level functions — `resolve_references` (entry point), `_resolve` (walk) and `_resolve_ref` (swap one node for its target). The unused `path` argument and the two no-op `else` branches went with it.

SmittieC: Fixed in 53e299e, same change as the complexity finding above: the nesting came from the `for k, v` / `if isinstance` pairs inside the walk, which are now single-expression comprehensions in `_resolve`.

apps/pipelines/nodes/base.py
```diff
@@ -522,6 +535,8 @@ def __call__(self, schema: JsonDict):
             schema["ui:optionsSource"] = self.options_source
         if self.flag_required:
             schema["ui:flagRequired"] = self.flag_required
+        if self.api_exclude:
+            schema["api:exclude"] = True
```
SmittieC, line 551: Fixed in 53e299e. The six fields that map straight to a `ui:` key are now read from a `_UI_SCHEMA_KEYS` table, leaving `__call__` with the loop and the two fields that need their own handling (`api_exclude`, `visible_when`).

One behaviour note: the table applies `is not None` uniformly, where `widget`/`enum_labels`/`options_source`/`flag_required` previously tested truthiness. That only differs for a field explicitly set to an empty value (`[]`, `""`), which no node does — the 30 committed node-schema fixtures and the regenerated `api-schemas/v2.yml` are both unchanged.

apps/api/v2/tests/test_pipeline_discovery.py
```diff
@@ -0,0 +1,703 @@
...
+        served = options[key]
+        entries = [entry for group in served.values() for entry in group] if isinstance(served, dict) else served
+        assert entries, f"{key} came back empty, so it holds the docs to nothing"
+
+        optional = {name for name, field in documented.fields.items() if not field.required}
+        assert {name for entry in entries for name in entry} == set(documented.fields), key
+        for entry in entries:
+            assert set(documented.fields) - set(entry) <= optional, (key, entry)
+            if "value" in documented.fields:
+                assert isinstance(entry["value"], value_types[type(documented.fields["value"])]), (key, entry)
```
SmittieC: Fixed in 412d28b. The test looped over every documented option shape; it is now parametrised per key, so the loop is gone and each key gets its own test id. The `value` type check that used to be nested inside it is a separate test, `test_every_option_value_has_the_documented_type`.

SmittieC: Fixed in 412d28b, same change as the complexity finding above — the nested `for entry in entries` / `if "value" in documented.fields` block is now its own parametrised test.

apps/pipelines/nodes/node_metadata.py
```diff
@@ -0,0 +1,205 @@
...
+            key=lambda v: v["label"],
+        )
+    }
+
+
+def _option(value, label, type_=None, edit_url: str | None = None, max_token_limit=None):
+    data = {"value": value, "label": label}
+    data = data | ({"type": type_} if type_ else {})
+    data = data | ({"edit_url": edit_url} if edit_url else {})
+    data = data | ({"max_token_limit": max_token_limit} if max_token_limit else {})
```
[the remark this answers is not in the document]
SmittieC: Valid, fixed in 34a917b. `max_token_limit` is a `PositiveIntegerField` and 0 is a documented setting (`"If 0, compression will be disabled"`), so the field is now dropped only when it is `None`. Added `test_a_zero_max_token_limit_is_served_rather_than_dropped`, and reworded the serializer help text to state what 0 means rather than "Absent where none is recorded".

The "also affects" pointer to `apps/evaluations/views/evaluator_views.py:24` is off by a couple of hundred lines — line 24 is an import. There is a genuine second copy of this helper at `evaluator_views.py:262` with the same check, but it is untouched by this PR, so I have left it for a separate change.

apps/api/v2/tests/test_pipeline_discovery.py
SmittieC: Done in 412d28b. It now asserts a non-empty `description` and non-empty `schema.properties` for every listed type, with the type name as the assertion message, and keeps a single check that `LLMResponseWithPrompt` is in the list so the loop cannot pass vacuously.

apps/api/v2/tests/test_pipeline_discovery.py
SmittieC: Added in 412d28b. It is checking that the translation layer leaves no namespaced key behind: the builder schemas carry `ui:`/`api:` keys, the handful that mean something to a client are re-served under plain names, so a `:` surviving anywhere in the response — at the schema level or on a single param — is an untranslated key.

apps/api/v2/tests/test_pipeline_discovery.py
SmittieC: Done in 412d28b — every test in the file that lacked a docstring now has one, and both this file and `apps/pipelines/tests/test_node_options.py` got a module docstring saying which half of the split they hold.

While in here I also acted on your other note and moved the option-content tests (which of a team's resources reach each option list) over to `test_node_options.py`, next to the helpers that build them. I moved them as-is rather than rewriting them against `get_node_parameter_values`: several of them cover filtering that lives in `PipelineOptionsView` (model type/deprecation, voice service, `tool_config` scoping), so rewriting them at helper level would have meant asserting against querysets the test itself constructed. Happy to take it further if you would rather that file stayed endpoint-free.

apps/api/v2/discovery/views.py
```diff
@@ -0,0 +1,296 @@
...
+    OptionsSource.jinja_node,
+    OptionsSource.text_editor_autocomplete_vars_llm_node,
+    OptionsSource.text_editor_autocomplete_vars_router_node,
+)
+
+
+class DiscoveryView(GenericAPIView):
+    """Shared auth for the discovery endpoints."""
+
+    permission_classes = [DjangoModelPermissionsWithView, TokenHasOAuthResourceScope]
```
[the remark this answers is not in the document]
SmittieC: Confirmed and fixed in 3a64e37. `permission_classes` now spreads `BASE_PERMISSION_CLASSES` in, matching `ChatbotViewSet`, and both views are in `API_KEY_VIEWS` so `test_api_key_views_keep_the_read_only_gate` covers the new surface.

Your read of the blast radius matches mine: not exploitable today, but `DiscoveryView` is deliberately the shared base for the rest of the pipeline write API, so the first non-GET subclass would have inherited the hole.

apps/api/v2/discovery/views.py
```diff
@@ -0,0 +1,296 @@
...
+                response=NodeTypeNotFoundSerializer,
+                description="No such node type, or the type is deprecated and therefore not listed.",
+            ),
+        },
+        examples=[
+            OpenApiExample(
+                name="NodeTypes",
+                summary="A router, showing the per-keyword outputs and a param-pairing rule.",
+                value=[
+                    {
```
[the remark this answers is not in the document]
SmittieC: Confirmed and fixed in 3a64e37. The example is one entry now, with a comment saying why, and the regenerated `v2.yml` reads `- type: RouterNode` instead of `- - type: RouterNode`.

apps/api/v2/discovery/views.py
```diff
@@ -0,0 +1,296 @@
...
+        ],
+    )
+    def get(self, request):
+        requested_type = request.query_params.get("node_type")
+        if requested_type:
+            wanted = option_keys_for_node_type(requested_type)
+            if wanted is None:
+                raise unknown_node_type(requested_type)
+        else:
+            wanted = served_option_keys() | API_ONLY_OPTION_KEYS
```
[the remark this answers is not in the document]
SmittieC: Confirmed and fixed in 3a64e37, taking the dependency-map approach you suggested. `contract.py` gains:

```python
OPTION_KEY_DEPENDENCIES = {"synthetic_voice_id": frozenset({"voice_provider_id"})}
API_ONLY_OPTION_KEYS = frozenset().union(*OPTION_KEY_DEPENDENCIES.values())
```

and the view expands `wanted` after both branches, so the unscoped behaviour is unchanged. New test `test_scoped_options_carry_the_key_that_resolves_a_voice` pins `?node_type=LLMResponseWithPrompt` carrying both keys, and `test_options_can_be_scoped_to_one_node_type` already pins `RenderTemplate` scoping to `template_variables` alone, so the over-serving case stays covered.

apps/api/v2/discovery/contract.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+
+# The exception to the same-name rule. No param is named for the variable list it draws on, and
+# `prompt` on an LLM node and `prompt` on a router accept different sets, so these keys name the
+# flavour of text being written rather than the param writing it.
+OPTIONS_KEY_RENAMES = {
+    "jinja_node": "template_variables",
+    "text_editor_autocomplete_vars_llm_node": "llm_prompt_variables",
+    "text_editor_autocomplete_vars_router_node": "router_prompt_variables",
+}
```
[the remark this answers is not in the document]
SmittieC: Confirmed. Took the first option in 3a64e37: `"agent_tools": "tools"` is now in `OPTIONS_KEY_RENAMES`, the serializer field is `tools` with `help_text` distinguishing it from `custom_actions` and `built_in_tools`, the response example and `v2.yml` are updated. The rename is API-visible but the endpoint is unreleased, and I would rather the stated rule hold than carry a second footnote. The builder key is untouched -- `OptionsSource.agent_tools` still names the list it consumes.

On the missing guard, you were right that reachability through `ui:optionsSource` cannot catch this, so `test_every_option_key_is_named_for_the_param_that_reads_it` now checks the rule the client actually applies: every served key must match a served param name, give or take an `_id`/`_ids` suffix, with the exceptions enumerated (the three variable lists, `voice_provider_id`, `default_llm_provider`). Verified it fails with `agent_tools` before passing with `tools`.

apps/api/v2/discovery/views.py
```diff
@@ -0,0 +1,297 @@
...
+            for provider_type, config in options[OptionsSource.tool_config].items()
+            if provider_type in llm_provider_types
+        }
+        # See `OPTION_KEY_DEPENDENCIES` -- no node param sources its options from this list.
+        options[OptionsSource.voice_provider_id] = [
+            {"value": provider.id, "label": provider.name, "type": provider.type} for provider in voice_providers
+        ]
+        options["default_llm_provider"] = get_node_default_values(llm_providers, llm_provider_models)
+        options = self._to_api_vocabulary(self._describe_prompt_vars(options))
+        return Response({key: value for key, value in options.items() if key in wanted})
```
SmittieC: Fixed in 56f27c4. The dependency expansion added the branch that tipped this over. `get()` is now the three steps it always was — pick the keys, build the options, filter — with `_wanted_keys()`, `_options_for_team()` and `_speakable_voices()` alongside it. No behaviour change; the regenerated `v2.yml` is identical and all 101 tests in the two files still pass.

apps/api/v2/discovery/views.py
```diff
@@ -0,0 +1,310 @@
+"""Team-level discovery endpoints for the chatbot write API: what a client can build
```
SmittieC, line 1: Leaving this one as-is. It surfaced because splitting `PipelineOptionsView.get` into named helpers (56f27c4, clearing the Complex Method finding on the same file) raised the module mean — more functions, and the mean is total over count. Every function in the module is comfortably under the per-function threshold.

I checked whether extracting the option reshaping into its own `options.py` — mirroring `node_types.py` — would clear it: `views.py` would drop to roughly 3.3, but the new module would land near 4.6, so the finding relocates rather than goes away. Not worth a package restructure at 4.25 against a threshold of 4.

apps/api/v2/discovery/contract.py
```diff
@@ -0,0 +1,60 @@
...
+
+# `builder option key -> the key the API serves it under`. These are the builder keys that would
+# otherwise break the same-name rule: `agent_tools` holds the values for a param named `tools`, and
+# the variable lists have no param named for them at all -- `prompt` on an LLM node and `prompt` on a
+# router accept different sets, so those keys name the flavour of text being written instead.
+OPTIONS_KEY_RENAMES = {
+    "agent_tools": "tools",
+    "jinja_node": "template_variables",
+    "text_editor_autocomplete_vars_llm_node": "llm_prompt_variables",
+    "text_editor_autocomplete_vars_router_node": "router_prompt_variables",
```
snopoke: I think we could just rename these in code

apps/api/v2/discovery/contract.py
```diff
@@ -0,0 +1,60 @@
...
+}
+
+# `option key -> the keys a client needs alongside it to make sense of its entries`. A dependency is
+# served whenever the key depending on it is, scoped responses included, and no node param reads one
+# -- which is why the schema-derived whitelist cannot discover them on its own.
+OPTION_KEY_DEPENDENCIES = {
+    # Resolves the `provider_id` carried on each `synthetic_voice_id` entry.
+    "synthetic_voice_id": frozenset({"voice_provider_id"}),
```
snopoke: you could potentially encode this in the schema directly using `visible_when=VisibleWhen("voice_provider_id", "is_not_empty")`. Though you may still need some code here to unpack what that means for the API

snopoke (commented):
some nice refactoring in here

apps/api/v2/discovery/contract.py
SmittieC: Done — `OPTIONS_KEY_RENAMES` is gone. The `OptionsSource` members now carry the names the API was translating to:

- `agent_tools` -> `tools` (matching the param it feeds)
- `jinja_node` -> `template_variables`
- `text_editor_autocomplete_vars_llm_node` -> `llm_prompt_variables`
- `text_editor_autocomplete_vars_router_node` -> `router_prompt_variables`

That drops the map, `PipelineOptionsView._to_api_vocabulary`, and the `OPTIONS_KEY_RENAMES.get(source, source)` in `_option_keys_by_type`. Regenerating `api-schemas/v2.yml` produces no diff, so the published surface is unchanged.

I had claimed on an earlier thread that the builder's JS reads these keys verbatim, which is why I put the rename in the API layer. That was wrong for these four. The builder looks options up generically:

```js
// assets/javascript/apps/pipeline/utils.tsx
if (schema["ui:optionsSource"]) {
  return parameterValues[schema["ui:optionsSource"]];
}
```

Both sides of that lookup are generated from the enum, so renaming a member moves them together. The hardcoded JS references (`built_in_tools`, `tool_config`, `llm_provider_id`, `llm_provider_model_id`, `synthetic_voice_id`) are all keys that weren't being renamed.

Since nothing in the API layer now bridges a mismatch, I added `test_every_declared_options_source_has_a_list_to_draw_on` — it catches a param declaring a source that no builder populates, which is a silently empty dropdown rather than an error. Sharing the enum already rules out the two spellings drifting apart, so the test doesn't try to cover that.

Verified in the builder as well as the test suite, since the `parameter-values` blob the browser indexes into changed: every `ui:optionsSource` on the page resolves against it, and the three widget types that read the renamed lists still populate — the LLM prompt autocomplete offers `participant_data`, the Jinja template autocomplete offers `session_state`, and the Tools multiselect renders all 13 entries.

SmittieC: See https://github.com/dimagi/open-chat-studio/pull/4149/changes/aa0036b49f8f0f7d42594c2e9140e591f9ad54e1

apps/api/v2/discovery/contract.py
SmittieC: Actually this can go. The original implementation is wrong I see. The UI only shows synthetic_voice_id when experiment.settings has voice_provider_id configured, and we should do the same here

SmittieC: https://github.com/dimagi/open-chat-studio/pull/4149/commits/454b9392999d10c6d552b4c5fe0f48f7d3c58d53

api-schemas/v2.yml
```diff
@@ -115,6 +115,175 @@ paths:
...
+
+        Static per deploy: revalidate a cached copy with `If-None-Match` against the `ETag`.
+      summary: List Pipeline Node Types
+      parameters:
+      - in: query
+        name: type
+        schema:
+          type: string
+        description: Return only this node type. An unknown or deprecated name returns
+          404.
```
snopoke: If this is going to 404 on an unknown value I think it should be a path param not a query param:

`/api/v2/pipeline/nodes/PythonNode`

Alternately just return an empty list. It is a bit unusual to 404 based on a query parameter that doesn't match.

api-schemas/v2.yml
```diff
@@ -115,6 +115,175 @@ paths:
...
+      description: |-
+        The values each node param accepts, scoped to the API key's team.
+
+        A key holds the values for the node param of the same name: write one of `source_material`'s entries into a node's `source_material_id`, one of `collection_index`'s into `collection_index_ids`.
+
+        The variable lists are the exception, because no param is named for the list it draws on and two different params are both named `prompt`. Jinja params -- `template_string` and `SendEmail`'s fields -- draw from `template_variables` and are written double-braced (`{{input}}`); an LLM node's `prompt` draws from `llm_prompt_variables` and a router's from `router_prompt_variables`, both written single-braced (`{source_material}`). Pass `?node_type=` to receive only the list that applies -- the sets are not interchangeable.
+      summary: List Pipeline Node Options
+      parameters:
+      - in: query
+        name: node_type
```
snopoke: similar comment as above about query param and 404 response

apps/api/v2/discovery/node_types.py
```diff
@@ -0,0 +1,153 @@
...
+        detail = f"Unknown node type: {requested_type}"
+    return NotFound({"detail": detail, "valid_types": _valid_type_names()})
+
+
+def etag(payload) -> str:
+    digest = hashlib.sha256(json.dumps(payload, sort_keys=True, default=str).encode()).hexdigest()
+    return f'W/"{digest[:32]}"'
+
+
+def _addable_schemas() -> list[dict]:
```
snopoke: nit: rename to 
```suggestion
def _available_schemas() -> list[dict]:
```

apps/api/v2/discovery/views.py
```diff
@@ -0,0 +1,312 @@
...
+        )
+
+        options = cls._clean_options(
+            get_node_parameter_values(
+                team=team,
+                llm_providers=llm_providers,
+                llm_provider_models=llm_provider_models,
+                synthetic_voices=cls._speakable_voices(team, voice_providers),
+            )
+        )
```
snopoke: Can we not consolidate this to avoid having to duplicate the fetch of the llm_providers etc here and in chatbot and pipeline views?

snopoke (commented):
Generally looks good. Some comments and thoughts.

api-schemas/v2.yml
```diff
@@ -115,6 +115,175 @@ paths:
...
+                      title: RouterNode
+                      type: object
+                      required:
+                      - llm_provider_id
+                      - llm_provider_model_id
+                      - name
+                      properties:
+                        llm_provider_id:
+                          type: integer
+                          title: LLM Model
```
SmittieC: this should be "LLM Provider"

apps/api/v2/discovery/node_types.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/4149/changes/bd7b829edde6c599c3701a0ae8d72b7d5a4f163a

api-schemas/v2.yml
SmittieC: See https://github.com/dimagi/open-chat-studio/pull/4149/changes/187d46091a3759a69ccec693cf18aa7aafed49f7

apps/api/v2/discovery/views.py
SmittieC: Something like https://github.com/dimagi/open-chat-studio/pull/4149/changes/a5612d4efede3ef073159194e5429a1b2a47ce28?

snopoke (approved):

PR #4157, opened 2026-08-11
Escape JSON rendered by the to_json filter

snopoke:
### Product Description
Data a chat participant supplies can no longer inject markup into the team-facing session detail, annotation and trace pages.

### Technical Description
The `to_json` template filter wrapped raw `json.dumps` output in `mark_safe`, which switches off autoescaping. `json.dumps` does not escape `<`, `>` or `&`, and the local `escape_script_tags` helper only rewrote a literal `</script>`, so stored values reached the browser as markup rather than text.

The filter now returns a plain string and lets the template engine escape it. All nine call sites render display-identical output; the one site that chains `|force_escape` inside an `x-data` attribute is byte-identical.

Found by an automated security review. `apps/web/tests/test_json_tags.py` covers the escaping through the real template engine and fails without the change.

### Migrations
None.

### Demo
n/a

### Docs and Changelog

### DIFF
--- a/apps/web/templatetags/json_tags.py
+++ b/apps/web/templatetags/json_tags.py
@@ -38,20 +38,20 @@ def pygments_json_css() -> str:
 
 @register.filter
 def to_json(obj):
-    """Source: https://gist.github.com/czue/90e287c9818ae726f73f5850c1b00f7f"""
+    """Render a Python value as indented JSON text.
 
-    def escape_script_tags(unsafe_str):
-        # seriously: http://stackoverflow.com/a/1068548/8207
-        return unsafe_str.replace("</script>", '<" + "/script>')
-
+    The returned string is deliberately *not* marked safe: the data may contain
+    attacker-controlled values (e.g. participant data or session state supplied
+    through the API), so Django's autoescaping must run over the JSON to prevent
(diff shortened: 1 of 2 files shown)


snopoke:
@claude Review the code at the location below. A potential bug has been identified by an AI
agent. Verify if this is a real issue. If it is, propose a fix; if not, explain why it's
not valid.

Location: apps/web/templatetags/json_tags.py#L52-L54

Potential issue: The `to_json` template filter uses a `try...except JSONDecodeError`
block to handle errors from `json.dumps()`. However, `json.dumps()` raises `TypeError`
or `ValueError` when it encounters an object it cannot serialize, not `JSONDecodeError`.
This means if a non-serializable object (e.g., a custom class instance or a set) is
passed to the filter, the exception will not be caught, leading to an unhandled
exception and a server error. The `highlight_json` filter in the same file correctly
catches `(TypeError, ValueError)` for an identical operation, confirming the incorrect
exception type is being used in `to_json`.

snopoke:
@claude please commit the changes you reference in your earlier comment (https://github.com/dimagi/open-chat-studio/pull/4157#issuecomment-5253133504)

SmittieC (approved):

PR #4160, opened 2026-08-11
Keep the read-only API key gate on views that override permission_classes

snopoke:
### Product Description
An API key issued as read-only can no longer perform writes on the sessions API.

### Technical Description
`ReadOnlyAPIKeyPermission` — the ADR-0021 "look but don't touch" gate — exists only in `REST_FRAMEWORK["DEFAULT_PERMISSION_CLASSES"]`. A view that sets `permission_classes` replaces that list wholesale, silently dropping the gate. `ExperimentSessionViewSet` did exactly that while exposing four write endpoints.

This adds `BASE_PERMISSION_CLASSES = [IsAuthenticatedOrMachineToken, ReadOnlyAPIKeyPermission]` in `apps/api/permissions.py` and spreads it into every view that overrides the defaults: `ExperimentSessionViewSet`, `ExperimentViewSet`, v2 `ChatbotViewSet`, v2 `UsageView`, and v2 `MeView` (which keeps its deliberate `IsAuthenticated`). Four of those five have no write endpoints today, so the addition is latent-hole closure there.

The OAuth scope class is deliberately not in the shared list: each view pairs its own (`TokenHasOAuthScope` vs `TokenHasOAuthResourceScope`) with `required_scopes`, and including both would demand two scopes from every OAuth caller.

Found by an automated security review. `apps/api/tests/test_session_api.py` covers all five write entry points and fails without the change; `test_permissions.py` guards against a future override dropping the gate again.

### Migrations
None.

### Demo
n/a

### Docs and Changelog

### DIFF
--- a/apps/api/permissions.py
+++ b/apps/api/permissions.py
@@ -190,6 +190,14 @@ class ReadOnlyAPIKeyPermission(BasePermission):
         return True
 
 
+# The non-scope half of REST_FRAMEWORK["DEFAULT_PERMISSION_CLASSES"] (config/settings.py). Setting
+# ``permission_classes`` on a view replaces the defaults wholesale, which silently drops the
+# read-only API-key gate (ADR-0021), so views that need extra permission classes must build on this
+# list rather than start from an empty one. The OAuth scope class is deliberately excluded: each view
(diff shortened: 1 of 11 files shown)


SmittieC (approved):
Looks good to me

PR #4162, opened 2026-08-11
Gate the participants API on the same permissions as the UI

snopoke:
### Product Description
The participants API now enforces the same role permissions as the participants UI, so a team role that cannot see participants in the app cannot read or rewrite them through the API either.

### Technical Description
`ParticipantView` was gated only on "is authenticated", "the key is not read-only" and (for OAuth tokens only) the participants scope. Unlike `ExperimentViewSet` and `ExperimentSessionViewSet`, it never consulted the caller's model permissions, so team membership plus a self-issued key was the whole authorization decision for a team-wide participant read and for the `Participant` / `ParticipantData` / `ScheduledMessage` writes.

New `ParticipantModelPermissions` in `apps/api/permissions.py`, appended to the view's existing `permission_classes`:

- GET/HEAD → `experiments.view_participant` + `experiments.view_participantdata`, matching the UI export view;
- POST → `experiments.{add,change}_participant` + `experiments.{add,change}_participantdata`, which is literally `IMPORT_PERMISSIONS` from `apps/participants/views.py`;
- OPTIONS left open (metadata only), matching `DjangoModelPermissions`.

The schedules branch gets no separate gate: it is reachable only from POST, which already requires `experiments.change_participant` — the permission the UI itself uses for schedule cancel and `trigger_bot`. Client-credentials machine tokens short-circuit as they do elsewhere.

Operational note: an existing API key belonging to a low-privilege member will start receiving 403 on `/api/participants`.

Found by an automated security review. `apps/api/tests/test_participant_api.py` covers the denial and allow matrix and fails without the change.

### Migrations
None.

### Demo
n/a

### Docs and Changelog


apps/api/permissions.py
```diff
@@ -214,6 +214,52 @@ def has_permission(self, request, view):
...
+        "POST": PARTICIPANT_WRITE_PERMISSIONS,
+    }
+
+    def get_required_permissions(self, method):
+        if method not in self.perms_map:
+            raise exceptions.MethodNotAllowed(method)
+        return self.perms_map[method]
+
+    def has_permission(self, request, view):
+        if is_client_credentials_request(request):
```
SmittieC, line 252: Not for this PR, just commenting: I'm starting to see a pattern with these custom permission classes checking `if is_client_credentials_request(request):` and returning True. Perhaps we should consider adding a base class for "custom" permission classes like this

SmittieC (dismissed):

SmittieC (approved):

PR #4164, opened 2026-08-11
Verify a shared secret on SureAdhere inbound webhooks

snopoke:
### Product Description
Adds an optional shared secret to the SureAdhere messaging provider and verifies it on inbound webhooks, so a caller must present a credential rather than just a tenant id.

### Technical Description
`new_sureadhere_message` is `@csrf_exempt` and sits outside auth middleware; its only check was that the path's `sureadhere_tenant_id` matched a channel. A tenant id is an identifier, not a credential. Every sibling webhook in the same file verifies something (Turn.io and Meta by HMAC, Telegram by a high-entropy path token) — this one verified nothing.

- New `apps/channels/sureadhere_webhook.py` with `get_presented_secret` (accepts `X-OCS-Webhook-Secret`, or `Authorization: Bearer`) and `verify_secret` using `constant_time_compare` behind a guard so empty-vs-empty cannot match.
- New optional `webhook_secret` on `SureAdhereMessagingConfigForm`, stored in the already-encrypted `MessagingProvider.config` and masked in the UI.
- The view guard runs before `json.loads` and before the task dispatch; a missing or wrong credential is a 401.

**Verify-when-set, deliberately.** No deployment has this config today, so a provider with no secret is still accepted, now with a `WARNING` per request naming the channel and team. Failing closed in one step would drop real patient messages. Status codes are unchanged for every request except one presenting no/wrong credential to a provider that has a secret set.

**Before this helps anything:** SureAdhere must be configured to send the secret in one of the two accepted locations, and whether its callback registration supports either is not documented anywhere in this repo — worth confirming with the vendor. Until an operator sets a secret per provider, that provider's exposure is unchanged.

Also worth a follow-up: the 401 carries no `WWW-Authenticate` and SureAdhere's retry behaviour is undocumented here, so a mistyped secret could mean silently lost messages.

Found by an automated security review. `apps/channels/tests/test_sureadhere_webhook.py` covers both credential locations, wrong/missing/whitespace/case/scheme variants, rejection before body parsing, and the unconfigured pass-through.

### Migrations
None — `webhook_secret` lands in the existing `config` JSON field.

### Demo
n/a

### Docs and Changelog

### DIFF
--- a/apps/channels/sureadhere_webhook.py
+++ b/apps/channels/sureadhere_webhook.py
@@ -1,0 +1,35 @@
+from django.http.request import HttpHeaders
+from django.utils.crypto import constant_time_compare
+
+SECRET_HEADER = "X-OCS-Webhook-Secret"
+
+
+def get_presented_secret(headers: HttpHeaders) -> str:
+    """Extract the shared secret an inbound SureAdhere webhook presents, or "" if it presents none.
+
+    SureAdhere does not sign its callbacks, so the only credential available is a secret
+    configured on both sides. It is accepted in either the ``X-OCS-Webhook-Secret`` header or
+    as an ``Authorization: Bearer`` token, because the callback is registered in SureAdhere's
+    own configuration and which of the two it can send is not recorded in this repo.
+
+    Lookups go through ``HttpHeaders``, which is case-insensitive, so header casing does not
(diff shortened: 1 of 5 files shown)


snopoke:
This requires coordination with the SureAdhere team.

SmittieC:
@snopoke is this change waiting on SureAdhere?

Issue #2140, opened 2025-09-04
Implement API rate limiting


snopoke:
Initial spec: https://docs.google.com/document/d/1iE6MYWSTQMBr64XQDEWMBmNR3pcnqT4V/edit
See #1408

PR #4174, opened 2026-08-11
Rate limit the widget chat API endpoints

barry47products:
### Product Description

The chat widget's `/api/chat/*` endpoints currently share the team's `api` rate limit bucket. (These are the live widget API views in `apps/api/views/chat.py`, not the legacy `*_embed` flow that #4168 retired.) One busy conversation on a chatbot can therefore spend the allowance of every other visitor to that same chatbot, and of the team's own API integrations.

This gives the widget its own `widget` scope, bucketed per chat session, so conversations are isolated from each other and from the team's API traffic.

Ships in log-only mode: `RATE_LIMIT_ENFORCE` stays `False`, so no request is blocked on default config. The only visible change is the `X-RateLimit-*` response headers.

Part of #2140.

### Technical Description

- Registers a `widget` scope in `RATE_LIMITS` at `300/5m`, fail-open, env-overridable via `RATE_LIMIT_WIDGET`. The rate comes from the signed-off design (observed peak 108/IP/5min across sessions).
- Adds `WidgetRateThrottle`, keying on the session id from the URL, then the widget channel from `request.auth`, then client IP for legacy clients. Session creation has no session to key on yet, so it buckets per chatbot channel, which is the step worth bounding there.
- Applies it to the five `/api/chat/*` views with `@throttle_classes`. Setting `throttle_classes` replaces `DEFAULT_THROTTLE_CLASSES` rather than adding to it, so every request still lands in exactly one bucket.
- Documents `RATE_LIMIT_WIDGET` in `.env.example` and the hosting configuration page.

Stacked on #4173, which is stacked on #4170. Review the diff from `bt/rate-limit-admin-api-scope-2140`.

Three things worth a reviewer's attention:

1. **`identity` signature widened.** `APIRateThrottle.identity` now takes `(request, view)`, because the session id is a URL capture and only the view carries it. `allow_request` is the sole production caller. The `webhook` scope PR needs the same widening, so whichever lands second drops that step.
2. **Five endpoints leave the `api` scope.** Its log-only baseline will drop accordingly once this is deployed, which matters when reading `rate_limit.would_block` data ahead of the enforcement flip.
3. **Known gap, not addressed here.** DRF resolves authentication before throttling, so a request with an invalid `X-Embed-Key` is rejected at authentication and never counted. Repeated forged-key probing is therefore not bounded by this scope. The fix belongs with the `credentials` scope's failed-authentication counting.

Not built: the design's optional team-level backstop for this scope. Nothing needs it yet, and a second bucket per request would break the one-bucket-per-request property the core relies on. Worth revisiting once log-only data shows whether a single chatbot gets driven hard across many sessions.

### Migrations

No migrations in this PR.

### Demo

Driven against a local dev server with `RATE_LIMIT_WIDGET=3/5m`, against a chatbot with an embedded-widget channel:

- `POST /api/chat/start/` returns 201 with `X-RateLimit-Limit: 3`, `X-RateLimit-Remaining: 2`, and `access-control-expose-headers` listing all four rate limit headers, so cross-origin widget clients can read them.
- Polling the resulting session starts at `Remaining: 2`, confirming the poll bucket is separate from the channel bucket that session creation used.
- Log-only: over-limit polls still return 200, and one sampled `rate_limit.would_block` line is emitted rather than one per request.
- With `RATE_LIMIT_ENFORCE=True`: 429 with `Retry-After: 135` and body `{"detail":"Rate limit exceeded.","available_in":95}`.
- Isolation: with one session exhausted and returning 429, a second session on the same chatbot polls 200 with `Remaining: 2`.

### Docs and Changelog

`RATE_LIMIT_WIDGET` is documented in `docs/hosting/configuration.md` alongside the other rate limit settings.


barry47products:
@coderabbitai review

docs/hosting/configuration.md
```diff
@@ -52,6 +52,7 @@ All configuration is via environment variables. In production, set `DJANGO_SETTI
...
 | `RATE_LIMIT_API` | `2000/5m` | Request limit for the `api` scope, format `count/window` with `s`/`m`/`h` units. Fails open: if the limiter's cache is unreachable, requests are served. |
 | `RATE_LIMIT_ADMIN_API` | `100/5m` | Request limit for the `admin_api` scope (the `/admin/api/*` autocomplete and provider-reporting endpoints). Keyed by authenticated user, then the provider-reporting token, then client IP, so anonymous traffic cannot spend a staff member's allowance. Set `RATE_LIMIT_TRUSTED_PROXY_COUNT` before enforcing, or all anonymous callers behind a proxy share one bucket. Fails open: if the limiter's cache is unreachable, requests are served. |
+| `RATE_LIMIT_WIDGET` | `300/5m` | Request limit for the `widget` scope (the embedded chat widget's `/api/chat/*` endpoints), keyed per chat session, and per widget channel on session creation since no session exists yet. Sits apart from `RATE_LIMIT_API` so one busy conversation cannot spend the team's interactive API allowance. Fails open: if the limiter's cache is unreachable, requests are served. |
```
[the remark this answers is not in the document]
barry47products: @coderabbitai added in commit 787dd2131fef43df95a4c61d496631708564f8ac

snopoke:
@barry47products some (or maybe all) of these endpoints have been removed: https://github.com/dimagi/open-chat-studio/pull/4168

barry47products:
> some (or maybe all) of these endpoints have been removed: #4168

@snopoke Thanks, that catches something real, though I want to check I've followed it to the right place.

#4174 here only touches the five `/api/chat/*` views in `apps/api/views/chat.py`, which aren't in #4168's diff and are still routed on main. But #4175, stacked on top of this branch, decorates five views that #4168 replaced with `embed_flow_gone`: `start_session_public_embed`, `chatbot_chat_embed`, `experiment_session_message_embed`, `poll_messages_embed` and `start_chatbot_session_public_embed`. It even has a commit keeping the 429 frameable inside the legacy iframe.

Is that the overlap you were pointing at, or is there something on the `/api/chat/*` side I should be looking at as well?

If it's #4175, I'll rebase the stack onto main and drop the embed work there. Two commits go entirely, and four decorator lines come out of a third, leaving the nine live public chat views.

One thing I'd rather not decide alone: should `embed_flow_gone` sit in the `public_chat` scope at all? Legacy callers will keep hitting those paths for a while, but the stub returns a constant response with no DB work, so the limiter's cache round trip probably costs more than serving the 410. My inclination is to leave it unlimited.

snopoke:
> > some (or maybe all) of these endpoints have been removed: #4168
> 
> @snopoke Thanks, that catches something real, though I want to check I've followed it to the right place.
> 
> #4174 here only touches the five `/api/chat/*` views in `apps/api/views/chat.py`, which aren't in #4168's diff and are still routed on main. But #4175, stacked on top of this branch, decorates five views that #4168 replaced with `embed_flow_gone`: `start_session_public_embed`, `chatbot_chat_embed`, `experiment_session_message_embed`, `poll_messages_embed` and `start_chatbot_session_public_embed`. It even has a commit keeping the 429 frameable inside the legacy iframe.

Sorry, I just read the PR title and assumed. This is related to changes in #4175. 

> One thing I'd rather not decide alone: should `embed_flow_gone` sit in the `public_chat` scope at all? Legacy callers will keep hitting those paths for a while, but the stub returns a constant response with no DB work, so the limiter's cache round trip probably costs more than serving the 410. My inclination is to leave it unlimited.

No need to rate limit the `embed_flow_gone` view.

config/settings.py
```diff
@@ -663,6 +663,7 @@ def _celery_task_queues() -> tuple[Queue, ...]:
 RATE_LIMITS = {
     "api": {"rate": env("RATE_LIMIT_API", default="2000/5m"), "fail_open": True},
     "admin_api": {"rate": env("RATE_LIMIT_ADMIN_API", default="100/5m"), "fail_open": True},
+    "widget": {"rate": env("RATE_LIMIT_WIDGET", default="300/5m"), "fail_open": True},
```
snopoke: Can we rename this to `RATE_LIMIT_CHAT_API` since it also used by non-widget clients.

snopoke (commented):
One small change but otherwise looks good

config/settings.py
barry47products: No problem. will push a small change

snopoke (approved):

Issue #4176, opened 2026-08-12
RDS Proxy pins sessions: .iterator() declares WITH HOLD cursors on the chat request path


snopoke:
## Symptom

RDS Proxy logs this for chat history reads:

```
[INFO] [proxyEndpoint=default] [clientConnection=3327650365] If this prepared statement runs,
the client session will be pinned to the database connection [dbConnection=4266821654].
Reason: SQL changed session settings that the proxy doesn't track.
Digest: "declare "_django_curs_140042145613568_sync_7" no scroll cursor with hold for
select "chat_chatmessage".… from "chat_chatmessage" where ("chat_chatmessage"."chat_id" = $1
and not ("chat_chatmessage"."id" = $2)) order by "chat_chatmessage"."created_at" desc"
```

## Root cause

`Chat.message_iterator` (`apps/chat/models.py:84-91`) calls `.iterator(100)`. On Postgres, Django implements `.iterator()` with a server-side named cursor unless `DISABLE_SERVER_SIDE_CURSORS` is set (`django/db/models/query.py:526`), and declares it with `withhold=self.connection.autocommit` (`django/db/backends/postgresql/base.py:419-420`). We don't set `ATOMIC_REQUESTS`, so request-path code runs in autocommit → the cursor is declared `WITH HOLD`.

A `WITH HOLD` cursor outlives its transaction, so RDS Proxy cannot return the backend connection to the shared pool and pins the session. Pinning lasts until the *client* connection closes — and with the psycopg pool (`DJANGO_DATABASE_POOL_MIN_SIZE=2` / `MAX_SIZE=35`, `config/settings.py:249-254`) client connections are long-lived. Over time most of the pool ends up pinned and the proxy's multiplexing is effectively lost.

This particular query is on the hottest path we have: `Chat.get_langchain_messages_until_marker` → `message_iterator`, i.e. once per chat turn.

## Secondary bug: the setting is a no-op in production

`config/settings.py:233-245`:

```python
if "DATABASE_URL" in env:
    DATABASES = {"default": env.db()}
else:
    DATABASES = {
        "default": {
            ...
            "CONN_HEALTH_CHECKS": True,
            "DISABLE_SERVER_SIDE_CURSORS": env.bool("DJANGO_DISABLE_SERVER_SIDE_CURSORS", default=False),
        }
    }
```

[...]

PR #4177, opened 2026-08-12
Apply CONN_HEALTH_CHECKS and DISABLE_SERVER_SIDE_CURSORS with DATABASE_URL

snopoke:
### Product Description
No change.

### Technical Description
`CONN_HEALTH_CHECKS` and `DISABLE_SERVER_SIDE_CURSORS` were set only inside the `else` branch of the `DATABASE_URL` check in `settings.py`, so neither reached Django in any environment that sets `DATABASE_URL` — i.e. all of them. `DJANGO_DISABLE_SERVER_SIDE_CURSORS=True` was a silent no-op, which matters because we need that knob to evaluate the RDS Proxy session pinning in #4176. Django reads both keys from the top level of the connection's settings dict, and django-environ never emits them from a URL, so they can't be smuggled in via the query string either.

No behaviour change intended beyond the two keys now being present: the default stays `False`, so server-side cursors are still enabled until someone sets the var. Verified against the local `.env` (which sets `DATABASE_URL`) that `connection.settings_dict` now carries `CONN_HEALTH_CHECKS: True` and `DISABLE_SERVER_SIDE_CURSORS: False`, where previously both keys were absent.

The config moved to `config/db.py` so both branches can be unit tested — `settings.py` isn't reachable from a test without reloading the module. I confirmed the test fails correctly by reinstating the bug: only the two `[database_url]` cases fail, with `KeyError` on each key.

Also documents `DJANGO_DATABASE_USE_POOL`, `CONN_MAX_AGE`, `SSLMODE` and `DISABLE_SERVER_SIDE_CURSORS`, and moves the pool/TLS knobs out of the "alternative to DATABASE_URL" section of the hosting docs, since they apply either way — the old placement implied the same thing this bug did.

### Docs and Changelog

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,9 +11,15 @@ DATABASE_URL='postgres://postgres:postgres@localhost:5432/open_chat_studio'
 # DJANGO_DATABASE_PORT=
 
 # Database pool settings
+# DJANGO_DATABASE_USE_POOL=
 # DJANGO_DATABASE_POOL_MIN_SIZE=
 # DJANGO_DATABASE_POOL_MAX_SIZE=
 # DJANGO_DATABASE_POOL_TIMEOUT=
+# DJANGO_DATABASE_CONN_MAX_AGE=
+
+# Applies with DATABASE_URL too. Set to True behind a connection proxy that pins
+# sessions on WITH HOLD cursors (e.g. AWS RDS Proxy).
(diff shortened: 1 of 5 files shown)


SmittieC (approved):
sneaky one that was

PR #4179, opened 2026-08-12
Exposed usage records data to trace,session and chatbot home

### LINKED ISSUE #3716: Expose UsageRecord data in trace view, session view, and chatbot home
## Background

The `cost_tracking` app records a `UsageRecord` row for every LLM call, keyed by `(team, experiment, session, trace, provider_type, model_name, service_kind)`. It captures token counts, unit price, computed cost, and confidence (`EXACT` / `ESTIMATED` / `UNKNOWN`).

Currently this data is only surfaced in two places (both gated by `flag_ai_cost_monitoring`):
- **Dashboard panel** — period spend + top-N chatbots
- **LLM Provider page** — per-model rates and overrides

All the FK linkages exist to expose this much more granularly, but nothing uses them yet.

## What we want

Surface `UsageRecord` data in the three views below. All additions should be gated by `flag_ai_cost_monitoring`.

### 1. Trace view
Each trace already has a `Trace` FK on `UsageRecord`. Show a token/cost breakdown in the trace detail panel — at minimum:
- Total input tokens, output tokens, cached tokens
- Total cost (or "no pricing data" if `pricing_rule` is null)
- Confidence badge if any row is `ESTIMATED` or `UNKNOWN`
- Break down by model if the trace involved multiple models (e.g. router + worker)

### 2. Session view (chat session history)
Each session maps via `UsageRecord.session`. Show a usage summary on the session detail / transcript view:
- Aggregate tokens and cost across all traces in the session
- Could be a collapsed section or a small stats bar at the top
- Per-message breakdown would be a nice-to-have if traces map 1:1 to messages

### 3. Chatbot home (`single_chatbot_home`)
`UsageRecord.experiment` enables per-chatbot roll-ups. Show a lightweight usage widget:
- Total cost this month (or last 30 days)
- Total sessions and message counts with cost
- Could reuse / extend `reporting.cost_summary()` which already aggregates by experiment

## Notes

- All the relevant indexes are in place: `(team, session, timestamp)`, `(team, experiment, timestamp)`, and trace is a direct FK.
- `reporting.py` already has `cost_summary()` and `top_n_bots()` — reuse where possible rather than raw ORM queries in views.
- For trace/session views, a simple `UsageRecord.objects.filter(trace=trace).aggregate(...)` is likely sufficient. Avoid N+1s on session list pages — aggregate at the queryset level.
- The confidence field matters: surface it so users know when token counts are estimated, not exact.
- Consider a shared template partial (e.g. `_usage_summary.html`) that all three views can include, to keep the display consistent.

brian-mecha:
### Product Description

Surfaces AI cost and token-usage data across three more views, gated behind flag_ai_cost_monitoring:

- `Trace detail` - token breakdown (input/output/cached), total cost, confidence badge (flags estimated/missing usage data), and a per-model split when a trace involved multiple models.
- `Session detail` - aggregate token/cost summary across all traces in the session, shown as a compact line in the session details panel.
- `Chatbot home` - a lightweight "last 30 days" widget: spend, session count, and message count.

All three share one partial (cost figure + confidence badge) so the display and copy are consistent everywhere, and reuse the `existing cost_summary()`/`session_usage()`/`trace_token_usage()` aggregation functions rather than new ad-hoc queries.

### Demo

Screenshot 2026-08-13 at 8 07 13 PM

Screenshot 2026-08-12 at 2 29 23 PM

Screenshot 2026-08-13 at 8 09 45 PM

Screenshot 2026-08-14 at 5 35 30 PM

### Docs and Changelog

Fixes #3716


snopoke:
@brian-mecha it looks like there are some AI review comments that aren't resolved yet. I'll review once you've resolved them.

brian-mecha:
@snopoke   AI review comments resolved.

SmittieC:
@brian-mecha thanks for this. Two things from my side:
1. on the chatbot detail page, would you mind making the cost stats a bit smaller please? It feels like it's taking over the page
2. on the trace detail page, we could potentially move the cost stats intobadge like the experiment / session / participant uses.

Thoughts? I'm basically trying to make it not use as much real estate as it is currently doing

brian-mecha:
@SmittieC  Changes effected on the UI to not use as much real estate. Screenshots updated.

templates/cost_tracking/_usage_summary.html
```diff
@@ -0,0 +1,38 @@
...
+    unpriced/estimated/unknown - a bool or a truthy count; either renders the same way.
+      unpriced: some usage has no matching PricingRule ("No pricing data" / "(partial pricing)").
+      estimated: some usage is Confidence.ESTIMATED, not provider-reported.
+      unknown: some usage is Confidence.UNKNOWN (no usage data recorded at all).
+{% endcomment %}
+<div class="flex items-center gap-2 flex-wrap">
+  {% if total_cost %}
+    <span class="font-semibold" data-testid="{{ testid_prefix }}-cost">${{ total_cost|cost_display }}</span>
+    {% if unpriced %}
+      <span class="text-xs text-base-content/40">(partial pricing)</span>
```
snopoke: This should have a tooltip to give the user more info about what it means.

apps/cost_tracking/services/reporting.py
```diff
@@ -339,6 +380,37 @@ def costs_by_experiment(
...
+    page's usage widget. Session/message counts come from `filtered_querysets` - the same canonical,
+    ADR-0051 activity definitions the dashboard's Bot Performance table uses - narrowed to this
+    experiment, rather than re-deriving the session base here.
+
+    That matters because "active in the window" is not "created in the window": `filtered_querysets`
+    counts a session if it has an in-window conversation message (via `Exists`), so a long-lived session
+    (e.g. a WhatsApp thread `channel_base.py` keeps reusing indefinitely) still counts when it was created
+    well before `start` but has turns inside [start, end). A `created_at` filter on the session would drop
+    that session's messages entirely - the join to its messages can't re-admit rows the outer session
+    filter already excluded. `filtered_querysets` also excludes SETUP sessions, per ADR-0051.
```
snopoke: Unnecessary explanation
```suggestion
```

apps/dashboard/views.py
```diff
@@ -79,6 +79,7 @@ def _cost_summary_from_cache(payload: dict) -> CostSummary:
         delta_pct=payload["delta_pct"],
         exact_cost=Decimal(payload["exact_cost"]),
         estimated_cost=Decimal(payload["estimated_cost"]),
+        estimated_call_count=payload["estimated_call_count"],
```
snopoke: I think this will break when loading existing data from the cache:
```suggestion
        estimated_call_count=payload.get("estimated_call_count", 0),
```

snopoke (commented):
Few small comments but mostly looks good. 

Did you consider caching the data for the experiment / session level? I think short caches there would make sense and reduce the query load.

brian-mecha:
@snopoke Tooltip added. Screenshot updated.

- For the chatbot home widget (`experiment-level`), yes. It scans a 30-day window and mirrors exactly what `DashboardCache` already does for the dashboard panel (_same cache model, same TTL pattern_), so wiring `chatbot_usage_summary()` through that with a short TTL (5–10 min, matching how often anyone would notice a stale number there) seems like a win.
- For Session-level, session_usage() is a single query already hitting the (`team`, `session`, `timestamp`) index, scoped to one session's handful of UsageRecord rows. It's about as cheap as a query gets, and it's viewed one session at a time rather than in a hot loop. Caching it would mean extra invalidation logic for a query that's already fast.

snopoke (approved):

snopoke:
> * For the chatbot home widget (`experiment-level`), yes. It scans a 30-day window and mirrors exactly what `DashboardCache` already does for the dashboard panel (_same cache model, same TTL pattern_), so wiring `chatbot_usage_summary()` through that with a short TTL (5–10 min, matching how often anyone would notice a stale number there) seems like a win.

@brian-mecha can you do this in a followup PR please?

Issue #2584, opened 2025-12-15
POC using the chat widget in OCS instead of the current chat UI


snopoke:
Use a feature flag to create a POC of using the chat widget in OCS instead of the current chat UI.

Document any gaps or issues that arise.

PR #4180, opened 2026-08-12
Release flag_chat_widget off the chatbot management UI

snopoke:
### Product Description
Chatting to a bot from the chatbot management pages now happens in the embedded widget popup for everyone, instead of navigating away to the full-page chat UI. This was behind `flag_chat_widget` (POC, #2584).

The versions table gains the same treatment — its per-version chat button opens a widget pinned to that version — and the redundant "new session" button is gone from the chatbots list.

### Technical Description
The flag is released **everywhere except `templates/chatbots/chat/web_chat.html`**, which still gates the widget on the full-page chat. That's the only remaining `{% flag %}`, and `Flags.CHAT_WIDGET` stays defined for it.

Worth review attention:

- **`chatbots:start_authed_web_session` is now unreachable from any template.** The view, route and load test are untouched — removing a public endpoint felt like a separate decision. Its last two callers were the versions table (migrated to the widget) and the chatbots-list button (removed).
- **Every page rendering `ChatbotSessionsTable` must include `session_chat_widget_launcher.html`.** Dropping the non-widget fallback made the Continue Chat button depend on `ocsContinueSessionChat`, and the table is also built in `apps/participants/views.py` and `apps/analysis/views.py` — it threw a `ReferenceError` there until those pages got the include. Tests on both pages now assert the launcher is present.
- **The site support widget is targeted by id (`ocs-support-widget`) rather than by tag.** `{% block app %}` precedes `{% block footer %}`, so once the chatbot home page rendered its own `<open-chat-studio-widget>` unconditionally, `document.querySelector('open-chat-studio-widget')` started matching *that* one — leaving the support bot without page context and sending app context (team slug, breadcrumbs, Django messages) into the user's own chat.
- **Version badge labels are computed server-side** (`_version_label` in `apps/chatbots/tables.py`). The JS default assumed any non-zero version is the working version, which mislabelled older snapshots as "Working Version (v3)". Sessions are always attached to the working row (`start_experiment_session` rejects anything else), so the working version number is already on the select_related'd `experiment` — no extra query.
- **`file_uploads_enabled` is versioned**, so the "Published version" launcher reads it off the published snapshot, not the working row. The view already resolved that object for `deployed_version`.

Known gap left in deliberately: `ocsApplyWidgetVersionTheme` still colours anything non-zero with the working-version (yellow) palette, so an older snapshot shows the right label over the wrong colour. Fixing it properly needs a decision on whether a published snapshot should read as green, which costs a prefetch on three tables.

### Demo
Chatbot home → the chat dropdown and the versions-table chat button open the widget in place. Sessions list → Continue Chat reopens that session in a widget headed with the version it's bound to ("Published Version" / "Working Version (vN)" / "Version N").

Verified in a browser beyond the test suite: widget version binding (a message sent from the v1 launcher created a session with `chat.metadata.experiment_version == 1`), all four label cases, attachment propagation with the working and published rows disagreeing, Continue Chat on the participants page, and that page context lands on the support widget rather than the chatbot's.

### Docs and Changelog

The user-visible change is that chatting from the chatbot management pages happens in the embedded widget popup rather than the full-page chat UI. The full-page chat itself is unchanged and still behind `flag_chat_widget`.


apps/chatbots/tables.py
```diff
@@ -27,26 +26,45 @@ def _show_chat_button(request, record):
...
+        ctxt.update(
+            {
+                "chatbot_id": record.experiment.public_id,
+                "session_external_id": record.external_id,
+                "session_token": issue_session_token(record),
+                "version_number": record.get_experiment_version_number(),
+                "version_label": _version_label(record),
+                "allow_attachments": record.experiment.file_uploads_enabled,
+            }
+        )
```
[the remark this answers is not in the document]
snopoke: Fixed in e67b19a64. `allow_attachments` now comes from `experiment.get_version(version_number)` rather than the working row, with a fallback to the working row when the target version has been archived (the default manager filters those out, so resolving one raises). Added parametrized regression tests over the published alias / older version / working version, plus one for the archived-version fallback.

*— Claude Code*

templates/analysis/detail.html
```diff
@@ -123,4 +123,6 @@ <h2 class="text-lg font-semibold mb-4">Sessions ({{ object.sessions.count }})</h
       </div>
     </div>
   </div>
+  {# The sessions table swapped in above renders Continue Chat buttons, which need ocsContinueSessionChat. #}
```
SmittieC, line 126: Do you generally find these inline comments useful?

SmittieC (approved):

templates/analysis/detail.html
snopoke: I think this one specifically is helpful because if you only look at this template there's no indication of why the include is here.

Issue #2349, opened 2025-10-22
[MEDIUM] Missing API Rate Limiting Exposes System to Abuse


snopoke:
## Summary
No rate limiting is configured for public-facing API endpoints, webhooks, or chat interfaces, allowing unlimited requests that could lead to resource exhaustion, brute-force attacks, and service degradation.

## Severity
**MEDIUM** (CVSS: 6.5)

## Affected Components
- Chat API endpoints (`/api/chat/*`)
- Webhook endpoints (all channels)
- File upload endpoints
- Authentication endpoints (API key usage)
- Public experiment sessions

### Current Configuration

**REST_FRAMEWORK settings (gpt_playground/settings.py:401-415)**:
```python
REST_FRAMEWORK = {
    "DEFAULT_AUTHENTICATION_CLASSES": [
        "apps.api.permissions.ApiKeyAuthentication",
        "apps.api.permissions.BearerTokenAuthentication",
    ],
    "DEFAULT_PERMISSION_CLASSES": [
        "rest_framework.permissions.IsAuthenticated",
        "apps.api.permissions.ReadOnlyAPIKeyPermission",
    ],
    # ❌ Missing: DEFAULT_THROTTLE_CLASSES
    # ❌ Missing: DEFAULT_THROTTLE_RATES
    "DEFAULT_RENDERER_CLASSES": ["rest_framework.renderers.JSONRenderer"],
    "DEFAULT_PARSER_CLASSES": ["rest_framework.parsers.JSONParser"],
    "DEFAULT_SCHEMA_CLASS": "drf_spectacular.openapi.AutoSchema",
    "DEFAULT_PAGINATION_CLASS": "apps.api.pagination.CursorPagination",
    "PAGE_SIZE": 100,
}
```

**No rate limiting on**:
- `apps/api/views/chat.py` - Chat endpoints
- `apps/channels/views.py` - Webhook handlers
- `apps/experiments/views/` - Public experiment access

### 1. API Key Brute Force
```bash
# Attacker attempts to guess valid API keys
for key in $(cat wordlist.txt); do
  curl -X POST https://chatbots.example.com/api/chat/$EXPERIMENT_ID/ \
    -H "Authorization: Bearer $key" \
    -d '{"message": "test"}' &
done
# Without rate limiting, attacker can try millions of keys
```

### 2. Resource Exhaustion via Chat API
```bash
# Flood chat endpoint with expensive LLM requests
for i in {1..100000}; do
[...]

PR #4182, opened 2026-08-12
Rate limit the inbound channel webhooks

barry47products:
### Product Description

Inbound channel traffic (Telegram, Twilio, Meta, Turn, SureAdhere, CommCare Connect, Slack) now counts against its own rate limit budget instead of being unlimited. Log-only, so nothing is rejected yet.

### Technical Description

New `channels` scope at 3000/5m, fail-open, `RATE_LIMIT_ENFORCE` stays `False`. Each route buckets on its own identity so no route can spend another's allowance:

| Route | Identity type | Keyed on |
| --- | --- | --- |
| Telegram | `telegram_channel` | `channel_external_id` |
| Turn | `turn_experiment` | `experiment_id` |
| SureAdhere | `sureadhere_tenant` | `sureadhere_tenant_id` |
| Twilio / Meta / Connect / Slack | `*_ip` | client IP |

**This keying is being reworked, see the review thread on `rate_limit_keys.py:47`.** A provider's source address identifies the provider rather than the tenant, so the four IP-keyed routes currently share one bucket per provider. The counting moves inside the views, after the channel resolves and after the signature check, keyed on `ExperimentChannel.pk`. That also picks up `team_id`, which is `None` on every log line today. The table above describes the branch as it stands.

The cache key is `rl:{scope}:{window}:{identity_type}:{identity}`, so identity type is the only separator. Sharing one type would let the free-text SureAdhere segment spend a Telegram channel's allowance.

Meta answers an over-limit POST with 200 rather than 429, matching every other drop path in that view. WhatsApp retries any non-200 for up to 7 days, so a 429 would buy a week of retries of the traffic that tripped the limiter. Counting still happens. The limiter is on `post` only, leaving the `hub.challenge` GET ungated.

### Migrations

- [x] No migrations.

### Demo

With `RATE_LIMIT_CHANNELS=3/5m` against a local stack: log-only served all six deliveries and logged one `rate_limit.would_block` with `scope=channels`, `identity_type=channel`, identity hashed. With `RATE_LIMIT_ENFORCE=True`, requests 4 and 5 returned 429 with the full `X-RateLimit-*` and `Retry-After` set, and a second channel was unaffected.

### Docs and Changelog

- [x] `RATE_LIMIT_CHANNELS` documented in `.env.example` and `docs/hosting/configuration.md` in this PR.

**Config note:** the scope was named `webhook` earlier in this branch. It now covers inbound channel ingress rather than a delivery mechanism, so it is `channels` and the variable is `RATE_LIMIT_CHANNELS`. Anyone who set `RATE_LIMIT_WEBHOOK` from an earlier build of this branch falls back to the `3000/5m` default. Nothing enforces yet, so the practical effect is nil, which is why the rename happens now rather than after the flip.

---

## Open decision: the API channel views

Chris's answer was (c), on the grounds that `NewApiMessageView` and `NewApiMessageVersionedView` are the API channel and belong with the other channels. One finding since then is worth weighing before it is settled.

Those two views are already counted. `DEFAULT_THROTTLE_CLASSES` is `APIRateThrottle` (`config/settings.py:452`) and neither view overrides `throttle_classes`, so both have been under `api` at 2000/5m since #4170. Moving them is a relocation, not new coverage, and the question is which number they sit under.

They also cost something different from the rest of the scope. Every other route resolves a channel, calls `.delay()` and returns, so the limit bounds chat queue depth. `_new_api_message` calls `tasks.handle_api_message` directly and returns the reply inline, so each request holds a web worker for an LLM round trip. Calibration will see far more WhatsApp and Telegram volume than API chat calls, so a single `channels` threshold gets set by enqueue-shaped traffic and then applied to a route bounded by worker concurrency.

That is an argument about the number rather than the grouping. Either works:

- **(d)** its own scope, named as part of the channel family, with a rate matched to what it consumes.
- **(c)** into `channels`, provided the rate is set from this route rather than from webhook volume.
- **(a)** leave them in `api`, where they already are.

## Second open decision: does this scope ever refuse?

`check()` does one thing once `RATE_LIMIT_ENFORCE` flips: over limit means `allowed=False`. Once counting moves inside `_dispatch_message_values`, that means skipping the `.delay()` for a message value, so a participant's WhatsApp message is dropped while Meta still receives its 200.

Counting without ever refusing is the safer behaviour for this scope and needs a small addition to the limiter core. Worth agreeing before the keying rework is built, since it decides whether this scope measures channel load or sheds participant messages.

## Before enforcement is switched on

- **Set `RATE_LIMIT_TRUSTED_PROXY_COUNT`.** With the default of 0, `client_ip()` returns the load balancer's address. The keying rework removes most of the exposure here, but `generate_key` stays IP-keyed by necessity.
- **Delivery-status callbacks are counted then discarded,** so observed volume overstates what enforcement would block.
- **Over-limit responses need a per-provider decision.** Twilio does not retry by default, so a 429 loses the notification; Slack, Meta and Telegram all retry, so a 429 regenerates the traffic. Meta is the only one decided here, and the count-without-refuse decision above may settle the rest.

## Follow-ups, to file separately

`X-RateLimit-*` headers on non-429 responses give an anonymous caller a cross-tenant traffic oracle. `generate_key`'s `httpx.get` sets no timeout where `CommCareConnectClient` sets `timeout=10`, so a hung upstream pins a worker regardless of any limit. The inbound email webhooks stay unlimited: `AmazonSESInboundWebhookView` and `MailgunInboundWebhookView` come from anymail's URLconf (`config/urls.py`), so limiting them needs the same shadowing approach #4184 uses for the OAuth views.

`is_exempt` sitting outside the fail-open try/except was filed here and is fixed in #4184.

Part of #2349


apps/channels/views.py
```diff
@@ -50,13 +59,15 @@
...
 from apps.teams.utils import set_current_team
+from apps.utils.rate_limit import rate_limited
 from apps.web.waf import WafRule, waf_allow
 
 log = logging.getLogger("ocs.channels")
 
 
 @waf_allow(WafRule.NoUserAgent_HEADER)
 @csrf_exempt
+@rate_limited(WEBHOOK_SCOPE, key_fn=channel_external_id_key)
```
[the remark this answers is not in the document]
barry47products: @coderabbitai Confirmed and fixed in 7a13461fd021916e0221f9ca205bfaa9c7c4ba42.
A GET to the Telegram webhook was reaching the secret-token check inside the view and returning 400, having already spent the channel's allowance. `@require_POST` now sits above `@rate_limited`, matching `new_connect_message`. The parametrised non-POST test covers Telegram.

apps/slack/views.py
```diff
@@ -40,6 +42,7 @@ def slack_oauth_redirect(request):
 
 @waf_allow(WafRule.SizeRestrictions_BODY)
 @csrf_exempt
+@rate_limited(WEBHOOK_SCOPE, key_fn=slack_ip_key)
```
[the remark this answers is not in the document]
barry47products: @coderabbitai Confirmed and fixed in 7a13461fd021916e0221f9ca205bfaa9c7c4ba42 with `@require_POST` above `@rate_limited`. A GET previously fell through `slack_bolt`'s handler to its 404 branch, after the
limiter had counted it. `slack_bolt` only uses the GET branch for its own `install_path` / `redirect_uri_path`, and neither resolves to `slack/events` (OCS routes those to `slack_install` and `slack_oauth_redirect`), so POST-only is the full surface of this route. Slack is in the non-POST test now.

apps/utils/tests/test_rate_limit.py
```diff
@@ -450,6 +450,13 @@ def test_public_chat_scope_is_configured():
     assert fail_open is True
 
 
+def test_webhook_scope_is_configured():
+    """The webhook scope parses to the design's 3000/5m, fail-open."""
+    limit, window_seconds, fail_open = _scope_config("webhook")
+
+    assert (limit, window_seconds, fail_open) == (3000, 300, True)
```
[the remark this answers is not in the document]
barry47products: @coderabbitai Fixed in 3bcb33050a2b0aed2f4952866daa91c94192dfc4. The test now asserts a positive limit, a positive window and `fail_open is True`, matching `test_admin_api_scope_is_configured`, `test_chat_api_scope_is_configured` and `test_public_chat_scope_is_configured`.

docs/hosting/configuration.md
```diff
@@ -65,6 +65,7 @@ These apply whether the connection comes from `DATABASE_URL` or the variables ab
...
+| `RATE_LIMIT_WEBHOOK` | `3000/5m` | Request limit for the `webhook` scope (inbound channel deliveries: Telegram, Twilio, Meta Cloud API, Turn, SureAdhere, CommCare Connect, Slack events). Keyed by the channel identifier the webhook URL carries (Telegram's `channel_external_id`, Turn's `experiment_id`, SureAdhere's `sureadhere_tenant_id`), then by client IP for webhooks whose URL carries no identifier. Set `RATE_LIMIT_TRUSTED_PROXY_COUNT` before enforcing behind a proxy, or those IP-keyed callers all share one bucket. Fails open: if the limiter's cache is unreachable, requests are served. |
```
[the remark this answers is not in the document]
barry47products: @coderabbitai Not applying this one. `docs/agents/adr_process.md` scopes the field to design docs specifically: "Design docs produced by the brainstorming skill (`docs/design/`, `docs/superpowers/specs/`) carry a `status` frontmatter field", and gives it lifecycle semantics: `active` gates ADR extraction off, `stable` means the decisions are settled and safe to extract, `extracted` means they already are. Every file in the repo carrying the field is such a doc.

`docs/hosting/configuration.md` is a published reference page, listed in `mkdocs.yml` under Hosting → Configuration. Marking it `stable` would assert it is a settled design source awaiting `/extract-adrs`, which is not what it is. The same doc records that ADRs themselves do not carry the field either, using the pill from `docs/adr/_template.md` instead.

barry47products:
@snopoke or @SmittieC, there's one open decision in the PR body that needs to be decided before the PR is complete.

apps/channels/rate_limit_keys.py
```diff
@@ -0,0 +1,65 @@
...
+provider address; a route whose URL carries no identifier buckets on the caller's
+address under an identity type of its own. A route reached without the identifier
+its URL normally carries falls back to the plain address bucket.
+"""
+
+from django.http import HttpResponse
+
+from apps.utils.rate_limit import RateLimitResult, client_ip, rate_limited
+
+WEBHOOK_SCOPE = "webhook"
```
SmittieC: To your open question in the PR description:

The reason for having `NewApiMessageView` and `NewApiMessageVersionedView` as a channel view is because they are seen as the "api" channel. In light of this, I am wondering if WEBHOOK_SCOPE should not be renamed to CHANNELS_SCOPE.

apps/channels/rate_limit_keys.py
```diff
@@ -0,0 +1,65 @@
...
+    """Builds a key function bucketing on the caller's address under `identity_type`."""
+
+    def key_fn(request, *args, **kwargs):
+        return identity_type, client_ip(request)
+
+    return key_fn
+
+
+twilio_ip_key = _client_address_key("twilio_ip")
+meta_ip_key = _client_address_key("meta_ip")
```
SmittieC: From what I gather from a quick search on the the interwebs, the IP from which the WA message comes might differ between webhook calls (I'm sure it's the same for the others?).

Perhaps in Meta's case we could use the [unique_phone_ids](https://github.com/dimagi/open-chat-studio/blob/7a13461fd021916e0221f9ca205bfaa9c7c4ba42/apps/channels/views.py#L550) which correspond to `ExperimentChannel.extra_data__phone_number_id`. In most cases there will be a single entry in there. If there are multiple, we can ignore a rate limited one. The penalty is that we parse before limiting, but I think that's fine, considering that we don't want tenant X to influence tenant Y.

For twilio, we could use the `message.to` field, which will mean that we'd have to move the rate limit logic into the view function. I think connect has a single IP, so that should be fine

apps/channels/rate_limit_keys.py
barry47products: Agreed on the rename. `WEBHOOK_SCOPE` becomes `CHANNELS_SCOPE`, `RATE_LIMIT_WEBHOOK` becomes `RATE_LIMIT_CHANNELS`. Nothing enforces it yet, so the key prefix moving from `rl:webhook:` to
`rl:channels:` is free.

The grouping I'd leave to you, but one thing worth knowing first: those two views are already counted. `DEFAULT_THROTTLE_CLASSES` is `APIRateThrottle` (`settings.py:452`), and neither view overrides `throttle_classes`, so they've sat under `api` at 2000/5m since #4170. Moving them isn't adding coverage. It's choosing a different number for them.

And they do cost something different from the rest. Everything else here resolves a channel, calls `.delay()`, and returns. That bounds queue depth. `_new_api_message` calls `tasks.handle_api_message` directly and hands back the model's reply inline, so the request holds a web worker for a full LLM round trip. We'll see far more WhatsApp and Telegram volume than API chat calls, so one `channels` number ends up being set by enqueue traffic and then applied to a route bounded by worker concurrency.

That's an argument about the number though, not about the grouping. Yours stands on its own: the API channel is a channel in the core concepts list, and naming scopes after the domain will age better than naming them after today's cost shape, which changes the day this route moves to Celery.

So (d) if you want the channel family visible in the config, with its own rate. Or (c), as long as we set that rate from this route rather than from webhook volume. Whichever reads better in `RATE_LIMIT_SCOPES`.

apps/channels/rate_limit_keys.py
barry47products: Agreed. I'm moving the count into the view after the channel resolves, keyed on `ExperimentChannel.pk` rather than on `phone_number_id` and `message.to`.

The reason is ordering. Neither view can check a signature before it has resolved the channel: Meta's `app_secret` lives on the resolved channel's provider, and `validate_twillio_request` (`views.py:115`) runs after `get_experiment_channel` (`:104`). Keyed on payload content at the door, the count fires before anything has authenticated the payload, so anyone supplying a real identifier drives that tenant's bucket. Resolving first also gets us `team_id`, which is `None` on every line today.

For Meta, that lands inside `_dispatch_message_values`, once per message value after the `not ch` and non-conversational skips, so a multi-number batch bills each tenant separately. Connect resolves a channel too (`views.py:341`), so it keys the same way. Slack can't resolve until Bolt has dispatched, so its count goes in the `new_message` listener (`slack_listeners.py:31`).

One thing I'd like your read on before I build it. Once `RATE_LIMIT_ENFORCE` flips, `check()` does exactly one thing when you're over: `allowed=False`. Inside that Meta loop, that means skipping the `.delay()`, so a participant's WhatsApp message disappears, and Meta still gets its 200. I'd rather this scope counted and never refused, which needs a small addition to the core. It's the difference between measuring channel load and quietly dropping people's messages.

barry47products:
@SmittieC or @snopoke, I'd like to pull `generate_key` (`apps/api/views/channels.py:42`) into this PR rather than leave it as a follow-up. It checks that an `Authorisation` header exists (`:44`) and nothing else, then makes an outbound `httpx.get` to CommCare Connect (`:49`), whose response is what tells us whether the token was any good (`:51`). So any non-empty header makes us send a request that lands at CommCare Connect under our name.

It has to be IP-keyed at the door. `channel_id` is in `request.POST` before the call, but the caller supplies it, so keying on that lets someone drain a known channel's budget, or slip the counter entirely by varying it.

Two, I'd rather you called:

- First, which scope: #4184 adds an IP-keyed `credentials` scope this could join, except that one is fail-closed, so a Redis outage would stop Connect key exchange completely. A separate fail-open scope avoids that and costs one more thing to calibrate.
- Second, whether `callback` comes along. It's HMAC-gated, and its body is a bare `return HttpResponse()`, so there's nothing there to bound, and I'd leave it out. `consent` is HMAC-gated too, but does a read and a write, so I've put it in `channels`, counted it, and not refused.

apps/channels/rate_limit_keys.py
barry47products: Renamed pushed in df8d420

snopoke:
> @SmittieC or @snopoke, I'd like to pull `generate_key` (`apps/api/views/channels.py:42`) into this PR rather than leave it as a follow-up. It checks that an `Authorisation` header exists (`:44`) and nothing else, then makes an outbound `httpx.get` to CommCare Connect (`:49`), whose response is what tells us whether the token was any good (`:51`). So any non-empty header makes us send a request that lands at CommCare Connect under our name.
> 
> It has to be IP-keyed at the door. `channel_id` is in `request.POST` before the call, but the caller supplies it, so keying on that lets someone drain a known channel's budget, or slip the counter entirely by varying it.
> 
> Two, I'd rather you called:
> 
> * First, which scope: [Rate limit credential endpoints and failed API-key attempts #4184](https://github.com/dimagi/open-chat-studio/pull/4184) adds an IP-keyed `credentials` scope this could join, except that one is fail-closed, so a Redis outage would stop Connect key exchange completely. A separate fail-open scope avoids that and costs one more thing to calibrate.

I think it's OK to fail closed here.

> * Second, whether `callback` comes along. It's HMAC-gated, and its body is a bare `return HttpResponse()`, so there's nothing there to bound, and I'd leave it out. `consent` is HMAC-gated too, but does a read and a write, so I've put it in `channels`, counted it, and not refused.

I agree with these choices.

barry47products:
Thanks @snopoke, both are done.

`generate_key` landed in #4184 rather than here, since that is where the `credentials` scope is defined, and this PR is its base. Keyed at the door by client IP, fail-closed with the rest of the scope. The test that matters asserts no outbound call is made once the bucket is spent, because the limit exists to bound the call rather than the view. `callback` left out.

`consent` is the one still to come. Counting it without refusing needs the limiter core to be able to express that, which it could not: `check()` had one behaviour once `RATE_LIMIT_ENFORCE` flips, and that is `allowed=False`. A scope can now set `refuse: False` alongside `fail_open`, which counts and logs but never turns a caller away, and holds through both the enforcement flip and a backend outage, so the two settings cannot combine into a dropped message. That is pushed here. No scope sets it yet.

Setting it on `channels` is the open question from Chris's thread, since it decides the same thing for every provider webhook, not just `consent`. Meta is the sharp case: with counting moved inside `_dispatch_message_values`, refusing means skipping a `.delay()`, so a participant's WhatsApp message disappears while Meta still gets its 200. Your answer on `consent` points the same way, and I would rather make it once for the scope than differently per route. @SmittieC, does that match your read?

apps/channels/rate_limit_keys.py
snopoke: > One thing I'd like your read on before I build it. Once RATE_LIMIT_ENFORCE flips, check() does exactly one thing when you're over: allowed=False. Inside that Meta loop, that means skipping the .delay(), so a participant's WhatsApp message disappears, and Meta still gets its 200. I'd rather this scope counted and never refused, which needs a small addition to the core. It's the difference between measuring channel load and quietly dropping people's messages.

I think this is correct. I think it is better to accept the message and log the over usage warning than to ignore / block the message entirely. Once logging is in place we can create log based alerts for monitoring and alerting.

snopoke (approved):

PR #4184, opened 2026-08-12
Rate limit credential endpoints and failed API-key attempts

barry47products:
### Product Description

Guessing a credential is currently free. This puts a limit on it.

One `credentials` scope covers two surfaces, both keyed by client IP, since a caller who fails authentication has no identity to key on:

- The OAuth client-credential endpoints: `/o/token/`, `/o/revoke_token/`, `/o/introspect/`
- API requests whose key or bearer token is rejected. Successes count under `RATE_LIMIT_API`, so a working integration never lands in this bucket.

Log-only, as with the rest of the series. `RATE_LIMIT_ENFORCE` stays `False`, so over-limit requests are served and logged as `rate_limit.would_block`. This is the last surface; only the docs and enforcement PR is left after it.

### Technical Description

Part of #2349 . Stacked on #4182, so the base branch is `bt/rate-limit-webhook-scope-2140`.

`credentials` is registered at `100/5m` with `fail_open: False`. It is the only fail-closed scope in the design, and the core's fail-closed branch has had nothing configured against it until now.

**Fail-closed means an OAuth outage, and the exemption flag cannot rescue it.** When the limiter's cache is unreachable and enforcement is on, `check()` refuses. For failed API-key attempts that costs nothing: those requests were being rejected anyway, so a 401 becomes a 429. For the OAuth endpoints it means no client can get or refresh a token until Redis is back. `flag_ignore_rate_limiting` is no help, because resolving the flag reads the same Redis and now answers "not exempt" (see below). That leaves `RATE_LIMIT_ENFORCE=False`, an env var needing a restart.

With Redis stopped and enforcement on:

```
HTTP/1.1 429 Too Many Requests
Retry-After: 300
{"detail": "Rate limit exceeded.", "available_in": 300}
```

Refused, `Retry-After` set, `X-RateLimit-*` absent because the result is degraded. That header split in `RateLimitHeadersMiddleware` shipped with #4170 and has been unreachable since. This is the first time anything has taken it.

The design signed off on fail-closed for the scope as a whole, so that is what is here. If the OAuth consequence is too sharp, the fallback is to split it: fail-closed `credentials` for failed attempts, fail-open `token` for issuance. Cheap to do in this PR rather than after. Either way nothing is at risk before the enforcement flip.

**`is_exempt` is hardened, which widens this PR into the shared core.** `rate_limited` calls `is_exempt` before the `check()` whose try/except gives every scope its failure mode. `is_exempt` resolves a Waffle flag against the default cache, which has no socket timeout configured, so during a Redis outage it could raise before the limiter ran at all and turn the request into a 500. For a fail-open scope that turns a served request into an error. For a fail-closed one it breaks the contract in exactly the case the contract was written for: a 500 is neither refused nor counted.

It now catches, logs `rate_limit.exemption_check_error`, and answers `False`. Fail-open scopes carry on to `check()` and are allowed as before; fail-closed ones carry on and are refused as designed. Nothing gets more permissive. This was filed as a follow-up on #4182, but the guarantee this PR rests on does not hold without it, so it is folded in. The Redis-down run above logged both `rate_limit.backend_error` and `rate_limit.exemption_check_error`, which is this path being taken.

**The OAuth routes are shadowed at the URLconf.** All three views are third-party, so there is nothing local to decorate. Each gets a `path()` declared before `include(oauth2_urls.base_urlpatterns)`, shadowing the upstream route and leaving every other OAuth endpoint alone. Names and namespace are unchanged, so `reverse()` is unaffected and `resolve()` returns the shadowing route by declaration order. `csrf_exempt` survives: `as_view()` copies `cls.dispatch.__dict__` onto its output and `rate_limited`'s `@wraps` copies it again. Checked on all three.

`revoke_token` and `introspect` went in after review, not in the original design. Both authenticate the same `client_id` and `client_secret` that `/o/token/` bounds, so limiting only `token` left a brute-forcer able to change one path segment and get the same signal unmetered.

Last thing worth flagging: the exemption flag is deliberately not consulted inside the authentication class. `flag_is_active` can read `request.user`, and reading `request.user` from inside an authentication class re-enters authentication. A rejected key has no team for a team-scoped exemption to apply to anyway, so the check is skipped rather than made re-entrant. The global kill switch still works, since setting the flag for everyone is checked on the `api` scope that all successful traffic takes, and `RATE_LIMIT_ENFORCE` is still the outer switch.

#### Two I left alone, for your call

**Rejected embed keys count in no scope at all.** `/api/chat/*` uses `EmbeddedWidgetAuthentication`, and DRF authenticates before it throttles, so an `AuthenticationFailed` short-circuits `initial()` and `ChatAPIRateThrottle` never runs. Those attempts land in neither `chat_api` nor `credentials`. Someone can probe `X-Embed-Key` values as fast as they like and read a clean answer off the 401. Different auth class, different app, and `chat_api` was #4174's ground, so I have not widened this PR again. Worth deciding whether it blocks closing #2140, given this is meant to be the last scope.

**`X-RateLimit-Remaining` goes out on 401s.** Someone brute-forcing reads their exact remaining budget and reset time off each rejection, so they can pace just under the threshold instead of probing for it. That saves them a probe rather than handing them anything, and normal limiters do expose these headers. Still, on this scope the reader is unauthenticated by definition. One-line change to suppress them on this path if you'd rather.

### Migrations

None.

### Demo

Local stack, `RATE_LIMIT_CREDENTIALS=3/5m`.

Log-only, five requests with a rejected key:

```
401 401 401 401 401
```

One `rate_limit.would_block`, sampled after the first crossing. Redis shows the scope and keying: `rl:credentials:<window>:ip:127.0.0.1`.

Enforcement on, same requests:

```
401 401 401 429 429
```

```
HTTP/1.1 429 Too Many Requests
Retry-After: 73
X-RateLimit-Limit: 3
X-RateLimit-Remaining: 0
X-RateLimit-Reset: 73
{"detail":"Rate limit exceeded.","available_in":73}
```

`POST /o/token/` with a `client_credentials` grant gives the same progression and body. The 401s come from the upstream views; the limiter counts them and takes over on the fourth. The Redis-stopped run is above.

### Docs and Changelog

`RATE_LIMIT_CREDENTIALS` is in `.env.example` and the hosting configuration reference: what the scope covers, all three OAuth endpoints, the IP keying, the `RATE_LIMIT_TRUSTED_PROXY_COUNT` precondition, and the failure mode. Rendered page checked locally.


.env.example
barry47products, line 1: @SmittieC, it feels like a lot of the comments here should go in the documentation instead. Thoughts?

snopoke: Looks like we already have docs for most of it so this is duplicating. I do like having the docs close to the usage but the duplication isn't great. I'm OK with moving the bulk of it and adding a reference to the docs in this file.

snopoke (dismissed):

.env.example
barry47products: Perfect, I have pushed up that change as part of the docs PR #4194

snopoke (approved):

PR #4185, opened 2026-08-13
fix: Generate provider test factory secrets per instance

barry47products:
### Product Description

Test-only change. `main` fails intermittently on `apps/api/v2/inspect/tests/test_secrets_exclusion.py`; this makes that guard depend only on real leaks.

### Technical Description

The ADR-0027 guard searches the whole serialized projection for each provider config value. Four factories used values short enough to appear there by coincidence: `"123"` in the LLM, voice and trace factories matches the provider's own primary key, and `"key"` in `AuthProviderFactory` is a substring of `AuthProviderType.api_key`.

Each secret now comes from a `factory.Sequence`, so it stays distinct from anything the projection can contain. `factory.Dict` also gives each instance its own dict; these configs were class attributes shared across every provider a factory built. Non-secret fields (`aws_region`, langfuse `host`) keep their fixed values.

`test_secrets_exclusion.py` and `apps/service_providers` pass (763 tests); the full suite passed at 2683.

### Migrations

None.

### Demo

N/A, test factory change with no runtime surface.

### Docs and Changelog

### DIFF
--- a/apps/utils/factories/service_provider_factories.py
+++ b/apps/utils/factories/service_provider_factories.py
@@ -35,7 +35,7 @@ class LlmProviderFactory(factory.django.DjangoModelFactory):
     team = factory.SubFactory(TeamFactory)
     type = str(LlmProviderTypes.openai)
     name = factory.Sequence(lambda n: f"Test LLM Provider {n}")
-    config = {"openai_api_key": "123"}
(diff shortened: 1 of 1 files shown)


barry47products:
@snopoke or @SmittieC, this is a small fix for the flaky test.

SmittieC (approved):

PR #4187, opened 2026-08-13
Show record counts on tables

### LINKED ISSUE #4186: Show record counts
We need to show the number of records for each table in the UI so users can see how many records there are. This came from a user who wanted to see the number of records that a filter matched

floze-the-genius:
### Product Description

Standard table pages now show the total number of matching records below the table. The count remains visible for empty and single-page results, uses singular/plural labels, and updates with applied filters.

### Technical Description

The shared django-tables2 Tailwind footer now renders the normal paginator's exact count. Lazy-paginated tables keep their existing behavior and do not request an expensive exact count.

Template tests cover zero, singular, single-page, multi-page, and lazy-paginator behavior.

Local verification:

- `uv run ruff check`
- `uv run ty check`
- `uv run prek run --files templates/table/tailwind.html apps/generics/tests/test_table_templates.py`
- `uv lock --check`
- `uv run pytest apps/generics/tests apps/chatbots/tests/test_chatbot_views.py -q` — 37 passed
- `pnpm test` — 18 passed
- `pnpm type-check`
- `pnpm build` on Node 24
- Browser smoke: the All Sessions table showed 5 records; filtering to Customer Support Bot showed two rows and 2 records, with no browser console errors.

### Demo

1. Open a standard table such as **All Sessions**.
2. Confirm the footer shows the total record count.
3. Apply a table filter and confirm the footer updates to the number of matching rows.

### Docs and Changelog

Closes #4186

### DIFF
--- a/apps/generics/tests/test_table_templates.py
+++ b/apps/generics/tests/test_table_templates.py
@@ -1,0 +1,53 @@
+import django_tables2 as tables
+from django.template import Context
+from django.template.loader import get_template
+from django.test import RequestFactory
+from django_tables2 import RequestConfig
+from django_tables2.paginators import LazyPaginator
+
+
+class NameTable(tables.Table):
+    name = tables.Column()
+
+
+def _render_table(names, *, per_page=20, lazy_pagination=False):
+    request = RequestFactory().get("/")
+    table = NameTable([{"name": name} for name in names])
+    paginate = {"per_page": per_page}
+    if lazy_pagination:
+        paginate["paginator_class"] = LazyPaginator
+    RequestConfig(request, paginate=paginate).configure(table)
+    template = get_template("table/tailwind_js_pagination.html").template
+    return template.render(Context({"table": table, "lazy_pagination": lazy_pagination, "request": request}))
+
+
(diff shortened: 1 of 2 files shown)


floze-the-genius:
@coderabbit review

snopoke:
@floze-the-genius thanks or your contribution!

Please could add a screenshot showing what this looks like and also list out which tables this change impacts. We have specifically removed counting from some tables due to the performance cost.

snopoke (commented):
@floze-the-genius I took a quick look here and I see you've made changes to the base tailwind table template but we only ever use `tailwind_js_pagination.html` which overrides the pagination block.

Looking that that file now I can see that there's a bug, the `if lazy_pagination` switch is around the `{% block %}` tag instead of inside it. This is a bug because Django evaluates `block` tags first so the block is always overridden regardless of the 'lazy' flag (https://docs.djangoproject.com/en/5.2/ref/templates/language/#template-inheritance).

I think the actual fix is to adjust the code in `tailwind_js_pagination.html`.

SmittieC:
Superseded by https://github.com/dimagi/open-chat-studio/pull/4255

PR #4189, opened 2026-08-13
Restructure dev docs around a development workflow page

snopoke:
### Product Description
No change — developer docs, plus one new invoke task.

### Technical Description
The getting-started docs described commands but never the loop we actually work in, and the AI-assisted development content was spread across three pages that overlapped and cross-linked in circles.

`common-tasks.md` becomes `dev-workflow.md`: worktree → agent session → logical commits → `/code-review` → draft PR → `/iterate-pr` → hand off. `ai_development.md` is folded into its step 2 (it duplicated the design-before-code phases), `ai-setup.md` is trimmed to setup only, and `claude_code_agent.md` is renamed `claude_github_automation.md` since it only ever covered the GitHub Actions side. Old command reference survives as "Everyday commands" at the bottom of the new page.

Worth a look:

- **`inv typecheck`** (`tasks.py`) is the only code here. It runs `ty` and `tsc`, and deliberately runs both even when the first fails so one invocation reports every type error. Exits non-zero naming which checkers failed — verified by planting a bad annotation in a `.py` and a `.ts` file. `AGENTS.md` now points at it and at `inv ruff` rather than the underlying tools.
- **All 13 `PROPOSED` ADRs are now `ACCEPTED`** (commit 2), in both the per-file status pill and the index table. There were no `REJECTED` or `SUPERSEDED` records in the set, so nothing recording a reversal was overwritten. Shout if any of those were still genuinely open.
- `contributing/pull_requests.md` had human reviewers being assigned *before* AI review; reordered to match how we actually work.

Content verified against the repo rather than memory: `dev-utils` no longer ships `resolve-ci-failures` / `resolve-pr-comments` and `plugins/README.md` is a 404, so those references are gone; superpowers now installs from the Dimagi marketplace that `.claude/settings.json` already registers.

One claim I could not verify: that Sentry posts a second-round AI review alongside Claude. The only Sentry reference in the repo is `allowed_bots` in `claude-code-review.yml`, which is about reviewing Seer-*authored* PRs. Left in on the basis that it happens in practice — remove if not.

### Migrations
N/A

### Demo
N/A — docs. `zensical build` is clean and every internal link added here resolves.

### Docs and Changelog

Developer docs only; no user-facing change to log.


SmittieC (dismissed):
Scanned it only, but LGTM

SmittieC (approved):

lisa-tarbo:
@snopoke appreciate this as it answers many of my questions about the process you are using

PR #4192, opened 2026-08-13
Show and filter session last activity with a created_at fallback

### LINKED ISSUE #4190: Set last_activity_at on session creation (e.g. trigger_bot_api)
## Problem

When a session is created via `trigger_bot_api` (or other creation paths), `last_activity_at` is never set at creation time. If the user never sends a message to the session, `last_activity_at` remains `null` indefinitely — it is only populated once actual **human** message activity occurs.

## Expected Behaviour

`last_activity_at` should be set at session creation time so that even sessions with no subsequent user messages have an accurate activity timestamp.

## Suggested Fix

Set `last_activity_at = now` when a session is created across all session creation paths.

## Impact

- Sessions created but never messaged will have `last_activity_at = null` forever.
- Any ordering, filtering, or TTL logic based on `last_activity_at` will treat these sessions as if they never had any activity.
- Could affect session expiry, dashboards, or any feature that relies on this field to determine recency.

snopoke:
### Product Description
The sessions table's **Last activity** column no longer renders blank for sessions whose participant never sent a message (for example sessions created via `trigger_bot_api`) - it shows the session's creation time instead, and those sessions now sort by that time instead of dropping to the bottom of the list.

The filter panel gains a **Last Activity** filter over the same value, and the two message-based filters are relabelled **Last Participant Message** and **First Participant Message** so it is clear they only match sessions where the participant actually wrote something.

Closes #4190

### Technical Description
`last_activity_at` is only written by an inbound human message (`apps/channels/signals.py`), so it is null for bot-initiated and legacy sessions. Rather than seeding it at creation (the approach in the earlier attempt on this issue - that changes what the field *means* for `MetaCloudAPIService._is_within_service_window` and for session-token expiry), the fallback lives at the points that present the value:

- `ExperimentSession.last_activity` (`apps/experiments/models.py`) returns `last_activity_at or created_at` and backs the table column; `last_activity_expression()` is its `Coalesce` order-by equivalent.
- `ChatbotSessionsTable` - the column is renamed `last_message` to `last_activity` (its header already read "Last activity"), and `order_last_activity` sorts by the expression. `get_table_queryset` uses the same expression for its default ordering, so the display and the sort agree.
- `LastActivityFilter` (`apps/experiments/filters.py`) filters on the same fallback using two OR'd `Q` branches rather than a `Coalesce` annotation: each branch can use the existing column indexes, and two filters on one column (the UI expresses a date range as `after X` plus `before Y`) do not collide on an annotation alias.

Two changes beyond the three requested items, easy to drop if you would rather not take them:

1. **`expsession_team_lastact_c_idx`** (migration `0149`). Ordering by the coalesced expression means `expsession_team_lastact_idx` no longer serves the default sort, which would leave large teams sorting their whole session set on every page. The new functional index on `(team, Coalesce(last_activity_at, created_at) DESC)` restores that. `AddIndexConcurrently`, matching migrations `0134` and `0146`.
2. **`date_range_column`** for the session filter moves from `last_message` to `last_activity`, so the quick "Last 7 Days" dropdown keeps sessions the participant never replied to. That constant also feeds the help agent's prompt, so the three generic-recency fixtures in `apps/help/evals/fixtures/filter.yml` ("within the last week", "from the last 30 days", "in the last hour") now expect `last_activity`. I could not run the evals (they need `SYSTEM_AGENT_MODELS_*`), so those expectations are reasoned, not verified.

Existing saved filters and URLs using `f_last_message` keep working - that filter is unchanged apart from its label.

### Migrations

`0149_expsession_team_lastact_c_idx` adds an index concurrently. No column or data changes, and rows with a null `last_activity_at` are read through the fallback rather than backfilled.

### Demo
No new UI surface. Verified at the view/HTML level rather than in a browser: `test_chatbot_sessions_table_view_last_activity_filter` drives `chatbots:sessions-list` over HTTP and asserts the filter selects the never-messaged session and that the rendered time cell carries its `created_at`.

New tests: `TestSessionsTableLastActivityColumn` (cell value for both cases, default ordering, click-sort in both directions), `TestLastActivityFilter` (every operator, the relative `range` fallback, no row multiplication, `after` plus `before` on one column), and a filter-label assertion in `test_dynamic_filters.py`.

Full suite on a fresh DB: **6144 passed, 2 skipped**. Ruff and `ty` clean on the changed files.

### Docs and Changelog

### DIFF
--- a/apps/chatbots/tables.py
+++ b/apps/chatbots/tables.py
@@ -9,7 +9,7 @@ from django.urls import reverse
 from django_tables2 import columns
 
 from apps.api.session_tokens import issue_session_token
-from apps.experiments.models import Experiment, ExperimentSession
+from apps.experiments.models import Experiment, ExperimentSession, last_activity_expression
 from apps.generics import actions, chips
 from apps.generics.actions import chip_action
 from apps.generics.tables import ArrayColumn, ColumnWithHelp, TimeAgoColumn
@@ -177,7 +177,7 @@ class ChatbotSessionsTable(tables.Table):
         accessor="message_count",
         orderable=True,
     )
-    last_message = TimeAgoColumn(accessor="last_activity_at", verbose_name="Last activity", orderable=True)
+    last_activity = TimeAgoColumn(accessor="last_activity", verbose_name="Last activity", orderable=True)
     tags = columns.TemplateColumn(verbose_name="Tags", template_name="annotations/tag_ui.html")
     versions = ArrayColumn(verbose_name="Versions", accessor="experiment_versions")
     state = columns.Column(verbose_name="State", accessor="status", orderable=True)
@@ -199,6 +199,15 @@ class ChatbotSessionsTable(tables.Table):
         ],
         align="right",
     )
+
+    def order_last_activity(self, queryset, is_descending):
+        """Sort by the same coalesced expression the column renders.
+
+        Without this, django-tables2 would order by the `last_activity` accessor, which is a
+        model property and not a database field.
+        """
+        order = last_activity_expression()
+        return queryset.order_by(order.desc() if is_descending else order.asc()), True
 
     def render_tags(self, record, bound_column):
         template = get_template(bound_column.column.template_name)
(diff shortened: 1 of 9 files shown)


SmittieC (approved):
Looks fine to me

PR #4198, opened 2026-08-14
Pin OAuth applications to the chatbots they may reach

### LINKED ISSUE #4197: Pin OAuth applications to the chatbots they may reach
A client-credentials OAuth application can currently converse with **every** chatbot in its team. `chatbots:interact` is team-scoped, and none of the endpoints it guards require anything chatbot-specific:

| Endpoint | Resolves | Capability |
|---|---|---|
| `_chat_completions` — `apps/api/openai.py:130` | path `experiment_id` | chat with any chatbot in the team |
| `_new_api_message` — `apps/channels/views.py:259` | path `experiment_id` | chat with any chatbot in the team |
| `handle_trigger_bot_message` — `apps/api/views/channels.py:150` | body `experiment` | send outbound WhatsApp/Telegram/Connect messages to arbitrary participants |

Add `OAuth2Application.allowed_chatbots` so an application reaches only the chatbots it was authorised for. **Empty means none** — an application authorises nothing until someone says so.

This is a prerequisite for OAuth on the chat widget (#3893), where the token is handed to a browser *by design* and the team boundary is too coarse to be the last line. Design: `docs/design/oauth-chat-widget.md`, D4.

### Scope

- Applies to **client-credentials** applications only. API-key, Django-session and authorization-code callers keep team-membership semantics untouched.
- Gates `chatbots:interact` — the three call sites above, covering six views (v1/v2 trigger, versioned/unversioned OpenAI and message endpoints).
- Read surfaces (`chatbots:read`, `sessions:*`) stay on team scope for now. Worth a follow-up: an application can currently be barred from talking to a chatbot while still reading its transcripts.

### No backfill

Client credentials shipped 2026-07-24, so the live population is small enough to migrate by announcement. Existing applications must name their chatbots after this lands. Changelog + API docs, not a data migration.

### Plan

**1. Model + migration** — `apps/oauth/models.py`

```python
allowed_chatbots = models.ManyToManyField(
    "experiments.Experiment", blank=True, related_name="oauth_applications",
    help_text="Chatbots this application may start chat sessions with. Empty means none.",
)
```

No `limit_choices_to` — the valid set depends on the application's team, so the form and the check enforce it. Migration needs an explicit dependency on the latest `experiments` migration.

**2. Auditing** — `@audit_fields("allowed_chatbots")`, first audited model in `apps/oauth`. `field_audit` registers `m2m_changed` handlers, so M2M edits genuinely record. `audit_special_queryset_writes` not needed.

**3. The check** — `apps/oauth/permissions.py`, sibling to `is_client_credentials_request()`

```python
def application_allows_chatbot(request, experiment) -> bool:
    if not is_client_credentials_request(request):
        return True
    return request.auth.application.allowed_chatbots.filter(
        pk=experiment.get_working_version_id()
    ).exists()
```

Plus `enforce_application_chatbot_access` raising `PermissionDenied`. **403, not 401** — the caller authenticated fine, it just isn't authorised for this chatbot.

**4. Wire into the three call sites.** Check before anything mutates:

- `_chat_completions` — resolve the `Experiment` explicitly at the top. Today it is resolved inside `ExperimentSessionCreateSerializer`, so checking after `serializer.save()` would create a session and then reject it.
- `_new_api_message` — once after the if/else, where both branches have bound `experiment`.
- `handle_trigger_bot_message` — right after `get_object_or_404`, **before** `_get_or_create_participant_data`, which creates participant data as a side effect.

**5. Form + views** — `RegisterApplicationForm.__init__` takes `team`; `CreateApplication` and `EditApplication` pass `self.request.team` via `get_form_kwargs` (neither overrides it today — `CreateApplication` sets the team in `form_valid`, too late to filter a queryset). Queryset is the team's working versions. `clean()` clears the field for authorization-code grants. `RegisterGlobalApplicationForm` omits it — global apps are authorization-code only. Add to `render_form_fields` in `templates/oauth2_provider/application_form.html`.

**6. Docs + changelog.**

### Two traps

**Version normalisation.** `public_id` is `unique=True` per row and `create_new_version` assigns a fresh one (`apps/experiments/models.py:874`), so a version snapshot has its own id. `_new_api_message` and `handle_trigger_bot_message` both resolve without a working-version filter, so a caller can address a version directly. The allowlist holds family heads, so the check must normalise via `get_working_version_id()` — otherwise legitimate versioned calls are denied.

**Archived chatbots in the form.** A `ModelMultipleChoiceField` silently drops selected values outside its queryset. Filtering `is_archived=False` naively would quietly revoke an archived chatbot on the next unrelated save. Union in the instance's current selections.

### Tests

- Per endpoint (parametrised over all six views): application lists the chatbot → success; does not list it → 403; empty list → 403.
- **Version normalisation**: allowlist holds the family head, caller addresses a version's own `public_id` → allowed.
- Non-machine callers unaffected: API key, authorization-code token and Django session all still reach a chatbot the application does not list.
- No side effects on denial: no `ParticipantData` from a denied trigger, no session from a denied completion.
- Cross-team chatbot neither offerable by the form nor reachable.
- Audit records on `.add()` / `.remove()` / `.set()`.
- Form: an archived-but-selected chatbot survives a save.

### Suggested sequencing

Steps 1–3 are inert and can land as one commit, leaving step 4 as a pure wiring diff. Step 4 is the behaviour change and where the changelog attaches.

snopoke:
Closes #4197

### Product Description
A client-credentials OAuth application could converse with every chatbot in its team. It now reaches only the chatbots it was authorised for, chosen when the application is registered or edited. **Empty means none** — existing applications must name their chatbots after this lands.

### Technical Description
`OAuth2Application.allowed_chatbots` (M2M to `Experiment`, audited) gates `chatbots:interact` at its three call sites — `_chat_completions`, `_new_api_message`, `handle_trigger_bot_message` — covering six views. Denial is a 403 and happens before anything with a side effect (no session from a denied completion, no `ParticipantData` from a denied trigger).

Only client-credentials callers are gated; API-key, Django-session and authorization-code callers are untouched.

Two things worth a look:

- The allowlist holds working versions, but a caller can address a version by its own `public_id`, so the check normalises via `get_working_version_id()`.
- `RegisterApplicationForm.save()` never called `save_m2m()`, so the field would not have persisted at all. Fixed.

Also noted while testing, not fixed here: a machine token reaching `new_api_message` still crashes on `request.user.email` (`AnonymousUser` has none). The endpoint has no way to supply a participant identifier, so it has no machine success path — the allowlist denial now fires before that point.

### Migrations

`oauth.0004` adds the M2M table only. No backfill: client credentials shipped 2026-07-24, so the live population migrates by announcement.

### Demo
Verified against a running app with a real client-credentials token:

```
POST /api/openai/<unlisted-chatbot>/chat/completions  -> 403
    {"detail":"This application is not authorized to interact with this chatbot."}
POST /api/openai/<listed-chatbot>/chat/completions    -> 200
```

Registration and edit round trips driven in the browser: the field renders as a checkbox list of the team's chatbots, pre-checks the current selection, and persists.

### Docs and Changelog

The new 403 is declared on all six endpoints; `api-schemas/v1.yml` and `v2.yml` regenerated.


ocs-percy (approved):
This is a solid, well-scoped security gate. Approving.

**What I looked at:**

- `permissions.py`: Clean two-function split (`application_allows_chatbot` predicate + `enforce_application_chatbot_access` enforcer). The version normalisation via `get_working_version_id()` is exactly right — allowlist holds working versions, versioned callers still get through.
- **All 6 enforcement sites** are covered: `_chat_completions`, `_new_api_message`, `handle_trigger_bot_message` (v1 + v2), versioned chat completions. Checks land before any side-effectful call (session creation, participant data).
- `forms.py`: `_init_allowed_chatbots` handles the archived-chatbot edge case properly — the union-back-in trick prevents silent revocation on an unrelated save. The `save_m2m()` bugfix is important; without it the field would never have persisted at all.
- `RegisterGlobalApplicationForm.Meta` correctly drops the field; global apps have no team to offer chatbots from.
- `clean()` zeroing `allowed_chatbots` for non-client-credentials grants is a good safeguard.
- Migration is additive only — no backfill, backwards compatible. ✅
- Test suite is thorough: denial cases (not listed, empty list), allowed case, version normalisation, cross-team isolation, side-effect safety (no session/participant-data on denial), non-machine callers unaffected. The `COMPLETABLE_ENDPOINTS` / `VERSION_ADDRESSABLE_ENDPOINTS` split is a clean way to handle the `new_api_message` limitation.

**One thing to communicate pre-deploy:** empty `allowed_chatbots` means *no access* for all existing client-credentials applications. The PR notes this, but worth a team heads-up / runbook step so existing integrations aren't silently broken after the migration lands.

PR #4202, opened 2026-08-14
Allow admins to disable channels to temporarily block access

### LINKED ISSUE #4200: Allow admins to disable channels to temporarily block access
## Summary
Allow chatbot admins to disable individual channels as a way to temporarily block user access, with an optional static response message returned to users while the channel is disabled.

## Motivation
Admins sometimes need to temporarily cut off access to a bot on a specific channel — e.g. during maintenance, an incident, a content freeze, or while investigating misuse — without deleting the channel or its configuration. Today there's no clean way to do this short of tearing down the channel.

## Requirements
- Admins can **disable/enable** a channel via a toggle in the channel settings UI.
- When a channel is disabled:
  - Incoming user messages are **not processed** by the bot (no LLM/pipeline invocation).
  - Admins can configure an optional **static response message**:
    - If set, the message is sent back to any user who messages the disabled channel.
    - If left blank, the bot does **not respond** at all (silent).
- The channels UI **visually highlights** disabled channels so their state is obvious at a glance (e.g. badge/label, muted styling).

## UX notes
- Disabled state and the static message should be editable together in the channel config form.
- Consider surfacing the disabled indicator both in the channel list and on the channel detail view.

## Out of scope
- Scheduled/automatic enable-disable windows.
- Per-user or per-group access blocking (this is channel-wide).

snopoke:
### Product Description
Chatbot admins can now temporarily disable an individual channel instead of deleting it — useful during maintenance, an incident, a content freeze, or while investigating misuse. While a channel is disabled, incoming messages are not processed by the bot. Admins can optionally configure a static response that is sent back to anyone who messages the channel; leaving it blank makes the bot stay silent. Disabled channels are visually highlighted in the channel list and flagged in the channel config dialog.

Resolves #4200

### Technical Description
- `ExperimentChannel` gains `enabled` (BooleanField, default True) and `disabled_message` (TextField, blank). Both are added to `EXPERIMENT_CHANNEL_FIELDS` so changes are audited.
- New `ChannelDisabledStage` runs first in the message pipeline. A disabled channel therefore does no work: no participant record, no session, no LLM/pipeline invocation, nothing persisted to chat history. With a static message configured it raises `EarlyExitResponse` so the terminal stages deliver it via the channel's normal sender; without one it raises `EarlyAbort` for a silent halt.
- Wired into `ChannelBase._build_pipeline()` and `ApiChannel._build_pipeline()` (the latter also backs the embedded widget). `WebChannel`/`EvaluationChannel` are deliberately untouched — they serve team-global channels that have no admin toggle.
- `ChannelForm` exposes both fields; the message textarea is revealed by Alpine only when the channel is off, mirroring the existing `form_attrs` pattern used by the extra forms.
- `api-schemas/export.yml` regenerated: the export surface derives its serializer from the model.

Out of scope (per the issue): scheduled enable/disable windows and per-user blocking. Note that bot-initiated outbound messages (scheduled messages, reminders, event triggers) use a separate mini-pipeline and are not affected by the toggle.

### Migrations

`0032_experimentchannel_enabled_disabled_message` adds two nullable-safe columns with defaults; existing channels stay enabled.

### Demo
N/A — covered by tests that render the dialog and channel list and drive a disabled channel end-to-end through `new_user_message`.

### Docs and Changelog

### DIFF
--- a/api-schemas/export.yml
+++ b/api-schemas/export.yml
@@ -3832,7 +3832,8 @@ components:
           description: Uncheck to temporarily block access to this bot on this channel.
         disabled_message:
           type: string
-          description: Optional message sent to anyone who messages this channel while it is disabled. Leave blank for the bot to stay silent.
+          description: Optional message sent to anyone who messages this channel while
+            it is disabled. Leave blank for the bot to stay silent.
         external_id:
           type: string
           format: uuid
--- a/api-schemas/export.yml
+++ b/api-schemas/export.yml
@@ -3827,6 +3827,12 @@ components:
           maxLength: 255
         deleted:
           type: boolean
+        enabled:
+          type: boolean
+          description: Uncheck to temporarily block access to this bot on this channel.
+        disabled_message:
+          type: string
+          description: Optional message sent to anyone who messages this channel while it is disabled. Leave blank for the bot to stay silent.
         external_id:
           type: string
           format: uuid
--- a/apps/channels/api_channel.py
+++ b/apps/channels/api_channel.py
@@ -10,6 +10,7 @@ from apps.channels.pipeline import MessageProcessingContext, MessageProcessingPi
 from apps.channels.sender import ChannelSender
 from apps.channels.stages.core import (
     BotInteractionStage,
+    ChannelDisabledStage,
     ChatMessageCreationStage,
     ConsentCheckStage,
     ConsentFlowStage,
@@ -90,6 +91,8 @@ class ApiChannel(ChannelBase):
     def _build_pipeline(self) -> MessageProcessingPipeline:
         return MessageProcessingPipeline(
             core_stages=[
+                # ApiChannel also backs the embedded widget, which admins can disable
+                ChannelDisabledStage(),
                 ParticipantValidationStage(),
                 ParticipantResolverStage(),
                 ConsentCheckStage(),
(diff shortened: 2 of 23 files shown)


ocs-percy (approved):
Solid implementation. Clean feature with well-reasoned stage placement.

**What I looked at:**
- Stage design: `ChannelDisabledStage` runs after participant/consent stages (so the static reply is addressable and consent-gated) but before session/bot work — the placement is correct and the comments in `channel_base.py` explain the tradeoffs clearly.
- The `last_activity_at` fix in `pipeline.py` is a nice catch — prevents a disabled-channel reply from going down the paid WhatsApp template path when there's no session yet.
- `EvaluationChannel` correctly omits the stage with a clear comment explaining why.
- `notifications.py` gracefully handles the sessionless case for delivery failure notifications.
- Test coverage is thorough: unit tests for `ChannelDisabledStage`, end-to-end tests per channel type, WhatsApp template billing test, consent-gate test.

**Outstanding CodeRabbit comment:** The migration defaults flag (`enabled=True`, `disabled_message=""`) is technically correct per `.coderabbit.yaml` policy, but for Postgres 11+ adding a column with a constant default is safe without a table rewrite. Given `ExperimentChannel` is not a firehose table this is low practical risk — worth being aware of but not a blocker.

LGTM 🚀

Issue #4199, opened 2026-08-14
Replace chat session inactivity window with an absolute session lifetime


snopoke:
## Problem

`session_token_expired()` (`apps/api/session_tokens.py`) is a **sliding** window: `CHAT_SESSION_TOKEN_INACTIVITY_WINDOW` (7 days) measured from `last_activity_at`, which advances on every user message. ADR-0040 deliberately excluded polling from that so a leaked token can't keep a session alive *without talking* — but a caller who keeps chatting slides the window forever.

Since admission is decided once, at session creation (ADR-0039, ADR-0053), the abuse budget of a single admitted caller is:

```
rate limit (ADR-0052, per-session bucket)  ×  session lifetime
```

The second factor is unbounded, which makes the first one moot. One session start buys unlimited chatbot usage at the throttle's rate, indefinitely.

This matters independently of any particular admission credential, but it is sharpest for hosts hardening a widget against abuse: tightening the door (short-lived credentials, stricter checks) does nothing about the room behind it, because the door is only crossed once. Context in #3893, though this is separable and needs nothing from it.

## Proposal

Expire on **age**, not dormancy:

```python
# apps/api/session_tokens.py
def session_token_expired(session: ExperimentSession) -> bool:
    """A session's token stops working a fixed time after the session was created."""
    channel = session.experiment_channel
    lifetime = (channel and channel.session_token_lifetime) or settings.CHAT_SESSION_TOKEN_LIFETIME
    return timezone.now() - session.created_at > lifetime
```

### Replace, don't add

Adding an absolute cap *beside* the inactivity window does not work. `last_activity_at >= created_at`, so `now - last_activity <= now - created_at`, and therefore any lifetime `L <= W` makes the inactivity branch **unreachable** — whenever it would fire, the age check has already fired. At the proposed `L = 7 days` against today's `W = 7 days`, the window is provably dead code. Dead code that reads like live policy is worse than no policy.

The two only compose when `L > W` — "a dormant session dies in a week, an active one dies at 90 days" — which is a coherent policy that nothing currently wants.

### Which makes the lifetime mandatory

With the window gone, a session with no lifetime has *no expiry at all* — the "no expiry" alternative ADR-0040 explicitly rejected, and the indefinite access ADR-0039 exists to prevent. So there is no "off": the global setting always has a value and the per-channel field only ever overrides it.

[...]

PR #4204, opened 2026-08-14
Expire chat session tokens on absolute age

snopoke:
Part of #4199 — the global setting only; the per-channel `session_token_lifetime` override, its migration and its form field are left for a follow-up (ADR-0054 records it as deferred, not rejected).

A session's token expired on dormancy, measured from `last_activity_at`, which every user message advanced. Admission is decided once at session creation (ADR-0039), so a caller who kept chatting held an admitted session forever — rate limit × unbounded lifetime. It now expires 7 days after `created_at`, and activity doesn't extend it.

Replacing rather than adding: since `last_activity_at >= created_at`, any lifetime at or below the old window makes the dormancy branch unreachable, so keeping both would have left dead code reading as live policy.

7 days is deliberately today's number — the new rule fires no later than the old one for any session, so nothing's access is extended.

### Worth review attention

**Bound widgets dead-end.** The `session_expired` recovery path starts a new conversation only for *unbound* widgets. A widget handed a `session-id` + `session-token` by its host page — how `web_chat.html` renders the full-page and kiosk chat — refuses to restart a session it doesn't own and shows "This chat session is no longer available". So an anonymous participant on a public chatbot page whose conversation is still in daily use hits a wall on day 7 with no restart affordance on that page. Previously that only befell already-abandoned sessions. Adding the affordance is the follow-up ADR-0054 names; shipping without it is the call to make here.

"Continue Chat" and the logged-in chat page are unaffected — both go through the participant-user bypass, which runs before the expiry check. `test_participant_user_bypass_outlives_the_lifetime` pins that, since this diff makes the bypass load-bearing.

**Announcement.** Sessions currently riding the sliding window past a week end at their next request. Small population, not empty.

### Migrations

None.

### Docs and Changelog

The user-visible change is that a conversation now ends a week after it started rather than a week after it went quiet.

### DIFF
--- a/apps/api/session_tokens.py
+++ b/apps/api/session_tokens.py
@@ -30,11 +30,10 @@ def validate_session_token(token: str, session_external_id: str) -> bool:
 
 
 def session_token_expired(session: ExperimentSession) -> bool:
-    """Sliding inactivity backstop: reject token access to long-inactive sessions.
+    """A session's token stops working a fixed time after the session was created.
 
-    Activity is the session's `last_activity_at` (updated on each user message;
-    polling does not count, so a leaked token cannot keep a session alive),
-    falling back to session creation when there has been no activity yet.
+    The lifetime is absolute: activity does not extend it, so an admitted caller's
+    access is bounded no matter how much they talk. Once it fires the caller must
+    start a new session and be re-admitted under whatever rules apply then.
     """
-    last_activity = session.last_activity_at or session.created_at
-    return timezone.now() - last_activity > settings.CHAT_SESSION_TOKEN_INACTIVITY_WINDOW
+    return timezone.now() - session.created_at > settings.CHAT_SESSION_TOKEN_LIFETIME
--- a/apps/api/tests/test_chat_session_token.py
+++ b/apps/api/tests/test_chat_session_token.py
@@ -9,6 +9,7 @@ from rest_framework.test import APIClient
 
 from apps.api.session_tokens import issue_session_token
 from apps.channels.models import ChannelPlatform
+from apps.chat.models import ChatMessage, ChatMessageType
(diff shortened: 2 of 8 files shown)


ocs-percy (approved):
Solid, focused change — approving.

**What I looked at:**

- `session_tokens.py` — the rewrite is as minimal as it could be. `timezone.now() - session.created_at > CHAT_SESSION_TOKEN_LIFETIME` is the whole policy; `last_activity_at` is gone from the expiry path entirely. The rationale in the PR (rate limit × unbounded lifetime = unlimited abuse budget) is correct and the fix is the right one.
- Setting rename (`INACTIVITY_WINDOW` → `LIFETIME`) is clean and accurately describes what the value now means.
- **Tests** cover the key cases: dormancy alone no longer expires, activity does not slide the window (`test_session_expired_after_lifetime_despite_recent_activity` with the `traveller.shift()` is the important regression catch), the lifetime setting is exercised, and the participant-user bypass is explicitly pinned (`test_participant_user_bypass_outlives_the_lifetime` — good that this exists given the bypass is now load-bearing).
- The decision to replace rather than add is right — with a lifetime ≤ old window, the dormancy branch is unreachable dead code that reads as live policy. ADR-0054 explains this clearly.
- **ADR-0054** is well-written, including the honest accounting of consequences: bound widgets hitting a hard wall at day 7 for active conversations is a real regression, and it's good that it's named explicitly rather than buried. Shipping without the restart affordance is acceptable given the security gain — just needs to be on the follow-up list alongside the per-channel `session_token_lifetime` override.
- One operational note: sessions currently living past 7 days by activity will end at their next request post-deploy. Worth flagging in release notes / internal announcement if any hosted instances have long-running anonymous kiosk sessions.

PR #4206, opened 2026-08-17
Feat: Adds caching for chatbot usage

brian-mecha:
### Product Description

Adds short-lived caching (5 min TTL) to the chatbot home page's usage widget, reducing repeated database load when a chatbot's page is viewed multiple times in quick succession. No visible behavior change.
Spend/session/message counts may lag actual usage by up to 5 minutes. Session and trace-level usage views are unaffected (left uncached, as their queries are already cheap single-row lookups).

Follow up PR for [4179](https://github.com/dimagi/open-chat-studio/pull/4179)

### Docs and Changelog

### DIFF
--- a/apps/cost_tracking/services/reporting.py
+++ b/apps/cost_tracking/services/reporting.py
@@ -15,7 +15,7 @@ from django.db.models.functions import Coalesce, TruncDate, TruncMonth, TruncWee
 from django.utils import timezone
 
 from apps.cost_tracking.models import Confidence, ServiceKind, UsageRecord, UsageSource
-from apps.experiments.models import Experiment, ExperimentSession
+from apps.experiments.models import ExperimentSession
 from apps.teams.models import Team
 from apps.trace.models import Trace
 from apps.usage_metrics.dashboard_querysets import filtered_querysets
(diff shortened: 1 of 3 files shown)


snopoke (commented):
I think the caching should probably happen at the view level or else in a new function that also applies the date filtering:
```python
def get_latest_chatbot_usage_summary(team, experiment_id):
    ...
```

apps/cost_tracking/services/reporting.py
```diff
@@ -396,6 +403,13 @@ def chatbot_usage_summary(experiment: Experiment, *, start: datetime, end: datet
     page's usage widget. Session/message counts come from `filtered_querysets` - the same canonical,
     ADR-0051 activity definitions the dashboard's Bot Performance table uses - narrowed to this
     experiment, rather than re-deriving the session base here.
+
+    Uncached and takes an arbitrary window - for a caller that wants a fixed date range. The chatbot
+    home page instead wants "the last 30 days as of now" specifically, cached for a few minutes to
+    cut its query load: see `get_latest_chatbot_usage_summary`, which owns both of those and calls
+    this function on a cache miss. Caching *this* function directly would be wrong - a cache key
+    that doesn't include `start`/`end` would silently serve one caller's window to another's, and a
+    key that does would never hit here since "now"-relative bounds differ on every call.
```
snopoke: unnecessary comment
```suggestion
```

apps/cost_tracking/services/reporting.py
```diff
@@ -404,6 +418,37 @@ def chatbot_usage_summary(experiment: Experiment, *, start: datetime, end: datet
...
+    write would defeat the cache (the same tradeoff `PricingResolver`'s cache makes).
+    """
+    cache_key = _chatbot_usage_summary_cache_key(team.id, experiment_id)
+    cached = cache.get(cache_key)
+    if cached is not None:
+        return cached
+
+    end = timezone.now()
+    start = end - timedelta(days=_CHATBOT_USAGE_SUMMARY_WINDOW_DAYS)
+    experiment = Experiment.objects.get(id=experiment_id, team=team)
```
snopoke: Rather than re-fetching the model here you could change `chatbot_usage_summary` to accept the experiment ID and the team separately.

snopoke (approved):

PR #4211, opened 2026-08-17
Resolve pipeline node inputs by arrival order

snopoke:
### Product Description
When several branches merge into one node, that node now takes its input from the branch that arrived most recently, and `node_inputs` holds every input that has arrived so far.

Previously the input was chosen by the order the connections happened to be stored, so the same graph could feed a merge node a different branch depending on how it was drawn — sometimes never surfacing a branch's output at all. `node_inputs` only ever held one branch's output, so the documented "wait until I have at least N inputs" pattern never completed and the pipeline silently returned nothing.

### Technical Description
`PipelineState.get_node_inputs` now walks `state["path"]` — one entry per node run, appended in execution order — and returns every input delivered to the node as `(source_node_id, output)` pairs, oldest first. A node that ran more than once contributes one input per run that targeted this node; a run that produced no output (a Python node aborting via `require_node_outputs`) contributes none, since it appends neither an output nor a path entry.

`_prepare_state` sets `last_node_input`/`node_source` from the most recent pair and `node_inputs` from all of them.

This restores the path-order selection that `8edc8de87` replaced with `reversed(incoming_nodes)`, and gives `node_inputs` the semantics `c8c81d263` named it for.

Existing test `test_code_node_wait_for_inputs` changed expectation: the merge node's source is now `B`, the branch that delivered last and unblocked the merge, rather than an edge-order artifact.

### Migrations
None.

### Demo
On the graph from the [uneven branches docs](https://docs.openchatstudio.com/concepts/pipelines/parallel/#uneven-branches) (`start → A → C → D`, `start → B → D`):

| | before | after |
|---|---|---|
| `D` run 1 | `input=B` | `input=B`, `node_inputs=[B]` |
| `D` run 2 | `input=B` if the C→D edge was drawn first, else `input=C` | `input=C`, `node_inputs=[B, C]` — either way |
| optional-branch merge counting `node_inputs` | pipeline returns nothing | returns the merged value |

Four regression tests added to `test_parallel_nodes.py`, including one where a merge node and a branch reach the same node in the same step — that case previously discarded the merged value outright.

### Docs and Changelog

Docs updated in a companion PR: https://github.com/dimagi/open-chat-studio-docs/pull/688 — the pages describing both merge patterns documented behaviour this code did not have.

:warning: Behaviour change for live pipelines: any node with multiple incoming connections may receive a different input than before, and `node_inputs` can now hold more entries.

### DIFF
--- a/apps/pipelines/nodes/base.py
+++ b/apps/pipelines/nodes/base.py
@@ -107,13 +107,14 @@ class PipelineState(dict):
     # List of (previous, current, next) tuples used for aiding in routing decisions.
     path: Annotated[Sequence[tuple[str | None, str, list[str]]], operator.add]
 
-    # inputs to the current node
+    # Every input delivered to the current node so far, oldest first. One entry per incoming node
+    # run that routed here, so a node on a merge accumulates an entry per branch that reaches it.
     node_inputs: list[str]
 
-    # input from the last executed node prior to this one
+    # The most recently delivered of those inputs: the current node's primary input.
     last_node_input: str
 
-    # source node for the current node
+    # The node that produced `last_node_input`.
     node_source: str
 
     intents: Annotated[list[Intents], _merge_intents]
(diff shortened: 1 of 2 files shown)


ocs-percy (approved):
Good fix. The root cause is clear (edge-draw order leaking into runtime semantics) and the solution is the right one.

**What I looked at:**
- `get_node_inputs` return-type change from `dict` to `list[tuple]` is the right shape now that order is the thing that matters. Walking `state["path"]` in execution order gives stable, deterministic inputs regardless of how edges were stored.
- `runs_seen` correctly handles the case where one upstream node runs multiple times — nth path entry maps to nth output at that index.
- `_prepare_state` is much cleaner: last pair → `last_node_input`/`node_source`, all values → `node_inputs`. The old `reversed()` loop + edge-case comment is gone because the new approach makes it unnecessary.
- `node_inputs` now actually has the semantics the docs described and the `wait_for_next_input()` pattern needs — accumulating across branches in arrival order.
- Test coverage: edge-order independence test (parametrized long-edge-first/short-edge-first), optional branch wait, downstream-of-merge, and the same-step arrival case. Solid.

**CodeRabbit nitpick:** Suggests a test for a single source node running twice and routing to the same target. Fair gap, but `runs_seen` is exercised implicitly and this is clearly a nitpick, not a blocker.

**Behaviour change** is correctly called out in the PR description — worth a prod comms note if any live pipelines use multi-branch merges.

LGTM 🚀

PR #4213, opened 2026-08-17
fix: make help tooltips keyboard accessible

### LINKED ISSUE #4171: Help tooltips are unreachable by keyboard
## Problem

`templates/generic/help.html` renders the help icon used across the app. Its trigger is a `<div role="button">` with no `tabindex`, so it cannot receive focus. The panel is revealed by `:hover` or `:focus-within` on the container, and the only elements carrying `tabindex="0"` sit inside `.dropdown-content`, which is `display: none` until the panel opens. Nothing in the component is focusable, so keyboard users cannot open any help tooltip in the application.

The trigger also has no accessible name, so assistive technology announces an unlabelled button.

WCAG 2.2: 2.1.1 Keyboard (Level A), 4.1.2 Name, Role, Value (Level A).

## Where

`templates/generic/help.html` lines 11-22:

```html
<div class="dropdown dropdown-right dropdown-hover">
  <div role="button" class="btn btn-circle btn-ghost btn-xs text-info">
    <i class="text-xs fa-regular fa-circle-question"></i>
  </div>
  <div tabindex="0" class="card card-sm dropdown-content ...">
```

The partial is included from 14 templates. Three of them are `templates/web/form/text.html`, `select.html` and `checkbox.html`, which are the generic Django form-field renderers, so this partial backs effectively every form help icon in the app.

`templates/dashboard/index.html` lines 132-135 hold an inline copy of the same markup. The dashboard stat cards are rendered by Alpine rather than Django, so the partial cannot be included there. The copy needs the same change to stay in step.

## Fix

Replace the div trigger with a real button carrying an accessible name:

```html
<button type="button"
        class="btn btn-circle btn-ghost btn-xs text-info"
        aria-label="{% translate 'More information' %}">
  <i class="text-xs fa-regular fa-circle-question" aria-hidden="true"></i>
</button>
```

A native `<button>` is focusable, so Tab reaches it and daisyUI's existing `:focus-within` rule opens the panel. No JavaScript is needed.

The governing reveal rule in daisyUI 5.7.14:

```css
.dropdown:not(details, .dropdown-open, .dropdown-hover:hover, :focus-within) .dropdown-content {
  display: none;
}
```

## The trap

Do not add `tabindex="0"` to the button, even though daisyUI's own documentation uses `<div tabindex="0" role="button">` for dropdown triggers. daisyUI 5.7.14 also ships:

```css
.dropdown.dropdown-hover:not(:hover) [tabindex]:first-child:focus:not(:focus-visible) ~ .dropdown-content {
  display: none;
}
```

This suppresses the panel when the first-child `[tabindex]` element has focus but not `:focus-visible`, which is exactly what pointer and touch focus produce. A `<button>` with no `tabindex` attribute does not match the `[tabindex]` selector, so tap-to-focus keeps the panel open on touch devices. Adding `tabindex="0"` reintroduces the touch problem this issue is meant to fix.

`templates/service_providers/components/llm_model_row.html` uses a `tooltip` class with `data-tip` on a `<span class="badge">`. That is not focusable either, so it is not a pattern to copy here.

## Consequence to expect

Making the trigger focusable inserts a tab stop next to every form field that carries help text. That is the correct behaviour, since the content is currently unreachable by keyboard at all, but it is a perceptible change to form navigation across the app and is worth calling out in the PR description.

## Verification

- Tab to a help icon on any form, confirm the panel opens and its content is announced.
- Tap the icon on a touch device, confirm the panel opens and stays open.
- Confirm hover still works with a mouse.
- Check both light and dark themes.

## Context

Raised by CodeRabbit on #4156 (metric definition convergence). Out of scope there: that PR's subject is metric definitions, and this touches an app-wide partial.

nanhe17:
## Summary
- Replace non-focusable help tooltip triggers with native buttons.
- Add accessible labels and hide decorative icons from assistive technology.
- Apply the same markup to the dashboard copy.

## Testing
- git diff --check

Fixes #4171

### DIFF
--- a/templates/dashboard/index.html
+++ b/templates/dashboard/index.html
@@ -130,9 +130,12 @@
                 {# Mirrors generic/help.html; inlined because the card list is rendered by Alpine, not Django. #}
                 <template x-if="stat.tooltip">
                   <div class="dropdown dropdown-top dropdown-hover">
-                    <div role="button" class="btn btn-circle btn-ghost btn-xs text-info">
-                      <i class="text-xs fa-regular fa-circle-question"></i>
-                    </div>
+                    <button
+                      type="button"
+                      class="btn btn-circle btn-ghost btn-xs text-info"
+                      aria-label="{% translate 'More information' %}">
+                      <i class="text-xs fa-regular fa-circle-question" aria-hidden="true"></i>
+                    </button>
                     <div
                       tabindex="0"
                       class="card card-sm dropdown-content bg-slate-300 dark:bg-slate-700 rounded-box z-1 w-64 shadow-sm">
--- a/templates/generic/help.html
+++ b/templates/generic/help.html
@@ -8,10 +8,14 @@ Using directly:
   {% include "generic/help.html" with help_content=help_text %}
 
 {% endcomment %}
+{% load i18n %}
 <div class="dropdown dropdown-right dropdown-hover">
-  <div role="button" class="btn btn-circle btn-ghost btn-xs text-info">
-    <i class="text-xs fa-regular fa-circle-question"></i>
-  </div>
+  <button
+    type="button"
+    class="btn btn-circle btn-ghost btn-xs text-info"
+    aria-label="{% translate 'More information' %}">
+    <i class="text-xs fa-regular fa-circle-question" aria-hidden="true"></i>
+  </button>
   <div
     tabindex="0"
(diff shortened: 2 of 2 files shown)


snopoke:
@nanhe17 thanks for the contribution. Please could you include a screen shot or short video / GIF of the working help button for visual validation.

SmittieC (approved):

PR #4214, opened 2026-08-17
Bound memory use in chat session exports

snopoke:
### Product Description
Large chat exports no longer risk OOM-killing the background worker.

### Technical Description
Investigating an overnight outage traced to a memory spike on a large session export. Profiling with tracemalloc found three separate terms, only one of which was the obvious suspect:

1. **`ContentFile(tmp.read())` in `async_export_chat`** — read the whole compressed export back into one contiguous `bytes`, directly negating the temp-file spooling on the line above. This is the only term that grew without bound with output size: measured 32.0 MB peak at a 31.8 MB gzip, 8.2 MB at 7.9 MB — tracking output exactly. Now flat at 0.3 MB.
2. **`session_cache`** — one entry per session, held for the whole run. It can't evict cleanly by construction: messages are ordered by global message pk, so sessions stay interleaved and no entry is ever safely "done". Now a bounded LRU; a miss just recomputes, with no extra query since the session and its chat tags/comments are already loaded on the message.
3. **Spent chunks lingering behind ORM reference cycles** (`_state`, related-object and prefetch caches), so 2–3 chunks were live at once instead of one.

Worth knowing: the message axis *plateaus* at ~95 MB (2k/8k/16k messages), so the keyset chunking from #3232 was working. Only the session axis kept climbing — 29.3 MB at 500 sessions to 97.0 MB at 4,000, to produce a 250 KB file. A wide export is a better fit for the incident than a deep one. Anyone re-measuring this should use the session axis; the message axis looks fine even when something is wrong.

**Worth a closer look:** `del chunk` before the collect is load-bearing, not tidying — the two do different work and both earn their place. Measured at 16k messages: 94.7 MB with neither, 46.0 MB with `del` alone, 28.8 MB with both.

The per-chunk `gc.collect()` has no measurable wall-time cost — 5 interleaved A/B reps give a median of 15.60s with vs 15.94s without (−2.1%), well inside a noise floor that threw 19–20s outliers. Holding less live memory seems to pay for the collect itself.

Also removes `filtered_export_to_csv` (buffered whole exports, no callers). Six of its tests weren't testing the buffering — they were the only coverage of participant-data snapshots via trace diffs, trace-id resolution, and the session-state column — so they're ported to `export_rows_to_csv_stream` rather than deleted with it.

Net: 94.7 → 28.7 MB on the message axis, 97.0 → 33.2 MB on the session axis.

### Migrations
N/A — no schema changes.

### Demo
N/A — no user-visible change beyond exports not falling over.

### Docs and Changelog

Internal reliability fix; no user-facing behaviour or API change to document.

### DIFF
--- a/apps/experiments/export.py
+++ b/apps/experiments/export.py
@@ -292,20 +292,6 @@ def export_to_tempfile(
     return tmp
 
 
-def filtered_export_to_csv(experiment, sessions_queryset, translation_language=None):
-    """Build a complete CSV in a StringIO buffer and return it.
-
-    For large datasets prefer generate_export_rows() + export_rows_to_csv_stream()
-    to avoid buffering the entire export in memory.  This function is retained for
-    backward compatibility with callers that expect a StringIO return value.
-    """
-    csv_in_memory = io.StringIO()
-    writer = csv.writer(csv_in_memory, delimiter=",", quotechar='"', quoting=csv.QUOTE_MINIMAL)
-    for row in generate_export_rows(experiment, sessions_queryset, translation_language):
-        writer.writerow(row)
-    return csv_in_memory
-
-
 def _get_trace_id_for_export(message):
     """Returns the trace info from the message.
     This will return the first non-OCS trace info if it exists.
--- a/apps/experiments/export.py
+++ b/apps/experiments/export.py
@@ -3,6 +3,7 @@ import gzip
 import io
 import json
 import tempfile
+from collections import OrderedDict
 from collections.abc import Generator, Iterator
 
 import dictdiffer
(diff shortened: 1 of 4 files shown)


ocs-percy (approved):
Thorough root-cause analysis and targeted fixes for all three memory-leak terms.

- `DjangoFile(tmp)` fix is minimal and correct — straightforward streaming win.
- `_SessionCache` LRU is well-implemented; eviction logic is right and the `move_to_end` on hit is clean.
- `del chunk` + `gc.collect()` is load-bearing as described; the benchmarks make a convincing case.

Test coverage is excellent — the weakref test actually verifies GC behaviour rather than just exercising the path, and `_ReadTracker` gives a good regression guard against unbounded reads sneaking back in.

Happy with this. 🚢

PR #4218, opened 2026-08-18
Provider spend reporting: tracing volume and project-based team joins

snopoke:
### Product Description

No change. Staff-only reporting APIs, consumed by the provider spend reconciliation script.

### Technical Description

Monthly provider spend is attributed back to OCS teams by joining each provider's cost report against OCS's key registry. Two providers can't be joined that way, and this adds what they need:

- **`provider-keys` gains `trace_providers` and `cloud_project`.** Tracing bills per project inside an organization, and Vertex authenticates with a service-account document rather than an API key, so neither has a key fingerprint to match a team on — the project id is the only join available. `TraceProvider.metadata` already recorded the Langfuse project/org, so this mostly just exposes it. Vertex's project id comes out of `credentials_json`; it's an identifier, not a credential, and no secret is returned by either block.

- **New `/admin/api/tracing-usage/`** returning per-team `traces`, `turns` and `toolcalls`. It combines none of them: a tracing invoice needs a single weight, but what counts as a billable unit is a fact about the vendor's pricing rather than about OCS, so the consumer decides. Worth knowing before weighting them — a trace records its LLM turns and tool calls but not its pipeline node steps, so anything billing per observation is billing for spans no column here counts.

Both endpoints reuse the existing `superuser_or_reporting_token` + rate-limit decorators and the dashboard date-range form, so all three of `range_type`, `start` and `end` are required — `_get_form` leaves the form unbound otherwise. Tests pin that down since it's easy to trip over.

### Demo

```
$ curl -H "Authorization: Bearer $TOKEN" \
    "$OCS/admin/api/tracing-usage/?range_type=custom&start=2026-07-01&end=2026-07-31"

{"start": "...", "end": "...",
 "teams": [{"team_id": 1, "team_name": "...", "team_slug": "...",
            "traces": 412, "turns": 1103, "toolcalls": 87}]}
```

### Docs and Changelog

### DIFF
--- a/apps/admin/provider_keys.py
+++ b/apps/admin/provider_keys.py
@@ -22,7 +22,7 @@ import logging
 from collections.abc import Iterator
 from dataclasses import dataclass
 
-from apps.service_providers.models import LlmProvider, TraceProvider
+from apps.service_providers.models import LlmProvider, LlmProviderTypes, TraceProvider
 from apps.teams.metadata import get_team_metadata_fields
 
 logger = logging.getLogger("ocs.admin")
@@ -144,17 +144,32 @@ def _cloud_project(provider) -> str | None:
     ``mask_secret`` has nothing to fingerprint and the key→team join can never fire for
     it. Its GCP project id is the join instead — that is what the Cloud Billing export
     keys cost by — and it is an identifier, not a credential.
+
+    Gated on the provider type rather than on the presence of ``credentials_json``:
+    ``config`` is a free-form JSON blob, so any provider could carry that key, and a
+    stray one would hand a spend report a project id to attribute cost by — quietly
+    billing the wrong team.
     """
-    credentials = provider.config.get("credentials_json")
-    if isinstance(credentials, str):
-        try:
-            credentials = json.loads(credentials)
-        except ValueError:
-            logger.warning("Provider id=%s has unparseable credentials_json", provider.pk)
-            return None
-    if not isinstance(credentials, dict):
+    if provider.type != str(LlmProviderTypes.google_vertex_ai):
         return None
-    return credentials.get("project_id") or None
+    credentials = provider.config.get("credentials_json")
+    project = _as_dict(credentials).get("project_id") or None
+    if credentials and not project:
+        # Configured but unusable -- malformed JSON, or a document with no project_id.
+        # Either way this team's Vertex spend cannot be joined, so say so rather than
+        # letting it silently fall through to the pro-rata bucket.
+        logger.warning("Vertex provider id=%s has no usable project_id in credentials_json", provider.pk)
+    return project
+
+
+def _as_dict(value) -> dict:
+    """``value`` as a dict, whether it is stored as one or as a JSON string."""
+    if isinstance(value, str):
+        try:
+            value = json.loads(value)
+        except ValueError:
+            return {}
+    return value if isinstance(value, dict) else {}
 
 
 def _secret_field_for(provider) -> str | None:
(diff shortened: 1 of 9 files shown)


apps/admin/provider_keys.py
```diff
@@ -61,32 +68,95 @@ def mask_secret(secret: str, provider_type: str) -> str:
...
+    credentials = provider.config.get("credentials_json")
+    if isinstance(credentials, str):
+        try:
+            credentials = json.loads(credentials)
+        except ValueError:
+            logger.warning("Provider id=%s has unparseable credentials_json", provider.pk)
+            return None
+    if not isinstance(credentials, dict):
+        return None
+    return credentials.get("project_id") or None
```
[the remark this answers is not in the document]
snopoke: Fixed in 0b40b4f78. Gated on `provider.type` — you're right that `config` is free-form, so nothing stopped a non-Vertex provider carrying a `credentials_json`, and the docstring already claimed otherwise. Added a test asserting `cloud_project is None` for an OpenAI provider holding a `project_id`.

The consumer already filtered on `provider_type` before joining, so no cost was actually misattributed, but the contract should hold on its own rather than depending on that.

While in there: a Vertex provider whose `credentials_json` is present but unusable (malformed, or no `project_id`) now logs a warning instead of silently falling through to the pro-rata bucket.

*— Claude Code*

apps/admin/provider_keys.py
```diff
@@ -6,16 +6,23 @@
 ``sk-...JrYA`` for OpenAI, ``sk-ant-api03-cLV...lAAA`` for Anthropic) so the
 masked key here can be matched against what the provider console/API shows.
 
+Not every provider bills against something a key can identify. Vertex authenticates
```
snopoke, line 9: Fixed in 49acb85f0 — `apps/admin/provider_keys.py` is now 10.0 with no findings.

Extracted the tracing-metadata projection into `_trace_metadata`, which replaced four repeated `or ""` fallbacks with one comprehension over a named tuple of keys, and pulled the JSON-or-dict coercion out of `_cloud_project` into `_as_dict`.

*— Claude Code*

apps/admin/views.py
```diff
@@ -28,8 +28,9 @@
 )
 from apps.admin.imports import import_team_metadata_from_csv
 from apps.admin.models import OcsConfiguration
-from apps.admin.provider_keys import get_provider_key_fingerprints
+from apps.admin.provider_keys import get_provider_key_fingerprints, get_trace_provider_records
```
snopoke, line 32: Fixed in 49acb85f0. Both views differed only in which report they build, so the range validation and 400 response moved into a shared `_range_report(request, build)` and each view is now a one-liner.

*— Claude Code*

ocs-percy (approved):
Clean, well-scoped additions to the staff reporting APIs.

- `_cloud_project` gated on provider type rather than key presence is the right call — any provider config blob could carry `credentials_json`, and silently attributing cost via the wrong join would be a quiet billing error. The warning log for a configured-but-unusable field is a good touch.
- `get_trace_provider_records` correctly exposes only host + unencrypted metadata; the test asserting the key pair is absent from the JSON is the right thing to pin.
- `build_tracing_volume_report` leaving the three counts separate is correct API design — the weighting is a consumer decision about how the tracing service prices ingestion, not an OCS concern.
- `_range_report` helper is a good DRY move.
- `TraceFactory.at` post-generation hook via `.update()` is the right way to set `auto_now_add` fields in tests.
- `conftest.py` extraction cleans up the duplicated `superuser_client` fixture nicely.

🚢

PR #4219, opened 2026-08-18
Deprecate retired Groq models

snopoke:
### Product Description

OCS's Groq models are being reconciled with Groq's official deprecation schedule. Three models Groq has retired (`gemma2-9b-it`, `llama-3.1-8b-instant`, `llama-3.3-70b-versatile`) are now flagged deprecated in OCS, so existing users see a deprecation warning and a recommended replacement, and the Groq provider default moves to a still-supported model (`openai/gpt-oss-120b`). No models are removed, so existing configurations keep working.

### Technical Description

Reconciles OCS's Groq model list with [Groq's official deprecations page](https://console.groq.com/docs/deprecations), following `docs/developer_guides/managing_models.md`.

- `apps/service_providers/llm_service/default_models.py`: marked `gemma2-9b-it` (replacement `openai/gpt-oss-20b`), `llama-3.1-8b-instant` (replacement `openai/gpt-oss-20b`), and `llama-3.3-70b-versatile` (replacement `openai/gpt-oss-120b`) as `deprecated=True` — all three are on Groq's deprecation schedule with shutdown dates now passed. Moved the Groq `is_default`/`is_translation_default` flags off the now-deprecated `llama-3.3-70b-versatile` onto the recommended survivor `openai/gpt-oss-120b`.
- Added migration `apps/service_providers/migrations/0073_deprecate_groq_models.py` running `llm_model_migration()` + `notify_deprecated_models` (notifies affected teams of the deprecations).
- Stripped the superseded one-shot `llm_model_migration()`/`notify_deprecated_models` calls from earlier migrations `0071`/`0072` (kept idempotent `load_pricing_data()` and dependency chains intact).

### Migrations

This PR adds a data migration (0073). It only sets deprecated flags / sends notifications and removes no models, so it is backwards compatible.

### Docs and Changelog

---
_Generated by [Claude Code](https://claude.ai/code/session_016yVrzTx9CroEHsAFLbagwJ)_

### DIFF
--- a/apps/service_providers/llm_service/default_models.py
+++ b/apps/service_providers/llm_service/default_models.py
@@ -107,11 +107,11 @@ DEFAULT_LLM_PROVIDER_MODELS = {
     ],
     "groq": [
         Model("whisper-large-v3-turbo", k(8)),
-        Model("gemma2-9b-it", k(8)),
+        Model("gemma2-9b-it", k(8), deprecated=True, replacement="openai/gpt-oss-20b"),
         Model("gemma-7b-it", k(8), deprecated=True),
-        Model("llama-3.3-70b-versatile", k(128), is_default=True, is_translation_default=True),
-        Model("llama-3.1-8b-instant", k(128)),
-        Model("openai/gpt-oss-120b", 131072),
+        Model("llama-3.3-70b-versatile", k(128), deprecated=True, replacement="openai/gpt-oss-120b"),
+        Model("llama-3.1-8b-instant", k(128), deprecated=True, replacement="openai/gpt-oss-20b"),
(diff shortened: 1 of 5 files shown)


ocs-percy (approved):
Clean. The migration consolidation holds up: stripping `llm_model_migration()` from `0072` and `notify_deprecated_models` from `0071` is safe because already-applied environments will not re-run those, and fresh environments get everything they need from `0073`. `load_pricing_data()` correctly stays in `0072` given its idempotency. The retained dependency annotations on `0071`/`0072` are correct — dropping them would shift the migration graph for environments already past those points.

Model changes and integration test cleanup are correct. 🚢

Issue #3893, opened 2026-07-22
Add OAuth support to the chat widget and chat API


ocs-percy:
## Summary

Allow the chat widget and chat API to be secured via OAuth. Team admins can configure a widget channel to require OAuth authentication, restricting access to the chatbot to authorised users/clients only.

## Motivation

Currently the chat widget and API have no OAuth-based access control. For deployments where the chatbot should only be accessible to authenticated users or authorised services, there is no way to enforce this at the channel level.

## Proposed behaviour

- Team admins can enable an "Require OAuth" option on a widget channel configuration
- When enabled, incoming requests to the chat widget / chat API must present a valid OAuth access token
- Invalid or missing tokens result in a `401 Unauthorized` response
- The channel uses the OAuth application infrastructure (scoped to the team) to validate tokens

## Acceptance criteria

- [ ] Widget channel has a configurable "Require OAuth" toggle
- [ ] Chat API validates OAuth bearer tokens when the toggle is enabled
- [ ] Unauthenticated requests are rejected with `401`
- [ ] Authenticated requests proceed to the chatbot as normal
- [ ] Docs updated to cover the OAuth widget setup flow

## Dependencies

- #3710 — Support client-credentials OAuth applications scoped to a team (required to issue tokens that can be validated here)

## Notes

The primary use-case is machine-to-machine access (e.g. an internal app embedding the widget) using the client-credentials flow introduced in #3710.

PR #4224, opened 2026-08-18
Chat API Channel: credential_mode and a per-channel session lifetime

### LINKED ISSUE #4199: Replace chat session inactivity window with an absolute session lifetime

snopoke:
First implementation slice of [the chat API admission design](https://github.com/dimagi/open-chat-studio/blob/sk/oauth-chat-widget/docs/design/oauth-chat-widget.md) (#3893). **Stacked on #4203** — merge that first.

The design's server work is one coherent phase, and slicing it below that mostly produces dead code or a footgun. The one clean cut is here: the two channel columns are separable from the path that uses them, because `credential_mode` defaults to `embed_key` and stays inert until something can select it. That takes the migration off the OAuth PR's critical path and lets the D7 override ship now.

### Product Description

A chat widget channel can now set its own session lifetime instead of always using the 7-day global — an abuse-facing channel can tighten to hours. Otherwise no change: `credential_mode` is not yet reachable from the UI.

### Technical Description

`credential_mode` is deliberately **not** on the channel form. Nothing resolves an OAuth token yet, so selecting `oauth` today would leave the channel unreachable; a test guards that it stays off the form until the OAuth path lands.

The mode/level pin is enforced twice on purpose. `save()` pins `required_auth_level` to `SESSION_TOKEN` (and extends `update_fields` so a partial save carries it), and a `CheckConstraint` makes the pair unrepresentable rather than merely un-enterable — that combination issues no session token, so the session would 201 on create and 403 on every follow-up.

Two things worth a closer look:

- **The migration's dependencies are hand-trimmed to `bot_channels.0032`.** The autodetector's generated cross-app deps reorder a from-scratch plan so `service_providers.0071` runs ahead of `pipelines.0029`, and its `notify_deprecated_models` data migration then queries `pipelines_node.position_x` before that column exists — `--create-db` fails outright. Confirmed latent on a clean tree. **The ordering bug in `0071` is still there** and will bite the next migration that happens to pull those apps forward; out of scope here.
- **`EmbeddedWidgetChannelForm.__init__` now copies `self.initial`.** `ExperimentChannel.extra_form` passes the live `extra_data` dict, which Django binds without copying — so seeding a `timedelta` into `initial` wrote it straight into the JSON column. Pre-existing aliasing (four other forms do the same), harmless until now only because every prior write was JSON-serializable.

`session_token_expired` reads the override off the session cached by `get_experiment_session_cached`, so a lifetime change takes up to 5 minutes to take effect. The 5-minute floor lives only in the form — the model and DB still accept `timedelta(0)`, which matters once the field gets writers beyond the form.

Closes #4199 (the per-channel override ADR-0054 deferred).

### Migrations

Additive: two columns with defaults, a choices-only `AlterField`, and one check constraint every existing row already satisfies. Old code ignores both columns.

### Docs and Changelog


apps/channels/forms.py
```diff
@@ -685,6 +711,13 @@ def clean(self):
         """Generate or preserve the widget token"""
         cleaned_data = super().clean()
 
+        # Popped for the same reason as allow_all_domains: whatever is left in cleaned_data
+        # becomes the channel's extra_data, and this one has a column of its own.
+        self._session_token_lifetime = cleaned_data.pop("session_token_lifetime", None)
+        if self._session_token_lifetime is not None and self._session_token_lifetime < MIN_SESSION_TOKEN_LIFETIME:
+            # A lifetime this short makes every session dead on arrival.
+            raise ValidationError({"session_token_lifetime": "The session lifetime must be at least 5 minutes."})
+
```
snopoke: Fixed in 79e3de42a — extracted the session-lifetime handling into `_pop_session_token_lifetime`, which drops `clean()` back under the threshold. Fair finding: `clean()` had grown to do three unrelated things (lifetime, domains, widget token) behind a docstring that only mentioned the token.

A local CodeScene review confirms `EmbeddedWidgetChannelForm.clean` is no longer flagged; the two findings still reported for this file are in `SlackChannelForm`, which this PR does not touch.

*— Claude Code*

docs/design/oauth-chat-widget.md
```diff
@@ -806,16 +806,16 @@ PR #4198 delivered the per-application allowlist, so none of those appear here a
...
 | 7 | `apps/api/exceptions.py` | `ChatApiAccessDenied(NotAuthenticated)` — body and `code` only; the `401` comes from the authenticator (D6). |
 | 8 | `apps/api/authentication.py` | `ChatOAuthAuthentication`: resolve chatbot → Chat API Channel in `oauth` mode, validate token + origin, return `(AnonymousUser(), channel)`; `authenticate_header` → `Bearer realm="api"` (D3). |
 | 9 | `apps/api/throttling.py` | **No change** — the existing `ExperimentChannel` branch buckets OAuth callers per channel. |
-| 10 | `apps/api/session_tokens.py` | ~~age check against `created_at`; `last_activity_at` branch deleted~~ — **shipped in PR #4204** (ADR-0054). Reading the per-channel override is what remains. |
+| 10 | `apps/api/session_tokens.py` | ~~age check against `created_at`; `last_activity_at` branch deleted~~ — **shipped in PR #4204** (ADR-0054). ~~Reading the per-channel override~~ — **shipped**. Note the lifetime is read off `get_experiment_session_cached`'s cached session, so a change to a channel's value takes up to `WIDGET_SESSION_CACHE_TTL` (5 min) to take effect. |
```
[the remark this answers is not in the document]
snopoke: Fixed in 5d3ee939d. The override does ship in this PR, so the trailing D7 paragraph was the stale half — rewritten to say nothing of D7 is outstanding.

Two more instances of the same contradiction that weren't flagged: the D7 section header ("Largely shipped… #4199 stays open for the per-channel override") and the configuration paragraph ("the piece #4199 still tracks" / "Until the override lands, an `oauth` channel cannot tighten below the global"). Both corrected in the same commit.

*— Claude Code*

CONTEXT.md
```diff
@@ -53,15 +53,15 @@ So "this Chatbot has no Chat API Channel" means "no outside embedder can reach i
...
+- **Credential Mode** — the column exists (`ExperimentChannel.credential_mode`) but nothing reads it yet, so it is not admin-selectable and every Channel sits at `embed_key`. Once the OAuth path lands ([oauth-chat-widget.md](docs/design/oauth-chat-widget.md) D1) it is the *admin's* choice of what an *external* caller must present: the **Embed Key**, or an OAuth token. (A signed-in team member reaches an in-app embed through membership, presenting neither.) Under the OAuth mode an Embed Key is *ignored rather than rejected*, so an existing snippet needs no edit beyond adding the token — but the token is required, and the key alone no longer admits anyone. Whether that mode serves a browser or a server integration is told by the Channel's allowed domains, not by a separate setting: a blank list means server-only.
 - **Widget Auth Level** — a *version floor*, raised automatically as the deployed widget is upgraded (ADR-0045). It describes what old widgets on the page are capable of, never what the admin wants required.
```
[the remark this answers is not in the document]
snopoke: Fixed in 174b8a0a2 — now "never what the admin requires".

*— Claude Code*

apps/channels/models.py
```diff
@@ -37,7 +38,9 @@ class ChannelPlatform(models.TextChoices):
     SLACK = "slack", "Slack"
     COMMCARE_CONNECT = "commcare_connect", "CommCare Connect"
     EVALUATIONS = "evaluations", "Evaluations"
-    EMBEDDED_WIDGET = "embedded_widget", "Embedded Widget"
+    # Label only — the stored value stays `embedded_widget` because it is also a
+    # Participant.platform value, and renaming it would fork every existing participant.
```
snopoke:
```suggestion
```

apps/channels/models.py
```diff
@@ -252,6 +267,28 @@ class ExperimentChannel(BaseTeamModel):
     # ratchet_widget_auth_levels; not audited (the audited change is required_auth_level itself).
     pending_auth_level = models.PositiveSmallIntegerField(choices=WidgetAuthLevel.choices, null=True, blank=True)
     auth_level_notified_at = models.DateTimeField(null=True, blank=True)
+    # Which credential admits a caller to `chat/start/` on this channel. Nothing resolves an
+    # OAuth token yet, so this is not admin-selectable and every row sits at EMBED_KEY —
+    # today's behaviour exactly.
```
snopoke:
```suggestion
```

apps/channels/forms.py
```diff
@@ -648,6 +654,17 @@ class EmbeddedWidgetChannelForm(ExtraFormBase):
         help_text="Enter the domains where this widget is allowed to be embedded (one per line).",
     )
 
+    session_token_lifetime = forms.DurationField(
```
SmittieC, line 657: What happens when I set `session_token_lifetime` to something like `12:800:1200`?

SmittieC (approved):
LGTM, one question.

apps/channels/forms.py
SmittieC: I think the 800 minutes will be treated as 13 hours and some. Not a super big deal, but something worth considering

snopoke: Yup, `12:800:1200` = `1 01:40:00`

The field get's broken up using a regex and passed to `timedelta` so it is equivalent to `timedelta(hours=12, minutes=800, seconds=1200)`

Issue #3979, opened 2026-07-28
Dashboard: add cost breakdowns by model, service kind, and time


ocs-percy:
## Context

The dashboard already has a cost-tracking panel (behind `flag_ai_cost_monitoring`) that shows:
- Cost over time (daily bar chart) 
- Total cost for the selected period
- Top-N bots by cost (`top_n_bots`)
- Confidence footnote (exact / estimated / unknown)

It is missing richer breakdowns that would make the panel more actionable.

## What to add

**Break down cost chat by source** (done)
- chat
- evals

**Provider breakdown**
- Bar chart showing split by llm provider

**Model breakdown**
- Bar or stacked chart / table showing cost split by `model_name` for the selected period

**Service-kind breakdown**
- Donut / stacked bar of input tokens vs. output tokens vs. cached input vs. cache-write cost
- Lets users understand cache savings at a glance

**Cost column added to the Participant table**

**p95 cost per trace**
- line chart
- one line per chatbot

## Data availability

All data lives in `UsageRecord` with `(team, experiment, timestamp)` and `(team, session, timestamp)` indexes. No new model fields required — only new query helpers in `apps/cost_tracking/services/reporting.py` and corresponding view endpoints / template partials.

## Flag

Keep everything behind `flag_ai_cost_monitoring` so it rolls out incrementally.

barry47products:
## Design

Everything here builds on the existing cost panel plumbing: helpers in `apps/cost_tracking/services/reporting.py`, flag-gated endpoints in `apps/dashboard/views.py`, `DashboardCache` on the usual 30-minute TTL, Chart.js on the frontend. No model changes; the `(team, model_name, timestamp)` and `(team, timestamp)` indexes already cover these queries.

On source semantics: provider, model, and service kind aren't entities, so those breakdowns follow the ADR-0048 team-total rule. An unfiltered read counts chat and evaluation spend; a read narrowed by the dashboard filters counts chat only, through the existing `_scoped_records` path. The participant column and the p95 chart attribute cost to entities, so they count chat only.

### Provider and model breakdown

A single helper grouping scoped records by `(provider_type, model_name)` with `Sum(cost)` feeds both charts. The provider bar chart sums per provider client-side; the model chart plots cost per model directly. Both are served from a new `api/cost-breakdown/` endpoint, shared with the service-kind data below.

### Service-kind breakdown

[...]

PR #4227, opened 2026-08-18
Add participant cost columns to the dashboard and participants page

barry47products:
### Product Description

Part 2 of 3 for #3979, behind `flag_ai_cost_monitoring`. Both participant tables now show what each participant's conversations cost:

- The dashboard's **Most Active Participants** table gains a **Cost** column that follows the page's filters (date range, chatbot, channel, participant, tag).
- The participants page gains a **Cost (30d)** column showing each participant's spend over the last 30 days (the dashboard's default window).

Both columns show chat spend only: evaluation spend is the team's, never a participant's (ADR-0048). With the flag off, neither page changes.

**Stacked on #4225** (`bt/cost-breakdown-charts-3979`) - merge that first; this PR's diff includes it until then.

### Technical Description

- `costs_by_participant` in `apps/cost_tracking/services/reporting.py` mirrors `costs_by_experiment`; both now delegate to a shared `_entity_cost_map` grouping helper. Chat-only via the existing `_attributable_records` path, null participants excluded.
- The dashboard wiring mirrors the Bot Performance cost column: `include_cost` pass-through from the view's flag check, a cost-prefixed `DashboardCache` key, and the cost map bounded to the table's top 10 participants via `CostFilters(participant_ids=...)`.
- The participants page computes cost per page of results in `ParticipantTableView.get_table` and excludes the column via `table.exclude` when the flag is off (the `ChatbotSessionsTable` pattern). The column is deliberately **not sortable**: `UsageRecord` has no `(team, participant)` index, so sorting every participant by cost is the one expensive query shape; the per-page read is bounded to the page's participant ids instead.

### Migrations

No migrations - no model changes, no new indexes.

### Demo

| Table | Screenshot |
| --- | --- |
| Dashboard - Most Active Participants with the Cost column | pr2-most-active-participants-card |
| Participants page - Cost (30d) column | pr2-participants-page-cost-30d |

### Docs and Changelog


apps/cost_tracking/tests/test_reporting.py
```diff
@@ -20,6 +20,7 @@
     cost_total,
     costs_by_experiment,
     costs_by_model,
+    costs_by_participant,
```
barry47products: The file follows a one-test-class-per-public-reporting-function structure, and the new `TestCostsByParticipant` class follows it. Splitting the file inside this PR would churn a file the base branch (#4225) also touches and create rebase conflicts across the #3979 stack, so I'd rather do it as a standalone cleanup once the stack merges - e.g. moving the entity-breakdown classes (`TestCostsByExperiment`, `TestCostsByParticipant`, `TestCostsByModel`, `TestCostsByServiceKind`) into a `test_reporting_breakdowns.py`. If PR 3 (the p95 chart) adds further classes here, the split can land there instead of growing the file a third time.

snopoke (approved):

Issue #4078, opened 2026-08-04
Add LiteLLM support


ocs-percy:
## Add LiteLLM Support

Integrate [LiteLLM](https://github.com/BerriAI/litellm) as a supported LLM provider/proxy in Open Chat Studio.

LiteLLM provides a unified interface to 100+ LLMs and can be self-hosted as a proxy, making it popular for teams that want to centralise model access, add spend controls, and maintain a consistent API surface regardless of underlying provider.

### What needs to happen

- [ ] Add LiteLLM (proxy) as a provider option in the model configuration UI
- [ ] Support configuring a custom LiteLLM base URL and API key per team/org
- [ ] Map available models from the LiteLLM `/models` endpoint (or allow manual entry)
- [ ] Handle LiteLLM-specific behaviours (e.g. streaming differences, error formats)
- [ ] Test against a self-hosted LiteLLM instance with multiple backend models

### Notes

LiteLLM exposes an OpenAI-compatible API, so this may share a code path with the OpenRouter integration. Consider whether both can be handled under a generic "OpenAI-compatible endpoint" provider type.

airouter-dev:
A generic OpenAI-compatible provider looks preferable to adding another enum member for every proxy. One data-model detail is worth settling first: `LlmProviderModel` is currently scoped by team, model name, and provider **type**, while the endpoint and credentials live on the `LlmProvider` instance. If two compatible endpoints expose the same model ID, a shared `openai_compatible` type could make their discovered models collide. Associating custom/discovered model rows with the provider instance (or a stable provider-scoped identifier) would avoid that.

A narrow first slice could be:

- Add an `openai_compatible` provider that reuses `OpenAIGenericService`, with required base URL and API key fields. Keep it separate from the first-party `openai` type, since that type currently advertises transcription and Assistants capabilities that a generic endpoint may not implement.
- Normalize the configured API root once, accepting either a host root or a root ending in `/v1`, so requests cannot become `/v1/v1/...`.
- Discover models from `GET /models` relative to that normalized API root, but retain manual model entry when discovery is disabled, filtered, or unavailable.
- Treat capabilities as explicit per-provider/per-model settings. Start with Chat Completions; do not infer Responses, Assistants, transcription, vision, or tool support from the model name.
[...]

PR #4233, opened 2026-08-19
feat(llm): add OpenRouter as an OpenAI-compatible LLM provider

### LINKED ISSUE #4077: Add OpenRouter support
## Add OpenRouter Support

Integrate [OpenRouter](https://openrouter.ai) as a supported LLM provider in Open Chat Studio.

OpenRouter provides a unified API to access models from many providers (OpenAI, Anthropic, Google, Meta, Mistral, etc.) under a single endpoint and billing account.

### What needs to happen

- [ ] Add OpenRouter as a provider option in the model configuration UI
- [ ] Support OpenRouter API key configuration (per team/org)
- [ ] Map OpenRouter model identifiers to OCS model metadata (name, capabilities, pricing)
- [ ] Handle OpenRouter-specific auth headers (`HTTP-Referer`, `X-Title`)
- [ ] Test with a representative set of models available via OpenRouter

### Notes

OpenRouter is OpenAI-compatible, so integration should be relatively straightforward via the existing OpenAI-compatible provider path.

ariefandw:
### Product Description
Adds OpenRouter as an LLM provider option. It's an aggregator that gives access to a ton of models (OpenAI, Anthropic, Meta, Google, DeepSeek, etc.) through one OpenAI-compatible API and one key. Closes #4077.

### Technical Description
OpenRouter only supports the OpenAI chat-completions API (no Responses API), so this just reuses the same path we already have for Groq/Perplexity/MiniMax:

- New `LlmProviderTypes.openrouter` in `apps/service_providers/models.py`, wired to `OpenAIGenericConfigForm`/`OpenAIGenericService` with `openai_api_base=https://openrouter.ai/api/v1`.
- A handful of default models seeded so you can pick one right away (`openai/gpt-4.1-mini` default, `anthropic/claude-sonnet-4.6` translation default).
- `_openrouter()` env loader for `OPENROUTER_API_KEY`, added to `.env.example`.
- No migration (choices don't touch the schema). Tests cover provider tagging, routing/base URL, and credential loading.

One honest caveat: going through **LiteLLM** would probably be the "right" long-term answer since it'd cover OpenRouter plus many more providers uniformly — but it's a much bigger refactor and #4078 is tracking it separately. I'd rather land this small, working OpenRouter integration now and treat LiteLLM as the follow-up.

### Migrations
None.

### Demo
Picked OpenRouter in the add-provider dropdown, added an API key, and it answered a chat message (tested against a local deployment).

### Docs and Changelog

Changelog: new LLM provider type. Refs #4078 for the future LiteLLM work.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -207,9 +207,6 @@ APP_IP=
 
 # OpenRouter
 # OPENROUTER_API_KEY=
-# Optional: customize the attribution headers sent to OpenRouter (see https://openrouter.ai/docs/api-reference/overview#headers)
-OPENROUTER_REFERER=https://github.com/dimagi/open-chat-studio
-OPENROUTER_TITLE=Open Chat Studio
 
 # LiteLLM
 # LITELLM_API_KEY=
--- a/.env.example
+++ b/.env.example
@@ -205,6 +205,9 @@ APP_IP=
 # Perplexity
 # PERPLEXITY_API_KEY=
 
+# OpenRouter
+# OPENROUTER_API_KEY=
+
 # LiteLLM
 # LITELLM_API_KEY=
 # LITELLM_API_BASE=
--- a/.env.example
+++ b/.env.example
@@ -207,6 +207,9 @@ APP_IP=
 
 # OpenRouter
 # OPENROUTER_API_KEY=
+# Optional: customize the attribution headers sent to OpenRouter (see https://openrouter.ai/docs/api-reference/overview#headers)
+OPENROUTER_REFERER=https://github.com/dimagi/open-chat-studio
+OPENROUTER_TITLE=Open Chat Studio
 
 # LiteLLM
 # LITELLM_API_KEY=
@@ -220,3 +223,4 @@ APP_IP=
 ## wrong container or the port mapping is invisible to `docker ps`.
 # OCS_POSTGRES_CONTAINER=
 # OCS_REDIS_CONTAINER=
+
--- a/apps/service_providers/llm_service/credentials.py
+++ b/apps/service_providers/llm_service/credentials.py
@@ -124,14 +124,19 @@ def _openrouter() -> ProviderCredentials | None:
     api_key = os.environ.get("OPENROUTER_API_KEY")
     if not api_key:
         return None
+
+    from django.contrib.sites.models import Site  # noqa: PLC0415 - local import to avoid circular dep at module load
+
+    from apps.web.meta import get_server_root  # noqa: PLC0415 - local import to avoid circular dep at module load
+
     return ProviderCredentials(
         LlmProviderTypes.openrouter,
         "OpenRouter",
         {
             "openai_api_key": api_key,
             "default_headers": {
-                "HTTP-Referer": os.environ.get("OPENROUTER_REFERER", "https://github.com/dimagi/open-chat-studio"),
-                "X-Title": os.environ.get("OPENROUTER_TITLE", "Open Chat Studio"),
+                "HTTP-Referer": get_server_root(),
+                "X-Title": Site.objects.get_current().name,
             },
         },
     )
--- a/apps/service_providers/llm_service/credentials.py
+++ b/apps/service_providers/llm_service/credentials.py
@@ -114,6 +114,13 @@ def _perplexity() -> ProviderCredentials | None:
     return ProviderCredentials(LlmProviderTypes.perplexity, "Perplexity", {"openai_api_key": api_key})
 
 
+def _openrouter() -> ProviderCredentials | None:
+    api_key = os.environ.get("OPENROUTER_API_KEY")
+    if not api_key:
+        return None
+    return ProviderCredentials(LlmProviderTypes.openrouter, "OpenRouter", {"openai_api_key": api_key})
+
+
 def _minimax() -> ProviderCredentials | None:
     api_key = os.environ.get("MINIMAX_API_KEY")
     if not api_key:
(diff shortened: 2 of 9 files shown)


apps/service_providers/llm_service/credentials.py
```diff
@@ -113,6 +113,29 @@ def _perplexity() -> ProviderCredentials | None:
...
+    api_key = os.environ.get("OPENROUTER_API_KEY")
+    if not api_key:
+        return None
+    return ProviderCredentials(
+        LlmProviderTypes.openrouter,
+        "OpenRouter",
+        {
+            "openai_api_key": api_key,
+            "default_headers": {
+                "HTTP-Referer": os.environ.get("OPENROUTER_REFERER", "https://github.com/dimagi/open-chat-studio"),
```
snopoke: This should use the actual server address. You can get it from the `apps.web.meta.get_server_root` helper

apps/service_providers/llm_service/credentials.py
```diff
@@ -113,6 +113,29 @@ def _perplexity() -> ProviderCredentials | None:
...
+                "X-Title": os.environ.get("OPENROUTER_TITLE", "Open Chat Studio"),
```
snopoke: You can use the display name from the Site object: `Site.objects.get_current().name`

apps/service_providers/llm_service/default_models.py
```diff
@@ -124,6 +124,14 @@ def k(n: int) -> int:
...
         Model("llama-3.1-70b-instruct", 131072),
     ],
+    "openrouter": [
+        Model("openai/gpt-4.1-mini", 1000000, is_default=True),
+        Model("openai/gpt-4.1", 1000000),
+        Model("anthropic/claude-sonnet-4.6", 1000000, is_translation_default=True),
+        Model("meta-llama/llama-3.3-70b-instruct", 131072),
+        Model("google/gemini-2.5-flash", 1048576),
+        Model("deepseek/deepseek-v4-flash", 1000000),
+    ],
```
snopoke: Since OpenRouter support a very large number of models I don't think we should provide any default models in OCS.

Rather, I think it would be better to provide the user with an easy way to add models. You can already do this in OCS but we could extend this for openrouter to use their model API to provide, search, autocomplete and model details including pricing.

This could be done as a followup PR in which case I would suggest leaving this list blank or leave `openrouter` out entirely.

An adjacent problem we need to solve how to handle model parameters for these models. The OpenRouter API does include a list of supported parameters for each model. It might be necessary to store those in OCS and refactor our parameter handling.

.env.example
```diff
@@ -209,3 +209,9 @@ APP_IP=
 
 # Perplexity
 # PERPLEXITY_API_KEY=
+
+# OpenRouter
+# OPENROUTER_API_KEY=
+# Optional: customize the attribution headers sent to OpenRouter (see https://openrouter.ai/docs/api-reference/overview#headers)
+OPENROUTER_REFERER=https://github.com/dimagi/open-chat-studio
+OPENROUTER_TITLE=Open Chat Studio
```
snopoke: Drop these in favour of the site config object.

snopoke (commented):
@ariefandw thanks for you contribution!

I have a few small comments but I think the primary one is to drop the default models from this PR.

apps/service_providers/llm_service/default_models.py
snopoke: Followup issue https://github.com/dimagi/open-chat-studio/issues/4258

An alternative is to mirror the top models we are already supporting for the other major platforms (OpenAI, Anthropic, Gemini). This might actually be a better place to start.

ariefandw:
Updated! Dropped the default models (tracking the discovery flow in #4258) and updated attribution headers (\HTTP-Referer\ and \X-Title\) to be derived dynamically from \get_server_root()\ and \Site.objects.get_current().name\.

apps/service_providers/llm_service/main.py
```diff
@@ -191,6 +191,10 @@ def replace_file_links(self, text: str, file: File, session: ExperimentSession)
 class OpenAIGenericService(LlmService):
     openai_api_key: str
     openai_api_base: str
+    # Extra HTTP headers forwarded verbatim to every request.
+    # Used by OpenRouter to carry attribution headers (HTTP-Referer / X-Title);
+    # None for all other generic providers so ChatOpenAI uses its own defaults.
+    default_headers: dict[str, str] | None = None
```
snopoke, line 197: This doesn't seem to get set anywhere except in `apps/service_providers/llm_service/credentials.py` which is only used by the `bootstrap_data` command.

If you want these to be set in production you need to create a subclass or else pass them in from `LlmProviderTypes`

snopoke (dismissed):
One comment but otherwise looks good. Feel free to merge and address the comment in a followup PR or in this one.

apps/service_providers/llm_service/main.py
snopoke: @ariefandw do you want to address this before merging?

snopoke: @ariefandw bump

snopoke (commented):
@ariefandw please fix CI and then this is GTG

snopoke (dismissed):

snopoke:
@ariefandw there are some lint errors that need to be addressed before this can be merged

snopoke:
@ariefandw bump

ariefandw:
@snopoke ready for re-review. main merged in, inline Site imports hoisted to module level (inline-imports CI), and added the choices migration (0082, LiteLLM-style). All checks green, nothing functional changed.

snopoke (approved):

PR #4245, opened 2026-08-20
feat: add the chatbot write API

SmittieC:
### Product Description
Chatbots can now be created and updated through the v2 API (`POST`/`PATCH /api/v2/chatbots/`), gated behind a new `chatbots:write` OAuth scope.

### Technical Description
- New `apps/api/v2/write/` package: serializers mirror `ChatbotSettingsForm`'s validation (voice as two top-level ids, participant allowlist normalized, unknown keys rejected — and published as `additionalProperties: false`). Parity is test-enforced.
- Gated by `DjangoModelPermissionsWithView` + `TokenHasOAuthResourceScope`, so a caller needs the same `add_experiment`/`change_experiment` the chatbot UI requires. Machine (client-credentials) tokens have no user, so their authorization rests on the scope and the token's pinned team.
- References resolve through `TeamScopedRelatedField`: a cross-team id misses the queryset and surfaces as a field-level 400, with no existence oracle.
- **Machine tokens are held to `OAuth2Application.allowed_chatbots` on write.** PATCH is gated (`ChatbotViewSet.partial_update`); POST is not, since a chatbot that doesn't exist yet can't be on any allowlist.
- PATCH takes a `select_for_update` row lock. **This serializes API PATCHes only** — `ChatbotSettingsForm.save()` takes no lock, so a settings-page save racing a PATCH is still last-write-wins. Flagging in case a reviewer wants the follow-up.

**Worth a reviewer's eye — the authenticators now write `team` onto the *Django* request** (see [here](https://github.com/dimagi/open-chat-studio/pull/4245/changes/22ec5405b3bf901203a7a9ae8f1f935c4af53297)). The one change here that reaches outside `apps/api/v2/`. An attribute assigned to DRF's `Request` lives on the wrapper alone, so `clone_request` — which OPTIONS metadata uses to re-check permissions — couldn't see `request.team`; it fell through to `TeamsMiddleware`'s lazy lookup, `None` on an API path, and team-scoped `get_queryset()` raised. Both authenticators now write through via `set_request_attrs` (`apps/teams/helpers.py`) and the `request_team()` workaround is gone.

- Shared plumbing: v1, v2 and every OAuth-authenticated view get this. I found no readers of `team` on the Django request outside `TeamsMiddleware`, but that's the blast radius to sanity-check.
- It's the pattern DRF already uses for `user`/`auth`, whose setters write through for the same reason. `user` deliberately stays on that setter — its getter re-enters authentication.
- Covered by `test_the_authenticated_team_survives_a_clone_request`; confirmed load-bearing by reverting the write-through, which also fails the two existing OPTIONS tests.

### Migrations

`Experiment.owner` widened to nullable (`0150_alter_experiment_owner`) so an API-created chatbot need not have an owner. `DROP NOT NULL` is the safe rolling-deploy direction: old code always sets `owner`. `owner` is in `VERSION_IDENTITY_FIELDS`, so version cloning is untouched.

**Worth a reviewer's eye:** `resolve_fk` (`apps/teams/export/importer.py:80-91`) returns `None` for a nullable field rather than raising, so a team-sync owner that can't be resolved on the target now **silently nulls** instead of failing hard. Users sync ahead of chatbots, so this should rarely fire.

**Pulling this branch:** the migration was renumbered `0149`→`0150` to clear a collision with `main`. Run pytest once with `--create-db`.

### Demo
N/A — API-only; see `apps/api/v2/write/tests/`.

### Docs and Changelog

Docs need: the create/update endpoints, the `chatbots:write` scope, references given as ids from the discovery endpoints, and the allowed-chatbots requirement for machine tokens.


apps/api/tests/test_permissions.py
```diff
@@ -54,3 +49,40 @@ def test_api_key_views_keep_the_read_only_gate(view):
     """Overriding ``permission_classes`` replaces the project defaults, which is how the read-only
     API-key gate (ADR-0021) gets dropped. Every API-key view must still carry it."""
     assert ReadOnlyAPIKeyPermission in view.permission_classes
+
+
+def test_chatbots_write_scope_is_declared():
```
SmittieC: low value test

apps/api/tests/test_permissions.py
```diff
@@ -54,3 +49,40 @@ def test_api_key_views_keep_the_read_only_gate(view):
...
+    """TokenHasOAuthResourceScope derives `chatbots:write` for write verbs from
+    `required_scopes = ["chatbots"]`; the scope has to exist in settings for a token to carry it."""
+    assert "chatbots:write" in settings.OAUTH2_PROVIDER["SCOPES"]
+
+
+def test_chatbots_write_is_grantable_to_machine_tokens():
```
SmittieC: also low value test

apps/api/v2/write/tests/test_auth.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def test_write_mixin_keeps_the_read_only_api_key_gate():
+    """Overriding permission_classes replaces the project defaults, which is how the read-only
+    API-key gate (ADR-0021) gets silently dropped."""
+    assert ReadOnlyAPIKeyPermission in ChatbotWriteMixin.permission_classes
+    assert ChatbotCompositionPermission in ChatbotWriteMixin.permission_classes
+
+
+def test_write_mixin_declares_the_chatbots_scope():
```
SmittieC: low value test

apps/api/v2/write/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+            # A client-credentials (machine) token has no user behind it.
+            owner=request.user if request.user.is_authenticated else None,
+            name=validated_data["name"],
+            description=validated_data["description"],
+            pipeline=pipeline,
+        )
+
+
+class ChatbotCreatedSerializer(serializers.Serializer):
+    """The create response: spec 5.1's three keys and nothing more."""
```
SmittieC: don't mention the spec in the docstring

apps/api/v2/write/tests/test_chatbot_create.py
```diff
@@ -0,0 +1,137 @@
...
+def client(team):
+    return ApiTestClient(team.members.first(), team)
+
+
+def test_create_url_is_the_registered_list_route():
+    assert reverse("api:v2:chatbot-list") == CREATE_URL
+
+
+@pytest.mark.django_db()
+def test_create_returns_exactly_the_three_spec_keys(client):
```
SmittieC: low value test

apps/api/v2/lookups.py
```diff
@@ -24,6 +24,11 @@ def get_working_chatbot(team, public_id, *, lock: bool = False) -> Experiment:
 
     ``lock`` takes a row lock for the rest of the transaction, which a writer needs and a reader
     does not.
+
+    DRF's ``get_object_or_404``, not Django's: ``public_id`` is a ``UUIDField`` and the router's
```
SmittieC: this comment needs to be at the top where the import happens

apps/api/v2/write/serializers.py
```diff
@@ -155,23 +112,82 @@ class ChatbotWriteSerializer(serializers.ModelSerializer):
...
 
-    def update(self, instance, validated_data):
-        # `voice` has no source, so ModelSerializer.update would set a junk attribute; apply the
-        # pair explicitly. The sentinel separates "omitted" (leave alone) from null (clear both).
-        voice = validated_data.pop("voice", _MISSING)
-        if voice is not _MISSING:
-            instance.voice_provider = voice["voice_provider"] if voice else None
-            instance.synthetic_voice = voice["synthetic_voice"] if voice else None
-        return super().update(instance, validated_data)
+    def validate(self, attrs):
```
SmittieC: note to self: look at this more carefully

apps/api/v2/write/tests/test_chatbot_patch.py
```diff
@@ -163,6 +163,50 @@ def test_a_nonexistent_reference_is_a_400_on_that_field(client, chatbot, field):
...
+    response = client.patch(_url(chatbot), {"nmae": "Typo"}, format="json")
+
+    assert response.status_code == 400
+    assert "nmae" in response.json()
+    chatbot.refresh_from_db()
+    assert chatbot.name == "Support bot"
+
+
+@pytest.mark.django_db()
+def test_an_unrecognised_settings_key_is_a_400_naming_it(client, chatbot):
```
SmittieC: merge this into test_an_unrecognised_key_is_a_400_naming_it

apps/api/v2/write/serializers.py
```diff
@@ -128,29 +118,37 @@ class ChatbotWriteSerializer(RejectsUnknownKeys, serializers.ModelSerializer):
...
 
+    name = NfcCharField(max_length=128)
+    # Labelled from the model because declaring the field explicitly drops the label ModelSerializer
+    # would have derived, and that label is the field's `title` in the published OpenAPI schema.
+    description = OptionalTextField(label=Experiment._meta.get_field("description").verbose_name)
     settings = ChatbotSettingsSerializer(source="*", required=False)
     consent_form_id = TeamScopedRelatedField(
         source="consent_form",
-        get_team_queryset=lambda team: ConsentForm.objects.working_versions_queryset().filter(team=team),
+        scoped_queryset=lambda request: ConsentForm.objects.working_versions_queryset().filter(team=request.team),
```
SmittieC, line 103: Should these not use `request_team` instead of doing `request.team`?

apps/api/v2/views.py
```diff
@@ -62,13 +62,24 @@ def get_queryset(self):
         return working_chatbots(self.team).select_related("team").prefetch_related("versions")
 
     def get_serializer_class(self):
-        # The actions below build their serializers directly, but OPTIONS metadata and any future
-        # `self.get_serializer()` call resolve the class through here, so it has to be right.
+        # The actions below build their serializers directly, but any `self.get_serializer()` call
+        # resolves the class through here, so it has to be right.
+        action = self.action
+        if action == "metadata":
```
SmittieC, line 67: I'd like to see a test for this please

apps/api/v2/views.py
```diff
@@ -161,7 +161,9 @@ def inspect(self, request, id=None):
     def partial_update(self, request, *args, **kwargs):
         with transaction.atomic():
             # Model.save() writes every column, so without the row lock two concurrent PATCHes
-            # naming different fields would silently clobber one another (spec W7).
+            # naming different fields would silently clobber one another (spec W7). This serialises
```
SmittieC: remove this comment

apps/api/tests/test_permissions.py
SmittieC: Fixed in 17b49e38c — deleted. It only restated that a constant is in a settings list; the end-to-end coverage is `test_machine_token_creates_an_ownerless_chatbot` (creates with `chatbots:write`) and `test_oauth_token_without_the_write_scope_cannot_create` (refused with `chatbots:read`), neither of which passes unless the scope is declared and grantable.

apps/api/tests/test_permissions.py
SmittieC: Fixed in 17b49e38c — deleted alongside its twin, same reasoning.

apps/api/v2/write/tests/test_auth.py
SmittieC: Fixed in 17b49e38c — deleted. `assert ChatbotWriteMixin.required_scopes == ["chatbots"]` just read the attribute back.

apps/api/v2/write/serializers.py
SmittieC: Fixed in 423988e82. Now reads "The create response: the ids needed to keep working with the new chatbot, and nothing more."

Applied the same rule to the other four spec references in this PR, since the spec is not something a reader of this code can follow: the `test_chatbot_create.py` and `test_chatbot_patch.py` module docstrings, `(spec W6, "lenient on structure")`, and `(spec W7)` in the `partial_update` row-lock comment.

apps/api/v2/write/tests/test_chatbot_create.py
SmittieC: Fixed in 17b49e38c — deleted. Its key-set assertion largely restated `ChatbotCreatedSerializer`'s three declared fields, and the 201-plus-`id` path is exercised by every other create test.

Worth flagging how I read the anchor: it points at line 43 of 26fe3d48, which is this test, not the `test_create_url_is_the_registered_list_route` tautology just above it. I checked which one you meant before deleting; that `reverse()` one is still there if you want it gone too.

apps/api/v2/lookups.py
SmittieC: Fixed in 423988e82. The paragraph now sits directly above `from rest_framework.generics import get_object_or_404` and is out of `get_working_chatbot`'s docstring.

apps/api/v2/write/serializers.py
SmittieC: Left as-is and left this thread open, since it is your note to self rather than a change request. Summary of what is here, to make the read cheaper:

- `validate` no-ops when neither voice key was sent, deliberately, so a PATCH of an unrelated field cannot fail on a half-set voice a legacy row already holds.
- When either key was sent it builds the pair as it *would be* after the write: incoming value where one was sent, stored value otherwise. `attrs.get(..., getattr(self.instance, ...))` keeps "sent as null" (clear it) distinct from "not sent" (keep it).
- `_voice_pair_error` rejects half-set pairs, a provider/voice type mismatch, and a voice owned by a different provider. Both-null is valid.
- The error is raised against every key the request actually sent, so the 400 names a field the client wrote.

The bit I would look hardest at: `getattr(self.instance, ...)` is only reached with `self.instance` set today (PATCH), so the instance-less fallback is untested defence.

apps/api/v2/write/tests/test_chatbot_patch.py
SmittieC: Fixed in 17b49e38c. The two are now one parametrized test with `at-the-root` and `under-settings` cases; the settings case's `initial_data` explanation moved into the shared docstring.

apps/api/v2/write/serializers.py
SmittieC: Yes — fixed in 00bd60627. All four `scoped_queryset` lambdas now use `request_team(request)`, and so does `ChatbotCreateSerializer.create`, which was reading `request.team` for the same reason.

These lambdas are not reachable through the OPTIONS `clone_request` path today (`determine_actions` describes only PUT and POST, and this viewset has no `update`, so the PATCH serializer is never introspected), but the serializers were the last readers of the raw attribute and there is no reason for them to differ from the viewset and the write mixin.

apps/api/v2/views.py
SmittieC: Already covered — `test_options_advertises_the_body_that_post_actually_accepts` in `apps/api/v2/write/tests/test_chatbot_create.py`, added in f171c4d2c after you left this comment.

It asserts the OPTIONS POST body is exactly `{name, description}` with `name` required and `description` not, which only holds if this branch resolves `metadata` + POST to `ChatbotCreateSerializer`. Without the branch it would advertise the read serializer's `id`/`url`/`version_number`/`versions` and omit `description`. No new test added.

apps/api/v2/views.py
SmittieC: Removed in 423988e82. The comment is back to the two lines about `Model.save()` writing every column, minus the `(spec W7)` reference; the settings-form caveat from e59598bf1 is gone.

SmittieC:
The `test_create_seeds_a_start_llm_end_pipeline` failure reported above is fixed in e34e9997c.

It was a test-ordering bug, not a flake in the endpoint. The fixture relied on the global `LlmProviderModel` rows that migration `service_providers.0020` seeds, and `apps/evaluations/tests/test_auto_population_task.py` runs with `transaction=True` and no `serialized_rollback` — its flush drops those rows for the rest of that xdist worker session. `get_first_llm_provider_model` then found no default model, so the seeded pipeline came back as Start + End with no LLM node.

Reproducible without xdist:

```
pytest apps/evaluations/tests/test_auto_population_task.py \
       apps/api/v2/write/tests/test_chatbot_create.py -p no:randomly
```

The fixture now creates the global row it needs instead of inheriting it. The root cause — a transactional test flushing shared migration data — is still there and can bite any other test that leans on those rows; worth a separate look.

apps/api/v2/views.py
```diff
@@ -165,6 +175,10 @@ def partial_update(self, request, *args, **kwargs):
             # API writes only: ChatbotSettingsForm.save() takes no lock, so a settings-page save
             # racing a PATCH can still clobber. Locking the form too is the fix, if that matters.
             chatbot = get_working_chatbot(self.team, self.kwargs[self.lookup_url_kwarg], lock=True)
+            # A machine token reaches only the chatbots its application was pinned to. Checked after
+            # resolution because the allowlist is keyed on the chatbot, and before any write, so the
+            # refusal costs nothing but the lock this block releases on the way out.
+            enforce_application_chatbot_write(request, chatbot)
```
SmittieC: Note to self: I don't like this being checked here. it should be in the middleware

api-schemas/v2.yml
```diff
@@ -36,6 +36,34 @@ paths:
               schema:
                 $ref: '#/components/schemas/PaginatedChatbotList'
           description: ''
+    post:
+      operationId: chatbot_create
+      description: 'Create a chatbot''s working (draft) version, seeded with a Start
+        -> LLM -> End pipeline. Nothing is published: use POST /chatbots/{id}/versions/
+        for that. On a team with no LLM provider the seed is Start + End with no edge
+        between them, so the new chatbot reports pipeline_valid: false until you wire
+        it. A key that is not listed below is rejected rather than ignored.'
```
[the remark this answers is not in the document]
SmittieC: Fixed in 9778fe102. There is no `servers` block, so every path key in this document is `/api/v2/...` and the prose was the only place missing the prefix.

Two more instances in the same operation descriptions had it too, so all three now carry it: `POST /api/v2/chatbots/{id}/versions/`, `GET /api/v2/pipeline/options/` and `GET /api/v2/chatbots/{id}/inspect/`.

api-schemas/v2.yml
```diff
@@ -902,6 +975,94 @@ components:
...
+      properties:
+        name:
+          type: string
+          maxLength: 128
+        description:
+          type: string
+          nullable: true
+          default: ''
+      required:
+      - name
```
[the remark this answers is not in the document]
SmittieC: Fixed in 4cd4c5a5f. Agreed — the prose said it and the schema did not, and this API's consumer reads the schema.

Rather than hand-listing the three components, `mirror_unknown_key_rejection` (a postprocessing hook in `apps/api/schema.py`) walks `RejectsUnknownKeys.__subclasses__()` and derives the component names from the class names, including drf-spectacular's `Patched` prefix for a PATCH body. A new closed serializer therefore needs no edit to the hook.

Safe to close all three: every field on those serializers is writable, so the declared properties are exactly the keys `RejectsUnknownKeys` accepts. `test_the_schema_closes_each_body_that_rejects_unknown_keys` asserts that equality against `serializer().fields` — the same set the runtime check uses — so the two cannot drift, and `test_every_serializer_that_rejects_unknown_keys_is_closed_in_the_schema` catches a component the name derivation misses.

apps/chatbots/tests/test_chatbot_forms.py
```diff
@@ -91,3 +91,31 @@ def test_chatbot_settings_form_consent_form_queryset_is_team_scoped(team_with_us
...
+    (`normalize_participant_allowlist`), so pinning it here pins both callers."""
+    request = RequestFactory().get("/")
+    request.team = team_with_users
+    request.user = team_with_users.members.first()
+    set_current_team(team_with_users)
+
+    form = ChatbotSettingsForm(request, data={"name": "Bot", "participant_allowlist": entered})
+    form.is_valid()
+
+    assert form.cleaned_data["participant_allowlist"] == cleaned
```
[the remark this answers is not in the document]
SmittieC: Removed in 234b3aeca — `ChatbotSettingsForm` reads `request.team` (`apps/chatbots/forms.py:81`), so the call was doing nothing; the test passes without it.

One correction on the reasoning, though: the ContextVar cannot leak into a later test. `apps/conftest.py:86` has an autouse `_reset_team_context` fixture that calls `unset_current_team()` both before and after every test, so a missing token reset is contained either way. The call was worth dropping because it implied a dependency this test does not have, not because of cross-test bleed.

apps/api/v2/write/fields.py
```diff
@@ -0,0 +1,76 @@
...
+    not always the only thing that narrows the set -- the voice fields also have to apply the same
+    feature flag ``ChatbotSettingsForm`` applies, and a flag can be active for reasons that have
+    nothing to do with the team (see ``excluded_voice_services``).
+
+    A cross-team id simply misses the queryset, exactly as a nonexistent id does, so both surface
+    as a 400 on the field. Telling them apart would need a second, global query whose only effect
+    is to leak whether the id exists in another team.
+    """
+
+    def __init__(self, scoped_queryset, **kwargs):
```
snopoke: add a type hint for scoped_queryset

apps/api/v2/views.py
```diff
@@ -45,12 +53,53 @@ class ChatbotViewSet(mixins.ListModelMixin, mixins.RetrieveModelMixin, GenericVi
...
+        action = self.action
+        if action == "metadata":
+            # `self.action` is "metadata" for the whole of an OPTIONS request, so the mapping below
+            # would answer the *read* serializer for every method being described. DRF's
+            # `SimpleMetadata.determine_actions` swaps in a `clone_request` carrying the method it
+            # is describing before calling `get_serializer()`, so read the method instead. Without
+            # this, OPTIONS advertises a POST body whose keys `RejectsUnknownKeys` then 400s -- and
+            # OPTIONS is exactly how the agent this API is built for discovers the body.
+            # Only POST needs mapping: `determine_actions` describes PUT and POST alone, and this
+            # viewset has no `update`, so PATCH is never described.
```
snopoke: Can we reduce the verbosity of this comment?

apps/api/v2/views.py
```diff
@@ -45,12 +53,53 @@ class ChatbotViewSet(mixins.ListModelMixin, mixins.RetrieveModelMixin, GenericVi
...
+        ),
+        tags=["Chatbots"],
+        request=ChatbotCreateSerializer,
+        responses={201: ChatbotCreatedSerializer},
+    )
+    def create(self, request, *args, **kwargs):
+        serializer = ChatbotCreateSerializer(data=request.data, context=self.get_serializer_context())
+        serializer.is_valid(raise_exception=True)
+        chatbot = serializer.save()
+        return Response(ChatbotCreatedSerializer(chatbot).data, status=status.HTTP_201_CREATED)
```
snopoke: This naming is confusing similar to the `ChatbotCreateSerializer`. How about calling it `ChatbotCreateResponseSerializer`?

apps/api/v2/write/serializers.py
```diff
@@ -0,0 +1,269 @@
...
+            team=team,
+            # A client-credentials (machine) token has no user behind it.
+            owner=request.user if request.user.is_authenticated else None,
+            name=validated_data["name"],
+            description=validated_data["description"],
+            pipeline=pipeline,
+        )
+
+
+class ChatbotCreatedSerializer(serializers.Serializer):
```
snopoke: What about just using the `ChatbotDetailSerializer` instead of this one?

apps/api/v2/write/fields.py
SmittieC: Fixed in 6bc8115 — `scoped_queryset: Callable[[Request], QuerySet]`.

*— Claude Code*

apps/api/v2/views.py
SmittieC: Fixed in 2ccbc96 — cut from eight lines to four, keeping why the method is read off the `clone_request` and why only POST needs mapping.

*— Claude Code*

apps/api/v2/views.py
SmittieC: Resolved by taking your other suggestion in 2c079a3 — `ChatbotCreatedSerializer` is gone entirely, so create now responds with `ChatbotDetailSerializer` and there is no near-twin name left to confuse.

*— Claude Code*

apps/api/v2/write/serializers.py
SmittieC: Fixed in 2c079a3 — create now returns `ChatbotDetailSerializer` and `ChatbotCreatedSerializer` is deleted. The detail serializer is a superset (it already carried `id`, `pipeline_id` and `version_number`), so create just returns more of the new chatbot than before. Schema regenerated; write + api tests pass.

*— Claude Code*

snopoke (dismissed):

SmittieC:
@snopoke I had so resolve some conflicts with main

snopoke (approved):

PR #4247, opened 2026-08-20
Track team creators across team management and admin reporting

snopoke:
### Product Description

Internal administrators can identify the user who created a team in team search and cross-team provider and tracing reports.

### Technical Description

Adds an audited, nullable `Team.created_by` relation, records it in both team-creation flows, and backfills historical teams from their earliest membership (`created_at`, then primary key). Teams without a membership remain unset, and deleting the creator sets the relation to null.

Admin team, provider-key, provider-usage, and tracing APIs expose the same nullable creator object (`id`, `username`, and `email`) and eager-load the relation. The generated export API schema is updated to reflect the new response field.

Backend verification:

- `uv run pytest --no-migrations apps/teams/tests/test_creation_deletion.py apps/teams/tests/test_migration_0015_backfill_team_created_by.py apps/admin/tests/test_dashboard_views.py apps/admin/tests/test_provider_keys.py apps/admin/tests/test_provider_usage_api.py apps/admin/tests/test_tracing_usage_api.py -q` — 90 passed
- Scoped Ruff and Python type checks passed.
- `uv run python manage.py makemigrations --check --dry-run` — no changes detected.

### Migrations

The new foreign key is nullable, so the schema and backfill can be deployed while the previous application version is still running.

### Demo

N/A — this is an admin API/reporting change with no new end-user UI.

### Docs and Changelog

No external documentation change is required for this internal admin/reporting change.


apps/teams/tests/test_migration_0015_backfill_team_created_by.py
```diff
@@ -0,0 +1,39 @@
...
+from django.utils import timezone
+from field_audit.models import AuditAction
+
+from apps.teams.models import Membership
+from apps.utils.factories.team import MembershipFactory, TeamFactory
+from apps.utils.factories.user import UserFactory
+
+
+@pytest.mark.django_db()
+def test_backfill_uses_earliest_membership_and_leaves_teams_without_members_null():
```
SmittieC, line 16: are there teams without members? If so, should we not rather delete them, since they cannot be redeemed?

SmittieC (approved):
LGTM. One non blocking comment

apps/teams/tests/test_migration_0015_backfill_team_created_by.py
snopoke: There shouldn't be but it's not impossible.

PR #4253, opened 2026-08-21
Broadcast message support

### LINKED ISSUE #4252: Broadcast message support
Add the ability for a bot builder to send broadcast messages to all participants. They should be able to choose a specific set of channels (or default to all) on which this should be sent, as well as specify what the message should be.

## Implementation
From the chatbot home page, there should be a "broadcast message" button. When clicked, a modal should pop up showing the available channels on which this should be sent (They should be able to multi-select channels.) (excluding the API, web and evaluations channels) and a text box - accepting the message to be sent -  limited to `MetaCloudAPIService.TEMPLATE_MESSAGE_CHAR_LIMIT` characters. There should also be a cancel and "send" button.

### Note
If in the modal the whatsapp channel is ticked, we should show a mini warning telling the user to ensure that they have a `new_bot_message` whatsapp template configured for the provider being used (this is so we can actually send out of service window messages). You can have a link to [this page](https://docs.openchatstudio.com/how-to/whatsapp_meta_cloud_api/#create-the-required-template-in-meta-business-manager) there informing users about the template.

Once they click "send", a background task should be spawned that sends this message to all participants on the selected platforms for this chatbot.

SmittieC:
### Product Description

Bot builders can now send a one-off broadcast message to every participant of a chatbot. A "Broadcast message" button on the chatbot home page opens a modal where they pick the channels to send on (multi-select) and type the message. Sending queues a background task that delivers the message to every participant on the selected channels.

If a WhatsApp channel is ticked, the modal warns that the provider needs a `new_bot_message` template configured (broadcasts land outside the 24-hour service window) and links to the docs page explaining how to create it.

### Technical Description

Three commits:

1. **Background task** (`apps/chatbots/tasks.py`) — `send_broadcast_message` resolves the latest session per `(participant, channel)` and fans out one `send_broadcast_message_to_session` per session, which calls `ExperimentSession.ad_hoc_bot_message(message_text=...)` to deliver the text verbatim. The coordinator runs on the `background` queue so walking a large audience can't hold up inbound chat; individual sends stay on the chat queue. A broadcast reaches a participant through a session they already have — that session holds the platform address to deliver to.

2. **Form, view, URL** — `BroadcastMessageForm` offers the chatbot's own channels, excluding `team_global_platforms()` (API/Web/Evaluations) and disabled channels (the send path drops bot-initiated traffic on those). Message capped at `MetaCloudAPIService.TEMPLATE_MESSAGE_CHAR_LIMIT` on every platform, so one message is deliverable on all selected channels rather than being split on one. `POST chatbots/<id>/broadcast/` is gated on `experiments.invite_participants`.

3. **UI** — daisyUI modal with channel checkboxes, textarea with a live character counter, Cancel + Send (Send disabled until a channel and message are chosen), and the Alpine-driven WhatsApp template warning.

### Migrations

No migrations.

### Demo

image

### Docs and Changelog

Resolves #4252

### DIFF
--- a/apps/chatbots/forms.py
+++ b/apps/chatbots/forms.py
@@ -119,17 +119,29 @@ class CopyChatbotForm(forms.Form):
     )
 
 
+# Platforms with no way to push a message at a participant out of band. API and web sessions
+# have no address to deliver to and evaluation sessions have no human on the other end; the
+# chat widget routes through `ApiChannel`, whose sender is a `NoOpSender`, so a broadcast there
+# would land in the chat history and be delivered nowhere.
+#
+# Listed out rather than derived from `ChannelPlatform.team_global_platforms()`, which answers a
+# different question -- which platforms allow only one channel per team -- and no longer matches.
+NON_BROADCASTABLE_PLATFORMS = [
+    ChannelPlatform.API,
+    ChannelPlatform.WEB,
+    ChannelPlatform.EVALUATIONS,
+    ChannelPlatform.EMBEDDED_WIDGET,
+]
+
+
 def get_broadcast_channels(experiment: Experiment):
     """The channels a broadcast can go out on.
 
-    The team-global platforms are excluded: API and web sessions have no address to push a
-    message to, and evaluation sessions have no human on the other end. Disabled channels are
-    excluded too -- the send path drops bot-initiated traffic on them, so offering one would
+    Disabled channels are excluded alongside the platforms that can't be pushed to at all --
+    the send path drops bot-initiated traffic on a disabled channel, so offering one would
     only ever be a broadcast that silently goes nowhere.
     """
-    return experiment.experimentchannel_set.exclude(platform__in=ChannelPlatform.team_global_platforms()).exclude(
-        enabled=False
-    )
+    return experiment.experimentchannel_set.exclude(platform__in=NON_BROADCASTABLE_PLATFORMS).exclude(enabled=False)
 
 
 class BroadcastChannelWidget(forms.CheckboxSelectMultiple):
--- a/apps/chatbots/forms.py
+++ b/apps/chatbots/forms.py
@@ -132,6 +132,28 @@ def get_broadcast_channels(experiment: Experiment):
     )
 
 
+class BroadcastChannelWidget(forms.CheckboxSelectMultiple):
+    """Checkboxes tagged with their platform so the dialog can react to what is ticked.
+
+    The `data-platform` attribute is what arms the WhatsApp template warning; `x-model` feeds
+    the same selection to the send button.
+    """
+
+    def __init__(self, attrs=None):
+        super().__init__(attrs={"x-model": "selectedChannels", **(attrs or {})})
+
+    def create_option(self, name, value, label, selected, index, subindex=None, attrs=None):
+        option = super().create_option(name, value, label, selected, index, subindex, attrs)
+        # `value` is a ModelChoiceIteratorValue wrapping the channel's pk; the instance is on it.
+        option["attrs"]["data-platform"] = value.instance.platform
+        return option
+
+
+class BroadcastChannelField(forms.ModelMultipleChoiceField):
+    def label_from_instance(self, obj):
+        return f"{obj.platform_enum.label} — {obj.name}"
+
+
 class BroadcastMessageForm(forms.Form):
     """A one-off message sent to every participant of a chatbot on the chosen channels."""
 
@@ -140,16 +162,21 @@ class BroadcastMessageForm(forms.Form):
     # all of the selected channels rather than being silently split on one of them.
     MESSAGE_CHAR_LIMIT = MetaCloudAPIService.TEMPLATE_MESSAGE_CHAR_LIMIT
 
-    channels = forms.ModelMultipleChoiceField(
+    channels = BroadcastChannelField(
         queryset=ExperimentChannel.objects.none(),
-        widget=forms.CheckboxSelectMultiple,
+        widget=BroadcastChannelWidget,
         error_messages={"required": "Select at least one channel to broadcast on."},
     )
     message = forms.CharField(
         max_length=MESSAGE_CHAR_LIMIT,
-        widget=forms.Textarea(attrs={"rows": 4}),
+        widget=forms.Textarea(attrs={"rows": 5, "x-model": "message"}),
     )
 
     def __init__(self, experiment: Experiment, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["channels"].queryset = get_broadcast_channels(experiment)
+
+    @property
+    def eligible_channels(self):
+        """The channels on offer. Empty means there is nothing to broadcast on, so no dialog."""
+        return self.fields["channels"].queryset
--- a/apps/chatbots/forms.py
+++ b/apps/chatbots/forms.py
@@ -142,8 +142,8 @@ class BroadcastChannelWidget(forms.CheckboxSelectMultiple):
     def __init__(self, attrs=None):
         super().__init__(attrs={"x-model": "selectedChannels", **(attrs or {})})
 
-    def create_option(self, name, value, label, selected, index, subindex=None, attrs=None):
-        option = super().create_option(name, value, label, selected, index, subindex, attrs)
+    def create_option(self, name, value, *args, **kwargs):
+        option = super().create_option(name, value, *args, **kwargs)
         # `value` is a ModelChoiceIteratorValue wrapping the channel's pk; the instance is on it.
         option["attrs"]["data-platform"] = value.instance.platform
         return option
--- a/apps/chatbots/forms.py
+++ b/apps/chatbots/forms.py
@@ -173,10 +173,8 @@ class BroadcastMessageForm(forms.Form):
     # all of the selected channels rather than being silently split on one of them.
     MESSAGE_CHAR_LIMIT = MetaCloudAPIService.TEMPLATE_MESSAGE_CHAR_LIMIT
 
-    DEFAULT_ACTIVE_WITHIN_DAYS = 90
-    # Bounds the audience query's lookback so a broadcast can't be pointed at years of stale
-    # sessions -- three years comfortably covers any real "recently active" cutoff.
-    MAX_ACTIVE_WITHIN_DAYS = 365 * 3
+    DEFAULT_ACTIVE_WITHIN_DAYS = 14
+    MAX_ACTIVE_WITHIN_DAYS = 90
 
     channels = BroadcastChannelField(
         queryset=ExperimentChannel.objects.none(),
--- a/apps/chatbots/forms.py
+++ b/apps/chatbots/forms.py
@@ -2,8 +2,10 @@ from django import forms
 from django.db import transaction
 from waffle import flag_is_active
 
+from apps.channels.models import ChannelPlatform, ExperimentChannel
 from apps.experiments.models import ConsentForm, Experiment, SyntheticVoice
 from apps.pipelines.models import Pipeline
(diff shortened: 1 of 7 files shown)


templates/chatbots/components/broadcast_dialog.html
```diff
@@ -0,0 +1,67 @@
+{% load i18n %}
+<dialog id="broadcast_message_modal" class="modal">
+  {# `whatsappChannels` is the subset of the rendered channels that needs the template warning below. #}
```
SmittieC: remove this comment

templates/chatbots/components/broadcast_dialog.html
```diff
@@ -0,0 +1,67 @@
...
+  <div class="modal-box" x-data="{
```
SmittieC: why not just render `BroadcastMessageForm` here instead?

apps/chatbots/tasks.py
```diff
@@ -12,3 +15,49 @@ def send_bot_message(session_id: int, instruction_prompt: str):
...
+    the address to deliver to -- and only their latest session on a channel is the live
+    conversation, so older ones are skipped. Someone on two channels is messaged on both.
+    """
+    return list(
+        ExperimentSession.objects.filter(
+            experiment_id=experiment_id,
+            experiment_channel_id__in=channel_ids,
+        )
+        .order_by("participant_id", "experiment_channel_id", "-created_at")
+        .distinct("participant_id", "experiment_channel_id")
```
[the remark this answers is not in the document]
SmittieC: Already addressed in c20248e0d. `-id` now follows `-created_at` in the `order_by` chain (`apps/chatbots/tasks.py:46`), so `distinct("participant_id", "experiment_channel_id")` picks the newest session unambiguously when two sessions in a group share a `created_at`.

apps/chatbots/tests/test_broadcast.py
```diff
@@ -0,0 +1,141 @@
...
+    other_participant = ExperimentSessionFactory(experiment=experiment, experiment_channel=whatsapp)
+    ExperimentSessionFactory(experiment=experiment, experiment_channel=unselected)
+    # A participant of a different chatbot is not part of this broadcast.
+    ExperimentSessionFactory(experiment=ExperimentFactory(team=team), experiment_channel=telegram)
+
+    send_broadcast_message(experiment_id=experiment.id, channel_ids=[telegram.id, whatsapp.id], message="Hi all")
+
+    messaged = {call.kwargs["session_id"] for call in delay.call_args_list}
+    assert messaged == {latest_telegram.id, on_whatsapp_too.id, other_participant.id}
+    assert {call.kwargs["message"] for call in delay.call_args_list} == {"Hi all"}
```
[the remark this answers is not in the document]
SmittieC: Already addressed in c20248e0d. `assert delay.call_count == 3` now runs before the recipient-set comparison, so a duplicate send for the same session fails the test instead of being absorbed by the set.

apps/chatbots/views.py
```diff
@@ -353,6 +361,28 @@ def single_chatbot_home(request, team_slug: str, experiment_id: int):
...
+    else:
+        channels = form.cleaned_data["channels"]
+        send_broadcast_message.delay(
+            experiment_id=experiment.id,
+            channel_ids=[channel.id for channel in channels],
+            message=form.cleaned_data["message"],
+        )
+        platforms = ", ".join(sorted(channel.platform_enum.label for channel in channels))
+        messages.success(request, f"Your message is being sent to all participants on {platforms}.")
+    return redirect("chatbots:single_chatbot_home", team_slug=team_slug, experiment_id=experiment.id)
```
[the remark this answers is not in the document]
SmittieC: No code change needed — this is already enforced on both halves.

**The endpoint:** the view loads via `get_object_or_404(Experiment, ...)`, and `Experiment.objects` goes through `VersionsObjectManagerMixin.get_queryset`, which filters `is_archived=False` (`apps/experiments/versioning.py:419`). Since `is_editable` is defined as `not is_archived`, a direct POST for a non-editable chatbot 404s before form validation or `.delay()`.

**The dialog:** the include is already gated — `single_chatbot_home.html:139` opens with `{% if experiment.is_editable %}` and the `broadcast_form.eligible_channels and perms...` check is nested inside it.

Rather than add a redundant check, c20248e0d added `test_broadcast_view_rejects_a_chatbot_that_is_not_editable`, which archives the chatbot, POSTs the endpoint directly, and asserts a 404 with nothing queued — so the protection is pinned down by a test rather than left implicit.

templates/chatbots/components/broadcast_dialog.html
SmittieC: Removed in 8d2417142 — the `whatsappChannels` array it was explaining is gone too (see the reply on the next comment).

templates/chatbots/components/broadcast_dialog.html
SmittieC: Good call — done in 8d2417142. No reason not to; the hand-rolled version was declaring the field names, the character limit and the required flag a second time.

The dialog now renders the form via `{% render_field %}`, matching the rest of the app, and `maxlength`/`required` come off the field itself (`CharField.widget_attrs`) instead of being passed through the context.

Two things needed moving into the form to make it render properly:

- `BroadcastChannelField.label_from_instance` keeps the nice `WhatsApp — My Channel` option label (`ExperimentChannel.__str__` gives `Channel: My Channel (whatsapp)`, with the raw platform value).
- `BroadcastChannelWidget` puts `x-model` and a `data-platform` attribute on each checkbox, following the same `create_option` pattern as `EvaluatorCheckboxWidget` in `apps/evaluations/forms.py:44`.

That `data-platform` attribute also let the WhatsApp warning key off what is actually ticked, so the `whatsappChannels` array of channel ids is gone, and with it the last reason for the template to loop over channels — the view now passes only `broadcast_form`, not `broadcast_channels` and `broadcast_char_limit` as well.

Verified the rendered HTML rather than just the tests: no duplicate `class` attribute on the checkboxes, `data-platform` present per option, and `maxlength="924"`/`required` on the textarea. All 95 tests in `apps/chatbots/` pass, ruff/djLint/ty clean.

apps/chatbots/tasks.py
```diff
@@ -12,3 +15,50 @@ def send_bot_message(session_id: int, instruction_prompt: str):
...
+    """The most recent session per (participant, channel) for a broadcast.
+
+    A participant is reachable only through a session they already have -- that is what holds
+    the address to deliver to -- and only their latest session on a channel is the live
+    conversation, so older ones are skipped. Someone on two channels is messaged on both.
+    """
+    return list(
+        ExperimentSession.objects.filter(
+            experiment_id=experiment_id,
+            experiment_channel_id__in=channel_ids,
```
SmittieC, line 66: update this filter so that only sessions with status = active should be matched

SmittieC: Done in f81321564 — `status=SessionStatus.ACTIVE` is now part of the filter.

One thing worth confirming, since it shrinks the audience in a way that may not be obvious: `SETUP`/`PENDING` are the pre-conversation states, so excluding those is unambiguously right. But `session.end()` sets `PENDING_REVIEW`, and that is called when the bot ends a conversation, when an event action ends it, via the API, and from the session UI — so a participant whose last conversation was deliberately ended is now skipped rather than messaged. That reads as intended to me (their conversation is over), but it does mean a broadcast reaches fewer people than "everyone who has ever chatted to this bot". Say the word if you wanted the wider audience.

The one case it does *not* skip: the status filter runs before the newest-per-group pick, so a participant whose newest session has been ended is still reached on an older session that is still active, rather than being dropped entirely. `test_broadcast_falls_back_to_an_older_session_that_is_still_active` pins that.

Also added `test_broadcast_skips_a_session_that_is_not_active`, parametrized over all five non-active statuses. I checked these fail without the filter (6 failures) and pass with it, so they are actually holding the behaviour down.

apps/chatbots/forms.py
```diff
@@ -115,3 +117,66 @@ class CopyChatbotForm(forms.Form):
...
+    """
+
+    def __init__(self, attrs=None):
+        super().__init__(attrs={"x-model": "selectedChannels", **(attrs or {})})
+
+    def create_option(self, name, value, label, selected, index, subindex=None, attrs=None):
+        option = super().create_option(name, value, label, selected, index, subindex, attrs)
+        # `value` is a ModelChoiceIteratorValue wrapping the channel's pk; the instance is on it.
+        option["attrs"]["data-platform"] = value.instance.platform
+        return option
```
SmittieC: Fixed in 27887feca. `create_option` only reads `value`; the other five positional arguments existed purely to forward them to `super()`, so it now takes `*args, **kwargs` and passes them straight through. Under the argument limit, and it makes it obvious that nothing else is inspected.

apps/chatbots/tasks.py
```diff
@@ -12,3 +15,57 @@ def send_bot_message(session_id: int, instruction_prompt: str):
...
+    in it yet, and one that has been ended sits at `PENDING_REVIEW` or `COMPLETE` -- broadcasting
+    into either would push a message at someone who is not in a conversation with the bot. The
+    status is filtered before the newest-per-group pick, so a participant whose newest session
+    has been ended is still reached on an older one that is still active.
+    """
+    return list(
+        ExperimentSession.objects.filter(
+            experiment_id=experiment_id,
+            experiment_channel_id__in=channel_ids,
+            status=SessionStatus.ACTIVE,
```
snopoke, line 67: Should we put an activity timeout on this - I think the intention is to only message active / recently active participants right. So I think we should limit it to sessions that have had activity in the last X days. Maybe the web user can decide what the cutoff is exactly.

SmittieC: See https://github.com/dimagi/open-chat-studio/pull/4253/commits/14de9f257f1bd1af9dded485eaeb37c273808a2a

image

apps/chatbots/forms.py
```diff
@@ -115,3 +117,88 @@ class CopyChatbotForm(forms.Form):
...
+    )
+    active_within_days = forms.IntegerField(
+        label="Active in the last (days)",
+        initial=DEFAULT_ACTIVE_WITHIN_DAYS,
+        min_value=1,
+        help_text=(
+            "Only participants whose last activity falls inside this many days are messaged. "
+            "A participant who has not chatted since then is left alone."
+        ),
+    )
```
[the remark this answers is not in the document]
SmittieC: Confirmed and fixed in d59cfa818: added `MAX_ACTIVE_WITHIN_DAYS` (3 years) and wired it up as `max_value` on the `active_within_days` field.

apps/chatbots/forms.py
```diff
@@ -115,3 +117,92 @@ class CopyChatbotForm(forms.Form):
...
+
+    # The limit is WhatsApp's: a broadcast lands outside the 24-hour service window, so it goes
+    # out as a template message. Applied to every platform so the same text is deliverable on
+    # all of the selected channels rather than being silently split on one of them.
+    MESSAGE_CHAR_LIMIT = MetaCloudAPIService.TEMPLATE_MESSAGE_CHAR_LIMIT
+
+    DEFAULT_ACTIVE_WITHIN_DAYS = 90
+    # Bounds the audience query's lookback so a broadcast can't be pointed at years of stale
+    # sessions -- three years comfortably covers any real "recently active" cutoff.
+    MAX_ACTIVE_WITHIN_DAYS = 365 * 3
```
snopoke: I would drop these both agressively:

```suggestion
    DEFAULT_ACTIVE_WITHIN_DAYS = 14
    MAX_ACTIVE_WITHIN_DAYS = 90
```

snopoke (approved):

PR #4264, opened 2026-08-24
Write API | Chatbot options endpoint

### LINKED ISSUE #4259: Write API | Chatbot options
Spec: https://docs.google.com/document/d/1oD25fmxCjdam728UOinAj7bry15ih1pxdhf5paokalw/edit?tab=t.9rysxi5a1wpm

We need an endpoint similar to [pipeline options](https://www.openchatstudio.com/api/v2/docs/#tag/Pipelines/operation/pipeline_options), but for a chatbot's settings.

Something like `/chatbot/options`. The option fields + values need to be derived from `ChatbotSettingsForm` so that the UI and API is kept in sync without maintenance.

SmittieC:
### Product Description
Adds `GET /api/v2/chatbot/options/`, which tells an API client what values each of a chatbot's settings accepts for its team — the counterpart of `/pipeline/options/` for chatbot settings.

Fixes #4259

### Technical Description
The option lists are read off `ChatbotSettingsForm` rather than listed by hand: the form is instantiated with the API request and its fields are walked — a `ModelChoiceField` yields `{value, label}` off its team-scoped queryset, a `ChoiceField` yields its choices. No settings field is named in that module, so the UI and the API stay in sync without maintenance — including the `flag_open_ai_voice_engine` gate, which the API now honours automatically.

- `apps/api/v2/discovery/chatbot_settings.py` — new, the form introspection
- `ChatbotOptionsSerializer` documents today's keys for the OpenAPI schema; the view merges the derived payload over the serialized one so a new choice field is served rather than silently dropped
- `synthetic_voice` entries carry `type` and `provider_id` so a client can pair a voice with a speakable provider
- Auth matches the other discovery endpoints: `chatbots:read` scope, `experiments.view_experiment`, read-only API keys allowed
- 24 new tests; `api-schemas/v2.yml` regenerated

### Migrations

### Demo
```json
{
  "voice_provider": [{"value": 2, "label": "AWS: Prod Polly", "type": "aws"}],
  "synthetic_voice": [{"value": 11, "label": "English, female: *Joanna", "type": "aws", "provider_id": 2}],
  "voice_response_behaviour": [{"value": "reciprocal", "label": "Reciprocal"}],
  "trace_provider": [{"value": 4, "label": "Prod Langfuse", "type": "langfuse"}],
  "consent_form": [{"value": 7, "label": "Returns consent"}]
}
```

### Docs and Changelog
- [ ] This PR requires docs/changelog update (the OpenAPI schema is the doc and is regenerated here)


apps/api/v2/discovery/serializers.py
```diff
@@ -188,3 +188,65 @@ class PipelineOptionsSerializer(serializers.Serializer):
...
+    provider_id = serializers.IntegerField(
+        allow_null=True,
+        help_text=(
+            "The voice provider that owns this voice, or null for a shared voice any provider of the "
+            "same `type` can speak. A voice owned by a provider can only be paired with that one."
+        ),
+    )
+
+
+class SettingChoiceOptionSerializer(serializers.Serializer):
```
snopoke: almost identical to `SettingResourceOptionSerializer`

apps/api/v2/discovery/serializers.py
```diff
@@ -188,3 +188,65 @@ class PipelineOptionsSerializer(serializers.Serializer):
     default_llm_provider = DefaultLlmProviderSerializer(
         required=False, help_text="A provider/model pair that already satisfies the `must_match` rule."
     )
+
+
+class SettingResourceOptionSerializer(serializers.Serializer):
```
snopoke: This is basically the same as `ResourceOptionSerializer`

snopoke (dismissed):
There's some overlap between this and the `pipline-options` view. Any chance for merging them or at least reusing?

apps/api/v2/discovery/serializers.py
SmittieC: Gone — `/chatbot/options/` now uses `ResourceOptionSerializer` directly, along with `ProviderOptionSerializer` and `SyntheticVoiceOptionSerializer`. Their `value`/`label` help text is generalised to cover a chatbot setting as well as a node param, and the wording that only held for one endpoint moved onto the containing field, where it already varies per endpoint — the `must_match` join-key note is now on `PipelineOptionsSerializer.llm_provider_id`.

While in there: `ProviderOption`, `LlmProviderModelOption` and `SyntheticVoiceOption` were each redeclaring `value` and `label` identically, so they inherit `ResourceOptionSerializer` now too. That's six duplicate field declarations gone on top of the four `Setting*` classes.

Field order is unchanged, so response payloads and the schema's property order are untouched — the only schema diffs are component names and descriptions.

5f8c431ad

apps/api/v2/discovery/serializers.py
SmittieC: Renamed to `ChoiceOptionSerializer` and kept separate, deliberately: its `value` is a `CharField` where `ResourceOption`'s is an `IntegerField`. Folding the two together would publish an untyped `value`, and a client could no longer tell from the schema whether to write `4` or `"always"`. The docstring says as much now.

Happy to merge them anyway if you'd rather have the single type — it just costs that distinction in the published schema.

5f8c431ad

SmittieC:
> There's some overlap between this and the `pipline-options` view. Any chance for merging them or at least reusing?

Reusing, done — the four `Setting*` serializers are gone and `/chatbot/options/` now describes its payload with the same `ResourceOption` / `ProviderOption` / `SyntheticVoiceOption` types as `/pipeline/options/` (5f8c431ad). Net −58 lines, and four schema components collapsed into the existing ones. Replied on the two inline threads with the detail.

Merging the views I've left alone, and I think that's the right call: `ChatbotOptionsView` builds its payload from `ChatbotSettingsForm`, the pipeline views build theirs from `get_node_parameter_values`. Every helper on the shared `TeamOptionsView` base — `_clean_options`, `_is_placeholder`, `_describe_prompt_vars` — exists to strip builder affordances and prompt-variable shapes that the settings-form path never produces, so inheriting it would only carry dead code into the chatbot endpoint. Happy to revisit if there's a seam I'm not seeing.

apps/api/v2/tests/test_chatbot_options.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def _enable_voice_engine_flag_for(team):
+    flag, _ = Flag.objects.get_or_create(name="flag_open_ai_voice_engine")
+    flag.teams.add(team)
+    flag.flush()
+
+
+@pytest.mark.django_db()
+@pytest.mark.parametrize("auth_method", ["api_key", "oauth"])
```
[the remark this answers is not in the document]
SmittieC: Done in 1767ad4e4 — both cases now use `pytest.param(..., id=...)` (`api-key` / `oauth`).

snopoke (approved):

PR #89, opened 2023-10-26
Voice providers part 2

snopoke:
PR on top of https://github.com/dimagi/open-chat-studio/pull/88

This is the final piece for per-team voice providers (no more global Azure / AWS voice keys).

This links team voice providers to experiments and updates the UI to support selecting them and filtering the list of synthetic voices (see demo below).

It also refactors the audio generation code to make it easier to call.

### UI demo
Peek 2023-10-26 15-56

### Notes

Can only be deployed after https://github.com/dimagi/open-chat-studio/pull/88 has been deployed.

Resolves https://github.com/dimagi/open-chat-studio/issues/82

PR #4267, opened 2026-08-24
Cs/pipeline facade 1

SmittieC:
### Migrations

### Docs and Changelog


apps/api/v2/pipeline_edit/references.py
```diff
@@ -0,0 +1,90 @@
...
+        source = sources[param]
+        offered = options.get(source, [])
+        if not isinstance(offered, list):
+            # A dict-shaped source nests its options under provider types, so there is no flat set
+            # to check a value against. The two that exist are excluded by name; this keeps one
+            # added later from raising here rather than being skipped.
+            continue
+        allowed = {option["value"] for option in offered}
+        supplied = value if is_list_param(node_class, param) else [value]
+        unknown = [item for item in supplied if not _is_offered(item, allowed)]
```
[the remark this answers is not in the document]
SmittieC: Fixed in 1ea8186fe — good catch, this was a real 500. `supplied` now only takes `value` as-is when the value actually is a list/tuple/set, so a scalar sent for a list-valued reference is reported as one unreachable value instead of raising `TypeError` while iterating an int. Matches how `Node._sync_resource_fk_fields` already normalises `collection_index_ids` on the write side.

Added a regression test (`test_a_scalar_sent_for_a_list_valued_reference_is_refused`) which reproduced the 500 before the fix.

apps/pipelines/tests/test_models.py
```diff
@@ -2,6 +2,7 @@
 from unittest.mock import Mock, patch
 
 import pytest
+from django.db import DatabaseError
```
SmittieC: Fixed in fb36a030d. The class this PR added to `test_models.py` (`TestNodeValidationErrors`) now lives in its own module, `apps/pipelines/tests/test_node_validation_errors.py`, which puts the file back under the threshold.

apps/api/v2/pipeline_edit/graph_editor.py
```diff
@@ -0,0 +1,340 @@
...
+    end. One already clear to the right is left where someone put it, and only its x ever moves.
+
+    The move carries content because that is the only kind of node diff ``apply_pipeline_patch``
+    writes position columns for, and its params are narrowed like an edit's -- writing the graph's
+    copy back would store the resource-id mirror ``to_flow_node`` merges into every node.
+    """
+    moved = []
+    for stored in flow.get("nodes", []):
+        node = FlowNode(**stored)
+        if node.type != REACT_FLOW_END_TYPE or (node.position.get("x") or 0) > new_node_x:
```
SmittieC: Fixed in 62d14aeca. The conditional is now a named predicate, `_is_overtaken_end_node(node, new_node_x)` — which also says what the test means rather than leaving the reader to invert it.

apps/api/v2/pipeline_edit/views.py
```diff
@@ -0,0 +1,229 @@
...
+    def post(self, request, id: str) -> Response:
+        body = self.get_serializer(data=request.data)
+        body.is_valid(raise_exception=True)
+        label = body.validated_data.get("label")
+        node_type = body.validated_data["type"]
+        node_class = get_node_class(node_type)
+        params = writable_params(node_class, body.validated_data["params"])
+        check_references(option_lists_for(request.team, params, node_class), node_class, params)
+        return Response(
+            edit_pipeline(request, id, lambda flow: plan_create(flow, node_type, label, params), self._write_response),
```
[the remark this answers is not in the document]
SmittieC: Dismissed as a false positive, consistent with #89/#92 on `apps/pipelines/views.py`, which @snopoke closed with "These errors have string representations designed to give user feedback". The taint here is `Pipeline.validate()`'s error report reaching the response — the same report, now reached through the façade's write endpoints.

That said, one branch of it genuinely was leaky, so it is fixed in fb36a030d: `_node_validation_errors` had an unrestricted `except Exception` reporting `f"{type(e).__name__}: {e}"`. Unlike the other branches, whose messages are written for whoever sent the params, that one catches anything at all, so its message was as likely to expose how the server is put together as to say something useful. It now logs the exception and reports a fixed message; the node stays reportable, which was the point of the branch.

apps/api/v2/pipeline_edit/views.py
```diff
@@ -0,0 +1,229 @@
...
+        # Building the option lists that check needs is the slow half and needs no graph, so it is
+        # done here and handed in rather than built in the critical section.
+        options = option_lists_for(request.team, params)
+        return Response(
+            edit_pipeline(
+                request,
+                id,
+                lambda flow: plan_update(flow, options, node_id, label, params),
+                self._write_response,
+            )
```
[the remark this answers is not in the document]
SmittieC: Dismissed as a false positive, consistent with #89/#92 on `apps/pipelines/views.py`, which @snopoke closed with "These errors have string representations designed to give user feedback". The taint here is `Pipeline.validate()`'s error report reaching the response — the same report, now reached through the façade's write endpoints.

That said, one branch of it genuinely was leaky, so it is fixed in fb36a030d: `_node_validation_errors` had an unrestricted `except Exception` reporting `f"{type(e).__name__}: {e}"`. Unlike the other branches, whose messages are written for whoever sent the params, that one catches anything at all, so its message was as likely to expose how the server is put together as to say something useful. It now logs the exception and reports a fixed message; the node stays reportable, which was the point of the branch.

apps/api/v2/pipeline_edit/views.py
```diff
@@ -0,0 +1,229 @@
...
+            200: PipelineWriteSerializer,
+            403: FORBIDDEN,
+            404: OpenApiResponse(description="No such chatbot or node."),
+            409: OpenApiResponse(description="The node is part of the pipeline's structure and cannot be deleted."),
+        },
+    )
+    def delete(self, request, id: str, node_id: str) -> Response:
+        # `plan_delete` names no node, so the response reports the pipeline alone: there is no node
+        # left to describe.
+        return Response(edit_pipeline(request, id, lambda flow: plan_delete(flow, node_id), self._write_response))
```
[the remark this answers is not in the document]
SmittieC: Dismissed as a false positive, consistent with #89/#92 on `apps/pipelines/views.py`, which @snopoke closed with "These errors have string representations designed to give user feedback". The taint here is `Pipeline.validate()`'s error report reaching the response — the same report, now reached through the façade's write endpoints.

That said, one branch of it genuinely was leaky, so it is fixed in fb36a030d: `_node_validation_errors` had an unrestricted `except Exception` reporting `f"{type(e).__name__}: {e}"`. Unlike the other branches, whose messages are written for whoever sent the params, that one catches anything at all, so its message was as likely to expose how the server is put together as to say something useful. It now logs the exception and reports a fixed message; the node stays reportable, which was the point of the branch.

apps/api/v2/pipeline_edit/serializers.py
```diff
@@ -0,0 +1,132 @@
...
+        child=serializers.ListField(child=OutputHandleSerializer()),
+        help_text=(
+            "Advisory 'what still needs wiring' map, keyed by ``node_id``: every output handle with "
+            "no outgoing edge and every implicit ``input`` handle with no incoming edge. Never an "
+            "error and never blocks a publish."
+        ),
+    )
+
+
+class NodeWriteSerializer(PipelineWriteSerializer):
```
SmittieC, line 133: thought: maybe we should just return the node's validation state so as to not pollute the caller's context window

apps/api/v2/discovery/node_types.py
```diff
@@ -32,16 +40,62 @@ def get_node_types() -> list[dict]:
...
+    return cast(type[BasePipelineNode], resolve_node_class(node_type))
+
+
 def option_keys_for_node_type(node_type: str) -> frozenset[str] | None:
     """The option keys a single node type reads, or ``None`` if no such type is served."""
     return _option_keys_by_type().get(node_type)
 
 
+def reference_sources_for_type(node_type: str) -> dict[str, str]:
+    """``param -> the /pipeline/options/ key its value must be drawn from``, for one node type.
```
SmittieC: shorten docstring

apps/api/v2/discovery/options.py
```diff
@@ -0,0 +1,69 @@
...
+
+# The option lists holding prompt variables rather than referenceable resource ids.
+PROMPT_VAR_OPTION_SOURCES = (
+    OptionsSource.template_variables,
+    OptionsSource.llm_prompt_variables,
+    OptionsSource.router_prompt_variables,
+)
+
+
+def options_for_team(team: Team) -> dict:
```
SmittieC, line 23: shorten docstring

api-schemas/v2.yml
```diff
@@ -169,6 +169,164 @@ paths:
               schema:
                 $ref: '#/components/schemas/ChatbotInspect'
           description: ''
+  /api/v2/chatbots/{id}/pipeline/nodes/:
```
SmittieC, line 172: we need payload examples for each accepted node type

apps/api/v2/discovery/options.py
```diff
@@ -0,0 +1,69 @@
...
+    A dozen queries and a few hundred milliseconds, mostly parsing custom actions' OpenAPI schemas,
+    so callers that need it twice should hold the result rather than ask again.
+    """
+    options = _clean_options(get_node_parameter_values(team=team, usable_models_only=True))
+    options["default_llm_provider"] = get_node_default_values(team, usable_models_only=True)
+    return _describe_prompt_vars(options)
+
+
+def _clean_options(value: Any) -> Any:
+    """Strip the builder-only affordances off every option list. Recurses -- ``built_in_tools``
```
SmittieC: What is the "builder"? Please be more specific about the UI builder or API builder. Same goes for every other comment

apps/api/v2/discovery/contract.py
```diff
@@ -15,6 +18,38 @@
 # `param -> the field whose chosen option selects which sub-list of its options apply`.
 OPTIONS_KEYED_BY = {"built_in_tools": PROVIDER_TYPE_MATCH, "tool_config": PROVIDER_TYPE_MATCH}
 
+
+def _resource_source(param: str) -> OptionsSource | None:
+    """The option list a resource param draws its values from, found by name.
```
SmittieC: trim docstring

apps/api/v2/discovery/contract.py
```diff
@@ -15,6 +18,38 @@
...
+    ``test_every_mirrored_resource_param_is_checked`` is what says none of the params that *are*
+    offered fell through this way.
+    """
+    for candidate in (param, param.removesuffix("_ids"), param.removesuffix("_id")):
+        if candidate in OptionsSource.__members__:
+            return OptionsSource[candidate]
+    return None
+
+
+#: The option lists a team could be denied a value from, so a param drawing from one is checked
```
SmittieC, line 37: trim comment

apps/api/v2/discovery/node_types.py
SmittieC: Trimmed in 19e1708f2 — `parameter_option_mapping` (the renamed `reference_sources_for_type`) is down from 16 lines to 5; the worked `LLMResponseWithPrompt` example went, since one `source_material_id -> source_material` pair makes the same point.

apps/api/v2/discovery/options.py
SmittieC: Trimmed in 19e1708f2 — `options_for_team` is down from 9 lines to 5, and the module docstring lost the "or discovery is lying" aside.

apps/api/v2/discovery/contract.py
SmittieC: Trimmed in 19e1708f2 — `_resource_source` is down from 13 lines to 8. The two test names it cited went; they are findable from the constant.

apps/api/v2/discovery/contract.py
SmittieC: Trimmed in 19e1708f2 — the `PARAMETER_OPTION_SOURCES` comment is down from 6 lines to 4. The `tools` aside moved out; `reachable_tools` already says it.

apps/api/v2/discovery/options.py
SmittieC: Fixed in 19e1708f2. "the builder" now reads "the UI builder" everywhere it meant the pipeline builder UI — across `options.py`, `node_types.py`, `contract.py`, `facade.py`, `graph_editor.py`, `examples.py`, `write/base.py` and the tests. Nothing in this change refers to an API builder, so there was no second sense to disambiguate against; the `help_text` strings already said "pipeline builder" and were left as they were, since they are client-facing.

api-schemas/v2.yml
SmittieC: Done in 05db21264 (`apps/api/v2/pipeline_edit/examples.py`) — one POST and one PATCH example per accepted node type, each naming *every* param that type declares, plus a "Minimal" `{"type": "CodeNode"}` body. They are in the schema under `requestBody.content.application/json.examples` for both operations.

`tests/test_schema_examples.py` holds them to the node schemas on the three counts that can go stale: every served type is covered, each example names exactly the params its type declares, and each is a body the endpoint actually accepts (sent for real, with placeholder ids swapped for the team's).

apps/api/v2/pipeline_edit/serializers.py
SmittieC: Won't do — this predates the current shape of the response, and the full pipeline state is what a caller needs after an edit: `pipeline_valid` is the publish signal, and a delete has no node to scope the errors to. A caller wanting less can read `node` alone and ignore the rest.

apps/api/v2/pipeline_edit/tests/test_node_create.py
```diff
@@ -0,0 +1,618 @@
...
+    }
+
+    response = client.post(nodes_url(chatbot), payload, format="json")
+
+    assert response.status_code == 201, response.content
+    body = response.json()
+    assert body["node"]["label"] == "Answer the question"
+    assert body["node"]["params"] == payload["params"]
+    assert body["pipeline_errors"]["node"] == {}
+    assert stored_node_params(chatbot, body["node"]["node_id"]).items() >= payload["params"].items()
```
SmittieC: Fixed in fab40211d — the five resource fixtures are bundled into an `LlmNodeResources` dataclass fixture, so the test takes `(client, chatbot, llm, llm_node_resources)`: 4 arguments.

apps/api/v2/pipeline_edit/tests/test_node_update.py
```diff
@@ -0,0 +1,614 @@
...
+    }
+
+    response = client.patch(node_url(chatbot, node_id), payload, format="json")
+
+    assert response.status_code == 200, response.content
+    body = response.json()
+    assert body["node"]["label"] == "Answer the question"
+    assert body["node"]["params"] == payload["params"]
+    assert body["pipeline_errors"]["node"] == {}
+    assert stored_node_params(chatbot, node_id).items() >= payload["params"].items()
```
SmittieC: Fixed in fab40211d — same `llm_node_resources` fixture, so this test is down to 4 arguments too.

apps/api/v2/pipeline_edit/views.py
```diff
@@ -0,0 +1,241 @@
...
+        label = body.validated_data.get("label")
+        return Response(
+            edit_pipeline(
+                request,
+                id,
+                # The node's type comes from the graph, so its params can only be checked under the
+                # lock -- the team is handed in so the check can look its references up there.
+                lambda flow: plan_update(flow, request.team, node_id, label, params),
+                self._write_response,
+            )
```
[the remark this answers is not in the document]
SmittieC: False positive, no code change. The only `str(exception)` reaching a response body on this path is `Pipeline._node_validation_errors`'s `PipelineNodeBuildError` branch and `validate`'s `PipelineBuildError` one — both curated, human-readable validation messages raised deliberately by pipeline node validators for the caller to read, not stack traces. Both already reach clients through `GET /chatbots/{id}/inspect/` on main; this PR only adds a second sink for the same report.

The one branch that *could* have leaked internals is the new catch-all in `_node_validation_errors`, and it deliberately does not: fb36a030d logs the exception and reports a fixed `"This node could not be read. Check the values of its params."` instead, for exactly the reason CodeQL is flagging.

Happy to dismiss the alert as "used in tests"/"won't fix" if you'd rather it not sit on the branch.

apps/api/v2/pipeline_edit/tests/test_node_create.py
```diff
@@ -0,0 +1,607 @@
...
+    assert "server" in str(response.json()["node_id"]).lower()
+    assert not Node.objects.filter(pipeline=chatbot.pipeline, flow_id="mine-1").exists()
+
+
+@pytest.mark.django_db()
+def test_an_unrecognised_body_key_is_refused(client, chatbot):
+    response = client.post(nodes_url(chatbot), {"type": "LLMResponseWithPrompt", "colour": "red"}, format="json")
+
+    assert response.status_code == 400, response.content
+    assert "colour" in response.json()
```
SmittieC: this is too vague

SmittieC:
Superseded by https://github.com/dimagi/open-chat-studio/pull/4289

PR #4249, opened 2026-08-20
Feat: Redesigns Evaluations Results UI

brian-mecha:
### Demo

Screenshot 2026-08-24 at 3 06 23 PM

Screenshot 2026-08-28 at 2 03 53 PM

Screenshot 2026-08-24 at 3 13 27 PM

Screenshot 2026-08-24 at 3 13 32 PM

Screenshot 2026-08-26 at 4 12 50 AM

Resolves #4244

PR #4271, opened 2026-08-24
Seed evaluation runs for local dev

snopoke:
### Product Description
No change — local dev tooling only.

### Technical Description
`bootstrap_data` creates an evaluation config but never runs it, so the runs list, run detail page and trend charts render empty locally. New `bootstrap_evaluation_runs` command seeds 12 runs chosen to cover the states the results UI has to cope with: a rising sequence of completed runs for the trend charts, preview and delta runs, one with bot generation, one with per-evaluator errors and mixed-confidence cost, one with no usage recorded (empty cost card), a failed run, and PROCESSING/PENDING for the progress bar. Also tops the dataset up to 14 messages so the results table paginates, and adds a binary LLM evaluator and a Python one so the aggregates block renders all three stat shapes.

It's a separate command (called from `bootstrap_data`) so the runs can be reseeded on their own while iterating on the results UI. Everything is derived from the spec's `quality` and the message index, so a reseed produces the same numbers — reviewing a UI change doesn't mean re-reading a new dataset.

Two things worth a look:
- Reseed order: tags are un-applied via `remove_applied_tags_for_runs` before the runs are deleted, matching `ClearEvaluationRuns`. The `AppliedTag` audit is what records which eval-applied tags to remove, and deleting the runs cascades it away — reversing the order orphans `CustomTaggedItem` rows on chats and messages.
- The PENDING/PROCESSING fixtures are non-terminal, which leaks past the page they illustrate: `raise_if_runs_in_flight` makes "Clear evaluation runs" 409 and blocks deleting dataset messages/evaluators, and under `inv celery --beat` the coordinator will claim them and dispatch real evaluator calls. Inherent to seeding those states, so the command prints a warning rather than pretending otherwise. The PREVIEW run is deliberately under a day old for the same class of reason — `cleanup_old_preview_evaluation_runs` deletes older ones.

### Demo
`uv run python manage.py bootstrap_evaluation_runs` (or a full `bootstrap_data`), then open the runs list link it prints.

Verified all 12 run detail pages, both result tables, the runs list and the trends view render with aggregates, tags and cost populated, and that the command reruns idempotently.

### Docs and Changelog

### DIFF
--- a/apps/web/management/commands/bootstrap_evaluation_runs.py
+++ b/apps/web/management/commands/bootstrap_evaluation_runs.py
@@ -1,0 +1,717 @@
+"""Seed evaluation runs, results, aggregates and cost rows for local development.
+
+`bootstrap_data` creates an evaluation config but never runs it, so the runs list, the
+run detail page and the trend charts all render empty. This command fills that gap with
+a spread of runs the UI has to cope with: several completed runs over the past few weeks
+(for the trends), a preview and a delta run, a failed one, and two still in flight, plus
+the `source=EVALUATION` UsageRecords the cost cards read (ADR-0048, ADR-0050).
+
+Everything is derived deterministically from the run's `quality` and the message index,
+so a reseed produces the same numbers and reviewing UI changes doesn't mean re-reading a
+new dataset each time.
+
+Usage:
+    python manage.py bootstrap_evaluation_runs
+    python manage.py bootstrap_evaluation_runs --team-slug dev-team
+"""
+
+import uuid
+from dataclasses import dataclass
+from datetime import timedelta
+from decimal import Decimal
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+from django.urls import reverse
+from django.utils import timezone
+
+from apps.annotations.models import Tag
+from apps.cost_tracking.models import Confidence, PricingRule, ServiceKind, UsageRecord, UsageSource
+from apps.evaluations.aggregation import compute_aggregates_for_run
+from apps.evaluations.models import (
+    AppliedTag,
+    EvaluationConfig,
+    EvaluationMessage,
+    EvaluationMode,
+    EvaluationResult,
+    EvaluationRun,
+    EvaluationRunStatus,
+    EvaluationRunType,
+    Evaluator,
+    EvaluatorTagRule,
+)
+from apps.evaluations.rule_validation import ConditionType
+from apps.evaluations.tagging import remove_applied_tags_for_runs
+from apps.experiments.models import Experiment
+from apps.service_providers.models import LlmProvider
+from apps.service_providers.utils import get_first_llm_provider_model
+from apps.teams.models import Team
+
+# Topped up to this many so the results table paginates (10 rows/page).
+_DATASET_SIZE = 14
+_EXTRA_DATASET_MESSAGES = [
+    ("Where is my refund?", "It was issued on Tuesday — it should land within 3 working days."),
+    ("Can I change my delivery address?", "Yes, I've updated it to the new address you gave me."),
+    ("Your app keeps crashing on login.", "Sorry about that — updating to 4.2.1 fixes the login crash."),
+    ("Do you ship to Kenya?", "We do, and delivery to Nairobi is usually 5-7 days."),
+    ("I was charged twice for one order.", "I can see the duplicate charge and have reversed it."),
+    ("This is the third time I'm asking!", "I'm sorry you've had to chase this — I'm fixing it now."),
+    ("Great support, thank you.", "Happy to help — shout if anything else comes up."),
+    ("How do I cancel my subscription?", "Settings → Billing → Cancel, and it stops at the period end."),
+    ("The size chart is wrong.", "Thanks for flagging it — I've passed it to the catalogue team."),
+    ("Is there a student discount?", "There is: 20% off with a valid student email."),
+    ("My parcel arrived empty.", "That shouldn't happen — a replacement is on its way today."),
+]
+_TOPICS = ["billing", "delivery", "product", "account"]
+# Free-text values for `string` output fields, so the results table's text columns read
+# like real judge output instead of the same sentence repeated down the page.
+_REASONING = [
+    "Answered the question directly and matched the expected response.",
+    "Correct, but omitted the timeframe the expected answer gives.",
+    "Acknowledged the complaint without resolving it.",
+    "Restates the question instead of answering it.",
+    "Accurate and appropriately apologetic for the failure.",
+    "Answers a different question than the one asked.",
+    "Right outcome, but the tone is too curt for a complaint.",
+]
+# Best-first, matching the seeded sentiment evaluator's own choice order.
+_SENTIMENTS = ["positive", "neutral", "negative"]
+
+_JUDGE_MODEL = ("openai", "gpt-4o-mini", Decimal("0.00000015"), Decimal("0.0000006"))
+_SECOND_JUDGE_MODEL = ("anthropic", "claude-sonnet-4-5", Decimal("0.000003"), Decimal("0.000015"))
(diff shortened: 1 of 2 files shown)


apps/web/management/commands/bootstrap_evaluation_runs.py
```diff
@@ -0,0 +1,717 @@
...
+                    # "unpriced" if any of its rows lacks a rule, so hanging this off a
+                    # priced judge would hide that judge's real cost too.
+                    evaluator_id=judges[-1].id,
+                    message_id=None,
+                    extra_keys={"missing_usage_calls": 2},
+                )
+            ]
+        UsageRecord.objects.bulk_create(records)
+        # `timestamp` is auto_now_add, so the rows can only be backdated after the insert.
+        UsageRecord.objects.filter(pk__in=[record.pk for record in records]).update(timestamp=run.created_at)
```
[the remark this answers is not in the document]
snopoke: Only PostgreSQL is supported

snopoke: Fixed in 2d53ff21e. The cost-profile branching moved out into `_judge_billing`, which returns the (model, confidence) pair from single-term guards, and the trailing unknown-usage row into `_unreported_usage_records`.

File is now Code Health 10.0 with no findings.

*— Claude Code*

apps/web/management/commands/bootstrap_evaluation_runs.py
```diff
@@ -0,0 +1,717 @@
...
+                position = int((1 - min(max(quality + jitter * 0.7, 0), 1)) * len(choices))
+                values[field_name] = choices[min(position, len(choices) - 1)]
+            elif field_type in ("int", "float"):
+                low = field_def.get("ge", 1)
+                high = field_def.get("le", 10)
+                scaled = low + (high - low) * min(max(quality + jitter * 0.15, 0), 1)
+                values[field_name] = int(round(scaled)) if field_type == "int" else round(scaled, 2)
+            else:
+                values[field_name] = _REASONING[index % len(_REASONING)]
+        return values
```
snopoke: Fixed in 2d53ff21e. The per-field-type dispatch is now three small functions (`_field_value`, `_choice_value`, `_numeric_value`), leaving `_result_values` as the schema lookup plus a comprehension.

*— Claude Code*

apps/web/management/commands/bootstrap_evaluation_runs.py
```diff
@@ -0,0 +1,717 @@
...
+                    experiment,
+                    model=_GENERATION_MODEL,
+                    input_tokens=310 + 9 * (message_index % 5),
+                    output_tokens=180 + 7 * (message_index % 5),
+                    evaluator_id=None,
+                    message_id=message.id,
+                    confidence=Confidence.EXACT,
+                    priced=True,
+                )
+        if spec.cost_profile == "mixed" and scored and judges:
```
snopoke: Fixed in 2d53ff21e. The `cost_profile == "mixed" and index == ...` conditionals are gone — `_judge_billing` decides the same thing with one-term guard clauses.

*— Claude Code*

apps/web/management/commands/bootstrap_evaluation_runs.py
```diff
@@ -0,0 +1,717 @@
+"""Seed evaluation runs, results, aggregates and cost rows for local development.
```
snopoke, line 1: Fixed in 2d53ff21e. Extracting the pricing decision and the per-field-type helpers brought the mean back under the threshold; the file now scores 10.0.

*— Claude Code*

apps/web/management/commands/bootstrap_evaluation_runs.py
```diff
@@ -0,0 +1,717 @@
...
+        EvaluationRun.objects.filter(pk=run.pk).update(**fields)
+        # `created_at` is read back for the usage timestamps and by the callers' logging.
+        run.refresh_from_db()
+
+        if spec.status == EvaluationRunStatus.COMPLETED:
+            compute_aggregates_for_run(run)
+        self._apply_tags(team, run, results)
+        if spec.cost_profile != "none":
+            self._create_usage_records(team, run, spec, evaluators, scored, experiment)
+        return run
```
snopoke: Fixed in 2d53ff21e. The four run-invariant arguments (team, config, user, generation experiment) are now a `_SeedContext`, so `_create_run` takes 4.

*— Claude Code*

apps/web/management/commands/bootstrap_evaluation_runs.py
```diff
@@ -0,0 +1,717 @@
...
+                    self._ensure_pricing_rule(provider_type, model_name, service_kind, unit_price) if priced else None
+                ),
+                evaluator_id=evaluator_id,
+                message_id=message_id,
+            )
+            for service_kind, quantity, unit_price in (
+                (ServiceKind.LLM_INPUT, input_tokens, input_price),
+                (ServiceKind.LLM_OUTPUT, output_tokens, output_price),
+            )
+        ]
```
snopoke: Fixed in 2d53ff21e. Down to 2: a `_RunContext` (team/run/experiment) and a `_ModelCall` describing what was billed for how many tokens and on whose behalf.

*— Claude Code*

apps/web/management/commands/bootstrap_evaluation_runs.py
```diff
@@ -0,0 +1,717 @@
...
+            model_name=model_name,
+            quantity=quantity,
+            unit_price=unit_price,
+            cost=cost,
+            confidence=confidence,
+            experiment=experiment,
+            evaluation_config=run.config,
+            pricing_rule=pricing_rule,
+            extra=extra,
+        )
```
snopoke: Fixed in 2d53ff21e. Down to 2 — `_RunContext` plus a `_UsageRow` dataclass holding the row's own fields.

*— Claude Code*

apps/web/management/commands/bootstrap_evaluation_runs.py
```diff
@@ -0,0 +1,717 @@
...
+    "Accurate and appropriately apologetic for the failure.",
+    "Answers a different question than the one asked.",
+    "Right outcome, but the tone is too curt for a complaint.",
+]
+# Best-first, matching the seeded sentiment evaluator's own choice order.
+_SENTIMENTS = ["positive", "neutral", "negative"]
+
+_JUDGE_MODEL = ("openai", "gpt-4o-mini", Decimal("0.00000015"), Decimal("0.0000006"))
+_SECOND_JUDGE_MODEL = ("anthropic", "claude-sonnet-4-5", Decimal("0.000003"), Decimal("0.000015"))
+_GENERATION_MODEL = ("openai", "gpt-4o", Decimal("0.0000025"), Decimal("0.00001"))
```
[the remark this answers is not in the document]
snopoke: Good catch — fixed in 0d8b90a7d, and both consequences were real.

Confirmed the diagnosis against the code: `recorder.py:87` computes `quantity / 1000 * unit_price`, and the seed JSON has the rates you quoted. Took the suggestion along with the `/1000` you flagged as required, extracted as a `_token_cost` helper that mirrors the recorder including its `quantize`.

On point 1 — my local dev DB already had exactly the two orphaned rules you predicted, `anthropic/claude-sonnet-4-5` at 0.000003/0.000015, i.e. 1/1000 of the real rate. They were `PROTECT`ed by the usage rows referencing them, so they only became deletable after a reseed. Cleaned up locally; the command now creates no `PricingRule` rows at all.

Verified on real data after reseeding: 473 usage rows, 0 with `unit_price != pricing_rule.unit_price` (point 2), 0 breaking `cost == quantity/1000 * unit_price`, and total cost unchanged at 0.4743 — the two errors did cancel, exactly as you said.

*— Claude Code*

brian-mecha (approved):
Tested against my current UI refactor on #4249 . Works like a dream.

Image

PR #4273, opened 2026-08-25
Chat widget: present an OAuth token at chat/start/

snopoke:
Server side of [#3893](https://github.com/dimagi/open-chat-studio/issues/3893) already landed — `ChatOAuthAuthentication`, the `chat:start` scope, `credential_mode` on the channel. This is the widget half: a chatbot whose Chat Widget & API channel is in `oauth` mode needs the caller to present a bearer token, and until now the widget had no way to send one.

Design context: [`docs/design/oauth-chat-widget.md`](https://github.com/dimagi/open-chat-studio/blob/main/docs/design/oauth-chat-widget.md) (row 15 of the implementation outline).

### Product Description

No user-visible change yet. `credential_mode` is deliberately not on the channel form, so `oauth` mode is not admin-selectable and no existing embed is affected. This unblocks that form landing.

### Technical Description

**The credential is pulled, not held.** The single new prop is `authTokenProvider` — a function the host installs on the element, which the widget calls at every session start. There is no stored token, so the widget can never carry a stale one, and ADR-0054's session expiry recovers on its own: the restart calls `chat/start/` again, which asks the host for a fresh token at exactly that moment. That is what makes admission recurring rather than one-shot without the host needing a timer or a signal back from the widget.

Worth knowing for review: **a static `auth-token` attribute was implemented first and then removed.** It read as the cheaper integration — a backend template renders the token into the snippet, no JS — and with `ACCESS_TOKEN_EXPIRE_SECONDS` at its 10-hour default it works for any visit that starts a chat the same day. It came out because anyone reaching `oauth` mode already runs a token-minting backend (client credentials require a secret the page cannot hold), because its failure mode passes integration testing and fails in production (past the token's life the visitor gets an opaque 401 with no recovery short of a reload), and because keeping both meant a precedence rule over a fallback branch almost nothing reached. The cost is real and worth naming: the provider is a function, so it is a JS property with no attribute equivalent — a pure-snippet embed cannot use `oauth` mode, and the channel dialog's copy-paste snippet will have to show a small JS block. Rationale is recorded in the design doc rather than only here.

Two other decisions the diff won't explain:

- **`Authorization` goes on `chat/start/` only.** The design doc originally said `getCommonHeaders()`, which would have put the host's credential on `message/`, `poll/` and `upload/` — three endpoints that ignore it. Corrected in this PR.
- **A refused token is retried exactly once, and only if the refresh returned a different one.** `chat/start/` is throttled per ADR-0052, so resending an identical request would spend two buckets per failure. The retry guard is on "is there a provider", not "did we send a token", so a provider that wasn't ready on the first send can still recover.

`ChatAuthError` (401) is deliberately separate from `SessionAccessError` (403): the latter's handler discards the session and, on a bound widget, reports it unavailable — wrong for a refusal where no session was ever created.

### Migrations

None.

### Demo

Not reachable from the UI yet — `oauth` mode has no form control, so exercising this end to end needs a channel switched in the shell. Covered by tests instead: 24 cases across `chat-session-service.spec.ts` and `ocs-chat_auth_provider.spec.tsx`, the component ones driving the real service through a mocked `fetch` so they assert on actual request headers. They cover header sent/omitted/rotated/never-persisted, the retry and its cap, a provider that throws, 401 vs 403, and the ADR-0054 restart pulling a fresh token.

### Docs and Changelog

Widget changelog entry needed for the new `authTokenProvider` prop, including that it is a JS property with no attribute form and that the host must mint tokens in its backend. **This is not usable until the widget is published** — needs the npm release, a `LATEST_VERSION` bump and `MIN_OAUTH_WIDGET_VERSION` (agreed as 0.12.0), which land together per `docs/developer_guides/widget_versioning.md`. Draft until then.

Left out of scope: `credential_mode` on `EmbeddedWidgetChannelForm`, and the bound-widget restart affordance (ADR-0054's own follow-up).


components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
```diff
@@ -107,6 +115,16 @@ export class OcsChat {
...
   @Prop() embedKey?: string;
 
+  /**
+   * Supplies the OAuth bearer token for a chatbot whose Chat Widget & API channel
+   * requires one. Mint it in the host application's backend and return it here;
+   * the widget asks once per session start, so it never holds a credential that can
+   * go stale, and never writes one to local storage. `forceRefresh` is set when the
+   * previous token was rejected, so a caching host knows to bypass its cache. Being
+   * a function, this is a JavaScript property with no HTML attribute equivalent.
+   */
```
snopoke, line 124:
```suggestion
  /**
   * Supplies the OAuth bearer token for a chatbot whose Chat Widget & API channel
   * requires one. Mint it in the host application's backend and return it here;
   * the widget asks once per session start and never writes one to local storage.
   * `forceRefresh` is set when the previous token was rejected. Being
   * a function, this is a JavaScript property with no HTML attribute equivalent.
   */
```

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
```diff
@@ -889,6 +917,15 @@ export class OcsChat {
     await this.clearSession();
   }
 
+  /**
+   * The service is built once and cached, so a provider installed later has to be
+   * pushed into the live instance -- rebuilding it would drop the session token.
+   */
```
snopoke:
```suggestion
```

components/chat_widget/src/services/chat-session-service.ts
```diff
@@ -12,6 +12,35 @@ export class SessionAccessError extends Error {
...
+    this.name = 'ChatAuthError';
+    this.status = status;
+    this.code = code;
+  }
+}
+
+/**
+ * Supplies the OAuth bearer token for `chat/start/`. Called once per session
+ * start, and a second time with `forceRefresh` when the server rejected the
+ * first token — a host that caches tokens should bypass its cache on that call.
```
snopoke:
```suggestion
 * Supplies the OAuth bearer token for `chat/start/`. Called once per session
 * start, and a second time with `forceRefresh` when the server rejected the
 * first token.
```

components/chat_widget/src/services/chat-session-service.ts
```diff
@@ -105,18 +136,27 @@ export class ChatSessionService {
...
-      headers: this.getJsonHeaders(),
-      body: JSON.stringify(requestBody),
-    });
+    const token = await this.resolveAuthToken(false);
+    let response = await this.startSessionRequest(requestBody, token);
+
+    // The provider may have handed back a cached token that has since expired, so
+    // ask again with `forceRefresh` before giving up. One retry only, and only for
+    // a token that actually changed -- a provider that returns the same token (or
+    // nothing) would spend the retry on a request already refused.
```
snopoke:
```suggestion
    // The provider may have handed back a cached token that has since expired, so
    // ask again with `forceRefresh` before giving up. One retry only, and only for
    // a token that actually changed.
```

components/chat_widget/src/services/chat-session-service.ts
```diff
@@ -383,6 +434,42 @@ export class ChatSessionService {
...
 
+  setAuthTokenProvider(provider?: AuthTokenProvider): void {
+    this.authTokenProvider = provider;
+  }
+
+  /**
+   * Headers for `chat/start/`. The bearer token goes here and nowhere else: it
+   * admits a *new* session, and every request after that is authorised by the
+   * session token instead. Sending it more widely would put the host's
+   * credential on three endpoints that ignore it.
```
snopoke:
```suggestion
   * Headers for `chat/start/`. The bearer token goes here and nowhere else: it
   * admits a *new* session, and every request after that is authorised by the
   * session token instead.
```

components/chat_widget/src/services/chat-session-service.ts
```diff
@@ -383,6 +434,42 @@ export class ChatSessionService {
...
+      headers['Authorization'] = `Bearer ${authToken}`;
+    }
+    return headers;
+  }
+
+  /**
+   * Asks the host for a bearer token at the moment one is needed. There is no
+   * stored token to fall back on: a credential the widget holds onto goes stale
+   * without anything noticing, and the host is the only party that can mint a
+   * fresh one. A channel in `embed_key` mode has no provider and needs none.
```
snopoke:
```suggestion
   * Asks the host for a bearer token at the moment one is needed.
```

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
snopoke: Fixed in d0e18e05b. Applied as suggested.

*— Claude Code*

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
snopoke: Fixed in d0e18e05b. Comment removed.

*— Claude Code*

components/chat_widget/src/services/chat-session-service.ts
snopoke: Fixed in 45c16a706. Applied as suggested.

*— Claude Code*

components/chat_widget/src/services/chat-session-service.ts
snopoke: Fixed in 45c16a706. Applied as suggested.

*— Claude Code*

components/chat_widget/src/services/chat-session-service.ts
snopoke: Fixed in 45c16a706. Applied as suggested.

*— Claude Code*

components/chat_widget/src/services/chat-session-service.ts
snopoke: Fixed in 45c16a706. Applied as suggested.

*— Claude Code*

components/chat_widget/src/services/chat-session-service.ts
```diff
@@ -383,6 +434,42 @@ export class ChatSessionService {
...
+   */
+  private async resolveAuthToken(forceRefresh: boolean): Promise<string | undefined> {
+    if (!this.authTokenProvider) {
+      return undefined;
+    }
+    try {
+      return (await this.authTokenProvider({ forceRefresh })) || undefined;
+    } catch (error) {
+      const detail = error instanceof Error ? error.message : String(error);
+      throw new ChatAuthError(401, 'auth_token_unavailable', `Could not obtain an authentication token: ${detail}`);
```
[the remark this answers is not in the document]
snopoke: Fixed in 28fb41da9. Confirmed the path is real: the message reaches `addErrorMessage`, and `saveSessionToStorage` writes `keys.messages` unconditionally, so host exception text does land in localStorage.

Went slightly further than the suggestion — the detail now goes to `console.error` rather than being dropped, so a host debugging its token endpoint can still see why the provider failed without the text being persisted or shown. The test asserts both halves.

*— Claude Code*

docs/design/oauth-chat-widget.md
```diff
@@ -794,11 +798,38 @@ The page token carries `chat:start` and nothing else ([D4](#d4-what-makes-a-toke
...
+token it returns is sent as `Authorization: Bearer` **on `chat/start/` only** (an earlier draft said
+`getCommonHeaders()`, which would have put the host's credential on `message/`, `poll/` and `upload/`
+— three endpoints that ignore it and that [D5](#d5-session-bound-endpoints-are-untouched) keeps out of
+the admission decision). Nothing is written to `localStorage`: the credential belongs to the page, not
+the conversation. `embed-key` stays optional, so an `oauth`-mode embed may set only `chatbot-id`.
+
+**Rotation is a pull, not a push, and that is what makes D7 work.**
+[D7](#d7-admission-is-bounded-in-time) says the host "must push a fresh token through the `@Watch`",
+which leaves the host to work out *that* it needs to. Pulling inverts the problem: the widget asks at
+the moment it needs a credential — including at the ADR-0054 restart — so recurring admission needs no
```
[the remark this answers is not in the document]
snopoke: Fixed in 65ec26a00 — good catch, though for a different reason than stated. D7 itself was already rewritten to describe the pull; the problem was this paragraph still *attributing* the "must push" wording to D7, which no longer says it. That framing came from an earlier draft, so it is now credited as such.

*— Claude Code*

components/chat_widget/src/components/ocs-chat/readme.md
```diff
@@ -9,38 +9,39 @@ For more information, see the [Open Chat Studio documentation](https://docs.open
...
+| `authTokenProvider`          | --                              | Supplies the OAuth bearer token for a chatbot whose Chat Widget & API channel requires one. Mint it in the host application's backend and return it here; the widget asks once per session start, so it never holds a credential that can go stale, and never writes one to local storage. `forceRefresh` is set when the previous token was rejected, so a caching host knows to bypass its cache. Being a function, this is a JavaScript property with no HTML attribute equivalent. | `(context: { forceRefresh: boolean; }) => string \| Promise<string>` | `undefined`                        |
```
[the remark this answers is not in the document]
snopoke: Not applicable — declined by author. The table is generated by `stencil build --docs` from the JSDoc, so the only way to add the `undefined` case is to re-lengthen the prop comment that was just trimmed in d0e18e05b. The exported `AuthTokenProvider` type states it, and a provider returning nothing is covered by tests (`omits the header when the provider yields nothing`, `retries an unauthenticated start once the provider can mint a token`).

*— Claude Code*

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
snopoke: Fixed in d0e18e05b. Applied as suggested.

*— Claude Code*

components/chat_widget/src/components/ocs-chat/ocs-chat.tsx
snopoke: Fixed in d0e18e05b. Comment removed.

*— Claude Code*

components/chat_widget/src/services/chat-session-service.ts
snopoke: Fixed in 45c16a706. Applied as suggested.

*— Claude Code*

components/chat_widget/src/services/chat-session-service.ts
snopoke: Fixed in 45c16a706. Applied as suggested.

*— Claude Code*

components/chat_widget/src/services/chat-session-service.ts
snopoke: Fixed in 45c16a706. Applied as suggested.

*— Claude Code*

components/chat_widget/src/services/chat-session-service.ts
snopoke: Fixed in 45c16a706. Applied as suggested.

*— Claude Code*

components/chat_widget/src/services/chat-session-service.ts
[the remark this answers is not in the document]
snopoke: Fixed in 28fb41da9. Confirmed the path is real: the message reaches `addErrorMessage`, and `saveSessionToStorage` writes `keys.messages` unconditionally, so host exception text does land in localStorage.

Went slightly further than the suggestion — the detail now goes to `console.error` rather than being dropped, so a host debugging its token endpoint can still see why the provider failed without that text being persisted or shown. The test asserts both halves.

*— Claude Code*

docs/design/oauth-chat-widget.md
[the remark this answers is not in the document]
snopoke: Fixed in 65ec26a00 — real inconsistency, though for a different reason than stated. D7 itself was already rewritten to describe the pull; the problem was that this paragraph still *attributed* the "must push" wording to D7, which no longer says it. That framing came from an earlier draft, so it is now credited as such rather than to D7.

*— Claude Code*

components/chat_widget/src/components/ocs-chat/readme.md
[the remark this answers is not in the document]
snopoke: Not applicable — declined by author. That table is generated by `stencil build --docs` from the JSDoc, so the only way to add the `undefined` case is to re-lengthen the prop comment that was just trimmed in d0e18e05b.

The exported `AuthTokenProvider` type already states it, and the behaviour is covered by tests: `omits the header when the provider yields nothing` and `retries an unauthenticated start once the provider can mint a token`.

*— Claude Code*

SmittieC (approved):

Issue #3682, opened 2026-06-23
Implement new 'public channel'


snopoke:
Similar to embedded channel but does not expose widget key or allowed hosts. Also allows the user to specify starter questions and welcome messages.

snopoke:
Closing this after reaching a decision to remove the public chat UI entirely.

snopoke:
Decision reversed

barry47products:
@snopoke, picking this up from where I left off previously. Here is the shape I think it takes, and the calls I would want before writing a full design.

## Two surfaces, not one

Worth separating:

1. The legacy public chat UI (`/e/<id>/start/`). What projects use today, and what #4209 removes.
2. Anonymous session start on the chat API. ADR-0053 left this open on purpose: a request with no credentials can start a session on any chatbot whose `public_id` it knows. `widget_versions.py` deprecates everything below 0.6.0 with a 1 October 2026 sunset; after that, no supported widget lacks an embed key, and closing the anonymous path becomes its own change.

The public channel moves group 1 onto mechanism 2 without requiring mechanism 2 to stay open for every chatbot in the system. Anonymous start becomes something a team switches on for one bot, instead of something tolerated everywhere until the sunset.

## Shape

A per-chatbot `PUBLIC` platform. No embed key, no allowed-domains list, and a declared `required_auth_level` so the ADR-0045 ratchet does not sweep it up in October.

That is three touches in `chat_start_session`: `WidgetDomainPermission` passes without a domain check; `_resolve_embed_key_channel` gains a sibling path so a chatbot with a public channel resolves to it with no key present, which is what stops public traffic falling through to `get_team_api_channel` and losing its attribution; version selection stays shut to anonymous callers.

Then two surfaces over the one channel: an OCS-hosted page, which is the shareable link the legacy UI provided, and a keyless snippet for projects hosting it themselves. Welcome message and starter questions as channel config.

## Calls I would like

**1. What goes in the public URL.** `public_id` costs no new state, but it is the chatbot's identity everywhere else, so a link cannot be revoked without disabling the channel. A rotatable token on the channel costs a field and gives you a regenerate action. Either way, it is an address rather than a secret, since it ships in the page. I would take the token. Public URLs end up in link previews and search results.

**2. Which version anonymous visitors get.** They cannot select one, so the channel decides. The working version means unpublished edits are live to the public as you make them. Published needs defined behaviour for a bot with nothing published yet. I would serve published, and have the channel refuse until a version exists.

[...]

PR #4279, opened 2026-08-26
Widget release A: tab persistence, kiosk restart, UUID visitor id, timezone on start

### LINKED ISSUE #4292: Chat widget: wrap stale-session callbacks in a `forCurrentSession` helper
#4279 added a `sessionEpoch` counter to `ocs-chat.tsx` so that callbacks from a previous session (poll results, history loads, in-flight sends) are dropped once a kiosk restart or a new session has started. The guard is a repeated inline check:

```ts
const epoch = this.sessionEpoch;
...
if (epoch !== this.sessionEpoch) return;
```

It now appears at twelve sites across `startSession`, `sendMessage`, `loadBoundSessionHistory` and both pollers (`startMessagePolling` and the task poller). CodeScene flagged `startMessagePolling` as crossing its complexity threshold because of the added guards.

### Proposed change

Add a small private helper:

```ts
private forCurrentSession<T extends unknown[]>(fn: (...args: T) => void): (...args: T) => void {
  const epoch = this.sessionEpoch;
  return (...args) => {
    if (epoch !== this.sessionEpoch) return;
    fn(...args);
  };
}
```

and wrap the six callback sites in the two pollers with it. The six inline guards inside the `async` methods stay as they are; the guard there sits between `await`s, and a wrapper does not express that as clearly.

### Acceptance

- The kiosk restart and session-handling specs (`ocs-chat_kiosk_restart.spec.tsx`, `ocs-chat_session_handling.spec.tsx`) pass unchanged.
- `startMessagePolling` and the task poller no longer carry inline epoch checks.
- No behaviour change: a poll callback that arrives after a restart is still dropped.

Related: #4279, #3682.

barry47products:
### Product Description

First widget slice for the public channel (#3682, design D5 / Phasing step 1). Nothing changes for existing embeds unless they opt in.

- `persistent-session="tab"`: conversation survives a reload of the same tab (sessionStorage); closing the tab forgets it.
- Kiosk mode shows a "Start new chat" button once the session has ended, so a shared device can move to the next person without a reload.
- Visitor ids are `ocs:<uuid v4>` via `crypto.randomUUID()`. Stored ids keep working.
- The widget sends the browser timezone on session start; the backend stores it on the participant.

### Technical Description

- One `getStorage()` accessor picks localStorage, sessionStorage or nothing from the `persistent-session` value; all reads and writes go through it. The prop type is now `boolean | 'tab' | 'true' | 'false'`, and Stencil stops coercing attributes once the type is a union, so the string `"false"` is handled explicitly.
- Clearing a session only touches the store the widget is using, so a kiosk page cannot wipe a session a standard widget holds on another same-origin page. Orphaned records are reaped by `persistent-session-expire`.
- Poll callbacks from a cleared session are dropped (epoch guard), so a restart cannot resurrect the previous transcript. The restart button reuses `clearSession()` and is hidden for bound sessions and read-only widgets.
- `POST /api/chat/start/` accepts an optional `timezone`, validated against `pytz.all_timezones_set` (what `pretty_date` resolves with). Unknown values are ignored, not rejected: optional telemetry should not fail an anonymous start. Stored on the started chatbot's `ParticipantData` only. Schema regenerated.
- Widget 0.12.0. `LATEST_VERSION` stays at 0.11.0 until the npm publish, as the 0.11.0 release did; a follow-up PR bumps it with the root dependency and the notification migration.

Publish blocker, pre-existing: since the pnpm migration (464d0248c) the widget build runs on Stencil 4.43.5, where `stencil-tailwind-plugin` silently does nothing and the bundle ships raw `@apply`. 0.11.0 predates that commit; `w_v0.12.0` would be the first broken publish. Needs its own fix before tagging. Screenshots below were built with Stencil pinned to 4.36.3.

CodeScene: `ocs-chat.tsx` improves overall (5.73 -> 5.80). The remaining deltas, `render` 47 -> 48 and `chat_start_session` 13 -> 14, are one conditional render for the kiosk button and the `timezone` write beside the existing `name` write.

### Demo

Two kiosk widgets for the same chatbot, differing only in `persistent-session`; the panel under each reads its storage once a second.

| Step | Screenshot |
|---|---|
| 1. First message in A: session id under sessionStorage, B has none. | step 1 |
| 2. Same tab reloaded: same conversation and id. | step 2 |
| 3. New tab: A fresh, B resumes from localStorage. | step 3 |
| 4. Session ended in OCS: "Start new chat" above the disabled composer. | step 4 |
| 5. After restart: new session id, B untouched. | step 5 |

### Docs and Changelog

Widget changelog for 0.12.0 comes from the docs automation on merge.

Closes #4292.


barry47products:
@snopoke or @SmittieC, a few things I'd like your read on, none of which block review:

1. Unknown `timezone` on `/api/chat/start/`: currently ignored so optional telemetry cannot fail an anonymous start. Would you rather it return a 400?
2. `pytz` isn't a direct dependency; it comes in via `django-tz-detect`, and the lock has it at 2023.3, so the validator's `all_timezones_set` lags `zoneinfo` (currently by one real zone, `America/Coyhaique`). Django and the newer apps here already use `zoneinfo`; the remaining `pytz` users are `utils/time.py`, `dynamic_filters/column_filters.py`, `events/models.py` and this validator. I can move those four to `zoneinfo` in a follow-up.
3. `widget_versioning.md` says to bump `LATEST_VERSION` in the same PR as the widget version. 0.11.0 did it in two: #3888 bumped `package.json`, the tag was cut, then #3897 bumped `LATEST_VERSION`. This PR follows the 0.11.0 sequence. Which of the two should change?
4. Two refactors I left out because they reshape code this slice doesn't otherwise touch. Happy to do either as a separate PR:
   - `_record_participant_name` and `_record_participant_timezone` each `get_or_create` the same `ParticipantData` row. One keyword-argument helper (still writing `Participant.name`) would save a query when a request carries both, and take `chat_start_session` to 13 or below.
   - The stale-callback epoch guard now sits at twelve sites across `startSession`, `sendMessage`, `loadBoundSessionHistory` and both pollers. A `forCurrentSession(callback)` wrapper would collapse the six callback sites in the pollers and take both under the threshold; the six inline guards in the async methods would stay.

barry47products:
Heads-up on #4273 (widget OAuth token at chat/start/): it and this PR both touch `ocs-chat.tsx` around `startSession`, plus the generated `readme.md` and `components.d.ts`. The server side doesn't interact (the public channel stays in embed-key mode), but whichever merges second will need a hand-merge of `startSession`. 

I'm happy to merge into this branch once #4273 lands, or the other way round if you would rather this go first; your call on order. If both are in before the 0.12.0 publish, `authTokenProvider` ships in the same widget release as the tab persistence and kiosk restart, so the release notes should cover both.

snopoke:
> @snopoke or @SmittieC, a few things I'd like your read on, none of which block review:
> 
> 1. Unknown `timezone` on `/api/chat/start/`: currently ignored so optional telemetry cannot fail an anonymous start. Would you rather it return a 400?

Ignoring is fine.

> 2. `pytz` isn't a direct dependency; it comes in via `django-tz-detect`, and the lock has it at 2023.3, so the validator's `all_timezones_set` lags `zoneinfo` (currently by one real zone, `America/Coyhaique`). Django and the newer apps here already use `zoneinfo`; the remaining `pytz` users are `utils/time.py`, `dynamic_filters/column_filters.py`, `events/models.py` and this validator. I can move those four to `zoneinfo` in a follow-up.

Seems useful to do as a followup.

> 3. `widget_versioning.md` says to bump `LATEST_VERSION` in the same PR as the widget version. 0.11.0 did it in two: [bump widget version #3888](https://github.com/dimagi/open-chat-studio/pull/3888) bumped `package.json`, the tag was cut, then [chore: upgrade open-chat-studio-widget to 0.11.0 #3897](https://github.com/dimagi/open-chat-studio/pull/3897) bumped `LATEST_VERSION`. This PR follows the 0.11.0 sequence. Which of the two should change?

We usually split the version bump and release into a separate step.

> 4. Two refactors I left out because they reshape code this slice doesn't otherwise touch. Happy to do either as a separate PR:
>    
>    * `_record_participant_name` and `_record_participant_timezone` each `get_or_create` the same `ParticipantData` row. One keyword-argument helper (still writing `Participant.name`) would save a query when a request carries both, and take `chat_start_session` to 13 or below.

This was one I flagged in my review as well.

>    * The stale-callback epoch guard now sits at twelve sites across `startSession`, `sendMessage`, `loadBoundSessionHistory` and both pollers. A `forCurrentSession(callback)` wrapper would collapse the six callback sites in the pollers and take both under the threshold; the six inline guards in the async methods would stay.

Good for a follow up.

barry47products:
@snopoke, all applied.

The two participant helpers are now one `_record_participant_details` that fetches the `ParticipantData` row once. While I was in there, I moved the new validator onto `zoneinfo.available_timezones()` rather than leaving a fourth `pytz` import for the follow-up to undo; the three existing ones are in #4291.

`package.json` is back at 0.11.0, so the bump and publish can go as their own PR, as you described.

I merged `main` after #4273 landed. The only real conflict was the generated `readme.md`, which I regenerated with `stencil build --docs` rather than merging by hand. `startSession` now sends both the OAuth token and the timezone.

On the CodeScene side: I did end up pulling the `forCurrentSession` wrapper in here because it's what actually clears the `startMessagePolling` finding, so this PR can close #4292. The rest of its complaints were about `render`, which was already at 47 before I touched it. The kiosk restart control, the composer input area and the attachment controls now render from their own methods, and the "is the composer locked" rule is written once instead of four times.

The one finding left is `chat_start_session` at 14. My contribution there is the single `if name or participant_timezone`; getting the view under the threshold needs a bigger split, which I wrote up in #4293 for a separate PR.

snopoke (approved):

Issue #4281, opened 2026-08-26
Surface the running OCS version


snopoke:
Self-hosted operators can't tell what version they're running, and neither can we. `pyproject.toml` says `version = "0.1.0"`, which is unused and stale. Recorded as known gap 2 in `RELEASING.md`.

### Where the version comes from

Derive it at build time from the git tag rather than bumping a tracked file. A tracked `__version__` fights the release soak rule: the bump would have to be committed, but a release tag must point at a commit that has already been running in production, and a fresh bump commit hasn't soaked.

```bash
git describe --tags --match 'v*' --always
```

| Build | Output |
|---|---|
| At a release tag | `v1.2.0` |
| Main, 37 commits later | `v1.2.0-37-gabc1234` |
| Before any `v*` tag exists | `f8eb278de` |

`--match 'v*'` is required, not cosmetic. Bare `git describe --tags` in this repo returns `w_v0.11.0-1539-gf8eb278de` — it latches onto the chat widget's tag train. All three rows above were verified against the repo.

Plumb it as `ARG OCS_VERSION` → `ENV` in the final Dockerfile stage, read into settings with a default of `unknown` so source-tarball builds degrade gracefully rather than crashing. It has to arrive as a build arg rather than being computed inside the image, because `.git` should not be in the image (see #4282).

### Surfaces

- [ ] `manage.py version` — what an operator actually reaches for from a container shell
- [ ] `sentry_sdk.init(release=...)` in `config/settings.py:732`
- [ ] `/status/` (`apps/web/views.py:33`) — already gated by `HEALTH_CHECK_TOKENS`, so it's the right home for a machine-readable version
- [ ] Django admin index and/or footer, **staff only**
- [ ] Note in `pyproject.toml` that its `version` field is unused, so it stops being a trap
- [ ] Flip known gap 2 in `RELEASING.md` and the corresponding note in `docs/hosting/releases.md`

Don't put the version in the public footer or any unauthenticated response: it tells an attacker precisely which CVEs apply to a self-hosted instance.

### Note

`deploy.yml:384` already runs `getsentry/action-release`, so Sentry releases exist but events aren't tagged to them — `release=` is absent from `sentry_sdk.init`. That one line is independently useful and can land ahead of the rest.

PR #4283, opened 2026-08-26
Publish versioned container images to GHCR

### LINKED ISSUE #4282: Publish versioned container images to GHCR
Third-party operators currently clone the repo and build their own image. Publishing a versioned image on tag push makes an upgrade a pull instead of a build. Recorded as known gap 1 in `RELEASING.md`.

Blocked on the base image bump below — see "Do this first".

### Do this first: the base image is EOL

`Dockerfile` and `Dockerfile.dev` both use `python:3.13-slim-bullseye`, which is Debian 11.11. Bullseye's LTS window closes around now. Internally that's a manageable risk; publishing it publicly means shipping unpatched OS CVEs to third parties under Dimagi's name.

I tested both upgrade targets against the six packages the Dockerfile installs (`libasound2`, `libpq5`, `ffmpeg`, `libmagic1`, `mailcap`, plus build deps):

- **bookworm — clean drop-in.** All six resolve unchanged. One-line change in each Dockerfile, no package edits.
- **trixie — not a drop-in.** `libasound2` is now `libasound2t64` and `libmagic1` isn't available at all.

Recommend bookworm. This is independently valuable and can merge on its own, ahead of the rest.

### Then

- [ ] Bump `Dockerfile` and `Dockerfile.dev` to `python:3.13-slim-bookworm`
- [ ] Add `.git` to `.dockerignore`. It isn't excluded today and the Dockerfile does `COPY . /code`, so full repo history lands in every image — bloat internally, and history we didn't mean to distribute in a public one
- [ ] Add `.github/workflows/publish_image.yml` on `push: tags: ['v*']`, publishing to `ghcr.io/dimagi/open-chat-studio`
- [ ] Pass the version through as a build arg (see #4281)
- [ ] Parameterize `docker-compose.prod.yml`: `image: ghcr.io/dimagi/open-chat-studio:${OCS_VERSION:-latest}`, keeping `build:` so both pull-and-run and build-from-source keep working
- [ ] Update `docs/hosting/docker.md` and `docs/hosting/releases.md` for the pull path
- [ ] Flip known gap 1 in `RELEASING.md`

### Implementation notes

- Auth is `GITHUB_TOKEN` plus `permissions: packages: write`. No new secrets.
- `publish_widget.yml` already has the tag-triggered shape to mirror (it fires on `w_v*`).
- `deploy.yml:140` already proves the `docker/build-push-action@v7` + `cache-from: type=gha` config for this Dockerfile.
- ECR stays exactly as it is; this is additive and doesn't touch Dimagi's own deploy path.
- **amd64 only initially.** arm64 under QEMU means compiling Python C extensions and running a full node asset build under emulation; it wants a native arm runner. Worth a follow-up issue rather than blocking this.
- Tag with `docker/metadata-action`: `1.2.3`, `1.2`, and `latest`.

### Scope note

Publishing a public image is a commitment, not just a workflow: base-image CVEs become ours to patch on a schedule, and `latest` becomes something operators will pin to whether we want them to or not.

snopoke:
Closes #4282 (partially — see "Left out" below).

### Product Description
No change for product users. Self-hosted operators gain a versioned image to pull instead of building their own.

### Technical Description
Three separable commits, deliberately ordered so the base-image bump can be reviewed and merged on its own if you'd rather not take the rest yet.

**Base image: bullseye → bookworm.** This is the one that needed care. `python:3.13-slim-bullseye` is Debian 11.11 and bullseye's LTS window has closed; publishing a public image on it means shipping unpatched OS CVEs to third parties under Dimagi's name. bookworm is a genuine drop-in — I checked all six installed packages resolve unchanged before touching anything. trixie is *not*: `libasound2` becomes `libasound2t64` and `libmagic1` isn't available at all, so bookworm is the right target today.

**GHCR publish on `v*` tags.** Mirrors `publish_widget.yml`'s tag-triggered shape and reuses the `docker/build-push-action@v7` + GHA-cache config already proven in `deploy.yml:140`. Auth is `GITHUB_TOKEN` + `packages: write` — no new secrets. `v*` does not match the widget's `w_v*` tags.

**Compose** now reads `ghcr.io/dimagi/open-chat-studio:${OCS_VERSION:-latest}`, keeping `build:` so pull-and-run and build-from-source both work.

Two deliberate departures from existing config, both worth a look:

- **No zstd compression.** `deploy.yml` uses `compression=zstd`, which is right for our own ECS but needs Docker 23+/containerd to pull. This image goes to operators on unknown Docker versions, so it stays on default gzip.
- **amd64 only.** arm64 under QEMU would mean compiling Python C extensions *and* a full node asset build under emulation. Wants a native arm runner; worth its own issue rather than blocking this.

ECR and Dimagi's deploy path are untouched — this is purely additive.

### Verification
Built the bookworm image locally (`docker build`, exit 0) and exercised it rather than trusting the build:

- `manage.py check --settings=config.settings_production` — no issues
- Base is `12.15`; `.git` confirmed absent from `/code`; `libasound`/`libmagic` both load; static files collected
- **ffmpeg went 4.x → 5.1**, which is the real behavioural risk since audio goes through pydub. Ran the actual `apps.channels.audio.convert_audio` path inside the image for ogg→mp3, ogg→opus and ogg→wav (the Telegram and WhatsApp targets at `telegram_channel.py:65` and `messaging_service.py:193`). All produce correct-duration mono output.
- `actionlint` clean on the new workflow

The `RequestsDependencyWarning` and transformers/PyTorch notices in container output also reproduce in the local venv — pre-existing and unrelated to the base image.

The publish workflow itself can't be verified without pushing a `v*` tag, which would publish a real public image. Worth reviewing that file closely rather than trusting me on it.

### Left out
The issue's last two checklist items target files that only exist on the unmerged #4280: `docs/hosting/releases.md` and `RELEASING.md`. `docs/hosting/docker.md` exists on `main`, but #4280 already rewrites the exact section that would need editing, so touching it here guarantees a conflict. Once #4280 merges, the pull path and `OCS_VERSION` want documenting in those two files — either as a follow-up or folded into #4280 while it's still a draft.

Also still open from #4282: passing the version through as a build arg, which belongs to #4281.

### Migrations
None.

### Docs and Changelog

### Operator Impact

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,5 @@
+.git
+.gitignore
 media
 static/css
 static/js
--- a/.github/workflows/publish_image.yml
+++ b/.github/workflows/publish_image.yml
@@ -1,0 +1,66 @@
+# Publishes a versioned container image for third-party self-hosted operators.
+# Triggered by the release tags described in RELEASING.md. Dimagi's own deploys
+# are unaffected: they build and push to ECR from deploy.yml.
+
+name: Publish Container Image
+
+on:
+  push:
+    tags:
+      # Release tags only. The chat widget's `w_v*` tags do not match.
+      - 'v*'
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish:
+    name: Build and publish image
+    runs-on: ubuntu-latest
+
+    steps:
+      # Audit only: records egress without blocking. Flip to `block` plus an
+      # `allowed-endpoints` list once the recorded traffic looks stable.
+      - name: Harden the runner
+        uses: step-security/harden-runner@b09bb98e06d4d774595224525879c09bc6e98c40 # v2.20.1
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@v7
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Derive image tags
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/${{ github.repository }}
+          # v1.2.3 -> `1.2.3` and `1.2`. `latest` follows automatically, and
+          # metadata-action withholds it from pre-releases (v1.2.3-rc1).
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+
+      - name: Build and push
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: Dockerfile
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          # Deliberately NOT using deploy.yml's zstd-compressed output: zstd
+          # layers need Docker 23+/containerd to pull. Fine for our own ECS,
+          # but this image goes to operators on unknown Docker versions, so
+          # it stays on the default gzip.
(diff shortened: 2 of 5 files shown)


SmittieC (approved):
Cool

PR #4284, opened 2026-08-26
Add the public channel behind flag_public_channel

barry47products:
### Product Description

Step 2 of #3682 (design: `docs/design/public-channel.md`, D1 to D6, D8, D9). Behind `flag_public_channel`, a chatbot can get a "Public link" channel: a `/c/<token>/` page on the OCS host with the chat widget in kiosk mode. Share the link and people can chat, no embedding needed.

What a builder sees:

- "Public link" in the add-channel dropdown (one per chatbot), with welcome messages and starter questions as channel settings.
- The channel dialog shows the link, a copy button and "Regenerate link". The chatbot home gets a "Copy link" chip next to the channel button.
- The page shows the chatbot name and description above the widget. If the channel is disabled or nothing is published, visitors get a banner instead of a live widget.

Visitors only ever reach the published version. Regenerating the link revokes the old one straight away and ends every live conversation on it. Disabling or deleting the channel ends live conversations too.

The old chat-widget prototype behind `flag_chat_widget` is gone; that flag row can be deleted from `/admin/flags/` after merge.

### Technical Description

- `ChannelPlatform.PUBLIC` maps to `ApiChannel` and reuses the embedded widget's `widget_token` embed key and session-token auth. `ChannelPlatform.widget_platforms()` covers the places both widget platforms share (auth lookups, `widget_auth_level`). The auth-level ratchet, migration 0029, `EmbeddedWidgetChannelForm`, `fetch_widget_embed_key`, the widget update status and the outdated-widget report stay `EMBEDDED_WIDGET` only, since the platform bundles the widget a public link serves.
- One origin rule, `channel_origin_allowed(request, channel)`, replaces the two `allowed_domains` reads: `allowed_domains` for the embedded widget, the `Site` hostname (port ignored) for a public channel.
- Start refuses a public visitor with `409 no_published_version`. Send and upload re-resolve the published version on every request and return the same 409 if it has gone. Team members are exempt at start and per request so they can preview an unpublished chatbot through its page.
- `ChatAPIRateThrottle` keys a public start on the visitor IP (so `RATE_LIMIT_TRUSTED_PROXY_COUNT` matters behind a proxy). The page itself is on the existing `public_chat` scope.
- "Regenerate link" posts `regenerate_link=1` to the edit dialog. `PublicChannelForm.clean()` mints the token; `post_save` calls `ExperimentChannel.end_live_sessions()`, which marks the channel's open sessions `COMPLETE` (what the widget and API key off). `soft_delete()` and disabling the channel do the same.
- The page is a standalone template with a CSP, `X-Robots-Tag: noindex` and `Referrer-Policy: origin`; no Sentry or analytics on it. `/c/` is disallowed in `robots.txt`. `user-id`/`user-name` are passed to team members only.
- `ChannelPlatform.for_dropdown` now tolerates a used platform whose flag is off (it used to `KeyError` on the chatbot home).

Things to know when running it:

- `persistent-session="tab"`, now that widget 0.12.0 is live on the site. The session lives in `sessionStorage`, so it survives a reload and clears when the tab closes, and a shared browser does not hand the next visitor the previous conversation.
- Everything is pinned to the Django `Site` domain. If that row does not name the deployed host, links 404 and starts 403. The hosting docs say so now.
- Turning the flag off hides the channel type from the dropdown but does not revoke existing links. Regenerate or delete the channel for that.
- The `/c/<token>/` lookup on `extra_data` is unindexed until the GIN index follow-up; the `public_chat` per-IP throttle bounds it.
- Unchanged platform shape worth knowing: an anonymous start with the working version's `chatbot_id` and no embed key still reaches the team API channel and its published-or-working version (that is the keyless-start sunset's job); after start, chat endpoints throttle per session rather than per IP; `RATE_LIMIT_ENFORCE` defaults to log-only; the start response includes version names and descriptions.

Follow-ups from the design, not in this PR: GIN index on `extra_data`, scrubbing `/c/<token>/` from logs and Sentry URLs before GA, step 3 (consent in the widget).

### Migrations

`0034_alter_experimentchannel_platform`: choices only, no data touched.

### Demo

Local dev server, flag on, `Site` domain `localhost:8000`, published chatbot.

| Step | Screenshot |
|---|---|
| 1. Create dialog | 01-create-dialog |
| 2. Edit dialog after create: link, copy, regenerate | 02-edit-dialog-with-link |
| 3. "Copy link" chip on the chatbot home | 03-home-copy-chip |
| 4. `/c/<token>/` as an anonymous visitor | 04-public-page-visitor |
| 5. A starter question answered (participant `platform=public`, trace on the published version) | 05-public-chat-reply |
| 6. "Link regenerated" | 06-after-regenerate-dialog |
| 7. The old tab: session ended, composer locked (old URL 404s, old key 401s) | 07-old-tab-chat-ended |
| 8. The new link chatting | 08-new-link-works |
| 9. After unpublishing: banner, widget disabled, live send and fresh start both `409 no_published_version` | 09-unpublished-banner |

The public page renders light regardless of the app theme: it is a standalone template with no `data-theme` handling. Worth a follow-up if we want it to follow the visitor's OS preference.

### Docs and Changelog

Channel docs gain the "Public link" channel (flag-gated); the automation opens the docs PR on merge. Hosting docs (rate-limit scopes, `RATE_LIMIT_TRUSTED_PROXY_COUNT`, `Site` domain) are updated here.

### Operator Impact

Nothing breaks on upgrade. The channel sits behind `flag_public_channel`, which is off by default, and no existing behaviour changes until a team is opted in at `/admin/flags/`. Before enabling it:

- **The Django `Site` row must name the deployed host.** Public links are built from it, and every request to `/c/<token>/` is checked against it. If it does not match, links 404 and their chat starts are refused with 403. The value is cached per process, so a change needs a restart. Written up in `docs/hosting/configuration.md`.
- **Set `RATE_LIMIT_TRUSTED_PROXY_COUNT` behind a proxy or tunnel.** A public link start is keyed on the visitor IP, and the page itself is on the `public_chat` scope. Left at `0` behind a proxy, every visitor shares one bucket.
- **Visitors' browsers fetch the widget from `unpkg.com`.** The page's CSP allows `unpkg.com` for scripts and `cdnjs.cloudflare.com` for fonts and styles. A deployment that expects visitor traffic to reach only its own origin needs to know that.
- `robots.txt` now disallows `/c/`.

Turning the flag off later hides the channel type from the dropdown but does not revoke links already handed out. Regenerate or delete the channel for that.

After merge the `flag_chat_widget` row can be deleted from `/admin/flags/`; the prototype it gated is gone.


apps/api/views/chat.py
```diff
@@ -498,6 +582,9 @@ def chat_start_session(request):
     if disabled := _channel_disabled_response(experiment_channel):
         return disabled
 
+    if refusal := _public_channel_refusal(request, experiment, experiment_channel):
+        return refusal
```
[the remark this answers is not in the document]
barry47products: @coderabbitai Fixed in c6d1b13fa: the start response now serializes the published version for public visitors (`_public_channel_admission` returns it). Test: `test_start_response_describes_the_published_version`.

apps/channels/forms.py
```diff
@@ -751,6 +752,96 @@ def post_save(self, channel: ExperimentChannel):
...
+
+    def post_save(self, channel: ExperimentChannel):
+        if self._previous_token and channel.extra_data.get("widget_token") != self._previous_token:
+            ended = channel.end_live_sessions()
+            self.success_message = (
+                f"Link regenerated. The old link no longer works and {ended} live conversation(s) were ended."
+            )
+        elif self._previously_enabled and not channel.enabled:
+            ended = channel.end_live_sessions()
+            self.success_message = f"Channel disabled. {ended} live conversation(s) were ended."
```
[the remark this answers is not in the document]
barry47products: @coderabbitai Fixed in c6d1b13fa: `ChannelFormWrapper.save()` runs the channel save and `post_save` in one `transaction.atomic()`. Test: `test_regeneration_rolls_back_when_ending_sessions_fails`.

apps/teams/flags.py
```diff
@@ -47,7 +47,7 @@ class Flags(FlagInfo, Enum):
 
     NOTIFICATIONS = ("flag_notifications", "User notifications", "", [], False, True)
 
-    CHAT_WIDGET = ("flag_chat_widget", "Use the embedded chat widget on the full-page web chat (POC)")
+    PUBLIC_CHANNEL = ("flag_public_channel", "Public link channel served by the chat widget", "", [], False)
```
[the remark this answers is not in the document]
barry47products: No change: removing the prototype and `flag_chat_widget` is part of the approved design for #3682 (D9). The flag row can be deleted from the admin after merge, as noted in the description.

apps/web/meta.py
```diff
@@ -16,6 +16,11 @@ def get_server_root(is_secure: bool = settings.USE_HTTPS_IN_ABSOLUTE_URLS) -> st
     return f"{get_protocol(is_secure)}://{Site.objects.get_current().domain}"
 
 
+def canonical_hostname() -> str:
+    """The hostname OCS is served from, for origin checks: the Site domain without a port."""
+    return Site.objects.get_current().domain.split(":")[0].lower()
```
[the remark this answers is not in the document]
barry47products: Fixed in c6d1b13fa: `hostname_of()` parses the value as a URL authority and both `canonical_hostname()` and the page host check use it, so IPv6 literals compare whole. Tests: `test_public_channel_origin_rule_on_an_ipv6_site`, `test_page_host_check_on_an_ipv6_site`.

barry47products:
CodeScene gate: the one remaining failure is Low Cohesion on `apps/channels/forms.py` (56 functions, 7+ responsibilities, threshold 4). The file already held one form per platform (Telegram, WhatsApp, SureAdhere, Facebook, Slack, CommCare Connect, embedded widget, email); adding the public link form pushed the count over. The fix is to split the module into a package with one module per platform, which is mechanical but touches every import of `apps.channels.forms` and rewrites blame on ~900 lines, so I have left it for a follow-up PR rather than fold it into this one. The other CodeScene findings from the first run (complexity in `chat_send_message`, `chat_start_session`, `chat_upload_file`, `for_dropdown`) were addressed in e870967c3 and now show as improvements.

apps/chatbots/public_link.py
```diff
@@ -0,0 +1,92 @@
...
+            "channel": channel,
+            "state": state,
+            "chatbot_name": shown.name,
+            "chatbot_description": shown.description,
+            "public_id": channel.experiment.public_id,
+            "token": token,
+            "api_base_url": get_server_root(),
+            "welcome_json": json.dumps(channel.extra_data.get("welcome_messages", [])),
+            "starters_json": json.dumps(channel.extra_data.get("starter_questions", [])),
+            "user": request.user if member else None,
```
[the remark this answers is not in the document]
barry47products: @claude Confirmed, and the chain is as described. Fixed in f128029 on the API side rather than the page: withholding `user-id` from non-members is deliberate (see `test_logged_in_non_member_gets_no_user_id`; a visitor from another team should not have their email land in this team's participant list), so the start view now treats a non-member on a public channel as an anonymous visitor even when the browser carries an OCS session cookie. Team members keep their identity so they can preview before publishing.

The masking test is split into two: the member case still sends `participant_remote_id=user.email`, and the non-member case sends a widget-style generated id and asserts an anonymous `public` participant with no linked user. The second one reproduced the 400 before the fix.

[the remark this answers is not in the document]
barry47products: @claude Agreed on both residuals. On the first: the membership check runs once per request (`_is_public_visitor` at start, `_public_session_version` on the following send/poll), the same count as before the fix, so there is nothing to fold within one request. The CSRF point is pre-existing for any widget on the OCS host; noted for public-link reports.

apps/api/views/chat.py
```diff
@@ -354,6 +359,83 @@ def _channel_disabled_response(experiment_channel) -> Response | None:
...
+
+    Public visitors only ever reach the published version, and a consent-form chatbot has no
+    live link until consent moves into the widget.
+    """
+    try:
+        published = resolve_chatbot_version(experiment, VersionSelectionRule.LATEST_PUBLISHED)
+    except NoPublishedVersion:
+        return None, Response(NO_PUBLISHED_VERSION, status=status.HTTP_409_CONFLICT)
+    if published.consent_form_id:
+        return None, Response(CONSENT_UNAVAILABLE, status=status.HTTP_409_CONFLICT)
```
snopoke: I missed this in the spec but this will break existing bots so I think we should leave it out.

apps/channels/tests/test_models.py
```diff
@@ -281,3 +282,40 @@ def test_webhook_url_for_telegram_channel():
...
     assert str(channel.external_id) in url
     assert url.startswith("https://")
+
+
+@pytest.mark.django_db()
+class TestPublicChannelPlatform:
+    """PUBLIC is a widget platform, one per chatbot, offered only behind flag_public_channel."""
+
+    @pytest.fixture()
+    def public_flag_enabled(self, experiment):
```
snopoke: You can use the `override_flag` decorator instead of creating your own fixture: https://waffle.readthedocs.io/en/stable/testing/automated.html#unit-tests

apps/chatbots/public_link.py
```diff
@@ -0,0 +1,92 @@
...
+            "widget_enabled": state.live or member,
```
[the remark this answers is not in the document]
barry47products: @claude Fixed in 98ea43ed4. widget_enabled now excludes the disabled state from the member exemption, matching the API ordering, and test_team_member_gets_a_disabled_widget_on_a_disabled_channel covers the member + disabled combination.

apps/channels/tests/test_public_channel.py
```diff
@@ -0,0 +1,257 @@
...
+
+def _operator(client, team, codename):
+    user = team.members.first()
+    user.user_permissions.add(Permission.objects.get(codename=codename))
+    client.force_login(user)
+    return user
+
+
+@pytest.fixture()
+def public_flag(experiment):
```
snopoke: same here about using `override_flag`

apps/channels/models.py
```diff
@@ -345,27 +362,48 @@ def is_disabled(self) -> bool:
 
     @property
     def widget_update_status(self) -> widget_versions.WidgetUpdateStatus | None:
-        if self.platform_enum != ChannelPlatform.EMBEDDED_WIDGET:
+        if self.platform_enum not in ChannelPlatform.widget_platforms():
```
snopoke: I think this should stay as just the embedded widget. The user is unable to change upgrade the widget used on the platform so informing them of the update is confusing.

apps/channels/models.py
```diff
@@ -345,27 +362,48 @@ def is_disabled(self) -> bool:
...
+            ended += 1
+        return ended
+
     @property
     def min_widget_version(self) -> str | None:
         """Minimum widget version required by this channel's current auth level.
 
         None for non-widget channels or a NONE-level widget channel (no floor).
         """
+        if self.platform_enum not in ChannelPlatform.widget_platforms():
```
snopoke: similar comment to above - I think this logic and below should only consider the embedded widget channel:

```suggestion
        if self.platform_enum not in ChannelPlatform.EMBEDDED_WIDGET:
```

apps/channels/management/commands/list_outdated_widget_versions.py
```diff
@@ -79,7 +79,7 @@ def _collect_rows(self, cutoff, team_slug=None, deprecated_only=False, include_u
             .values("count")[:1]
         )
         channels = (
-            ExperimentChannel.objects.filter(platform=ChannelPlatform.EMBEDDED_WIDGET)
+            ExperimentChannel.objects.filter(platform__in=ChannelPlatform.widget_platforms())
```
snopoke: This should only consider the EMBEDDED_WIDGET channel since the widget used by public channels is bundled with the platform.

apps/chatbots/public_link.py
snopoke, line 1: Small issue for a followup PR. The patter we have in other apps when we want to split views into multiple modules is to create a `views` package and place all the view modules in that package:

```
apps/chatbot/views/
 |- __init__.py  # optionally import views here for compatibility with existing code
 |- views.py  # or rename it to something else
 |- public_link.py
```

apps/chatbots/public_link.py
```diff
@@ -0,0 +1,92 @@
...
+    if published.consent_form_id:
+        banner = "This chatbot needs your consent, which the public link cannot collect yet."
+        return PageState("consent_unavailable", banner), published
+    return PageState("live", None), published
+
+
+@waf_allow(WafRule.NoUserAgent_HEADER)
+@public_chat_rate_limited
+def public_link_page(request, token: str):
+    if hostname_of(request.get_host()) != canonical_hostname():
```
snopoke, line 74: This could be tough to debug. Can I suggest adding a message or some other indication if the user accessing the page is authenticated.

templates/chatbots/public_link.html
```diff
@@ -0,0 +1,42 @@
...
+        {% if chatbot_description %}<p class="opacity-80">{{ chatbot_description }}</p>{% endif %}
+      </header>
+      {% if not state.live %}
+        <div role="alert" class="alert alert-warning w-full max-w-3xl mb-4" data-state="{{ state.code }}">
+          <i class="fa-solid fa-triangle-exclamation"></i>
+          <span>{{ state.banner }}</span>
+        </div>
+      {% endif %}
+      <div class="w-full max-w-3xl flex-1 relative">
+        {# "tab" once widget 0.12.0 is LATEST_VERSION: the 0.11.0 widget coerces any string to true #}
```
snopoke: this can be done now since 0.12.0 is released and live on the site

snopoke (commented):
Looking good.

templates/chatbots/public_link.html
barry47products: Done in 4f99fc07f: `persistent-session="tab"`, comment removed.

This also changes behaviour. The 0.11.0 widget read `"false"` as true, so the session went to `localStorage`, and a shared browser could hand the next visitor the previous conversation. `"tab"` uses `sessionStorage`, so it survives a reload and clears when the tab closes.

apps/channels/management/commands/list_outdated_widget_versions.py
barry47products: Reverted in 494d2902f. The test that asserted the opposite is now inverted, so a public channel on an old version stays out of the report.

apps/channels/models.py
barry47products: Narrowed back to `EMBEDDED_WIDGET` in d0509ce59, along with `min_widget_version` and `pending_min_widget_version` from your other comment.

apps/channels/models.py
barry47products: Done in d0509ce59, together with `widget_update_status` above.

One deviation from the suggestion: I used `!=` rather than `not in`. Since `ChannelPlatform` is a `TextChoices`, its members are `str`, so `x not in ChannelPlatform.EMBEDDED_WIDGET` is a substring test against `"embedded_widget"` rather than a membership check. It gives the right answer for all twelve platforms today, but would quietly break on a platform valued `"widget"` or `"embed"`.

apps/chatbots/public_link.py
barry47products: Done in 45d274786, following the `TeamAccessDenied` convention: a tagged `Http404` subclass that `404.html` recognises through the `exception` context variable, the same way `apps/experiments/decorators.py:110` does it for superusers.

A signed-in viewer now gets both hosts named on the 404 page. An anonymous visitor still gets the bare 404, so the request is refused before any lookup, and nothing about the deployment reaches the public.

apps/api/views/chat.py
barry47products: Removed in 19e4d3021, from the API and the page.

Nothing else in the Chat API reads `consent_form`, so the public link was the only surface applying it, and a chatbot carrying its team's form would have had a link that never opened.

I took the page banner out with it, since it would otherwise refuse on a page whose widget the API now serves. Happy to put it back as a warning if you would rather keep it.

`409` on start-session is now `no_published_version` only. Tests cover a consent-form chatbot starting a session, a send continuing after a consent form is published mid-session, and the page rendering the widget with no banner.

apps/channels/tests/test_models.py
barry47products: Tried it, and the flag stays off. The platform gate runs through `Flag.is_active_for_team` (`apps/teams/models.py:246-263`), which matches the team M2M and never reads `everyone`, which `override_flag` sets. A probe with the decorator, asserting `PUBLIC in for_dropdown(set(), team)`, fails.

The duplication is fair, though: aaac6feee moves the fixture into `apps/channels/tests/conftest.py` so both files share one.

Worth a look beyond this PR. `apps/documents` restores the `everyone` precedence in a helper of its own, so `override_flag` works there and not here. Filed as #4321.

apps/api/views/chat.py
barry47products: One correction to the above: I should have said the only enforcement in the Chat API path. The web chat still collects consent through `_run_public_consent_flow`, and this PR doesn't touch it.

apps/channels/tests/test_public_channel.py
barry47products: Same as the other thread: `override_flag` doesn't reach the team gate, so the fixture stays. It moved into `apps/channels/tests/conftest.py` in aaac6feee, and both files now share it.

apps/chatbots/public_link.py
```diff
@@ -0,0 +1,92 @@
...
+    banner: str | None
+
+    @property
+    def live(self) -> bool:
+        return self.banner is None
+
+
+def _page_state(channel: ExperimentChannel) -> tuple[PageState, Experiment | None]:
+    if channel.is_disabled:
+        return PageState("disabled", channel.disabled_message or "This chatbot is temporarily unavailable."), None
```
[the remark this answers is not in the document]
barry47products: @claude Confirmed, and fixed in 971b841f0. Reproduced it first: enabled, you get the published name; disabled, you get the draft.

`_page_state` now resolves the published version before any refusal and hands it back on the disabled branch. `no_published_version` still falls back to the working version, since there's no published row to name there. The regression test covers both states and renames the draft after publishing, which the old helper couldn't do.

apps/chatbots/public_link.py
barry47products: Filed as #4322. `apps/chatbots/views.py` becomes `apps/chatbots/views/`, `public_link.py` moves alongside it, and `__init__.py` re-exports so existing imports and `urls.py` keep working.

snopoke (approved):

PR #4285, opened 2026-08-26
Run psql and redis-cli from the dev containers in worktree setup

pxwxnvermx:
### Product Description

Worktree setup and teardown no longer need `psql` or `redis-cli` installed on the host — both run inside the containers `docker-compose-dev.yml` already defines. Without this, setup on such a host dies with `redis-cli: command not found` after having already created the branch database.

### Technical Description

`scripts/worktree-environment.sh` gains `ocs_psql` / `ocs_redis_cli`, and every existing call site goes through them:

- The container is found by the host port it publishes, read out of `docker ps --format '{{.Names}} {{.Ports}}'`. Not `--filter publish=`, which podman's docker CLI shim rejects as an invalid filter.
- No container → the host client, so a locally installed server or a CI service container still works. Neither available → a message naming the missing client, instead of a bare "command not found".
- A container that cannot start a client (`docker exec` 125/126/127) is skipped and the next candidate tried. Any other status is the client's own and is returned untouched, so a statement the server rejected is never replayed elsewhere.
- `OCS_POSTGRES_CONTAINER` / `OCS_REDIS_CONTAINER` pin a container, read from the environment or the worktree's `.env` (nothing sources `.env` for these scripts, so they parse it, as `OCS_WORKTREE_ID` already did). Pre-existing knobs are unchanged.

Safe to run the client in the container because every `ocs_psql` caller passes SQL inline with `-c`/`-tAc` — no `-f`, no stdin, no `\copy`. `-h localhost` is correct either way: the container's own server, or the published port.

Rationale for the individual decisions is in the commit messages.

### Migrations

None.

### Docs and Changelog

Included: `.env.example` documents the two knobs; `docs/getting-started/dev-workflow.md` describes the container clients under *What worktree setup does*. No changelog entry — developer tooling, no user-facing surface.

### DIFF
--- a/scripts/test_worktree_environment.py
+++ b/scripts/test_worktree_environment.py
@@ -574,6 +574,74 @@ def test_psql_recovers_from_a_stale_container_override(
     assert "psql:-h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 -c SELECT 1" in command_output
 
 
+@pytest.mark.parametrize(
+    ("env_line", "expected_container"),
+    [
+        pytest.param("OCS_POSTGRES_CONTAINER=ocs_env_postgres", "ocs_env_postgres", id="bare"),
+        pytest.param("OCS_POSTGRES_CONTAINER='ocs_env_postgres'", "ocs_env_postgres", id="quoted"),
+    ],
+)
+def test_psql_reads_its_container_from_the_worktree_env_file(
+    worktree_fixture: tuple[Path, Path, dict[str, str], Path],
+    env_line: str,
+    expected_container: str,
+) -> None:
+    _, worktree, env, command_log = worktree_fixture
+    # Nothing sources `.env` before these scripts run, so a knob set there only works if
+    # they read the file themselves — the way OCS_WORKTREE_ID always has.
+    with (worktree / ".env").open("a") as env_file:
+        env_file.write(f"{env_line}\n")
+
+    _run_worktree_helper(worktree, "ocs_psql", "postgres", "-c", "SELECT 1", env=env)
+
+    assert f"docker-exec:{expected_container}" in command_log.read_text()
+
+
+def test_an_environment_variable_beats_the_worktree_env_file(
+    worktree_fixture: tuple[Path, Path, dict[str, str], Path],
+) -> None:
+    _, worktree, env, command_log = worktree_fixture
+    with (worktree / ".env").open("a") as env_file:
+        env_file.write("OCS_POSTGRES_CONTAINER=ocs_env_postgres\n")
+    env["OCS_POSTGRES_CONTAINER"] = "ocs_exported_postgres"
+
+    _run_worktree_helper(worktree, "ocs_psql", "postgres", "-c", "SELECT 1", env=env)
+
+    command_output = command_log.read_text()
+    assert "docker-exec:ocs_exported_postgres" in command_output
+    assert "ocs_env_postgres" not in command_output
+
+
+def test_a_container_override_that_is_not_a_container_name_is_ignored(
+    worktree_fixture: tuple[Path, Path, dict[str, str], Path],
+) -> None:
+    _, worktree, env, command_log = worktree_fixture
+    # The value is interpolated into `docker exec`, and `.env` is a file anything can
+    # write, so a name-shaped check is what keeps the rest of a line out of the command.
+    env["OCS_POSTGRES_CONTAINER"] = "nope; rm -rf /"
+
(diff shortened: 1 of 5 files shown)


snopoke (commented):
Please update the docs with the new env vars that are used here:
* `OCS_REDIS_CONTAINER`
* `OCS_REDIS_PORT`
* `OCS_POSTGRES_CONTAINER`
* `OCS_POSTGRES_PORT`

scripts/test_worktree_environment.py
```diff
@@ -4,6 +4,7 @@
 
 import json
 import os
+import shutil
```
pxwxnvermx, line 7: Fixed in 92f1765c6. The bash fakes, driver helpers and fixture moved out to `scripts/worktree_harness.py`, re-exported through a new `scripts/conftest.py` (matching the shared-fixture convention in `apps/conftest.py`), which takes the test module from 1079 raw lines to 842 — comfortably back under the 600-LOC threshold, with the harness itself at ~371.

The split is along a real seam rather than a size one: assertions in the test module, the scaffolding they need in the harness.

*— Claude Code*

scripts/worktree-environment.sh
```diff
@@ -89,6 +89,65 @@ ocs_worktree_resource_name() {
...
+        docker exec "$container" redis-cli "$@"
+        return
+    fi
+    # A Redis reachable on the host port without a container of its own -- a service
+    # container in CI, or a locally installed server -- still answers a host client.
+    if ! command -v redis-cli >/dev/null 2>&1; then
+        ocs_report_missing_client redis-cli "${OCS_REDIS_PORT:-6379}"
+        return 1
+    fi
+    redis-cli "$@"
```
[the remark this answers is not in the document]
pxwxnvermx: Real issue, fixed in 92f1765c6 — but by removing the knob rather than by passing `-p`.

`OCS_REDIS_PORT` and `OCS_POSTGRES_PORT` are gone; discovery now matches literal 6379/5432. Adding `-p` to the host clients would have left the other half of the mismatch in place: `setup-worktree.sh` hardcodes 5432/6379 into the `DATABASE_URL` and `REDIS_URL` it writes to `.env`, so a non-default port would still have pointed Django at the wrong server even once the shell helpers agreed. A knob that only half-moves a service is worse than no knob — the half left behind is a `FLUSHDB` against whatever else answers on 6379.

No non-default-port tests, since there is no longer a non-default port to configure.

*— Claude Code*

scripts/test_worktree_environment.py
pxwxnvermx: Correction to the reply above: that split is reverted in 72cf4ef8c, and this file-size issue is back.

Extracting the harness dropped the file under 600, but left 40 test functions referencing nothing in their own module — read as 40 responsibilities against a threshold of 4. That traded this advisory violation (10.00 → 9.39) for a critical one (10.00 → 8.03), so it was the wrong trade.

*— Claude Code*

pxwxnvermx, line 7: Not resolvable by rearranging this module — leaving it as a human call.

- Extracting the fakes, helpers and fixture into a harness + `conftest.py` gets under 600, but then the 40 remaining test functions share nothing module-local: Low Cohesion, 40 responsibilities vs a threshold of 4, and a worse score (10.00 → 8.03 critical, versus 10.00 → 9.39 advisory here). Tried in 92f1765c6, reverted in 72cf4ef8c.
- Moving the file under `scripts/tests/` does not help. `.codescene/custom-quality-gates.json` waives every gate for `**/tests/**`, but the moved path was gated all the same, so that repo-local config is not what the "Clean Code Collective" profile applies. Tried in fc11c3c73, also reverted.
- Moving the embedded bash fakes into real `.sh` files would not move the number: the count already excludes string-literal interiors (1079 raw lines report as 630).

So the two rules cannot both hold for a 40-test module. The choices are to suppress this biomarker for the file, or to make the profile honour the test-path exemption this repo already declares — which would also cover the four other `scripts/test_*.py` files, mis-gated as production code for the same reason.

*— Claude Code*

snopoke:
> Please update the docs with the new env vars that are used here:
> 
> * `OCS_REDIS_CONTAINER`
> * `OCS_POSTGRES_CONTAINER`

@pxwxnvermx this PR is still in draft state. Also don't forget to document these vars.

pxwxnvermx:
@snopoke yeah, I was trying to confirm that my changes are working correctly locally. I will update the PR soon with the docs for the variables.

pxwxnvermx:
I have verified these changes locally, setting up and tearing down the worktrees works using the docker containers pre-packaged psql and redis-cli binaries.

snopoke (approved):

PR #4286, opened 2026-08-26
docs: document global (team-less) OAuth applications

snopoke:
### Product Description
No user-facing change. Documents an existing superuser feature: OAuth applications registered with no team, from Global Admin > OAuth Apps.

### Technical Description
Global applications have been supported since the team-scoping work, but the only documentation was code comments in `apps/oauth/`, and `CONTEXT.md` actively contradicted them — the glossary entry opened "A **Team-scoped** registration", which reads as ruling the team-less case out.

Two things worth checking as domain claims rather than prose:

- The guide leans hard on **application-is-global / token-never-is**, since that's the part most likely to be misread by anyone wiring up a client. `CONTEXT.md` gets a matching `_Avoid_` entry for "a global token".
- It states the client secret **cannot be rotated** — `RegisterApplicationForm` swaps the field for a `HiddenInput` once the app exists, and `ClientSecretField.pre_save` hashes it. Deliberately omitted: it *is* settable via `/django-admin/`, which I'd rather not document as a supported path.

`allowed_chatbots` isn't mentioned — client-credentials only, which global apps can't be.

Still absent, and out of scope: an ADR for the authorization-code-only decision and the `prompt=none` open-redirect guard in `_refuse_non_member`. Happy to follow up.

### Demo
N/A — docs only. `zensical build --clean` passes with no new warnings.

### Docs and Changelog

Internal developer/admin docs, already in this repo — nothing for the public changelog.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -118,9 +118,10 @@ A Team-scoped, versioned wrapper around a resource in OpenAI's Assistants API. P
 _Avoid_: bare "Assistant" — it overloads with the colloquial sense ("the chatbot as an assistant").
 
 **OAuth Application**:
-A Team-scoped registration that lets an external system authenticate to OCS. Two kinds, fixed at registration: **machine** applications (client credentials — no human, no login, a synthetic service identity acting for the Team) and **user-facing** applications (authorization code — a real User signs in and consents).
+A registration that lets an external system authenticate to OCS. Two kinds, fixed at registration: **machine** applications (client credentials — no human, no login, a synthetic service identity acting for the Team) and **user-facing** applications (authorization code — a real User signs in and consents).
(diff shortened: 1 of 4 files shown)


ocs-percy (approved):
LGTM — clean docs-only change.

- CONTEXT.md fix is correct: removing "Team-scoped" from the glossary opener and adding the global-app explanation (plus the `_Avoid_: "a global token"` entry) closes a real gap.
- The new guide is well-structured: the comparison table up front, the explanation of *why* client-credentials is unavailable for global apps, and especially the "what a token is scoped to" section ("The application is global; a token is not") are exactly the things someone wiring up a client would otherwise get wrong.
- Good call omitting the `/django-admin/` secret rotation path — not a supported workflow.

Nothing to block here.

PR #4288, opened 2026-08-26
Inbound message deduplication

### LINKED ISSUE #3240: Handle duplicate incoming messages from commcare-connect
There have been some instances where we received duplicate messages from CommCare Connect resulting in duplicate messages being sent to the user.

CC includes a 'message_id' field in the payload. We should use this to abort early if we have already received this message.

We don't currently store the `message_id` but I think we can store it in the `ChatMessage.metadata` field or a new `external_id` field which may be better for querying.

## Extra
Also check other incoming webhooks to see if a similar approach should be take.

zandre-eng:
### Product Description

Messaging providers retry inbound webhooks, and until now every retry was processed as a fresh message, so a participant could get the same answer from the bot twice.

Replayed deliveries are now recognised and dropped silently. This covers Connect, Meta, Turn.io, Telegram, Twilio, Email and Slack, with SureAdhere being excluded because it sends no message identifier. A partly-overlapping Connect batch still has its new messages processed in full, so nothing a participant sent is dropped in order to deduplicate.

No UI changes.

### Technical Description

[Link to issue here](https://github.com/dimagi/open-chat-studio/issues/3240)

One field and one check. `ChatMessage.external_ids` records the namespaced provider ids a message was built from (`telegram:-100123:4471`), and a new `DuplicateDeliveryStage` at the head of `ChannelBase._build_pipeline` raises the existing `EarlyAbort` when every id on a delivery is already recorded.

`external_ids` is saved as an array because Connect joins N provider messages into a single `ChatMessage`. An empty ID list is never treated as a duplicate, so an unidentifiable delivery is always processed.

### Migrations

- [X] The migrations are backwards compatible

Both migrations are additive, with `chat.0026` adding a field with a constant default and `chat.0027` building the GIN index with `AddIndexConcurrently` to avoid locking writes on `ChatMessage`.

### Demo

No UI changes to demo.

### Docs and Changelog

### Operator Impact


apps/channels/deduplication.py
```diff
@@ -0,0 +1,79 @@
...
+    candidates = set(external_ids)
+    if not candidates:
+        return candidates
+
+    seen = chain.from_iterable(
+        ChatMessage.objects.filter(chat__team_id=team_id, external_ids__overlap=list(candidates)).values_list(
+            "external_ids", flat=True
+        )
+    )
+    return candidates - set(seen)
```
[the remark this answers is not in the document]
zandre-eng: A boundary claim has to be taken before the work, so it's orphaned whenever the pipeline dies and needs a claim/completion timestamp, staleness window and purge task to recover. Writing the record only alongside the ChatMessage means a dead pipeline leaves no trace and the retry just works. Not changing it here.

apps/chat/migrations/0026_chatmessage_external_ids.py
```diff
@@ -0,0 +1,22 @@
...
+        migrations.AddField(
+            model_name="chatmessage",
+            name="external_ids",
+            field=django.contrib.postgres.fields.ArrayField(
+                base_field=models.CharField(max_length=600),
+                blank=True,
+                default=list,
+                help_text="Provider message IDs this message was built from, namespaced by platform.",
+                size=None,
+            ),
```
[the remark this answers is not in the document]
snopoke: This is unnecessary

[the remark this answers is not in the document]
zandre-eng: Not applicable. The model field is non-null with `default=list`, so the migration matches it; `AddField` on a NOT NULL column needs the default. Django drops the DB default after backfill anyway (metadata-only on PG11+), so existing rows are {}, never NULL.

apps/channels/deduplication.py
```diff
@@ -0,0 +1,79 @@
...
+
+    False for an empty list: a delivery the provider did not identify is processed, never dropped.
+    That is why callers must use this rather than negating `unseen_message_ids`, which would invert
+    the intent with no visible symptom.
+    """
+    return bool(external_ids) and not unseen_message_ids(external_ids, team_id)
+
+
+def connect_external_ids(messages: list[dict]) -> list[str]:
+    return [namespaced_id("connect", message["message_id"]) for message in messages]
```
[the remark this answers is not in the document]
zandre-eng: `message_id` is a required `serializers.UUIDField()` and the only caller reads `serializer.data` post-validation, so null/empty is a 400 at the boundary. A missing key would also raise `KeyError`, not silently produce `"connect:None"`.

Applying `external_ids_for` here would actually regress: it returns `[]` for an unidentified message, and `unseen_connect_messages` keeps only messages whose id is in unseen, so that message would be dropped entirely. The asymmetry is deliberate. Leaving as is.

apps/channels/deduplication.py
```diff
@@ -0,0 +1,79 @@
...
+    Every channel goes through here rather than calling `namespaced_id` directly, because storing a
+    `None` sentinel is this feature's worst failure mode: a literal "slack:None" would be recorded
+    for the first message that lacks an id and silently drop every later one.
+    """
+    if any(part is None or part == "" for part in parts):
+        return []
+    return [namespaced_id(platform, *parts)]
+
+
+def unseen_message_ids(external_ids: list[str], team_id: int) -> set[str]:
```
SmittieC: super nit: make this a private method

apps/channels/tests/test_email_channel.py
```diff
@@ -1344,6 +1406,24 @@ def test_handler_persists_files_before_enqueue(self, team_with_users):
         assert kwargs["channel_id"] == channel.id
         assert len(kwargs["email_data"]["attachment_file_ids"]) == 1
 
+    @override_settings(EMAIL_CHANNEL_ALLOWED_DOMAINS=["chat.openchatstudio.com"])
+    def test_replay_is_dropped_before_files_are_persisted(self, team_with_users, record_delivery):
```
SmittieC, line 1410: A small comment about this in the inbound handler might be a good idea as well?

apps/channels/tests/channels/concrete/test_whatsapp_channel.py
```diff
@@ -93,3 +96,23 @@ def test_predicate_is_wired_into_response_sending_stage(self):
         pipeline = channel._build_pipeline()
         stage = next(s for s in pipeline.terminal_stages if isinstance(s, ResponseSendingStage))
         assert stage._should_voice_fallback_to_text == channel._should_voice_fallback_to_text
+
+
+@pytest.mark.django_db()
+class TestReplayedDelivery:
```
SmittieC, line 102: Might be a good idea to add tests for the other channels as well?

SmittieC (commented):
LGTM. A few small comments

apps/channels/stages/core.py
```diff
@@ -613,6 +648,7 @@ def process(self, ctx: MessageProcessingContext) -> None:
             message_type=ChatMessageType.HUMAN,
             content=ctx.user_query,
             metadata=metadata,
+            external_ids=ctx.message.external_ids,
```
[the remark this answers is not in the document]
SmittieC: > Meta doesn't get a timely 200 (the EventBot LLM call on this path makes that more likely)

I'm not sure that this is 100% correct. Sure, message handling happens async so there is a miniscule chance of the bot's response reaching the user before the messaging provider received the 200 response from our server. Maybe the request failed after the task was dispatched, but the chances of that happening is again, miniscule. I'm not convinced that fixing this is worth the effort.

apps/channels/datamodels.py
```diff
@@ -135,6 +142,7 @@ def parse(update_obj) -> "TelegramMessage":
             content_type=update_obj.message.content_type,
             media_id=update_obj.message.voice.file_id if update_obj.message.content_type == "voice" else None,
             message_id=update_obj.message.message_id,
+            external_ids=external_ids_for("telegram", update_obj.message.chat.id, update_obj.message.message_id),
```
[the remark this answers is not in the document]
SmittieC: @zandre-eng this seems like it might be legit, albeit I am not sure how likely it is to happen. I think an easy fix might be to scope `is_duplicate_delivery` to a chatbot as well when we have one, otherwise fall back to the team wide scope

apps/channels/deduplication.py
zandre-eng: Addressed in [13bf3e1](https://github.com/dimagi/open-chat-studio/pull/4288/commits/13bf3e1c842846f9090dc945493475d38df4d82c).

apps/channels/tests/test_email_channel.py
zandre-eng: Good idea, addressed in [5f84b06](https://github.com/dimagi/open-chat-studio/pull/4288/commits/5f84b0661bbeec3465863710ed1e0d2be525311f).

apps/channels/tests/channels/concrete/test_whatsapp_channel.py
zandre-eng: Yeah this could be good to have. I've added replay tests for Telegram and Twilio.

Addressed in [234f0ed](https://github.com/dimagi/open-chat-studio/pull/4288/commits/234f0eddcad5f0f94817be068f2252be2a3c4dca)

apps/channels/datamodels.py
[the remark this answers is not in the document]
zandre-eng: Yeah this seems like it might be worth addressing. I've added chatbot scope for the channels, where a chatbot is available.

Addressed in [324aeba](https://github.com/dimagi/open-chat-studio/pull/4288/commits/324aeba963b871b7e750d37f5851816fc1fa93d7).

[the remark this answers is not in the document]
zandre-eng: Already addressed in https://github.com/dimagi/open-chat-studio/pull/4288#discussion_r3863977542.

apps/chat/migrations/0027_chatmessage_external_ids_idx.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Migration(migrations.Migration):
+    atomic = False
+
+    dependencies = [
+        ("chat", "0026_chatmessage_external_ids"),
+    ]
+
+    operations = [
+        AddIndexConcurrently(
```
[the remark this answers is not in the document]
zandre-eng: Added.

apps/channels/deduplication.py
```diff
@@ -0,0 +1,89 @@
...
+    candidates = set(external_ids)
+    if not candidates:
+        return candidates
+
+    messages = ChatMessage.objects.filter(chat__team_id=team_id, external_ids__overlap=list(candidates))
+    if chatbot_id is not None:
+        # A session's experiment may be the working version or any of its published versions.
+        messages = messages.filter(
+            Q(chat__experiment_session__experiment_id=chatbot_id)
+            | Q(chat__experiment_session__experiment__working_version_id=chatbot_id)
```
SmittieC: Sessions (and by extension chats + chatmessage records) can only be scoped to working version chatbots, so you can drop the `chat__experiment_session__experiment__working_version_id` branch.

apps/channels/deduplication.py
```diff
@@ -31,43 +33,51 @@ def external_ids_for(platform: str, *parts) -> list[str]:
     return [namespaced_id(platform, *parts)]
 
 
-def _unseen_message_ids(external_ids: list[str], team_id: int) -> set[str]:
-    """Return the ids with no ChatMessage in this team already recording them."""
+def chatbot_id_for(channel) -> int | None:
+    """The dedup scope for a delivery that is only known by its `ExperimentChannel`."""
+    return channel.experiment.get_working_version_id() if channel.experiment_id else None
```
SmittieC, line 38: Both experiment sessions and channels always points to the "canonical" or "working" version of the experiment. The experiment version only exists to provide information that informs a bot's behavior (stuff like prompt / pipeline etc) whereas the "orchestration" layer always sticks to the original experiment. cc @pxwxnvermx

apps/channels/deduplication.py
```diff
@@ -31,43 +33,51 @@ def external_ids_for(platform: str, *parts) -> list[str]:
...
 
-    seen = chain.from_iterable(
-        ChatMessage.objects.filter(chat__team_id=team_id, external_ids__overlap=list(candidates)).values_list(
-            "external_ids", flat=True
+    messages = ChatMessage.objects.filter(chat__team_id=team_id, external_ids__overlap=list(candidates))
+    if chatbot_id is not None:
+        # A session's experiment may be the working version or any of its published versions.
+        messages = messages.filter(
+            Q(chat__experiment_session__experiment_id=chatbot_id)
+            | Q(chat__experiment_session__experiment__working_version_id=chatbot_id)
```
SmittieC, line 52: Sessions (and by extension chats + chatmessage records) can only be scoped to working version chatbots, so you can drop the `chat__experiment_session__experiment__working_version_id` branch.

apps/channels/deduplication.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+    seen = chain.from_iterable(messages.values_list("external_ids", flat=True))
+    return candidates - set(seen)
+
+
+def is_duplicate_delivery(external_ids: list[str], team_id: int, chatbot_id: int | None = None) -> bool:
```
SmittieC, line 58: Sorry if this causes some work to be reverted, but I just realized that a potential less "invasive" way to get the proper chatbot scope would be to pass in the `chatbot_id` to [TelegramMessage.parse](https://github.com/dimagi/open-chat-studio/blob/324aeba963b871b7e750d37f5851816fc1fa93d7/apps/channels/tasks.py#L61). You can get it from `experiment_channel.experiment_id`. That way we know that `external_ids` are already scoped to the user + correct chat. Thoughts on this?

zandre-eng: I like this idea. The chatbot scoping was primarily done for Telegram, so if we build it into the key we can get rid of some unnecessary code to scope by chatbot for all channels.

zandre-eng: Addressed in [413caeb](https://github.com/dimagi/open-chat-studio/pull/4288/commits/413caeb2323f28f5667c5435040a0b31310d3e6a).

apps/channels/deduplication.py
zandre-eng: Ah, that is good to know. Based on another comment I've gone with a different approach here so chatbot scoping in the query has now completely been removed.

apps/channels/deduplication.py
zandre-eng: I went with a different approach here, so this chatbot filtering has now completely been removed.

apps/channels/tests/test_email_channel.py
```diff
@@ -124,6 +125,18 @@ def _make_session(team, channel, external_id, participant_email="user@example.co
...
 
 class TestEmailMessageParse:
+    @pytest.mark.parametrize(
+        ("message_id", "expected"),
+        [
+            pytest.param("<msg1@example.com>", ["email:<msg1@example.com>"], id="with-a-message-id"),
+            pytest.param("", [], id="without-a-message-id"),
+        ],
+    )
+    def test_parse_records_the_provider_id(self, message_id, expected):
```
SmittieC: Is this test name a misnomer?

SmittieC (dismissed):
LGTM

apps/channels/tests/test_email_channel.py
zandre-eng: That's a fair point, I agree this name should be updated.

Addressed in [d60fa0c](https://github.com/dimagi/open-chat-studio/pull/4288/commits/d60fa0c770e61032f7b4fed1607474d3c4269e3b).

SmittieC (approved):

PR #4289, opened 2026-08-26
Write API | Edit a chatbot's pipeline through the API, a node at a time

SmittieC:
### Product Description

An API client — in practice an agent — can now build and edit a chatbot's pipeline one node at a time, instead of only reading it. Three new endpoints under the existing `Pipelines` tag:

| Verb | Route |
|---|---|
| `POST` | `/api/v2/chatbots/{id}/pipeline/nodes/` |
| `PATCH` | `/api/v2/chatbots/{id}/pipeline/nodes/{node_id}/` |
| `DELETE` | `/api/v2/chatbots/{id}/pipeline/nodes/{node_id}/` |

The server holds the graph: a client never reproduces a whole document to change one setting, and two edits to different nodes cannot overwrite each other. `type` alone is a valid create body — the node is written with that type's defaults, and whatever is still missing comes back in `pipeline_errors` rather than being refused, so a graph can be built up over several calls.

No UI change. Nothing existing changes behaviour except the fix below.

### Technical Description

Sixteen commits, small and worth reviewing in order.

`facade.py` is the one shape every edit takes: lock the `Pipeline` row, hand the UI builder's own patch engine a one-item diff, persist the way the builder's save does. Reusing `apply_pipeline_patch` keeps the API off a second persistence path, and building the response *inside* the lock means a node the server cannot describe fails the write rather than persisting and then 500ing every later read. `edit_revision` is bumped so an open builder session sees a conflict instead of silently overwriting an API edit.

Design points a reviewer may want to push on:

- **`fix: report a node the server cannot parse instead of raising`** is a **behaviour change to an existing path.** `Pipeline.validate()` parses every node on every read, so a param a validator chokes on before pydantic can wrap it used to 500 `/inspect/` and every write to that pipeline, permanently. Now reported and logged. `DatabaseError` still propagates — caught inside an atomic block it would leave the transaction aborted.
- **Team-scoped references are refused, not reported** — the one place writes refuse, because nothing downstream would tell the caller: another team's id lands in the FK column untouched and `Pipeline.validate` says nothing about it. A nonexistent id and another team's id answer the same on purpose. Resolvers ask after the ids they were handed rather than building the team's lists, since a PATCH runs its check under the lock; `test_a_reference_check_reads_only_the_ids_it_was_sent` pins the query count.
- **Router edges follow keywords by branch label, not position.** Handles are positional (`output_i` serves `keywords[i]`), so dropping the second of three keywords renumbers the third; matching by position would hand the third branch's target to the second. A branch that is gone takes its edge with it, as the builder's `deleteKeyword` does.
- **PATCH takes no `type`** — a node's type decides what its params mean, so changing it in place would reinterpret every stored value. Params merge key by key. Start and End answer 409, not 404.
- **`ChatbotCompositionPermission`** replaces the stock `DjangoModelPermissions` verb map, since deleting a node is a *change* to the chatbot, not a deletion of it (the stock map would demand `delete_experiment`). All three verbs are covered in `test_auth.py`.
- **`DescribesPatch`** — stock DRF metadata describes only PUT and POST, so a PATCH-only route answers OPTIONS with no body, and OPTIONS is how the agent this API is built for learns what it may send.

**Verification.** Every commit was linted and tested, and carries a regenerated `api-schemas/v2.yml` where its API surface changed. At the tip: `apps/api` 988 passed, `apps/pipelines` 637 passed, `ty` clean.

<sub>Note: running `apps/api` and `apps/pipelines` in one pytest process intermittently errors on a `users_customuser_username_key` duplicate from `UserFactory` — a pre-existing Faker collision, a different test each run, and both suites are clean run separately.</sub>

### Demo

```bash
# add a node — `type` alone is enough
curl -XPOST "$OCS/api/v2/chatbots/$ID/pipeline/nodes/" \
  -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -d '{"type": "LLMResponseWithPrompt"}'

# edit one setting; everything else is left as it is
curl -XPATCH "$OCS/api/v2/chatbots/$ID/pipeline/nodes/$NODE_ID/" \
  -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -d '{"params": {"prompt": "Be terse."}}'
```

https://www.loom.com/share/4a71d6092df34a48a6d1badf159dd9b5

Every write answers with the node as stored plus `pipeline_valid`, `pipeline_errors` and `unwired_handles`, so the next call needs no re-read.

### Docs and Changelog


apps/api/v2/pipeline_edit/views.py
```diff
@@ -0,0 +1,240 @@
...
+        label = body.validated_data.get("label")
+        return Response(
+            edit_pipeline(
+                request,
+                id,
+                # The node's type comes from the graph, so its params can only be checked under the
+                # lock; the team goes in so the check can look its references up there.
+                lambda flow: plan_update(flow, request.team, node_id, label, params),
+                self._write_response,
+            )
```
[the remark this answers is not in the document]
SmittieC: Fixed in 9ddbf7d97 (test in 0aabc394f).

The alert is right — the flow was real. `graph.py:318` raises `PipelineNodeBuildError(ex)` wrapping a raw pydantic `ValidationError`, so `str(e)` at `models.py:256` put internal model class names (and the `errors.pydantic.dev` URL) into `pipeline_errors`, which this endpoint serves.

That branch now logs the exception and appends a generic line, the same treatment the catch-all in `_node_validation_errors` already gets in this PR. Nothing user-actionable is lost: `build_runnable()` only runs when every node has already validated cleanly, so the curated messages (e.g. `"LLM provider model with id 999999 does not exist"`) still reach the client through per-node validation, which reports pydantic errors field-by-field rather than dumping the exception.

apps/pipelines/models.py
```diff
@@ -617,6 +617,15 @@ def resource_fk_fields(cls):
...
         ]
 
+    @classmethod
+    def resource_param_names(cls) -> frozenset[str]:
+        """The param names ``resource_params`` produces, whatever the node's type.
+
+        ``to_flow_node`` merges all of these into every node's params, its type declaring them or
+        not, so a caller writing params back has to tell the mirrored ids from the declared ones.
+        """
+        return frozenset({f"{field_name}_id" for field_name in cls.resource_fk_fields()} | {"collection_index_ids"})
```
SmittieC: Custom action operation have a FK to Node. Can we include that in here?

apps/pipelines/nodes/node_metadata.py
```diff
@@ -305,3 +308,130 @@ def _get_node_schema(node_class: type) -> dict:
...
+    if not ids:
+        return set()
+    return set(queryset.filter(id__in=ids).values_list("id", flat=True))
+
+
+#: The resolver behind each option list, keyed on the list rather than the param, because the list is
+#: what decides the permitted values -- ``source_material_id`` is the param, ``source_material`` the
+#: list it chooses from. ``contract.PARAMETER_OPTION_SOURCES`` is the set of sources a write checks;
+#: ``test_every_checked_param_has_a_resolver`` says every one of those is a key here.
+RESOLVERS: dict[OptionsSource, Callable[[Team, list], set]] = {
```
SmittieC, line 414: can we not move the resolvers onto OptionsSource so I can do `OptionsSource.get_resolver`

apps/pipelines/models.py
SmittieC: Done in fb26dde2a.

You're right that it belongs there — and reading it back actually fixes something. `update_from_params` writes the `CustomActionOperation` rows, and `CustomActionOperation.custom_action` is `CASCADE`, so deleting a `CustomAction` takes the row away while the stale `"<action id>:<operation id>"` stays in params. That stale entry was what `to_flow_node` served, to the builder and over the API both — exactly the case the FK columns exist to correct.

So `resource_params` now reads `custom_actions` off the rows (sorted, since they have no ordering of their own here) and `resource_param_names` reports it. `contract.PARAMETER_OPTION_SOURCES` picks it up from that derivation now instead of naming it outright — the comment claiming no FK backs it was wrong.

Two things worth flagging:

* The prefetch was a string literal at seven call sites, so it's now `NODE_RESOURCE_PREFETCHES` in `pipelines.models`. Missing one is an N+1 per node rather than per pipeline.
* `QUERIES_UNDER_THE_LOCK` goes 27 -> 29: one prefetch for the locked read, one for the rebuild after the write. Both flat in the number of nodes. Noted against the constant, as it asks.

I used the unfiltered `custom_action_operations`, not `get_custom_action_operations()`, which additionally drops operations the action no longer allows. That one is a join, so it can't be satisfied by a plain prefetch, and dropping those here would delete them from params on the next PATCH. Left alone deliberately.

Covered by `test_custom_actions_come_from_the_operation_rows`, `test_deleted_custom_action_reads_as_empty`, and `test_editor_load_reads_each_resource_table_once` (which now asserts one read per resource table, not just the indexes).

apps/pipelines/nodes/node_metadata.py
SmittieC: Done in dbc79cd04 — it's `source.get_resolver()` now, so `OptionsSource.custom_actions.get_resolver()` reads as you'd expect and the free function is gone.

The resolver functions themselves stayed in `node_metadata`. Moving them onto the enum would mean `nodes/base.py` importing `Team`, `CustomAction`, `Collection` and the rest at module scope, which is both circular and a lot of weight on the module every node class imports. So the method reads the `RESOLVERS` registry through a local import instead, and the functions stay beside the querysets the option lists are built from — the property that keeps what a client is offered and what a write accepts from coming apart.

The `NotImplementedError` for a source that can deny nothing moved with it, and now has a test of its own (`test_a_source_that_can_deny_nothing_has_no_resolver`) rather than resting on the two tests that only assert the positive case.

apps/api/v2/pipeline_edit/tests/test_node_create.py
```diff
@@ -0,0 +1,247 @@
...
+        )
+
+        assert response.status_code == 201, response.content
+        node_id = response.json()["node"]["node_id"]
+        assert "tempreture" not in response.json()["node"]["params"]
+        assert "tempreture" not in Node.objects.get(pipeline=chatbot.pipeline, flow_id=node_id).params
+
+
+@pytest.mark.django_db()
+class TestReportedRatherThanRefused:
```
SmittieC, line 211: This follows the UI's behaviour where we allow structural invalid states while the person is editing. Allowing this in the API gives us a "drafting" capability where the client can make edits and at aany time come back (without previous context) to finish up, which is why I opted to keep the same behaviour.

apps/api/tests/test_permissions.py
```diff
@@ -67,6 +71,64 @@ def test_api_key_views_keep_the_read_only_gate(view):
...
+        )
+
+    @pytest.mark.parametrize("has_perms", [True, False], ids=["holder-allowed", "non-holder-denied"])
+    def test_requires_team_permission_follows_the_users_permissions(self, has_perms):
+        assert _NeedsChange().has_permission(_request(has_perms=has_perms), view=None) is has_perms
+
+    def test_requires_team_permission_refuses_a_subclass_that_declares_nothing(self):
+        """`has_perms([])` is `all([])`, so a forgotten declaration would silently admit everyone. It
+        fires when the class is defined rather than when its endpoint is first called: an open door that
+        only shows itself under traffic is one that ships.
```
snopoke, line 116:
```suggestion
        """`has_perms([])` is `all([])`, so a forgotten declaration would silently admit everyone. It
        fires when the class is defined rather than when its endpoint is first called.
```

apps/api/v2/discovery/contract.py
```diff
@@ -15,6 +17,31 @@
...
+    for candidate in (param, param.removesuffix("_ids"), param.removesuffix("_id")):
+        if candidate in OptionsSource.__members__:
+            return OptionsSource[candidate]
+    return None
+
+
+#: The option lists a team could be denied a value from, so a param drawing from one is checked
+#: before it is written. Keyed on the list, not the param: the list is what decides the permitted
+#: values. Derived from the resource mirror so a new resource relation on ``Node`` needs no edit
+#: here; ``tools`` is named because nothing in the database backs it.
```
snopoke, line 40:
```suggestion
# Option sources for which we need to check provided values against the available values for a team. e.g. ``llm_provider``; ``tools`` is named because nothing in the database backs it.
```

apps/api/v2/discovery/node_types.py
```diff
@@ -32,11 +40,41 @@ def get_node_types() -> list[dict]:
...
 def option_keys_for_node_type(node_type: str) -> frozenset[str] | None:
     """The option keys a single node type reads, or ``None`` if no such type is served."""
     return _option_keys_by_type().get(node_type)
 
 
+def parameter_option_mapping(node_type: str) -> dict[str, OptionsSource]:
+    """One node type's reference params, each mapped to the option list it draws from.
+
+    Worth carrying around because a param is not always named for its list --
+    ``source_material_id`` draws from ``OptionsSource.source_material``.
```
snopoke, line 69:
```suggestion
```

apps/api/v2/pipeline_edit/examples.py
```diff
@@ -0,0 +1,161 @@
...
+                summary=f"{node_type}: {NOTES[node_type]}",
+                value={"type": node_type, "label": _label(node_type), "params": params},
+                request_only=True,
+            )
+            for node_type, params in FULL_PARAMS.items()
+        ),
+    ]
+
+
+def update_examples() -> list[OpenApiExample]:
```
snopoke, line 143: I wonder if we need the full examples here - in the schema it is basically duplicating the same thing as in create.

apps/api/v2/pipeline_edit/facade.py
```diff
@@ -0,0 +1,92 @@
...
+    since ``Pipeline.data`` no longer lists them, ADR-0049) and returns the edit to make. It runs
+    inside the lock, so what it reads is what gets written.
+
+    ``respond`` runs inside the lock too, so that a body that cannot be built takes the write down
+    with it: building it after the commit is how a node the server could not parse used to persist
+    and then 500 every later read of the pipeline.
+    """
+    with transaction.atomic():
+        chatbot = get_working_chatbot(request.team, public_id)
+        enforce_application_chatbot_write(request, chatbot)
```
snopoke, line 50: Do these need to be in the transaction?

snopoke (commented):
This was a tough review. I'm not done yet but some initial feedback - mostly very minor.

apps/api/v2/pipeline_edit/node_params.py
```diff
@@ -0,0 +1,73 @@
...
+        return False
+    return _is_list_annotation(field.annotation)
+
+
+def _is_writable(node_class: type[BasePipelineNode], name: str) -> bool:
+    field = node_class.model_fields.get(name)
+    if field is None:
+        return False
+    extra = field.json_schema_extra
+    return not (isinstance(extra, UiSchema) and extra.api_exclude)
```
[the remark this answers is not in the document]
SmittieC: Fixed in 4a9fd96 — `_is_writable` now rejects any field with `exclude` set, which is the suggested change verbatim. `node_id` and `django_node` are the only excluded fields across all eight served node types, so nothing settable is caught by it.

The finding is real, and I reproduced it end to end. Two corrections to the analysis, both of which made the regression test harder to write than the report implied:

1. **`Pipeline.validate` does not raise on its own.** `_node_validation_errors` and `_router_output_map` both build with `model_validate({**params, "node_id": ..., ...})` — a dict spread, so the explicit key wins and there is no collision. And `graph.build_runnable()` is staged behind `has_errors(report)`, which a poisoned node always trips (poisoning requires the validation-failure path), so it is never reached from `validate`.

2. **The `TypeError` needs the router to have an *outgoing* edge.** `_output_map_for` is the one caller that splats `**self.params`, and it is only asked for a router with conditional edges. With just `Start -> router` wired, `validate()` returns cleanly (`route_key: Field required`) and `/inspect/` answers 200 — I checked. Add `router -> target` on `output_0` and `validate()` raises `TypeError: got multiple values for keyword argument 'node_id'`, taking `/inspect/`, `/pipelines/<pk>/data` and every façade write with it.

So the wedge is real but one wiring step further out than described. Both tests are in `test_node_create.py::TestNodeId`: one asserts the internals never reach the row, the other wires the outgoing edge and asserts `/inspect/` still answers.

*— Claude Code*

apps/api/tests/test_permissions.py
SmittieC: Fixed in a94c84f — applied verbatim.

*— Claude Code*

apps/api/v2/discovery/node_types.py
SmittieC: Fixed in a94c84f — applied verbatim, leaving the one-line summary.

*— Claude Code*

apps/api/v2/discovery/contract.py
SmittieC: Fixed in a94c84f — applied as suggested, rewrapped across two lines to fit the 120-char limit (ruff would have flagged the single line at 168):

```python
# Option sources for which we need to check provided values against the available values for a team.
# e.g. ``llm_provider``; ``tools`` is named because nothing in the database backs it.
```

*— Claude Code*

apps/api/v2/pipeline_edit/facade.py
SmittieC: Fixed in 0781fba — no, and they are now outside it.

`get_working_chatbot` is a read and `enforce_application_chatbot_write` only reads too, so neither needs the transaction, and a 404 or a 403 no longer opens one. Nothing is lost by moving them out: the lock is on the `Pipeline` row rather than the `Experiment` (`get_working_chatbot` is called with `lock=False`), so being inside `atomic()` was never protecting the chatbot read from a concurrent archive — under READ COMMITTED it saw the same snapshot either way. `_locked_pipeline` restates `team_id=chatbot.team_id` under the lock, so the tenancy boundary still holds where the write happens.

*— Claude Code*

apps/api/v2/pipeline_edit/examples.py
SmittieC: Fixed in da94f62 — you were right, and it was worse than duplication.

The eight full PATCH payloads were the create ones with `type` dropped: 128 lines of near-identical YAML in `api-schemas/v2.yml`. And a body naming every param reads as the opposite of what an endpoint that merges key by key does, which the endpoint description then had to walk back ("for completeness, but a PATCH writes only the ones you send").

So they are gone, replaced by three partial bodies — one param, label only, router keywords — one per thing a PATCH actually does. The description now points at the POST examples for the full vocabulary.

One thing I kept: `test_a_documented_update_example_is_accepted` was the only assertion that every param of every type is settable *by PATCH* (merged onto a bare node, as opposed to written at create). That test now builds its body from `FULL_PARAMS` directly rather than from a published example, so the coverage survives the examples going away.

*— Claude Code*

apps/api/v2/pipeline_edit/graph_editor.py
```diff
@@ -0,0 +1,326 @@
...
+    The move carries content because that is the only node diff ``apply_pipeline_patch`` writes
+    position columns for, with params narrowed as an edit's are (see :func:`stored_params`).
+    """
+    moved = []
+    for stored in flow.get("nodes", []):
+        node = FlowNode(**stored)
+        if not _is_overtaken_end_node(node, new_node_x):
+            continue
+        content = cast(FlowNodeData, node.data)
+        content.params = stored_params(content)
```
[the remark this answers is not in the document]
SmittieC: Don't think this is an issue, since by now we have already resolved the node type (and raise a 404 if not found)

snopoke (approved):

Issue #4254, opened 2026-08-21
Remove the OpenAI assistant feature


snopoke:
Following the deprecation of the feature from OpenAI, the full shutdown is due for Aug 26 2026.

The removal from OCS should be done in two phases:

## Feature removal

Remove all views, UI, and logic related to assistants except for the following:

* Data models
* Django admin configuration

Pipelines that contain an assistant node should still be able to load in the UI though it should display that node as a 'Remove Node' with the previous node type. It should not be possible to build a pipeline in that state.

## Data removal after cool off period

After 30 days:

* Remove datamodels
* Remove django admin config

lisa-tarbo:
And remove documentation in 2nd phase: 

There seems to be sections about Assistants in multiple places
https://docs.openchatstudio.com/concepts/assistants/
https://docs.openchatstudio.com/concepts/experiment/
https://docs.openchatstudio.com/how-to/add_a_knowledge_base/?h=assistan#assistant
https://docs.openchatstudio.com/how-to/assistants_migration/
https://docs.openchatstudio.com/concepts/pipelines/?h=assistan#asking-the-assistant-about-a-pipeline
https://docs.openchatstudio.com/concepts/versioning/?h=assistan#versioning-for-openai-assistants

PR #4305, opened 2026-08-27
Remove the OpenAI Assistants feature (phase 1)

pxwxnvermx:
### Product Description

The OpenAI Assistants feature is removed. OpenAI shut its Assistants API down on
26 August 2026, so the feature could no longer function regardless.

For teams that used it:

* The Assistants section, its pages and its team-nav entry are gone.
* A chatbot pipeline that still contains an Assistant node **opens normally** and shows
  that node as a **Removed Node**, with an explanation and the original node type. The
  pipeline can still be edited and saved, but it will not run until the node is deleted.
* Assistant-related chat attachment UI and citation/file-download links for assistant
  files are gone.

This is phase 1 of two. The underlying data models and their Django admin stay in place
for a 30-day cool-off period; phase 2 drops them.

### Technical Description

Scope follows #4254: remove views, UI and runtime logic, keep data models and admin.

**Removed node rendering.** `REMOVED_NODE_TYPES` serves a stub schema for removed types.
This is load-bearing rather than cosmetic: the editor does `nodeSchemas.get(data.type)!`,
so a legacy `AssistantNode` with no schema would crash the pipeline editor outright.
`resolve_node_class()` still returns `None`, so the pipeline reports "Unknown node type"
and cannot build — matching the issue's "should not be possible to build a pipeline in
that state". Note the node does **not** block *saving*; the editor POSTs and gets `errors`
back alongside a 200, so the warning text says the pipeline will not *run*.

**Breaking API change (v2).** `feat(api)!` — the `assistant` object is removed from
`GET /api/v2/chatbots/{id}/inspect/`, along with the `Assistant` and `AssistantNodeParams`
schema components. Scope is narrow: `to_representation` only ever emitted `assistant` for
nodes declaring `assistant_id`, so no other node type's output changes. A follow-up commit
suppresses `assistant_id` from the generic `params` bag as well — without it, legacy rows
would have leaked the internal `OpenAiAssistant` primary key to API consumers, which is
exactly what that filter exists to prevent. It needs a separate `_HIDDEN_PARAM_KEYS`
because `test_resource_param_fields_are_real_node_fields` pins every
`_RESOURCE_PARAM_KEYS` entry to a real pydantic field on some node type, and no node
declares `assistant_id` any more.

**`Attachment.upload_to_assistant` is kept, deliberately inert.** `Attachment` instances
are handed to user-authored Python nodes, and `restricted_write` permits attribute writes
on them, so a stored node doing `att.upload_to_assistant = True` would start raising if the
field vanished. It is excluded from `PYTHON_NODE_HELP_PROMPT` so it is not advertised going
forward, and goes away with phase 2.

**`CustomActionOperation`.** A node is now the only version holder. Assistant-attached
operations deliberately do **not** block `CustomAction` deletion: there is no assistants UI
left to clear the reference from, so blocking would leave the action permanently
undeletable. The trade-off — deleting the action CASCADEs those rows away before phase 2
reaches them — is recorded in the docstring rather than left looking accidental.

**Also:** `check_llm_model_usage --include-archived` was left dead by the removal (the
assistant counts were its only consumer) and is now functional against pipeline nodes;
several helpers left with no callers are deleted.

### Demo

To exercise the removed-node path against an existing pipeline that has an Assistant node:

1. Open the chatbot's pipeline in the editor — it loads, with the node shown as
   **Removed Node** and a warning triangle carrying the explanation.
2. Edit an unrelated node and save — the save succeeds.
3. Try to run the chatbot — it fails with "Unknown node type" until the node is deleted.

### Docs and Changelog

Notes for the changelog:

* User-facing: the OpenAI Assistants feature is removed, following OpenAI's own shutdown of
  the Assistants API on 26 August 2026. Pipelines containing an Assistant node still open,
  but must have that node removed before they will run.
* **Breaking, API consumers:** the `assistant` object is no longer returned by
  `GET /api/v2/chatbots/{id}/inspect/`, and the `Assistant` / `AssistantNodeParams` schema
  components are removed from the v2 OpenAPI schema. Only assistant nodes are affected; no
  other node type's output changes.
* This is phase 1: data models and Django admin remain for a 30-day cool-off period.

Refs #4254


pxwxnvermx:
Closing this PR. Going to create multiple smaller PRs to make these changes more reviewable.

PR #4308, opened 2026-08-27
Diff chatbot versions on the details view

### LINKED ISSUE #4207: Allow users to see the 'diff' for chatbot versions
The details view for a chatbot version is not very useful as it does not show what changed between this version and the previous one. It also does not show any details about the pipeline itself.

Ideally this view should display the diff of this version with the previous version (in the same way that we do on the 'create new version' page.

It would also be great to be able to view a diff between two arbitrary chatbot versions.

preston-56:
### Product Description

The version details view now shows what changed since the previous version, the way the create-version page already does. A **Compare with** selector picks any earlier version to diff against.

resolves #4207.

### Technical Description

`versions/compare.html` was already in the template, but nothing called `VersionDetails.compare()`, so every field showed as unchanged.

The selector lists earlier versions only. That keeps the diff oldest-to-newest, and any pair is still reachable from the newer version.

### Demo

Defaults to the previous version; any earlier one can be picked.

Compare with selector open, listing v2 and v1

Comparing v3 against v1, skipping v2.

Prompt diff between v3 and v1

v1 has no predecessor. No selector is shown.

v1 details with no comparison selector

### Docs and Changelog

---

The issue also mentions the view not showing details about the pipeline itself. That group was there before but collapsed since `compare.html` only opens a group when it has changed fields. It now expands and diffs like the rest, so this may already be covered. Let me know if you meant something more. @snopoke


apps/chatbots/views.py
```diff
@@ -564,7 +564,42 @@ def chatbot_version_details(request, team_slug: str, experiment_id: int, version
...
+    earlier_versions = list(
+        Experiment.objects.get_all()
+        .filter(working_version_id=experiment_id, version_number__lt=version_number)
+        .order_by("-version_number")
+    )
+
+    requested = request.GET.get("compare_to")
+    if requested:
+        try:
+            requested_version_number = int(requested)
```
SmittieC, line 578: If `requested` is not a valid integer, then the query below will throw a 404 anyways, so perhaps this check can be dropped (I'll leave it to you to decide)? The downside of dropping it is that it will essentially take a DB query to figure out whether or not that is a valid integer, but I doubt that the requested version will ever be invalid, since we control this.

SmittieC (commented):
Thanks for the contribution! Generally looks good to me.

apps/chatbots/views.py
preston-56: Good catch, though it throws a 500 rather than a 404 here. `get_object_or_404` doesn't catch the ValueError Django raises when it tries to coerce a non-integer for `version_number`. I think the lookup above is safe because `<int:version_number>` in the URL pattern coerces it first and `?compare_to=` has no converter, so the raw string reaches the field. Happy to drop it if you'd still rather.

SmittieC: Nah it's ok. Thanks!

SmittieC (approved):

apps/chatbots/views.py
```diff
@@ -564,7 +564,42 @@ def chatbot_version_details(request, team_slug: str, experiment_id: int, version
     except Experiment.DoesNotExist:
         raise Http404() from None
 
-    context = {"version_details": experiment_version.version_details, "experiment": experiment_version}
+    # Earlier versions only. The diff then always reads oldest-to-newest, and any
+    # pair stays reachable by opening the newer of the two.
+    earlier_versions = list(
+        Experiment.objects.get_all()
```
SmittieC, line 570: Why call `.get_all()` here and not just `Experiment.objects.all()`?

preston-56: `.all()` filters out archived experiments, so an archived earlier version would disappear from the selector. `get_all()` skips that, matching the other lookups in this view.

SmittieC: The thing I was wondering about is if we want to present archived versions as well here, but I can see a usecase for that.

SmittieC:
From the [issue](https://github.com/dimagi/open-chat-studio/issues/4207):

> The details view for a chatbot version is not very useful as it does not show what changed between this version and the previous one. It also does not show any details about the pipeline itself.

@snopoke did you have anything specific in mind that should change on the version display modal to show more pipeline details?

snopoke:
> From the [issue](https://github.com/dimagi/open-chat-studio/issues/4207):
> 
> > The details view for a chatbot version is not very useful as it does not show what changed between this version and the previous one. It also does not show any details about the pipeline itself.
> 
> @snopoke did you have anything specific in mind that should change on the version display modal to show more pipeline details?

No, when the diff is rendered it is fine. The issue was with the normal, non-diff view which only shows the pipeline name. Fine to leave that as it is for now.

PR #4311, opened 2026-08-28
Cache the Meta provider's WhatsApp template and numbers, and use them across the UI

SmittieC:
### Product Description

Creating a Meta Cloud API messaging provider now kicks off a background check against the Graph API: whether the WhatsApp Business Account has a usable `new_bot_message` template, and which WhatsApp numbers the account owns, each with its phone number ID. All of it is cached in the provider's `extra_data`. The provider page reports what it found and offers a form to send a real test message through the template, with a Refresh button that re-runs both checks on demand. Everything downstream reads that cache rather than calling Meta: adding a WhatsApp channel offers the account's numbers as a picker instead of a free-text box, and the broadcast dialog warns before sending when a ticked channel's provider has no usable template — or when the message contains line breaks, which Meta flattens inside a template parameter.

Nothing in production has this cache yet and there is no backfill, so the whole thing fills itself in. A provider with no cached numbers keeps the free-text number box it has today; saving that channel syncs the provider first and matches what was typed against what Meta returned, which leaves the cache populated — so from then on that provider offers a picker. The two ways a provider gets populated are saving a channel and clicking Refresh on its page. A provider nobody has checked says exactly that rather than claiming its template is broken, and existing channels keep working untouched. Twilio and Turn.io channels are unaffected throughout.

### Technical Description

- `MessagingProvider.extra_data` gains two keys: `whatsapp_numbers` (`state`, `synced_at`, `error`, `numbers[]`) and `whatsapp_template` (`ok`, `checked_at`, `problems`, `error`, `template`). Both are written through a `select_for_update` helper so they never clobber each other or the webhook verify-token hash that already lives there.
- `sync_whatsapp_provider_task` fetches the numbers and re-checks the template in one pass, recording each independently so a failure in one does not lose the other. It is queued from `run_post_create_hook` and from the provider page's Refresh button. Nothing else calls Meta — the provider page and the broadcast dialog both render from the cache.
- `check_message_template` is strict about what `send_template_message` actually needs: the template must exist, be `APPROVED`, be in the provider's configured language, and carry a `bot_message` body parameter. A failed call is reported in the panel rather than raised.
- Sends use the cached phone number ID directly, and Meta's response body is shown verbatim when it refuses.
- The broadcast dialog's channel checkboxes carry `data-platform`, `data-template-status` and `data-provider-url`; Alpine derives both warnings from whatever is ticked. Provider links are behind `change_messagingprovider`, since a chatbot admin can broadcast but would get a 403 on the provider page.
- Disabled channels are now listed in the broadcast dialog greyed out with a tooltip rather than omitted; they cannot be ticked, and `clean_channels` rejects one submitted anyway.

### Migrations

No migrations — both caches live inside the existing `MessagingProvider.extra_data` JSONField.

### Demo

Provider without the `new_bot_message` template configured
Screenshot from 2026-08-28 14-42-12

Provider with the `new_bot_message` template configured
Screenshot from 2026-08-28 14-42-57

Notice when you select the Whatsapp Channel and your broadcast message contains newlines which are not permitted by WA (`new_bot_message` is configured)
Screenshot from 2026-08-28 14-43-23

Notice when you select the Whatsapp Channel (`new_bot_message` not configured)
Screenshot from 2026-08-28 14-43-44

### Docs and Changelog

The WhatsApp how-to on docs.openchatstudio.com walks through typing the phone number in by hand; that step is now a picker for any provider whose numbers have been synced.


apps/service_providers/models.py
```diff
@@ -569,6 +570,12 @@ def platform_supported_provider_types(platform: ChannelPlatform) -> list["Messag
         return provider_types
 
 
+#: Key under ``MessagingProvider.extra_data`` holding the cached WhatsApp numbers.
```
SmittieC: remove these comments

apps/service_providers/views.py
```diff
@@ -557,3 +579,131 @@ def sync_voices(request, team_slug: str, provider_type: str, pk: int):
         log.exception("Failed to sync voices for provider %s", pk)
         messages.error(request, "Voice sync failed. Please check your API key and try again.")
     return redirect("service_providers:edit", team_slug=team_slug, provider_type=provider_type, pk=pk)
+
+
+def _get_meta_provider(request, pk: int) -> MessagingProvider:
```
SmittieC: Nit: it's weird having this indirection. Rather call `get_object_or_404` inside the views itself

templates/channels/partials/whatsapp_channel_fields.html
```diff
@@ -0,0 +1,7 @@
+{% comment %}
```
SmittieC: remove comment

apps/channels/forms.py
```diff
@@ -252,7 +256,12 @@ def clean_bot_token(self):
         return bot_token
 
 
+NUMBER_NOT_FOUND = "{number} was not found at the provider. Please make sure it is there before proceeding"
```
SmittieC: move this to the top of the file

apps/channels/forms.py
```diff
@@ -275,25 +346,46 @@ def post_save(self, channel: ExperimentChannel):
     def clean(self):
         cleaned_data = super().clean()
         number = cleaned_data.get("number")
-        if not number or not self.messaging_provider:
+        provider = self.messaging_provider
+        if not number or not provider:
+            return cleaned_data
+
+        if provider.type == MessagingProviderType.meta_cloud_api:
+            self._clean_meta_number(cleaned_data, provider, number)
```
SmittieC: Hmm, this logic branches based on the provider type. I think we should still call the generic "serivce.resolve_number", but in there do what the `_clean_meta_number`` logic does. I don't want to pollute this form with provider specific logic.

templates/chatbots/components/broadcast_channel_option.html
```diff
@@ -0,0 +1,20 @@
+{% load i18n %}
+{% comment %}
+  One channel checkbox in the broadcast dialog.
```
SmittieC: remove comment

templates/chatbots/components/broadcast_dialog.html
```diff
@@ -1,7 +1,26 @@
 {% load form_tags i18n %}
+{% comment %}
```
SmittieC: remove comment

apps/service_providers/models.py
```diff
@@ -590,6 +597,117 @@ def type_enum(self):
...
+            "service_providers:edit",
+            kwargs={
+                "team_slug": get_slug_for_team(self.team_id),
+                "provider_type": const.MESSAGING,
+                "pk": self.id,
+            },
+        )
+
+    @property
+    def whatsapp_numbers_status(self) -> dict:
```
SmittieC: the word `status` makes one expect a single value, not a `dict`. Rename `whatsapp_numbers_status` to `whatsapp_numbers_info`. The same goes for `whatsapp_template_status`

apps/service_providers/messaging_service.py
```diff
@@ -1,3 +1,4 @@
+import dataclasses
```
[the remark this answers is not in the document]
SmittieC: Not addressed in this pass. The module is genuinely large, but bringing it under 600 lines means splitting the provider services into separate modules -- a refactor well beyond the scope of this PR, and one that would make this diff much harder to review. Tracking as follow-up debt rather than rushing it here.

*— Claude Code*

apps/service_providers/models.py
```diff
@@ -590,6 +597,117 @@ def type_enum(self):
...
 
+    def get_absolute_url(self) -> str:
+        return reverse(
+            "service_providers:edit",
+            kwargs={
+                "team_slug": get_slug_for_team(self.team_id),
+                "provider_type": const.MESSAGING,
+                "pk": self.id,
+            },
+        )
```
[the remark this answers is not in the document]
SmittieC: Not applicable -- checked the five flagged methods (`AuthProviderType.form_cls`, `MessagingProvider.get_absolute_url`, `MessagingProviderType.form_cls`, `VoiceProvider.add_file_url`). They are short, unrelated accessors that share a shape because they are model/enum boilerplate, not duplicated logic. Abstracting over them would add indirection without removing anything real.

*— Claude Code*

apps/channels/forms.py
```diff
@@ -10,6 +10,7 @@
 from django.conf import settings
 from django.contrib.postgres.forms import SimpleArrayField  # ty: ignore[unresolved-import]
 from django.core.exceptions import ValidationError
+from django.urls import reverse
```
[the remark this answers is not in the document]
SmittieC: Not addressed in this pass -- same root cause as the file-length finding on `messaging_service.py`: `channels/forms.py` holds a form per platform, so responsibilities scale with platforms supported. Splitting it per platform is a reasonable follow-up but too broad for this PR. Tracking as debt.

*— Claude Code*

apps/channels/forms.py
```diff
@@ -262,6 +271,68 @@ class WhatsappChannelForm(WebhookUrlFormBase):
...
+        options = {}
+        for provider in MessagingProvider.objects.filter(team=team, type=MessagingProviderType.meta_cloud_api):
+            numbers = [
+                {"value": number["number"], "label": whatsapp_number_label(number)}
+                for number in provider.whatsapp_numbers
+                if number.get("number")
+            ]
+            # Keep a number the channel already uses selectable, even if Meta no longer lists it.
+            is_channels_provider = bool(self.channel) and self.channel.messaging_provider_id == provider.id
+            if saved_number and is_channels_provider and saved_number not in {n["value"] for n in numbers}:
```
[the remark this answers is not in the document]
SmittieC: Not applicable -- the flagged conditional is a flat three-term guard (`saved_number and is_channels_provider and saved_number not in {...}`) that reads as one idea: keep a saved number selectable when the provider no longer lists it. Extracting it to clear a metric threshold would cost more readability than it buys.

*— Claude Code*

apps/channels/forms.py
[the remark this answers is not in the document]
SmittieC: Fixed in a364cab3c. Confirmed real: `ExtraFormBase.messaging_provider` resolved the submitted ID with no team filter, and `ChannelFormWrapper.is_valid()` validates both forms before either result is checked -- so a foreign Meta provider with an empty cache got synced (an outbound call plus a write to that provider's row) before the request was rejected.

The lookup is now scoped to `self.experiment.team`. Regression test `test_another_teams_provider_is_never_reached` submits a foreign provider ID through `ChannelFormWrapper` and asserts `sync_whatsapp_numbers()` is never called; it fails on the previous commit.

*— Claude Code*

apps/channels/forms.py
```diff
@@ -275,25 +346,46 @@ def post_save(self, channel: ExperimentChannel):
...
+            except Exception:
+                logger.exception("Could not fetch the WhatsApp numbers for provider %s", provider.id)
+                self.add_error("number", "Could not validate this number right now. Please try again.")
+                return
+
+        match = next((entry for entry in provider.whatsapp_numbers if entry["number"] == number), None)
+        if not match:
+            self.add_error("number", NUMBER_NOT_FOUND.format(number=number))
+            return
+        cleaned_data["phone_number_id"] = match["phone_number_id"]
```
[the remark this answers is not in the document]
SmittieC: Fixed in 6aad68065. Confirmed real -- `_numbers_by_provider` deliberately keeps a saved number selectable after Meta stops listing it, and validation then rejected that same number, so an edit could never be saved again (not even a rename).

Took a slightly broader fix than the suggested diff: `_unchanged_number_config()` short-circuits validation whenever the channel is being edited and both the provider and the number are unchanged, preserving the stored `phone_number_id`. It was resolved when it was saved, so re-checking it only creates a dead end. Two tests cover it: the unchanged stale number now saves, and changing the number is still checked against the provider.

*— Claude Code*

apps/service_providers/forms.py
```diff
@@ -361,12 +363,51 @@ def clean_template_language_code(self):
     def save(self, instance):
         instance = super().save(instance)
         verify_token = self.cleaned_data["verify_token"]
-        instance.extra_data = {
-            "verify_token_hash": hashlib.sha256(verify_token.encode()).hexdigest(),
-        }
+        # Merge: extra_data also holds this provider's cached WhatsApp numbers.
+        extra_data = instance.extra_data or {}
+        extra_data["verify_token_hash"] = hashlib.sha256(verify_token.encode()).hexdigest()
+        instance.extra_data = extra_data
```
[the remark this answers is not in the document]
SmittieC: Fixed in 51f958e32. Confirmed the gap: `_update_extra_data()` already locks and re-reads for the sync and template-check paths, but `MetaCloudAPIMessagingConfigForm.save()` merged onto the in-memory `instance.extra_data`, which the view then wrote back wholesale.

`save()` now re-reads `extra_data` under `select_for_update()` before merging the token hash. The caller (`_save_provider`) saves inside the same transaction, so the lock covers the write. Regression test `test_saving_the_config_form_keeps_numbers_synced_while_the_form_was_open` holds a stale instance, commits a sync from a second instance, then saves the form -- it drops the synced numbers on the previous commit and passes now.

*— Claude Code*

apps/service_providers/forms.py
```diff
@@ -361,12 +363,51 @@ def clean_template_language_code(self):
...
+        self.fields["from_number_id"].choices = [
+            (number["phone_number_id"], whatsapp_number_label(number)) for number in numbers
+        ]
+
+    def clean_to_number(self):
+        try:
+            number = phonenumbers.parse(self.cleaned_data["to_number"])
+        except phonenumbers.NumberParseException:
+            raise forms.ValidationError("Enter a valid phone number (e.g. +12125552368).") from None
+        return phonenumbers.format_number(number, phonenumbers.PhoneNumberFormat.E164)
```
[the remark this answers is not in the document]
SmittieC: Fixed in 771e6240b. `clean_to_number` now rejects the number when `phonenumbers.is_valid_number()` is false, so an unreachable number is a field error instead of a Meta send failure.

Also applied the same fix to `WhatsappChannelForm.clean_number` in `apps/channels/forms.py`, which had the identical parse-without-validate pattern. Both are covered by parametrized tests using numbers that parse but are not dialable (`+1234`, `+10000000000`).

*— Claude Code*

apps/service_providers/models.py
SmittieC: Fixed in c9a4d1199 -- dropped the `#:` comments above `WHATSAPP_NUMBERS_KEY` and `WHATSAPP_TEMPLATE_KEY`.

*— Claude Code*

templates/channels/partials/whatsapp_channel_fields.html
SmittieC: Fixed in c9a4d1199.

*— Claude Code*

apps/channels/forms.py
SmittieC: Fixed in b02bbf872 -- `NUMBER_NOT_FOUND` now sits just below the logger at the top of the module.

*— Claude Code*

templates/chatbots/components/broadcast_channel_option.html
SmittieC: Fixed in c9a4d1199.

*— Claude Code*

templates/chatbots/components/broadcast_dialog.html
SmittieC: Fixed in c9a4d1199.

*— Claude Code*

apps/service_providers/models.py
SmittieC: Fixed in 916381d03 -- renamed to `whatsapp_numbers_info` and `whatsapp_template_info`, along with the local variables in `views.py` that mirrored them.

One thing worth knowing: `apps/chatbots/forms.py` has a module-level function also called `whatsapp_template_status`. I left it alone -- it returns a single string (`"ok"` / `"problem"` / `"missing"`), so "status" is accurate there, and it reads `whatsapp_template_ok` rather than the renamed property.

*— Claude Code*

apps/service_providers/views.py
SmittieC: Fixed in 07cbb56be -- dropped `_get_meta_provider` and inlined `get_object_or_404(...)` into the three views. Reads consistently with the other provider lookups in the module now.

*— Claude Code*

apps/channels/forms.py
SmittieC: Fixed in ae27bd9cc -- the form no longer branches on provider type. `clean()` now always calls `provider.resolve_number(number)` and merges the result into `cleaned_data`; `_clean_meta_number` is gone.

One deviation from your suggestion worth flagging: I put `resolve_number` on `MessagingProvider` rather than on the service. The Meta numbers are cached on the provider row, and `get_messaging_service()` builds the service from `config` alone, so a service-level implementation cannot see the cache -- it would have to call Meta on every validation. That would undo the caching this PR adds, and `test_a_cached_number_resolves_its_id_without_calling_meta` pins it (it asserts `get_messaging_service` is never called on a warm cache).

So the dispatch sits on the model, which owns the cache: `MessagingProvider.resolve_number()` delegates to the service for non-Meta providers and to `_resolve_meta_number()` (cache first, sync if empty) for Meta. It returns the channel config the number implies -- `{"number": ...}` plus `phone_number_id` for Meta -- or `None` when the provider does not have the number. The form stays provider-agnostic either way; happy to move it onto the service if you would rather take the per-validation API call.

*— Claude Code*

apps/service_providers/models.py
```diff
@@ -7,6 +7,7 @@
 from django.core.exceptions import ValidationError as DjangoValidationError
 from django.db import IntegrityError, models, transaction
 from django.urls import reverse
+from django.utils import timezone
```
SmittieC, line 10: Not addressed. This one is new since the last review, and it is the `resolve_number` / `_resolve_meta_number` pair added in ae27bd9cc that tipped it over -- the module went from 74 to 76 functions.

The metric is measuring the module, not those two methods: `models.py` holds every provider model (LLM, messaging, voice, auth, tracing), so "10 responsibilities" is the file's existing shape at a threshold of 4. The two new methods belong on `MessagingProvider` -- they read the WhatsApp number cache stored on that row, which is the whole reason they cannot live on the service.

Bringing this under the threshold means splitting `models.py` per provider type, which is the same follow-up as the `messaging_service.py` and `channels/forms.py` findings on this PR. Not something to fold into this diff.

*— Claude Code*

apps/channels/forms.py
```diff
@@ -159,8 +165,14 @@ def __init__(self, experiment, channel=None, **kwargs):
 
     @cached_property
     def messaging_provider(self) -> MessagingProvider | None:
+        """The submitted provider, scoped to the experiment's team.
+
+        ChannelForm rejects a provider from another team on its own, but both forms are
+        validated before either is checked, so an unscoped lookup here would let a submitted
+        ID reach another team's provider and its service.
```
snopoke, line 172:
```suggestion
```

apps/service_providers/models.py
```diff
@@ -590,6 +595,153 @@ def type_enum(self):
...
+            "service_providers:edit",
+            kwargs={
+                "team_slug": get_slug_for_team(self.team_id),
+                "provider_type": const.MESSAGING,
+                "pk": self.id,
+            },
+        )
+
+    @property
+    def whatsapp_numbers_info(self) -> dict:
```
snopoke, line 609: I think it would be good to move these to a mixin.

snopoke (dismissed):
No major comments. LGTM

apps/service_providers/models.py
SmittieC: Done — moved to `WhatsAppProviderMixin` in `apps/service_providers/whatsapp.py`: the numbers/template/refresh caches and their accessors, plus the three `WHATSAPP_*_KEY` constants. `MessagingProvider` keeps `get_absolute_url`, `resolve_number` (the type dispatch), `_update_extra_data` (shared with the verify-token write) and `run_post_create_hook`.

—Claude Code

apps/service_providers/forms.py
```diff
@@ -361,12 +364,66 @@ def clean_template_language_code(self):
...
+class WhatsappTestMessageForm(forms.Form):
+    """Sends one template message through a Meta Cloud API provider to see if Meta accepts it."""
+
+    from_number_id = forms.ChoiceField(label=_("From"))
+    to_number = forms.CharField(
+        label=_("To"),
+        # A placeholder rather than help text: the help icon makes this label taller than the
+        # "From" label beside it, which knocks the two fields out of alignment.
+        widget=forms.TextInput(attrs={"placeholder": "+27812345678"}),
+    )
```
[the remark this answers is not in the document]
SmittieC: Not a bug. If the numbers cache changes between GET and POST, the form should re-validate against the current cache: sending from a `phone_number_id` the provider no longer has would just fail at the Meta API instead. Caching the GET-time choices in the session would let a submission through for a number that's already gone. Leaving as-is.

—Claude Code

apps/service_providers/models.py
SmittieC: Addressed as part of the mixin extraction below (this PR) — the WhatsApp cache accessors moved out of `models.py` into `apps/service_providers/whatsapp.py`, which should reduce the responsibility count here.

—Claude Code

snopoke (approved):

PR #4313, opened 2026-08-29
feat(teams): redesign the team settings page

### LINKED ISSUE #4210: Redesign the team settings page
Mockups from Claude design (https://claude.ai/design/p/a1ec46c3-62b5-4748-acc3-417c8febdbbd?file=Team+Admin.dc.html&via=share)

Standalone HTML mockup: [Team Admin.html](https://github.com/user-attachments/files/31172097/Team.Admin.html)

Image

Image

Image

preston-56:
### Product Description

Redesigns Team Settings to match the #4210 mockup: one page with a section nav (Integrations / Members / Developer / Data), merged Integrations table (LLM, Speech, Messaging, Auth, Tracing in one filterable table), merged Members table (active members + pending invites, searchable/filterable), inline team-name editing, and a single `Save key` button covering public key + migration mode.

Closes #4210

### Technical Description

- New: `integrations_views.py`, `members_views.py`, `tables.py` (merged row-building + tables)
- `ServiceProviderType` gained a `category` field for the pills/column
- `set_public_key` now saves the migration-mode form too (one button, one request)
- Removed `service_provider_home.html`, `team_invitations.html`, `invitation_row.html` (superseded by the merged tables, no remaining references)
- `object_home_content.html` and `app_nav_menu_items.html`: small backward-compatible additions (button label param, sidebar link)
- Status badge is a hardcoded "Connected" for now. No health-check infra exists yet
- Stats card counts refresh on page load only, not after htmx actions

### Migrations

No migrations in this PR.

### Demo

Integrations table, category pills, and the merged Members table (search/role/status filters):

01-integrations

Members table with a live invite showing the dashed avatar and Invited status:

02-members

Developer access: Custom actions and OAuth applications together:

03-developer

Data & migration: the single "Save key" button, and Danger Zone below it:

04-data

### Round 2: addressing review feedback

Dark mode fixed (was rendering half-light-half-dark before):

01-dark-mode-fixed

02-light-mode

Scroll-spy nav highlighting fixed (was activating the wrong section):

07-scrollspy-scrolling

Add integration dropdown grouped by category:

03-integration-dropdown-grouped

Invite people is now a real dialog, including fixing a latent bug where a validation error used to silently close the form:

04-invite-dialog

05-invite-dialog-error-state

08-invite-dialog-reopens-on-error

Checkbox spacing fixed globally, not just on this page:

06-oauth-checkbox-spacing

### Docs and Changelog

### Operator Impact

UI-only change. No new settings, no migration, no config change required.


templates/oauth2_provider/application_form.html
```diff
@@ -1,6 +1,17 @@
 {% extends "generic/object_form.html" %}
 {% load form_tags i18n %}
 
+{% block page_head %}
+  {{ block.super }}
+  <style>
+    /* CheckboxSelectMultiple falls through to the plain text-field template, which renders
+       each option as an unspaced div. Space them out for the "Allowed chatbots" list. */
```
snopoke: Is this an issue everywhere we use this widget or just on this page? If it's an issue everywhere we should look for a global fix e.g. override the default Django template.

apps/teams/views/members_views.py
```diff
@@ -0,0 +1,111 @@
...
+    table_class = MembersTable
+    template_name = "teams/components/members_table.html"
+
+    def get_queryset(self):
+        rows = get_member_rows(self.request.team)
+        return filter_member_rows(rows, self.request.GET)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["total_count"] = len(get_member_rows(self.request.team))
```
snopoke: This calls `get_member_rows` a second time. Either cache the result or override some other methods to avoid the double call.

apps/teams/views/members_views.py
```diff
@@ -0,0 +1,111 @@
...
+    display_name = membership.user.get_display_name()
+    return {
+        "id": f"member-{membership.pk}",
+        "kind": "member",
+        "membership_pk": membership.pk,
+        "name": display_name,
+        "email": membership.user.email,
+        "initials": _initials(display_name),
+        "role_list": [group.name for group in groups],
+        "role_names": {group.name for group in groups},
```
snopoke: These fields have identical values (same in the invitation row values)

templates/teams/manage_team.html
```diff
@@ -2,176 +2,355 @@
...
+  <link href="https://fonts.googleapis.com/css2?family=Plus+Jakarta+Sans:wght@400;500;600;700&display=swap" rel="stylesheet">
   <style>
     /* Adds fade out effect when invitations are canceled */
     tr.htmx-swapping td {
       opacity: 0;
       transition: opacity 1s ease-out;
     }
+
+    #team-settings-page {
+      font-family: 'Plus Jakarta Sans', ui-sans-serif, system-ui, sans-serif;
```
snopoke: We shouldn't be using a different from from the rest of the site

templates/teams/manage_team.html
snopoke, line 1: I notice that there is quite a lot of custom CSS here including static colors etc. 

We should be sticking to the DaisyUI styling as much as possible and avoiding custom CSS.

templates/teams/manage_team.html
```diff
@@ -2,176 +2,355 @@
...
+      font-size: 14px;
+      font-weight: 700;
+    }
+    #automation h1.pg-subtitle {
+      font-size: 14px;
+      font-weight: 700;
+    }
   </style>
+  <script>
+    document.addEventListener('DOMContentLoaded', function () {
```
snopoke, line 108: This code doesn't really work very well. Try click on 'Members' and you'll notice that 'Developer' get's activated. 

If it's not easy to get it working well we should rather drop the activation for all together.

templates/teams/manage_team.html
```diff
@@ -2,176 +2,355 @@
...
-                   class="btn btn-outline"
-                   href="{% url 'files:base' team.slug files_export_file.id %}?allow_s3">
-                  {% translate "Download Export" %}
-                </a>
-                <button type="submit" class="btn btn-outline">{% translate "Regenerate export" %}</button>
-              {% else %}
-                <button type="submit" class="btn btn-outline">{% translate "Download team files" %}</button>
-              {% endif %}
+        <section class="app-card" id="automation">
+          <h2 class="pg-subtitle font-bold">{% translate "Developer access" %}</h2>
```
snopoke:
```suggestion
          <h2 class="pg-subtitle font-bold">{% translate "Developers" %}</h2>
```

snopoke (commented):
I know this is still in draft but I took a quick pass and tried it out locally. In addition to the PR comments I also put together some UI feedback:

https://docs.google.com/document/d/1yQJtqClIJEEWxW3ts6_q6XSpMCX-L7XO-ELcHWChdj0/edit?usp=sharing

templates/oauth2_provider/application_form.html
preston-56: Yes, everywhere. Fixed it globally: added a case for `CheckboxSelectMultiple` in the `render_field` dispatch instead of patching each page, and removed this page's local style block.

apps/teams/views/members_views.py
preston-56: Fixed. Caching it from `get_queryset` and reusing that in `get_context_data` instead of calling it again.

apps/teams/views/members_views.py
preston-56: Dropped `role_names`. Using `role_list` for both the template loop and the filter check now.

templates/teams/manage_team.html
preston-56: Removed it.

templates/teams/manage_team.html
preston-56: > I notice that there is quite a lot of custom CSS here including static colors etc.
> We should be sticking to the DaisyUI styling as much as possible and avoiding custom CSS.

Converted the static colors to DaisyUI's own theme tokens `(var(--color-base-100)` etc.), so it's `dark-mode-safe` now. This was actually a real bug too: the card rendered `half-light-half-dark` before. Kept the distinct background/border treatment since that matches the mockup, just theme-aware instead of hardcoded.

preston-56: > If it's not easy to get it working well we should rather drop the activation for all together.

Fixed it rather than dropping it. Root cause was the `IntersectionObserver's` band being too thin for short sections. Switched to tracking scroll position directly against each section's offset, which handles sections of any height. Verified by scrolling the full page and clicking every nav link.

templates/teams/manage_team.html
preston-56: Applied, thanks.

preston-56:
> I know this is still in draft but I took a quick pass and tried it out locally. In addition to the PR comments I also put together some UI feedback:
> 
> https://docs.google.com/document/d/1yQJtqClIJEEWxW3ts6_q6XSpMCX-L7XO-ELcHWChdj0/edit?usp=sharing

Thanks for the heads-up!

apps/teams/views/integrations_views.py
```diff
@@ -0,0 +1,154 @@
...
+    def get_queryset(self):
+        rows = get_integration_rows(self.request, self.request.team)
+        category = self.request.GET.get("category")
+        if category:
+            rows = [row for row in rows if row["category"] == category]
+        return rows
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        all_rows = get_integration_rows(self.request, self.request.team)
```
snopoke: here's another example of repeating the same query twice in the view

apps/teams/views/manage_team_views.py
```diff
@@ -160,22 +173,25 @@ def send_invitation_view(request, team_slug):
...
 
 @require_POST
 @permission_required("teams.change_team", raise_exception=True)
 def set_public_key(request, team_slug):
+    """Saves the public key and the migration-mode toggle together, matching the mockup's
+    single "Save key" action for the whole Migration public key card."""
     form = TeamPublicKeyForm(request.POST, instance=request.team)
-    if form.is_valid():
+    migration_form = TeamMigrationForm(request.POST, instance=request.team)
+    if form.is_valid() and migration_form.is_valid():
```
snopoke: What about combining these into a single form?

templates/teams/components/member_row_name.html
```diff
@@ -0,0 +1,11 @@
+<div class="flex items-center gap-2">
+  {% if record.kind == "invitation" %}
+    <span class="w-8 h-8 rounded-full border border-dashed border-neutral-400 text-neutral-400 flex items-center justify-center text-xs font-bold shrink-0">{{ record.initials }}</span>
```
snopoke: DaisyUI has an avatar placeholder component that we should use: https://daisyui.com/components/avatar/#avatar-placeholder

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
       opacity: 0;
       transition: opacity 1s ease-out;
     }
+
+    .xl\:container:has(#team-settings-page) {
+      max-width: none;
+      background: var(--color-base-200);
+    }
```
snopoke: This is what is giving this page a different width and BG color from the rest of the site. I do like the look but we should drop it from this PR and separately apply that to the site as a whole.

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+    #general.app-card, #integrations.app-card, #members.app-card,
+    #automation.app-card, #data.app-card,
+    #automation .app-card, #data .app-card,
+    #team-settings-nav .app-card {
+      border: 1px solid var(--color-base-300);
+      border-radius: 16px;
+      background: var(--color-base-100);
```
snopoke: I think we can apply this to the main `app-class` definition in `assets/styles/site-tailwind.css`

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+    }
+    #team-settings-nav .nav-count {
+      font-size: 13px;
+      font-weight: 600;
+      color: color-mix(in oklch, var(--color-base-content) 55%, transparent);
+    }
+    #team-settings-nav a.settings-nav-active .nav-count {
+      color: var(--color-primary);
+    }
+    #team-settings-page .category-pill {
```
snopoke: You can replace `category-pill` with normal button styling:

```
<button class="btn btn-xs btn-soft btn-primary rounded-full {% if pill.active %}btn-active{% endif %}" >
```

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+    #team-settings-page .section-subtitle,
+    #automation .app-card .text-neutral-500 {
+      margin-top: 0;
+      display: block;
+    }
+    #team-settings-page input[name="is_migrating"] {
+      border-radius: 4px;
+      width: 1.1rem;
+      height: 1.1rem;
+    }
```
snopoke: unnecessary styles

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+    #data label[for="id_public_key"] {
+      font-size: 12px;
+      font-weight: 600;
+    }
+    #data h3.pg-subtitle {
+      font-size: 14px;
+      font-weight: 700;
+    }
+    #automation h1.pg-subtitle {
```
snopoke, line 50: I was confused why this is `h1` but I see it comes from `object_home_content.html`. Maybe add a comment.

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+      padding: 1rem;
+    }
+    #data label[for="id_public_key"] {
+      font-size: 12px;
+      font-weight: 600;
+    }
+    #data h3.pg-subtitle {
+      font-size: 14px;
+      font-weight: 700;
+    }
```
snopoke: apply this directly to the elements with `font-bold text-sm`

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+      width: 1.1rem;
+      height: 1.1rem;
+    }
+    #team-settings-page .app-card {
+      padding: 1rem;
+    }
+    #data label[for="id_public_key"] {
+      font-size: 12px;
+      font-weight: 600;
+    }
```
snopoke: I didn't notice any difference when I removed this styling.

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+    #automation .app-card .text-neutral-500 {
+      margin-top: 0;
+      display: block;
+    }
+    #team-settings-page input[name="is_migrating"] {
+      border-radius: 4px;
+      width: 1.1rem;
+      height: 1.1rem;
+    }
+    #team-settings-page .app-card {
```
snopoke, line 44: add a comment
```suggestion
    #team-settings-page .app-card {
      /* compact padding for better fit in smaller area */
```

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+    }
+    #team-settings-page .section-subtitle {
+      font-size: 12px;
+      white-space: nowrap;
+    }
+    #team-settings-page h2.pg-subtitle,
+    #team-settings-page h3.pg-subtitle {
+      margin-bottom: 0;
+      font-size: 17px;
+    }
```
snopoke, line 43: this can be merged with `#team-settings-page .section-subtitle`

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+      border-color: var(--color-primary);
+      background: color-mix(in oklch, var(--color-primary) 12%, var(--color-base-100));
+      color: var(--color-primary);
+    }
+    #team-settings-page .category-pill-active {
+      border-color: var(--color-primary);
+      background: color-mix(in oklch, var(--color-primary) 12%, var(--color-base-100));
+      color: var(--color-primary);
+    }
+    #team-settings-page #id_name {
```
snopoke, line 26: Make a note that this is for the inline edit input element to keep the text styling matching the non-edit text.

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+      font-size: 1.875rem;
+      font-weight: 700;
+      letter-spacing: -0.02em;
+      height: auto;
+      padding: 0.25rem 0.75rem;
+    }
+    #automation .app-card .text-neutral-500 {
```
snopoke: if you add the `section-subtitle` class to the span in `object_home_content.html` you won't need two selectors to apply this style.

templates/teams/manage_team.html
```diff
@@ -8,170 +8,348 @@
...
+    #team-settings-page h2.pg-subtitle,
+    #team-settings-page h3.pg-subtitle {
+      margin-bottom: 0;
+      font-size: 17px;
+    }
+    #team-settings-page .section-subtitle,
+    #automation .app-card .text-neutral-500 {
+      margin-top: 0;
+      display: block;
+    }
```
snopoke: Merge this with the section-subtitle styling above

templates/teams/manage_team.html
snopoke, line 1: Can I suggest that you ask Claude or Codex to review the CSS here and help with cleaning it up. It should also move to `assets/styles/app/team-settings.css` and be imported into `assets/styles/site-tailwind.css` in the same way as `dashboard.css`

snopoke (commented):
I've tested this out and it is working better.

Most of my comments are now related to the styling which fall into two categories:
1. Custom styles / components that should be replaced with DaisyUI components e.g. category pills & avatars
2. Utility overrides

apps/teams/views/integrations_views.py
preston-56: Fixed the same way, caching the rows from `get_queryset` and reusing them in `get_context_data`.

apps/teams/views/manage_team_views.py
preston-56: Combined. `TeamPublicKeyForm` now includes `is_migrating` directly, and `set_public_key` validates and saves just the one form.

templates/teams/components/member_row_name.html
preston-56: Switched to DaisyUI's avatar-placeholder component for both states.

templates/teams/manage_team.html
preston-56: Dropped it from this PR

templates/teams/manage_team.html
preston-56: Moved the border, radius, and background into the real @utility app-card definition in site-tailwind.css. Checked it against a couple of other pages that use app-card, renders cleanly everywhere, not just here.

templates/teams/manage_team.html
preston-56: Replaced with the markup you gave, and dropped the `category-pill` CSS entirely.

templates/teams/manage_team.html
preston-56: removed

templates/teams/manage_team.html
preston-56: Added a comment explaining it comes from `object_home_content.html`.

templates/teams/manage_team.html
preston-56: Applied font-bold text-sm directly to the three h3 elements this was targeting, and dropped the CSS rule.

templates/teams/manage_team.html
preston-56: removed

templates/teams/manage_team.html
preston-56: added using the suggested wording

templates/teams/manage_team.html
preston-56: merged into one `.section-subtitle` rule

templates/teams/manage_team.html
preston-56: done

templates/teams/manage_team.html
preston-56: Added the class there. Down to one selector now instead of two.

templates/teams/manage_team.html
preston-56: Same fix as above, merged into that one .section-subtitle rule.

templates/teams/manage_team.html
preston-56: consider it done. thank you!

snopoke: This could do with the help text from before: 

"While enabled, scheduled messages and event triggers (including timeout triggers) for this team are paused and will not fire."

preston-56: Restored. Thank you!

snopoke (commented):
I had Claude go over this again and it found some actionable items:

**HIGH**

1. apps/teams/views/manage_team_views.py:187 — migration-mode checkbox can display a value that was never saved
TeamPublicKeyForm now covers both public_key and is_migrating, but set_public_key re-renders from request.team, which ModelForm._post_clean has already mutated in memory with the submitted is_migrating — even when clean_public_key rejected the key and nothing was saved. The dangerous direction: an admin ticks "Migration mode", pastes a malformed key, sees only "Could not save the public key", and the checkbox stays ticked — so they believe outbound messages/event triggers are frozen when they aren't. Fix: render the checkbox from the bound form, or don't let a key error swallow the toggle.

**MEDIUM**

2. templates/teams/components/members_table.html:2 — pagination nests a duplicate header inside the table wrapper
tailwind_js_pagination.html swaps closest div.table-container by outerHTML, so the response must be the container. These two views return a fragment that wraps it in extra markup. With 31 members, clicking "next" injects the "25 of 31" `<p>` and a second overflow-x-auto div inside the existing one, compounding per click. integrations_table.html:15 has the same shape and duplicates the whole category pill bar (with stale counts/active pill) past 25 integrations.

**LOW**

3. provider_form.html:8 — three links point at anchors deleted with service_provider_home.html (also service_providers/usages.html:7, prompt_builder_toolbox.html:83); retarget to #integrations.
4. status_badge.html:1 — dead status_kind == "invited" branch; integration rows never set it.
5. members_section.html:40 — the length-guard hx-trigger on the input is inert (no hx-* verb on that element); the form-level trigger fires on every keystroke.

preston-56:
> I had Claude go over this again and it found some actionable items:
> 
> **HIGH**
> 
> 1. apps/teams/views/manage_team_views.py:187 — migration-mode checkbox can display a value that was never saved
>    TeamPublicKeyForm now covers both public_key and is_migrating, but set_public_key re-renders from request.team, which ModelForm._post_clean has already mutated in memory with the submitted is_migrating — even when clean_public_key rejected the key and nothing was saved. The dangerous direction: an admin ticks "Migration mode", pastes a malformed key, sees only "Could not save the public key", and the checkbox stays ticked — so they believe outbound messages/event triggers are frozen when they aren't. Fix: render the checkbox from the bound form, or don't let a key error swallow the toggle.
> 
> **MEDIUM**
> 
> 2. templates/teams/components/members_table.html:2 — pagination nests a duplicate header inside the table wrapper
>    tailwind_js_pagination.html swaps closest div.table-container by outerHTML, so the response must be the container. These two views return a fragment that wraps it in extra markup. With 31 members, clicking "next" injects the "25 of 31" `<p>` and a second overflow-x-auto div inside the existing one, compounding per click. integrations_table.html:15 has the same shape and duplicates the whole category pill bar (with stale counts/active pill) past 25 integrations.
> 
> **LOW**
> 
> 3. provider_form.html:8 — three links point at anchors deleted with service_provider_home.html (also service_providers/usages.html:7, prompt_builder_toolbox.html:83); retarget to #integrations.
> 4. status_badge.html:1 — dead status_kind == "invited" branch; integration rows never set it.
> 5. members_section.html:40 — the length-guard hx-trigger on the input is inert (no hx-* verb on that element); the form-level trigger fires on every keystroke.

Thanks for the detailed pass, Simon. All five items are fixed and covered by new tests:

**HIGH**: `set_public_key` now refreshes the team from the database when the form is invalid, so a rejected public key no longer leaves the migration-mode checkbox showing a value that was never saved.

**MEDIUM**: Both `MembersTableView` and `IntegrationsTableView` now return the bare table fragment on pagination/sort requests, matching the outerHTML-swap contract the pagination links expect, so clicking through pages no longer nests duplicate wrappers.

**LOW**: Retargeted the three dead anchor links to `#integrations`, removed the dead `status_kind == "invited"` branch, and moved the `members-search` length guard onto the form's own trigger (it was declared on the input, which has no `hx-get` of its own, so it never actually fired).

While verifying the last one live I also found the role/status dropdowns were sizing to their longest option text and squeezing the search input down to about 66px, so I capped their widths too.

Pushed in five separate commits, one per item, for easier review.

templates/teams/manage_team.html
```diff
@@ -8,170 +8,301 @@
...
+        {% endif %}
+        <label for="edit-team-toggle" class="btn btn-outline whitespace-nowrap">{% translate "Cancel" %}</label>
+      </form>
+      <div class="flex flex-wrap gap-2">
+        <a href="{% url "single_team:feature_flags" request.team.slug %}"
+           class="btn btn-outline whitespace-nowrap">
+          <i class="fa-solid fa-flag mr-1 text-neutral-400"></i>{% translate "Feature flags" %}</a>
+        {% if request.user.is_staff %}
+          <a href="{% url "single_team:internal_metadata" request.team.slug %}"
+             class="btn btn-outline whitespace-nowrap">
```
SmittieC:
```suggestion
             class="btn btn-sm btn-outline whitespace-nowrap">
```

templates/teams/manage_team.html
```diff
@@ -8,170 +8,301 @@
...
+        {% csrf_token %}
+        {{ team_form.name }}
+        {% if request.team_membership.is_team_admin %}
+          <input class="btn btn-primary whitespace-nowrap" type="submit" value="{% translate "Save" %}">
+        {% endif %}
+        <label for="edit-team-toggle" class="btn btn-outline whitespace-nowrap">{% translate "Cancel" %}</label>
+      </form>
+      <div class="flex flex-wrap gap-2">
+        <a href="{% url "single_team:feature_flags" request.team.slug %}"
+           class="btn btn-outline whitespace-nowrap">
```
SmittieC:
```suggestion
           class="btn btn-sm btn-outline whitespace-nowrap">
```

apps/teams/tables.py
```diff
@@ -0,0 +1,56 @@
...
+    if record["kind"] == "mcp":
+        return reverse("mcp_integrations:delete", args=[request.team.slug, record["pk"]])
+    return reverse("service_providers:delete", args=[request.team.slug, record["provider_type"], record["pk"]])
+
+
+class IntegrationsTable(Table):
+    name = TemplateColumn(template_name="teams/components/integration_row_name.html", verbose_name="Name")
+    category = Column(verbose_name="Category")
+    provider = Column(verbose_name="Provider")
+    status = TemplateColumn(template_name="teams/components/status_badge.html", verbose_name="Status", orderable=False)
```
SmittieC: Forcing the status to be "connected" by default is a bit misleading. What is the credentials are wrong? I think we should defer the status field to a followup where we can reliably determine the correct status.

FYI in [this PR](https://github.com/dimagi/open-chat-studio/pull/4373) we determine credential validity which can be used for this status, albeit only for LLM providers. Perhaps we can / should do it for the others as well. This will need some investigation

templates/teams/components/settings_nav.html
```diff
@@ -0,0 +1,23 @@
+{% load i18n %}
+<nav class="flex flex-col gap-1 mb-4">
+  <a href="#integrations" class="flex justify-between items-center rounded-lg px-3 py-2 text-sm font-semibold hover:bg-base-200">
```
SmittieC, line 3: A thought:  the settings page feels cluttered with all tabs below each other. I quite like how github's settings page works where the sidebar dictates what gets rendered inside the main space. So in this case, only one of Integrations, Members, Developer or Data gets rendered, not all at once. Thougths? cc @snopoke

templates/teams/manage_team.html
```diff
@@ -8,170 +8,301 @@
...
+          <i class="fa-solid fa-flag mr-1 text-neutral-400"></i>{% translate "Feature flags" %}</a>
```
SmittieC: can we make `Feature Flags` and `Internal metadata` sidenar tabs as well?

apps/teams/tests/test_integrations_table.py
```diff
@@ -0,0 +1,143 @@
...
+        request = RequestFactory().get("/")
+        request.team = team
+        request.user = user
+        return request
+
+    return _make
+
+
+@pytest.mark.django_db()
+def test_get_integration_rows_composite_ids_are_prefixed_by_provider_type(team, request_for):
```
SmittieC, line 31: nit: would you mind adding a bit more context in this docstring?

apps/teams/tables.py
preston-56: > Forcing the status to be "connected" by default is a bit misleading. What is the credentials are wrong? I think we should defer the status field to a followup where we can reliably determine the correct status.
> 
> FYI in [this PR](https://github.com/dimagi/open-chat-studio/pull/4373) we determine credential validity which can be used for this status, albeit only for LLM providers. Perhaps we can / should do it for the others as well. This will need some investigation

Agreed, and this was flagged as a known gap in the PR description (`"Status badge is a hardcoded Connected for now. No health-check infra exists yet"`). I'd rather not build real per-provider connectivity checks into this redesign PR though, since that's a genuinely separate piece of work (different check logic per provider category, plus a caching/refresh strategy so we're not hitting 5+ external services on every table render). 

I've already scoped this as its own follow-up ticket. Happy to file it now if that resolves this for the current PR, or fold your LLM-credential-validity work in as the starting point for it. Thanks for the heads-up Smit!

templates/teams/manage_team.html
preston-56: Added, they're now in the sidebar alongside Integrations/Members/Developer/Data.

apps/teams/tests/test_integrations_table.py
preston-56: Done, expanded it to explain why the composite `id` format matters (different provider models share independent `pk` sequences, so the type prefix is what keeps row `ids` unique across the merged table, not just cosmetic).

templates/teams/components/settings_nav.html
preston-56: My own take is that this doesn't need to block the current PR, though that's just my read and not a final call since Simon's weighing in too.

Worth naming the real cost either way: only rendering one of `Integrations`, `Members`, `Developer` or `Data` at a time means the page has to track which section is active across a reload, probably via a URL fragment or query param, and each section's own htmx and pagination logic currently assumes it's already sitting in the DOM. Not a huge lift, but it's a second pass on top of what's already here, not a small tweak to it.

Given that, I'd lean toward merging this PR as-is and picking up the single-panel rendering as a fast follow-up, but if Simon would rather hold this PR for it, that works for me too.

apps/teams/tables.py
SmittieC: It's fine to do it in a followup, but I do think we should exclude the status field as a whole for this PR

templates/teams/components/settings_nav.html
snopoke: I'm fine with that though I'd prefer to push it to a new PR

snopoke (dismissed):

templates/teams/components/settings_nav.html
preston-56: Sounds good. I'll leave the follow-up ticket to you two.

apps/teams/tables.py
preston-56: Fair enough, I'll pull the status column out of this PR entirely and pick it up properly in the followup once we've worked out how to determine it for real.

preston-56:
@snopoke this had your approval before, but pushing the `nav/docstring/status-column` changes above reset it per branch protection. Could you take another look when you get a chance? The diff since your approval is just those three commits, nothing else changed.

snopoke (approved):

PR #4317, opened 2026-08-30
Add connection test for LLM providers

### LINKED ISSUE #449: Test LLM service on create
When a user adds a new LLM service to their team, it could be useful if we allow them to quickly test the integration (we can either do it autmatically in the background or allow them to enter some text to test it). This will reveal any issues with the integration there and then. If we don't do this, the first failure will happen when someone chat with the bot, but the error will be masked, since we show some generic error message to the user (or just tell them something went wrong if they use an external channel).

Davemaina1:
### Product Description

Adds a way to verify an LLM provider's credentials actually work, instead of only finding out when a bot's first real chat fails. A "Test Connection" button on the provider edit page sends one minimal message and reports success or a specific failure reason. The same check also runs automatically every time a provider is saved (created or updated), surfacing a non-blocking warning on failure.

Closes #449

### Technical Description

Design doc, including two open questions and the reasoning behind building both manual and automatic instead of one: #449 (comment).

Two decisions worth flagging directly:

- The automatic check runs inline (nested savepoint, same shape as `VoiceProvider.run_post_save_hook`), not as a Celery task. Happy to move it async if the 10s worst case proves too slow in practice — noted as a real option in the design doc, not built here.
- Voyage AI is excluded from the test entirely (embeddings-only, can't run a chat completion). Caught a real bug here during self-review: the original `self.type == LlmProviderTypes.voyage.value` comparison was always `False` (comparing a string to a dataclass instance), silently letting the short-circuit never fire. Fixed to `self.type_enum == LlmProviderTypes.voyage`, matching the pattern already used elsewhere in this file, with a regression test that fails without the fix.

The test call uses a model the provider already has configured (preferring a team-scoped model over a global default, same priority as pricing-rule resolution elsewhere in this app) rather than a hardcoded per-type model list.

### Demo

1. Edit an existing LLM provider, scroll to the bottom, click "Test Connection".
image

2. With bad/fake credentials: a "Test failed. Check your credentials and try again." message appears; the page doesn't crash.
image

3. Change the provider's name (no credential change) and click Update: the same test runs automatically, same failure message appears, and the name change still saves (non-blocking).
image

Verified both in a local browser session against a provider seeded with fake test credentials.

### Docs and Changelog

New: a "Test Connection" button on LLM provider settings, plus an automatic check on save.

### Operator Impact
- [ ] Self-hosted operators must know about or act on this change.


apps/service_providers/tests/test_views.py
```diff
@@ -1,11 +1,21 @@
 from unittest import mock
 
+import httpx
+import openai
```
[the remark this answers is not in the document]
Davemaina1: Fixed in 1c1a9eb5b: moved the `openai` import into the one test that needs it, with a `# noqa: PLC0415` comment matching the same local-import convention already used elsewhere in this file for heavy libraries.

apps/service_providers/urls.py
```diff
@@ -39,4 +39,5 @@
     path("<slug:provider_type>/<int:pk>/remove-file/<int:file_id>", views.remove_file, name="delete_file"),
     path("<slug:provider_type>/<int:pk>/upload-file/", views.AddFileToProvider.as_view(), name="add_file"),
     path("<slug:provider_type>/<int:pk>/sync-voices/", views.sync_voices, name="sync_voices"),
+    path("<slug:provider_type>/<int:pk>/test-connection/", views.test_llm_connection, name="test_llm_connection"),
```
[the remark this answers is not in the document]
Davemaina1: Fixed in 1c1a9eb5b: the view now rejects any provider_type other than `llm` with a 404, matching the existing `_resolve_subtype` pattern in this same file. The redirect at the end now also uses the validated `llm` slug directly rather than the raw URL parameter.

apps/service_providers/models.py
```diff
@@ -202,6 +205,58 @@ def create_remote_index(self, name: str, file_ids: list | None = None) -> str:
...
+
+        if self.type_enum == LlmProviderTypes.voyage:
+            raise ConnectionTestNotSupportedError(self.type)
+
+        # A team-configured model wins over a global default, same priority as pricing-rule
+        # resolution elsewhere in this app.
+        model = (
+            LlmProviderModel.objects.for_team(self.team)
+            .filter(type=self.type)
+            .order_by(models.F("team_id").desc(nulls_last=True))
```
SmittieC, line 279: We have a utility called [get_default_model](https://github.com/dimagi/open-chat-studio/blob/4bf9704a1ab9c66eafadc95faf2f4bab71ba3826/apps/service_providers/llm_service/default_models.py#L247) to get the name of the default model to use in cases like this.

apps/service_providers/models.py
```diff
@@ -202,6 +205,58 @@ def create_remote_index(self, name: str, file_ids: list | None = None) -> str:
...
+            .first()
+        )
+        if model is None:
+            raise NoTestableModelError(self.type)
+
+        chat_model = self.get_llm_service().get_chat_model(model.name, timeout=CONNECTION_TEST_TIMEOUT_SECONDS)
+        chat_model.invoke([HumanMessage(content="Hello")])
+
+    def run_connection_test_hook(self) -> list[str]:
```
SmittieC, line 291: After testing this, I'm thinking that perhaps we should not do the "automatic" testing anymore and only stick to the manual test. The reason being that it would be nice to show the user why a connection test failed i.e. 400 < status code < 500 = permission issue and 500 < status code = connection issue

SmittieC: It would also be nice if we can inform users if / when they change the config or when they first create an LLM provider, that the credentials haven't been tested yet and that they can test it. I'm thinking we show a banner when this is detected. Thoughts?

SmittieC (commented):
Thanks for this. Overall it looks good. A few comments

apps/service_providers/models.py
Davemaina1: Good catch, thanks. Went with `get_default_model` directly rather than `get_first_llm_provider_model` though. That one always filters down to the registered default's exact name with no fallback, so a team that configured e.g. `gpt-4o` but not `gpt-4.1-mini` would get `NoTestableModelError` even though they do have a testable model. Now it prefers the registered default when the team has it, otherwise falls back to whatever else they've configured. Also has to be a local import: `default_models` imports `LlmProviderModel`/`LlmProviderTypes` from this same file, so a module-level import here would be circular. Pushed in 0dd8b041d.

apps/service_providers/models.py
Davemaina1: Went ahead and built the status-code classification, that was a real gap. Both the manual button and the automatic save-time check now say why a test failed: rate limit/timeout, bad credentials (or an invalid saved config, same bucket), or a connection issue on the provider's side, using one shared classifier so both paths stay consistent. Didn't drop the automatic check: the richer message applies there too now, so hopefully that covers what you were after without losing the instant-feedback-on-typo case the design doc argued for. Let me know if there's something else behind the "drop automatic" idea I'm missing. Pushed in 0dd8b041d.

Davemaina1: Like the idea, but I'd rather keep it out of this PR. It's really a separate feature (persistent per-provider state, not a point-in-time test), and #449 was scoped to just verifying credentials, not tracking staleness over time. Happy to open a follow-up issue once this lands, unless you'd rather fold it in now.

Davemaina1:
One more thing while testing this: saving an LLM provider always redirects to the team list, not back to the provider's own page, so the "connection test failed... use Test Connection below" warning was pointing at a button that wasn't on the page you land on. Unrelated to any of the comments above, found it while manually testing the classification changes. Fixed it here too: a save with a connection-test warning now redirects back to the provider's own edit page instead; every other save (other provider types, a passing test) keeps the existing team-list redirect. Pushed in 0dd8b041d.

apps/service_providers/models.py
SmittieC: It would be nice if it can be rolled with this PR, but also we don't have to do this with persistence changes. Here's my thoughts:

- When adding a new LLM provider, we know the config has not been tested yet, so we know that the banner should be shown nudging the user to test the connection.
- When updating an LLM provider, we can easily in the UI detect field changes on the config fields. If we do detect a change, then we know to show the banner.

One weirdness to be aware of: If I update the config, navigate away and come back again, we will _not_ show a banner, because we assume that the config is ok. I think this is ok, since we did nudge the user to test the connection.

Thoughts?

Davemaina1: Built both. One deviation worth flagging: not "always show on create". Standard provider types already have global default models seeded, so a fresh create runs a real automatic test and gets a genuine pass/fail immediately. Showing "not yet tested" right after a passing test would be wrong. Instead: the banner fires specifically when there's no model to test against (create or update), which is the actual condition where credentials are unverified. Update-time field changes are covered exactly as you described, client-side, with no persistence. Pushed in 6ae545ce6.

apps/service_providers/models.py
```diff
@@ -2,7 +2,7 @@
 import logging
 from collections.abc import Callable
 from enum import Enum
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
```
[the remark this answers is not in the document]
Davemaina1: Pre-existing structure, not introduced by this PR. models.py has held six unrelated provider domains (LLM, Voice, Messaging, Auth, Trace, Embedding) since well before this branch, one models.py per Django app is the existing convention here. This PR only added a small function and edited two methods inside the LLM section. Splitting the file is a separate refactor touching about 90 importing files across the app, out of scope here.

templates/service_providers/provider_form.html
```diff
@@ -37,7 +37,19 @@ <h1 class="pg-title">{{ title }}</h1>
       {% if config_form.custom_template %}
         {% include config_form.custom_template with object=object form=config_form %}
       {% else %}
-        {% render_form_fields config_form %}
+        {% if test_llm_connection_url %}
+          {# Nudges toward re-testing after an edit, not a saved "untested" state: resets on
```
SmittieC: Remove this comment please. It showed up in the UI:

Image

templates/service_providers/provider_form.html
```diff
@@ -59,6 +71,12 @@ <h1 class="pg-title">{{ title }}</h1>
           <button type="submit" class="btn btn-outline">Sync Voices from ElevenLabs</button>
         </form>
       {% endif %}
+      {% if test_llm_connection_url %}
+        <form method="post" action="{{ test_llm_connection_url }}" class="mt-4">
+          {% csrf_token %}
+          <button type="submit" class="btn btn-outline">Test Connection</button>
```
SmittieC, line 106: A few observations:

1. When I change my credentials from invalid to valid ones and I click this button, it still uses the invalid credentials.
2. We test the credentials every time I click update, even when I just changed the name

 A thought which is a slight pivot:
1. We don't really need this button anymore if we have automatic testing.
2. We should only fire auto-testing when the credentials changed. It should also be obvious that we're going to test the credentials, so I propose that we show one of three button texts: "Create and Test", "Update and Test" and "Update".

**Create and Test** - Show in create views only
**Update and Test** - show in update form, only when a credential field changed
**Update** - Show in the update form by default

Naturally, we should also in the backend only fire a test if a credential field changed.

Thoughts on this?

apps/service_providers/models.py
```diff
@@ -164,6 +164,54 @@ def _build_llm_service(self, llm_service, config: dict) -> "llm_service.LlmServi
...
+        return "retryable"
+
+    if isinstance(exc, ServiceProviderConfigError):
+        return "permission"
+
+    status = getattr(exc, "status_code", None) or getattr(exc, "code", None)
+    if isinstance(status, int) and 400 <= status < 500:
+        return "permission"
+
+    return "connection"
```
SmittieC, line 217: Testing a gemini provider with invalid credentials told me that it was a connection error, but in the logs I saw `message: "API key not valid. Please pass a valid API key."`.

templates/service_providers/provider_form.html
Davemaina1: Fixed, and confirmed live, not just in code. Django's `{# ... #}` comment tag doesn't span multiple lines (its tokenizer regex doesn't match across a newline), so mine never got recognized as a comment and rendered as literal text instead. Switched to `{% comment %}...{% endcomment %}`, already used elsewhere in this codebase. Pushed in 44a9d7433.

apps/service_providers/models.py
Davemaina1: Found it. Our Gemini integration (langchain_google_genai) catches the real error -google.api_core.exceptions.InvalidArgument, which does carry a status code, and re-raises its own wrapper (`ChatGoogleGenerativeAIError`) around it, `from e`. The wrapper itself has no status of its own, only the exception it's chained to does, which is what our classifier was missing. It now checks `exc.__cause__`/`__context__` too. Verified against the actual classes with your exact error message ("API key not valid. Please pass a valid API key."). Pushed in 44a9d7433.

templates/service_providers/provider_form.html
Davemaina1: Good catches, both real. The stale-credentials one predates this PR - Test Connection is a separate form posting to a separate endpoint that just re-tests whatever's already saved, not what's in the field. Not fixed yet - this and the retest-only-on-credential-change/button-text proposal is a real change to how the whole feature works, not a quick patch, so I want to scope it properly rather than fold it in alongside the two bug fixes I just pushed. Picking this up next.

Davemaina1: Built this. Went with a version close to your proposal, with one deliberate change: the manual "Test Connection" button isn't removed, it's hidden whenever there are unsaved credential changes, so it can never test stale data (that was the actual bug you found) but stays available to re-verify a provider you haven't touched today, in case a key was revoked or rotated since it was last saved. That was the one case the original design doc built the button for that "test on save" alone can't cover.

What changed:
- Automatic test now only fires when the config actually differs from what's saved, or it's a new create. A name-only edit no longer triggers it.
- Submit button says up front what's about to happen: "Create and Test", "Update and Test" once you touch a credential field, or plain "Update"/"Create" otherwise.
- Test Connection button hides itself the moment credentials are edited, reappears once saved.

Screenshots (before/after): https://docs.google.com/document/d/1AeVlz7Pi6G9dq15nD3CwAWzZ83SFbYtd4QKGUgTrLbA/edit?usp=sharing

Pushed in ed1d734a9.

api-schemas/v1.yml
```diff
@@ -2024,6 +2024,8 @@ components:
             participants:read: Read Participant Data
             participants:write: Update Participant Data
             usage:read: Read usage and activity data
+            openid: OpenID Connect scope
```
SmittieC: These changes should not be in here. Do you have `OIDC_RSA_PRIVATE_KEY` set locally? If so, please remove it and regenerate the schema

SmittieC:
LGTM. Just the failing test

apps/service_providers/views.py
```diff
@@ -298,10 +337,25 @@ def _save_provider(self, request, primary_form, config_form, file_formset):
             if isinstance(obj, VoiceProvider):
                 for warning in obj.run_post_save_hook():
                     messages.warning(request, warning)
-            if is_new and isinstance(obj, MessagingProvider):
+            if _should_run_whatsapp_post_create_hook(obj, is_create):
```
SmittieC: A more generic name here would be "_should_run_post_create_hook"

api-schemas/v1.yml
Davemaina1: Confirmed - I had OIDC_RSA_PRIVATE_KEY set locally, which flips OIDC_ENABLED on and adds these scopes. Regenerated without it in d8045454a; matches main now.

apps/service_providers/views.py
Davemaina1: Renamed to _should_run_post_create_hook in d8045454a.

SmittieC (approved):

PR #4318, opened 2026-08-30
feat(llm): add LiteLLM as an OpenAI-compatible LLM provider

### LINKED ISSUE #4078: Add LiteLLM support

Davemaina1:
Closes #4078

### Product Description

Teams running their own OpenAI-compatible model gateway (a self-hosted or
third-party LiteLLM proxy) can now add it as a chatbot LLM provider the same
way they'd add OpenAI or Groq: an API key and the proxy's base URL, then pick
or add the backend model(s) it serves.

### Technical Description

New `LlmProviderTypes.litellm`, routed through the existing `OpenAIGenericService`
(same as groq/perplexity/minimax/openrouter in #4233). Unlike those, LiteLLM has
no fixed endpoint, so `LiteLLMConfigForm` adds a required `openai_api_base` field
with normalization (strip trailing slash, append `/v1` only if missing) shared
between the form and the new `_litellm()` credentials env-loader, so there's one
implementation. No default models are shipped, same reasoning as OpenRouter:
users point this at their own backend models.

While building this, found the create-page template silently rendered nothing
for any provider with zero default or custom models: `for_type` is an enum
member, and a dict-membership check against it raises `TypeError`, which
Django's `{% if %}` swallows as `False`. Fixed with an explicit empty-state
message, scoped per provider type. Not LiteLLM-specific, but only became
visible once a provider shipped zero defaults.

Also surfaced, documented but out of scope here: `LlmProviderModel` has no FK
to `LlmProvider`, so two instances of the same provider type on one team can
collide on an identical model name. Pre-existing across every provider type,
not introduced by this PR. Filing separately.

### Migrations

`0075_alter_embeddingprovidermodel_type_and_more`: choices-only change on
`LlmProvider.type`, `LlmProviderModel.type`, `EmbeddingProviderModel.type` to
add the `litellm` choice. No column, index, or data change.

### Demo

1. Create form
image

2. Filled form
image
image

3. Base URL normalization

Entered `https://litellm.example.com` (no `/v1`) on create. Confirmed stored
normalized via `LlmProvider.objects.get(name="Test LiteLLM").config` in the
Django shell:

```
litellm {'openai_api_key': 'sk-test-litellm', 'openai_api_base': 'https://litellm.example.com/v1'}
```

### Docs and Changelog

New provider type: LiteLLM, for self-hosted or proxied OpenAI-compatible
models. Same setup flow as OpenAI/Groq, with an added required Base URL field
pointing at the operator's proxy.

### Operator Impact

Carries a migration and two new optional env vars (`LITELLM_API_KEY`,
`LITELLM_API_BASE`). CHANGELOG.md entry to follow in a separate commit once
the PR number is known.


apps/service_providers/forms.py
```diff
@@ -259,6 +259,31 @@ class DeepSeekConfigForm(ObfuscatingMixin, ProviderTypeConfigForm):
...
+    url = url.rstrip("/")
+    if not url.endswith("/v1"):
+        url = f"{url}/v1"
+    return url
+
+
+class LiteLLMConfigForm(ObfuscatingMixin, ProviderTypeConfigForm):
+    obfuscate_fields = ["openai_api_key"]
+
+    openai_api_key = forms.CharField(label=_("API Key"))
```
SmittieC: Since you're creating a custom form, you might as well drop the openai prefixes and rename these fields to `api_key` and `api_base`

SmittieC (commented):
Thanks for this. One comment from my side. There's also a few coderabbitai comments

apps/service_providers/forms.py
Davemaina1: Done - fields are now `api_key`/`api_base`. Stored config still uses `openai_api_key`/`openai_api_base` under the hood though, that's what `OpenAIGenericService` expects (same as groq/perplexity/minimax/openrouter), so I added a translation in `__init__`/`save()` rather than touching the shared service class. Also had to fix the edit page: it pulls `initial` straight from the stored config, so a bare field rename would've shown a blank key instead of the masked one when editing an existing provider.

On the CodeRabbit comments: the missing-assert one is already fixed. Not actioning the HTTPS-only or migration-rollback ones, no other provider form enforces HTTPS (would break self-hosted proxies on a private network with no TLS), and the embedding-provider concern doesn't hold up: `OpenAIGenericService.get_local_index_manager` (`main.py:273`) already has a real implementation, it doesn't fall through to the base class's `NotImplementedError`. Happy to revisit any of these if you see it differently.

SmittieC: @Davemaina1 I didn't realize that the form parms maps directly to the `OpenAIGenericService`, so the way you had it was better retrospectively. If you want you can revert that change, but please do add a comment in the form about this

Davemaina1: Reverted - fields are back to `openai_api_key`/`openai_api_base`, and added a comment above them explaining the `OpenAIGenericService` constraint so it doesn't trip anyone else up. Pushed in ef7346c55.

SmittieC:
@Davemaina1 Looks like you just need to run `inv schema` to fix the failing test, otherwise everything looks good

SmittieC (approved):

PR #4319, opened 2026-08-30
Add toggle filters, error count, and team display to notifications

### LINKED ISSUE #4075: Notifications UI improvements
## Notifications UI Improvements

A set of UX improvements to the notifications list and detail pages.

### List page

- [ ] **Filter by team** — toggle buttons if < 3 teams, dropdown with multi-select if 3 or more
- [ ] **Filter by level** — toggle buttons (e.g. info / warning / error)
- [ ] **Display error count** — show the number of errors on each notification row

### Detail page

- [ ] **Show team** — display the associated team on the notification detail page
- [ ] **Link to trace for bot errors** — when the notification is a bot error, include a link to the relevant trace

Davemaina1:
### Product Description

Four small UX improvements to the notifications list and detail pages:

- **Toggle-button filters** for team (when the user belongs to fewer than 3 teams) and level (Info/Warning/Error), alongside the existing dropdown filter panel.
- **Error count**: each notification row now shows how many times that error has recurred (e.g. `×3`), when it's happened more than once.
- **Team on the detail page**: the breadcrumb now shows which team a notification belongs to, since this page is reachable from the cross-team notifications list.

Closes #4075.

Not in this PR: the trace-link gap for `tool_error_notification` (see the design-doc comment on the issue) — that needs new plumbing to access the current trace from inside tool execution, so it's scoped as a separate follow-up per the open question raised there.

### Technical Description

- Toggle buttons are a page-local addition (`notification_toggle_filters.html`) driving the existing `f_<column>`/`op_<column>` query-param convention (`FilterParams`) directly, rather than modifying the shared `experiments/filters.html` panel every other filterable page also uses.
- Each click swaps `#data-table` and, out-of-band, re-renders the toggle-button row itself (`notification_table_with_toggles.html`) with the query string/active state for the filter it just applied. Without that, a button's own `hx-get`/`hx-push-url` would stay frozen at whatever it was on the last full page load, and a second click on the same button could never undo what the first one did.
- That re-render surfaced a real edge case in `FilterParams.from_request`: it falls back to the `HX-Current-URL` header whenever a request's own query string has no `f_` params (a convenience for sort/pagination links that intentionally omit filter state). htmx only updates the address bar from `hx-push-url` *after* a response settles, so a request that clears the last active filter arrives with a header that still names the pre-click state. Toggle-button requests now carry an `explicit_filters` marker so `resolve_notification_filter_params` treats their (possibly empty) query string as authoritative instead of consulting that header — sort/pagination links are unaffected.
- The `explicit_filters` marker also needs to run the same legacy `filter_<n>_*` query-param translation `FilterParams.from_request` does before trusting the query string directly — otherwise a bookmarked legacy-format filter link could be dropped by a toggle click. (The underlying translation helper itself has a separate, pre-existing bug — it silently drops modern `f_*` params when a legacy param is also present in the same query string — that predates this PR and affects every page using this filter system, not just notifications; out of scope here.)
- Error count is a `Subquery`/`Count` annotation on `EventUserQuerySet` (`with_event_count`), counting `NotificationEvent` rows for the same `EventType` - the interpretation proposed in the design doc's open question, since no reply landed on the issue.
- No migrations, no schema changes.

### Demo

1. Notifications list – toggle-button filters (team + level) and the error-count badge (`×N`) on a row.
image

2. Notifications list – level filter toggled active (button highlighted, table filtered).
image

3. Notification detail page – the team is shown in the breadcrumb.
image

### Docs and Changelog

### Operator Impact


Davemaina1:
Addressed both findings:

- **Legacy filter dropped on toggle click** - confirmed real. Reproduced it directly: `resolve_notification_filter_params` was skipping the legacy `filter_<n>_*` translation that `FilterParams.from_request()` normally does. Fixed in 07a4c7f5c with a regression test. Also found the root cause runs deeper than this PR: the shared `_translate_legacy_query_params`/`convert_saved_filter_data` helper (`apps/web/dynamic_filters/`) silently drops *any* modern `f_*` param that's mixed with a legacy one in the same query string, app-wide, independent of this PR — my fix makes the notifications toggle path consistent with the rest of the app's existing (if imperfect) behaviour rather than worse.
- **Linked issue / trace-link scope** -  deferral is noted directly in the PR description, per the design doc's own proposal (open question 2 on the issue).

Docstring coverage: left as-is. Measured it against untouched sibling files in this app (views.py/filters.py/test files all sit at 12-30%) - the 80% bar doesn't match this repo's own established norm here.

apps/ocs_notifications/filters.py
```diff
@@ -88,3 +93,67 @@ class UserNotificationFilter(MultiColumnFilter):
...
+    `EXPLICIT_FILTERS_PARAM` -- consulting that header for them would be wrong, since it
+    reflects the browser's address bar as of *before* this response's own `hx-push-url` takes
+    effect, one click stale.
+
+    Still runs the same legacy `filter_<n>_*` translation `FilterParams.from_request` does
+    before trusting the query string, so a bookmarked legacy-format filter link isn't dropped
+    out from under a user who then clicks a toggle button.
+    """
+    if EXPLICIT_FILTERS_PARAM in request.GET:
+        return FilterParams(_translate_legacy_query_params(request.GET))
```
snopoke: I don't think we need to support legacy filters here. We'll also be removing support for legacy filters soon (https://github.com/dimagi/open-chat-studio/issues/4329)

apps/ocs_notifications/filters.py
```diff
@@ -88,3 +89,63 @@ class UserNotificationFilter(MultiColumnFilter):
...
+        option_id = str(choice["id"] if isinstance(choice, dict) else choice)
+        label = choice["label"] if isinstance(choice, dict) else choice
+        is_active = option_id in active_values
+        new_values = active_values - {option_id} if is_active else active_values | {option_id}
+
+        query = base_query.copy()
+        if new_values:
+            query[f_key] = serialize_csv_tilde_values(sorted(new_values))
+            query[op_key] = "any of"
+        query[EXPLICIT_FILTERS_PARAM] = "1"
```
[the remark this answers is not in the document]
snopoke: No need to support legacy filter formats from now on: https://github.com/dimagi/open-chat-studio/issues/4329

apps/ocs_notifications/views.py
```diff
@@ -23,7 +29,33 @@
...
+TEAM_TOGGLE_BUTTON_THRESHOLD = 3
+
+
+def get_notification_toggle_context(request) -> dict:
+    """Build the level/team toggle-button options for the notifications list (see
+    build_toggle_options). Shared by the page's initial render and by the table endpoint's own
+    response, so every click re-renders the buttons with the query strings/active state that
+    match the filter it just applied -- htmx swaps this back in out-of-band alongside the table
+    (see notification_table_with_toggles.html), otherwise a button's own href would stay frozen
+    at whatever it was on the last full page load and a second click could never undo the first.
```
snopoke:
```suggestion
    (see notification_table_with_toggles.html).
```

Issue #4140, opened 2026-08-07
Write API Phase 1 — 5: Pipeline façade — node endpoints


SmittieC:
Part of the Phase 1 write API. **Spec:** [Write Chatbot Management API](https://docs.google.com/document/d/1oD25fmxCjdam728UOinAj7bry15ih1pxdhf5paokalw/edit?tab=t.9rysxi5a1wpm)

## Scope

`POST`/`PATCH`/`DELETE /chatbots/{id}/pipeline/nodes/…` (spec §5.2, W2, W5, W6, W7). The biggest chunk; edges split out into a separate ticket to keep it reviewable.

## Deliverables

- Façade write plumbing (shared with the edge-endpoints ticket): `select_for_update` on the Pipeline row inside `transaction.atomic()` (W7); read `Pipeline.data` → mutate → `pipeline.save()` → `update_nodes_from_data()` — the builder's exact sequence (`apps/pipelines/views.py:385`); response envelope = resource + `pipeline_valid` + normalized `errors` + `unwired_handles`.
- **POST node**: server-assigned `node_id` (`Node.flow_id`) and parking position (`max(x) + 200`, `y = 200`); `type` alone suffices, Pydantic defaults materialized into the response (not persisted); client-supplied `node_id` → 400; unknown type → 400; missing type → 400.
- **PATCH node**: merge at `params` level, raw flat Pydantic field names, no nested→flat mapping; router keyword edits regenerate `output_handles`; stranded edges land in `errors.edge`, never auto-pruned.
- **DELETE node**: removes the node + its dangling edges; unknown `node_id` → 404; deleting the only Start persists + reports `errors.pipeline`.
- Strict reference validation: nonexistent resource id → 400, cross-team id → 404 — checked up front, never silently nulled (spec §6.2 boundary tables).

## Files

New façade module under `apps/api/v2/` (e.g. `pipeline_edit/`); urls; tests.

## Depends on

Task 1 (envelope), Task 4 (permissions).

## Test focus

The §6.2 node-create boundary table, row by row; concurrent-write safety via the row lock; lenient persistence (missing required param → 201 + reported); direct `Node`-row writes never happen (blob is source of truth).

---

## Note: pipeline graph guards to add when the write API lands

[...]

Issue #4135, opened 2026-08-07
Write Chatbot Management API — Phase 1


SmittieC:
## Goal

Ship the Phase 1 write API as a sequence of independently deployable changes.

**Spec:** [Write Chatbot Management API](https://docs.google.com/document/d/1oD25fmxCjdam728UOinAj7bry15ih1pxdhf5paokalw/edit?tab=t.9rysxi5a1wpm)

## Approach

Ten sub-tickets. Each is additive, lands behind existing auth/team-scoping, and leaves the deployed system fully working — read-only behaviour is never broken, and write surface appears endpoint-group by endpoint-group. Shared plumbing (build-state reporting, write permissions) ships first so later tickets stay small.

## Dependency graph

```
Task 1 (build-state core) ──> Task 2 (inspect additions)
        │
        └──────────────────> Task 5 (façade: nodes) ──> Task 6 (façade: edges)
        └──────────────────> Task 7 (publish + status)

Task 3 (discovery) ─ independent

Task 4 (write auth + create/PATCH) ──> Task 5, Task 7, Task 8 (archive),
                                        Task 9 (triggers), Task 10 (content)
```

Parallelizable lanes after Task 1 + Task 4 land: `{2, 3}`, `{5→6}`, `{7}`, `{8}`, `{9}`, `{10}`.

## Sub-tickets

- #4136 — Pipeline build-state reporting core (no API surface change)
- #4137 — Additive inspect fields (read side of the write loop)
- #4138 — Discovery endpoints
- #4139 — Write-API auth plumbing + chatbot create/PATCH
- #4140 — Pipeline façade: node endpoints
- #4141 — Pipeline façade: edge endpoints
- #4142 — Publish + version status polling
- #4143 — Chatbot archive with channel guard
- #4144 — Triggers (static + timeout)
- #4145 — Content resources (source material, consent forms)

## Final end-to-end test

After Task 7 lands: the spec §4 "agent's loop" — discover → inspect → build → trigger → publish → poll → verify snapshot — as one integration test.

## Explicitly out of scope for Phase 1

Collections + file upload, custom-action authoring, canvas auto-layout on wiring (W11), channel anything, provider writes, idempotency keys, rate limiting (spec §3, §9, §10).

PR #4331, opened 2026-09-01
Write API: edge endpoints

### LINKED ISSUE #4141: Write API Phase 1 — 6: Pipeline façade — edge endpoints

## Scope

`POST /chatbots/{id}/pipeline/edges/` + `DELETE …/edges/{edge_id}/` (spec §5.2, W5, §8.2).

## Deliverables

- **POST edge**: server-generated edge id (`FlowEdge.id`); `source`/`target` must be nodes in this pipeline (400), `source_handle` must be offered by the node's `output_handles` (400); `target_handle` optional when the target has a single input; duplicate edge (same source/handle/target/handle) → 400, making the most retry-prone write idempotent (§8.2); cycles / unreachable End persist + report in `errors.pipeline`.
- **DELETE edge**: by the id inspect emits; unknown → 404.
- Phase 1 layout: no repositioning on wire (W11 is Phase 2) — nodes keep their parking position.

## Files

Same façade module as the node-endpoints ticket; tests.

## Depends on

Task 5 (pipeline façade: node endpoints).

## Test focus

The §6.2 edge-create boundary table; duplicate-edge retry safety; wiring updates `unwired_handles` on both endpoints of the edge.

SmittieC:
Closes #4141. Part of the Phase 1 write API (#4135), stacked on the node endpoints (#4140) — **base branch is `cs/pipeline_facade_redone`, not `main`.**

### Product Description

An agent building a chatbot's pipeline over the API can now wire and unwire its nodes:

```
POST   /api/v2/chatbots/{id}/pipeline/edges/
DELETE /api/v2/chatbots/{id}/pipeline/edges/{edge_id}/
```

`source` and `target` are usually the whole body; a handle is named only when the node offers a choice, which today means a router's branches. Both endpoints run the same locked read-modify-write as the node endpoints and answer with the same envelope: the edge written, then the state of the pipeline it went into. A wire the server can act on lands even when it leaves the pipeline unbuildable — an agent is mid-build, and `pipeline_errors` is how it learns what is still wrong. A wire naming something the pipeline does not have is refused with a 400 keyed to the field at fault.

### Technical Description

Not obvious from the diff:

- **Edge ids are derived, so they can collide.** They come from react-flow's own `getEdgeId` formula, so an API-wired edge reads like a hand-wired one — but a router keyword edit leaves an edge holding the id of a wiring it no longer has (`graph_editor._rewired_edges`), and `patching._apply_edge_diff` treats an add whose id already exists as a no-op, i.e. 201 having stored nothing. So a taken id gets a suffix, and `_write_response` reads the edge back out of the saved graph rather than reporting the planned one. Ids are not byte-identical to builder-drawn ones: the builder renders no id on target handles, so its edges stop at the target's node id where these end in `input`.
- **Duplicate wires are refused rather than stored twice**, which is what makes the most retry-prone write in the API safe to retry (spec §8.2) — a repeat leaves the graph exactly as the first call left it, and the 400 names the edge already wiring the pair so a client that never saw the first response can carry on. Detection normalises absent handles, because the builder stores a null `targetHandle` where the server stores `input`.
- **`source_handle` is optional**, defaulting to the source's only handle. The ticket said that of `target_handle` alone; extending it is symmetric and saves an agent needing to know that `output` is the magic name. The cost, worth recording: a node type that later gains a second output turns previously-valid calls into 400s. Equally true of `target_handle`.
- **An unreachable End node reports under `pipeline_errors.node[<end>].root`, not `.pipeline`** — a deliberate departure from the ticket's shorthand. `PipelineBuildError` carries the End node's id, and that bucketing is already what the shipped `/inspect/` endpoint publishes, so changing it would break a read contract. Cycles do land in `pipeline_errors.pipeline`.

### Demo

Driven end-to-end against a live server on a dev DB: wired four edges, exercised every refusal, deleted by the id `/inspect/` emitted, retried the delete (404), closed a cycle (201 + reported). The builder's own data endpoint hands back API-wired edges in the shape it writes its own.

```
$ POST /pipeline/edges/ {"source": "<router>", "target": "<end>"}
400 {"source_handle": "'RouterNode-746f7' offers more than one output handle;
      name the one to wire from: output_0, output_1."}

$ POST /pipeline/edges/ {"source": "<llm>", "target": "<end>"}     # retried
400 {"non_field_errors": ["These nodes are already wired this way, by edge
      'reactflow__edge-...output-endinput'. Nothing was changed."]}
```

### Docs and Changelog

The OpenAPI schema is the documentation for this surface and is regenerated in-tree (`api-schemas/v2.yml`), including the exact duplicate-wire message so a client can anchor on it.

### Operator Impact

Purely additive API surface: no migrations, no new settings or env vars, nothing removed.

---


apps/pipelines/build_state.py
```diff
@@ -56,15 +56,25 @@ def unwired_handles(pipeline: Pipeline) -> dict:
...
     for handle in node_output_handles(node):
         if (node.flow_id, handle["handle"]) not in wired_outputs:
             dangling.append(handle)
     return dangling
 
 
+def input_handles(node_type: str) -> list[str]:
+    """The input handles a node of this type accepts an edge on.
+
+    Every type has exactly one, implicit, ``input`` handle -- bar Start, which has none: nothing runs
```
SmittieC: shorten this docstring

api-schemas/v2.yml
```diff
@@ -169,6 +169,126 @@ paths:
               schema:
                 $ref: '#/components/schemas/ChatbotInspect'
           description: ''
+  /api/v2/chatbots/{id}/pipeline/edges/:
+    post:
+      operationId: pipeline_edge_create
+      description: |-
+        Add an edge to the chatbot's working (draft) pipeline, wiring one node's output handle to another node's input.
```
SmittieC: this needs to be less verbost. Stick to the core of what this endpoint does

api-schemas/v2.yml
```diff
@@ -169,6 +169,126 @@ paths:
...
+
+        `source` and `target` are enough for most nodes: a handle you leave out means the only one the node has. A router is the exception — it exposes one handle per branch, so name the branch to wire in `source_handle`. Both nodes keep the position they were parked at; nothing is moved on the canvas.
+
+        The edge's `id` is assigned by the server and cannot be chosen. It is the address `DELETE .../pipeline/edges/{edge_id}/` takes, and the same id `GET /api/v2/chatbots/{id}/inspect/` reports.
+
+        Wiring a pair that is already wired the same way is refused rather than stored twice, so retrying a wire whose response you never saw is safe — the 400 names the edge that already wires the pair.
+
+        A wire that leaves the *graph* wrong is not refused: it lands, and comes back in `pipeline_errors` for you to repair. Note the two graph-level problems land in different buckets — a cycle in `pipeline_errors.pipeline`, while an End node nothing reaches is reported against the End node itself, under `pipeline_errors.node[<end node id>].root`.
+      summary: Wire two Pipeline Nodes
+      parameters:
```
SmittieC, line 246: It would be nice if we can wire multiple nodes in the same call. @snopoke thoughts? Should we rather stick to adding a single wire per call?

apps/api/v2/pipeline_edit/tests/conftest.py
```diff
@@ -125,6 +125,25 @@ def client(chatbot):
     return ApiTestClient(chatbot.team.members.first(), chatbot.team)
 
 
+@pytest.fixture()
+def start(chatbot) -> str:
```
SmittieC:
```suggestion
def start_node(chatbot) -> str:
```

apps/api/v2/pipeline_edit/tests/conftest.py
```diff
@@ -125,6 +125,25 @@ def client(chatbot):
...
+    return boundary_node(chatbot, "StartNode")
+
+
+@pytest.fixture()
+def end(chatbot) -> str:
```
SmittieC:
```suggestion
def end_node(chatbot) -> str:
```

apps/api/v2/pipeline_edit/views.py
```diff
@@ -1,12 +1,15 @@
-"""The pipeline façade's node endpoints (#4140).
+"""The pipeline façade's node and edge endpoints (#4140, #4141).
 
-A chatbot's pipeline is edited one node at a time rather than replaced wholesale: the server holds
-the graph and applies the change, so a client never has to reproduce a whole document to alter one
-setting, and two edits to different nodes cannot overwrite each other.
+A chatbot's pipeline is edited one node or one edge at a time rather than replaced wholesale: the
```
SmittieC, line 3: This comment is for this PR in general: Keep the docstings / module strings succinct

apps/api/v2/pipeline_edit/edge_editor.py
```diff
@@ -0,0 +1,225 @@
...
+from .facade import PipelineEdit, graph_diff
+from .ids import with_free_suffix
+
+#: The four values that identify an edge: ``(source, source handle, target, target handle)``, as
+#: :attr:`~apps.pipelines.flow.FlowEdge.wiring` reads them off a stored edge.
+Wiring = tuple[str, str, str, str]
+
+
+def plan_create(
+    flow: dict, source: str, target: str, source_handle: str | None, target_handle: str | None
```
SmittieC: nit: it would be nice if flow can be an instance of `Flow` instead of a dict. This goes for plan_delete as well (same for similar methods in graph_editor.py)

apps/pipelines/flow.py
```diff
@@ -84,6 +111,18 @@ def react_flow_node_type(node_type: str) -> str:
     return REACT_FLOW_NODE_TYPE
 
 
+def react_flow_edge_id(source: str, source_handle: str | None, target: str, target_handle: str | None) -> str:
```
SmittieC: this is only used in one place, so move the logic into that function please

apps/api/v2/pipeline_edit/tests/test_schema_examples.py
```diff
@@ -164,3 +166,21 @@ def _with_real_ids(params: dict, reference_ids: dict) -> dict:
     unknown = PLACEHOLDER_ID_PARAMS.symmetric_difference(reference_ids)
     assert not unknown, f"no id to substitute for {sorted(unknown)}"
     return {name: reference_ids.get(name, value) for name, value in params.items()}
+
+
+class TestWireExamples:
```
SmittieC: low value test. remove

apps/pipelines/build_state.py
SmittieC: Fixed in ccc19c392. Trimmed to two lines: `input_handles` now says every type has one implicit `input` handle bar Start, and why it returns a list.

*— Claude Code*

api-schemas/v2.yml
SmittieC: Fixed in ccc19c392. The `pipeline_edge_create` description is down to three short paragraphs — what it wires, the handle rule, and that the id is server-assigned plus the duplicate/invalid-graph behaviour. The canvas-position aside and the errors-bucket tour are gone. `api-schemas/v2.yml` regenerated to match.

*— Claude Code*

apps/api/v2/pipeline_edit/views.py
SmittieC: Fixed in ccc19c392 (app code, OpenAPI descriptions and field help text) and 1b83d2a1f (tests). Cut the restatement and rhetorical framing throughout — module strings, class and function docstrings — keeping the facts a reader cannot get from the code. Verified prose-only: an AST comparison with docstrings stripped shows no logic change in either commit.

*— Claude Code*

apps/api/v2/pipeline_edit/edge_editor.py
SmittieC: Fixed in 0e2b1fe0b. `edit_pipeline` now validates the locked read into a `Flow` once and hands that to the planner, so `plan_create`/`plan_delete` in both `edge_editor.py` and `graph_editor.py` (plus `plan_update`, `find_node`, `parking_position`, `_reparked_end_nodes`, `_rewired_edges`) take `Flow` and read attributes off the models instead of `.get("nodes", [])`. The patch engine still takes the raw graph. `find_node` and `_reparked_end_nodes` now copy the node they hand back, since both edit it in place to build the diff — the old per-node re-parse made that safe for free.

*— Claude Code*

apps/api/v2/pipeline_edit/tests/test_edge_create.py
```diff
@@ -0,0 +1,472 @@
...
+        response = client.post(
+            edges_url(chatbot), {"source": llm_node, "target": end_node, "source_handle": "output"}, format="json"
+        )
+
+        assert response.status_code == 400, response.content
+        assert len(edges_from(chatbot.pipeline, llm_node)) == 1
+
+    def test_a_second_edge_from_the_same_handle_to_another_node_is_allowed(self, client, chatbot, llm, end_node):
+        """Not a duplicate: one output handle may fan out, and the pipeline runs both branches."""
+        source = add_llm_node(client, chatbot, llm)
```
SmittieC, line 338: this is the same as test_wiring_the_same_pair_twice_is_refused.

apps/pipelines/flow.py
SmittieC: Fixed in 513494ccc. Dropped `react_flow_edge_id` from `apps/pipelines/flow.py` and inlined the f-string into `_unused_edge_id`, its only caller — which is also the only place that can make the id unique. The `or ''` handle fallbacks went with it: `Wiring` is a 4-tuple of non-empty strings, since `FlowEdge.source_handle_name`/`target_handle_name` already fill in the standard names.

*— Claude Code*

apps/api/v2/pipeline_edit/tests/test_schema_examples.py
SmittieC: Fixed in 772836713. Removed the whole `TestWireExamples` class, and with it the now-unused `EdgeCreateSerializer` / `WIRE_EXAMPLES` imports.

*— Claude Code*

apps/api/v2/pipeline_edit/tests/test_edge_create.py
SmittieC: Fixed in ee7fc3a96. Removed `test_a_duplicate_of_an_edge_the_ui_builder_wrote_is_refused`.

Note: this comment was anchored on line 343, inside `test_a_second_edge_from_the_same_handle_to_another_node_is_allowed` — which asserts the opposite (a fan-out from one handle *is* allowed). Confirmed with you that the builder-edge test just above it was the intended target, so that is the one that went.

*— Claude Code*

apps/api/v2/pipeline_edit/edge_editor.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def _target_handle(content: FlowNodeData, requested: str | None) -> str:
+    """Which of the target node's input handles the edge points at.
+
+    Never required: every node type has exactly one, implicit, input handle -- bar Start, which has
+    none. The field exists all the same so an edge read back from ``GET /inspect/`` can be sent
+    straight back, and so a multi-input node type would not need a new field.
+    """
+    accepted = input_handles(content.type)
```
snopoke: the logic below here is almost identicial to `_source_handle` with the difference of using `_why_no_output_handles` and the assumption here there is only ever one input handle. 

The assumption can be dropped without impact and the error message function can be adjusted to cater for input types.

apps/api/v2/pipeline_edit/facade.py
```diff
@@ -16,24 +16,45 @@
...
 
     diff: PipelineDiffPayload
     node_id: str | None = None
+    edge: FlowEdge | None = None
+
+    def __post_init__(self) -> None:
+        # Checked rather than merely documented: each view's `_write_response` reads only its own
+        # field, so a plan that set both would answer with one resource's envelope and drop the
+        # other, with nothing anywhere saying so.
+        if self.node_id is not None and self.edge is not None:
```
snopoke: should this also verify that at least one is not None?

apps/api/v2/pipeline_edit/serializers.py
```diff
@@ -184,15 +174,22 @@ class EdgeCreateSerializer(RejectsServerAssignedKeys, RejectsUnknownKeys, serial
 
     server_assigned_keys = SERVER_ASSIGNED_EDGE_KEYS
 
-    source = serializers.CharField(help_text="``node_id`` the edge leaves from. " + NODE_IDS)
+    source = serializers.CharField(
+        help_text=(
+            "``node_id`` the edge leaves from. Node ids are the ones a node write returns and "
+            "`GET /api/v2/chatbots/{id}/inspect/` reports."
```
snopoke: I'm not sure about putting URLs in the help text like this. If a new version is released we can't use this serializer without removing the URL.

api-schemas/v2.yml
snopoke: I think allow multiple would be nice but we should take an all or nothing approach (either all succeed or you get an error response and we make no changes).

snopoke (commented):
This LGTM

api-schemas/v2.yml
SmittieC: Done in dc170f428 — `POST .../pipeline/edges/` now takes a **top-level JSON array** of wires (a single wire is a list of one), applied all-or-nothing.

The planner runs inside the existing row-locked `transaction.atomic()`, so nothing is written when any wire is refused — the `edit_revision` is not even bumped. Rather than stopping at the first fault, every wire is checked and the refusals come back positioned like the body (`{}` where a wire is fine), which is the shape DRF's own list serializer refuses in:

```json
[{"source": "..."}, {}, {"source_handle": ["..."]}]
```

The duplicate rule now also holds *within* one body: a repeat has no existing edge id to name, so it names the index of the first of the two instead. The response leads with `edges`, in the order the body named them.

*— Claude Code*

apps/api/v2/pipeline_edit/edge_editor.py
SmittieC: Fixed in 004745077. Both ends now go through one `_handle()` — what differs between them is data on a `_Side` (the body field names, whether its handles are called "input" or "output", where they come from, and what to say when there are none). The single-input assumption is gone: an omitted handle is filled in only when the node offers exactly one, on either side.

*— Claude Code*

apps/api/v2/pipeline_edit/serializers.py
SmittieC: Fixed in 043a08f35. Paths in help text and schema descriptions are now relative to the version root (`GET /chatbots/{id}/inspect/`), matching how the docstrings in this package already wrote them. Done for every instance in `apps/api/v2/pipeline_edit/` — serializers, the `extend_schema` descriptions, the `node_id`/`edge_id` path params, and the reference-check error message in `references.py`.

Two more sit outside this PR's files, in `apps/api/v2/views.py` (lines 86 and 142). Left alone to keep the diff to this package — happy to fix them here or in a follow-up, whichever you prefer.

*— Claude Code*

apps/api/v2/pipeline_edit/references.py
```diff
@@ -61,7 +61,7 @@ def _reference_errors(team: Team, node_class: type[BasePipelineNode], params: di
         if unknown := [item for item in requested_values if _is_not_allowed(item, available_values)]:
             errors[param] = (
                 f"Not available to this team: {', '.join(repr(item) for item in unknown)}. "
-                f"Choose from the '{param_option_map[param]}' list in GET /api/v2/pipeline/options/{node_type}/."
+                f"Choose from the '{param_option_map[param]}' list in GET /pipeline/options/{node_type}/."
```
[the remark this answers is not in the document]
snopoke: +1 this is worse than not including a URL at all. Either drop the URLs entirely (maybe replace with endpoint IDs?) or else use the correct URL.

apps/api/v2/pipeline_edit/edge_editor.py
```diff
@@ -107,32 +138,54 @@ def _endpoints(flow: Flow, source: str, target: str) -> tuple[FlowNodeData, Flow
     )
 
 
-def _source_handle(content: FlowNodeData, requested: str | None) -> str:
-    """Which of the source node's output handles the edge leaves from.
+@dataclass(frozen=True)
+class _Side:
```
snopoke: I don't think the `_` convention with classes is necessary

apps/api/v2/pipeline_edit/edge_editor.py
```diff
@@ -173,34 +226,28 @@ def _no_output_handles_message(content: FlowNodeData) -> str:
...
+#: ``target_handle`` is never required, since only the Start node lacks an input handle and it
+#: cannot be a target at all. The field exists so an edge read back from ``GET /inspect/`` can be
+#: sent straight back, and so a multi-input node type would need no new field.
+_TARGET = _Side(
+    endpoint="target",
+    handle_field="target_handle",
+    kind="input",
+    offered=lambda content: input_handles(content.type),
+    none_offered=lambda content: f"'{content.id}' has no input handles, so no edge can point at it.",
+)
```
snopoke: This is very compact but it makes it harder to read the code that uses it. Might I suggest subclasses:

```python
class TargetEdgeSide(Side):
    endpoint="target"
    handle_field="target_handle"
    kind="input"

    def get_handles_offered(content):
       return input_handles(content.type)

    def get_non_offered_message(content):
        return f"'{content.id}' has no input handles, so no edge can point at it."
```

It is more code but I think it's a lot more readable. You can also add type hints to the methods and have better method names.

apps/api/v2/pipeline_edit/references.py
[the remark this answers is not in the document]
SmittieC: see here: https://github.com/dimagi/open-chat-studio/commit/383e6cd350b8abe2228bd99b1abc438efdddd6cf

apps/api/v2/pipeline_edit/edge_editor.py
SmittieC: see here: https://github.com/dimagi/open-chat-studio/commit/1063f845b271c9a46dba99e3e82ac9171eb14d55

apps/api/v2/pipeline_edit/edge_editor.py
SmittieC: see here: https://github.com/dimagi/open-chat-studio/commit/b63e3e0d9a99d9bc18585655cd7b51495cc5890a

api-schemas/v2.yml
```diff
@@ -219,17 +219,151 @@ paths:
...
+        required: true
+      tags:
+      - Pipelines
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                $ref: '#/components/schemas/EdgeCreate'
```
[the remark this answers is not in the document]
snopoke: @SmittieC this is a good point. Limiting the number of edges that can be created in one go to a reasonable number.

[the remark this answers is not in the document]
SmittieC: Valid finding, fixed in c51faedf1 — with a request-body change worth flagging, since this endpoint is not on `main` yet.

**The mechanism.** `edge_editor._refuse_duplicate` compares each wire against every wire already planned, and `plan_create` runs inside `facade.edit_pipeline`'s `transaction.atomic()` holding a `select_for_update` on the `Pipeline` row. So body length drove O(n²) work while every other writer to that pipeline blocked.

**Why the body moved.** Adding `max_length` to `WireListSerializer` and regenerating produced no schema change at all. drf-spectacular reads length constraints from a *field's* validators (`_insert_field_validators`, called while iterating a serializer's declared fields), but a top-level array body goes through `_unwrap_list_serializer` → `build_array_type(result)` with no bounds — and `build_array_type` would in any case emit `maxLength`/`minLength`, the string keywords a validator ignores on an array. A bare top-level array can therefore enforce a bound it has no way to state.

So the wires moved under a key, which makes the bound declarative and publishes it natively:

```python
class WireBodySerializer(RejectsUnknownKeys, serializers.Serializer):
    wires = serializers.ListField(
        child=EdgeCreateSerializer(),
        min_length=1,
        max_length=MAX_WIRES_PER_CALL,   # 100
    )
```

`POST` bodies are now `{"wires": [...]}` instead of `[...]`. DRF checks a `ListField`'s length before validating its children, so an oversized body is refused before the view opens its transaction.

**Refusals moved with it**, or the endpoint would report two shapes depending on which layer refused. They are now keyed by the position of the wire at fault, under `wires` — the shape the body serializer already refuses a wire in — and a wire that is fine has no entry rather than a `{}` placeholder:

```json
{"wires": {"0": {"source": "This pipeline has no node 'CodeNode-nope1'."},
           "2": {"source": "'end' offers no output handles, so no edge can leave it."}}}
```

**Schema.** `api-schemas/v2.yml` regenerated: `WireBody.properties.wires` carries `maxItems: 100` and `minItems: 1` (so CKV_OPENAPI_21 is addressed), plus `additionalProperties: false` from the existing `mirror_unknown_key_rejection` hook; `EdgeCreate` keeps its `$ref`. `v1.yml` and `export.yml` are unchanged.

**Tests.** Driven end-to-end against the endpoint, not just asserted: over-limit → `{"wires": ["Name at most 100 wires per call; ..."]}`; exactly 100 → read as wires, pinning the off-by-one; a bare list → a refusal naming the wrapper. `test_the_wire_body_publishes_the_limit_it_enforces` holds the committed `maxItems`/`minItems` to the constant, so raising the limit without regenerating fails naming the limit.

On CKV_OPENAPI_4 in the same comment: `security` is set per operation throughout this schema rather than globally, so the global-field rule is not one this API means to satisfy.

snopoke (approved):

PR #4357, opened 2026-09-01
Remove the assistant object from the inspect API

pxwxnvermx:
Part of splitting #4305 (OpenAI Assistants removal, phase 1). **Stacked on #4328** — review that first; this PR's diff is only correct once #4328 merges.

### Product Description
**Breaking, for API clients.** `/api/v2/.../inspect/` no longer returns an `assistant` key on a node, and `AssistantNodeParams` is gone from the node params union. The key was already conditional — omitted for nodes not declaring `assistant_id` — so only clients inspecting assistant-bearing pipelines are affected. Those nodes still render: the type is reported verbatim and the stored params come through the generic shape.

Nothing else changes. The assistant node still exists and still runs in this PR; the removal itself is part 4.

### Technical Description
This lands **before** the node and runtime removal, and that order is forced. Deleting `AssistantNode` orphans `assistant_id` in the inspect serializer's resource map, which `test_resource_param_fields_are_real_node_fields` catches immediately — so the removal PR depends on this one. The reverse is not true: dropping the `assistant` key is a pure contract change that works fine while the node still exists. I tried it the other way round first and it left 6 failing API tests on `main`.

Gone: the `AssistantSerializer`, the `assistant` entry in the resource map and its `assistant_id` lifting rule, `AssistantNodeParams`, and the `assistant` select-related on the inspect queryset. `api-schemas/v2.yml` is the regenerated artifact — it has to move in the same PR or the committed schema goes stale.

`assistant_id` is suppressed from the generic params rather than falling through to them: with no resource key left to lift it into, it would otherwise leak an internal database id to API clients that never had access to it before.

Two tests used the assistant node as their example of a resource-bearing node (the embedded-pipeline test and the query-count guard); both now use source material, keeping the coverage. `test_full_response_body` keeps its assistant node on purpose and asserts the shape *before* the class is removed; part 4 updates it to the after shape. That is the only test this split touches twice.

### Demo
Not applicable — API response shape only. The schema diff in `api-schemas/v2.yml` shows the removed component.

### Docs and Changelog

### Operator Impact


CHANGELOG.md
```diff
@@ -56,6 +56,13 @@ version section when a release is cut.
      removal may land, and links its deprecation tracking issue. -->
 
 ### Removed
+- **Breaking (API):** `/api/v2/.../inspect/` no longer returns an `assistant`
```
SmittieC: I don't think this type of change should be in here, since it doesn't pertain to self-hosting itself? cc @snopoke do you agree?

apps/api/v2/tests/test_chatbots_inspect.py
```diff
@@ -918,7 +910,7 @@ def _adversarial_bot():
     provider = LlmProviderFactory.create(team=team, name="Shared", type="openai")
     model = LlmProviderModelFactory.create(team=team, name="gpt-4o", deprecated=False)
     source = SourceMaterialFactory.create(team=team)
-    assistant = OpenAiAssistantFactory.create(team=team)
+    embedded_source = SourceMaterialFactory.create(team=team)
```
SmittieC: why not just use `source` which is also `SourceMaterial`?

apps/api/v2/inspect/serializers.py
```diff
@@ -505,12 +495,15 @@ class InspectNodeSerializer(serializers.ModelSerializer):
...
-        "assistant": ("assistant_id",),
         "custom_actions": ("custom_actions",),
         "media_collection": ("collection_id",),
         "indexed_collections": ("collection_index_ids",),
     }
     _RESOURCE_PARAM_KEYS = frozenset(param for params in _CONDITIONAL_KEY_PARAMS.values() for param in params)
+    # ``assistant_id`` is suppressed rather than lifted: ``_RESOURCE_PARAM_KEYS`` is derived from
+    # ``_CONDITIONAL_KEY_PARAMS`` and no render key claims it, so hiding it keeps an internal id
+    # out of the generic ``params`` payload.
+    _HIDDEN_PARAM_KEYS = _RESOURCE_PARAM_KEYS | {"assistant_id"}
```
SmittieC, line 508: I think we should have a data migration that drops the `assistant_id` key from `Node.params` rather than having to cater for legacy assistant keys.

SmittieC (commented):
Generally LGTM. A few comments

apps/api/v2/inspect/serializers.py
pxwxnvermx: This is planned for the Phase 2 for this removal. It would include removing all the data related to the Assistants feature.

apps/api/v2/tests/test_chatbots_inspect.py
pxwxnvermx: Fixed. https://github.com/dimagi/open-chat-studio/pull/4357/commits/528e44ec063cc75df2d10045891cdff3428d6bab

CHANGELOG.md
snopoke: I agree - this changelog should only include changes that require intervention when upgrading an OCS instance

pxwxnvermx: Thanks, I will remove this changelog.

SmittieC (approved):

Issue #4321, opened 2026-08-31
Team-scoped flag checks ignore waffle's `everyone` value


barry47products:
`Flag.is_active_for_team` (`apps/teams/models.py:246`) decides a flag by team membership alone:

```python
team_ids = self._get_team_ids()
return team.pk in team_ids
```

Waffle's own precedence puts `everyone` above every other rule, as a tri-state (`waffle/models.py`):

```python
if self.everyone:
    return True
elif self.everyone is False:
    return False
# else None: fall through to users / groups / percent
```

Three consequences.

**1. `override_flag` does not work for team-scoped flags.** `waffle.testutils.override_flag` sets `everyone` and nothing else, so a test decorated with it gets a flag row that `is_active_for_team` still reports as off. Tests have to hand-roll a `Flag.objects.create(...)` plus `teams.add(...)` fixture instead. There are at least six copies of that fixture across `apps/`.

**2. Two implementations disagree.** `Collection._flag_active_for_team` (`apps/documents/models.py:534`, added in `d64bdcfda`) restores the precedence locally before falling back to teams:

```python
if flag.everyone is not None:
    return flag.everyone
return flag.is_active_for_team(self.team)
```

Its docstring notes that indexing and retrieval run without a request, so waffle's request-based path is unreachable. That is true of any team-scoped check, so the workaround applies far wider than one app, but it is private to `Collection`. A flag with `everyone=True` is on for hybrid search and off for the channel platform dropdown at the same time.

**3. `everyone=False` does not disable a flag.** `Flag.is_active` (`apps/teams/models.py:234`) only short-circuits on a truthy result from `super()`:

```python
is_active = super().is_active(request, read_only)
if is_active:
    return is_active
...
return self.is_active_for_team(team)
```

Waffle returns `False` for `everyone is False`, which falls through to the team check, and any team in the M2M keeps the flag. The kill switch does not switch anything off. Read from source, not yet reproduced with a test.

## What needs deciding

Whether `is_active_for_team` omits `everyone` deliberately, so that a globally set flag cannot expose a feature to every tenant, or whether it is an oversight that `d64bdcfda` worked around.

- If deliberate: document it on the method, and give tests a shared team-flag fixture instead of six copies. `Collection._flag_active_for_team` then needs a second look, because it is the one place a global flag does reach every team.
[...]

PR #4359, opened 2026-09-01
Honour a global everyone=True in team-scoped flag checks

barry47products:
### Product Description

Feature flags switched on for everyone now reach team-scoped features. Most visibly: a team that enables Hybrid search or Contextual retrieval in Settings > Feature flags actually gets it. Until now the settings screen reported those flags as on while indexing and retrieval read the flag row a different way and ignored them.

### Technical Description

`Flag.is_active_for_team` decided by team membership alone, so `everyone=True` did nothing on team paths, and `override_flag` could not reach team-scoped checks in tests. `Collection._flag_active_for_team` carried a private workaround that read the tri-state directly, which inverted the bug for the two flags it gates: the team settings screen stores `everyone=False` to mean "no global override" (every creation path does, see #4321), and the workaround read that as "off for everyone".

- `everyone=True` now wins in `is_active_for_team`, checked before the team guard to match waffle's precedence. `everyone=False` keeps deferring to the team M2M.
- `Collection._flag_active_for_team` is deleted; both callers go through `Flag.get(...).is_active_for_team(...)`, the same shape as the channels platform gate. One behaviour note: with `WAFFLE_CREATE_MISSING_FLAGS` on, the first check lazily creates a missing flag row, where the old helper returned `False` without creating.
- `FeatureFlagForm.save()` only writes M2M changes for flags the admin actually toggled. Previously every save re-enrolled the team in each ticked flag, and with `everyone=True` now rendering ticked, an unrelated save would have written invisible team memberships that outlive the rollout.
- Known limit, accepted on #4321: the settings screen shows an `everyone=True` flag ticked for every team, and unticking it is inert because the global value overrides the M2M. This replaces the previous inconsistency, where the screen showed such flags off while request paths served the feature.
- Tests: a shared `team_flag` fixture (writes the row the way the settings screen does, flushes waffle's cache on teardown, and refuses a flag that is already on for everyone); the hand-rolled fixtures that existed to dodge the gap now use `override_flag`, which works for team-scoped checks from this change on; a precedence table over the `everyone` x membership matrix plus form-save transition tests.

First of three PRs on #4321: this one fixes precedence, the next reduces the flag inputs in `/admin/flags`, the last migrates `everyone=False` to `None` and turns `everyone` into a working tri-state. Related to #4321.

### Migrations

No migrations.

### Demo

No UI changes. The fix is observable in Settings > Feature flags: enabling Hybrid search there now changes `Collection.hybrid_search_enabled` for that team's collections, which the new retrieval test pins.

### Docs and Changelog

### Operator Impact

Flag rows with `everyone=True` now activate their feature on team-scoped paths, where they were previously ignored. Operators should review any `everyone=True` rows to confirm a global rollout is intended; `everyone=False` rows and rows with only team assignments behave exactly as before.


apps/documents/models.py
```diff
@@ -518,7 +518,7 @@ def _build_contextualizer(self):
                 },
             )
             return None
-        if not self._flag_active_for_team(Flags.CONTEXTUAL_RETRIEVAL):
+        if not Flag.get(Flags.CONTEXTUAL_RETRIEVAL.slug).is_active_for_team(self.team):
```
snopoke: A small utility to match `waffle.flag_is_active` could be useful:

```python
flag_is_active_for_team(team, 'flag')
```

snopoke (dismissed):
One thought but otherwise LGTM

apps/documents/models.py
barry47products: Good call, added in bfce75442

snopoke (approved):

PR #4369, opened 2026-09-02
Reduce flag admin inputs to everyone and teams

barry47products:
Related to #4321. Second of three PRs; step 1 is #4359, step 3 follows this one.

### Product Description

A feature flag is now configured by two settings: on for everyone, or on for specific teams. The flag admin no longer offers the settings that only work when there is a browser request to inspect (specific users, superusers, staff, authenticated, testing mode, percentage rollout, languages, groups), because team-scoped flag checks - indexing, retrieval, Celery tasks - have no request to read them from. Those settings appeared to work in the admin while silently doing nothing on the paths that matter, so a flag could read as on in one place and off in another.

Existing flags keep their `everyone` and team assignments. The values stored for the removed settings are reset, so a flag decision now reads the same way everywhere.

Superusers no longer get automatic access to every flag on request paths. Per #4321: anywhere superusers must always see something, gate it with an `is_superuser` check rather than a flag.

### Technical Description

`FlagUpdateForm` and `update_flag` accept and write only `everyone` and `teams`. The flag list and detail templates drop the request-only badges, controls, the users section, and the matching Alpine form state. `FlagAdmin` excludes the same fields from the Django admin and trims its list display and filters to match.

The columns stay on the model - they come from waffle's `AbstractUserFlag` - so writes to a neutralised field remain possible at the DB level. `FLAG_FIELDS` is therefore extended with `staff`, `authenticated`, `languages` and `groups` rather than trimmed: the audit log is the tripwire that records any such write. The flag history view name-maps `groups` deltas alongside `teams` and `users` so those entries stay legible.

Why this precedes step 3: with `everyone=None`, waffle's `is_active` reaches `_is_active_for_user`, where `superusers=True` (waffle's field default, set on every row) returns True for superusers. Step 3 migrates `everyone=False` to `None`; without this PR first, that migration would hand superusers every flag on request paths. After this PR, `None` and `False` resolve identically and step 3's migration is a behavioural no-op.

Known limitations, all deliberate:

- The data migration runs on historical models and so bypasses `field_audit`. The neutralisation is not recorded as audit events.
- The cache flush runs inside the migration transaction, so a concurrent request can miss the cache, read the pre-migration rows and re-cache them until TTL. Waffle's own save-time flush has the same window.
- The reverse migration is a no-op: the old values are deliberately unrecoverable. `elidable=True` because the operation is safe to drop on squash.
- `users_api` (`apps/admin/urls.py`) loses its only consumer with the users selector; it is left in place rather than widening this PR, and is now referenced only by the rate-limit tests.

### Migrations

`teams/0016_neutralise_request_only_flag_fields` sets `superusers=False`, `staff=False`, `authenticated=False`, `testing=False`, `rollout=False`, `percent=None`, `languages=""` and clears the `users` and `groups` M2Ms, leaving `everyone` and `teams` untouched. No schema change, so old and new code both run against migrated rows.

It flushes waffle's per-flag cache keys for each row. Waffle caches whole `Flag` instances plus the user, group and team ID sets under separate keys, so without the flush a stale copy with `superusers=True`, or a cleared user list, would keep answering flag checks until TTL. Every flush key derives from the flag name alone, which is why an unsaved live-model instance can compute them for historical rows - and using the live model includes the OCS-specific teams key that a historical model would miss.

### Demo

| | |
| --- | --- |
| **Flag list.** Only the `Everyone` badge is left. The badges for testing, superusers, staff, authenticated and rollout are gone, and a flag's scope now reads as its `Everyone` badge plus its team count. | Flag list |
| **Flag detail.** The form offers `Everyone` and `Teams`, and nothing else. Change History records each `everyone` transition, so a write to a flag stays traceable. | Flag detail |
| **Django admin change form.** The same two settings. The request-only fields are excluded here too, so the Django admin cannot reintroduce what the flag admin no longer offers. | Django admin change form |

### Docs and Changelog

### Operator Impact

Two behaviour changes on deploy:

1. Superusers lose automatic access to flags on request paths. An operator relying on a superuser account to preview gated features will stop seeing them unless the flag is enabled for `everyone` or for their team.
2. Values previously stored for the request-only settings are reset. An operator using percentage rollout, testing mode, or per-user flag assignment loses those configurations, and the admin no longer offers them.


apps/teams/migrations/0016_neutralise_request_only_flag_fields.py
```diff
@@ -0,0 +1,47 @@
...
+    copy (notably `superusers=True`, waffle's field default) would keep answering flag
+    checks until its TTL. The flush keys derive from the flag name alone, which is why an
+    unsaved live-model instance can compute them for the historical rows.
+    """
+    flag_model = apps.get_model("teams", "Flag")
+    flush_keys = set()
+    for flag in flag_model.objects.all():
+        for field, value in NEUTRALISED_FIELDS.items():
+            setattr(flag, field, value)
+        flag.save(update_fields=list(NEUTRALISED_FIELDS))
```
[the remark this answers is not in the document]
barry47products: @coderabbitai Narrower than described in this PR and handled in the stacked follow-up.

`everyone=False` returns before `_is_active_for_user` runs, so `superusers` is only read when `everyone is None`. Both creation paths write `False` - `FLAG_DEFAULT` is unset, and the team settings screen passes it explicitly - so as of this PR, a superuser outside a flag's teams cannot activate one.

It matters from #4370, which migrates `everyone=False` to `None`. That PR re-runs this neutralisation over every row and pins `defaults={"everyone": None, "superusers": False}` on the create-missing-flag path, so new flags are not minted with superuser access.

apps/teams/migrations/0016_reduce_flag_inputs_to_everyone_and_teams.py
```diff
@@ -0,0 +1,47 @@
...
+    unsaved live-model instance can compute them for the historical rows.
+    """
+    flag_model = apps.get_model("teams", "Flag")
+    flush_keys = set()
+    for flag in flag_model.objects.all():
+        for field, value in NEUTRALISED_FIELDS.items():
+            setattr(flag, field, value)
+        flag.save(update_fields=list(NEUTRALISED_FIELDS))
+        flag.users.clear()
+        flag.groups.clear()
```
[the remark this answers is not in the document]
barry47products: @coderabbitai Confirmed - the irreversible clear is intended, agreed on #4321 before this was written.

These settings only resolve when there is a request to inspect, so on team-scoped checks they read as configured while doing nothing. Leaving the values in place would keep that discrepancy alive, so the reverse is a deliberate no-op. Disclosed in the PR body under Migrations and Operator Impact.

apps/teams/model_audit_fields.py
```diff
@@ -1,3 +1,16 @@
 TEAM_FIELDS = ["name", "slug", "created_by", "public_key", "metadata", "is_migrating"]
 MEMBERSHIP_FIELDS = ["team", "user"]
-FLAG_FIELDS = ["name", "everyone", "percent", "testing", "superusers", "rollout", "teams", "users"]
+FLAG_FIELDS = [
+    "name",
```
SmittieC: should the fields that cannot be changed from the UI still be audited?

apps/teams/tests/test_flag_audit.py
```diff
@@ -0,0 +1,57 @@
...
+        pytest.param("superusers", False, id="superusers"),
+        pytest.param("staff", True, id="staff"),
+        pytest.param("authenticated", True, id="authenticated"),
+        pytest.param("testing", True, id="testing"),
+        pytest.param("rollout", True, id="rollout"),
+        pytest.param("percent", 50, id="percent"),
+        pytest.param("languages", "en", id="languages"),
+    ],
+)
+def test_writes_to_request_only_fields_are_audited(field, value):
```
SmittieC: why is this?

SmittieC (dismissed):

apps/teams/migrations/0017_everyone_false_to_none.py
```diff
@@ -0,0 +1,49 @@
...
+    the old values are deliberately unrecoverable. Cache flushing works as in 0016: flush
+    keys derive from the flag name alone, so an unsaved live-model instance can compute
+    them for the historical rows.
+    """
+    flag_model = apps.get_model("teams", "Flag")
+    flush_keys = set()
+    for flag in flag_model.objects.all():
+        if flag.everyone is False:
+            flag.everyone = None
+        for field, value in NEUTRALISED_FIELDS.items():
```
SmittieC: @barry47products Since these migrations haven't run yet, can we merge this one and 0016?

SmittieC (dismissed):

apps/teams/migrations/0010_create_flag_ai_cost_monitoring.py
snopoke, line 1: Side note: We should drop this migration (we can just remove it's operations) since it isn't necessary to create feature flags via migrations.

apps/teams/migrations/0017_everyone_false_to_none.py
snopoke: +1 to merge these together if possible

snopoke (commented):
LGTM pending the migration merge.

As a followup to this work I think it would be useful to create an ADR with the decisions to use a tri-state flag and remove support for the other fields.

apps/teams/migrations/0017_everyone_false_to_none.py
barry47products: Done in 45b272519. 0017 is folded into 0016, now `0016_reduce_flag_inputs_to_everyone_and_teams`, since it repeated 0016 in full and added only the everyone False to None line.

apps/teams/migrations/0010_create_flag_ai_cost_monitoring.py
barry47products: @snopoke do you want to include here, or should I open a new PR

barry47products:
> As a followup to this work I think it would be useful to create an ADR with the decisions to use a tri-state flag and remove support for the other fields.

I'll do a follow-up PR covering both.

apps/teams/migrations/0010_create_flag_ai_cost_monitoring.py
snopoke: A separate PR

snopoke (approved):

PR #4373, opened 2026-09-02
feat: a merged Models tab and persisted LLM credential verification

SmittieC:
### Product Description

The LLM provider page gets one role-filtered **Models** tab in place of three tables (default chat, custom, and embedding models).

Credential verification already ran on save, but its result was a flash message that vanished on the next request. The page now says where the credentials stand and keeps saying it: verified, never checked, or rejected with the provider's own words. The save button says up front whether saving will verify. The manual **Test Connection** button is gone — it tested whatever was in the database rather than what was on screen, and saving again now re-runs a failed check.

### Technical Description

Verification state moves into a new `LlmProvider.extra_data` bag (`verified_credentials`, `verification_error`). Two things worth a look:

- **`extra_data` is part of the team export.** The export serializer takes every field by default, so it joins the export contract — deliberately: verification state travels with the credentials it describes, since a key the provider accepted is still that key on the target. It was excluded at first, which left an existing target row holding its own verification result over credentials a re-import had overwritten; `test_import_carries_provider_verification_state` covers both the created and the updated row.
- **`classify_connection_test_failure` is gone.** Failures surface the provider's own error text rather than a category, truncated to 2000 chars since it renders on the page.

`_should_test_connection` now also fires when credentials have never passed — that is what makes the manual button redundant, at the cost of an external call on every save of a still-unverified provider. Voyage AI is embeddings-only and is never offered verification.

The Models tab filters rendered rows in the browser rather than re-querying, with the counts derived from the same pass that does the filtering, so a chip can never disagree with the list it filters. Row changes arrive over htmx and don't re-render the wrapper, so the pass re-runs off a `MutationObserver` — an htmx event won't do, because a delete swaps the row away and `htmx:afterSwap` fires on the row as it detaches.

Commits are split by concern and are individually green.

### Issue Link

None.

### Migrations

`0078_llmprovider_extra_data` adds a **nullable** jsonb column. Reversible and non-locking (metadata-only `ADD COLUMN` on PG 11+, no table rewrite), no manual steps.

Nullable rather than `NOT NULL` because Django's `AddField` drops the column default straight after adding it, so the previous release's inserts — which omit the column — would hit a not-null violation for the length of the deploy window (AGENTS.md, b15fe0c03). Every read already goes through `(self.extra_data or {})`, so no calling code changed. Verified against Postgres: after migrating, an INSERT omitting `extra_data` succeeds and leaves `NULL`; tests cover `NULL` reading as unverified and a `NULL` row surviving its first check.

### Demo

https://www.loom.com/share/45ad4458394646c98568c7255e965e5e

### Docs and Changelog

Two user-facing changes: the Models tab replacing the three model tables, and credential verification becoming visible state on the page with the manual Test Connection button removed.

### Operator Impact

Carries a migration, but it is backwards compatible and needs no operator action, so there is nothing for an operator to know or do on upgrade.


apps/teams/export/manifest.py
```diff
@@ -123,6 +123,10 @@ class ManifestEntry:
     "users.customuser": ["password", "user_permissions", "is_staff", "is_superuser"],
     # System-managed OAuth token cache; the target refetches it from the copied client credentials.
     "service_providers.authprovider": ["_auth_data"],
+    # Credential verification state describes the credentials as stored on this instance: the
+    # verified flag says nothing once they are re-sealed for another team, and the stored provider
+    # response is raw error text from an external service, which has no place in an export payload.
+    "service_providers.llmprovider": ["extra_data"],
```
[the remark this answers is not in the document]
SmittieC: No, this should not be reset. A credential set is valid (or not) no matter on which server it sits. Please do not ignore this field

apps/service_providers/migrations/0077_llmprovider_extra_data.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("service_providers", "0076_add_claude_fable_5_1"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="llmprovider",
+            name="extra_data",
+            field=models.JSONField(blank=True, default=dict),
```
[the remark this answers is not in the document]
SmittieC: Fixed in 7b658ac2f. The migration now restores a database-level default after Django drops the one it adds the column with.

Verified on a throwaway database: `sqlmigrate` confirms the `SET DEFAULT` lands after Django's `DROP DEFAULT`; after migrating, `information_schema` reports `column_default = '{}'::jsonb` with `is_nullable = NO`, and an INSERT that omits `extra_data` (what previous-release code emits) succeeds and gets `{}`. Reverse and re-apply both work.

The CHANGELOG entry that recorded this as a known deploy-window breakage is updated in d4cbfe56f.

*— Claude Code*

apps/teams/export/manifest.py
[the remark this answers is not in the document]
SmittieC: Addressed in 9372827e4, but the other way round — following the author's call on this thread rather than the suggestion.

The inconsistency is real: `_build_values` stripped `extra_data` (importer.py:356) and `_collect_concrete_field` only assigns fields present in the row (importer.py:410), so a re-import overwrote `config` while an existing target row kept its own verification result. Resetting is one way to close that; exporting the field is the other, and it is the one that matches the author's position that a credential set is valid or not regardless of which server it sits on.

So `service_providers.llmprovider` is out of `EXCLUDE_REGISTRY` entirely — `extra_data` now travels with `config` and is overwritten alongside it. `test_import_carries_provider_verification_state` covers the created and the updated row, and fails if the exclusion is put back. `api-schemas/export.yml` regenerated.

A reset commit (22960fe3c) was pushed first and is reverted by 9372827e4.

*— Claude Code*

templates/service_providers/provider_form.html
```diff
@@ -17,21 +17,42 @@
...
-    </div>
+  <div class="app-card max-w-5xl mx-auto"{% if can_verify_credentials %} x-data="{ configChanged: false }"{% endif %}>
+    <h1 class="pg-title">{{ title }}</h1>
+
+    {% comment %}
+      The provider's own words about the credentials saved here, from whenever they were
+      last checked. Above the tabs because it is about the provider, not about whichever tab
+      you happen to be on. Stays until a check passes, so the reason is still here after a
+      reload; saving successfully clears it.
+    {% endcomment %}
```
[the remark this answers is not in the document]
SmittieC: Partly declined, one part fixed in e0e9b091d.

Not applicable as written: these are already `{% comment %}{% endcomment %}` blocks, which is what this repo's AGENTS.md requires — a `{# ... #}` comment "must not be spread over multiple lines". Collapsing them to inline comments would mean either dropping the context or breaking that rule. The content is also the kind AGENTS.md says earns its place: an unselected daisyUI tab panel is `display:none` (which is why usages load on `intersect once` and why modals sit outside the tab wrapper), and Alpine does not bind directives on htmx-swapped content. A reader of the markup alone could not recover any of that.

One genuine problem inside the same set, though: line 1 said "Replaces generic/type_select_form.html", which narrates the diff rather than describing the code as it stands — explicitly called out in AGENTS.md. Removed.

*— Claude Code*

apps/service_providers/migrations/0077_llmprovider_extra_data.py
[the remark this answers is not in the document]
SmittieC: Reverted in 611782af0 — following repo convention instead: 15 other migrations in this codebase add a NOT NULL JSONField(default=dict) column with no database default, and this one now matches them. The deploy-window gap is real and accepted, not mitigated; CHANGELOG.md and the PR description are back to recording it as not backwards compatible.

*— Claude Code*

templates/service_providers/components/llm_models_tab.html
```diff
@@ -0,0 +1,96 @@
...
+  One pass owns both the filtering and the counts, reading the rendered rows, so a chip can
+  never disagree with the list it filters. Adding a custom model replaces the rows and
+  deleting one removes a row, neither of which re-renders this wrapper, so it re-runs on
+  any change to the row list. That is watched with a MutationObserver rather than an htmx
+  event because a delete swaps the row away, and htmx:afterSwap fires on the row as it
+  detaches - too late to bubble anywhere useful. Rows are hidden from here rather than by a
+  per-row x-show because Alpine does not bind directives on htmx-swapped content.
+{% endcomment %}
+<div x-data="{
+       role: 'all',
```
zandre-eng: Nit: This is a lot of JS code in the HTML element. What are your thoughts about having this in a `<script>` element instead? This could make it easier for the IDE to parse and identify JS issues as well.

zandre-eng (commented):
I'm not too well-versed with the codebase to be able to review the approach taken here, but the code itself seems generally fine. One small nit from my side.

templates/service_providers/components/llm_models_tab.html
```diff
@@ -0,0 +1,105 @@
...
+      We use
+      <a href="https://llm-stats.com" target="_blank" rel="noopener noreferrer" class="link link-hover font-semibold">llm-stats.com</a>
+      to source new model releases as they become available.
+    </span>
+  </div>
+
+  <div class="flex flex-wrap items-center justify-between gap-3 mt-4">
+    <div class="join" role="group" aria-label="Filter by role">
+      {% for key, label in role_filters %}
+        <button type="button" class="btn btn-sm join-item"
```
snopoke: Would be good to keep this consistent with the filter pills on the main team settings page: https://github.com/dimagi/open-chat-studio/blob/d44421f68bc597107dd3b008fab24c546ba882e5/templates/teams/components/integrations_table.html#L7

Maybe we can make that template into a component?

templates/service_providers/components/llm_model_row.html
```diff
@@ -1,78 +1,91 @@
...
+    {% if row.is_llm and not row.deprecated %}
+      <button class="btn btn-xs btn-ghost"
+              title="Override pricing for this team"
+              hx-get="{% url 'service_providers:pricing_override_form' request.team.slug row.id %}"
+              hx-target="#pricing_override_modal_body"
+              hx-swap="innerHTML"
+              onclick="pricing_override_modal.showModal()">
+        <i class="fa-solid fa-pencil"></i>
+      </button>
+      {% if row.rates.has_team_override %}
```
snopoke: This doesn't seem to be working. If you add a custom model and set pricing this button still shows. Clicking it returns the row to 'unpriced' with the warning.

templates/service_providers/components/llm_model_row.html
```diff
@@ -1,78 +1,91 @@
...
+    {% if row.deprecated %}class="text-base-content/50"{% endif %}
+    data-role="{{ row.role }}"
+    data-custom="{{ row.custom|yesno:"true,false" }}"
+    data-deprecated="{{ row.deprecated|yesno:"true,false" }}"
+    {# Matches the tab's default filter, so deprecated rows don't flash before it applies. #}
+    {% if row.deprecated %}hidden{% endif %}>
+  <td class="font-mono">{{ row.name }}</td>
+  <td>
+    <span class="badge badge-xs badge-outline">{{ row.role }}</span>
+    {% if row.custom %}<span class="badge badge-xs badge-outline">custom</span>{% endif %}
```
snopoke: nit: add a tooltip to this badge: "This model was added manually and is not managed by the platform" or something.

snopoke (commented):
Looks great. A few small comments. Also where does the 'verified' state show up? I couldn't find it anywhere in the ui

templates/service_providers/components/llm_models_tab.html
SmittieC: Done in 915c32640 — moved to a `<script>` block registering `llmModelFilters` on `alpine:init`, which is the pattern the other 13 Alpine components in `templates/` already use. The element is just `x-data="llmModelFilters"` now, and the two `{% comment %}` blocks that explained the JS became `//` comments inside it, next to the code they describe.

Worth noting for anyone doing this elsewhere: it only works because the Models tab renders with the page. If the tab were htmx-swapped in, `alpine:init` would have already fired and the registration would never run.

Verified in the browser rather than just by eye — filters (All 25 / Chat 22 / Embedding 3 / Custom 1), the deprecated toggle (25→35), and the MutationObserver re-applying the active filter after both an add (rows replaced) and a delete (single row swapped away). Counts stayed in sync, no console errors.

*— Claude Code*

templates/service_providers/components/llm_model_row.html
SmittieC: Confirmed and fixed in 45eac9a41.

The guard was `has_team_override`, which is true for any team-scoped rule. But reverting closes the team's rules and re-resolves against the *global* ones — and a custom model has no global rule to land on, so the revert emptied the row's pricing instead of falling back. A delete wearing a revert's label, exactly as you saw. Same applies to a global model that only ever had team pricing.

`has_team_override` is now `can_revert` — team override **and** a global rate to fall back to — which is the question the template was really asking. `_pricing_lookup` reads the global keys off the rules before team-scoped ones overwrite them on the same key.

Verified in the browser: a custom model with team pricing renders no revert button; `gpt-4o-mini` with a team override still does, and clicking it falls back $1.00 → $0.15/1M, drops the `manual` badge and removes the button. Five tests, all confirmed failing before the fix.

Removing pricing from a custom model is still possible — clear the fields in the override form and save.

*— Claude Code*

templates/service_providers/components/llm_model_row.html
SmittieC: Added in d17cef660: `data-tip="Added by this team, not managed by the platform"`.

Left the `deprecated` badge alone — I didn't want to assert what deprecation means for an existing configuration without checking.

*— Claude Code*

templates/service_providers/components/llm_models_tab.html
SmittieC: Styling matched in f72aff166 — `flex flex-wrap gap-2` wrapper, `btn btn-xs btn-soft btn-primary rounded-full` pills, count as `<span class="ml-1 opacity-60">`. The `join` group and the neutral count badges are gone, so the two rows read as the same control now.

I did **not** extract a shared component, for two reasons:

- `templates/teams/components/integrations_table.html` isn't on this branch — it landed on `main` while this PR was open. (Now merged in, but it wasn't when I made the change.)
- The two are only superficially the same. Yours are htmx links that reload the table with a `category` query param and server-computed counts. These filter already-rendered rows in the browser and derive their counts from the DOM, which is deliberate — it's what keeps a chip from ever disagreeing with the list it filters. A Django `{% include %}` can't carry per-pill behaviour, so a shared partial would end up presentation-only: effectively a shared class list with an awkward call signature at both sites.

Happy to do the extraction as a follow-up if you still want it — or if you'd rather I find a shape that carries both mechanisms, say so and I'll take another run at it.

*— Claude Code*

apps/service_providers/migrations/0077_llmprovider_extra_data.py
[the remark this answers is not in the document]
SmittieC: Correcting my earlier reply on this thread: the deploy-window gap is no longer accepted, and the migration is backwards compatible as of 075fad9ba.

What changed since that reply is that `main` gained an explicit rule in AGENTS.md (b15fe0c03): *"Adding a `NOT NULL` column is the usual trap — Django's `AddField` drops the DB default afterwards, so the running release's inserts, which omit the column, fail. Add it `null=True`, or keep a DB-level default."* This migration was exactly that case, so "15 other migrations do it this way" stopped being a good reason.

Took the `null=True` branch rather than the `RunSQL` default suggested here. Every read already went through `(self.extra_data or {})`, so no calling code changed; the field, the migration and the regenerated `api-schemas/export.yml` are the whole diff.

Verified against Postgres rather than by inspection: after migrating, an INSERT that omits `extra_data` — what the previous release emits — succeeds and leaves `NULL` (rolled back afterwards). Covered by tests too: `NULL` reads as unverified, and a `NULL` row survives its first credential check.

The migration is also renumbered to `0078_llmprovider_extra_data`, since `0077` was taken by `0077_deprecate_gpt5_family` on `main`.

The CHANGELOG entry that recorded this as an operator-visible breakage is removed in b3ecf2699 — there's nothing left for an operator to know or act on.

*— Claude Code*

apps/service_providers/models.py
```diff
@@ -317,50 +292,66 @@ def test_connection(self) -> None:
...
-                f"Provider saved, but the connection test failed. {reason} Use Test Connection below to retry."
-            )
-        return warnings
+            log.exception("Could not verify credentials for LLM provider %s", self.pk)
+            warnings.append("Provider saved, but the credentials could not be verified.")
+            detail = _error_detail(exc)
+        else:
+            verified = True
+        self._record_connection_test(verified, detail)
+        return warnings, detail
```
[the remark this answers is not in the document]
SmittieC: The observation is correct, the suggested fix isn't. Not changing this.

**The observation.** Confirmed by running it: with `{"verified_credentials": False, "verification_error": "..."}` stored, a save that raises `NoTestableModelError` leaves `{"verified_credentials": False}` — the reason is gone. So the report is accurate about what happens.

It is also deliberate, not an oversight: `_record_connection_test`'s docstring says a check "that never reached the provider" clears the reason rather than leaving a stale one, and `test_no_configured_model_stores_no_response` pins it.

**Why the suggested fix is worse.** Returning early without calling `_record_connection_test` also preserves `verified_credentials`, and that flag is what decides whether a later save re-checks. I applied the suggestion on a scratch copy to be sure:

- provider verified (`verified_credentials: True`), user replaces the API key, no model configured for that type
- `_should_test_connection` fires because the config changed, `test_connection` raises `NoTestableModelError`, early return
- result: `{"verified_credentials": True}` — the **new, never-tested** key is now recorded as verified

And it stays that way: `_should_test_connection` is `config changed or not verified`, so with the config now matching what's saved and the flag reading True, no future save re-checks it. Trading a lost error string for silently trusting unverified credentials isn't a good trade.

**What's actually left.** There is a narrower real issue: whether clearing the reason is right depends on something the hook can't see. If the credentials are unchanged (the retry path), the old reason still describes them and clearing it loses information — that's the report's case. If the user just replaced the key, the old reason describes a key that no longer exists and *must* be cleared, which is what today's code gets right. `_has_config_changed` is computed in `CreateServiceProvider` and isn't passed down, so distinguishing them means plumbing that signal into `run_connection_test_hook`.

Low value for the cost, as I read it: the user gets a warning naming the exact next step ("Add one on the Models tab, then save again"), and the save after that produces a fresh, accurate error. Happy to plumb it through if a maintainer disagrees.

*— Claude Code*

SmittieC:
@snopoke on the review question — *"where does the 'verified' state show up? I couldn't find it anywhere in the ui"*:

It didn't. You weren't missing it. The only trace of the stored flag was the submit button reading "Update" instead of "Update and Verify", which asks the reader to infer the state from the offer to change it. `test_views.py` even had a docstring admitting it: *"The label is the only place the stored flag surfaces - there is no badge."*

Fixed in 2611f20e5. There's now a line above the tabs, in the same slot as the rejection alert, so the two are one statement about where the credentials stand:

- verified → green check, "Credentials verified"
- never checked, or last check failed with nothing to quote → muted, "These credentials have not been checked yet - saving will check them"
- provider rejected them → the existing error alert, unchanged

It's hidden as soon as a credential field is edited, since from then on the stored flag describes something other than what's on screen. Provider types that can't be checked at all (Voyage AI) show nothing — there's no state to report.

Four tests cover the three states plus the untestable type.

*— Claude Code*

apps/service_providers/views.py
```diff
@@ -424,19 +410,67 @@ def get_success_url(self, obj=None, redirect_to_own_edit_page=False):
...
 
 
-def _get_models_by_type(queryset):
-    models_by_type = defaultdict(list)
-    for model in queryset:
-        models_by_type[model.type].append(model)
-    # Deprecated models sink to the bottom; otherwise newest first.
+def _active_provider_tab(request) -> str:
+    tab = request.GET.get("tab")
+    return tab if tab in VALID_PROVIDER_TABS else "configuration"
```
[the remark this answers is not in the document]
SmittieC: Confirmed and fixed in 3a14267c6. Good catch — reproduced exactly as described.

Wrote the failing test first: `GET /a/<team>/service_providers/voice/<pk>/?tab=models` renders **zero** checked radios, and `_checked_tab()` fails with `expected exactly one checked tab, got []`. Then drove it in a browser to confirm the CSS consequence, since a template test can't see `display: none` — the page renders the heading, the Configuration/Usages tab strip, and no form.

Took the suggested approach: the param is now resolved against the tabs the page will actually render rather than a global tuple. `_get_context` builds the available set (`configuration` always; `models` only for LLM; `usages` only with the permission — and neither on a create page, where the tab strip isn't rendered at all) and passes it to `_active_provider_tab`, which falls back to `configuration` for anything else. `VALID_PROVIDER_TABS` had no other callers and is gone.

Parametrised over a voice provider as suggested: `""`, `?tab=models` and `?tab=nonsense` all open Configuration, and `aria-label="Models"` is absent. Browser check across all four cases:

```
(none)           checked=["Configuration"]  config form visible=true
?tab=models      checked=["Configuration"]  config form visible=true
?tab=nonsense    checked=["Configuration"]  config form visible=true
?tab=usages      checked=["Usages"]         config form visible=false
```

Agreed on the `can_view_usages` variant not being reachable with the default roles; the same fix covers it regardless, since the available set is built from the permission.

*— Claude Code*

apps/service_providers/models.py
```diff
@@ -317,50 +292,66 @@ def test_connection(self) -> None:
...
+        extra_data = {**(self.extra_data or {}), VERIFIED_CREDENTIALS_KEY: verified}
+        if detail:
+            extra_data[VERIFICATION_ERROR_KEY] = detail
+        else:
+            extra_data.pop(VERIFICATION_ERROR_KEY, None)
+        self.extra_data = extra_data
+        self.save(update_fields=["extra_data"])
 
 
 class LlmProviderModelManager(models.Manager):
```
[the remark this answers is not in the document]
SmittieC: Real, and fixed in 0594e81e2.

Reproduced deterministically rather than by inspection: load an instance, write to `extra_data` through a second queryset, then run the hook on the stale instance. The concurrent write is gone — `{'verified_credentials': True}` instead of `{'something_else': ..., 'verified_credentials': True}`. `test_a_write_that_landed_during_the_check_is_not_clobbered` covers it and fails without the fix.

Worth adding that the window is unusually wide here, not the microseconds a read-modify-write usually spans: `_record_connection_test` is called after `test_connection()`, which is a synchronous external LLM request with a 10s timeout. The instance's `extra_data` is whatever was loaded before that call.

Took the suggested approach — `select_for_update()` inside `transaction.atomic()`, matching `MessagingProvider._update_extra_data`. The lock is only around the read-modify-write, not the external call: `run_connection_test_hook` deliberately runs outside a transaction so a multi-second provider request never holds a connection open, and that still holds.

*— Claude Code*

templates/service_providers/llm_provider_form.html
```diff
@@ -1,89 +1,48 @@
...
+        </div>
+        <div class="modal-action">
+          <form method="dialog">
+            <button class="btn btn-primary"
+                    hx-post="{% url "service_providers:llm_provider_model_new" request.team.slug %}"
+                    hx-include="#new_model_form"
+                    hx-target="#custom_model_list"
+                    hx-swap="innerHTML"
+                    hx-on::after-request="new_custom_model.close()">
+              Save
```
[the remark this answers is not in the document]
SmittieC: Fixed in c29b4475e, though the diagnosis needed one correction and the suggested fix on its own wouldn't have worked.

**The bug is real.** The dialog closes on every save attempt and takes the entered values with it, so a rejected save leaves you with nothing to correct.

**But `hx-on::after-request` was only half of it.** The Save button was a plain `<button>` inside `<form method="dialog">`, so it is a submit button, and submitting a dialog form closes the dialog *natively on click* — before the response is anywhere near coming back. Making the handler conditional would have left that path untouched and the dialog would still have closed. The fix is `type="button"` **and** `if (event.detail.successful)`. The regression test asserts both, and fails if either is reverted.

**One correction:** *"The user is not made aware of the error"* isn't right. `templates/web/components/ajax_errors-js.html` handles `htmx:responseError` globally and shows the response text in an alertify alert, so the message does surface. It surfaces badly — the endpoint returns `str(form.errors)`, so what you get is a raw Django `ErrorDict` in a modal rather than errors against the fields — but it isn't silent.

Verified in a browser: an invalid save now leaves the dialog open with the typed values intact; correcting the name and saving closes it and adds the row.

While chasing this I also went after an apparent whole-page wipe on the 400 — worth recording that it's **not a bug**: `django-htmx.js` replaces `document.children[0].innerHTML` with the response body on a 400/403/404/500, guarded by `DEBUG`, so Django's technical error page is readable for htmx requests. Dev-only, by design.

*— Claude Code*

apps/service_providers/views.py
```diff
@@ -339,19 +332,35 @@ def _save_provider(self, request, primary_form, config_form, file_formset):
...
+        return f"{verb} and Verify"
 
     def _get_context(self, primary_form, config_form, subtype, instance):
+        can_view_usages = self.request.user.has_perm(self.provider_type.get_permission("view"))
+        available_tabs = {"configuration"}
+        if instance:
+            if self.provider_type == ServiceProvider.llm:
+                available_tabs.add("models")
+            if can_view_usages:
+                available_tabs.add("usages")
```
SmittieC: Fixed in 327fdf1a8.

Fair flag — I put this there. Resolving the tab against the tabs the page actually renders (3a14267c6) added three branches to `_get_context`, which took it to the threshold.

The branches have moved into `_active_provider_tab`, which is where they belong: it is the function that consumes them, and the caller passes what it already had to hand. `_get_context` is back to one call with no added branching, and `_active_provider_tab` stays well under on its own.

*— Claude Code*

SmittieC:
I also updated the `usages` page to use chips

image

apps/service_providers/tests/test_pricing_override.py
```diff
@@ -116,6 +116,22 @@ def test_rejects_prices_outside_the_unit_price_domain(self, price):
         assert not PricingRule.objects.filter(team=team, model_name="test-pe").exists()
 
 
+@pytest.mark.django_db()
+class TestOverrideModalOpening:
```
snopoke: probably not a useful test

snopoke (dismissed):

apps/service_providers/tests/test_pricing_override.py
SmittieC: https://github.com/dimagi/open-chat-studio/pull/4373/commits/9cf683b5eeb69e912bcb77c8ec643fda297943b4

snopoke (approved):

PR #4379, opened 2026-09-02
Redesign session details page

### LINKED ISSUE #4232: Redesign session details page
Primary goals:

* Reduce space taken up by the details
* Present the details in a cleaner way that highlights important parts

[Session Detail (standalone).html](https://github.com/user-attachments/files/31223948/Session.Detail.standalone.html)

[Claude Design](https://claude.ai/design/p/263087cd-89f0-425e-96da-3e50ace56462?file=Session+Detail.dc.html&via=share)

Image

Image

Image

Image

Image

preston-56:
### Product Description
Redesigns the chatbot session detail page to match the approved mockup, so it takes up less space and surfaces the important parts of a session up front instead of burying them in a long list.

- Compact header: title, status badge, bot/participant/start-date info line, and message/usage/duration/tags stats all in one row, with the rest of the metadata (Session ID, Remote ID, Ended, Platform, Annotation Queues, Comments) tucked behind a collapsed "Session details" toggle.
- Messages now render as chat bubbles (avatar, timestamp, hover-revealed actions for trace/copy-link/tags/annotation) instead of a plain table.
- Participant Data is now editable in place (Format/Revert/Save, live JSON validation) instead of a read-only dump.
- Participant Schedules and Chatbot Events are now real, dense tables instead of stacked cards, each with a "Manage"/"Log" action that opens the details in a modal.
- Session export ("Export" button on the Messages tab) now produces a real per-session CSV download instead of being a disabled placeholder.

### Technical Description
- `apps/generics/views.py` (`render_session_details`) is the single context-builder behind the page; it now also supplies `participant_chip`, `session_start`/`session_duration_display` (compact duration via `apps.utils.time.seconds_to_human(..., compact=True)`), `message_count`, `participant_data`/`participant_data_updated_at`, and `has_event_logs`.
- Participant data is read via `ParticipantData.objects.for_experiment(experiment)` rather than a raw `experiment=` filter — `ParticipantData` rows are always keyed to the *working* version's id, so a session running a published version needs the same version resolution every other caller in the codebase already uses. The save path (`EditParticipantData.post` in `apps/participants/views.py`) needed the identical fix, or editing from this page would silently write an orphaned row the live chat runtime never reads.
- The participant-data JSON editor is the existing shared `JsonEditor` component (`assets/javascript/editors.js`); this page adds a scoped `<style>` block to flatten its line numbers/syntax highlighting to match the mockup's plain muted look, and a `formatEditor` custom event so an external header button can trigger the same reformat the editor already does internally.
- New Session / End Session buttons are always shown (per the mockup) but `disabled` with an explanatory tooltip when the action isn't valid for this session (web-platform session, disabled channel, or already-ended session) — consistent with how the rest of the page treats unavailable actions.
- Session export reuses the existing bulk-export helpers (`apps/experiments/export.py`) scoped to a single session, behind the existing `experiments.download_chats` permission.
- Two small site-wide changes ride along with this page's needs: `app-card` (`assets/styles/site-tailwind.css`) gained a visible border/radius/background (previously invisible on `main`; already applied and reviewed on a sibling branch), and the app's outer `xl:container` (`templates/web/app/app_base.html`) no longer caps out at 1536px, so wide screens use the full available width instead of leaving a large unused margin on either side.

Closes #4232

### Migrations
No migrations.

### Demo
See attached screenshots and `session-detail-demo.gif` (walks through every tab: Messages, Session details expanded, Participant Data, Participant Schedules + Manage modal, Session State, Chatbot Events + Log modal).

Messages tab

Session details expanded

Participant Data tab

Participant Schedules tab

Manage schedule modal

Session State tab

Chatbot Events tab

Event log modal

Full walkthrough

### Docs and Changelog

Note for changelog: user-facing redesign of the chatbot session detail page (issue #4232) plus a new per-session CSV export. No behavior change to existing exports/APIs.

### Operator Impact


apps/generics/views.py
```diff
@@ -42,29 +53,28 @@ def render_session_details(
...
-                for trigger in experiment.event_triggers
-            ],
+            "event_triggers": event_triggers,
+            "has_event_logs": any(item["event_logs"] for item in event_triggers),
             "participant_schedules": session.participant.get_schedules_for_experiment(
                 experiment.id, as_dict=True, include_inactive=True
             ),
             "participant_id": session.participant_id,
+            "participant": participant,
+            "participant_data": json.dumps(participant_data_row.data if participant_data_row else {}, indent=4),
```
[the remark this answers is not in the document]
preston-56: Checked the codebase for precedent: `@never_cache` (which sets `Cache-Control: no-store` among other directives) is used in two places: Django's own admin-framework default (`apps/web/admin.py`) and a Celery task-status polling endpoint (`apps/web/views.py`) that needs it for freshness, not privacy. Neither is about protecting sensitive data from caching, so there's no existing pattern here for that specific concern.

Given that, I'd rather not bolt a one-off `no-store` onto just these two views. If this is worth doing, it reads more like an app-wide caching policy (probably middleware covering authenticated/team-scoped views generally) than a per-view decorator, and deserves its own deliberate decision rather than being decided inside a page-redesign PR. Filing it separately.

templates/chatbots/chatbot_session_view.html
```diff
@@ -112,64 +172,195 @@ <h3 class="font-bold text-lg">{% translate "End Session?" %}</h3>
...
+          {% endflag %}
+        </dl>
+        {% if perms.annotations.add_usercomment %}
+          <div x-data="{showComments: false}" class="mt-4">
+            <dt class="text-xs uppercase tracking-wide text-base-content/60">
+              {% translate "Comments" %} (<span id="{{ experiment_session.chat.comment_count_element_id }}">{{ experiment_session.chat.get_user_comments|length }}</span>)
+            </dt>
+            <dd class="text-sm mt-1">
+              <button class="btn btn-sm btn-ghost pl-0" @click="showComments = !showComments;">
+                <i class="fa-regular fa-comment"></i>
```
SmittieC: I noticed that the "add to dataset" and "traces" icons are gone. I do think we need to bring those back

image

templates/chatbots/chatbot_session_view.html
SmittieC: I know it's not part of this issue's requested change, but can we rename the button to "next session" and "previous session"? The reason being that in the current UI running in prod, it's clearer that these buttons do not relate to the chat messages, but rather to the session, but since the session details table is gone now, these appear closer to the messages, which confused me a bit. Thougths?

templates/web/app/app_base.html
```diff
@@ -13,7 +13,7 @@
   {% block notifications %}
   {% endblock notifications %}
   <section class="m-2 lg:m-4">
-    <div class="xl:container flex flex-row">
+    <div class="xl:container 2xl:max-w-none flex flex-row">
```
SmittieC: why this change?

apps/chatbots/views.py
```diff
@@ -673,10 +674,21 @@ def chatbot_session_details_view(request, team_slug: str, experiment_id: uuid.UU
         session_id,
         active_tab="chatbots",
         template_path="chatbots/chatbot_session_view.html",
-        session_type="Chatbot",
     )
 
 
+@login_and_team_required
+@permission_required("experiments.download_chats", raise_exception=True)
+def export_chatbot_session_messages(request, team_slug: str, experiment_id: uuid.UUID, session_id: str):
```
SmittieC, line 682: I'm not sure about single session exports. I know the mockup showed a button, but I don't think it's useful. I might be wrong. Looping in @snopoke

templates/chatbots/chatbot_session_view.html
```diff
@@ -1,5 +1,39 @@
 {% extends "web/app/app_base.html" %}
-{% load form_tags i18n json_tags static waffle_tags %}
+{% load form_tags humanize i18n json_tags static waffle_tags %}
+
+{% block page_head %}
+  {{ block.super }}
+  <style>
+    /* The Format/Revert/Save row above the editor already covers this; the
+       JS-injected floating Format button inside the editor would duplicate it. */
+    #participant-data-panel .json-editor > button {
```
SmittieC: why not add these in assets/styles/site-tailwind.css?

SmittieC (commented):
This looks great! Some inital comments

templates/web/app/app_base.html
snopoke: +1 to leaving this out - I think we should take on the gutter removal as a separate issue since it impacts all the screens.

apps/generics/views.py
```diff
@@ -32,6 +34,15 @@ def render_session_details(
     # Usage/cost is team-internal: participants viewing their own session (no team
     # membership) don't see it.
     show_usage_summary = bool(request.team_membership)
+    session_start = session.consent_date or session.created_at
```
snopoke: Always use `created_at` (also `consent_date` is going to be retired)

apps/generics/views.py
```diff
@@ -42,29 +53,28 @@ def render_session_details(
             "annotation_queue_names": annotation_queue_names,
             "show_usage_summary": show_usage_summary,
             "usage_summary": session_usage(session) if show_usage_summary else None,
+            "participant_chip": session.get_participant_chip(
+                include_link=request.user.has_perm("experiments.view_participant")
+            ),
+            "session_start": session_start,
+            "session_duration_display": seconds_to_human(
+                max(0, ((session.ended_at or timezone.now()) - session_start).total_seconds()), compact=True
```
snopoke: The end date should either be `session.ended_at` or the date of the last message to avoid an ever growing duration for a session with no activity. You can actually use `session.latest_trace` to get the end date.

templates/chatbots/chatbot_session_view.html
```diff
@@ -182,9 +373,11 @@ <h3 class="text-base font-semibold leading-7 mt-8 pl-4">Session details</h3>
     data-unlinkTagUrl="{% url 'annotations:unlink_tag' request.team.slug %}"
   >
   </script>
+  <script src="{% static 'js/editors-bundle.js' %}"></script>
   <script>
     window.addEventListener('load', () => {
       SiteJS.tagMultiselect.setupTagSelects();
+      SiteJS.editors.initJsonEditors();
```
snopoke: I know the spec had the participant data view as an editable element but I don't actually think that makes sense because the current display shows the participant data at the time of the last message.

Let's drop the editor and revert to showing `experiment_session.latest_participant_data`

templates/experiments/chat/components/trace_icons.html
```diff
@@ -1,21 +1,19 @@
...
-  {% for trace_info in trace_infos %}
-    <div class="tooltip tooltip-right" data-tip="Open Trace on {{ trace_info.trace_provider|title }}">
-      {% if trace_info.trace_provider == "ocs" %}
-        <a class="btn btn-sm btn-square btn-ghost" href="{{ trace_info.trace_url }}" target="_blank">
-          <img class="rounded-md max-w-4 max-h-4" src="{% static 'images/favicons/favicon.svg' %}" alt="OCS Trace">
+  {% if trace_infos %}
+    {% for trace_info in trace_infos %}
+      <div class="tooltip tooltip-right" data-tip="Open Trace on {{ trace_info.trace_provider|title }}">
+        <a class="btn btn-xs btn-square btn-ghost" href="{{ trace_info.trace_url }}" target="_blank">
+          <i class="fa-solid fa-diagram-project"></i>
```
snopoke: This should use the langfuse icon for langfuse traces

templates/experiments/components/session_messages.html
```diff
@@ -1,5 +1,6 @@
-{% load chat_tags form_tags json_tags pagination_tags waffle_tags %}
-<div id="table-top" class="max-w-5xl mt-4 border rounded-2xl" x-data="{selectedDataset: null}">
+{% load chat_tags default_tags form_tags json_tags pagination_tags waffle_tags %}
```
snopoke: you don't need to load `default_tags` since they are always available

templates/participants/partials/participant_data.html
snopoke, line 1: These changes seem unrelated to the session detail page changes and should be reverted

snopoke (commented):
Thanks for these changes. My main feedback is to drop the JsonEditor and related code for participant data on this page and revert to using `session.latest_participant_data`. This also means you can undo the changes to the participant data edit view and template.

apps/chatbots/views.py
snopoke: I'm OK keeping this in

templates/chatbots/chatbot_session_view.html
preston-56: Renamed to "Previous session" and "Next session".

templates/web/app/app_base.html
preston-56: Reverted this, agree it's better as its own issue since it touches every screen.

templates/chatbots/chatbot_session_view.html
preston-56: Removed that whole style block instead, it was only scoped to the CodeMirror classes for the editable participant-data view, which I'm dropping per snopoke's comment below.

apps/generics/views.py
preston-56: Fixed, using created_at now.

apps/generics/views.py
preston-56: Fixed, now falling back to session.latest_trace.timestamp for the end date so the duration doesn't keep growing on reload for a session with no ended_at.

templates/chatbots/chatbot_session_view.html
preston-56: Agreed, dropped the editable view. Now showing experiment_session.latest_participant_data read-only, which already reflects the state as of the last message.

templates/experiments/chat/components/trace_icons.html
preston-56: Fixed, shows the langfuse icon for langfuse traces now and the generic icon otherwise.

templates/experiments/components/session_messages.html
preston-56: Removed, thanks for catching that.

templates/participants/partials/participant_data.html
preston-56: Reverted. That was leftover scaffolding for the editable participant-data view, which I'm dropping per your other comment.

assets/javascript/editors.js
```diff
@@ -213,6 +213,7 @@ class JsonEditor extends BaseEditor {
     this.element.addEventListener('resetEditor', () => {
       setTimeout(() => this.reset(), 10);
     })
+    this.element.addEventListener('formatEditor', () => this.formatJSON());
```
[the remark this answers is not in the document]
preston-56: This is already handled, JsonEditor listeners go through addTrackedListener now and destroy() removes them.

templates/experiments/components/session_messages.html
```diff
@@ -10,197 +11,174 @@
...
-          {% include "experiments/components/pagination_buttons.html" %}
+          </div>
+          <div class="pg-help flex items-center gap-3 flex-wrap">
+            <span>
+              <i class="fa-regular fa-clock mr-1"></i>
+              <time datetime="{{ message.created_at.isoformat }}" title="{{ message.created_at.isoformat }}">
+                {{ message.created_at|date:"DATETIME_FORMAT" }}
+              </time>
+            </span>
+            <div class="flex items-center gap-0 opacity-0 group-hover:opacity-100 focus-within:opacity-100 transition-opacity">
```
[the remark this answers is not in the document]
preston-56: This is already handled, the action group has [@media(hover:none)]:opacity-100 alongside the hover/focus classes so it stays visible on touch devices.

snopoke (approved):

apps/generics/views.py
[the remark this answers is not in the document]
preston-56: Follow-up: went with the middleware approach mentioned above. #4380 has the details, fix is in progress on a branch now.

PR #4381, opened 2026-09-03
Feat/scheduled trigger - #3343 

### LINKED ISSUE #3343: Events: add absolute-time schedule trigger
## Summary

The events framework currently supports two trigger types:

- **`StaticTrigger`** — fires on a discrete event (conversation start/end, new message, participant joined, etc.)
- **`TimeoutTrigger`** — fires after a relative delay from some anchor point (last message, first message, inactivity)

Neither supports firing at a **specific absolute datetime**. This is a gap for use-cases like:

- "Send a follow-up message to all participants on 2025-06-01 at 09:00"
- "End conversations that haven't completed by a deadline"
- Study or programme cohorts where events must fire at fixed calendar dates regardless of when each session started

## Proposed new trigger type: `ScheduledTrigger`

A new model alongside `StaticTrigger` and `TimeoutTrigger` that fires at an explicitly specified datetime.

### Model fields (sketch)

| Field | Type | Notes |
|---|---|---|
| `experiment` | FK → Experiment | |
| `action` | OneToOne → EventAction | same as existing triggers |
| `scheduled_at` | DateTimeField | the absolute UTC datetime to fire |
| `timezone` | CharField | optional display timezone for the UI |
| `is_active` | BooleanField | |
| `is_archived` | BooleanField | |
| `working_version` | self FK | versioning, consistent with other triggers |

### Scheduling / execution

- A periodic Celery beat task (or an extension of the existing `trigger_bot_message_task` approach) polls for `ScheduledTrigger` instances whose `scheduled_at <= now` and `is_active=True` and haven't already fired.
- Alternatively, use `django-celery-beat` to create a one-shot `PeriodicTask` when the trigger is saved, and delete it after firing.
- Firing behaviour is identical to existing triggers: instantiate the `ACTION_HANDLERS` handler for the action type and call `invoke()` for each eligible session.

### Scope questions to resolve

1. **Per-session vs. per-experiment**: does a `ScheduledTrigger` fire once globally (e.g. for all active sessions of an experiment) or is it pinned to a single session? The existing `TimeoutTrigger` is per-experiment-but-evaluated-per-session, which is probably the right model here too.
2. **Repeat / recurrence**: out of scope for the first version — one-shot only.
3. **UI**: a new row in the events table with a datetime picker instead of a delay input.
4. **Versioning**: should follow the same `VersionsMixin` pattern as `TimeoutTrigger`.

## Acceptance criteria

- [ ] `ScheduledTrigger` model created with migration
- [ ] Admin registration
- [ ] Celery task that fires due triggers and logs results via `EventLog`
- [ ] Create / edit / delete UI in the events section
- [ ] The trigger is included in experiment versioning / copy
- [ ] Tests covering: trigger fires at the right time, does not double-fire, respects `is_active`

yusra-muhtar:
### Product Description
Adds a new third event trigger type — Scheduled Trigger — for Chatbots. When one fires at its configured date, time and timezone, it runs the same actions as existing event triggers (log, end conversation, send a message to the bot, schedule another trigger, or start a pipeline). Triggerable actions keep working in the event logs.
Closes #3343 

### Technical Description
Implements ScheduledTrigger in apps/events/models.py following the existing StaticTrigger/TimeoutTrigger pattern (BaseModel, VersionsMixin, OneToOneField to EventAction, experiment-scoped, working_version self-FK, is_archived/is_active).

Key design decisions:

- Due-time querying: a UTC scheduled_at DateTimeField is computed at save from the configurer's trigger_date/trigger_time/timezone (via a new scheduled_datetime_for helper) so the poll task can query with a simple scheduled_at__lte=now() index lookup. The raw date/time/timezone are stored separately for display and DST handling. DST ambiguity is resolved with pytz.localize(is_dst=False), retrying with is_dst=True on NonExistentTimeError.
- Single-fire guarantee: fire() claims the trigger atomically (filter(id=..., fired_at__isnull=True, is_active=True).update(fired_at=now())) to prevent double-firing, then logs success/failure to EventLog via the working version.
- Session resolution: uses the latest experiment session (experiment.sessions.order_by("-created_at").first()), not participant-scoped, per confirmed decision. EventLog.session was made nullable so a trigger scheduled with no session yet can still log a failure gracefully.
- Celery: poll_due_scheduled_triggers (60s cadence, matching poll_scheduled_messages) enqueues fire_scheduled_trigger for due+active+published triggers, excluding migrating teams; both tasks run on Queues.BACKGROUND.
- Full CRUD surface: form (ScheduledTriggerForm) with date/time HTML widgets and a pytz.all_timezones timezone choice + past-time validation; create/edit/delete/toggle-active/logs views, URLs, EventsTable rendering, the events-tab "Create Scheduled Event" button, and the shared event_logs.html display branch.
- Versioning/copy: sync_triggers copies the EventAction and resets fired_at; the experiment version details/TRIGGER_ACCESSORS include scheduled_triggers.
- Permissions/usages: scheduledtrigger added to CONTENT_TYPES["events"] in apps/teams/backends.py; pipeline_start scheduled triggers participate in pipeline-usage searches.

### Migrations

0028_alter_eventlog_session_scheduledtrigger creates the new ScheduledTrigger table and makes EventLog.session nullable. The session nullability is a widening change (existing rows unaffected) and the new table is additive, so old code can run safely while the migration is applied.

### Demo
Screenshot 2026-09-14 at 3 55 14 PM
Screenshot 2026-09-14 at 3 50 46 PM
Screenshot 2026-09-14 at 3 55 45 PM

### Docs and Changelog

### Operator Impact


kpoudosu:
@yusra-muhtar 
Good work overall. The circular-import fix (moving EventLog to its own module) is a clean call, the DST handling in scheduled_datetime_for is careful and well-tested, and the atomic claim in fire() is the right approach to prevent double-firing. Adding @require_POST to the delete/toggle views is a good catch and also fixes the pre-existing gap on static and timeout views.

A few things to look at before marking ready:

Verify: save() does not set scheduled_at for new objects. The guard if self.pk and self.scheduled_at != new_scheduled_at: means scheduled_at is only updated when an existing trigger changes schedule. For a first-time save (pk is None), self.scheduled_at is never assigned. Either the view or the form must be setting it explicitly — confirm this is handled and add a comment if so, or move self.scheduled_at = new_scheduled_at outside the if block to make the first-save path explicit:

def save(self, *args, **kwargs):
    new_scheduled_at = self.scheduled_datetime
    if self.pk and self.scheduled_at != new_scheduled_at:
        self.fired_at = None
    self.scheduled_at = new_scheduled_at  # always set, not just on schedule change
    super().save(*args, **kwargs)

fire_scheduled_trigger task is missing select_related. Inside fire() it accesses self.action and self.experiment. Add .select_related("action", "experiment") to the task's objects.get() call to avoid two extra DB round-trips per trigger firing.

_validate_team_experiment return value is discarded. The function fetches the experiment but _create_event_view doesn't use the returned object. Either use it downstream or rename it to _assert_team_experiment to make clear it is a guard, not a getter.

No index on scheduled_at. Fine for the pilot scale, but the poll task does scheduled_at__lte=now() on every 60s cycle. Worth a follow-up ticket to add a partial index on (scheduled_at, is_active, fired_at) before we go beyond the pilot.

Otherwise the implementation is solid. Mark ready once the save() question is resolved.

apps/events/models/models.py
```diff
@@ -77,30 +78,6 @@ def save(self, *args, **kwargs):
...
-    )
-    status = models.CharField(choices=EventLogStatusChoices.choices)
-    log = models.TextField(blank=True)
-
-    class Meta:
-        indexes = [
-            models.Index(fields=["content_type", "object_id"]),
-        ]
-
-
```
snopoke, line 102: If you're going to split the models file you should move the modules to a new `models` package:

```
/models
|- __init__.py  #import the other models here
|- scheduled_trigger.py
|- event_log.py
|- models.py

File: apps/events/scheduled_trigger.py [hunk 2]
```diff
@@ -0,0 +1,153 @@
...
+
+    def fire(self):
+        working_version = self.get_working_version()
+        claimed = ScheduledTrigger.objects.filter(id=self.id, fired_at__isnull=True, is_active=True).update(
+            fired_at=timezone.now()
+        )
+        if not claimed:
+            return None
+
+        session = self._resolve_session()
```
snopoke: This only fire's for the first session. It should send a message to very 'active. session. The definition of 'active' might need to include a user configurable window e.g. 'open sessions with messages in the last X days' where X is a config parameter. 

Also open to other suggestions for what constitutes 'active' sessions.

apps/events/models/models.py
yusra-muhtar: Done, the events models now live in a models/ package.

apps/events/scheduled_trigger.py
yusra-muhtar: Done, fire() now targets all active sessions instead of just the newest one. _resolve_sessions() filters to ACTIVE sessions and there's a new user-configurable active_session_window_days field: when set, it only fires against sessions with activity within the last N days (falling back to created_at for sessions with no participant message, matching the broadcast active_within_days behaviour).
Screenshot 2026-09-04 at 2 46 51 PM

snopoke:
@yusra-muhtar please drop e848fe0f9eaac75b87fbf15cd8e02fc71bb2d69a from this branch as it isn't related to the other changes in the PR.

yusra-muhtar:
@snopoke I reverted those Session API work, they were meant for the other branch.

apps/events/models/scheduled_trigger.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def scheduled_datetime_for(trigger_date, trigger_time, timezone_name) -> datetime:
+    """Convert a naive local date/time in ``timezone_name`` into an aware UTC datetime.
+
+    Ambiguous or non-existent local times (DST transitions) are resolved deterministically:
+    an ambiguous time maps to the standard-time occurrence, and a non-existent time is
+    advanced through the spring-forward gap to preserve the intended wall-clock schedule.
+    """
+    tz = pytz.timezone(timezone_name)
```
snopoke: nit: prefer `ZoneInfo` over `pytz`

apps/events/tests/test_views.py
```diff
@@ -31,3 +82,92 @@ def test_action_params_form_view_400_for_invalid_action_type(experiment, authed_
...
+
+    toggle_url = reverse(
+        "chatbots:events:scheduled_event_toggle",
+        args=[team_with_users.slug, other_experiment.id, other_trigger.id],
+    )
+    assert client.post(toggle_url).status_code == 404
+
+
+@pytest.mark.django_db()
+def test_scheduled_delete_and_toggle_reject_get_requests(client, team_with_users):
```
snopoke: drop this test

apps/events/tests/test_views.py
```diff
@@ -31,3 +82,92 @@ def test_action_params_form_view_400_for_invalid_action_type(experiment, authed_
     )
     response = authed_client.get(url, {"action_type": "bogus"})
     assert response.status_code == 400
+
+
+@pytest.mark.django_db()
+def test_scheduled_views_are_scoped_to_request_team(client, team_with_users):
```
snopoke: drop this test

apps/events/tests/test_views.py
```diff
@@ -31,3 +82,92 @@ def test_action_params_form_view_400_for_invalid_action_type(experiment, authed_
...
+
+    toggle_url = reverse(
+        "chatbots:events:scheduled_event_toggle",
+        args=[team_with_users.slug, team_experiment.id, trigger.id],
+    )
+    assert client.get(toggle_url).status_code == 405
+
+
+@pytest.mark.django_db()
+def test_toggle_rejects_trigger_belonging_to_different_experiment_in_same_team(client, team_with_users):
```
snopoke: drop this test

apps/events/forms.py
```diff
@@ -158,3 +168,54 @@ class Meta:
             "delay": "Wait time",
             "trigger_from_first_message": "Trigger from first message",
         }
+
+
+class ScheduledTriggerForm(BaseTriggerForm):
+    timezone = forms.ChoiceField(
+        choices=[(tz, tz) for tz in pytz.all_timezones],
```
snopoke: Use `zoneinfo.available_timezones()`

apps/events/forms.py
```diff
@@ -158,3 +168,54 @@ class Meta:
...
+        label="Timezone",
+        initial="UTC",
```
snopoke: ideally this should default to the user's current timezone (`timezone.get_current_timezone()`)

apps/events/tasks.py
```diff
@@ -68,6 +75,32 @@ def fire_trigger(trigger_id, session_id):
     return triggered
 
 
+@shared_task(ignore_result=True, queue=Queues.BACKGROUND)
+def poll_due_scheduled_triggers():
+    # TODO: add a partial index on (scheduled_at, is_active, fired_at) before scaling beyond pilot.
```
snopoke: drop this TODO

apps/events/views.py
```diff
@@ -207,23 +258,33 @@ def _get_event_logs_context(trigger):
 
 @login_and_team_required
 @permission_required("events.change_statictrigger")
+@require_POST
```
snopoke, line 261: Thanks for these and the team scope checks

templates/events/events_actions_column_buttons.html
```diff
@@ -1,10 +1,13 @@
 {% include "events/components/toggle_active_flag_modal.html" with event=event toggle_active_flag_url=toggle_active_flag_url %}
 <a class="btn btn-sm btn-outline btn-primary" href="{{ view_log_url }}" aria-label="View logs">View Logs</a>
 <a class="btn btn-sm btn-outline btn-primary" href="{{ edit_url }}" aria-label="Edit event">Edit</a>
-<a class="btn btn-sm btn-outline btn-primary" href="{{ delete_url }}" aria-label="Delete event">Delete</a>
+<form method="post" action="{{ delete_url }}" style="display: inline;">
+  {% csrf_token %}
+  <button type="submit" class="btn btn-sm btn-outline btn-primary" aria-label="Delete event">Delete</button>
+</form>
```
snopoke, line 7: 🥇

snopoke (commented):
a few small comments otherwise LGTM

apps/events/models/scheduled_trigger.py
yusra-muhtar: Done

apps/events/tests/test_views.py
yusra-muhtar: Done, Test has been dropped

apps/events/tests/test_views.py
yusra-muhtar: Test has been dropped

apps/events/tests/test_views.py
yusra-muhtar: Done

apps/events/forms.py
yusra-muhtar: Done

apps/events/forms.py
yusra-muhtar: Done

apps/events/tasks.py
yusra-muhtar: Dropped

apps/events/forms.py
```diff
@@ -172,9 +173,9 @@ class Meta:
 
 class ScheduledTriggerForm(BaseTriggerForm):
     timezone = forms.ChoiceField(
-        choices=[(tz, tz) for tz in pytz.all_timezones],
+        choices=[(tz, tz) for tz in sorted(available_timezones())],
         label="Timezone",
-        initial="UTC",
+        initial=getattr(timezone.get_current_timezone(), "key", "UTC"),
```
snopoke: I don't think you can do this at the field level since this only happens once when the class is defined. It needs to be set when the form is initialized (in the context of a request)

apps/events/migrations/0029_scheduledtrigger_active_session_window_days.py
snopoke, line 1: @yusra-muhtar please merge this with the previous migration

snopoke (commented):
@yusra-muhtar please can you add a short demo / screenshots showing the UI components of this

apps/events/migrations/0029_scheduledtrigger_active_session_window_days.py
yusra-muhtar: Screenshots have been added and the migration files have been merged.

components/chat_widget/src/index.html
snopoke, line 1: Please revert this change as it is unrelated to the PR. This isn't even a Django template

templates/chatbots/single_chatbot_home.html
```diff
@@ -153,7 +153,9 @@ <h2 id="chatbot-description" class="flex-1 pg-subtitle">{{ experiment.descriptio
         <input type="radio" name="tab_group" role="tab" class="tab" aria-label="Sessions" id="tab-allsessions" {% if not highlight_version_id %}checked{% endif %}/>
       {% endif %}
       <input type="radio" name="tab_group" role="tab" class="tab" aria-label="Versions" id="tab-versions" {% if highlight_version_id %}checked{% endif %}/>
-      <input type="radio" name="tab_group" role="tab" class="tab" aria-label="Events" id="tab-events" />
+      {% flag "flag_events" %}
```
snopoke: This flag has been removed

components/chat_widget/src/index.html
yusra-muhtar: Done

templates/chatbots/single_chatbot_home.html
yusra-muhtar: Done, flag_events gate removed to match main.

snopoke (approved):

PR #4389, opened 2026-09-03
Retire the pre-login marketing pages

snopoke:
### Product Description

The marketing pages (`/about/`, `/applications/`, `/contact/`, `/open-opportunities/`)
now live at <https://openchatstudio.dimagi.com>, in
`dimagi-internal/open-chat-studio-prelogin`. This is the teardown half: `/` serves a
small landing page for anonymous visitors instead of the full marketing home, and the
retired paths `301` to the new site. Signed-in users are unaffected — `home()` keeps its
dashboard/create-team branch untouched.

### Technical Description

Phase 5 of the migration plan in the prelogin repo (`docs/MIGRATION_PLAN.md`).

**`apps/prelogin` stays**, despite serving almost nothing now, because `prelogin:home` is
reversed from six places outside it: `apps/sso/views.py`,
`apps/teams/views/{invitation,membership,manage_team}_views.py`,
`templates/account/logout.html` and `apps/teams/tests/test_teams_middleware.py`. The plan
had listed three; the other three turned up on inspection.

Worth a reviewer's attention:

- **`PROJECT_METADATA["MARKETING_SITE_URL"]`** is the single source for the new host —
  `urls.py` builds the redirects from it and the frame links to it. It is hardcoded to
  Dimagi's host, which means a **self-hosted** deployment's `/about/` now redirects to our
  marketing site. `PROJECT_METADATA` is the documented place for a fork to override, but
  say the word if this should be env-driven instead.
- **`/platform/` redirects straight to `…/#how-it-works`**, not to `…/platform/`, because
  that path is itself a 301 on the marketing site and chaining costs a hop.
- **The landing page canonicalises to the marketing home.** Without it this host serves an
  uncanonicalised near-duplicate and the two compete.
- **`StaticViewSitemap` drops to `prelogin:home`.** This is the urgent bit: until it
  merges, `www.openchatstudio.com/sitemap.xml` is still advertising the four retired URLs
  to Google while the new host has no history, so the old host stays the preferred
  canonical — the inverse of the point of the move.

Not carried over: `PreloginTemplateView` (no users left — every route is `home` or a
redirect) and the `prelogin/{home,about,applications,contact,open_opportunities}.html`
templates. `auth_base.html` is untouched, and the frame keeps the two blocks it overrides.

### Migrations

None.

### Demo

Screenshot from 2026-09-03 16-11-09

### Operator Impact

Four settings are removed, and the app no longer serves marketing pages. Logged under
`[Unreleased]` in `CHANGELOG.md` — Configuration for the settings, Deployment for the
redirect target, which a fork will want to override.

---

**Sequencing.** `ocs-deploy` still names the removed settings
(`ocs_deploy/config.py:102` and `304-307`, plus its `.env.example`) — that has to land
*after* this ships, so a deploy never reads a parameter that vanished mid-flight. Their
SSM parameters go last.

**Pre-existing test failures on `main`**, unrelated to this branch:
`test_schema_is_up_to_date_and_valid[export]` and
`test_repeated_metric_params_are_deduplicated`. I confirmed both fail on an untouched
`main` checkout. Separately,
`test_delete_view_blocks_when_only_versioned_pipeline_references_action` is flaky in a
full-suite run — `ExperimentFactory.name` is a session-global `factory.Sequence` and the
test substring-matches `f"{name} (v1)"`, so `"Test Experiment 5 (v1)"` matches inside
`"Test Experiment 51 (v1)"`. Worth its own fix.


config/settings.py
```diff
@@ -710,6 +702,12 @@ def _celery_task_queues() -> tuple[Queue, ...]:
     "PRIVACY_POLICY_URL": env("PRIVACY_POLICY_URL", default=""),
     "ACCEPTABLE_USE_POLICY_URL": env("ACCEPTABLE_USE_POLICY_URL", default=""),
     "DOCS_URL": env("DOCS_URL", default="https://docs.openchatstudio.com"),
+    # The marketing site. Separate host, separate repo
+    # (dimagi-internal/open-chat-studio-prelogin, an Astro site on Cloudflare
+    # Workers). Lives here rather than being hardcoded because both the pre-login
+    # frame and the redirects in apps/prelogin/urls.py need it, and a second copy
+    # is a link that rots the next time it moves.
```
snopoke:
```suggestion
```

config/settings.py
```diff
@@ -710,6 +702,12 @@ def _celery_task_queues() -> tuple[Queue, ...]:
...
+    "MARKETING_SITE_URL": "https://openchatstudio.dimagi.com",
```
SmittieC, line 710: Confirming if this should be hardcoded?

templates/prelogin/base.html
```diff
@@ -25,7 +32,7 @@
     <div class="nav-wrap">
       <div class="container">
         <nav class="primary">
-          <a class="logo" href="{% url 'prelogin:home' %}">
+          <a class="logo" href="{{ project_meta.MARKETING_SITE_URL }}/">
             <img src="{% static 'prelogin/images/ocs-logo.png' %}"
```
SmittieC, line 36: How far should we go with advertising Dimagi in all of this? If I self-host this, I probably want to be able to load this icon from some other source with my own branding.

SmittieC (approved):
LGTM. Two questions

templates/prelogin/base.html
snopoke: I'm not sure. Supporting full 'white labelling' would be a large feature. The main goal of this PR was to remove the Dimagi marketing site from the OCS repo.

PR #4392, opened 2026-09-03
Add reranking to collection retrieval

VicvekSr9485:
Phase 3 of the RAG extension. After dense and lexical retrieval have produced their candidates, a reranker rescores them against the query and the best of them are returned.

Adds a Reranker interface with a Voyage AI implementation, built from a collection's provider credentials through LlmService.get_reranker, and a rerank stage in search_collection gated by the flag_reranking waffle flag. The stage is independent of hybrid search: it reorders a dense-only ranking as readily as a fused one, and falls back to the ranking it was given whenever the reranker fails or is misconfigured.

Collection gains enable_reranking, reranker_provider, rerank_model and rerank_top_n, with a check constraint on rerank_top_n. search_collection gains the context parameter from issue #2681; the chat search tools pass recent conversation turns so the reranker can tell which question is being asked.

### Migrations


apps/chat/agent/tools.py
```diff
@@ -150,8 +150,42 @@ def _get_search_tool_footer(with_citations: bool):
...
+    Returns None unless reranking is actually active for the collection. The rerank stage is the
+    only consumer and it is off by default, so the query this costs must not be paid by every
+    search. `reranking_enabled` is a cached waffle lookup, which is the cheaper of the two.
+    """
+    if session is None or not collection.reranking_enabled:
+        return None
+
+    # System messages are excluded: they are prompts and summaries, not turns, and would crowd
+    # out the exchange the context exists to carry.
+    recent = list(
```
snopoke: We shouldn't need to reload this data since it is already loaded for the llm.

The correct approach here is to have langgraph inject the graph state into the tool call using the [InjectedState](https://reference.langchain.com/python/langgraph.prebuilt/tool_node/InjectedState) schema annotation. We already do this in [other places](https://github.com/dimagi/open-chat-studio/blob/main/apps/chat/agent/schemas.py#L70)

apps/documents/rerankers.py
```diff
@@ -0,0 +1,84 @@
...
+        self._timeout = timeout
+
+    def rerank(self, query: str, documents: Sequence[str], *, limit: int) -> list[RerankedDocument]:
+        documents = list(documents)
+        if not documents or limit <= 0:
+            # Voyage rejects an empty document list, and there is nothing for the caller to
+            # reorder either way, so this is answered without a billed round trip.
+            return []
+
+        import voyageai  # noqa: PLC0415 - TID253: heavy lib, slow startup
```
snopoke: Is this really a heavy lib? Maybe ask for some evidence of that.

apps/documents/rerankers.py
```diff
@@ -0,0 +1,84 @@
...
+        response = client.rerank(
+            query=query,
+            documents=documents,
+            model=self._model,
+            top_k=min(limit, len(documents)),
+            # Clip the pair to the model's context window rather than have the call rejected
+            # outright. Documents are chunk-sized and the query is bounded by its caller (see
+            # `apps.documents.retrieval`), so this is a guard, not the normal path.
+            truncation=True,
+        )
```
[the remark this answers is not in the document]
snopoke: `voyageai` is already a transitive dependency so OK to add it with the same version pin.

apps/documents/retrieval.py
```diff
@@ -75,6 +114,117 @@ def search_collection(
...
+        return candidates[:top_k]
+
+    logger.info(
+        "Reranked collection retrieval candidates",
+        extra={
+            "candidate_count": len(candidates),
+            "result_count": len(ranked),
+            "duration_ms": round((time.monotonic() - started) * 1000),
+        },
+    )
```
snopoke: Drop this - only operators will see it and it doesn't mean much out of context. If we want timing info on this we should using tracing which would mean pushing the tracing service down to tools. Let's leave that as out of scope.

apps/documents/retrieval.py
```diff
@@ -75,6 +114,117 @@ def search_collection(
...
+        return candidates[:top_k]
+
+    started = time.monotonic()
+    try:
+        ranked = reranker.rerank(
+            _rerank_query(query, context),
+            [candidate.contextualized_text for candidate in candidates],
+            limit=top_k,
+        )
+    except Exception:
```
snopoke: What kind of errors could we expect here?

apps/teams/flags.py
```diff
@@ -90,6 +90,14 @@ class Flags(FlagInfo, Enum):
         True,
     )
 
+    RERANKING = (
```
snopoke: Let's merge this with the hybrid search flag rather than introduce a new one

apps/documents/models.py
```diff
@@ -267,6 +268,45 @@ class Collection(BaseTeamModel, VersionsMixin):
         validators=[MinValueValidator(1)],
         help_text=("How many candidates to retrieve from each of the dense and lexical searches before fusing them."),
     )
+    # Reranking. Literal defaults for the same reason as the hybrid search knobs above, and off
+    # the pipeline node UI for the same reason too.
+    enable_reranking = models.BooleanField(
```
snopoke: Do we need a separate field for this? Could we not infer this from the presence of a provider and model?

VicvekSr9485: Thanks for the review! I’ve addressed the comments: reused the hybrid-search flag, inferred reranking from provider/model configuration, switched conversation context to LangGraph InjectedState, removed the timing log, pinned voyageai, and regenerated the export schema.

apps/documents/retrieval.py
```diff
@@ -75,6 +113,108 @@ def search_collection(
...
+    Every failure path returns the candidates in the order retrieval already put them in, cut to
+    `top_k`. Reranking is a quality stage on top of a ranking that is already usable, so a
+    reranker that errors, times out, or answers with something that does not describe the list it
+    was sent must cost the search its improvement, not its results.
+
+    Chunks that were reranked carry a `rerank_score`; on a fallback path they do not, which is
+    how the collection query preview knows which number it is looking at.
+    """
+    if top_k <= 0:
+        # A caller that wants nothing back gets it without a billed call.
```
snopoke:
```suggestion
```

apps/documents/retrieval.py
```diff
@@ -75,6 +113,108 @@ def search_collection(
...
+        return []
+    if len(candidates) <= 1:
+        # A single candidate cannot be reordered, and the provider would still bill the call.
```
snopoke:
```suggestion
```

apps/documents/retrieval.py
```diff
@@ -75,6 +113,108 @@ def search_collection(
...
+        return candidates[:top_k]
+
+    query = query.strip()
+    if not query:
+        # Nothing to score the candidates against. Dense search still ranks a blank query by
+        # whatever its embedding came out as, but a reranker has no such fallback.
```
snopoke:
```suggestion
```

snopoke (approved):

PR #4393, opened 2026-09-03
Redesign participant details page

### LINKED ISSUE #4231: Redesign participant details page
Primary goals:

1. Improve usability of page
2. Improve layout and presentation of details
3. Improve page performance

### Design guide:

This is intended as guidance for the design and not necessarily the finalized design.

[Participant Details (standalone).html](https://github.com/user-attachments/files/31788898/Participant.Details.standalone.1.html)

[Claude design](https://claude.ai/design/p/263087cd-89f0-425e-96da-3e50ace56462?file=Participant+Details.dc.html&via=share)

Image

preston-56:
### Product Description

Redesigns the participant details page to fix three long-standing usability problems:

1. The page required picking one chatbot before showing anything. It now aggregates sessions, data, and schedules across every chatbot the participant has used, with per-chatbot filter pills inside each tab.
2. Switching chatbots or opening a filter caused a full page reload with a visible color flicker. Both are now htmx-driven, with no reload and no flicker.
3. The participant's JSON data (Format/Revert/Save) is now editable in place per chatbot, with live save-status feedback and inline validation.

Closes #4231.

### Technical Description

`apps/participants/views.py` gains three htmx panel views (`ParticipantSessionsPanel`, `ParticipantDataPanel`, `ParticipantSchedulesPanel`), one per tab. Each tab's chatbot-filter pills target only that tab's own panel rather than the whole page, which is what makes pill-switching htmx-driven instead of a real navigation. Getting there surfaced three bugs worth flagging for review:

- The Sessions panel initially re-rendered `participant_sessions_table.html` wholesale on every pill click, and that template still pulled in `experiments/filters.html`'s Alpine root (`x-data="filterComponent"`). htmx's swap inserted the whole response, so each click mounted a second live Alpine instance whose own `init()` raced the first and clobbered the filtered table. Fixed by splitting the pills, table, and htmx-only response into three templates (`participant_sessions_pills.html`, `participant_sessions_table.html`, `participant_sessions_panel.html`) so the panel endpoint never re-includes `filters.html`.
- The Participant Data tab's `hx-swap="outerHTML"` made `htmx:afterSettle`'s `evt.detail.target` a detached pre-swap node: its `.id` still read correctly, but `isConnected` was `false`. The shared `editors.js` htmx handler mounted CodeMirror onto that orphaned element, which was never in the visible DOM. Switched to `hx-swap="innerHTML"` with the `id`'d wrapper living in the parent template, not re-rendered by the panel itself, so the swap target stays the same live node across requests. The participant-data save `<form>` had the identical problem one level down: its default swap target was itself, but the response included sibling content that lived outside the `<form>` tag. Fixed the same way.
- `experiments/filters.html`'s Save/Clear form carried a hidden `chatbot` input rendered once at page load. Pill clicks deliberately never touch that file, to avoid re-mounting Alpine per the bug above, so the hidden field went stale the moment a pill was clicked and Save/Clear would silently revert the chatbot selection. Replaced it with `hx-vals` reading the live `chatbot` query param off `window.location.search` at request time.

`filters.html` gets a `simple_mode` flag rather than a fork, so the participant Sessions tab renders a lightweight Search/State/Tag panel while every other page using this shared component (chatbot home, All Sessions, broadcast, consent, public-link pages) stays unaffected. Missing `{% load i18n %}` on that file's `{% translate %}` tags broke the template parser for every page including it, not just this one. Caught by running the full suite rather than the narrowly-relevant tests, which is the case for testing shared components broadly.

`Participant.get_schedules_for_all_experiments()` and `get_message_trend()` (`apps/experiments/models.py`) aggregate across every experiment the participant has used by looping the existing per-experiment methods (`get_schedules_for_experiment`, the bucket-and-zero-fill pattern from `Experiment.get_trend_data`) instead of duplicating their filtering logic. `ParticipantSessionsTable` (`apps/chatbots/tables.py`) is a participant-scoped `ChatbotSessionsTable` subclass for the same reason: it adds a Chatbot column and swaps the actions column, and inherits the rest.

CSS: audited the diff for anything that should have been a DaisyUI/Tailwind utility instead of hand-rolled CSS. The one avoidable rule, the active tab's underline and text color, now uses the `checked:` variant already established in this codebase (`experiment_session_messages_control_panel.html`) instead of a `<style>` block. Worth noting for review: the underline is actually drawn via `border-top: 3px solid currentColor` in DaisyUI's own tabs-border implementation, not `background-color`, so an earlier attempt to override the background had no visible effect despite `getComputedStyle` reporting the right color. What remains is a small block scoped to CodeMirror's own internal classes (`.cm-editor`, `.cm-gutters`, `.cm-activeLine`). CodeMirror is a third-party editor with no DaisyUI equivalent, so that is a deliberate exception, not an oversight.

### Migrations

No migrations. Both new `Participant` methods and `initials` are computed, not stored.

### Demo

**Sessions tab.** Aggregates every chatbot's sessions in one table, with an "All chatbots" pill plus one pill per chatbot. Table text sized down (`text-xs`) so columns stop overlapping.

Sessions tab

**Identity details.** Collapsed by default, expands in place instead of cluttering the header.

Identity details expanded

**Filter panel.** The existing Filter button now opens a lightweight Search transcripts / State / Tag panel scoped to this page only. Search queries message content; Clear resets the fields and closes the panel.

Filter panel open

**Participant Data tab.** Per-chatbot JSON editor, editable in place (Format/Revert/Save), grey background with no line-number gutter, live "All changes saved" and "Saved. Valid JSON, 0 errors." feedback.

Participant Data tab

**Schedules tab.** Aggregated across all chatbots, with all three status badges (Active, Completed, Cancelled) rendering correctly.

Schedules tab

**Smooth chatbot switching.** Clicking a chatbot pill swaps just that tab's content via htmx. No full-page reload, no color flicker (previously every pill click was a real page navigation).

Smooth pill switching

**Save flow.** Switching chatbots and saving updates the JSON editor and the save-status text live, with no page reload and no duplicated content.

Save flow

### Docs and Changelog

### Operator Impact


templates/participants/partials/participant_sessions_pills.html
```diff
@@ -0,0 +1,22 @@
+{% load i18n %}
+<div id="sessions-panel" class="contents">
+  <div class="w-px h-5.5 bg-base-300 mx-1"></div>
+  <a href="?chatbot="
+     hx-get="{% url 'participants:sessions-panel' request.team.slug participant.id %}?chatbot="
+     hx-target="#sessions-panel"
+     hx-swap="outerHTML"
+     hx-push-url="?chatbot="
```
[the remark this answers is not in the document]
preston-56: Fixed. Pills now read the live `q/state/tag` off the URL via `hx-vals` and carry them forward, in both the htmx request and the plain `href/hx-push-url` fallback.

templates/participants/partials/participant_schedule_modal.html
```diff
@@ -0,0 +1,68 @@
...
+    </div>
+    <div class="modal-action">
+      {% if perms.experiments.change_participant %}
+        <button id="cancel-schedule-{{ schedule.external_id }}"
+                class="btn btn-outline btn-error"
+                {% if schedule.is_cancelled or schedule.is_complete %}disabled{% endif %}
+                hx-post="{% url 'participants:cancel_schedule' request.team.slug participant_id schedule.external_id %}"
+                hx-trigger="confirmed"
+                hx-target="#schedule_{{ schedule.external_id }}"
+                hx-swap="outerHTML"
```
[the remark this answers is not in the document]
preston-56: Fixed, the dialog now closes on a successful cancel via `hx-on::after-request`. While verifying this I found the cancel action itself was 500ing on every attempt, unrelated to this PR: `cancel_schedule` called `.prefetch_related()` on a single model instance instead of a queryset, broken on main since #4232. Fixed that too since it was blocking verification of this one.

templates/experiments/filters.html
```diff
@@ -1,253 +1,314 @@
...
+        </div>
+        <div class="w-40">
+          <label class="block text-sm font-semibold mb-1">{% translate "Tag" %}</label>
+          <input type="text"
+                 name="tag"
+                 value="{{ request.GET.tag|default:'' }}"
+                 placeholder="{% translate 'e.g. follow-up' %}"
+                 class="input input-sm input-bordered w-full">
+        </div>
+        <button type="submit" class="btn btn-sm btn-primary">{% translate "Save filters" %}</button>
```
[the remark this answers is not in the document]
preston-56: Fixed, renamed to "Apply filters".

apps/participants/views.py
```diff
@@ -212,6 +268,43 @@ def get_context_data(self, *args, **kwargs):
...
+class ParticipantDataPanel(LoginAndTeamRequiredMixin, PermissionRequiredMixin, TemplateView):
+    """Participant Data tab content, re-rendered via htmx when a chatbot pill is clicked."""
+
+    permission_required = "experiments.view_participant"
+    template_name = "participants/partials/participant_data_panel.html"
+
+    def get_context_data(self, *args, **kwargs):
+        return single_participant_home_context(self.request, {}, participant_id=self.kwargs["participant_id"])
+
+
```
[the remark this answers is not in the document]
preston-56: Agreed, real cost, `single_participant_home_context` loads all three tabs' data on every panel request. Splitting it per panel is a bigger restructure than I want to fold into this PR blind. @SmittieC @snopoke, thoughts on scope, in this PR or a follow-up issue?

apps/participants/views.py
```diff
@@ -47,31 +47,87 @@
...
 
-    context["participant"] = participant
+    filter_context = get_filter_context_data(
+        team=team,
+        columns=ExperimentSessionFilter.columns(team),
+        filter_class=ExperimentSessionFilter,
+        table_url=reverse("chatbots:participant_sessions_list", args=[team.slug, participant_id]),
+        table_container_id="participant-sessions-table",
+        table_type=FilterSet.TableType.ALL_SESSIONS,
+    )
```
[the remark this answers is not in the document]
preston-56: Fair, these are two separate filter systems right now (`date-range widget` vs. the simple `Search/State/Tag panel`) and each can drop what the other applied. Fixed the narrower case below, pills now carry `q/state/tag` forward. Unifying both into one filter implementation is bigger. @SmittieC @snopoke, should that unification happen here, or is it its own issue?

apps/participants/views.py
```diff
@@ -47,31 +47,87 @@
...
+    if state_filter == "active":
+        sessions = sessions.filter(ended_at__isnull=True)
+    elif state_filter == "ended":
+        sessions = sessions.filter(ended_at__isnull=False)
+    tag_filter = request.GET.get("tag", "").strip()
+    if tag_filter:
+        sessions = sessions.filter(chat__tags__name__icontains=tag_filter).distinct()
+    table = ParticipantSessionsTable(sessions)
+    # set request (no pagination) so the chatbot chip can permission-gate its link
+    session_table = RequestConfig(request, paginate=False).configure(table)
```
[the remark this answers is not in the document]
preston-56: Agreed, this is real: matches the approved mockup's "N of M" display, but a participant with a long session history renders every row today. Real pagination is more than a quick fix here, it needs UI, not just a queryset change. @SmittieC @snopoke, worth doing in this PR, or should I open a tracked issue for it?

apps/participants/views.py
```diff
@@ -47,31 +47,87 @@
...
-        context["session_table"] = RequestConfig(request, paginate=False).configure(table)
-        context["selected_experiment_id"] = experiment_id
-        data = participant.get_data_for_experiment(experiment_id)
-        context["participant_data"] = json.dumps(data, indent=4)
-        context["participant_schedules"] = participant.get_schedules_for_experiment(
-            experiment_id, as_dict=True, include_inactive=True
+    experiments = list(participant.get_experiments_for_display())
+
+    filter_experiment_id = request.GET.get("chatbot")
+    filter_experiment_id = int(filter_experiment_id) if filter_experiment_id else None
```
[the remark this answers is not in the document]
preston-56: Fixed. Extracted `_parse_chatbot_filter`, which catches `ValueError` and treats an unparsable value as "All chatbots" instead of raising.

apps/experiments/models.py
```diff
@@ -1260,6 +1268,50 @@ def get_schedules_for_experiment(
...
+        """Like `get_schedules_for_experiment`, but across every chatbot this participant has used.
+
+        Each returned dict carries `experiment` (the source `Experiment` instance) alongside the
+        usual `as_dict()` fields, so callers can render a chatbot column without a second lookup.
+        """
+        schedules = []
+        for experiment in self.get_experiments_for_display():
+            for schedule in self.get_schedules_for_experiment(
+                experiment.id, as_dict=True, as_timezone=as_timezone, include_inactive=include_inactive
+            ):
```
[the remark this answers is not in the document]
preston-56: Fixed. Replaced the per-experiment loop with one `experiment_id__in=...` query, `select_related('action')` and `prefetch_related('attempts')`, grouped by experiment in Python afterward. The existing aggregation test still passes.

templates/participants/single_participant_home.html
```diff
@@ -1,78 +1,227 @@
 {% extends "web/app/app_base.html" %}
-{% load form_tags i18n static %}
+{% load form_tags humanize i18n static %}
+{% block page_head %}
+  {{ block.super }}
+  <style>
```
SmittieC: Can these styles move to components/chat_widget/src/tailwind.css?

templates/participants/single_participant_home.html
```diff
@@ -1,78 +1,227 @@
...
-    <h3 class="text-base font-semibold leading-7 mt-8 pl-4">Chatbots</h3>
-    <div class="max-w-5xl mt-4 border rounded-2xl">
-      {% include "participants/partials/participant_experiments.html" %}
+  </div>
+  <div role="tablist" class="tabs tabs-border">
+    <input type="radio"
+           name="participant_tabs"
+           role="tab"
+           class="tab checked:text-primary"
+           aria-label="{% translate 'Sessions' %} {{ total_session_count }}"
```
SmittieC: We should either put the count in parentheses or it should be a different colour / size. The claude design from the issue seems to use a different font / color / size.

templates/participants/single_participant_home.html
```diff
@@ -1,78 +1,227 @@
...
+          </div>
+          <div class="flex items-center gap-2 flex-nowrap whitespace-nowrap text-xs text-base-content/70">
+            <span><i class="fa-regular fa-window-maximize text-base-content/50 mr-1"></i>{{ participant.get_platform_display }}</span>
+            <span class="text-base-content/30">|</span>
+            <span>{{ participant.identifier }}</span>
+            <span class="text-base-content/30">|</span>
+            <span>{{ experiments|length }} {% translate "chatbots" %}</span>
+          </div>
+        </div>
+        <div class="stats stats-horizontal">
```
SmittieC: I notice the design from the issue has a space between this section and the previous one

apps/chatbots/tables.py
```diff
@@ -241,3 +242,76 @@ class Meta:
...
+        if self._user_has_perm("experiments.view_experiment"):
+            return format_html(
+                '<a href="{}" class="font-medium link link-hover">{}</a>', chatbot.get_absolute_url(), str(chatbot)
+            )
+        return format_html('<span class="font-medium">{}</span>', str(chatbot))
+
+    def render_state(self, record):
+        return format_html('<span class="badge badge-ghost">{}</span>', record.get_status_display())
+
+    def render_tags(self, record, bound_column=None):
```
snopoke: This should use the `tag_ui.html` template for consistency with other UIs. See https://github.com/dimagi/open-chat-studio/blob/main/apps/chatbots/tables.py#L215

apps/chatbots/tables.py
```diff
@@ -241,3 +242,76 @@ class Meta:
...
+        attrs = {
+            **settings.DJANGO_TABLES2_TABLE_ATTRS,
+            "th": {"class": "py-3 px-3 text-left text-xs"},
+            "td": {"class": "py-3 px-3 text-left overflow-hidden text-xs"},
+        }
+        row_attrs = settings.DJANGO_TABLES2_ROW_ATTRS
+        orderable = False
+        empty_text = "No sessions yet!"
+
+    def render_chatbot(self, record):
```
snopoke: Keep this consistent with other tables: https://github.com/dimagi/open-chat-studio/blob/main/apps/chatbots/tables.py#L224

apps/chatbots/tables.py
```diff
@@ -241,3 +242,76 @@ class Meta:
...
+    def render_tags(self, record, bound_column=None):
+        tags = record.chat.non_skipped_tags()
+        if not tags:
+            return "-"
+        return format_html_join(
+            " ", '<span class="badge badge-outline badge-info">{}</span>', ((tag["name"],) for tag in tags)
+        )
+
+    def render_versions(self, value):
+        return f"v{value[-1]}" if value else "-"
```
snopoke: why only the last version?

apps/participants/views.py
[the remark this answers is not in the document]
SmittieC: personally I think it should be rolled with this PR

apps/chatbots/tables.py
```diff
@@ -241,3 +242,76 @@ class Meta:
...
+        verbose_name="",
+    )
+
+    class Meta:
+        sequence = ("chatbot", "started", "last_activity", "message_count", "state", "tags", "versions", "actions")
+        exclude = ("participant", "remote_id")
+        attrs = {
+            **settings.DJANGO_TABLES2_TABLE_ATTRS,
+            "th": {"class": "py-3 px-3 text-left text-xs"},
+            "td": {"class": "py-3 px-3 text-left overflow-hidden text-xs"},
```
snopoke, line 291: should `text-xs` be applied to all tables?

templates/participants/single_participant_home.html
```diff
@@ -1,78 +1,227 @@
...
+{% load form_tags humanize i18n static %}
+{% block page_head %}
+  {{ block.super }}
+  <style>
+    /* The header row above the editor already has a Format button; the JS-injected
+       floating one inside the editor would duplicate it. */
+    #participant-data-tab .json-editor > button {
+      display: none;
+    }
+    #participant-data-tab .json-editor .cm-editor,
```
snopoke: This editor shouldn't use custom styling - all the editors should have uniform styling to either we drop this or apply this to all the editors. I think we should drop this for now.

templates/participants/single_participant_home.html
```diff
@@ -1,78 +1,227 @@
 {% extends "web/app/app_base.html" %}
-{% load form_tags i18n static %}
+{% load form_tags humanize i18n static %}
+{% block page_head %}
+  {{ block.super }}
+  <style>
+    /* The header row above the editor already has a Format button; the JS-injected
+       floating one inside the editor would duplicate it. */
+    #participant-data-tab .json-editor > button {
```
snopoke: This should be accomplished using a data attribute on the element which the JsonEditor JS class looks for e.g. `data-showFormatButton="false"`

Or else just use the in-editor format button and drop the custom one from this page

templates/participants/single_participant_home.html
```diff
@@ -1,78 +1,227 @@
...
-    <div class="grid grid-cols-6">
-      <div class="col-span-5">
-        <div class="flex">
-          <h1 class="pg-title">Participant details</h1>
+  <div class="app-card rounded-2xl border border-base-300 bg-base-100">
+    <div class="flex flex-wrap items-start justify-between gap-5">
+      <div class="flex items-start gap-5 flex-wrap">
+        <div class="avatar avatar-placeholder">
+          <div class="bg-primary/10 text-primary rounded-xl w-13">
+            <span class="text-lg font-bold">{{ participant.initials }}</span>
```
snopoke: The original design had `initials` but I think we can drop this avatar element since most participants don't have names set.

templates/participants/partials/participant_sessions_pills.html
```diff
@@ -1,19 +1,21 @@
...
   <div class="w-px h-5.5 bg-base-300 mx-1"></div>
-  <a href="?chatbot="
-     hx-get="{% url 'participants:sessions-panel' request.team.slug participant.id %}?chatbot="
+  <a href="?chatbot={% if request.GET.q %}&q={{ request.GET.q|urlencode }}{% endif %}{% if request.GET.state %}&state={{ request.GET.state|urlencode }}{% endif %}{% if request.GET.tag %}&tag={{ request.GET.tag|urlencode }}{% endif %}"
+     hx-get="{% url 'participants:sessions-panel' request.team.slug participant.id %}"
+     hx-vals='js:{chatbot: "", q: new URLSearchParams(window.location.search).get("q") || "", state: new URLSearchParams(window.location.search).get("state") || "", tag: new URLSearchParams(window.location.search).get("tag") || ""}'
      hx-target="#sessions-panel"
      hx-swap="outerHTML"
-     hx-push-url="?chatbot="
+     hx-push-url="true"
```
snopoke: This shouldn't push the URL otherwise the URL becomes the HTMX url and if the user refreshes the page they only see the unstyled htmx partial.

apps/experiments/models.py
```diff
@@ -1235,6 +1243,71 @@ def get_schedules_for_experiment(
...
+            messages = messages.filter(is_complete=False, cancelled_at=None)
+
+        schedules = []
+        for message in messages:
+            schedule = message.as_dict(as_timezone=as_timezone)
+            schedule["experiment"] = experiments_by_id[message.experiment_id]
+            schedules.append(schedule)
+        return schedules
+
+    def get_message_trend(self, days: int = 30) -> list[int]:
```
snopoke, line 1258: What made you decide to query `ChatMessages` here instead of traces, like we do in `Experiment.get_bulk_trend_data`?

Seems like we could use almost the same code with slightly different filters.

apps/participants/views.py
```diff
@@ -212,6 +292,52 @@ def get_context_data(self, *args, **kwargs):
...
+    re-fires its own auto-load) on every swap if included here too, racing this endpoint's
+    own htmx-swapped table content -- no full page reload, so no scroll jump or reload flash.
+    """
+
+    permission_required = "experiments.view_participant"
+    template_name = "participants/partials/participant_sessions_panel.html"
+
+    def get_context_data(self, *args, **kwargs):
+        initial_context = super().get_context_data(*args, **kwargs)
+        return single_participant_home_context(
```
snopoke: Is it necessary to fetch the full context for these stubs? What about breaking the function up so that you can fetch each sections content individually?

apps/participants/views.py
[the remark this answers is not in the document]
SmittieC: I think that the existing filter should be ported to a [dynamic filter](https://github.com/dimagi/open-chat-studio/blob/20d62ced6f21ca89a8ee10f9a200d001f10b72e0/docs/developer_guides/code_systems/dynamic_filters.md). Why was the `ExperimentSessionFilter` introduced?

templates/participants/partials/participant_data.html
```diff
@@ -1,19 +1,71 @@
+{% load i18n %}
+{% if just_saved is not None %}
+  <span id="data-save-status"
```
snopoke: Where does this render? I couldn't see it in the UI

templates/experiments/filters.html
```diff
@@ -1,253 +1,314 @@
...
+       class="{% if simple_mode %}w-full p-4 bg-base-100 border border-base-300 rounded-lg mt-2{% else %}absolute left-0 mt-2 p-4 bg-base-100 border border-base-200 rounded-lg shadow-lg z-10 min-w-max max-w-screen-lg{% endif %}"
+       x-cloak>
+    {% if simple_mode %}
+      <form hx-get="{{ simple_filter_url }}"
+            hx-target="{{ simple_filter_target }}"
+            hx-swap="outerHTML"
+            hx-vals='js:{chatbot: new URLSearchParams(window.location.search).get("chatbot") || ""}'
+            class="flex gap-4 items-end">
+        <div class="w-64">
+          <label class="block text-sm font-semibold mb-1">{% translate "Search transcripts" %}</label>
```
SmittieC: I don't think we should be doing this. See [my comment](https://github.com/dimagi/open-chat-studio/pull/4393#discussion_r3948457375) about using a dynamic filter for the participant page

SmittieC (commented):
Some initial feedback

templates/participants/partials/participant_data_panel.html
```diff
@@ -0,0 +1,27 @@
...
+         hx-get="{{ data_panel_url }}?chatbot={{ experiment.id }}"
+         hx-target="#data-panel"
+         hx-swap="innerHTML"
+         hx-push-url="?chatbot={{ experiment.id }}"
+         class="btn btn-sm rounded-lg no-animation transition-none shadow-none border border-base-300 focus:outline-none focus-visible:outline-none active:outline-none
+                {% if selected_data_experiment == experiment %}btn-soft btn-primary hover:bg-primary/10 hover:text-primary hover:border-base-300{% else %}bg-base-100{% endif %}">{{ experiment }}</a>
+    {% endfor %}
+  </div>
+  {% if selected_data_experiment %}
+    <span id="data-save-status" class="text-sm text-base-content/60">{% translate "All changes saved" %}</span>
```
snopoke: This get's displayed even after editing the data (and before saving).

templates/participants/partials/participant_schedule_modal.html
snopoke, line 1: Is this not the same as the modal in `session_schedules_table.html`?

templates/participants/partials/participant_schedule_row.html
snopoke, line 1: Again here, can we not share the template from `session_schedules_table.html`?

templates/participants/single_participant_home.html
```diff
@@ -1,78 +1,227 @@
...
+          </div>
+        </div>
+        <div class="flex-1 min-w-70 flex flex-col gap-2">
+          <div class="flex items-center gap-2 flex-wrap">
+            <h1 class="pg-title text-lg shrink-0" id="participant-name">
+              {% include "participants/partials/participant_name.html" %}
+            </h1>
+            <span class="badge badge-sm whitespace-nowrap
+                         {% if latest_session.status == 'active' %}badge-success{% else %}badge-ghost{% endif %}">
+              {% if latest_session.status == 'active' %}
```
snopoke: I think this should be based on any usage in the last 30 days rather than the last session status.

templates/participants/single_participant_home.html
```diff
@@ -1,78 +1,227 @@
...
+        </div>
+        <div class="stats stats-horizontal">
+          <div class="stat py-2 px-0 pr-8 border-e-0">
+            <div class="stat-value text-lg">{{ total_session_count }}</div>
+            <div class="stat-title uppercase tracking-[.08em] text-[10px] font-semibold">{% translate "Sessions" %}</div>
+          </div>
+          <div class="stat py-2 px-0 pr-8 border-e-0">
+            {% with last_seen=participant.last_seen %}
+              <div class="stat-value text-lg">{{ last_seen|naturaltime|default:"-" }}</div>
+              <div class="stat-title uppercase tracking-[.08em] text-[10px] font-semibold">
```
snopoke: since this style is reused in a few places consider creating a `stat-title-upper` style in `site-tailwind.css`.

Also, the text size shouldn't be absolute pixel sizing - always use `rem` to support zoom etc.

templates/participants/single_participant_home.html
```diff
@@ -1,78 +1,227 @@
...
-          <h1 class="pg-title">Participant details</h1>
+  <div class="app-card rounded-2xl border border-base-300 bg-base-100">
+    <div class="flex flex-wrap items-start justify-between gap-5">
+      <div class="flex items-start gap-5 flex-wrap">
+        <div class="avatar avatar-placeholder">
+          <div class="bg-primary/10 text-primary rounded-xl w-13">
+            <span class="text-lg font-bold">{{ participant.initials }}</span>
+          </div>
+        </div>
+        <div class="flex-1 min-w-70 flex flex-col gap-2">
```
snopoke: Can we make the bottom row aligned?

Image

snopoke (commented):
This is looking good. One comment not raised inline is that I think we should drop the 'simple' filters and just use the default filter widget for now. If we're going to implement a simple filter option I think it needs it's own design time. Apologies for not noticing this in the mock up.

templates/participants/single_participant_home.html
preston-56: Fixed, switched to label-wrapped tabs so the count can be styled independently. It's now text-base-content/50 and a size smaller than the tab label.

templates/participants/single_participant_home.html
preston-56: Fixed, added ml-4 between the identity block and the stats.

apps/participants/views.py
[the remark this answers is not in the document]
preston-56: Done. Split into _sessions_panel_context/_schedules_panel_context/_data_panel_context, and each panel view now only builds its own tab's context instead of the whole page's.

apps/participants/views.py
preston-56: Same fix as the thread on line 340, these three views were the stubs it's talking about. Each now calls its own panel context builder instead of the full-page one.

apps/chatbots/tables.py
preston-56: Fixed, removed the override so it inherits the base table's tag_ui.html rendering.

apps/chatbots/tables.py
preston-56: Fixed, removed the override so it inherits the base table's chip rendering.

apps/chatbots/tables.py
preston-56: Fixed, now shows all versions instead of just the last one.

apps/chatbots/tables.py
preston-56: Left it scoped to this table for now. It has more columns than most (Chatbot, Started, Last activity, Messages, State, Tags, Version, Actions), so the smaller text helps it stay readable. Applying it site-wide feels like a separate call outside this PR's scope.

templates/participants/single_participant_home.html
preston-56: Fixed, dropped it. Also removed the custom header Format button so it just uses the standard floating one now, matching every other json-editor on the site.

templates/participants/single_participant_home.html
preston-56: Went with dropping the custom button entirely rather than adding the data attribute, same reasoning as the thread above.

templates/participants/single_participant_home.html
preston-56: Ended up dropping the custom styling entirely instead of moving it, per a suggestion below. This editor now just uses the standard styling like every other one.

templates/participants/single_participant_home.html
preston-56: Fixed, dropped it.

templates/participants/single_participant_home.html
preston-56: Fixed, added Participant.is_recently_active() (30-day window based on last message) and switched the badge to use that instead of the last session's status.

templates/participants/single_participant_home.html
preston-56: Fixed, added a stat-title-upper utility in site-tailwind.css using text-xs instead of the raw px value, and reused it for all three stat titles.

templates/participants/single_participant_home.html
preston-56: Fixed. The two rows had different font sizes (text-xs vs text-[10px]) which threw off the baseline. Fixing the stat-title size to match, plus switching the header to items-center, lines them up now.

image

apps/participants/views.py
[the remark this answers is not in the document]
preston-56: Update: went further than this. Per the thread below, the simple filter is gone entirely now in favor of the dynamic filter widget, so this specific gap doesn't apply anymore.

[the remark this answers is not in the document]
preston-56: Went with dropping the simple filter entirely instead of porting it, per a review comment suggesting exactly that. The dynamic filter widget was already being built in the backend (get_filter_context_data was computed but never rendered), just switched the template to use it like every other page does.

templates/experiments/filters.html
preston-56: Moot now, the whole simple filter form is gone.

templates/participants/partials/participant_sessions_pills.html
preston-56: Fixed, pills now push an explicit relative URL matching their href instead of `true`, so a refresh lands on the full page, not the bare htmx fragment. Also found the same gap one level up: once the dynamic filter widget is active, its own AJAX fetch only forwarded f_/op_ params, so it silently dropped the chatbot pill scope too. Fixed that in the shared filter component so it forwards any non-filter query param.

templates/participants/partials/participant_data.html
preston-56: Found it, there were two elements with the same id (a static one in participant_data_panel.html, and this OOB one), so the browser only ever picked up the first, static one. Removed the duplicate, now this is the only one and it updates correctly on save.

templates/participants/partials/participant_data_panel.html
preston-56: The duplicate-id bug above was the direct cause, that static span was frozen and never reflected an in-progress edit. It's fixed in the sense that the text is now accurate immediately after a save. It still won't say "unsaved" while you're actively typing before hitting Save though, that needs real dirty-tracking in JS. Happy to add that if you want it in scope here, otherwise I'd treat it as a follow-up.

apps/experiments/models.py
preston-56: Went with ChatMessage on purpose: this is meant to show raw conversation volume per participant, not LLM call success/error rate, and the two aren't the same number (one message can trigger multiple traces, and not every message path is traced). get_bulk_trend_data's Trace query is also scoped by experiment, not participant, so reusing it here would need a different join anyway.

templates/participants/partials/participant_schedule_modal.html
preston-56: Yeah, this is basically the same modal as session_schedules_table.html's, the participant version just also shows which chatbot each schedule belongs to since it aggregates across all of them. Didn't want to merge them in the same pass as everything else here, happy to do it as a follow-up if that's preferred.

templates/participants/partials/participant_schedule_row.html
preston-56: Same as the modal thread, real overlap. Proposing as a follow-up rather than merging templates in this pass.

preston-56:
@snopoke Done, the simple filters are gone. The dynamic filter widget was already being built server-side (get_filter_context_data / ExperimentSessionFilter) but never actually rendered, so this was mostly a matter of switching the template to use it like every other page does, plus removing the now-dead simple filter code. Kept the chatbot quick-filter pills since they're a one-click scope switch the mockup called for specifically, separate from the generic column filter. While wiring this up I found the pills and the dynamic filter widget didn't talk to each other (the widget's own AJAX fetch dropped the pill's chatbot scope), fixed that too, details in the thread on participant_sessions_pills.html.

apps/experiments/models.py
snopoke: OK but let me just correct some points:

> not LLM call success/error rate

traces show error / success for the chatbot as a whole, not just LLM calls

> and the two aren't the same number (one message can trigger multiple traces, and not every message path is traced)

A single message can not trigger more than one trace and I do think that every message path is traced (open to be proved wrong).

apps/experiments/models.py
```diff
@@ -1116,6 +1116,14 @@ def global_data(self):
             return {"name": self.name}
         return {}
 
+    @property
+    def initials(self) -> str:
```
snopoke: I think this is now unused outside of tests

apps/participants/views.py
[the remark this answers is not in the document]
snopoke: Open as a new issue

templates/participants/partials/participant_data.html
```diff
@@ -1,19 +1,64 @@
+{% load i18n %}
+<span id="data-save-status"
```
snopoke, line 2: This can be removed - it isn't rendered anywhere

templates/participants/partials/participant_schedule_modal.html
snopoke: +1 to merge - followup PR is fine - maybe consider a django-tables2 table which is easier to modify for different uses.

templates/participants/single_participant_home.html
```diff
@@ -1,78 +1,197 @@
...
-          <h1 class="pg-title">Participant details</h1>
+  <div class="app-card rounded-2xl border border-base-300 bg-base-100">
+    <div class="flex flex-wrap items-start justify-between gap-5">
+      <div class="grid grid-cols-[minmax(0,1fr)_auto] items-start gap-x-8 gap-y-2 flex-1">
+        <div class="grid grid-rows-subgrid row-span-2 gap-2 min-w-70">
+          <div class="flex items-center gap-2 flex-wrap">
+            <h1 class="pg-title text-lg shrink-0" id="participant-name">
+              {% include "participants/partials/participant_name.html" %}
+            </h1>
+            <span class="badge badge-sm whitespace-nowrap
```
snopoke: let's add a tooltop to this badge:

`Has been active in the last 30 days`

snopoke (commented):
A few small cleanups but otherwise looks good. There is still one interactivity issue which we should look into:

When the user selects a chatbot on one of the tabs the change does not propagate to the other tabs to clicking to a different tab results in a difference between the URL `chatbot` parameter and what the pill filter shows.

apps/experiments/models.py
preston-56: Fixed, removed it and its dedicated test class.

apps/participants/views.py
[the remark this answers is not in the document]
preston-56: Done, opened #4441 to track it: https://github.com/dimagi/open-chat-studio/issues/4441

templates/participants/partials/participant_data.html
preston-56: It does render, just above the chatbot pills on the Data tab: "All changes saved" / "Save failed", updated via an OOB swap on form submit. Confirmed by loading the page directly, correct in both position and text. Might have been reviewed against an earlier commit, happy to pair if it's still not showing for you.

templates/participants/partials/participant_schedule_modal.html
preston-56: Thanks, opened #4442 to track it and noted your django-tables2 suggestion there: https://github.com/dimagi/open-chat-studio/issues/4442

templates/participants/single_participant_home.html
preston-56: Fixed, added the tooltip.

apps/experiments/models.py
preston-56: Fair, you're right on both counts. Trace has a direct participant FK, and both the ad-hoc and normal pipeline paths wrap in trace_or_span unconditionally, so message paths aren't untraced. Switched get_message_trend to query Trace scoped by participant, same bucket-and-zero-fill pattern as get_bulk_trend_data, just daily over 30 days instead of hourly over 24h. Verified against real trace data, still renders correctly.

preston-56:
Fixed. Each tab's radio input now fires an htmx GET (on change) to refresh just that tab's own panel using the current ?chatbot= value, so switching tabs always shows fresh data with the correct pill highlighted instead of whatever was rendered on the last full page load. Verified live: select a chatbot on the Sessions tab, then switch to Schedules and Participant Data, both immediately reflect the same chatbot with the correct pill highlighted, and switching back to Sessions keeps it consistent too.

SmittieC (commented):
@preston-56 thanks for the updates. I'm looking through it now. One thing I'm noticing is that the filter on the participant screen is cut off. I think we just need a z-index on it

Image

preston-56:
> @preston-56 thanks for the updates. I'm looking through it now. One thing I'm noticing is that the filter on the participant screen is cut off. I think we just need a z-index on it
> 
> Image

Couldn't reproduce a cut-off at a few widths/themes. Dropdown renders on top fine. What's your browser width and zoom when this happens? @SmittieC

SmittieC:
> > @preston-56 thanks for the updates. I'm looking through it now. One thing I'm noticing is that the filter on the participant screen is cut off. I think we just need a z-index on it
> > Image
> 
> Couldn't reproduce a cut-off at a few widths/themes. Dropdown renders on top fine. What's your browser width and zoom when this happens? @SmittieC

if you couldn't reproduce it, then I think it's fine

preston-56:
> if you couldn't reproduce it, then I think it's fine

Thanks.

apps/chatbots/views.py
```diff
@@ -635,13 +635,32 @@ class ChatbotSessionsTableView(LoginAndTeamRequiredMixin, PermissionRequiredMixi
     """View for rendering chatbot sessions table with filtering support."""
 
     model = ExperimentSession
-    table_class = ChatbotSessionsTable
     template_name = "table/single_table.html"
     permission_required = "experiments.view_experimentsession"
 
+    def get_table_class(self):
```
SmittieC: It's feels a bit strange to render a different table when a certain path parameter is present (i.e. the `participant_id` one). I suggest you create a dedicated view that subclasses `ChatbotSessionsTableView`. I'd build in a hook inside `get_queryset` to get extra_filters from the class. Something akin to this:

```python
class ChatbotSessionsTableView(...):
    def extra_filters(self):
        return Q()

class ParticipantScopedSessionsTableView(ChatbotSessionsTableView):
    def extra_filters(self):
        participant_id = self.kwargs.get("participant_id")
        chatbot_id = int(self.request.GET.get("chatbot", ""))
        return Q(participant_id=participant_id, experiment_id=chatbot_id)
```

apps/experiments/models.py
```diff
@@ -1235,6 +1240,69 @@ def get_schedules_for_experiment(
                 scheduled_messages.append(message.as_string(as_timezone=as_timezone))
         return scheduled_messages
 
+    def get_schedules_for_all_experiments(self, as_timezone: str | None = None, include_inactive=False) -> list[dict]:
```
SmittieC: Suggestion: How about replacing `get_schedules_for_experiment` with `get_schedules_for_all_experiments`, and updating all current calls to `get_schedules_for_experiment` to pass in the experiment to scope it to? The logic seems very similar

SmittieC (commented):
Looking good

apps/chatbots/views.py
preston-56: Fixed, added ParticipantScopedSessionsTableView with an extra_filters() hook, per the suggestion.

apps/experiments/models.py
preston-56: Fixed, merged into one method with experiment_id optional.

apps/participants/views.py
```diff
@@ -113,7 +113,7 @@ def _sessions_panel_context(
 def _schedules_panel_context(
     request, participant: Participant, experiments: list[Experiment], filter_experiment_id: int | None
 ) -> dict:
-    schedules = participant.get_schedules_for_all_experiments(include_inactive=True)
+    schedules = participant.get_schedules_for_experiment(as_dict=True, include_inactive=True)
```
SmittieC: I meant for you to keep `get_schedules_for_experiments` (the plural method name). Sorry if this wasn't clear

preston-56: Fixed, renamed to the plural.

templates/experiments/filters.html
```diff
@@ -1,253 +1,264 @@
+{% load i18n %}
```
SmittieC: this doesn't seem to be used?

SmittieC (dismissed):
LGTM

templates/experiments/filters.html
preston-56: Fixed, removed.

SmittieC (approved):

PR #4395, opened 2026-09-03
Add operator-level Langfuse tracing to the System Agent help bot

### LINKED ISSUE #2024: Add tracing with langfuse to the python code help bot
See code in apps/help/agent.py

Davemaina1:
Closes #2024

### Product Description
The System Agent (`apps/help/`: code generation, filter building, progress messages) had no observability. No record of what was asked, what came back, latency, or failures. It now optionally traces every call to Langfuse.

### Technical Description
Tracing is operator-level (`LANGFUSE_PUBLIC_KEY`/`LANGFUSE_SECRET_KEY`/`LANGFUSE_HOST`/`LANGFUSE_SAMPLE_RATE`), matching the existing `SYSTEM_AGENT_MODELS_HIGH/LOW` pattern, rather than reusing a team's own Trace Provider. `TraceProvider` is only reachable via `Experiment.trace_provider` (a per-chatbot FK, no "team default" concept), and `Tracer`/`TracingService` are shaped around `ExperimentSession`. The System Agent has neither an Experiment nor a Session. `BaseHelpAgent` gains a `_trace()` context manager (`apps/help/base.py`) that all three agents wrap their `run()` body in. It's a no-op when tracing isn't configured, or when trace setup itself fails. `apps/help/tracing.py` builds the tracer by reusing the existing `LangFuseTracer` class directly (no new SDK integration). Its sample-rate normalization moved to `apps.service_providers.tracing.langfuse` and is now shared with `TraceProvider.get_service()`'s own copy of the samelogic, which landed on `main` partway through this PR.

### Demo
To enable and see it: set `LANGFUSE_PUBLIC_KEY`/ `LANGFUSE_SECRET_KEY` (`LANGFUSE_HOST`/`LANGFUSE_SAMPLE_RATE` optional), restart, then use the "Generate" button on a Python Node's Code editor. A trace named `help_agent:code_generate` appears in the configured Langfuse project; filter building and progress-message generation trace the same way under `help_agent:filter` and `help_agent:progress_messages`. Verified locally with `inv dev` that leaving the env vars unset (the default) traces nothing and changes no behavior.

### Docs and Changelog
- [x] This PR requires docs/changelog update: new env vars, `CHANGELOG.md`   updated, `docs/hosting/configuration.md` updated.

### Operator Impact
- [x] Self-hosted operators must know about or act on this change: new optional env vars for System Agent tracing (no action required if unset).


apps/help/base.py
```diff
@@ -1,10 +1,19 @@
 from __future__ import annotations
 
+import logging
+import uuid
+from collections.abc import Iterator
+from contextlib import ExitStack, contextmanager
 from typing import ClassVar, Literal
 
+from langchain_core.runnables import RunnableConfig
```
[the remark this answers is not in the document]
Davemaina1: Fixed in dec588b95: moved to a TYPE_CHECKING-only import at module level, with the real import inside _trace() where RunnableConfig is actually constructed.

apps/help/tracing.py
```diff
@@ -0,0 +1,27 @@
...
+        return None
+
+    sample_rate = normalize_sample_rate(settings.LANGFUSE_SAMPLE_RATE)
+    if sample_rate is None:
+        return None
+
+    config = {
+        "public_key": settings.LANGFUSE_PUBLIC_KEY,
+        "secret_key": settings.LANGFUSE_SECRET_KEY,
+        "host": settings.LANGFUSE_HOST,
```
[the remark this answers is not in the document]
Davemaina1: Skipping: the existing LangfuseTraceProviderForm.host (the per-team equivalent of this same config value) is a plain URLField with no scheme restriction, and a self-hosted Langfuse instance without TLS is a deployment that form already permits. Enforcing HTTPS only on this new operator env var would be inconsistent with that established, already-reviewed precedent for the identical value.

apps/service_providers/tracing/langfuse.py
```diff
@@ -30,6 +30,23 @@
...
+    ``LANGFUSE_SAMPLE_RATE`` env var or 1.0, so "trace nothing" has to be enforced here
+    rather than trusted to the SDK. A blank rate is normalized to ``1.0`` for the same
+    reason: passing ``None`` through lets the SDK fall back to that env var if one happens
+    to be set, silently overriding "leave blank to trace every call".
+    """
+    if sample_rate == 0.0:
+        return None
+    if sample_rate is None:
+        return 1.0
+    return sample_rate
```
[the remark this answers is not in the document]
Davemaina1: Skipping the suggested move: verified empirically (django.setup()) that apps.service_providers.tracing.langfuse is already imported unconditionally at plain Django startup, via existing forms.py imports, independent of this change, so the module-level ClientManager and its background thread are already running before this code path ever executes. The suggested fix would not actually change that either: apps/service_providers/tracing/__init__.py already does "from .langfuse import LangFuseTracer" eagerly, and a package's __init__.py always runs before any of its submodules, so moving the function to another module inside the same package would not avoid triggering it.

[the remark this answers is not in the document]
Davemaina1: Fixed in dec588b95: normalize_sample_rate() now raises ValueError outside 0.0-1.0, matching Langfuse's own validation but surfaced earlier. Added parametrized tests. This also revealed a related gap: get_help_agent_tracer() itself could now raise and was not covered by _trace()'s resilience handling, fixed that too, with its own regression tests.

config/settings.py
```diff
@@ -1006,6 +1006,15 @@ def _before_create(task: dict):
 # 'low' models used for simple tasks
 SYSTEM_AGENT_MODELS_LOW = get_system_agent_models(agent_models_low, agent_api_keys)
 
+# Operator-level Langfuse tracing for the system agent (apps/help/). Separate from a team's own
+# Trace Provider: the system agent has no Experiment/Session to attach one to, and its model
+# choice is already operator-configured above, not per-team.
+LANGFUSE_PUBLIC_KEY = env("LANGFUSE_PUBLIC_KEY", default="")
+LANGFUSE_SECRET_KEY = env("LANGFUSE_SECRET_KEY", default="")
+LANGFUSE_HOST = env("LANGFUSE_HOST", default="https://cloud.langfuse.com")
```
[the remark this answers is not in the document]
Davemaina1: Same finding as the apps/help/tracing.py thread: skipping for consistency with LangfuseTraceProviderForm.host (no scheme restriction on the identical per-team config value), and because a non-TLS self-hosted Langfuse instance is a deployment that form already permits.

SmittieC (commented):
LGTM. There's just some merge conflicts

apps/help/base.py
```diff
@@ -47,13 +58,68 @@ def get_system_prompt(cls, input: TInput) -> str:
...
+        pointing LANGFUSE_HOST at an unresolvable domain and exercising this code path in a
+        real browser session: the request completed normally, and the export failure showed
+        up only as an "opentelemetry.sdk._shared_internal" log line, never reaching here.
+        get_help_agent_tracer() itself is covered too, in its own try/except that contains
+        no yield: an out-of-range LANGFUSE_SAMPLE_RATE raises there (see
+        normalize_sample_rate), and it must not break the request either. Keeping that
+        outside the trace-entry try/except matters: a yield inside a try/except can receive
+        an exception thrown back in from the caller's own code (e.g. a real bug in
+        agent.invoke()) when the `with` block exits, and that must propagate untouched, not
+        get misreported as a tracing failure.
```
snopoke: this can be reduced a lot

apps/help/base.py
```diff
@@ -47,13 +58,68 @@ def get_system_prompt(cls, input: TInput) -> str:
...
+        """
+        from langchain_core.runnables import RunnableConfig  # noqa: PLC0415 - lazy: avoids loading langchain_core
```
snopoke: this isn't a 'heavy' module listed in pyproject.toml (`banned-module-level-imports`). There's no reason to inline it.

config/settings.py
```diff
@@ -1006,6 +1006,15 @@ def _before_create(task: dict):
...
+# Fraction of system agent calls to trace, from 0.0 to 1.0. Leave unset to trace every call.
+LANGFUSE_SAMPLE_RATE = env.float("LANGFUSE_SAMPLE_RATE", default=None)
```
snopoke: I think we should prefix these env vars since the langfuse client also looks for these exact keys: https://langfuse.com/docs/observability/sdk/overview.

I would suggest `OCS_...`

snopoke (commented):
Please try to reduce the comment verbosity. Consider adding this to your local claude file: https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/prompting-claude-fable-5-1#writing-density

Davemaina1:
Hey @SmittieC, just a heads up, the merge conflicts are sorted now. Pulled in the latest from main and pushed, so the PR should be clean and up to date again.

SmittieC (approved):

apps/help/base.py
Davemaina1: That is a good catch. I have cut it down to what it returns and why a tracing failure doesn't break the underlying call.

apps/help/base.py
Davemaina1: Fair point, I have looked into this. Moved RunnableConfig back to a normal top level import.

config/settings.py
Davemaina1: I have actioned this. I have renamed all four to the OCS_ prefixed versions across settings, tracing.py, the tests, .env.example and the docs.

snopoke (approved):

PR #4407, opened 2026-09-04
Allow teams to require MFA for their members

### LINKED ISSUE #147: Allow teams to require 2fa for users
This should be quite easy with https://django-allauth-2fa.readthedocs.io/en/latest/advanced/

Davemaina1:
### Product Description
Team admins can now require two-factor authentication for their team, from the team settings page. Once turned on, every member of that team has to set up MFA before they can use Open Chat Studio, regardless of which team's pages they're on.

### Technical Description
Adds an audited `Team.require_mfa` boolean rather than a feature flag: flags here are for staged rollout, not a standing setting (see the design-doc thread on the issue). It follows the same pattern as the existing `is_migrating` field, its own migration, a small `TeamMfaForm`, a `set_require_mfa` view gated to team admins (`teams.change_team`), and a checkbox on the team settings page under Members & access. `require_mfa` is added to `TEAM_FIELDS` so changes go through `field_audit` like the rest of `Team`.

`RequireMfaForStaffMiddleware` is renamed to `RequireMfaMiddleware`: its enrolment check now covers the existing staff/superuser gate (`REQUIRE_MFA_FOR_STAFF`, unchanged) or membership in a team with `require_mfa` set (`user.teams.filter(require_mfa=True).exists()`). Gating is site-wide, the same all-or-nothing shape the staff gate already had, so a user in two teams where only one requires MFA is still gated everywhere.

The redirect message changed from "required for staff accounts" to "required for this account", the old copy was wrong once a non-staff member could see it too.

`django-allauth-2fa`, linked from the issue, isn't used. The repo runs allauth's native `mfa` app, and the two packages don't run together; the one useful idea from `django-allauth-2fa` (its `BaseRequire2FAMiddleware` pattern) is already what this middleware follows, per its own docstring.

Design doc: https://github.com/dimagi/open-chat-studio/issues/147#issuecomment-5533167902 (updated after review: https://github.com/dimagi/open-chat-studio/issues/147#issuecomment-5538778026)

### Issue Link
Closes #147

### Migrations

Adds `require_mfa` to `Team` with a DB-level default of `False`, so existing rows and the currently-running release are both unaffected.

### Demo
1. As a team admin, go to Team Settings and check "Require two-factor authentication" under Members & access, then save.
image
image
image

2. Any member of that team, including the admin who just turned it on, is redirected to MFA setup on their next page load until they enrol.
3. A member of a different team that hasn't enabled the setting is unaffected.

Manual check: throwaway team, admin and regular member. Turning the setting on gated both immediately; a member of a different, unflagged team stayed unaffected.

### Docs and Changelog

Notes for changelog: new team-level setting on the team settings page — "Require two-factor authentication." No new environment variable, a self-service in-app toggle.

### Operator Impact

New `Team.require_mfa` column, logged under Migrations in `CHANGELOG.md`. No action required, default is off, existing rows are unaffected.


snopoke (commented):
Looks good but I think this field should be shown within the "Members & Access" section of the settings page.

Davemaina1:
@snopoke, I have actioned this and moved the field to the Members & access section. I have added the necessary screenshots to the demo section of the PR description.

snopoke (approved):

PR #4411, opened 2026-09-04
Handle the DRF 3.17 request body cap and fix stale usage artefacts

### LINKED ISSUE #4365: DRF 3.17.2 follow-ups: 2.5 MB JSON body cap and MultipleChoiceField list change
Follow-ups from the DRF 3.15.2 → 3.17.2 bump ([#4360](https://github.com/dimagi/open-chat-studio/pull/4360)). See the [dependency review comment](https://github.com/dimagi/open-chat-studio/pull/4360#issuecomment-5501996011) for the full analysis.

## 1. Handle the new 2.5 MB cap on JSON request bodies

DRF 3.17.2 enforces `DATA_UPLOAD_MAX_MEMORY_SIZE` when parsing `request.data`. OCS never sets it, so Django's 2.5 MB default now caps **every** API endpoint (`DEFAULT_PARSER_CLASSES` is `JSONParser` only). Multipart uploads are exempt.

`RequestDataTooBig` is a `SuspiciousOperation`; DRF's handler returns `None` for it, so `apps/api/exception_handlers.py` passes it through and clients get a plain non-JSON HTTP 400 plus an ERROR on the `django.security` logger (Sentry noise).

- [ ] Set the limit explicitly in `config/settings.py` so the API contract is a deliberate choice, e.g. `DATA_UPLOAD_MAX_MEMORY_SIZE = env.int("DATA_UPLOAD_MAX_MEMORY_SIZE", default=10 * 1024 * 1024)`
- [ ] Handle `RequestDataTooBig` in `apps/api/exception_handlers.py`, mapping it to a JSON 413 (or JSON 400) with a `detail` message

Endpoints with genuinely unbounded JSON bodies: session create `messages` + `state` (`apps/api/serializers.py:246-247`), participant data update (`apps/api/serializers.py:302-312`), OpenAI-compatible completions `messages` (`apps/api/openai.py:19`), `metadata` (`apps/api/serializers.py:142`).

## 2. `MultipleChoiceField` now returns a list, not a set

Nothing breaks at runtime — every consumer uses `in` / iteration and there are no set operations on `metrics` — but two artefacts are stale:

- [ ] `apps/api/v2/usage/param_serializers.py:52-53` — comment says "Returns a set, so duplicates collapse"; it now returns an ordered list (duplicates still collapse)
- [ ] `apps/api/v2/usage/services.py:131` — `metrics: set[str]` no longer matches what `apps/api/v2/usage/views.py:180` passes; change to `Collection[str]` (or `list[str]`). `ty` doesn't flag it because the value arrives as `Any`. Note the tests construct `UsageQuery(metrics=set(...))` (`apps/usage_metrics/tests/test_equality.py:83`, `test_characterisation.py:56,194`), so tests and production currently differ in container type

## Testing

- POST a >2.5 MB JSON body to session-create and to participant-data update; confirm the response shape is acceptable to API consumers
- `GET /api/v2/usage/?metric=cost&metric=messages` — confirm response blocks and dedup are unchanged
- Multipart upload on `apps/api/views/chat.py:158` — confirm large files still succeed

## Heads-up for the next bump

DRF 3.18.0 (7 Aug 2026) changes list-serializer `many=True` errors from list format to dict format — a real breaking change for any client or test asserting on bulk-validation error bodies. Budget review time when Dependabot proposes it.

zandre-eng:
### Product Description

API clients sending a large JSON body now get a parseable JSON error instead of an HTML page. Any body over 2.5 MB previously failed with a plain-text `400` and logged an error to `django.security`. The limit is now a deliberate 10 MB and the response states it:

```json
{"detail": "Request body too large. The maximum is 10485760 bytes."}
```

File uploads are unaffected. No UI changes.

### Technical Description

DRF 3.17.2 reads `request.body` when parsing JSON, so Django's `DATA_UPLOAD_MAX_MEMORY_SIZE` began capping every endpoint at its 2.5 MB default. Three changes:

- `config/settings.py` — set the limit explicitly (10 MB, env-overridable) so the API contract is ours rather than an inherited default.
- `apps/api/exception_handlers.py` — map `RequestDataTooBig` to a JSON `413`, *before* delegating to `drf_exception_handler`, which returns `None` for a `SuspiciousOperation`.
- `apps/web/request_logging_middleware.py` — the non-obvious one. The middleware re-reads `request.body` after the view, and Django's `body` raises before caching `_body`, so it re-raised and Django replaced the new `413` with its plain-text `400`. Invisible to tests: `JSON_LOGGING = not (DEBUG or IS_TESTING)`, so only a run with `ENABLE_JSON_LOGGING=True` exercises it.

Part two of the ticket is artefact-only: `MultipleChoiceField` now returns an ordered list, so a stale "returns a set" comment is corrected and `UsageQuery.metrics` widens to `Collection[str]` — what its consumers actually require. `ty` missed the mismatch because the value arrives as `Any`.

Note: the ticket lists `metadata` (`serializers.py:142`) as unbounded, but that field is `read_only=True`, so clients cannot submit it.

Tests cover oversized bodies on the three unbounded endpoints, a multi-megabyte body succeeding under the limit, multipart staying exempt, and the middleware surviving an unreadable body. Each was checked to fail when its own fix is removed.

### Issue Link

https://github.com/dimagi/open-chat-studio/issues/4365.

### Demo

No user-facing changes.

### Docs and Changelog

Consumers gain a documented body limit and a new `413`. Left unchecked since ticking it dispatches to the docs repo on merge.

### Operator Impact

The setting is global to Django, so it also bounds web-UI form posts and unauthenticated webhooks. Going from 2.5 MB to 10 MB quadruples what one unauthenticated request can make a worker buffer. Override `DATA_UPLOAD_MAX_MEMORY_SIZE` to lower it. Logged under `[Unreleased] > Configuration` in `CHANGELOG.md`.


CHANGELOG.md
```diff
@@ -37,6 +37,13 @@ version section when a release is cut.
 - `OCS_VERSION`: new, optional, defaults to `latest`. Read by
   `docker-compose.prod.yml` to select which published image tag to run. Pin it
   to the release you intend to run rather than tracking `latest`. (#4283)
+- `DATA_UPLOAD_MAX_MEMORY_SIZE`: new, optional, defaults to `10485760` (10 MB).
```
snopoke, line 44: This should also be mentioned in the `docs/hosting/configuration.md` file and then rather point this changelog to that entry instead of duplicating the explanation.

config/settings.py
```diff
@@ -447,6 +447,11 @@ def _public_media_url(custom_domain: str, bucket_name: str, location: str) -> st
 
 SITE_ID = 1
 
+# Global cap on any request body Django buffers in memory, so it bounds web-UI form posts and
+# unauthenticated webhooks as well as the JSON API (DRF reads ``request.body`` to parse JSON).
+# Multipart uploads stream to disk and are bounded by the per-endpoint file size checks instead.
```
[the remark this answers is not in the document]
snopoke: We can probably drop the comment or just link to the Django docs

snopoke (commented):
Small comments but otherwise LGTM

config/settings.py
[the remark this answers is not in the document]
zandre-eng: I shortened the comment here to rather just link to the Django docs.

Addressed in [cd2eb42](https://github.com/dimagi/open-chat-studio/pull/4411/commits/cd2eb4248a9ecb4b2fb793da2e935888035ef4ba).

CHANGELOG.md
zandre-eng: Ah, thanks for flagging. I added a section in `configuration.md` and then linked the changelog to it.

Addressed in [cd2eb42](https://github.com/dimagi/open-chat-studio/pull/4411/commits/cd2eb4248a9ecb4b2fb793da2e935888035ef4ba).

config/settings.py
```diff
@@ -447,6 +447,11 @@ def _public_media_url(custom_domain: str, bucket_name: str, location: str) -> st
 
 SITE_ID = 1
 
+# Set explicitly because Django's 2.5 MB default silently became the JSON API's request size limit
+# when DRF 3.17 started reading ``request.body`` to parse JSON.
```
snopoke:
```suggestion
```

docs/hosting/configuration.md
```diff
@@ -55,6 +55,7 @@ These apply whether the connection comes from `DATABASE_URL` or the variables ab
 | `OIDC_RSA_PRIVATE_KEY` | — | RSA private key (PEM format) for the built-in OAuth2/OIDC provider. Required to register an OAuth2 application at all: the registration form signs every application with `RS256`, which django-oauth-toolkit rejects when this is unset. Setting it also enables OIDC, signs ID tokens, and adds the `openid`/`profile` scopes. |
 | `OAUTH_PKCE_REQUIRED` | `True` | Require PKCE for OAuth2 flows. |
 | `HEALTH_CHECK_TOKENS` | `[]` | Comma-separated tokens for the `/status` health check endpoint. |
+| `DATA_UPLOAD_MAX_MEMORY_SIZE` | `10485760` (10 MB) | Largest request body Django will buffer in memory, bounding the JSON API, web-UI form posts and channel webhooks alike. Set explicitly because Django's own 2.5 MB default silently became the API's request size limit when DRF 3.17 started reading `request.body` to parse JSON. Lower it to tighten what a single request can buffer; raise it if legitimate API payloads are rejected with HTTP 413 (non-API requests get Django's own HTTP 400). See the [Django docs](https://docs.djangoproject.com/en/stable/ref/settings/#data-upload-max-memory-size) for exactly what the cap covers. |
```
snopoke:
```suggestion
| `DATA_UPLOAD_MAX_MEMORY_SIZE` | `10485760` (10 MB) | Largest request body Django will buffer in memory, bounding the JSON API, web-UI form posts and channel webhooks alike. See the [Django docs](https://docs.djangoproject.com/en/stable/ref/settings/#data-upload-max-memory-size). |
```

CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ version section when a release is cut.
...
+  Caps the request body Django buffers in memory. See the
+  [configuration reference](docs/hosting/configuration.md#security) for why it
+  is now set explicitly and when to change it. (#4411)
```
snopoke:
```suggestion
  Caps the request body Django buffers in memory. See the
  [configuration reference](docs/hosting/configuration.md#security). (#4411)
```

snopoke (dismissed):

snopoke (approved):

PR #4412, opened 2026-09-04
Show progress while saving an LLM provider verifies credentials

### LINKED ISSUE #4410: No UI feedback while saving an LLM provider runs the connection test
### Problem

On the LLM provider create/edit page, submitting the form with "Create and Test" / "Update and Test" runs a synchronous connection test against the provider before redirecting. That's a real LLM request with a 10s timeout (`CONNECTION_TEST_TIMEOUT_SECONDS`, `apps/service_providers/models.py:173`), so the page sits there for several seconds with no indication anything is happening.

The submit button also stays enabled, so it can be clicked repeatedly — each click re-posts the form and fires another external call.

### Where

`templates/service_providers/provider_form.html:78` — the LLM submit button has no submitting state. `templates/generic/object_form.html:26` already does this (`x-data="{ submitted: false }" @submit="submitted = true"` + `:disabled="submitted"`), the provider form just doesn't use it.

The standalone "Test Connection" button (`templates/service_providers/provider_form.html:105`) has the same issue — same test, same wait, no feedback.

### Suggested fix

Disable the button on submit and show a spinner/"Testing connection…" label while the request is in flight, for both the submit button and the Test Connection button. Reuse the pattern from `generic/object_form.html`; note the provider form already has an Alpine scope (`configChanged`) that the label depends on.

SmittieC:
### Product Description
On the LLM provider create/edit page, saving runs a real connection test against the provider (up to a 10s timeout) before redirecting. The page used to sit there with no feedback and an enabled button, so a repeated click fired another external call.

The submit button now disables itself on submit, shows a spinner, and changes its label to "Verifying credentials…".

### Technical Description
- The card's Alpine scope in `templates/service_providers/provider_form.html` now always exists and carries `submitted`; the main form sets it via `@submit`.
- The LLM submit button is `:disabled="submitted"` and renders a spinner plus a reactive label.
- `willVerify` (`configChanged || !credentialsVerified`) mirrors the condition the view uses to decide whether to run the connection test, so a save that skips the check says "Saving…" rather than promising a verification.
- Provider types with no credential check get `:disabled="submitted"` as well, matching `generic/object_form.html`, but no verification wording.
- Two existing tests asserted the old exact button markup and were updated for the span-wrapped label; new tests cover the disabled state and the verified/unverified label split.

### Issue Link
Fixes #4410

### Migrations

No migrations in this PR. Note: `main` currently has two leaf migrations in `service_providers` (`0078_add_gpt_6_astra` and `0078_llmprovider_extra_data`) which blocks the test DB from building - that needs a separate merge migration.

### Demo
This button is shown when clicked (and verification is needed)

image

### Docs and Changelog

### Operator Impact


apps/service_providers/tests/test_views.py
```diff
@@ -493,6 +493,49 @@ def test_a_save_never_fails_because_the_provider_did(self, team_with_users, auth
         assert provider.name == "Saved Anyway"
         assert provider.config.get("openai_api_key") == "a-new-key"
 
+    def test_the_button_reports_the_check_it_is_waiting_on(self, team_with_users, authed_client):
```
SmittieC: remove this test. low value

templates/service_providers/provider_form.html
```diff
@@ -19,8 +19,12 @@
   configChanged is scoped to the card rather than to the config fields, because the submit
   button reads it too and sits outside them. Only a provider whose credentials can be
   verified has anything to re-verify, so nothing else carries the scope.
+
```
SmittieC: remove this whole comment

SmittieC (commented):
@claude please look at all feedback

snopoke (approved):

PR #4418, opened 2026-09-05
Surface index failure reasons for remote-index collections

### LINKED ISSUE #3971: Surface index failure reasons for remote-index collections

barry47products:
### Product Description

When a file fails to index, it gets a red badge with a tooltip that should say why. On remote-index collections it usually didn't. The tooltip was empty, or just the class name `FileUploadError`, or "Failed to link files to OpenAI vector store", which is the same text whether the key was rejected, the quota ran out or the connection dropped. You had to go to the server log to find out.

| What failed | Tooltip before | Tooltip now |
| --- | --- | --- |
| Upload to the provider | `FileUploadError` | `FileUploadError: Incorrect API key provided: sk-not-a**********3971. ...` |
| Linking to the vector store | `UnableToLinkFileException: Failed to link files to OpenAI vector store` | `UnableToLinkFileException: Incorrect API key provided: ...` |
| Anything `add_files` does not catch | nothing, and the row never left `IN_PROGRESS` | the provider's message, and a terminal status |
| Switching a collection's provider | the previous provider's reason, against a pending row | cleared with the reset |

In the third case the badge also kept spinning, polling every five seconds for a job that had already died.

### Technical Description

A new helper, `provider_error_message(exc)`, returns `exc.body["message"]` if `body` is a mapping that has one, and `str(exc)` otherwise. It duck-types the exception, so `RemoteIndexManager` still imports no provider SDK. It reads `body["message"]` rather than `body["error"]["message"]` because `openai/_client.py` unwraps the error (`data = body.get("error", body)`) before building the exception. A test against a real `openai.AuthenticationError` pins this down: if the SDK changes shape, that test fails, instead of every reason quietly falling back to `str(exc)`.

Both remote raise sites now pass the message into their exception, and the handlers write what they're given. `EditCollection.form_valid` also clears the reason when a collection switches provider, the same way `retry_failed_uploads` does.

No model, template or migration changes; it's all Python. I moved the collection-indexing tests out of `test_tasks.py` into `test_collection_indexing.py`. Their shared fixture now lives in a package `conftest.py` under the name `remote_collection_index`, so it doesn't shadow the `collection` fixture in `test_document_sources.py`.

A few things outside the issue's scope that are worth a look:

- `add_files` filtered its batch updates on `file_id__in` without scoping them to the collection. A file can be in two collections, so a failure in one overwrote the other collection's status and reason. The fix is one extra filter keyword, in code this PR was already touching.
- I didn't wrap the indexing call in `transaction.atomic()`, even though the handler catches `DatabaseError` and then writes. `add_files` saves per-file status as it goes, and an outer atomic block would roll that back along with embeddings that each cost a provider call. The recovery write catches `DatabaseError` itself instead, like the per-file save in `LocalIndexManager.add_files`. That matters because `create_new_version` calls this synchronously inside `@transaction.atomic()`, where an aborted transaction would hide the original error. It doesn't cover everything: if the connection is dead, the row stays `IN_PROGRESS` and the failure is logged.
- `index_collection_files` now closes its `.iterator(100)` in a `finally`. Before, an exception leaving the loop left the cursor open until garbage collection. The close is wrapped in `except Exception` on purpose, because a failing `CLOSE` comes back as a raw `psycopg` error rather than a Django one. Happy to split this into its own PR if you'd prefer.

Not included: per-file reasons when a batch link partly fails. That needs tracking which files went into which chunk.

CodeScene still flags `test_index_managers.py` for being slightly over its 600-line limit. Getting under it means moving `RemoteIndexManagerMock` and two shared fixtures that older tests depend on, which felt like a lot of churn for an advisory in a feature PR. I can do it if you'd like it green.

### Issue Link

Closes #3971. Split out of #3164, which did the local half.

### Migrations

None. No model fields changed.

### Demo

A collection whose provider has a rejected key, showing both failure stages.

| Stage | Before | After |
| --- | --- | --- |
| **Upload** | before-upload-failure | after-upload-failure |
| **Link** | before-link-failure | after-link-failure |

The "after" tooltips come from a real 401. For the "before" shots I wrote the strings main produces onto the same rows, so both could be screenshotted.

1240 tests pass across `apps/documents` and `apps/service_providers`.

### Docs and Changelog

A failed file in a remote-index collection now shows the provider's explanation instead of an exception class name.

### Operator Impact

Provider error text now shows up in the collection file list for remote indexes, where before it only went to the log. OpenAI masks the key in its own 401 message; other providers may not. This PR adds no redaction. If it did, it would need to cover the local path too, which has been showing this text for a while.


apps/documents/tests/test_collection_indexing.py
```diff
@@ -0,0 +1,277 @@
...
+
+    index_collection_files_task([collection_file.id])
+
+    collection_file.refresh_from_db()
+    assert collection_file.status == FileStatus.IN_PROGRESS
+    assert collection_file.failure_reason == ""
+
+
+@pytest.fixture()
+def assistant_with_remote_file(remote_collection_index):
```
SmittieC: All assistant-related tests are redundant as of this morning when [assistants were removed](https://github.com/dimagi/open-chat-studio/pull/4372).

SmittieC (dismissed):

apps/documents/tests/test_collection_indexing.py
barry47products: Thanks, removed in 92602a3 along with the task itself after merging main. The merge dismissed your approval, so it needs another tick when you have a moment.

SmittieC (approved):

PR #4423, opened 2026-09-06
Fix translation and tag filter state lost on page refresh

### LINKED ISSUE #4397: Translation settings lost on page refresh
## Root cause (confirmed)

From #2585: "Translation context is lost if you refresh the page."

The language `<select>` and "Show Original Messages" toggle in `experiment_session_messages_control_panel.html` both read their state from `request.GET.language` / `request.GET.show_original_translation`, and the whole "Translated" badge block is gated on `{% if language %}`, all server-side context derived from the URL. But the form that submits them has no `hx-push-url`:

```html
hx-get="{% url 'experiments:experiment_session_messages_view' ... %}"
hx-target="#messages-container"
hx-trigger="change, submit">
```

Picking a language or toggling the option updates `#messages-container` but never touches the browser's address bar. On a real refresh, the URL has no `?language=` param, `request.GET.language` comes back empty, and the entire translated view silently resets to the original messages.

## Proposed work

Add `hx-push-url="true"` to this form, the same fix already used elsewhere in the participant-details redesign for its search/state/tag filters. Verify the `search` hidden input and any other control-panel field still survive the push correctly, since they rely on the same GET-param round-trip.

preston-56:
### Product Description
Fixes translation settings and tag/search filters on the session detail page resetting when the page is refreshed.

### Technical Description
`experiment_session_messages_view` only ever renders the messages fragment, `experiments/components/session_messages.html`, never the full session page. The issue's proposed fix (`hx-push-url="true"` on the control panel form) would technically make the address bar reflect the filter state, but it would point the browser at this bare fragment's own URL. A refresh on that URL renders the fragment standalone, with no page layout, navigation, or stylesheet, since Django never wraps it in the base template for a direct request. Confirmed this live before proposing the real fix: screenshot in the before-image, a fully unstyled page.

Used the `HX-Push-Url` response header instead of the `hx-push-url` attribute, since htmx gives the header priority over the attribute regardless of what any element sets. The view now computes the full `chatbots:chatbot_session_view` URL plus the current querystring and pushes that instead, only on htmx requests (`request.htmx`). That page's own `#messages-container` already re-fetches this same view with `{% querystring %}` on load, so landing back there on refresh reconstructs the exact same filtered/translated view.

fixes #4397

### Demo
<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
before-refresh
</td>
<td>
after-refresh
</td>
</tr>
</table>

Both screenshots show the same session after applying the "billing" tag filter and then refreshing the page. Before: the filter chip still looks selected (the browser remembers the checkbox state), but all 3 messages show since the server never received the filter, since the URL never changed. After: the filter round-trips correctly and only the tagged message shows.

### Docs and Changelog

### Operator Impact


apps/experiments/views/experiment.py
```diff
@@ -661,11 +675,13 @@ def experiment_session_messages_view(request, team_slug: str, experiment_id: uui
         "highlight_message_id": highlight_message_id,
     }
 
-    return TemplateResponse(
+    response = TemplateResponse(
```
preston-56: Already 107 lines before this change (threshold is 80), pre-existing. The new logic is already isolated in its own helper; what's left here is just the call site. Shrinking the rest is out of scope for this fix.

SmittieC (approved):

Issue #2420, opened 2025-11-06
[spec] spec out changes to table pagination


snopoke:
### Goal:
Give users more flexibility to view data in tables without having reload / page back and forth a lot.

Must:
* Allow loading all data into the table

Potential 
* Load new pages at the bottom of the table instead of replacing the entire table

Davemaina1:
.take

Davemaina1:
# Design: incremental table pagination (#2420)

## Problem

Every django-tables2 table using `tailwind_js_pagination.html` (members, integrations, evaluations, queues, notifications, and more) pages by fully replacing the table on every Next/Prev click. Nothing accumulates; each click throws away what was already loaded. That's the reload/back-and-forth friction the issue describes.

## Scope

django-tables2 list views only. Session/message transcript pagination is a separate, non-django-tables2 widget and out of scope here.

## Approach

**Revised per review feedback on #4424.** The initial approach below (repeated bounded "load more" requests) was reworked to a single "Load all" action instead.

`MembersTableView.paginate_by = 5` pages the table at 5 rows. A "Load all" control below it requests page 1 with `per_page` set to the total row count, returned through the same full-container swap already used for pagination and sort. django-tables2's `per_page_field` already defaults to `per_page`, so no `Table.Meta` change was needed.

This drops the earlier "unbounded fetch" concern below for team membership specifically: team rosters are bounded by team size, not the open-ended datasets (integration logs, evaluation runs) the original concern was about. That concern still applies to any other table adopting this template and should be reconsidered per table, not assumed to carry over.

A client-side grid (DataTables, AG Grid) was the other option. Ruled out: it duplicates server-side sorting/filtering already in every `tables.py`, and adds a second table framework next to django-tables2.

<details>
<summary>Original approach (superseded)</summary>

Add `tailwind_js_pagination_append.html`, extending the existing template, with a trailing "load more" row instead of prev/next. It uses `hx-select="tbody > tr"`, so the server keeps returning the same full response it already does for pagination and sort clicks, and htmx just pulls the new rows out of it. No new fragment template, no new view branch, one line changed in `get_template_names`.

This read "allow loading all data" as repeated bounded requests, not one unbounded fetch. django-tables2 pages by offset, and these are open-ended, team-scoped querysets that can reach thousands of rows, so a single unbounded fetch risked a slow query and a browser trying to render all of it at once.

</details>

## Edge cases

[...]

PR #4424, opened 2026-09-07
Add load-more pagination to the team members table

Davemaina1:
### Product Description

The team members table now shows the first 5 members with a "Load all" button below. Clicking it loads every remaining member in a single action instead of paging through them.

### Technical Description

Per review feedback, reworked from the "load more" append mechanism to a single "Load all" control:

- `MembersTableView.paginate_by = 5`. django-tables2's `per_page_field` already defaults to `per_page`, so no `Table.Meta` change was needed.
- `tailwind_js_pagination_loadall.html` replaces the prev/next pagination block with one "Load all" link, shown only while there is a next page. It requests page 1 with `per_page` set to the full row count, through the same full-container swap (`hx-target="closest div.table-container"`, `hx-swap="outerHTML"`) already used by every pagination and sort click on this table, so no new fragment template or view branch was needed.
- Removed `tailwind_js_pagination_append.html` and `single_table_append.html`, the row-append templates from the earlier version of this PR.
- Fixed `templates/teams/components/members_table.html`: the row count read `table.rows|length` (full unpaginated dataset) instead of `table.paginated_rows|length` (current page only). Harmless under the old default page size, wrong once `paginate_by=5` took effect.

Scope is unchanged: members table only. Other tables on `tailwind_js_pagination.html` are follow-up work.

### Issue Link

Progress on #2420. Covers the members table only, not the full rollout, so left without a closing keyword.

### Demo

Covered by `test_members_table_initial_load_shows_first_page_only_with_load_all_control` (first page shows exactly 5 rows plus a "Load all" control) and `test_members_table_load_all_returns_every_row_with_no_further_control` (requesting `per_page` covering every row returns all of them with the control gone). Full suite: 10 passed.

The screenshots previously here showed the earlier "load more" append behavior and no longer match this implementation, so they've been removed rather than left stale.

### Docs and Changelog

### Operator Impact


templates/table/tailwind_js_pagination_append.html
```diff
@@ -0,0 +1,39 @@
...
+                <tr><td colspan="{{ table.columns|length }}">{{ table.empty_text }}</td></tr>
+            {% endblock table.tbody.empty_text %}
+        {% endif %}
+    {% endfor %}
+    {% if table.page.has_next %}
+        <tr hx-get="{{ request.path_info }}{% querystring table.prefixed_page_field=table.page.next_page_number %}"
+            hx-target="this"
+            hx-select="tbody > tr"
+            hx-swap="outerHTML"
+            style="cursor: pointer;">
```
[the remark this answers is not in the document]
Davemaina1: Fixed in a6e966b2f: moved the trigger to a real button inside the cell, with hx-target="closest tr" so the request still swaps the row. Confirmed keyboard-focusable and functional.

apps/teams/views/members_views.py
snopoke, line 100: Let's change the pagination to 5:

```suggestion
    template_name = "teams/components/members_table.html"
    paginate_by = 5
```

(It would be nice if we show 5 initially but load all the rest when the user clicks the button e.g. 'Load All' instead of 'Load More')

Davemaina1: I have actioned this and pushed 159862b1f. paginate_by = 5 on the view, and "Load All" now requests page 1 with per_page set to the full row count, in one click, instead of paging further. It reuses the existing pagination swap, so no new template was needed.

snopoke (approved):

PR #4426, opened 2026-09-07
Write API: publish, poll and archive a chatbot

### LINKED ISSUE #4142: Write API Phase 1 — 7: Publish + version status polling

## Scope

`POST /chatbots/{id}/versions/` + `GET /chatbots/{id}/versions/status/{task_id}/` (spec §5.1, W3).

## Deliverables

- **POST `versions/`**: replicate `start_version_creation`'s ~8 lines so the API keeps the uuid (lock token = Celery task id) as the poll handle; with `make_default: true`, run the synchronous validity gate — normalized `Pipeline.validate()` and nothing more; unwired handles never block — and reject 422 with the three-bucket `errors` before dispatch; otherwise dispatch `async_create_experiment_version` → 202 + `task_id`. Snapshot-without-default stays lenient.
- **GET `versions/status/`**: `create_version_task_id` lock holds the id → `pending`; released → `AsyncResult(task_id)`: `FAILURE` → `failed` + error string, else `completed` + `latest_version.version_number`.

## Files

`apps/api/v2/views.py` (or a versions module) + urls; tests.

## Depends on

Task 1 (gate normalization), Task 4 (permissions).

## Test focus

Invalid pipeline + `make_default` → 422 with normalized errors, no task dispatched; valid publish → 202, poll transitions `pending` → `completed` with correct version number; snapshot without `make_default` skips the gate. (Accepted: eager-test mode can't assert `failed` — spec W3.)

### LINKED ISSUE #4143: Write API Phase 1 — 8: Chatbot archive with channel guard

## Scope

`DELETE /chatbots/{id}/` (spec §5.1, W9). Small, self-contained.

## Deliverables

- Non-team-global, non-deleted `ExperimentChannel` rows attached → 409 with `detail` + `channels: [{id, platform}]` (team-global API/WEB/EVALUATIONS channels never trip the guard).
- Otherwise `Experiment.archive()`; count the scheduled messages it hard-deletes and return `{archived: true, cancelled_scheduled_messages: N}`.
- No channel endpoints of any kind.

## Files

`apps/api/v2/views.py`; tests.

## Depends on

Task 4 (`delete_experiment` permission).

## Test focus

Web/API-only bot archives freely; telegram channel attached → 409 listing it; scheduled-message count reported matches deletions.

SmittieC:
### Product Description
Three new v2 endpoints: publish a chatbot version, poll that publish for its outcome, and archive a chatbot.

### Technical Description

- **`start_version_creation` now returns the task id.** #4142 asked the endpoint to replicate its lock-and-dispatch lines to keep hold of the uuid; returning it avoids a second copy of that protocol. The three existing callers used the result as a bool, so truthiness carries them.
- **Two names diverge from the spec sketch:** the publish body takes `version_description` (`description` already means the chatbot's own on create/patch, and it's what inspect reads back), and the 422 carries `pipeline_errors`, matching the façade and inspect.
- **An unattributable `task_id` is a 404 when the chatbot has no versions** — the spec left this open. Nothing records the ids of finished operations, and a completed publish always leaves a version behind. Where versions do exist an unknown id still reads `completed` against the newest, which is the ambiguity W3 accepts.
- **The 409 on a busy version lock names no task id** on purpose: a UI revert holds the same lock under a token that isn't a Celery task id, so handing it out would report a publish that never happened.
- **The 409 channel list adds `name`** to the `id`/`platform` #4143 specifies. Nothing in that list is addressable — there is no channel endpoint — so it exists for a person to find the channel in the web app.

Worth a look: `ChannelsAttached` sets `self.detail` rather than passing it to `super().__init__()`, because DRF coerces every leaf of a structured detail to `ErrorDetail` (a `str` subclass) and that rendered each channel's integer id as a string. The 422 next door goes through `super()` and is unaffected only because every leaf of the errors report is already a string.

### Issue Link
fixes #4142
fixes #4143

### Docs and Changelog

### Operator Impact

### DIFF
--- a/api-schemas/v2.yml
+++ b/api-schemas/v2.yml
@@ -756,6 +756,109 @@ paths:
         '409':
           description: The node is part of the pipeline's structure and cannot be
             deleted.
+  /api/v2/chatbots/{id}/versions/:
+    post:
+      operationId: chatbot_version_create
+      description: |-
+        Snapshot the chatbot's working (draft) version as a new, immutable version. Nothing writes to a version: read it back with the `chatbot_inspect` endpoint's `version` parameter, and every other write endpoint keeps targeting the working version.
+
+        A worker takes the snapshot, so the answer is `202` with no body. Poll the `chatbot_version_status` endpoint for the version number and the outcome.
+
+        One version operation runs at a time per chatbot, creating a version and reverting alike; a second is refused with `409`.
+      summary: Create a new version
+      parameters:
+      - in: path
+        name: id
+        schema:
+          type: string
+          format: uuid
+        description: Chatbot ID
+        required: true
+      tags:
+      - Chatbots
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/VersionCreate'
+      security:
+      - OAuth2:
+        - chatbots:write
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '202':
+          description: The version creation was accepted. Poll `chatbot_version_status`
+            for its outcome.
+        '400':
+          description: The body carries a key this endpoint does not accept.
+        '403':
+          description: 'The caller is authenticated but not authorised for this chatbot''s
+            versions: either its role lacks permission to change chatbots, or it is
+            a machine (client-credentials) token whose application is not authorised
+            for this chatbot.'
+        '404':
+          description: 'No such chatbot. Archived chatbots and existing versions are
+            both outside this endpoint''s reach: only a live working version can be
+            snapshotted.'
+        '409':
+          description: A version operation (another version creation, or a revert
+            started in the UI) is already running for this chatbot.
+        '422':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/VersionCreateRefused'
+          description: 'Either the pipeline has errors and this version would be served
+            to participants -- `make_default: true`, or a chatbot''s first version,
+            which is always the published one -- or the working version is identical
+            to the newest one, so the snapshot would record no change. Nothing is
+            snapshotted either way. Dangling handles are not errors: `unwired_handles`
+            is advisory.'
+  /api/v2/chatbots/{id}/versions/status/:
+    get:
+      operationId: chatbot_version_status
+      description: |-
+        Report where this chatbot's version history stands, which is how a version the `chatbot_version_create` endpoint accepted is followed to its end.
+
+        Poll this until `status` is terminal:
+
+        - **`pending`** — a version operation is still running. Poll again.
+        - **`completed`** — no operation is running and `version_number` names the chatbot's newest version. Read it back with the `chatbot_inspect` endpoint's `version` parameter to check what landed in it.
+
+        **`completed` is not a receipt for your request.** This reports the chatbot's own state, not one task's outcome, so it cannot say which request produced the version it names. A version creation the worker raised on releases the lock without snapshotting anything, and polling then reports `completed` on the version that was already there — so check `version_number` against the one you expected rather than reading `completed` as proof the snapshot landed.
+
+        A chatbot with no version at all has nothing to report and answers `404`.
+      summary: Version creation status
+      parameters:
+      - in: path
+        name: id
+        schema:
+          type: string
+          format: uuid
+        description: Chatbot ID
+        required: true
+      tags:
+      - Chatbots
+      security:
+      - OAuth2:
+        - chatbots:read
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/VersionStatus'
+          description: ''
+        '403':
+          description: 'The caller is authenticated but not authorised for this chatbot''s
+            versions: either its role lacks permission to change chatbots, or it is
+            a machine (client-credentials) token whose application is not authorised
+            for this chatbot.'
+        '404':
+          description: No such chatbot, or it has never created a version.
   /api/v2/me/:
     get:
       operationId: me
@@ -3683,6 +3786,67 @@ components:
       - type: array
         items:
           $ref: '#/components/schemas/UsageBucketResult'
+    VersionCreate:
+      type: object
+      description: 'The version-create body: whether the snapshot goes live, and what
+        to label it.'
+      properties:
+        make_default:
+          type: boolean
+          default: false
+          description: Whether the new version becomes the one participants are served,
+            which takes effect from the next message on. Left false, the version is
+            a checkpoint that nothing serves -- except a chatbot's first version,
+            which is always the published one and so is always held to the pipeline-validity
+            gate.
+        version_description:
+          type: string
+          nullable: true
+          default: ''
+          description: Free-text label for this version -- what changed since the
+            last one. The `chatbot_inspect` endpoint reads it back under the same
+            name.
+      additionalProperties: false
+    VersionCreateRefused:
+      type: object
+      description: 'The 422: why the chatbot''s state made the request unanswerable.'
+      properties:
+        detail:
+          type: string
+        pipeline_errors:
+          allOf:
+          - $ref: '#/components/schemas/PipelineBuildErrors'
+          description: The repair list, present only when going live was refused for
+            a pipeline that does not validate. Absent when the refusal is that there
+            is nothing to snapshot.
+      required:
+      - detail
+    VersionStatus:
+      type: object
+      description: 'A version poll''s answer: whether an operation is running, and
+        the newest version if not.'
+      properties:
+        status:
+          allOf:
+          - $ref: '#/components/schemas/VersionStatusEnum'
+          description: |-
+            `pending` while a version operation is running -- poll again. `completed` is terminal.
+
+            * `pending` - Pending
+            * `completed` - Completed
+        version_number:
+          type: integer
+          description: The chatbot's newest version. Present on `completed` only.
+      required:
+      - status
+    VersionStatusEnum:
+      enum:
+      - pending
+      - completed
+      type: string
+      description: |-
+        * `pending` - Pending
+        * `completed` - Completed
     VoiceResponseBehaviourEnum:
       enum:
       - always
--- a/api-schemas/v2.yml
+++ b/api-schemas/v2.yml
@@ -815,6 +815,61 @@ paths:
             to the newest one, so the snapshot would record no change. Nothing is
             snapshotted either way. Dangling handles are not errors: `unwired_handles`
             is advisory.'
+  /api/v2/chatbots/{id}/versions/{version_number}/:
+    patch:
+      operationId: chatbot_version_publish
+      description: |-
+        Make this version the one participants are served, without snapshotting anything.
+
+        Channels resolve the published version per message, so a live chatbot switches over from the next message on. The version that held it stops being served and is otherwise untouched -- a chatbot holds exactly one published version, so promoting one demotes the other.
+
+        This is how an already-published version is served again; the `chatbot_version_create` endpoint's `make_default` covers the other case, a snapshot that goes live as it is taken. Promoting a version this way creates nothing, so it is never refused for having no changes to record.
+      summary: Make chatbot version the published one
+      parameters:
+      - in: path
+        name: id
+        schema:
+          type: string
+          format: uuid
+        description: Chatbot ID
+        required: true
+      - in: path
+        name: version_number
+        schema:
+          type: integer
+        description: The version's number, as the `chatbot_inspect` endpoint reports
+          it.
+        required: true
+      tags:
+      - Chatbots
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/PatchedPublishVersion'
+      security:
+      - OAuth2:
+        - chatbots:write
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/VersionPublished'
+          description: ''
+        '400':
+          description: The body carries a key this endpoint does not accept, or asks
+            to un-publish a version -- a chatbot always needs a published one.
+        '403':
+          description: 'The caller is authenticated but not authorised for this chatbot''s
+            versions: either its role lacks permission to change chatbots, or it is
+            a machine (client-credentials) token whose application is not authorised
+            for this chatbot.'
+        '404':
+          description: No such chatbot, or it has no version under this number --
+            an archived chatbot and an archived version included.
   /api/v2/chatbots/{id}/versions/status/:
     get:
       operationId: chatbot_version_status
@@ -3176,6 +3231,21 @@ components:
             does not declare is dropped rather than refused, and the response reports
             the params the node ended up holding.
       additionalProperties: false
+    PatchedPublishVersion:
+      type: object
+      description: |-
+        The PATCH body: make this version the one participants are served.
+
+        Named as ``chatbot_inspect`` reports it, so an agent that read
+        ``is_published_version: false`` writes back the same key.
+      properties:
+        is_published_version:
+          type: boolean
+          description: Must be `true`. A chatbot may hold only one published version
+            and always needs one, so this promotes a version; it cannot un-publish
+            one. Create a new version that goes live, or make a different version
+            the published one, to move it off.
+      additionalProperties: false
     PipelineBuildErrors:
       type: object
       description: The normalized three-bucket errors report the publish gate rejects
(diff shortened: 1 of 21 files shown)


apps/api/v2/versions/views.py
```diff
@@ -0,0 +1,237 @@
...
+    Once the lock is free, a recorded failure is read before the versions -- a chatbot that has
+    published before still has a ``latest_version``, which would otherwise be reported as this
+    publish's.
+    """
+    if chatbot.create_version_task_id == task_id:
+        return {"status": VersionStatus.PENDING}
+    result = AsyncResult(task_id)
+    if result.state == states.FAILURE:
+        return {"status": VersionStatus.FAILED, "error": str(result.result)}
+    version = chatbot.latest_version
```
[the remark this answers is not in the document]
SmittieC: Fixed in a397851 + b0a34b5.

Verified the premise first: `create_new_version` clones the working row from the DB while the lock is held, so the version row was already inheriting `create_version_task_id` — but only by accident, and it contradicted the documented intent of `VERSION_IDENTITY_FIELDS` ("never copied between rows in a family"). It also left `version_operation_in_progress` reading `True` on a finished snapshot.

So the association is now explicit rather than incidental: a `created_by_task_id` field records the publish task on the version it produced, and `create_version_task_id` is cleared on the new row. `_version_status` matches on that and raises `NotFound` when no version records the id — stale, mistyped, or belonging to another chatbot. Migration keeps a DB-level default (`db_default=""`) so the previous release's inserts survive it.

One note: this changes the contract issue #4142 specified ("released → `AsyncResult(task_id)`: FAILURE → failed, else completed + `latest_version.version_number`"), so the endpoint description and schema are updated to match, and the author has signed off on the divergence.

*— Claude Code*

apps/api/v2/views.py
```diff
@@ -182,6 +183,53 @@ def partial_update(self, request, *args, **kwargs) -> Response:
...
+                    "(client-credentials) token whose application is not authorised for this "
+                    "chatbot."
+                )
+            ),
+            404: OpenApiResponse(description="No such chatbot, or it is archived already."),
+            409: ChannelsAttachedSerializer,
+        },
+    )
+    def destroy(self, request, *args, **kwargs) -> Response:
+        chatbot = get_working_chatbot(self.team, self.kwargs[self.lookup_url_kwarg])
```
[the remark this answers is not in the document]
SmittieC: Partly fixed in 7b5ffc7.

The count race is real and is fixed: resolution, the channel guard, the count and `archive()` now share one `transaction.atomic()` with `get_working_chatbot(..., lock=True)`. Confirmed the loser now blocks on the row and answers 404 rather than reporting schedules it did not delete — the new test in `test_chatbot_archive.py` was checked against the unlocked code first and failed there with `assert 200 == 404`, so it genuinely covers the race.

**Not done: the channel-attachment half.** Making `ExperimentChannel` creation lock the same `Experiment` row and recheck `is_archived` means changing the web app's channel form, which is a different app and surface from this API-only PR, and the race predates it. Flagged to the author rather than folded in here.

*— Claude Code*

api-schemas/v2.yml
```diff
@@ -712,6 +756,120 @@ paths:
...
+        '409':
+          description: A version operation (a publish, or a revert started in the
+            UI) is already running for this chatbot.
+        '422':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PipelineNotValid'
+          description: ''
+  /api/v2/chatbots/{id}/versions/status/{task_id}/:
```
SmittieC: Experiment has a `create_version_task_id`, so we don't need the client to pass this arg.

apps/api/v2/versions/views.py
```diff
@@ -0,0 +1,238 @@
...
+        body = PublishSerializer(data=request.data)
+        body.is_valid(raise_exception=True)
+        chatbot = get_working_chatbot(request.team, id)
+        # A machine token reaches only the chatbots its application was pinned to. After resolution
+        # because the allowlist is keyed on the chatbot, and before the gate, which reads the graph.
+        enforce_application_chatbot_write(request, chatbot)
+        make_default = body.validated_data["make_default"]
+        if make_default:
+            _check_pipeline_can_go_live(chatbot)
+        task_id = start_version_creation(
```
SmittieC: Creating a new version should not be possible when there are no diffs. First check if there are diffs

api-schemas/v2.yml
```diff
@@ -186,6 +186,50 @@ paths:
             a machine (client-credentials) token whose application is not authorised
             for this chatbot. Retrying will not help; the application''s chatbot list
             has to change.'
+    delete:
```
SmittieC, line 189: We should allow archiving any version. The request path should be DELETE api/v2/chatbots/{id}/versions/{version_number}

api-schemas/v2.yml
SmittieC: Fixed in 7646cba + 36bf914. The route is now `GET /chatbots/{id}/versions/status/` and reads the chatbot: `version_operation_in_progress` → `pending`, otherwise `latest_version.version_number` → `completed`. `created_by_task_id`, its migration and the clone bookkeeping are gone with it.

Worth recording what the simplification costs, since `create_version_task_id` is the lock and is cleared the moment the publish ends — so it can only answer "in progress", never "how it went":

- **`failed` is gone.** It was read from `AsyncResult(task_id)`, which needs the id. The enum is now `pending`/`completed` only.
- **`completed` is no longer a receipt.** A publish the worker raises on frees the lock without snapshotting, so a poll then reports `completed` on the version that was already there. The endpoint description says this outright and tells clients to check `version_number` against the one they expected; `test_a_failed_publish_on_a_chatbot_with_history_reports_the_version_already_there` pins the behaviour so it can't drift silently.
- A chatbot that has never published answers `404`, which is also what a publish that failed before creating anything now looks like.

This diverges from #4142, which specified `{task_id}` in the path and `AsyncResult(task_id)` for `FAILURE` — you may want to update the issue.

*— Claude Code*

apps/api/v2/versions/views.py
SmittieC: Fixed in 8f424e3. `_check_there_is_something_to_publish` refuses with `422` (`NothingToPublish`) when the working version matches the newest one, and it runs before the pipeline gate as you asked — an unchanged chatbot is refused for having nothing to publish without the graph also being walked to find fault with it.

Two things that decided the shape:

- The guard is `chatbot.has_versions and not chatbot.compare_with_latest()`. `compare_with_latest()` returns `False` when there is no `latest_version`, so "no baseline" and "no difference" arrive as the same value — checking the comparison alone would have refused every first publish.
- `compare_with_latest()` does see pipeline edits: `Experiment._get_version_details` appends a `pipeline` `VersionField` when `pipeline_id` is set. Since node edits are what this API is mostly used for, `test_a_pipeline_edit_since_the_last_version_may_be_published` covers it explicitly rather than trusting that.

`pipeline_errors` is now `required=False` on the 422 serializer (renamed `PublishRefused`, component `PublishRefused`), because both refusals share the status code and only the pipeline one carries the repair list. A machine client parsing 422 against the old required-`pipeline_errors` schema would have broken on this body.

*— Claude Code*

api-schemas/v2.yml
SmittieC: Added in 579245a: `DELETE /chatbots/{id}/versions/{version_number}/` soft-archives one published version and the pipeline snapshot it owns.

I kept the chatbot-level `DELETE /chatbots/{id}/` alongside it rather than moving it, and checked with you first, because the two are not the same operation: the channel guard and `cancelled_scheduled_messages` — the bulk of #4143 — read channels and schedules, which hang off the working row and never off a snapshot. So there was nothing to carry over to a version path.

Rules, mirroring the web app's `archive_experiment_version`:

- The default version is refused with `409`. There is a partial unique constraint of one default per family, so archiving it would leave the chatbot with none; the UI silently redirects here, the API says why.
- An unknown number, an already-archived version and a version of an archived chatbot are all `404` — `chatbot.versions` excludes archived rows, so a retry after an answer you never saw is safe rather than a second archive.
- Gated on `delete_experiment`, not `change_experiment` like the rest of the sub-resources: a version is an `Experiment` row and this hides one.
- Lookup, the default check and `archive()` share one `transaction.atomic()` with `lock=True` on the working chatbot, so two archives of the same version can't both read it as live.
- The working version has no number and stays unreachable here.

One non-obvious thing the tests document: `create_new_version()` makes v1 the default automatically, so a chatbot's first version is only archivable once a later one takes the default over.

*— Claude Code*

apps/api/v2/versions/views.py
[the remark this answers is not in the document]
SmittieC: Correction to my reply above — that fix has been reverted, and this finding is now moot rather than fixed.

The author has since decided the client should not pass a task id at all, so as of 7646cba + 36bf914 the endpoint is `GET /chatbots/{id}/versions/status/` and `created_by_task_id` no longer exists. There is no task id to mis-attribute, so the "stale id returns an unrelated version" path this finding described is gone.

The underlying concern does survive in a different form, and is now documented rather than fixed: `completed` reports `latest_version.version_number` and cannot say which publish produced it, so a publish that fails after a chatbot already has versions reports `completed` on the version that was already there. The endpoint description states this and tells clients to check `version_number` against the one they expected. Tracked by `test_a_failed_publish_on_a_chatbot_with_history_reports_the_version_already_there`.

*— Claude Code*

api-schemas/v2.yml
SmittieC: Superseding my reply above — the dedicated endpoint is gone, replaced by a query parameter on this one in b87b636 + 738719c.

`DELETE /chatbots/{id}/?version_number=N` archives that one version; omitted, it archives the chatbot as before. `ChatbotVersionArchiveView`, its route, its permission class and `VersionArchived` are deleted, and the version tests moved into `test_chatbot_archive.py`.

Two things fell out of the query-parameter form that the path form didn't have, both now covered by tests:

- **A blank `?version_number=` was archiving the whole chatbot.** DRF's `Field.get_value` turns an empty query value into `empty` for a non-required field, so a client interpolating an unset variable into the URL would have destroyed the chatbot and every version under it. `_VersionNumberField` keeps the blank value so it fails validation with a `400`.
- **An unknown query key is now a `400`** rather than silently falling through to the whole-chatbot archive — the same reasoning, applied to `?vesrion_number=1`. This needed splitting `RejectsUnknownKeys`: a test holds its subclasses to a bijection with the schema components closed by `additionalProperties: false`, and OpenAPI has no way to say "and no other query parameters", so query serializers use a sibling mixin (`RejectsUnknownQueryKeys`).

Because both subjects now share one response, `cancelled_scheduled_messages` and `channels` became optional (`Archived`, and `ChannelsAttached` → `ArchiveRefused`). A version archive returns `{"archived": true}` with no count: schedules hang off the working row, so nothing is cancelled, and reporting `0` would read as "this version had no schedules" rather than "archiving a version cancels none".

One inconsistency worth a look: `chatbot_inspect` selects a version with `?version=` (accepting a number or `default`), while this takes `?version_number=`. Two names for the same idea on neighbouring endpoints. Say the word and I'll align them.

*— Claude Code*

apps/api/v2/versions/views.py
```diff
@@ -0,0 +1,472 @@
...
+FORBIDDEN_DELETE = OpenApiResponse(
+    description=(
+        "The caller is authenticated but not authorised to archive this chatbot's versions: either "
+        "its role lacks permission to delete chatbots, or it is a machine (client-credentials) "
+        "token whose application is not authorised for this chatbot."
+    )
+)
+
+
+class VersionOperationInProgress(APIException):
```
SmittieC: move exceptions into an exceptions file

apps/api/v2/versions/views.py
```diff
@@ -0,0 +1,472 @@
...
+    """Publish the working version as a new immutable version."""
+
+    permission_classes = [*BASE_PERMISSION_CLASSES, ChatbotCompositionPermission, TokenHasOAuthResourceScope]
+    required_scopes = ["chatbots"]
+
+    @extend_schema(
+        operation_id="chatbot_version_create",
+        summary="Publish a Chatbot Version",
+        description=(
+            "Snapshot the chatbot's working (draft) version as a new version.\n\n"
```
SmittieC: shorten this description

apps/api/v2/versions/serializers.py
```diff
@@ -0,0 +1,106 @@
...
+from apps.api.v2.write.fields import OptionalTextField
+
+
+class PublishSerializer(RejectsUnknownKeys, serializers.Serializer):
+    """The publish body: whether the snapshot goes live, and what to label it."""
+
+    make_default = serializers.BooleanField(
+        default=False,
+        help_text=(
+            "Whether the new version becomes the one participants are served. Channels resolve the "
```
SmittieC: shorten help text

apps/api/v2/versions/views.py
```diff
@@ -0,0 +1,472 @@
...
+    """
+    latest = chatbot.latest_version
+    if latest is None:
+        return
+    if chatbot.compare_with_latest() or _wiring_differs(chatbot, latest):
+        return
+    raise NothingToPublish()
+
+
+def _wiring_differs(chatbot: Experiment, latest: Experiment) -> bool:
```
SmittieC: rather update the pipeline model's `_get_version_details` to include its data dict

apps/api/v2/versions/views.py
```diff
@@ -0,0 +1,472 @@
...
+    chatbot.versions.filter(is_default_version=True).exclude(id=version.id).update(
+        is_default_version=False, audit_action=AuditAction.AUDIT
+    )
+    if not version.is_default_version:
+        version.is_default_version = True
+        version.save(update_fields=["is_default_version"])
+
+
+def _version_status(chatbot: Experiment) -> dict:
+    """Where ``chatbot``'s version history stands: an operation in flight, or the newest version.
```
SmittieC: make docstring 1 line

apps/api/v2/versions/views.py
SmittieC: Moved in ef85cee. `apps/api/v2/versions/exceptions.py` now holds all four — `VersionOperationInProgress`, `PipelineIsNotValid`, `NothingToPublish` and `VersionIsDefault`; the views import them. Verbatim move, no behaviour change.

*— Claude Code*

apps/api/v2/versions/views.py
SmittieC: Shortened in b7285b2 — the `chatbot_version_create` description drops from ~20 lines to 3 short paragraphs (what it does, that it is async, and the one-operation-at-a-time rule).

The two `422` rules moved onto the `422` response entry rather than being dropped: it had `description: ''` in the schema before, so this is where they belonged anyway. The `make_default` first-version rule stays in that field's own help text.

*— Claude Code*

apps/api/v2/versions/serializers.py
SmittieC: Shortened in b7285b2, 7 lines to 4. Kept the two facts a client acts on — it takes effect from the next message, and a first version is always published and so always gated — and dropped the "since a chatbot holding versions but none published would serve nothing" rationale.

*— Claude Code*

apps/api/v2/versions/views.py
SmittieC: Done in b9bc1d5, and it turned out to be a real bug fix rather than just a move.

`Pipeline._get_version_details` now carries an `edges` field, and `_wiring_differs`/`_wiring` are gone from the view — `_check_there_is_something_to_publish` is just `compare_with_latest()` again. The web app's "unreleased changes" badge and the version-comparison page read the same comparison, so they were blind to a pure rewiring too; they are not any more.

One deviation from the literal suggestion, worth flagging: the field is the *normalised wiring* derived from `data`, not the raw `data` dict. Comparing the dict directly would report changes that are not changes — edge ids are client-generated (`_unused_edge_id` can hand the same wire a different id), `sourceHandle`/`targetHandle` are null on builder-drawn edges but named on seeded ones, edge order is not stable, and `data` also carries the legacy `errors` key. The view already normalised past all four; keeping that in the model preserves it. `Pipeline.wiring` is the property, and `VersionFieldDisplayFormatters.format_wiring` renders it as sorted `source.handle -> target.handle` lines so the comparison UI diffs something readable instead of a JSON blob.

Covered by 7 new tests in `TestPipelineWiringVersionDetails` (unwire / wire up / repoint are changes; a new id for the same wire, an absent handle, and edge reordering are not; plus one that the rewiring reaches `Experiment.compare_with_latest`). The guard in `test_a_rewiring_since_the_last_version_may_be_published` flipped from `compare_with_latest() is False` to `is True` — that assertion was the blind spot, so it is now the proof.

*— Claude Code*

apps/api/v2/versions/views.py
SmittieC: Done in b7285b2 — `_version_status` is one line now.

The dropped paragraphs were arguing against alternatives nobody proposed (why the lock rather than Celery's state, which the function never touches), which AGENTS.md rules out. The part that is client-facing — `pending` means "some version operation is running", not "yours" — is already in the endpoint description.

*— Claude Code*

PR #4429, opened 2026-09-07
refactor: remove the "Create from assistant" collection flow

SmittieC:
### Product Description
The "Create from assistant" button and its form are gone from the Collections page. It built an indexed collection from an OpenAI assistant's file-search files — a migration path off a feature that's already been removed.

### Technical Description
Phase-1 cleanup for #4254 that got missed: the assistants UI went in #4328, but this view survived and still listed `OpenAiAssistant` rows the user can't otherwise see.

Worth a look:
- `create_collection_from_assistant_task` is deleted, so a task queued across the deploy fails with `NotRegistered`. Only ever enqueued by a manual press of the button being removed here — happy to leave a no-op stub for one release if you'd rather.
- I also dropped the now-dead `migrate_from_assistant` entry from `DOCUMENTATION_LINKS`. Its docs page (`/how-to/assistants_migration/`) lives in the docs repo and probably wants deleting there too.

No tests removed — the flow had none.

### Issue Link
Part of #4254

### Docs and Changelog

User-facing note is just that the button is gone. Also remove `/how-to/assistants_migration/` in the docs repo.

### Operator Impact

Covered by the existing "OpenAI Assistants are fully removed" entry — nothing separate for operators to act on.


CHANGELOG.md
```diff
@@ -70,6 +70,13 @@ version section when a release is cut.
      removal may land, and links its deprecation tracking issue. -->
 
 ### Removed
+- The "Create from assistant" collection flow is gone:
```
SmittieC: remove this entry

SmittieC: Done — dropped the whole CHANGELOG entry and the commit that added it, so it is out of the diff. Also unchecked Operator Impact in the description, since that entry was the only thing it was claiming.

pxwxnvermx (approved):
Thanks 🙌🏼

PR #4444, opened 2026-09-07
Add a no-store Cache-Control policy for authenticated pages

### LINKED ISSUE #4380: Add a cache-control policy for sensitive-data pages
## Background

Raised by CodeRabbit during review of PR #4379 (session-details redesign). The session detail page and its per-session CSV export don't set `Cache-Control: no-store`. A shared or public browser could retain participant chat content and PII in its cache after the user navigates away.

Checked for precedent before responding. `@never_cache` (which sets `no-store` among other directives) is used twice in this codebase today:

- `apps/web/admin.py`, wrapping Django's own admin views. This is a framework default, not an OCS-specific decision.
- `apps/web/views.py`'s Celery task-status polling endpoint. Needed for freshness, not privacy.

Neither is about protecting sensitive data from caching. There's no existing pattern here for that specific concern. Adding `no-store` to just the two views CodeRabbit flagged would be a one-off rather than closing a real, previously-decided gap. Plenty of other existing pages already render the same kind of participant/chat data (the Participant home page, admin views, the old experiment session view) without this header either. Declined to bolt this onto the redesign PR; reasoning posted as a [reply](https://github.com/dimagi/open-chat-studio/pull/4379#discussion_r3918357360) on the CodeRabbit thread instead of a code change.

## Proposed work

- Decide the actual scope: does this need to cover every authenticated/team-scoped view that can render participant data, or just a specific subset (session/chat-history pages, exports, participant-data editors)?
- Most likely a small piece of middleware (or a decorator applied at the `URLconf` level to a defined set of view names/patterns) rather than a per-view decorator repeated across dozens of views. A per-view approach doesn't scale and is easy to forget on new pages.
- Confirm this doesn't collide with anything relying on browser caching today (any page using `@cache_control(max_age=..., private=True)`, currently only `apps/experiments/views/experiment.py`'s trend chart).
- Write a short test asserting the header is present on at least one representative page in scope, so the policy doesn't silently regress.

preston-56:
### Product Description
Every authenticated page now tells the browser not to cache its response (`Cache-Control: no-store`). Previously only two unrelated views set this header for their own reasons; every other authenticated page, including ones rendering participant chat history and PII, had no explicit caching policy at all, so a shared or public browser could retain that content after the user navigated away.

### Technical Description
New middleware `apps/web/cache_control_middleware.py`'s `SensitiveDataCacheControlMiddleware`, added to `MIDDLEWARE` in `config/settings.py` right after `AuthenticationMiddleware`. Applies `django.utils.cache.add_never_cache_headers` (the same helper `@never_cache` uses) to any response for a logged-in user, but only when the view hasn't already set its own `Cache-Control` header. That guard matters: `apps/experiments/views/experiment.py`'s `trends_data` view deliberately sets `Cache-Control: private, max-age=900` for a real browser-side cache of chart data, and this must not be layered with `no-store`.

Went with a blanket policy (any authenticated response) rather than enumerating a specific list of "sensitive" pages. Checked the codebase for precedent first: `@never_cache` was only used for Django admin (a framework default) and a task-status polling endpoint (needed for freshness, not privacy), so there was no existing convention for this. An enumerated allowlist means deciding, page by page, whether something counts as sensitive, and that's easy to forget on new pages later. A blanket policy for anything behind login sidesteps that judgment call: authenticated pages are private by definition, so there's little reason to let a browser cache any of them client-side.

Closes #4380

### Migrations

### Demo
Verified live against the running dev server: an authenticated participant page returns `Cache-Control: max-age=0, no-cache, no-store, must-revalidate, private`, while the trend chart endpoint keeps its own `Cache-Control: max-age=900, private` untouched.

### Docs and Changelog

### Operator Impact


apps/web/cache_control_middleware.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class SensitiveDataCacheControlMiddleware:
+    """Adds a no-store Cache-Control policy to every authenticated response.
+
+    Applied broadly rather than enumerating specific "sensitive" pages: any authenticated
+    response can carry participant or team data, and a per-page allowlist is the kind of
+    thing that gets forgotten on new pages. A view that has already set its own
+    Cache-Control header (e.g. the experiment trend chart's deliberate browser-side cache)
+    is left alone.
```
snopoke: nit: overly verbose comment
```suggestion
```

apps/web/cache_control_middleware.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+
+    def __init__(self, get_response):
+        if get_response is None:
+            raise ValueError("get_response must be provided.")
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        is_authenticated = request.user and not request.user.is_anonymous
```
snopoke: `request.user` is always available:

```suggestion
        is_authenticated = request.user.is_anonymous
```

apps/web/cache_control_middleware.py
preston-56: Fixed, trimmed to one line.

apps/web/cache_control_middleware.py
preston-56: Fixed, dropped the redundant `request.user and`. Kept the `not` though, the suggested line drops it too and would flip the meaning (test_leaves_anonymous_responses_untouched catches this if it's wrong).

snopoke:
@preston-56 is this ready for review? The PR is still in draft state.

preston-56:
@snopoke it's ready. I've got a threshold of 2 open reviews at a time, so I'm waiting on those to be reviewed and merged before I can open this one for review.

snopoke (approved):

PR #4445, opened 2026-09-07
Load session messages on scroll instead of pagination buttons

### LINKED ISSUE #4400: Replace message pagination with load-on-scroll
## Root cause / current state (confirmed)

From #2585: "When reviewing transcripts, we always click the 'load all messages' button. Pagination makes it hard to review." The issue itself names the fix: "Load the first page on load and then load the rest of the data if the user scrolls down (e.g. using `revealed` or `intersect once` htmx triggers."

Checked both `main` and the #4232 branch (`po/session-details-redesign-4232`, PR #4379) directly: both still include the classic page-by-page widget, `pagination_buttons.html`, "Page N of M" with first/previous/next/last buttons. #4232 changed how individual messages look (chat bubbles instead of table rows) but never touched pagination, so this is still open regardless of that redesign merging.

The exact mechanism the issue asks for already exists in this same template, used for a different purpose:

```html
hx-get="{% url 'experiments:get_image_html' ... %}"
hx-trigger="intersect once"
hx-swap="innerHTML">
```

`session_messages.html` already lazy-loads an image attachment's preview only once it scrolls into view. This ticket extends that same, already-proven pattern to the next page of messages, not a new mechanism.

## Proposed work

Replace the Next/Previous/Last pagination controls with a trigger element at the bottom of the currently-loaded messages that, on `intersect once`, fetches the next page and appends it, rather than replacing `#messages-container` wholesale the way the current page-jump links do. Keep the existing `search`/`language` querystring params flowing through each subsequent fetch, the same way `pagination_buttons.html` already does via `{% querystring %}`.

preston-56:
### Product Description
Session transcripts now load-on-scroll instead of paging through Next/Previous/First/Last buttons. The first page loads normally; scrolling to the bottom quietly loads and appends the next one, until the whole transcript is loaded.

### Technical Description
Extends the same `intersect once` htmx pattern already used for lazy-loading image attachments to the next page of messages. The message loop and its "load more" trigger now live in a new partial, `session_messages_list.html`, included on first render and returned directly for scroll-triggered fetches (`experiment_session_messages_view` checks a `next_page=1` flag to decide which). Existing `search`/`tag_filter`/`language` params keep flowing through each fetch via `{% querystring %}`, same as before.

Removed the now-unused `pagination_buttons.html` and its Next/Previous controls. "Show all messages" is untouched, it's a separate manual override.

Closes #4400

### Migrations

### Demo
Verified live on a 72-message session: first 10 load, then scrolling appends 10 at a time (10 → 20 → ... → 72) with no jank or duplicates, stopping cleanly at the true total. No pagination controls remain.

### Docs and Changelog

### Operator Impact


apps/experiments/views/experiment.py
```diff
@@ -549,7 +549,7 @@ def _add_time_gap_info(messages, gap_threshold_hours=4):
 @experiment_session_view()
 @verify_session_access_cookie
 def experiment_session_messages_view(request, team_slug: str, experiment_id: uuid.UUID, session_id: str):
-    """View for loading paginated messages with HTMX"""
+    """View for loading a session's messages with HTMX, one page at a time via scroll."""
```
preston-56: This function was already 107 lines before this PR, well over the 80-line threshold on its own. My change adds 5 lines for the fragment-vs-full-page branch, not the reason it's large. Not pulling it apart in this PR, that's a separate refactor unrelated to load-on-scroll.

apps/experiments/views/experiment.py
```diff
@@ -549,38 +549,19 @@ def _add_time_gap_info(messages, gap_threshold_hours=4):
...
     session = request.experiment_session
     experiment = request.experiment
     page = int(request.GET.get("page", 1))
     selected_tags = list(filter(None, request.GET.getlist("tag_filter")))
     language = request.GET.get("language", "")
     show_original_translation = request.GET.get("show_original_translation") == "on" and language
+    is_fragment_request = bool(request.GET.get("next_page"))
```
SmittieC: Could this not also work? (then we can get rid of `next_page`, which seems to be used only here)

```python
 is_fragment_request = bool(request.GET.get("page"))
```

apps/experiments/views/experiment.py
```diff
@@ -648,22 +633,53 @@ def experiment_session_messages_view(request, team_slug: str, experiment_id: uui
...
         "default_message": default_message,
-        "default_translation_models_by_providers": get_default_translation_models_by_provider(),
-        "llm_provider_models_dict": get_models_by_team_grouped_by_provider(request.team),
-        "all_tags": all_tags,
         "highlight_message_id": highlight_message_id,
     }
 
+    # A scroll-triggered fragment only renders messages, not the control panel, so skip the
+    # translation forms and provider/model lookups that panel needs, they're real queries.
+    if not is_fragment_request:
```
SmittieC: Have you considered using a separate (dedicated) view  + endpoint for the fragment requests?

apps/experiments/views/experiment.py
preston-56: `page` and `next_page` answer different questions. The view already checks `not request.GET.get("page")` to detect a full page load that deep-links to one page, no `next_page` involved. Using `page` alone would misread that as a fragment request and skip the control panel setup it needs.

apps/experiments/views/experiment.py
preston-56: SGTM, split into a dedicated fragment view: d65cc12fa. Shared query/filter/pagination logic moved into a helper both views call, `next_page` is gone entirely (the new URL is the signal instead), and the scroll trigger now points at it (7df6acdc3). Tests updated in 53a4317d5.

apps/experiments/views/experiment.py
```diff
@@ -546,12 +546,12 @@ def _add_time_gap_info(messages, gap_threshold_hours=4):
...
-def experiment_session_messages_view(request, team_slug: str, experiment_id: uuid.UUID, session_id: str):
-    """View for loading paginated messages with HTMX"""
-    session = request.experiment_session
-    experiment = request.experiment
+def _build_session_messages_context(request, session, experiment, *, full_page: bool) -> dict:
+    """Shared context for a session's message list, at either the full page or the
+    scroll-fragment endpoint. `full_page` controls the parts only the full page needs:
+    the missing-translations check and the control panel's tags/translation-form/provider
+    data, both real queries a scroll-triggered fragment fetch shouldn't pay for.
+    """
```
preston-56: Checked with radon. The pre-split function (experiment_session_messages_view before this PR) was already complexity 13, over the threshold. This PR's split reduces it to 12 in the shared helper, plus two complexity-1 view wrappers. Net improvement, pre-existing debt this diff didn't introduce and doesn't fully clear. Leaving as is rather than extracting further for a partial win on debt that predates this PR.

apps/experiments/views/experiment.py
```diff
@@ -612,7 +593,10 @@ def experiment_session_messages_view(request, team_slug: str, experiment_id: uui
                 output_field=CharField(),
             )
         )
-        has_missing_translations = messages_queryset.exclude(**{f"translations__{language}__isnull": False}).exists()
+        if full_page:
```
SmittieC: Since `full_page=True` is only ever `True` for `experiment_session_messages_view`, how about adding another method (something like `_build_session_message_translation_context`) and calling that from `experiment_session_messages_view`? That way, this method here can stay focussed on one thing i.e. session messages.

So all logic inside `if full_page` checks should go into the new context builder method.

preston-56: makes sense, split this out. `_build_session_messages_context` now returns the queryset alongside the context, and the full-page-only logic (missing-translations check plus the control panel's tag/translation-form/provider data) moved into `_build_full_page_message_context`, called from `experiment_session_messages_view` only. No `full_page`-gated branches left in the shared function.

SmittieC (approved):

PR #4453, opened 2026-09-08
Feat: Adds session hyperlink

### LINKED ISSUE #2143: [Evals] Ability to see more information about a session when selecting to create clones for evals dataset
When creating a dataset for clones of existing sessions, it's a bit hard to know which sessions to select, given the limited information that's available in the session selection table (see screenshot below).

things that could be helpful:
**- Ability to click a hyperlink that opens the session, so you can confirm it's the right one**

_(i think the above is the most helpful, and perhaps then you wouldn't need additional columns in the table. but other columns to consider adding, that would be helpful)_
- Include session ID 
- Ability to see tags. i know you can filter by them, which is helpful, but seeing would be useful too

brian-mecha:
### Product Description

When creating a dataset by cloning existing sessions, the session-selection table only showed the experiment, participant, and message count: not enough to confirm you'd picked the right session before selecting it. 
This adds a Session ID column (_with copy-to-clipboard_) and a Tags column so tags are visible at a glance, not just filterable. The existing `View Session` link is now the first column after the checkbox, so you can confirm a session before selecting it without needing to scroll. Also fixes the table overflowing its container on narrow screens.

### Issue Link

Resolves #2143 

### Migrations

### Demo

Screenshot 2026-09-08 at 3 02 24 PM

### Docs and Changelog

### Operator Impact


templates/evaluations/session_id_column.html
snopoke, line 1: Can this 'component' be used in other places where we present the same data? Currently I think it is just in `templates/chatbots/chatbot_session_view.html`

snopoke (dismissed):
one thought but otherwise looks good

templates/evaluations/session_id_column.html
brian-mecha: Yes it can @snopoke . I have extracted it to a new shared partial: `templates/generic/session_id_chip.html`.

- `templates/evaluations/session_id_column.html` now just includes it (external_id=record.external_id).
- `templates/chatbots/chatbot_session_view.html` - the only other place this pattern existed and now includes it too (`id_prefix="session-external-id"` to keep its DOM id distinct).

The partial takes `external_id` and an optional `id_prefix` (defaults to `"session-id"`), and always suffixes the id with the external_id itself so it's safe to reuse in a loop (`table rows`) or standalone (`a single detail page`) without callers having to think about uniqueness.

templates/evaluations/session_id_column.html
```diff
@@ -0,0 +1 @@
+{% include "generic/session_id_chip.html" with external_id=record.external_id %}
```
SmittieC: isn't `templates/evaluations/session_id_column.html` a bit redundant if it only calls one line, which callers of this file can just call themselves?

SmittieC (commented):
LGTM. One comment

templates/evaluations/session_id_column.html
brian-mecha: Nice catch @SmittieC . Fixed

snopoke (approved):

PR #4455, opened 2026-09-08
Write API: publish, poll and archive a chatbot- #4426

### LINKED ISSUE #4142: Write API Phase 1 — 7: Publish + version status polling

### LINKED ISSUE #4143: Write API Phase 1 — 8: Chatbot archive with channel guard

SmittieC:
### Product Description
Three new v2 endpoints: publish a chatbot version, poll that publish for its outcome, and archive a chatbot.

### Technical Description

- **`start_version_creation` now returns the task id.** #4142 asked the endpoint to replicate its lock-and-dispatch lines to keep hold of the uuid; returning it avoids a second copy of that protocol. The three existing callers used the result as a bool, so truthiness carries them.
- **Two names diverge from the spec sketch:** the publish body takes `version_description` (`description` already means the chatbot's own on create/patch, and it's what inspect reads back), and the 422 carries `pipeline_errors`, matching the façade and inspect.
- **The status endpoint polls the chatbot, not the task.** The spec sketched `versions/status/{task_id}/`; nothing records the ids of finished operations, so an id that is not the one currently holding the lock could only be answered against the newest version anyway — the ambiguity W3 accepts. Rather than take a handle it cannot honour, the route takes none and reports the chatbot's own state: `pending` while the lock is held, `completed` with the newest `version_number` otherwise, and `404` when there is no version to report. A client checks the `version_number` it gets back against the one it expected, which is what the schema description tells it to do.
- **The 409 on a busy version lock names no task id** on purpose: a UI revert holds the same lock under a token that isn't a Celery task id, so handing it out would report a publish that never happened.
- **The 409 channel list adds `name`** to the `id`/`platform` #4143 specifies. Nothing in that list is addressable — there is no channel endpoint — so it exists for a person to find the channel in the web app.

Worth a look: `ChannelsAttached` sets `self.detail` rather than passing it to `super().__init__()`, because DRF coerces every leaf of a structured detail to `ErrorDetail` (a `str` subclass) and that rendered each channel's integer id as a string. The 422 next door goes through `super()` and is unaffected only because every leaf of the errors report is already a string.

### Issue Link
fixes #4142
fixes #4143

### Docs and Changelog

### Operator Impact

### DIFF
--- a/api-schemas/v2.yml
+++ b/api-schemas/v2.yml
@@ -756,6 +756,109 @@ paths:
         '409':
           description: The node is part of the pipeline's structure and cannot be
             deleted.
+  /api/v2/chatbots/{id}/versions/:
+    post:
+      operationId: chatbot_version_create
+      description: |-
+        Snapshot the chatbot's working (draft) version as a new, immutable version. Nothing writes to a version: read it back with the `chatbot_inspect` endpoint's `version` parameter, and every other write endpoint keeps targeting the working version.
+
+        A worker takes the snapshot, so the answer is `202` with no body. Poll the `chatbot_version_status` endpoint for the version number and the outcome.
+
+        One version operation runs at a time per chatbot, creating a version and reverting alike; a second is refused with `409`.
+      summary: Create a new version
+      parameters:
+      - in: path
+        name: id
+        schema:
+          type: string
+          format: uuid
+        description: Chatbot ID
+        required: true
+      tags:
+      - Chatbots
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/VersionCreate'
+      security:
+      - OAuth2:
+        - chatbots:write
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '202':
+          description: The version creation was accepted. Poll `chatbot_version_status`
+            for its outcome.
+        '400':
+          description: The body carries a key this endpoint does not accept.
+        '403':
+          description: 'The caller is authenticated but not authorised for this chatbot''s
+            versions: either its role lacks permission to change chatbots, or it is
+            a machine (client-credentials) token whose application is not authorised
+            for this chatbot.'
+        '404':
+          description: 'No such chatbot. Archived chatbots and existing versions are
+            both outside this endpoint''s reach: only a live working version can be
+            snapshotted.'
+        '409':
+          description: A version operation (another version creation, or a revert
+            started in the UI) is already running for this chatbot.
+        '422':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/VersionCreateRefused'
+          description: 'Either the pipeline has errors and this version would be served
+            to participants -- `make_default: true`, or a chatbot''s first version,
+            which is always the published one -- or the working version is identical
+            to the newest one, so the snapshot would record no change. Nothing is
+            snapshotted either way. Dangling handles are not errors: `unwired_handles`
+            is advisory.'
+  /api/v2/chatbots/{id}/versions/status/:
+    get:
+      operationId: chatbot_version_status
+      description: |-
+        Report where this chatbot's version history stands, which is how a version the `chatbot_version_create` endpoint accepted is followed to its end.
+
+        Poll this until `status` is terminal:
+
+        - **`pending`** — a version operation is still running. Poll again.
+        - **`completed`** — no operation is running and `version_number` names the chatbot's newest version. Read it back with the `chatbot_inspect` endpoint's `version` parameter to check what landed in it.
+
+        **`completed` is not a receipt for your request.** This reports the chatbot's own state, not one task's outcome, so it cannot say which request produced the version it names. A version creation the worker raised on releases the lock without snapshotting anything, and polling then reports `completed` on the version that was already there — so check `version_number` against the one you expected rather than reading `completed` as proof the snapshot landed.
+
+        A chatbot with no version at all has nothing to report and answers `404`.
+      summary: Version creation status
+      parameters:
+      - in: path
+        name: id
+        schema:
+          type: string
+          format: uuid
+        description: Chatbot ID
+        required: true
+      tags:
+      - Chatbots
+      security:
+      - OAuth2:
+        - chatbots:read
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/VersionStatus'
+          description: ''
+        '403':
+          description: 'The caller is authenticated but not authorised for this chatbot''s
+            versions: either its role lacks permission to change chatbots, or it is
+            a machine (client-credentials) token whose application is not authorised
+            for this chatbot.'
+        '404':
+          description: No such chatbot, or it has never created a version.
   /api/v2/me/:
     get:
       operationId: me
@@ -3683,6 +3786,67 @@ components:
       - type: array
         items:
           $ref: '#/components/schemas/UsageBucketResult'
+    VersionCreate:
+      type: object
+      description: 'The version-create body: whether the snapshot goes live, and what
+        to label it.'
+      properties:
+        make_default:
+          type: boolean
+          default: false
+          description: Whether the new version becomes the one participants are served,
+            which takes effect from the next message on. Left false, the version is
+            a checkpoint that nothing serves -- except a chatbot's first version,
+            which is always the published one and so is always held to the pipeline-validity
+            gate.
+        version_description:
+          type: string
+          nullable: true
+          default: ''
+          description: Free-text label for this version -- what changed since the
+            last one. The `chatbot_inspect` endpoint reads it back under the same
+            name.
+      additionalProperties: false
+    VersionCreateRefused:
+      type: object
+      description: 'The 422: why the chatbot''s state made the request unanswerable.'
+      properties:
+        detail:
+          type: string
+        pipeline_errors:
+          allOf:
+          - $ref: '#/components/schemas/PipelineBuildErrors'
+          description: The repair list, present only when going live was refused for
+            a pipeline that does not validate. Absent when the refusal is that there
+            is nothing to snapshot.
+      required:
+      - detail
+    VersionStatus:
+      type: object
+      description: 'A version poll''s answer: whether an operation is running, and
+        the newest version if not.'
+      properties:
+        status:
+          allOf:
+          - $ref: '#/components/schemas/VersionStatusEnum'
+          description: |-
+            `pending` while a version operation is running -- poll again. `completed` is terminal.
+
+            * `pending` - Pending
+            * `completed` - Completed
+        version_number:
+          type: integer
+          description: The chatbot's newest version. Present on `completed` only.
+      required:
+      - status
+    VersionStatusEnum:
+      enum:
+      - pending
+      - completed
+      type: string
+      description: |-
+        * `pending` - Pending
+        * `completed` - Completed
     VoiceResponseBehaviourEnum:
       enum:
       - always
--- a/api-schemas/v2.yml
+++ b/api-schemas/v2.yml
@@ -815,6 +815,61 @@ paths:
             to the newest one, so the snapshot would record no change. Nothing is
             snapshotted either way. Dangling handles are not errors: `unwired_handles`
             is advisory.'
+  /api/v2/chatbots/{id}/versions/{version_number}/:
+    patch:
+      operationId: chatbot_version_publish
+      description: |-
+        Make this version the one participants are served, without snapshotting anything.
+
+        Channels resolve the published version per message, so a live chatbot switches over from the next message on. The version that held it stops being served and is otherwise untouched -- a chatbot holds exactly one published version, so promoting one demotes the other.
+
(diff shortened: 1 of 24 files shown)


apps/api/v2/write/archive.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class ChannelsAttachedSerializer(serializers.Serializer):
+    """The 409: why the archive was refused, and what a person has to do about it."""
+
+    detail = serializers.CharField()
+    channels = AttachedChannelSerializer(many=True)
+
+
+class ChannelsAttached(APIException):
```
SmittieC: this should move to an exceptions.py file

apps/experiments/models.py
```diff
@@ -863,6 +863,17 @@ def acquire_version_operation_lock(self, task_id: str) -> bool:
     def release_version_operation_lock(cls, experiment_id: int):
         cls.objects.get_all().filter(id=experiment_id).update(create_version_task_id="", audit_action=AuditAction.AUDIT)
 
+    def publish_goes_live(self, *, make_default: bool) -> bool:
```
SmittieC: this is weird to have as a method. move this logic inline at the call site

apps/api/v2/write/archive.py
SmittieC: Fixed in c2c9524. `ChannelsAttached` now lives in `apps/api/v2/write/exceptions.py`, alongside the versions package's equivalent.

Its three serializers moved to `write/serializers.py` in the same commit, because the exception builds its response body with `AttachedChannelSerializer` — leaving them in `archive.py` would have made `exceptions.py` and `archive.py` import each other. `archive.py` is now just `archive_chatbot` and `attached_channels`.

Two generated schema descriptions shifted with them (`AttachedChannel`, `ChannelsAttached`): "standing in the way" and "the 409" had no antecedent once the classes left the archive module, so they name archiving explicitly now. Regenerated in the same commit.

*— Claude Code*

apps/experiments/models.py
SmittieC: Fixed in d97b312. `publish_goes_live` is gone from the model; the call site now reads `if make_default or chatbot.version_number == 1:`, with the docstring's reasoning kept as a comment there.

*— Claude Code*

api-schemas/v2.yml
```diff
@@ -2982,6 +3279,21 @@ components:
...
+
+        Named as ``chatbot_inspect`` reports it, so an agent that read
+        ``is_published_version: false`` writes back the same key.
+      properties:
+        is_published_version:
+          type: boolean
+          description: Must be `true`. A chatbot may hold only one published version
+            and always needs one, so this promotes a version; it cannot un-publish
+            one. Create a new version that goes live, or make a different version
+            the published one, to move it off.
```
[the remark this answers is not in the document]
SmittieC: Not applicable — both bodies are already refused at runtime.

`is_published_version` is a `BooleanField` with no `required=False`, so `{}` is a 400, and `validate_is_published_version` refuses `false`. Both are covered by tests: `test_an_empty_body_is_refused` and `test_un_publishing_is_refused` in `test_version_publish.py`.

The schema shows the property as optional because drf-spectacular emits a `Patched*` component for every PATCH body, in which all fields are optional by convention — the component name here is `PatchedPublishVersion`. That is the generator's PATCH convention, not the serializer's contract, and changing it would mean flipping `COMPONENT_SPLIT_PATCH` for the whole schema.

*— Claude Code*

apps/api/v2/versions/views.py
```diff
@@ -0,0 +1,362 @@
...
+        dispatched = start_version_creation(
+            chatbot,
+            version_description=body.validated_data["version_description"],
+            make_default=make_default,
+        )
+        # The task id it hands back goes nowhere: the status endpoint takes none. It is read only
+        # for the `None` that means another version operation already holds the lock.
+        if dispatched is None:
+            raise VersionOperationInProgress()
+        return Response(status=status.HTTP_202_ACCEPTED)
```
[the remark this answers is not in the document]
SmittieC: You read the spec correctly — §5.1/W3 does ask for `202 + task_id`, a `{task_id}` path parameter, and `failed` from `AsyncResult`. The divergence is deliberate, and the author has confirmed keeping it.

The reason is that nothing records the ids of finished operations. Only the id currently holding `create_version_task_id` is knowable, so any other id could be answered only against the newest version — the ambiguity W3 already accepts. Taking a handle it cannot honour would make the endpoint look more precise than it is, so the route takes none and reports the chatbot's own state.

On a worker crash reporting `completed`: that is real and is why the schema description says, at length, that **`completed` is not a receipt for your request** and tells clients to check `version_number` against the one they expected. A `failed` state was considered and left out with the task-scoped handle it would need to be attributable.

The PR description bullet that implied a `task_id` path parameter was stale and has been rewritten — that inconsistency is what your finding caught.

*— Claude Code*

api-schemas/v2.yml
```diff
@@ -712,6 +756,212 @@ paths:
...
+              $ref: '#/components/schemas/VersionCreate'
+      security:
+      - OAuth2:
+        - chatbots:write
+      - apiKeyAuth: []
+      - tokenAuth: []
+      responses:
+        '202':
+          description: The version creation was accepted. Poll `chatbot_version_status`
+            for its outcome.
```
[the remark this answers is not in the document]
SmittieC: Same finding as the one on `apps/api/v2/versions/views.py` — answered there. In short: the divergence from §5.1/W3 is deliberate and the author has confirmed keeping the chatbot-scoped poll, because nothing records the ids of finished operations. No schema change, so nothing to regenerate.

*— Claude Code*

apps/api/v2/versions/tests/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+    Written straight into ``Pipeline.data`` because that is where the edges live (ADR-0049), and
+    because a test about publishing should not fail when the unwire endpoint changes.
+    """
+    pipeline = chatbot.pipeline
+    pipeline.data["edges"] = []
+    pipeline.save(update_fields=["data"])
+
+
+def status_url(chatbot) -> str:
+    return f"/api/v2/chatbots/{chatbot.public_id}/versions/status/"
```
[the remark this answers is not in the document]
SmittieC: Same finding as the one on `apps/api/v2/versions/views.py` — answered there. The chatbot-scoped `status_url` in this conftest matches the route as intended, so the fixture stays as it is.

*— Claude Code*

apps/api/v2/versions/views.py
```diff
@@ -0,0 +1,366 @@
...
+
+    required_permissions = ["experiments.delete_experiment"]
+
+
+class ChatbotVersionView(APIView):
+    """The two writes that address one published version: promote it, or archive it."""
+
+    # Set per verb by `get_permissions`; declared so a misconfiguration is a refusal, not an
+    # open door, if that override is ever removed.
+    permission_classes = [*BASE_PERMISSION_CLASSES, ChatbotVersionDeletePermission, TokenHasOAuthResourceScope]
```
snopoke, line 237: this doesn't seem necessary

apps/api/v2/views.py
```diff
@@ -182,6 +185,60 @@ def partial_update(self, request, *args, **kwargs) -> Response:
...
+        # because the reported count is read before `archive()` deletes the rows it counts. Two
+        # unlocked archives would both count the same schedules and both report deleting them,
+        # when only the first did. Under the lock the second blocks, then re-evaluates the
+        # working-chatbot predicate against the committed `is_archived` and answers 404.
+        with transaction.atomic():
+            chatbot = get_working_chatbot(self.team, self.kwargs[self.lookup_url_kwarg], lock=True)
+            # A machine token reaches only the chatbots its application was pinned to. Checked
+            # after resolution because the allowlist is keyed on the chatbot, and before anything
+            # is written.
+            enforce_application_chatbot_write(request, chatbot)
```
snopoke, line 239: @coderabbitai I'm seeing this manually added in a lot of views. Please create an issue to see if we can make this a decorator or permission class that doesn't require manual including in the view body.

snopoke (approved):

apps/api/v2/versions/views.py
SmittieC: Fixed in 63c94c2. Removed — `get_permissions` sets the write permission per verb and never consults the class attribute, and nothing else reads it (`ChatbotVersionView` is not in `API_KEY_VIEWS`). The read-only API-key gate still holds because `get_permissions` builds on `BASE_PERMISSION_CLASSES`, which `test_a_read_only_key_cannot_promote` and `test_a_read_only_key_cannot_archive` cover.

*— Claude Code*

apps/api/v2/versions/views.py
```diff
@@ -0,0 +1,366 @@
...
+        dispatched = start_version_creation(
+            chatbot,
+            version_description=body.validated_data["version_description"],
+            make_default=make_default,
+        )
+        # The task id it hands back goes nowhere: the status endpoint takes none. It is read only
+        # for the `None` that means another version operation already holds the lock.
+        if dispatched is None:
+            raise VersionOperationInProgress()
+        return Response(status=status.HTTP_202_ACCEPTED)
```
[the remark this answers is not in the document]
SmittieC: The window is real, but the suggested fix does not close it, so leaving the code as it stands.

Moving the check after `acquire_version_operation_lock` would narrow the gap by the few statements between the two calls. The gap that matters is longer: the snapshot is taken by `async_create_experiment_version` in a Celery worker, after the request has already answered `202`, and the version-operation lock guards version operations against each other — it does not block pipeline edits at all. A concurrent edit can land after the lock is taken just as easily as before it.

Closing it properly means re-validating at snapshot time, inside `create_new_version`’s transaction and under the `Pipeline` row lock that `apps/api/v2/pipeline_edit/facade.py` already takes. That is a change to the shared publish path the web app runs too, and the web app deliberately lets a person publish a pipeline whose errors they can see (`_check_pipeline_can_go_live`’s docstring). So it is a design change rather than a fix to this endpoint, and it is out of scope for this PR — which, as it stands, is strictly more validation than the UI path performs, not less.

*— Claude Code*

SmittieC:
All four addressed.

**`test_version_status.py` (test · medium) — fixed in ad12219.**
Correct: the test stood in for the task's `finally` instead of exercising it. Both cases in `TestNothingToReport` now run `async_create_experiment_version` with `create_new_version` raising, through a `run_the_worker_and_let_it_raise` helper, so the lock is freed by the task itself. Verified with a mutation: deleting `release_version_operation_lock` from the task's `finally` turns both tests red, where before they stayed green.

**`team_with_roles` duplication (maintainability · low) — fixed in 082f65a.**
Moved into `apps/api/v2/versions/tests/conftest.py` and dropped from `test_version_archive.py` and `test_version_publish.py`. The third copy, in `apps/api/v2/write/tests/test_auth.py`, is left alone: it pre-dates this PR and sits in a different package, so sharing it would mean a conftest at `apps/api/v2/` — out of scope here.

**`test_chatbot_archive.py` thread (test · low) — fixed in 0d9dc28.**
The exception is now recorded on `answer` and asserted before the status code, so a thread that dies reports what killed it rather than a `KeyError`.

**`test_version_publish.py` allowlist (test · low) — fixed in 08fbe71.**
Parametrized over listed/unlisted like its siblings, so the permitted path is covered too.

*— Claude Code*

snopoke (approved):

PR #4457, opened 2026-09-08
feat: add ended_at and participant_data to session API

### LINKED ISSUE #2871: Add full participant data and session state to session API
Update the session API to include the session state and participant data. This should match the data shown in the session transcript UI.

yusra-muhtar:
### Product Description
The session API (`GET /api/sessions/` and `GET /api/sessions/{id}/`) now returns two additional fields that were already visible in the session transcript UI but missing from the API response:

- `ended_at`: when the session ended (nullable datetime), matching the "Ended" row in the session details panel
- `participant_data`: the participant's data dict as it existed after the last message in the session, matching the "Participant Data" tab

Closes #2871 

### Technical Description
**Serializer (`apps/api/serializers.py`):**
- `ended_at` is a plain model field added to `Meta.fields` — no custom field needed.
- `participant_data` is a `SerializerMethodField` that replicates the logic of `ExperimentSession.latest_participant_data`: takes the most recent trace's participant data snapshot and applies any diff, falling back to the live `ParticipantData` row when no traces exist. The method reads from the prefetched `_prefetched_traces` attribute when available to avoid N+1 queries on the list endpoint.

**Viewset (`apps/api/views/sessions.py`):**
- Added a `Prefetch("traces", ...)` to `get_queryset()` using `to_attr="_prefetched_traces"`, ordered by `-timestamp, -id` and scoped to only the columns needed for participant data resolution. This avoids one extra query per session on list responses.

**Tests (`apps/api/tests/test_session_api.py`):**
- Updated `get_session_json()` helper to include the two new fields in the canonical expected shape.
- `test_session_includes_ended_at`: verifies the field serializes correctly when set.
- `test_session_participant_data`: parametrized across three cases — no trace + no participant data, no trace + with participant data, with trace + with participant data. Covers both list and retrieve endpoints.

### Migrations
_No migrations — serializer-only change._

### Demo
Before:
```json
{ "id": "...", "status": "active", "state": {}, ... }
After:
{ "id": "...", "status": "active", "ended_at": null, "state": {}, "participant_data": {"name": "Alice"} }
```

### Docs and Changelog

The session API now surfaces ended_at and participant_data, bringing it to parity with the session transcript UI.

### Operator Impact


kpoudosu:
## Code Review: Add ended_at and participant_data to Session API

**Status:** 🔴 **BLOCKS MERGE** – Critical data consistency bug + missing test coverage + secondary issues

---

### Critical Issues (MUST FIX)

#### 1. Working Version ID Resolution Bug
**Location:** `apps/api/serializers.py`, lines 256–261
**Severity:** HIGH | **Blocks merge**

When a session is pinned to a published experiment version (not the working version), the prefetched fallback comparison fails:

```python
if pd.experiment_id == instance.experiment_id:  # ← BUG
    return pd.data
return {}  # ← Silently returns empty
```

ParticipantData is always stored against the working version, but `instance.experiment_id` returns the version row's ID. So the comparison never matches, breaking API parity with the UI (which uses `for_experiment()`).

**Fix:** Add one line before the loop:
```python
experiment_id = instance.experiment.working_version_id or instance.experiment_id
for pd in prefetched:
    if pd.experiment_id == experiment_id:  # Now works for versions
```

**Impact:** API returns empty participant_data for versioned experiments, violating the stated goal: *"bringing it to parity with the session transcript UI."*

---

#### 2. Incomplete Test Coverage for Versioned Sessions
**Location:** `apps/api/tests/test_session_api.py`
**Severity:** HIGH | **Test gap**

The list endpoint uses the prefetched path (where the bug lives), but tests don't validate it. Add a parametrized test case:

```python
def test_session_participant_data_list_with_versioned_session():
    # Create session pinned to published version
    # Ensure ParticipantData exists on working version
    # Call GET /api/sessions/ (list)
    # Assert participant_data != {}, matches expected data
```

---

### Moderate Issues

#### 3. OpenAPI Schema Contradiction
**Location:** `api-schemas/v1.yml`
**Severity:** MEDIUM | **Documentation**

```yaml
participant_data:
  required: true
  nullable: true  # Contradictory in OpenAPI 3.0
```

A required field cannot be null. Fix by removing from `required` (since `{}` is a valid, non-null response).

---

#### 4. Out-of-Scope Test Refactoring
**Location:** PR description mentions tag test relocation
**Severity:** MEDIUM-LOW | **Code review cleanliness**

PR mixes session API feature work (ended_at, participant_data) with unrelated tag test relocation. Split tag test changes into a separate PR.

---

#### 5. Missing Docstring Coverage
**Location:** Multiple functions (serializers.py, views/sessions.py)
**Severity:** LOW | **Policy requirement**

Docstring coverage is 26.47% vs. 80% required. Add docstrings to:
- `_participant_data_fallback()` – explain version resolution logic
- - Prefetch config comment – explain N+1 avoidance
---

### What Works Well ✅

- **ended_at:** Simple, correct implementation
- - **participant_data_from_trace() refactoring:** Good DRY extraction
- - **Prefetch optimization:** Subquery pattern avoids common N+1
- - **Test structure:** Parametrized tests cover snapshots, diffs, fallbacks
- - **Backward compatible:** Purely additive changes
---

### Suggested Action

Fix the critical items (2–3 min work), re-run tests, then this is ready to merge. The fixes are straightforward:

1. One-line bug fix in serializer
2. 2. One parametrized test case for list endpoint
3. 3. Schema field cleanup
4. 4. Docstring additions
CodeRabbit already flagged item #1; please address it along with #2.

snopoke (commented):
@yusra-muhtar what are the performance implications of this change?

yusra-muhtar:
> what are the performance implications of this change?

@snopoke this change removes the main N+1 performance issue. Without the prefetches, participant_data could trigger up to two extra queries per session. We now have two fixed prefetch queries per page, with the results resolved in Python.
The queries are also bounded and use existing indexes, so the cost is predictable.

There are two trade-offs:
- participant__data_set currently loads data across experiments, even though we only need the working-version row. Filtering by experiment could be a follow-up improvement, what do you think?
- Returning participant_data on the list endpoint increases the response size, especially for participants with large data payloads.

snopoke:
> * participant__data_set currently loads data across experiments, even though we only need the working-version row. Filtering by experiment could be a follow-up improvement, what do you think?

Sessions and participant data should only ever be linked to the working version so I don't think this is an issue.

> * Returning participant_data on the list endpoint increases the response size, especially for participants with large data payloads.

I did wonder about this and I see that we only include 'messages' in the detail endpoint. I think it would make sense to do the same here.

yusra-muhtar:
> I did wonder about this and I see that we only include 'messages' in the detail endpoint. I think it would make sense to do the same here.

@snopoke Makes sense, I'll follow the same pattern as messages/usage: pop participant_data from the serializer when include_messages=False, so it only appears on the detail endpoint. Tests and schema will be updated to match.

One question on the prefetches though, the trace and participant-data prefetches currently live in get_queryset() which runs for both list and retrieve. With participant_data detail-only, they'd still execute on every list request (2 extra queries per page) even though the data isn't used. Happy to move them to the retrieve path for cleanliness, or leave them in place if you'd rather keep the diff small, they're bounded and index-backed either way.

snopoke:
> > I did wonder about this and I see that we only include 'messages' in the detail endpoint. I think it would make sense to do the same here.
> 
> @snopoke Makes sense, I'll follow the same pattern as messages/usage: pop participant_data from the serializer when include_messages=False, so it only appears on the detail endpoint. Tests and schema will be updated to match.
> 
> One question on the prefetches though, the trace and participant-data prefetches currently live in get_queryset() which runs for both list and retrieve. With participant_data detail-only, they'd still execute on every list request (2 extra queries per page) even though the data isn't used. Happy to move them to the retrieve path for cleanliness, or leave them in place if you'd rather keep the diff small, they're bounded and index-backed either way.

Let's move them to the retrieve path

snopoke (approved):

PR #4459, opened 2026-09-09
chore: delete the flag_events waffle row

SmittieC:
### Product Description
No user-visible change. Housekeeping after the events rollout: the `flag_events` row is deleted from the waffle flag table.

### Technical Description
#4458 removed the last `flag_events` gate and the `Flags.EVENTS` registry entry, but deliberately left the DB row alone: previous-release nodes keep evaluating the template gate through a rolling deploy, and deleting the row would hide Events on those nodes until they drain. This PR is that deferred deletion.

* `teams/0018_delete_flag_events` deletes the row, and the team grants cascade with it.
* It also flushes waffle's cache. The historical model has no `flush()`, so the keys come from the concrete model — otherwise the cached instance under `flag:flag_events` and the stale `flags:all` list outlive the row.
* Deleting the row rather than leaving it keeps `flag_events` out of the admin's Legacy tab, where flags land once their DB row outlives its registry entry, so self-hosted operators don't have to clear it by hand.

**Merge order matters:** this has to ship in a release *after* #4458, not alongside it. Merging both into the same release recreates the rolling-deploy problem the split exists to avoid.

### Issue Link
None.

### Migrations

Compatible with the release that contains #4458 — by then nothing reads `flag_events`, so deleting the row changes nothing for the running or the incoming release. It is not compatible with the release *before* that one, hence the merge-order note above.

### Demo
No UI change. Verified by:
* `test_migration_0018_delete_flag_events` — the row and its team grants go, nothing is left in the cache, and a second run is a no-op.
* On a dev DB: created `flag_events` with a team grant, ran `migrate teams`, ended with 0 rows, 0 grants and no cached flag.

### Docs and Changelog

Nothing user-facing here; the rollout itself is covered by #4458.

### Operator Impact

No action required — this migration is what saves operators from deleting the legacy flag by hand in the admin.


snopoke (commented):
We don't use migrations to create the flags. What about just using the `waffle_delete` management command.

PR #4460, opened 2026-09-09
Keep pipelines running when their LLM model is deprecated

### LINKED ISSUE #4430: Deprecated LLM models break pipelines instead of warning
## Problem

Marking an `LlmProviderModel` as `deprecated=True` breaks every pipeline that references it. The node fails pydantic validation, so:

1. **Runtime**: the pipeline cannot be built. `PipelineNode.pipeline_node_instance` raises, `PipelineGraph._add_nodes_to_graph` wraps it in `PipelineNodeBuildError`, and the chatbot answers with the canned configuration-error reply instead of running. A live bot stops working the moment a deploy deprecates its model.
2. **Editor**: `Pipeline.validate()` reports the node as errored, which also blocks the test-message run (`get_response_for_pipeline_test_message`).
3. **Versioning**: `has_errors(pipeline.validate())` blocks creating a new chatbot version (`apps/chatbots/views.py:489`).

Source: `apps/pipelines/nodes/mixins.py:127`

```python
if model.deprecated:
    raise PydanticCustomError(
        "deprecated_model",
        f"LLM provider model '{model.name}' is deprecated.",
        {"field": "llm_provider_id"},
    )
```

## Why this is wrong

Deprecation is documented as a *warning* stage, not a removal stage. `docs/developer_guides/managing_models.md`:

> Deprecation marks a model as unsupported while keeping it accessible. Deprecated models show a warning in the UI. If a replacement is specified, users are notified so they can migrate before the model is eventually deleted.

The notification we send says the same (`apps/ocs_notifications/notifications.py:277`):

> The model '<name>' has been deprecated and will be removed soon. N chatbots are still using it. Please update to '<replacement>' before it is removed.

Deletion is the separate stage that actually repoints or clears references (`remove_deprecated_models`, `DELETED_MODELS`). Deprecation should give teams a migration window; today it is a hard cutover with no window at all.

## Repro

```python
provider = LlmProviderFactory.create()
model = LlmProviderModelFactory.create(team=provider.team, type=provider.type, deprecated=True)
pipeline = PipelineFactory.create(team=provider.team)
nodes = [start_node(), llm_response_node(str(provider.id), str(model.id)), end_node()]
create_runnable(pipeline, nodes).invoke(PipelineState(messages=["hi"]), config={"configurable": {"repo": ORMRepository()}})
```

```
apps.pipelines.exceptions.PipelineNodeBuildError: 1 validation error for LLMResponse
  LLM provider model 'test-model-0' is deprecated. [type=deprecated_model, ...]
```

`pipeline.validate()` on the same pipeline:

```
{'node': {'llm_response-e83d': {'llm_provider_id': "LLM provider model 'test-model-0' is deprecated."}}, 'edge': [], 'pipeline': []}
```

## Proposed fix

- Drop the `deprecated_model` raise from `LlmProviderModelMixin.validate_llm_model`. A deprecated model keeps running until it is deleted.
- Surface it as a non-blocking warning on the node in the editor instead. `ErrorReport` has no warning channel; either add a fourth advisory bucket or follow the `unwired_handles` precedent in `apps/pipelines/build_state.py` — an advisory map keyed by node id that never feeds `pipeline_valid`. The node-type deprecation path already has `ui:deprecated` / `ui:deprecation_message` (`apps/pipelines/nodes/base.py:616`) as a model for the shape of the message.
- Keep the existing option-list behaviour: the builder still offers deprecated models so a node pointing at one renders its selection, and `LlmProviderModel.__str__` already labels them `"<name> (Deprecated)"`.
- The write API (`usable_models_only=True`) should keep rejecting deprecated models as *new* values — a client should not be able to newly select one — while an existing reference stays valid.

## Also worth fixing while in here

The error is attached to `llm_provider_id`, not `llm_provider_model_id`, so the editor highlights the provider field rather than the model field. Same for the `invalid_model` raise just above it (`apps/pipelines/nodes/mixins.py:120`).

zandre-eng:
### Product Description

Deprecating an LLM model used to break every chatbot using it the moment the deploy landed: bots replied with the canned configuration-error message, and teams could no longer send a test message or version the pipeline. A deprecated model now keeps working until it is actually deleted, which is what the deprecation notice already promised.

Affected nodes carry a warning triangle in the header, naming the model and what to migrate to. Additionally, the node itself will be highlighted with an amber border to indicate it has a warning. If the node has errors the red border colour will take precedence over the warning colour.

Screenshot_20260909_150508

A node that is both a deprecated *type* and on a deprecated model shows one triangle carrying both messages.

Screenshot_20260909_150558

Furthermore, the LLM Model field will now also show a deprecated message if the currently selected model is deprecated:
Screenshot_20260909_150409

### Technical Description

The check lived in pydantic validation, so it failed the node build — one raise took out the run, the editor, the test message and version creation together.

The replacement warning is deliberately not a fourth `ErrorReport` bucket: `has_errors()` is "any bucket non-empty", so anything there would re-block what this fixes. It follows the `unwired_handles` precedent — an advisory map keyed by node id that never feeds `pipeline_valid`. Same reason it is composed in by the editing endpoints rather than returned from `pipeline_build_state()`: it costs a query, and `/inspect/` already publishes the fact as `llm.deprecated`.

The map reads `params`, not the `Node.llm_provider_model` FK column — that column is a mirror, and a stale one would let the warning disagree with the model the node is built from.

### Issue Link

fixes #4430

### Demo
See attached screenshots.

### Docs and Changelog

Notes for the changelog: deprecating an LLM model no longer stops chatbots that use it. Bots keep running, and the editor warns on affected nodes naming the model and its replacement so teams can migrate before removal. Previously a deprecation broke those bots immediately and blocked test messages and new versions.

### Operator Impact

### DIFF
--- a/apps/api/v2/pipeline_edit/serializers.py
+++ b/apps/api/v2/pipeline_edit/serializers.py
@@ -5,7 +5,11 @@ from typing import Any
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 
-from apps.api.v2.inspect.serializers import OutputHandleSerializer, PipelineBuildErrorsSerializer
+from apps.api.v2.inspect.serializers import (
+    DeprecatedModelSerializer,
+    OutputHandleSerializer,
+    PipelineBuildErrorsSerializer,
+)
 from apps.api.v2.write.base import RejectsUnknownKeys
 from apps.pipelines.build_state import node_output_handles
 from apps.pipelines.models import Node
@@ -120,16 +124,6 @@ class NodeUpdateSerializer(RejectsServerAssignedKeys, RejectsUnknownKeys, serial
         required=False,
         default=dict,
         help_text="The params to change; the ones you leave out are left as they are. " + PARAM_NAMES,
-    )
-
-
-class DeprecatedModelSerializer(serializers.Serializer):
-    """The model a node still points at, and what to move it to."""
-
-    model = serializers.CharField(help_text="Name of the deprecated model the node references.")
-    replacement = serializers.CharField(
-        allow_null=True,
-        help_text="Name of the model to migrate to, or null where none is declared.",
     )
 
 
--- a/apps/api/v2/pipeline_edit/serializers.py
+++ b/apps/api/v2/pipeline_edit/serializers.py
@@ -123,12 +123,18 @@ class NodeUpdateSerializer(RejectsServerAssignedKeys, RejectsUnknownKeys, serial
     )
 
 
-class PipelineWriteSerializer(serializers.Serializer):
-    """What every façade write reports back about the pipeline it just changed.
+class DeprecatedModelSerializer(serializers.Serializer):
+    """The model a node still points at, and what to move it to."""
 
-    The same three fields the `chatbot_inspect` endpoint publishes, so one shape is parsed across
-    read and write.
-    """
+    model = serializers.CharField(help_text="Name of the deprecated model the node references.")
+    replacement = serializers.CharField(
+        allow_null=True,
+        help_text="Name of the model to migrate to, or null where none is declared.",
+    )
+
+
+class PipelineWriteSerializer(serializers.Serializer):
+    """What every façade write reports back about the pipeline it just changed."""
 
     pipeline_valid = serializers.BooleanField(
         help_text="Whether the pipeline validates cleanly: all three error buckets empty, and nothing more."
@@ -140,6 +146,14 @@ class PipelineWriteSerializer(serializers.Serializer):
             "Advisory 'what still needs wiring' map, keyed by ``node_id``: every output handle with "
             "no outgoing edge and every implicit ``input`` handle with no incoming edge. Never an "
             "error and never blocks a publish."
+        ),
+    )
+    deprecated_models = serializers.DictField(
+        child=DeprecatedModelSerializer(),
+        help_text=(
+            "Advisory map, keyed by ``node_id``, of nodes referencing a deprecated LLM model. The "
+            "model keeps working until it is removed, so this is never an error and never blocks a "
+            "publish -- but a new node may not be pointed at one."
         ),
     )
 
--- a/apps/pipelines/build_state.py
+++ b/apps/pipelines/build_state.py
@@ -1,4 +1,4 @@
-"""Build-state reporting for a pipeline: the errors report and the advisory unwired-handles map.
+"""Build-state reporting for a pipeline: the errors report and the advisory maps beside it.
 
 ``Pipeline.validate()`` returns a complete :class:`~apps.pipelines.exceptions.ErrorReport`, which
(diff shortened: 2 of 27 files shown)


apps/pipelines/build_state.py
```diff
@@ -34,6 +36,37 @@ def pipeline_build_state(pipeline: Pipeline) -> dict:
...
+    replacements = get_deprecated_models()
+    deprecated = LlmProviderModel.objects.for_team(pipeline.team_id).filter(id__in=nodes_by_model_id, deprecated=True)
+    for model in deprecated.values_list("id", "type", "name", named=True):
+        warning = {"model": model.name, "replacement": replacements.get((model.type, model.name))}
+        for flow_id in nodes_by_model_id[model.id]:
+            warnings[flow_id] = warning
+    return warnings
+
+
+def _referenced_model_id(node: Node) -> int | None:
```
SmittieC: nit: how about making this a property on the node model?

```python
node.llm_model_id
```

SmittieC (commented):
LGTM. One request: when the model is deprecated, can we make the node's outine yellow to "highlight" the node (some pipelines are quite big and the warning triangle might be missed) - similar to what we do when there's an error. When there's a warning + error, the error's colour should take precedence. It would be nice if we can show a warning message for the model field, similar to how errors are shown.

Image

zandre-eng:
> One request: when the model is deprecated, can we make the node's outine yellow to "highlight" the node (some pipelines are quite big and the warning triangle might be missed) - similar to what we do when there's an error. When there's a warning + error, the error's colour should take precedence. It would be nice if we can show a warning message for the model field, similar to how errors are shown.

@SmittieC Great suggestion! I've incorporated your suggestion so that nodes with a warning will now show an amber border. I've also added a warning message for the LLM model field itself so that users can explicitly note when the field has a deprecated model selected (see PR description for visual examples).

apps/api/v2/pipeline_edit/serializers.py
```diff
@@ -142,6 +148,14 @@ class PipelineWriteSerializer(serializers.Serializer):
             "error and never blocks a publish."
         ),
     )
+    deprecated_models = serializers.DictField(
```
SmittieC, line 145: Can we add this property to `ChatbotInspectSerializer` as well please?

SmittieC (commented):
LGTM. One more comment

snopoke (dismissed):

apps/api/v2/pipeline_edit/serializers.py
zandre-eng: Addressed in [fd2bf70](https://github.com/dimagi/open-chat-studio/pull/4460/commits/fd2bf70fa80810afa7dfb85e8c9e8d91c506a9db).

SmittieC (approved):

PR #4461, opened 2026-09-09
Fix keyboard focus on self-swapping htmx controls

### LINKED ISSUE #4425: Keyboard focus is lost after htmx outerHTML swaps that replace their own trigger
## Problem

Any control that does `hx-get` on itself with `hx-target="this"` (or an ancestor) and `hx-swap="outerHTML"` loses keyboard focus after the swap. The old DOM node is gone, so focus resets to `<body>`. A keyboard user has to tab back down from the top of the page to find where they were.

Confirmed with a minimal repro: a button doing `hx-get` on itself, `hx-target="this"`, `hx-swap="outerHTML"`. After the swap, `document.activeElement` is `<body>`.

## Scope

`hx-swap="outerHTML"` shows up in 45 templates. Several are this exact pattern, a button replacing itself or its own row: `ocs_notifications/components/read_button.html`, `mute_button.html`, `do_not_disturb_pills.html`, and the pagination/sort links in `table/tailwind_js_pagination.html` (the members table's own controls render through this same template, not a separate one).

## Background

We know these as htmx limitations:

- htmx only auto-preserves focus for inputs with a stable `id` ([hx-swap docs](https://htmx.org/attributes/hx-swap/)); there's a long-open htmx issue on this ([htmx#75](https://github.com/bigskysoftware/htmx/issues/75)).
- `hx-preserve` doesn't help here. It only matches elements by `id` in the response, and even then its own docs say focus/caret isn't reliably preserved ([hx-preserve docs](https://htmx.org/attributes/hx-preserve/)). It does nothing for a control that's fully replaced.
- Wagtail's engineering team found the same gap across real htmx sites and recommended morph-based swapping (morphdom) for controls like this instead of `outerHTML` ([Wagtail: htmx accessibility gaps](https://wagtail.org/blog/htmx-accessibility-gaps-data-and-recommendations/)).

## Fix direction

Morph-based swapping (idiomorph, via `htmx-ext-morph`) fixes this at the mechanism level: it patches the DOM in place instead of destroying and rebuilding it, so focus, scroll position, and any other live state on the node survive the swap automatically. A generic `htmx:afterSwap` handler only fixes the specific controls it's written for, and has to be re-applied by hand to every new self-replacing control going forward, or the same bug comes back.

Scope: the controls listed above, plus `table/tailwind_js_pagination.html`'s pagination/sort links, which several other tables' controls render through. Not all 45 templates need this -- most of them swap some other element, not their own trigger.

This is a new dependency (`htmx-ext-morph` isn't installed, and no template uses `hx-ext` today) and a cross-cutting change. It needs a design doc before implementation, not a patch folded into an unrelated PR.

### LINKED ISSUE #4463: Table sort headers and pagination links aren't keyboard-focusable
## Problem
The shared table template `table/tailwind_js_pagination.html` renders sort headers as `<th>` with no `tabindex`, and pagination prev/next links as `<a>` with no `href` (they carry only `hx-get`). Neither is reachable by keyboard: confirmed in a real browser that both a scripted `.focus()` call and an actual mouse click leave `document.activeElement` unchanged on these elements. A keyboard-only user cannot sort a table or turn a page without a mouse.

## Scope
Every table rendered through `table/tailwind_js_pagination.html` (and `tailwind_js_pagination_lazy.html`, which extends it) is affected -- confirmed on the team members table, the Custom Actions table, and the OAuth Applications table, all rendered on the same manage-team page.

## Fix direction
- Pagination links: add a real `href` (the same querystring the `hx-get` already builds), so they're natively focusable and still work with JS disabled.
- Sort headers: either add `tabindex="0"` plus a keydown handler for Enter/Space, or move the click handling onto a real `<button>` inside the `<th>` instead of the cell itself.

## Test plan
A template test asserting `href`/`tabindex` presence, plus a manual keyboard-only click-through (Tab to the control, activate with Enter/Space) on at least one table.

Davemaina1:
### Product Description
Clicking "Mark Read", the mute dropdown's duration links, the Do Not Disturb pill cancel button, or a table's sort/pagination controls used to drop keyboard focus to the top of the page after each click. It now stays on the control (or its container) instead, for controls whose layout doesn't change shape between clicks. Table sort headers and pagination links were also not reachable by keyboard at all, they are now.

### Technical Description
These controls each `hx-get`/`hx-post` on themselves with `hx-swap="outerHTML"`, destroying the node that held focus. Switched them to `hx-swap="morph"` via `idiomorph`'s htmx extension, which patches the DOM in place instead of replacing it. `table/tailwind_js_pagination.html` backs every table's sort headers and pagination links (members, integrations, evaluation runs, notifications); fixing the one shared template covers all of them.

`idiomorph`'s extension has to register against the htmx instance actually running the page. That's `django-htmx`'s own vendored copy (loaded via `{% htmx_script %}` with `defer`), not the `htmx.org` npm package a couple of other modules import separately for unrelated calls. It loads from its own webpack entry (`morph-extension.js`) via a deferred `<script>` tag placed right after `{% htmx_script %}`, so both scripts run in the same guaranteed document order, with no gap where a control could be clicked before the extension registers.

Confirmed by hand in a real browser: `read_button` keeps the literal same DOM node focused across the swap. `mute_button`'s dropdown-to-single-button transition changes shape entirely, so there's nothing for morph to match onto, and focus still drops to `<body>` there -- same as before; this mechanism alone doesn't fix that case.

Sort headers and pagination links (`table/tailwind_js_pagination.html`) had no `tabindex`/`href` at all -- filed separately as #4463, included here instead of its own PR: same three template blocks this PR already touches, and the contributor pipeline on this repo is at its cap.

Sort headers now wrap their text in a real `<a href>`; pagination links get a real `href` alongside `hx-get`. `href` matches `hx-get`'s own `{{ request.path_info }}` prefix, not a bare relative querystring -- these fragments are always embedded via htmx into an outer page whose URL never changes, so a bare relative href resolves against the wrong page.

A real `href` also makes a modified click (Ctrl/Cmd-click) a genuine non-htmx navigation. `MembersTableView`/`IntegrationsTableView.get_template_names()` picked the fragment template from the query string alone; both now also check `request.htmx`.

### Issue Link
Closes #4425
Closes #4463

### Demo
Manual click-through in a local dev instance: focus on the "Mark Read" button survives the swap (`document.activeElement` is the same node before and after); focus on a mute-dropdown link does not (structural change, expected). Tab to a sort header or pagination link, press Enter: htmx intercepts it (no real navigation, `window.location` unchanged) and the same link keeps focus afterward.

### Docs and Changelog

### Operator Impact


assets/javascript/site.js
```diff
@@ -6,3 +6,13 @@ import './tom-select';
...
 import 'open-chat-studio-widget';
+
+// Registers `hx-ext="morph"` against the page's real htmx instance. `{% htmx_script %}`
+// loads that instance with `defer`, so it isn't defined yet when this (non-deferred) bundle
+// runs -- wait for DOMContentLoaded, which always fires after deferred scripts finish.
+// Importing the classic build rather than "idiomorph/htmx" also matters: the latter
+// re-imports its own private copy of `htmx.org` and would register the extension on that
+// copy instead of the one actually processing the page's hx-* attributes.
+document.addEventListener('DOMContentLoaded', () => {
+  import('idiomorph/dist/idiomorph-ext.js');
```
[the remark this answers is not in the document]
Davemaina1: Fixed: registration was racing DOMContentLoaded via a dynamic import. Moved it to its own webpack entry loaded via a deferred script tag, same as htmx_script, so ordering is now guaranteed by the defer spec instead of async timing. See templates/web/base.html and assets/javascript/morph-extension.js.

package.json
```diff
@@ -76,6 +76,7 @@
     "date-fns": "^4.1.0",
     "dompurify": "^3.4.13",
     "htmx.org": "^2.0.8",
+    "idiomorph": "^0.8.0",
```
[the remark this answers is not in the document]
Davemaina1: Idiomorph is the actual htmx morph extension (not htmx-ext-morph, which doesn't exist as a package), chosen over a per-control htmx:afterSwap handler because it fixes the swap mechanism once instead of needing reapplication to every new self-replacing control.

Davemaina1:
@coderabbitai full review

Davemaina1:
@snopoke, when working on this, I found that manual browser testing (not just the automated tests) surfaced one edge case: the members table's "Load all" control now uses `hx-ext="morph"`/`hx-swap="morph"` like the other self-swapping controls here, and the content swap works correctly. But clicking it still drops keyboard focus to `<body>`.

The cause differs from what this PR otherwise fixes. Every other control (sort headers, prev/next, the read/mute/DND buttons) exists in both the pre- and post-swap DOM, so Idiomorph matches the old node to its replacement and keeps focus. "Load all" doesn't: clicking it makes `has_next` false, so the block containing the link is dropped from the response entirely. There's no equivalent node for focus to land on, so it falls back to `<body>`. This is a "control eliminates itself" problem, not a "control gets replaced" problem, and morph alone can't fix it.

I checked for an existing convention before deciding how to handle it (ADRs, this control's commit history, similar controls) and found none. The closest pattern, the "Mark all as read" button, avoids the issue structurally by targeting a separate element instead of itself.

Leaving this as a known limitation rather than folding a fix into this PR. It's a distinct problem from what's fixed here and probably deserves its own design discussion. Can open a follow-up issue if wanted.

apps/ocs_notifications/tests/test_views.py
```diff
@@ -738,3 +739,88 @@ def test_404s_for_a_team_the_user_is_not_a_member_of(self, client, team_with_use
...
+        client.force_login(user)
+        session = client.session
+        session["team"] = team_with_users.id
+        session.save()
+
+        response = client.get(reverse("ocs_notifications:notifications_table"))
+
+        assert response.status_code == 200
+        assert b'hx-ext="morph"' in response.content
+        assert b'hx-swap="morph"' in response.content
```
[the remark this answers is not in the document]
Davemaina1: Fixed. Scoped all three assertions in TestNotificationButtonsUseMorphSwap to their actual control: the read button via its toggle_notification_read hx-post, each mute-duration button via its hx-vals duration payload, and the unmute button via its unmute_notification hx-post.

templates/ocs_notifications/components/mute_button.html
```diff
@@ -25,31 +25,31 @@
...
                         hx-vals='{"duration": "1m"}'
-                        hx-swap="outerHTML"
+                        hx-swap="morph"
                         hx-target="#btn-mute-{{ record.id }}">1 month</a>
                 </li>
                 <li>
                     <a hx-post="{{ url_for_mute }}"
                         hx-vals='{"duration": "forever"}'
-                        hx-swap="outerHTML"
+                        hx-swap="morph"
```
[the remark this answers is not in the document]
Davemaina1: Fixed. Replaced the five mute-duration anchors with <button type="button">, keeping every hx-post/hx-vals/hx-swap/hx-target attribute and label unchanged. Matches the existing daisyUI dropdown-menu pattern (e.g. templates/chatbots/single_chatbot_home.html); no extra classes needed since menu styles li > button the same as li > a.

templates/table/tailwind_js_pagination.html
```diff
@@ -26,15 +31,19 @@
 {% endblock table.thead %}
 
 {% block prev-page-link-attr %}
+    href="{{ request.path_info }}{% querystring table.prefixed_page_field=table.page.previous_page_number %}"
```
SmittieC: what is this for?

apps/web/tests/test_table_pagination_templates.py
```diff
@@ -44,3 +44,37 @@ def test_lazy_pager_never_asks_for_a_count(rf):
...
+    `window.location`) produces a visibly different, wrong href, not a coincidental match."""
+    html = render("table/tailwind_js_pagination.html", rf.get("/members/table/"))
+
+    assert '<a href="/members/table/?sort=name"' in html
+    assert 'hx-get="/members/table/?sort=name"' in html
+    assert 'hx-ext="morph"' in html
+    assert 'hx-swap="morph"' in html
+
+
+def test_pagination_links_carry_a_real_href(rf):
```
SmittieC, line 63: not sure how valuable these tests are?

templates/table/tailwind_js_pagination.html
```diff
@@ -26,15 +31,19 @@
...
     hx-get="{{ request.path_info }}{% querystring table.prefixed_page_field=table.page.previous_page_number %}"
     hx-trigger="click"
     hx-target="closest div.table-container"
-    hx-swap="outerHTML"
+    hx-ext="morph"
+    hx-swap="morph"
 {% endblock prev-page-link-attr %}
 
 {% block next-page-link-attr %}
+    href="{{ request.path_info }}{% querystring table.prefixed_page_field=table.page.next_page_number %}"
```
SmittieC, line 43: what is this for?

SmittieC (commented):
LGTM. one question

apps/web/tests/test_table_pagination_templates.py
Davemaina1: These catch the #4463 bug: a link with no href isn't keyboard reachable, and a wrong relative href would resolve against the outer page's URL instead of this fragment's own endpoint, since an htmx swap never changes window.location. Neither regression would show up anywhere else in the suite.

templates/table/tailwind_js_pagination.html
Davemaina1: This is the fix implemented. The link only had hx-get before, so it wasn't in the tab order and had no keyboard activation. href makes it a real link: keyboard and no-JS users get it, and a Ctrl/Cmd-click now opens a real navigation to the right URL instead of doing nothing. hx-get still handles a normal click.

apps/teams/views/integrations_views.py
```diff
@@ -155,10 +155,15 @@ def get_context_data(self, **kwargs):
...
         table = self._table
-        if table.prefixed_page_field in self.request.GET or table.prefixed_order_by_field in self.request.GET:
-            # Pagination/sort links target `closest div.table-container` with an
-            # outerHTML swap (see table/tailwind_js_pagination.html), so the response
-            # must be exactly that container -- not the filter-pill bar and wrapper
-            # around it, or each click nests another copy of them inside the last one.
+        wants_fragment = (
+            table.prefixed_page_field in self.request.GET or table.prefixed_order_by_field in self.request.GET
+        )
+        if wants_fragment and self.request.htmx:
```
SmittieC, line 161: Is this pattern needed for every table with pagination?

Davemaina1: No, these two are the only views that override get_template_names like this. Every other table view (Trace, Collections, Datasets, etc.) has a fixed template_name, always the fragment, so there's no second branch for a non-htmx request to hit.

SmittieC (approved):

apps/teams/views/integrations_views.py
snopoke: Is there something special about these views? Could we not use the same approach as the other views?

snopoke:
@Davemaina1 is there a general principal that we should extract from this and apply to future UI work?

apps/teams/views/integrations_views.py
Davemaina1: For Integrations, yes there is.  The filter pills need the whole container back, not just the table, so copying Members' fix won't work there.

For Members, no. ChatbotSessionsTable already does this with show_record_count, it puts the count inside div.table-container so it stays correct through self-swaps, and the view only needs one template. It also fixes a stale count bug we have right now after Load all. We have not touched it yet.

Davemaina1:
> @Davemaina1 is there a general principal that we should extract from this and apply to future UI work?

@snopoke yes, here is a proposal: we can use show_record_count if a table needs a count shown, don't hand roll one. It lives inside div.table-container by design, so it stays correct through self-swaps. Members built its own outside the container, that's actually why it ended up needing two response shapes in the first place.

PR #4464, opened 2026-09-09
Merge participant and session schedule templates

### LINKED ISSUE #4442: Merge participant and session schedule templates
## Current behavior

The #4231 participant details page added `templates/participants/partials/participant_schedule_modal.html` and `participant_schedule_row.html`, which are near-duplicates of the existing `templates/experiments/components/session_schedules_table.html` (same `cancel_schedule` action, near-identical row/modal markup). The participant version additionally shows which chatbot each schedule belongs to, since it aggregates across every chatbot the participant has used, whereas the session-scoped version is already pinned to one experiment.

## Proposed work

Merge the two into one shared template/component. This was suggested as a django-tables2 table instead of the current hand-rolled `<table>` markup, since a table class is easier to reuse/extend for the two different scopes (single-experiment vs. participant-wide) than parameterizing a raw template.

## Blocked on

#4393. The participant-side templates don't exist on `main` yet.

## Context

Raised during review on #4393, agreed as a good follow-up rather than folding into that PR:
- https://github.com/dimagi/open-chat-studio/pull/4393#discussion_r3948535669
- https://github.com/dimagi/open-chat-studio/pull/4393#discussion_r3948545237

preston-56:
### Product Description
No visible change to the session or participant schedules tabs. Under the hood, both now share one table component instead of two near-duplicate hand-rolled templates, and a bug is fixed where cancelling a schedule from a session page briefly rendered the wrong number of columns.

### Technical Description
Adds `SchedulesTable` (`apps/events/tables.py`), a django-tables2 table shared by the session-scoped schedules tab and the participant-wide one. `table.exclude = ("experiment",)` drops the Chatbot column for the single-session view, and `table.empty_text` is set per view. Row markup, the Manage modal, and the single-row htmx swap template now live once under `templates/events/components/`.

`cancel_schedule` always rendered the 6-column participant row template, even when called from the session page's 5-column table, mismatching columns after cancel. The Cancel button now sends a `show_chatbot` hx-vals flag reflecting whether its own row had a Chatbot column, and the view builds a table with matching columns for the swap.

Removed `templates/participants/partials/participant_schedule_row.html` and `participant_schedule_modal.html`, the templates this replaces.

Closes #4442

### Migrations

### Demo
Verified live on participant "Osaikou K." (2 chatbots, 5 schedules): Schedules tab shows the shared table with a Chatbot column.

participant schedules tab with shared table

Also verified on a session detail page's Participant Schedules tab: same table, Chatbot column excluded, both on initial load and after cancelling a schedule (row swaps to "Cancelled" and keeps the correct 5-column layout, no mismatch).

### Docs and Changelog

### Operator Impact


preston-56:
> Nitpick comments:
> In `@apps/events/tables.py`:
> - Around line 116-124: Use Django’s lazy translation function for the table
> column headers and the relevant Meta.empty_text, including the verbose_name
> values used by the columns around render_cadence and render_status, so headers
> and translated cell labels follow the active locale consistently.
> - Around line 160-163: Update SchedulesTable.Meta row attributes to start from
> DJANGO_TABLES2_ROW_ATTRS and then override only the schedule-specific id while
> preserving the existing class behavior; locate the change through the row_attrs
> definition and SchedulesTable.Meta.


Fixed the `row-attrs` one, restored `DJANGO_TABLES2_ROW_ATTRS` to match `EventsTable` above it.
Skipping the translation one, no table in the repo translates `verbose_name/empty_text`.

templates/events/components/schedule_row.html
snopoke, line 1: Can you use the existing `templates/table/table_row.html` template for this instead of creating a new one?

snopoke (commented):
LGTM pending one comment

templates/events/components/schedule_row.html
preston-56: @snopoke Fixed.

snopoke (approved):

PR #4466, opened 2026-09-09
fix: treat voice notes with no speech as user input, not a fault

### LINKED ISSUE #4405: Voice transcription no-speech results are reported as Sentry errors and give users no useful feedback
## Problem

Sentry issue OPEN-CHAT-STUDIO-2C7 has 24 events in 90 days, and not one of them is a bug. Every event is Azure returning `NoMatch` on a WhatsApp voice note:

| Reason | Count |
|---|---|
| `NoMatchReason.InitialSilenceTimeout` | 15 |
| `NoMatchReason.NotRecognized` | 9 |

That is "the participant sent a silent or unintelligible voice note" — a user-input condition. No operator can act on it. Meanwhile the participant gets an unhelpful reply and the team gets an ERROR notification.

Two changes wanted: stop reporting it, and tell the user something useful.

## 1. Don't send it to Sentry

`SpeechService.transcribe_audio` (`apps/service_providers/speech_service.py:63`) catches everything and flattens it:

```python
except Exception as e:
    log.exception(e)
    raise UserReportableError("Unable to transcribe audio") from e
```

From there `QueryExtractionStage.process` re-raises, `MessagePipeline._run_core_stages` catches it in the generic `except Exception` (`apps/channels/pipeline.py:251`), returns it, and `process()` re-raises after terminal stages — into Celery, into Sentry.

The precedent for the fix is right above that branch: `PipelineBuildError` and `CodeNodeRunError` / `NodeUserConfigRunError` log a warning, set `ctx.early_exit_response`, record `ctx.processing_errors`, and deliberately do **not** re-raise. A no-speech result belongs in the same category.

**Constraint:** real transcription failures must still reach Sentry. The `except Exception` in `transcribe_audio` currently erases the difference between "user sent silence" and "Azure key is invalid" / quota exhausted / `MessageMediaError`. The Azure `Canceled` branch (`speech_service.py:204`), which carries `error_details`, has produced zero events — it must keep reporting. So the distinction has to be made at the raise site, not by broadening a suppression rule.

Sketch: give no-speech its own exception type (e.g. a `NoSpeechDetected` subclass of `AudioTranscriptionException`), let `transcribe_audio` pass it through rather than flattening it, and add a pipeline branch that handles it like the user-config errors.

## 2. Give the user useful feedback

`_generate_error_message` (`apps/channels/pipeline.py:257`) feeds `ChatException` text to EventBot with the prompt "...adjust the message type or contents according to the following error message: {exception}". The exception text is `Unable to transcribe audio`, so the LLM has nothing to work with and the participant learns nothing.

The reason code is already known at the raise site and is genuinely actionable if it survives:

- `InitialSilenceTimeout` — no speech detected at all. "I couldn't hear anything in that voice note, could you record it again?"
- `NotRecognized` — audio was present but not matched. "I couldn't make out what was said."

So preserve the reason through to the user-facing message instead of discarding it.

## 3. Team notification

`QueryExtractionStage.process` calls `audio_transcription_failure_notification` on every failure, creating a `LevelChoices.ERROR` team notification titled "Audio Transcription Failed". A participant sending a silent voice note should not raise an error notification for the team. This should be gated the same way — or at least dropped to a lower level for the no-speech case.

## Notes

- 0 users impacted / 23-24 occurrences, `handled: no`, level `error` — this issue is pure noise in Sentry today.
- Related: the recognition language is hardcoded `en-US` (`speech_service.py:185`) while the affected participants are on Indian numbers (`whatsapp:IN.*`). Non-English speech surfaces as exactly `NotRecognized`, so some of those 9 events may be a language-config problem rather than unintelligible audio. Worth a separate issue if confirmed.

SmittieC:
### Product Description

A participant who sends a silent or unintelligible voice note now gets a reply telling them what to do differently — "I couldn't hear anything in that voice note, could you record it again?" instead of the generic "something went wrong". The error bot generates it, so it arrives in the participant's own language.

The turn is also recorded properly. Previously the voice note vanished and only the bot's reply was saved, so the session history showed an answer to nothing and the trace had no message to navigate from.

For the team this stops being an incident: no "Audio Transcription Failed" notification, no Sentry event, no failed Celery task, and it no longer counts against the chatbot's error rate. Genuine transcription failures — an invalid key, exhausted quota, a cancelled recognition — behave exactly as before.

### Technical Description

`SpeechService.transcribe_audio` flattened everything into `UserReportableError`, erasing the difference between "the participant sent silence" and "the Azure key is invalid". Both then hit the pipeline's generic `except Exception`, which re-raises after terminal stages, so both reached Sentry.

* **`NoSpeechDetected(AudioTranscriptionException)`** carries a `NoSpeechReason` — `SILENCE` or `NOT_UNDERSTOOD`. `speech_service.py` re-raises it instead of flattening it. Azure's `NoMatch` maps to a reason, and the base class raises `SILENCE` for a blank transcript, which is how OpenAI, ElevenLabs and Voice Engine report silence.
* **`MessageProcessingPipeline`** replies from `NO_SPEECH_PROMPTS[reason]` via the error bot and does not re-raise, matching the existing `PipelineBuildError` / `CodeNodeRunError` precedent. `QueryExtractionStage` skips the failure notification.
* **`NoSpeechDetected` joins `_CONTROL_FLOW_SIGNALS`** so the stage's span closes cleanly. Without this the trace was still written as `TraceStatus.ERROR` and counted on the operator error-rate charts.
* **Recording the turn.** `QueryExtractionStage` sits immediately before `ChatMessageCreationStage`, so raising there skipped the only stage that creates the human message, attaches the voice note and links the trace input. It now defers instead: it sets an empty `user_query`, records the reason on the context, and `NoSpeechGuardStage` raises once the turn exists. The reason needs its own context field — an attachment-only message with no caption produces the same empty query.
* **Linking the AI message.** `PersistenceStage` never called `set_output_message_id` for early-exit replies. That affected every early-exit path, so consent-flow, channel-disabled, unsupported-type and pipeline-build-error replies were all unreachable from their trace. Fixed where the message is created, so all of them benefit.

Two deviations from the issue's sketch, both pinned by tests:

1. **No `ctx.processing_errors` entry.** Its only consumer surfaces recovered errors on the span, and this span is deliberately not marked as errored. The pipeline tier owns the "not a fault" decision.
2. **Azure `NoMatchReason` bucketing is inverted from the obvious reading.** `InitialBabbleTimeout` means noise with no speech, so it maps to `SILENCE`; `NotRecognized` is the only member meaning audio was heard. A test asserts the covered set equals `set(speechsdk.NoMatchReason)`, so an SDK bump that adds a reason fails here rather than silently taking a default.

Not addressed here: **#4467** (unsupported message types have the same missing-human-message shape — the deferral fix does not transfer), and the hardcoded `en-US` recognition language, which makes non-English speech surface as exactly `NotRecognized`. Those participants get a better reply, but their speech still is not transcribable.

Tests: `ruff` and `ty` clean; 3137 passing across `apps/channels/`, `apps/service_providers/`, `apps/chat/`, `apps/experiments/` and `apps/pipelines/`. Both turn-recording defects were pinned with failing tests first.

### Issue Link

Closes #4405

### Migrations

No migrations in this PR.

### Demo

Covered by `apps/channels/tests/channels/test_voice_transcription_no_speech.py`, which drives a real voice message through `ChannelBase.new_user_message` against the DB and asserts the participant reply, the error-bot prompt for each reason, the absence of the team notification, and — against the real tracing service rather than a mock — that the human message, its voice attachment and `Trace.input_message` / `Trace.output_message` are all written.

### Docs and Changelog

### Operator Impact

Self-hosted operators will see fewer "Audio Transcription Failed" notifications and a lower chatbot error rate, because silent voice notes stop being counted. They now also create a human message, so the `NEW_HUMAN_MESSAGE` static trigger fires for them where it previously did not — worth knowing if a deployment has triggers wired to it. No action required.

### DIFF
--- a/apps/channels/pipeline.py
+++ b/apps/channels/pipeline.py
@@ -8,7 +8,7 @@ from django.utils import timezone
 
 from apps.channels.exceptions import EarlyAbort, EarlyExitResponse
 from apps.chat.bots import EventBot
-from apps.chat.exceptions import ChatException
+from apps.chat.exceptions import ChatException, NoSpeechDetected, NoSpeechReason
 from apps.pipelines.exceptions import (
     CodeNodeRunError,
     NodeUserConfigRunError,
@@ -156,13 +156,30 @@ class MessageProcessingPipeline:
        DEFAULT_ERROR_RESPONSE_TEXT, runs terminal stages, and is logged as
        a warning WITHOUT being re-raised (which would report it to Sentry
        and fail the task with no useful retry).
-    4. Unexpected Exception -- catch-all generates an error message
+    4. NoSpeechDetected -- the transcriber found no speech in the
+       participant's voice note. Replies with a message generated from
+       NO_SPEECH_PROMPTS via EventBot, falls back to
+       DEFAULT_ERROR_RESPONSE_TEXT, runs terminal stages, and is NOT
+       re-raised: it is user input, not a fault.
+    5. Unexpected Exception -- catch-all generates an error message
        via EventBot (preserving ChatException distinction), falls back
        to DEFAULT_ERROR_RESPONSE_TEXT, runs terminal stages, then
        RE-RAISES so the caller knows processing failed.
     """
 
     DEFAULT_ERROR_RESPONSE_TEXT = "Sorry, something went wrong while processing your message. Please try again later"
+
+    # Fed to the error bot, which turns them into advice in the participant's language.
+    NO_SPEECH_PROMPTS = {
+        NoSpeechReason.SILENCE: (
+            "Tell the user that no speech could be heard in the voice message they sent"
+            " and that they should try recording it again."
+        ),
+        NoSpeechReason.NOT_UNDERSTOOD: (
+            "Tell the user that the speech in the voice message they sent could not be made out"
+            " and that they should try recording it again, speaking clearly."
+        ),
+    }
 
     def __init__(
         self,
@@ -189,10 +206,13 @@ class MessageProcessingPipeline:
            error text, log a warning, and set ctx.early_exit_response --
            but do NOT re-raise (misconfiguration is not a bug worth
            reporting).
-        5. If any raises an unexpected exception, generate an error message
+        5. If any raises NoSpeechDetected, reply with a message generated
+           from the reason and set ctx.early_exit_response -- but do NOT
+           re-raise (silence is user input, not a bug).
+        6. If any raises an unexpected exception, generate an error message
            and set ctx.early_exit_response.
-        6. Run terminal stages unconditionally (they always fire).
-        7. If there was an unexpected exception, re-raise it after terminal
+        7. Run terminal stages unconditionally (they always fire).
+        8. If there was an unexpected exception, re-raise it after terminal
            stages complete.
         """
         try:
@@ -248,6 +268,9 @@ class MessageProcessingPipeline:
             logger.warning("Node user config error: %s", e)
             ctx.early_exit_response = self.DEFAULT_ERROR_RESPONSE_TEXT
             ctx.processing_errors.append(str(e))
+        except NoSpeechDetected as e:
+            logger.info("No speech detected in voice message: %s", e.reason)
+            ctx.early_exit_response = self._user_message(ctx, self.NO_SPEECH_PROMPTS[e.reason], e)
         except Exception as e:
             ctx.early_exit_response = self._generate_error_message(ctx, e)
             ctx.processing_errors.append(str(e))
@@ -264,7 +287,6 @@ class MessageProcessingPipeline:
         Maps to the old _inform_user_of_error() but WITHOUT sending --
         sending is ResponseSendingStage's responsibility.
         """
-        trace_info = TraceInfo(name="error", metadata={"error": str(exception)})
         prompt = (
             "Tell the user that something went wrong while processing their message"
             " and that they should try again later."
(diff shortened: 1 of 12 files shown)


apps/channels/stages/base.py
```diff
@@ -8,15 +8,17 @@
...
     from apps.channels.pipeline import MessageProcessingContext
 
 # Control-flow signals are not failures: they steer the pipeline (early exit,
-# silent abort, generation cancellation) rather than indicate something broke.
-# They must not mark a stage's span as errored.
-_CONTROL_FLOW_SIGNALS = (EarlyExitResponse, EarlyAbort, GenerationCancelled)
+# silent abort, generation cancellation, a voice note with no speech in it)
+# rather than indicate something broke. They must not mark a stage's span as
+# errored, which would also count them on the operator error-rate charts.
+_CONTROL_FLOW_SIGNALS = (EarlyExitResponse, EarlyAbort, GenerationCancelled, NoSpeechDetected)
```
SmittieC, line 21: `NoSpeechDetected` was added here to prevent this exception from marking the trace as an error

apps/channels/pipeline.py
snopoke, line 314: Can we customize the default too since a specific default message is better than a blanket one.

snopoke (approved):
One comment but otherwise LGTM

apps/channels/pipeline.py
SmittieC: Do you want the user to customize this on the bot level?

zandre-eng (approved):

apps/channels/pipeline.py
snopoke: No, I mean just using a static default based on the error type

Issue #4399, opened 2026-09-04
Add a scroll-progress bar to session messages


preston-56:
## What's wanted

From #2585: a floating progress bar at the top of the screen showing scroll progress through the message transcript, so a reviewer can gauge how far through a long session they are without counting pages. Nothing like this exists in the codebase yet, no bug to root-cause.

## Approach

Small, self-contained JS component: a fixed-position bar (`position: fixed; top: 0`) whose width tracks `scrollTop / (scrollHeight - clientHeight)` on the messages container, updated on scroll via `requestAnimationFrame` rather than a raw scroll listener. No backend change.

Should land after #4400 (load-on-scroll pagination), since a progress bar only makes sense once the whole transcript is loadable/scrollable in one place rather than split across pages.

PR #4469, opened 2026-09-09
Show compression checkpoints in the session transcript

### LINKED ISSUE #4012: Expose history summary/compression checkpoints in the session transcript view
## Problem

When a session uses history compression (Summarize, Truncate Tokens, or Max History Length), the transcript view gives no indication that compression occurred. Operators and developers have no way to:

- See _where_ in a conversation history was compressed
- Read the actual LLM-generated summary that the bot is using as its "memory" of earlier messages
- Distinguish between a message that was kept verbatim and one that has been summarised away

This makes it hard to debug bot behaviour — particularly cases where the bot seems to have "forgotten" earlier context, or where the summary may have captured the wrong things.

---

## Background

OCS has two chat history models, both with summary support:

**`ChatMessage`** (simple/non-pipeline bots):
- `summary` — `TextField` storing the LLM-generated summary of the conversation up to that point
- `compression_marker` — stored in `metadata` under `ChatMessageMetadataKeys.COMPRESSION_MARKER`; value is one of `PipelineChatHistoryModes` (`summarize`, `truncate_tokens`, `max_history_length`)

**`PipelineChatMessages`** (pipeline bots):
- `summary` — `TextField` on the paired message record
- `compression_marker` — `CharField` with `PipelineChatHistoryModes` choices

The three compression modes behave differently:

| Mode | Summary text? | What happens to older messages |
|---|---|---|
| `SUMMARIZE` | ✅ LLM-generated summary text | Replaced by the summary in context sent to model |
| `TRUNCATE_TOKENS` | ❌ (empty marker) | Dropped when token limit exceeded |
| `MAX_HISTORY_LENGTH` | ❌ | Oldest messages dropped when count exceeded |

`ChatMessage.make_summary_message()` / `message_iterator(with_summaries=True)` already know how to emit a synthetic system message for the summary — they just aren't used by the transcript view.

---

### 1. Visual compression checkpoint indicator

In the session message table (`templates/experiments/components/session_messages.html`), when iterating messages, detect if a message has a `compression_marker` and render a separator row, similar to the existing `time_gap_text` separators. This should appear _before_ the message that carries the checkpoint.

For **`SUMMARIZE`**, the separator should be expandable/collapsible to show the summary text — it's the actual content the bot is reasoning with. Something like:

```
┌─────────────────────────────────────────────────────────────────┐
│  📝  History summarised here  ▾                                 │
│  ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─  │
│  The user described their symptoms as fatigue and headache.     │
│  The assistant asked about duration and suggested hydration...  │
└─────────────────────────────────────────────────────────────────┘
```

For **`TRUNCATE_TOKENS`** / **`MAX_HISTORY_LENGTH`**, a simpler non-expandable indicator is enough (no summary text exists):

```
━━━━━━━━━━  History truncated here  ━━━━━━━━━━
```

### 2. Pass summary data to the template

The `experiment_session_messages_view` (or wherever messages are fetched for the transcript) should include the `summary` and `compression_marker` on each message object. This may require:

- For `ChatMessage`: querying the `summary` and `metadata` fields (already on the model)
- Passing them through to the template context — currently the template only gets `message.content`, `message.is_ai_message`, timestamps etc.

### 3. Pipeline history (`PipelineChatMessages`) — future scope

The transcript view currently shows `ChatMessage` records. `PipelineChatMessages` are stored separately. Surfacing pipeline history compression is a related but potentially separate piece of work — at minimum the issue should note it as out of scope for this ticket.

---

## Acceptance criteria

- [ ] A compression checkpoint is visually indicated in the transcript view at the correct position in message order
- [ ] For `SUMMARIZE` mode: the summary text is visible (collapsed by default, expandable inline)
- [ ] For `TRUNCATE_TOKENS` / `MAX_HISTORY_LENGTH` mode: a simple "truncated here" separator is shown (no text to expand)
- [ ] The indicator is clearly styled so it's not confused with a regular message (e.g. full-width, muted, distinct background)
- [ ] No regressions for sessions with no compression

preston-56:
### Product Description
The session transcript now shows where a conversation's history was compressed. A summarize checkpoint shows a collapsible "History summarized here" row with the actual LLM-generated summary text; a truncate/max-length checkpoint shows a plain "History truncated here" separator, since there's no summary text for those modes.

### Technical Description
`ChatMessage.summary` and `.compression_marker` were already fetched on every message the transcript view queries, no new query or migration needed. The template checks `message.summary` first (a real persisted summarize checkpoint only ever sets `summary`, never `compression_marker`, since that metadata key is reserved for the synthetic, unsaved message `ChatMessage.save()` refuses to persist), falling back to `message.compression_marker` for the truncate/max-length case. The summarize row is a DaisyUI `collapse collapse-arrow`, matching the pattern already used for version-diff groups.

Closes #4012

### Migrations

### Demo
Verified live with a 6-message session carrying both checkpoint types: the summarize checkpoint appears before message 3 (collapsed by default, expands to show the summary text), the truncate checkpoint appears before message 5 as a plain separator. A session with no compression shows neither.

collapsed, both checkpoint types visible

expanded, summary text shown

### Docs and Changelog

### Operator Impact


apps/chat/tests/test_participant_data_diff.py
```diff
@@ -108,3 +109,103 @@ def test_selecting_multiple_tags_keeps_all_of_them(self, client, experiment):
...
+        assert "History truncated here" not in content
+
+    @pytest.mark.parametrize(
+        "mode",
+        [
+            pytest.param(PipelineChatHistoryModes.TRUNCATE_TOKENS, id="truncate_tokens"),
+            pytest.param(PipelineChatHistoryModes.MAX_HISTORY_LENGTH, id="max_history_length"),
+        ],
+    )
+    def test_truncation_checkpoint_shows_a_plain_separator(self, client, experiment, mode):
```
SmittieC: in my opinion, this isn't a useful test

apps/chat/tests/test_participant_data_diff.py
```diff
@@ -108,3 +109,103 @@ def test_selecting_multiple_tags_keeps_all_of_them(self, client, experiment):
...
+
+        client.force_login(experiment.owner)
+        response = client.get(f"{self._url(experiment, session)}?show_all=on")
+
+        assert response.status_code == 200
+        content = response.content.decode()
+        assert "History truncated here" in content
+        assert "History summarized here" not in content
+
+    def test_no_checkpoint_indicator_without_compression(self, client, experiment):
```
SmittieC: also not super useful

apps/chat/tests/test_participant_data_diff.py
```diff
@@ -108,3 +109,103 @@ def test_selecting_multiple_tags_keeps_all_of_them(self, client, experiment):
...
+
+        client.force_login(experiment.owner)
+        response = client.get(f"{self._url(experiment, session)}?show_all=on")
+
+        assert response.status_code == 200
+        content = response.content.decode()
+        assert "History summarized here" not in content
+        assert "History truncated here" not in content
+
+    def test_checkpoint_shows_on_the_default_paginated_response(self, client, experiment):
```
SmittieC: also not very useful

SmittieC (commented):
LGTM, pending the conflict resolution

apps/chat/tests/test_participant_data_diff.py
preston-56: @SmittieC Fair, trimmed it. Dropped the parametrize over both truncation modes, they render identically since the template only checks `compression_marker` is set, not which mode. Kept one representative case.

apps/chat/tests/test_participant_data_diff.py
preston-56: @SmittieC Removed, agreed it's trivial.

apps/chat/tests/test_participant_data_diff.py
preston-56: @SmittieC Removed, pagination behavior is already covered elsewhere and isn't specific to this feature.

SmittieC:
@preston-56 there's another conflict

preston-56:
@SmittieC Fixed, was from #4399 merging into main and touching the same lines.

SmittieC (approved):

PR #4470, opened 2026-09-10
refactor: source model data from LiteLLM only and fix the auto-update-models workflow

SmittieC:
### Product Description

The nightly **Auto Update LLM Models** workflow has been failing with `HTTP Error 429` since 2026-09-09. Investigating that turned up an older problem: its price diff has never matched a single model, because it reads pricing fields llm-stats.com does not return.

This drops llm-stats entirely and sources everything from LiteLLM's public price table. A run goes from ~56 credentialed, rate-limited requests to one anonymous fetch, and the price diff starts working.

### Technical Description

**One upstream.** `model_prices_and_context_window.json` carries pricing, token limits, serving provider, model kind and deprecation dates for ~3,900 models in one file. llm-stats gave us none of the last three — `context_window` was `null` for every model — and metered the first. `LLM_STATS_BEARER_TOKEN` is no longer referenced.

**One comparison, six signals.** `scripts/reconcile_models.py` becomes `scripts/auto_sync_models`, a layered package: our catalogue and the price table are read into the same shape, so every signal is set algebra over the two — `added`, `removed`, `deprecated` (upstream end-of-life passed), `repriced`, `backfilled`, `unpriced`. `removed` and `deprecated` are new.

**Provider mapping comes from the data.** `ORG_TO_OCS_PROVIDERS` guessed that any OpenAI model is on Azure. LiteLLM tags each entry with the provider that serves it, so a model is offered for azure only when `azure/<model>` exists.

**A ledger, so nothing is offered twice or stranded.** `model_ledger.json` records every candidate. Claude either registers it — the entry is deleted, `default_models.py` is then the record — or rejects it with a reason, and only a rejection removes it from later runs. A failed run strands nothing.

**Two guards before anything destructive.** The script exits non-zero if the table translates to under 300 models, or would retire more than 25% of the catalogue, so a LiteLLM schema change cannot produce a mass-deletion PR.

**One job, one PR.** The catalogue and pricing halves land together under a single migration, gated on no `auto-models/*` PR already being open and on the script finding work.

### Issue Link

No issue. Failing run: https://github.com/dimagi/open-chat-studio/actions/runs/34436833898

### Migrations

None in this PR — the workflow's own runs write them.

### Demo

Dry run against the live price table and the real repo seed:

```
558 translated model(s), 315 chat-capable and current
226 added, 8 removed, 0 newly deprecated, 16 repriced, 3 to backfill, 6 uncostable
```

`110 passed`; ruff, ty and actionlint clean.

### Docs and Changelog

Internal CI tooling, not user-facing.

### Operator Impact

Affects only the `dimagi/open-chat-studio` scheduled workflow.

### Review notes

- **The first run opens a bigger PR than the steady state**: 16 rate corrections the diff should have been catching all along, plus 226 candidates to work through the ledger. Worth reading rather than rubber-stamping.
- **`azure gpt-4o-mini` is the one seeded rate to eyeball.** Every other Azure rate mirrors its OpenAI counterpart exactly; this one sits at 1.1× on input and output but 1.0× on cached input. That reads like a data-entry artifact, but if it is a deliberate markup this PR overwrites it.
- **`LLM_STATS_BEARER_TOKEN` can be deleted** from repo settings once this merges.


scripts/reconcile_models.py
```diff
@@ -256,103 +319,169 @@ def _rates_from_raw(
...
+    """Chat-capable models from providers OCS supports, keyed by model name.
+
+    One model can be served by several providers (``gpt-4o`` on both openai and
+    azure), so entries are merged: which providers offer it comes from the data
+    rather than from a hardcoded org mapping.
+    """
+    today = today or datetime.datetime.now(datetime.UTC).date()
+    merged: dict[str, dict] = {}
+    for key, entry in litellm_data.items():
+        if not isinstance(entry, dict) or entry.get("mode") not in CHAT_MODES or not _emits_text(entry):
```
SmittieC: Fixed in 38e4178a2. The three-clause guard is gone: the `isinstance` / `mode` / output-modality checks moved into a `_serves_chat` predicate (which absorbs the old single-use `_emits_text`), so the loop guard is now `if not _serves_chat(entry) or _is_deprecated(entry, today)`.

*— Claude Code*

scripts/reconcile_models.py
```diff
@@ -256,116 +410,253 @@ def _rates_from_raw(
...
+        ocs_provider, name = mapped
+        record = merged.setdefault(name, {"id": name, "providers": [], "keys": {}, "deprecation_date": None})
+        if ocs_provider not in record["providers"]:
+            record["providers"].append(ocs_provider)
+        if _prefers_key(key, record["keys"].get(ocs_provider), ocs_provider):
+            record["keys"][ocs_provider] = key
+        record["deprecation_date"] = record["deprecation_date"] or entry.get("deprecation_date")
+    for record in merged.values():
+        record["providers"].sort()
+    return merged
```
SmittieC, line 399: Fixed in 38e4178a2. `eligible_models` is down to a filter and a fold: eligibility moved to `_serves_chat` and the per-provider merge to `_merge_provider_key`.

*— Claude Code*

scripts/reconcile_models.py
```diff
@@ -909,29 +1223,44 @@ def _build_arg_parser() -> argparse.ArgumentParser:
...
+    print(f"  -> {len(data)} LiteLLM entries")
+    return data
 
 
 def _commit_price_changes(
     results: _ReconcileResults,
     repo_root: Path,
     output_path: Path,
     today: datetime.date,
+    dry_run: bool = False,
```
SmittieC, line 1085: Fixed in 38e4178a2. The backfill merge — the nested loop over `results.backfilled` and its `rules_by_kind` bookkeeping — moved to `_merge_backfilled`, which no-ops on an empty list, so the caller lost a branch too.

*— Claude Code*

scripts/reconcile_models.py
```diff
@@ -1,31 +1,45 @@
...
 """
-Reconcile OCS's in-repo model catalogue and pricing seed against upstream
-sources (llm-stats.com via the zeroeval Stats API, with LiteLLM as a fallback
-for pricing). One daily run produces three signals consumed by the
-``auto-update-models`` workflow:
-
-* **new_models** - candidates newly published upstream that OCS hasn't
-  registered yet. Feeds the Claude Code job that opens a "Register new models"
-  PR.
+Reconcile OCS's in-repo model catalogue and pricing seed against LiteLLM's
```
SmittieC, line 3: Fixed in d7424997e. Split into sibling modules: the rate-limit-aware HTTP client is now `scripts/reconcile_http.py` (5 functions) and the AST reader for OCS's own model catalogue is `scripts/reconcile_catalogue.py` (8 functions). Both are leaf layers — nothing in them reaches back into the pipeline — so the split is acyclic. `reconcile_models.py` is down from 81 functions to 68.

Verified behaviour-preserving: a `--dry-run` against the live LiteLLM price table produces a byte-identical output payload before and after, and the 155 script tests pass.

*— Claude Code*

scripts/reconcile_models.py
```diff
@@ -104,13 +109,97 @@
...
+RATE_LIMIT_JITTER_SECONDS = 1.0
+
+# Backstop against a server that 429s forever. Reaching it means something is
+# wrong beyond ordinary burst pressure, so the run fails rather than hangs.
+MAX_TOTAL_BURST_WAIT_SECONDS = 900.0
+
+
+class QuotaExceeded(RuntimeError):
+    """The daily quota is spent. It resets at UTC midnight, so retrying now
+    cannot help - the caller should stop making data requests."""
```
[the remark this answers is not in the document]
SmittieC: Not applicable — `_enrich_model`, `fetch_candidates` and `QuotaExceeded` no longer exist. This was reviewed at 74f8591d, when discovery still called the llm-stats API; the later commits on this branch dropped llm-stats entirely and source everything from the LiteLLM price table, which has no daily quota. There is no `except QuotaExceeded` branch in `main` left to reach.

*— Claude Code*

scripts/reconcile_models.py
```diff
@@ -104,13 +109,97 @@
...
 
 
+def _error_limit_type(exc: urllib.error.HTTPError) -> str:
+    """Which limit a 429 hit. Unlabelled 429s are treated as burst: burst is
+    what a run pushes against, and sleeping is the recoverable guess."""
+    try:
+        body = json.loads(exc.read() or b"{}")
+    except (ValueError, OSError):
+        return BURST_LIMIT_TYPE
+    limit_type = (body.get("error") or {}).get("limit_type")
```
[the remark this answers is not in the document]
SmittieC: Not applicable — the `limit_type` body parse is gone. Distinguishing a burst 429 from a daily one was an llm-stats concern; with that source removed, `_get_json` (now in `scripts/reconcile_http.py`) never reads the 429 body, so there is no `body.get("error")` to guard. Rate-limit classification is header-only, in `_is_rate_limited`.

*— Claude Code*

scripts/test_reconcile_models.py
```diff
@@ -1,62 +1,74 @@
...
 import pytest
+import reconcile_http
+import reconcile_models
+from reconcile_catalogue import (
+    IGNORED_MODELS_REL_PATH,
+    load_active_default_models,
+    load_ignored_models,
+    load_registered_models,
+)
+from reconcile_http import (
```
SmittieC: why is this called reconcile_http?

scripts/reconcile_ignored_models.json
```diff
@@ -0,0 +1 @@
+[]
```
SmittieC: what is the purpose of this?

.github/workflows/auto-update-models.yml
```diff
@@ -1,24 +1,30 @@
 # Auto Update LLM Models Workflow
 #
-# Daily reconciliation against llm-stats.com (via the zeroeval Stats API):
-# scripts/reconcile_models.py runs once and produces three signals:
+# Daily reconciliation of the in-repo model catalogue and pricing seed against
+# LiteLLM's model_prices_and_context_window.json, which carries pricing, token
+# limits, provider, model kind and deprecation dates for ~3900 models and needs
+# no credentials. scripts/reconcile_models.py produces four signals:
```
SmittieC:
```suggestion
# no credentials. The scripts/auto_sync_models package produces four signals:
```

.github/workflows/auto-update-models.yml
```diff
@@ -55,10 +55,13 @@ jobs:
       contents: write
       pull-requests: write
       issues: write
+    timeout-minutes: 30
     outputs:
       has_new_models: ${{ steps.reconcile.outputs.has_new_models }}
       new_model_count: ${{ steps.reconcile.outputs.new_model_count }}
       new_model_ids: ${{ steps.reconcile.outputs.new_model_ids }}
+      has_deprecated_upstream: ${{ steps.reconcile.outputs.has_deprecated_upstream }}
+      deprecated_upstream_count: ${{ steps.reconcile.outputs.deprecated_upstream_count }}
```
SmittieC:
```suggestion
      deprecated_upstream_count: ${{ steps.reconcile.outputs.deprecated_upstream_count }}
      has_catalogue_work: ${{ steps.reconcile.outputs.has_catalogue_work }}
```

.github/workflows/auto-update-models.yml
```diff
@@ -80,16 +83,15 @@ jobs:
...
+          # Lifts the commit lookup off the 60-per-hour unauthenticated limit
+          # that every job sharing the runner's IP draws on.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           # Stdlib-only - no dependency install needed.
-          python3 scripts/reconcile_models.py \
-            --bearer-token "$LLM_STATS_BEARER_TOKEN" \
-            --days "${{ inputs.days || '1' }}" \
-            --output reconciliation.json
+          python3 scripts/reconcile_models.py --output reconciliation.json
```
SmittieC:
```suggestion
          python3 -m scripts.auto_sync_models.run --output reconciliation.json
```

.github/workflows/auto-update-models.yml
```diff
@@ -80,16 +83,15 @@ jobs:
...
 
       - name: Upload reconciliation payload
-        if: steps.reconcile.outputs.has_new_models == 'true'
+        if: steps.reconcile.outputs.has_new_models == 'true' || steps.reconcile.outputs.has_deprecated_upstream == 'true'
```
SmittieC:
```suggestion
        if: steps.reconcile.outputs.has_catalogue_work == 'true'
```

.github/workflows/auto-update-models.yml
```diff
@@ -155,8 +157,9 @@ jobs:
 
   new-models-pr:
     needs: reconcile
-    if: needs.reconcile.outputs.has_new_models == 'true'
+    if: needs.reconcile.outputs.has_new_models == 'true' || needs.reconcile.outputs.has_deprecated_upstream == 'true'
```
SmittieC:
```suggestion
    if: needs.reconcile.outputs.has_catalogue_work == 'true'
```

.github/workflows/auto-update-models.yml
```diff
@@ -253,88 +256,125 @@ jobs:
...
-            The payload covers the last ${{ inputs.days || '1' }} day(s). Key fields:
-            - `new_models[]` - candidates to consider. Each has: `id`, `org`,
-              `context_window`, `ocs_providers` (the default OCS provider mapping),
-              `already_registered_providers`, `already_priced_providers`,
-              `source_url`, `sources`, `details_error`, and `pricing`.
-            - `pricing.has_pricing` / `pricing.rates` (per-1K-token) /
-              `pricing.source` (`llm_stats` or `litellm`) /
-              `pricing.llm_pricing_entries` - ready-to-paste `llm_pricing.json`
-              entries, one per provider that still needs pricing.
+            `scripts/reconcile_models.py` has already read the LiteLLM price
```
SmittieC:
```suggestion
            `scripts/auto_sync_models` has already read the LiteLLM price
```

.github/workflows/auto-update-models.yml
```diff
@@ -253,88 +256,125 @@ jobs:
...
-                 generally available on Azure OpenAI - otherwise drop `azure` and
-                 add to `openai` only. Skip any provider already in
-                 `already_registered_providers`.
+               - `ocs_providers` comes from LiteLLM's provider tagging. It is
+                 usually right but it is still upstream metadata: check each
+                 provider actually serves the model under that exact name
+                 before registering it, and drop any that doesn't. Skip any
+                 provider already in `already_registered_providers`.
+               - When you deliberately skip a model, append it to
+                 `scripts/reconcile_ignored_models.json` as
```
SmittieC:
```suggestion
                 `scripts/auto_sync_models/ignored_models.json` as
```

.github/workflows/auto-update-models.yml
```diff
@@ -348,12 +388,15 @@ jobs:
...
-                ### `<model-id>` (org: `<org>`)
+                ### `<model-id>`
                 - **Verdict:** Added | Skipped | Needs follow-up
                 - **Mapped providers:** <e.g. openai, azure>  (if added)
-                - **Pricing:** <source: llm_stats | litellm | unpriced>
+                - **Pricing:** <litellm | partial | unpriced>  (name the
+                  providers left unpriced, if any)
                 - **Reason:** <one-sentence rationale - why added/skipped, or
                   what info was missing>
+                - **Skip recorded:** <yes, in reconcile_ignored_models.json |
```
SmittieC:
```suggestion
                - **Skip recorded:** <yes, in auto_sync_models/ignored_models.json |
```

apps/cost_tracking/README.md
```diff
@@ -36,11 +36,14 @@ Reads split on one rule ([ADR-0048](../../docs/adr/0048-evaluation-spend-is-team
 
 The canonical pricing seed lives in `seed_data/llm_pricing.json` (per-1K-tokens, one entry per `(provider_type, model_name)`). Migration `0002_seed_pricing.py` and any subsequent `NNNN_rate_update_*.py` migrations load it via `load_pricing_data()` from `migration_utils.py`, which calls the `load_ai_pricing` management command. The loader is idempotent and handles supersession on rate changes.
 
-`.github/workflows/auto-update-models.yml` runs `scripts/reconcile_models.py` daily; one upstream pass produces three signals that the workflow acts on:
+`.github/workflows/auto-update-models.yml` runs `scripts/reconcile_models.py` daily; one pass over LiteLLM's `model_prices_and_context_window.json` produces four signals that the workflow acts on:
```
SmittieC:
```suggestion
`.github/workflows/auto-update-models.yml` runs `python3 -m scripts.auto_sync_models.run` daily; one pass over LiteLLM's `model_prices_and_context_window.json` produces four signals that the workflow acts on:
```

apps/cost_tracking/README.md
```diff
@@ -36,11 +36,14 @@ Reads split on one rule ([ADR-0048](../../docs/adr/0048-evaluation-spend-is-team
...
 
-- `new_models` - candidates not yet in `default_models.py`. Claude Code opens a PR to register them following `docs/developer_guides/managing_models.md`.
+- `new_models` - candidates not yet in `default_models.py`. Claude Code opens a PR to register them following `docs/developer_guides/managing_models.md`. Models a reviewer looked at and rejected are recorded in `scripts/reconcile_ignored_models.json`, which keeps them out of later runs.
```
SmittieC:
```suggestion
- `new_models` - candidates not yet in `default_models.py`. Claude Code opens a PR to register them following `docs/developer_guides/managing_models.md`. Models a reviewer looked at and rejected are recorded in `scripts/auto_sync_models/ignored_models.json`, which keeps them out of later runs.
```

.github/workflows/auto-update-models.yml
```diff
@@ -77,306 +104,236 @@ jobs:
...
-            echo "cc @dimagi/open-chat-studio"
-            echo
-            cat "$ISSUE_BODY_PATH"
-          } > issue-body.md
-          EXISTING=$(gh issue list \
-            --state open \
-            --label cost-tracking \
-            --search "Missing pricing for in:title" \
-            --json number --jq '.[0].number // empty')
+          # The ledger advances only when a PR merges, so the same models are
```
SmittieC: remove comment

.github/workflows/auto-update-models.yml
```diff
@@ -77,306 +104,236 @@ jobs:
...
+            a third tuple element naming a replacement where one exists so
+            references move rather than clear, and add
+            `RunDataMigration("remove_deprecated_models", command_options={"force": True})`
+            to your migration. That repoints or clears live references and
+            emails the affected teams, so anything doubtful goes under
+            `## Needs follow-up` instead.
+
+            Keep a deleted model's pricing rows so past usage stays costable.
+
+            ## 4. Write the migration
```
SmittieC: This step should direct claude to docs/developer_guides/managing_models.md

.github/workflows/auto-update-models.yml
```diff
@@ -77,306 +104,236 @@ jobs:
...
+            Where you are confident the script is wrong, correct it. Every
+            correction goes in the report and in the PR description, with what
+            you changed and why. Where you are unsure, leave it and flag it
+            under `## Needs follow-up`. Silent edits to billing data are not
+            acceptable.
+
+            ## 6. Commit and open one PR
+
+            Commit `default_models.py`, `llm_pricing.json`, the migration and
+            the ledger. `reconciliation.json`, `reconciliation.pricing-body.md`
```
SmittieC: Do not commit reconciliation.pricing-body.md. Add it to gitignore

scripts/auto_sync_models/model_ledger.json
```diff
@@ -0,0 +1 @@
+{}
```
SmittieC: currently this is used as both a staging ground for pending models and a "memory" of models to ignore (perhaps those that are timestamped). I do want to eventually move to a world where this ledger is the single source of truth for ALL models: live, pending, rejected + their pricing and supported params. We can build a "translation layer" or "presentation layer" that dictates to the pipeline how to render a certain parameter. The code can then read from here instead of `default_models.py`. That's the high level vision.

@snopoke I'm curious what you think about this.

scripts/auto_sync_models/catalogue.py
```diff
@@ -0,0 +1,191 @@
...
+    catalogue = {record.key: record for record in _model_records(assignments.get("DEFAULT_LLM_PROVIDER_MODELS"))}
+    return catalogue, set(_deleted_keys(assignments.get("DELETED_MODELS")))
+
+
+def _module_assignments(path: Path) -> dict[str, ast.expr]:
+    """``{assigned_name: value_node}`` for every top-level assignment."""
+    tree = ast.parse(path.read_text())
+    return {
+        target.id: node.value
+        for node in tree.body
```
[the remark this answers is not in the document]
SmittieC: Real issue, fixed in 7235efe30 — though the AnnAssign gap was the smaller half of it.

`_module_assignments` now collects `ast.AnnAssign` as well as `ast.Assign` (via a `_bindings` helper, which also skips a bare annotation that binds no value), so a type hint on `DEFAULT_LLM_PROVIDER_MODELS` **or** `DELETED_MODELS` no longer hides it.

The root cause is that an empty parse was returned as a valid answer. Any reason the name goes missing — a rename, a move into a function, a relocated file — produces the same silent empty catalogue, and the consequence you describe is worse than "bypasses a safety check": `check_removal_scale` is guarded by `if ours and ...`, so with an empty catalogue it does not run at all. So `read_default_models` now raises `CatalogueUnreadable` when the parse yields no models, and `main` catches it alongside `UpstreamUnavailable` for the same exit code 1.

Covered by `test_catalogue_reads_an_annotated_assignment` and `test_a_catalogue_that_parses_to_nothing_is_unreadable` (four cases: name absent, empty dict, annotated-but-unassigned, non-literal value).

*— Claude Code*

scripts/auto_sync_models/upstream.py
```diff
@@ -0,0 +1,287 @@
...
+    try:
+        return datetime.date.fromisoformat(str(raw)[:10]) <= today
+    except ValueError:
+        return False
+
+
+def _token_limit(entry: dict) -> int | None:
+    for field_name in ("max_input_tokens", "max_tokens"):
+        value = entry.get(field_name)
+        if isinstance(value, int) and not isinstance(value, bool) and value > 0:
```
SmittieC: Fixed in 07a38fc24. The int/bool/positive test moves out of `_token_limit` into `_positive_int`, which mirrors the shape `catalogue._const_int` already uses, leaving the loop as "first field that yields a limit". Pinned by a new `bool-is-not-a-limit` case on `test_translate_reads_the_token_limit`.

*— Claude Code*

.github/workflows/auto-update-models.yml
SmittieC: Already in place — the suggestion landed with the single-job rewrite (ed8646fef, c93705db1, 752adeb8f). The reconcile step now runs exactly `python3 -m scripts.auto_sync_models.run --output reconciliation.json`; the bearer token and `--days` are gone with `scripts/reconcile_models.py`.

*— Claude Code*

.github/workflows/auto-update-models.yml
SmittieC: Fixed in a3e8f1e29 — the two-line comment above the `gh pr list` call is gone. The header block already states the skip-rather-than-stack rule and the step echoes it at runtime.

*— Claude Code*

.github/workflows/auto-update-models.yml
SmittieC: Fixed in f2ea428a5. Step 4 no longer restates the migration recipe; it now reads "Read `docs/developer_guides/managing_models.md` and follow it. It is the authority on what the migration contains and on stripping `llm_model_migration()` from the earlier migrations that still call it."

*— Claude Code*

.github/workflows/auto-update-models.yml
SmittieC: Already handled — `.gitignore:62` carries `reconciliation.*.md` (added in f11b4afcb), which covers it. Confirmed locally:

```
$ git check-ignore -v reconciliation.pricing-body.md
.gitignore:62:reconciliation.*.md	reconciliation.pricing-body.md
```

The line you commented on is the prompt telling Claude the file is gitignored and not to force-add it, so both halves are in place.

*— Claude Code*

scripts/auto_sync_models/model_ledger.json
SmittieC: No code change — leaving this open for @snopoke, since it is a design question about where the ledger should sit rather than feedback on this PR.

One note on the current state, for the thread: the ledger is deliberately *not* a record of live models today. Registering a model deletes its entry, so `default_models.py` stays the single record that we serve it. Making the ledger the source of truth for live models as well would invert that, which is a real change rather than an extension.

*— Claude Code*

scripts/auto_sync_models/dispatch.py
```diff
@@ -0,0 +1,239 @@
...
+    diff = reconciliation.diff
+    _write_pricing_update(diff, repo_root, output, dry_run)
+    if diff.added and not dry_run:
+        advanced = advance_ledger(reconciliation.ledger, diff.added, reconciliation.today)
+        path = write_ledger(repo_root, advanced)
+        print(f"  -> recorded {len(diff.added)} model(s) as pending in {path.name}")
+
+    run_date = datetime.datetime.now(datetime.UTC).isoformat()
+    payload = build_payload(diff, reconciliation.orphan_rows, run_date=run_date)
+    output.write_text(json.dumps(payload, indent=2) + "\n")
```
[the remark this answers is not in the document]
SmittieC: Real, and fixed in 0e29394a3 — by removing `--output` rather than by creating the directory.

The suggested `output.parent.mkdir(parents=True, exist_ok=True)` treats the symptom. The flag was the problem: nothing ever passed anything but the default, and the workflow passed the default explicitly. A configurable output path bought no flexibility and carried this failure mode, plus a second one you did not flag — `--output` was resolved against the *current working directory* while everything else in the run resolves against `--repo-root`, so running from anywhere but the repo root wrote the payload somewhere the Claude Code step does not read.

Both paths are now fixed constants in `dispatch.py` resolved against `repo_root`:

```python
RECONCILIATION_REL_PATH = "reconciliation.json"
PRICING_BODY_REL_PATH = "reconciliation.pricing-body.md"
```

`repo_root` must already exist — layer 1 has read `default_models.py` out of it before dispatch runs — so there is no parent directory left to be missing, and the derived `.pricing-body.md` name (previously `output.with_name(output.stem + ...)`) now matches the literal name `.gitignore` and the workflow prompt already use.

Verified end-to-end against the live LiteLLM table from a different working directory: both artifacts landed under `--repo-root`, none in the cwd. Pinned by `test_dispatch_writes_both_artifacts_under_the_repo_root`.

*— Claude Code*

snopoke (commented):
didn't review closely but seems low risk

snopoke (approved):

PR #4471, opened 2026-09-10
Ability to change the node type of an existing node

### LINKED ISSUE #1452: Ability to change the node type of an existing node
It will save a lot of time when users don't have to re-create a node when wanting to swap out a node with another type.

Davemaina1:
### Product Description

Lets you change an existing pipeline node's type from the builder canvas, keeping its position, id, name, color, and edges wherever the new type still offers a matching output. No more deleting the node and recreating it from scratch.

### Technical Description

Params reset to the new type's own defaults rather than carrying over same-named values across types. `apps/api/v2/pipeline_edit/serializers.py`'s `NodeUpdateSerializer` already refuses a type change in place on the v2 API for exactly this reason (`test_patch_refuses_to_change_a_nodes_type`), a shared field name isn't proof two types give it the same meaning. Resetting sidesteps that: nothing crosses the type boundary except `name` and `color`, neither of which is a domain param.

The remaining real problem is output handles: `RouterNode`/`StaticRouterNode`/`BooleanNode` expose multiple named handles, everything else exposes one. `graph_editor.py`'s `plan_update` already rewired or dropped edges when a *param* edit changed a node's handles, but only on the v2 API, the human builder's own save path (`patching.py::apply_pipeline_patch`) never did, for any update. Extracted that logic into `apps/pipelines/build_state.py` (`output_handle_labels`, `rewired_edges_for_node`) so both callers share one implementation, and wired it into `patching.py`'s node-update path. This also fixes a pre-existing, unrelated gap: editing a router's keywords through the visual builder could already dangle an edge before this PR.

Two bugs turned up only by clicking through the feature in a running pipeline, not by tests or static review, both fixed with a regression test:
- `GetInputWidget.tsx` unconditionally wrote `params.keywords` on every node render regardless of schema. Harmless normally; a type change hands it a fresh, already-frozen (immer auto-freeze) params object with no `keywords` key, which crashed on write.
- `build_state.py`'s router fallback path called `enumerate()` on `keywords` without checking it wasn't `None`  reachable because the frontend has no static JSON-schema default to offer for a `default_factory=list` field, so a swap into a router type sends `keywords: null`.

Also: `Node.update_from_params()` only cleared `CustomActionOperation` rows when the node's *current* type is `LLMResponseWithPrompt`, so a type change away from it left old rows orphaned. Added the missing `elif` branch.

### Issue Link

Closes #1452

### Demo

change-node-type-1452-demo

Swapping an LLM node to LLM Router: params reset and the stale edge to the next node is dropped since the new type offers no matching output.

1. Open any pipeline in the builder and select a node.
2. Click the new swap icon in the node's toolbar (next to delete/edit).
3. Pick a different type from the list, the node keeps its id, position, and any edges its new type can still offer a matching output for; other params reset to the new type's defaults.

### Docs and Changelog

User-facing builder feature: a "change type" control on each pipeline node, letting a user swap it to a different node type in place instead of deleting and recreating it. Params reset to the new type's defaults (except name/color); edges are kept where the new type still offers a matching output, dropped otherwise.

### Operator Impact

### DIFF
--- a/apps/pipelines/build_state.py
+++ b/apps/pipelines/build_state.py
@@ -208,5 +208,10 @@ def _router_output_map(
         fallback = dict(params)
         if isinstance(fallback.get("keywords"), list):
             fallback["keywords"] = [str(keyword).upper() for keyword in fallback["keywords"]]
+        else:
+            # Anything else -- missing, or explicitly None (what a type change (#1452) sends,
+            # since the frontend has no static default to offer for a default_factory field) --
+            # is "no keywords yet", not a value get_output_map() can enumerate().
+            fallback["keywords"] = []
         instance = node_class.model_construct(**fallback)
     return instance.get_output_map()
--- a/apps/pipelines/build_state.py
+++ b/apps/pipelines/build_state.py
@@ -18,7 +18,7 @@ import pydantic
 
 from apps.pipelines.const import STANDARD_INPUT_NAME, STANDARD_OUTPUT_NAME
 from apps.pipelines.exceptions import PipelineNodeBuildError, has_errors
-from apps.pipelines.flow import Flow
+from apps.pipelines.flow import Flow, FlowEdge, FlowNodeData
 from apps.pipelines.models import Node, Pipeline
 from apps.pipelines.nodes.base import PipelineRouterNode, resolve_node_class
 from apps.pipelines.nodes.nodes import EndNode, StartNode
@@ -136,6 +136,58 @@ def output_handles(node_type: str, params: dict, node_id: str, django_node: Node
         output_map = _router_output_map(node_class, params, node_id, django_node)
         return [{"handle": handle, "label": label} for handle, label in output_map.items()]
     return [{"handle": STANDARD_OUTPUT_NAME, "label": None}]
+
+
+def output_handle_labels(content: FlowNodeData) -> dict[str, str | None]:
+    """``output_handles()`` as a ``{handle: label}`` map, for matching handles across an edit."""
+    return {handle["handle"]: handle["label"] for handle in output_handles(content.type, content.params, content.id)}
+
+
+def handle_remap(before: dict[str, str | None], after: dict[str, str | None]) -> dict[str, str]:
+    """Where each handle a node used to offer has ended up, keyed by the handle it was.
+
+    A handle whose branch the edit removed is absent: its edge has nowhere to go. A rename counts
+    as a removal -- inheriting the old branch's target would wire the new one somewhere nobody
+    chose.
+    """
+    if _labels_are_distinct(before) and _labels_are_distinct(after):
+        destinations = {label: handle for handle, label in after.items()}
+        return {handle: destinations[label] for handle, label in before.items() if label in destinations}
+    # Duplicate branch labels: keywords have to be unique, but a router that breaks that is still
+    # writable, and which edge belongs to which of two identical branches is a guess. So handles
+    # are followed by position instead, and only an edge left with no handle at all is dropped.
+    return {handle: handle for handle in before if handle in after}
+
+
+def _labels_are_distinct(handles: dict[str, str | None]) -> bool:
+    """Whether every handle in the map carries a different branch label."""
+    return len(set(handles.values())) == len(handles)
+
+
+def rewired_edges_for_node(
+    edges: list[FlowEdge], node_id: str, before: dict[str, str | None], after: dict[str, str | None]
+) -> tuple[list[FlowEdge], list[str]]:
+    """Edges sourced from ``node_id`` after its output handles change from ``before`` to ``after``.
+
+    Returns ``(updated, deleted_ids)``: ``updated`` are copies of the affected edges with
+    ``sourceHandle`` moved to follow the same branch label; ``deleted_ids`` are edges whose branch
+    is gone. An edge not sourced from this node, or already on a handle the node didn't offer
+    before the change, is left out of both -- it is either unaffected, or was already stranded
+    before this edit and is not this edit's problem to clean up.
+    """
+    moved_to = handle_remap(before, after)
+    updated: list[FlowEdge] = []
+    deleted: list[str] = []
+    for edge in edges:
+        handle = edge.source_handle_name
+        if edge.source != node_id or handle not in before:
+            continue
+        destination = moved_to.get(handle)
+        if destination is None:
+            deleted.append(edge.id)
+        elif destination != handle:
+            updated.append(edge.model_copy(update={"sourceHandle": destination}))
+    return updated, deleted
 
 
 def _router_output_map(
--- a/apps/pipelines/models.py
+++ b/apps/pipelines/models.py
@@ -630,6 +630,11 @@ class Node(BaseModel, VersionsMixin, CustomActionOperationMixin):
                 custom_action_infos.append({"custom_action_id": custom_action_id, "operation_id": operation_id})
 
             set_custom_actions(self, custom_action_infos)
+        elif self.custom_action_operations.exists():
+            # A type change (#1452) can leave this node no longer LLMResponseWithPrompt while its
+            # old CustomActionOperation rows still exist -- the branch above that would normally
+            # keep them in sync with `params` never runs for a type it doesn't recognise.
+            set_custom_actions(self, [])
 
     @classmethod
     def resource_fk_fields(cls):
--- a/apps/pipelines/patching.py
+++ b/apps/pipelines/patching.py
@@ -77,7 +77,10 @@ def _rewire_edges_for_update(flow: Flow, node_id: str, previous: FlowNode | None
 
     Covers a param edit that changes a router's branches (the pre-existing gap this closes) and a
     type change (#1452) the same way, since both are just an update whose handles differ before
-    and after -- neither is special-cased.
+    and after -- neither is special-cased. Runs before the edge diff below, so a caller that also
+    sends its own correct edge diff for the same node (the v2 API always does) just re-applies the
+    same values afterward -- harmless. A deleted edge is the one thing that step must not bring
+    back; see _apply_edge_diff.
     """
     previous_data = previous.data if previous else None
     if previous_data is None or updated.data is None:
@@ -101,9 +104,13 @@ def _apply_edge_diff(flow: Flow, diff: EdgeDiff) -> None:
     for edge_id in diff.delete:
         edge_map.pop(edge_id, None)
 
-    # Update: replace in-place
+    # Update: replace in-place -- but never resurrect an id the node-update rewiring above
+    # already dropped from this same flow (a caller's edge diff can carry a stale value for
+    # an edge whose handle no longer exists once the node update lands; see
+    # _rewire_edges_for_update).
     for updated in diff.update:
-        edge_map[updated.id] = updated
+        if updated.id in edge_map:
+            edge_map[updated.id] = updated
 
     # Add: insert, skip if already present (idempotent)
     for added in diff.add:
--- a/apps/pipelines/patching.py
+++ b/apps/pipelines/patching.py
@@ -79,9 +79,10 @@ def _rewire_edges_for_update(flow: Flow, node_id: str, previous: FlowNode | None
     type change (#1452) the same way, since both are just an update whose handles differ before
     and after -- neither is special-cased.
     """
-    if previous is None or previous.data is None or updated.data is None:
+    previous_data = previous.data if previous else None
+    if previous_data is None or updated.data is None:
         return
-    before = output_handle_labels(previous.data)
+    before = output_handle_labels(previous_data)
     after = output_handle_labels(updated.data)
     if before == after:
         return
--- a/apps/pipelines/patching.py
+++ b/apps/pipelines/patching.py
@@ -5,6 +5,7 @@ Never touches the database directly — the caller (the PATCH view) is responsib
 for persisting the merged graph and calling update_nodes_from_data().
 """
 
+from apps.pipelines.build_state import output_handle_labels, rewired_edges_for_node
 from apps.pipelines.flow import (
     EdgeDiff,
     Flow,
@@ -52,9 +53,11 @@ def _apply_node_diff(flow: Flow, diff: NodeDiff) -> None:
     for node_id in diff.delete:
         node_map.pop(node_id, None)
 
-    # Update: replace in-place
+    # Update: replace in-place, following or dropping edges whose output handles moved
     for updated in diff.update:
+        previous = node_map.get(updated.id)
         node_map[updated.id] = updated
+        _rewire_edges_for_update(flow, updated.id, previous, updated)
 
     # Add: insert, skip if already present (idempotent)
     for added in diff.add:
@@ -67,6 +70,27 @@ def _apply_node_diff(flow: Flow, diff: NodeDiff) -> None:
     deleted_ids = set(diff.delete)
     if deleted_ids:
         flow.edges = [edge for edge in flow.edges if edge.source not in deleted_ids and edge.target not in deleted_ids]
+
+
+def _rewire_edges_for_update(flow: Flow, node_id: str, previous: FlowNode | None, updated: FlowNode) -> None:
+    """Follow or drop ``node_id``'s outgoing edges when this update changes its output handles.
+
+    Covers a param edit that changes a router's branches (the pre-existing gap this closes) and a
+    type change (#1452) the same way, since both are just an update whose handles differ before
+    and after -- neither is special-cased.
+    """
+    if previous is None or previous.data is None or updated.data is None:
+        return
+    before = output_handle_labels(previous.data)
+    after = output_handle_labels(updated.data)
+    if before == after:
+        return
+    changed, deleted_ids = rewired_edges_for_node(flow.edges, node_id, before, after)
+    if not changed and not deleted_ids:
+        return
+    changed_by_id = {edge.id: edge for edge in changed}
+    deleted = set(deleted_ids)
+    flow.edges = [changed_by_id.get(edge.id, edge) for edge in flow.edges if edge.id not in deleted]
 
 
 def _apply_edge_diff(flow: Flow, diff: EdgeDiff) -> None:
--- a/apps/pipelines/tests/test_build_state.py
+++ b/apps/pipelines/tests/test_build_state.py
@@ -112,6 +112,19 @@ class TestNodeOutputHandles:
             {"handle": "output_1", "label": "B"},
         ]
 
+    def test_router_with_none_keywords_still_reports_handles(self):
+        # #1452: a type change to a router sends no keywords param at all (the frontend has no
+        # empty-list default to offer for a default_factory field), so params gets an explicit
+        # `None` rather than a missing key. route_key is also missing here, so full validation
+        # fails and falls back to the unvalidated instance -- which must treat None the same as
+        # "no keywords yet", not enumerate() it directly.
+        node = Node(
+            flow_id="router-1",
+            type="StaticRouterNode",
+            params={"name": "router", "keywords": None},
+        )
+        assert node_output_handles(node) == []
+
     def test_unknown_node_type_has_no_output_handles(self):
         node = Node(flow_id="ghost-1", type="GhostNode", params={"name": "ghost"})
         assert node_output_handles(node) == []
--- a/assets/javascript/apps/pipeline/PipelineNode.tsx
+++ b/assets/javascript/apps/pipeline/PipelineNode.tsx
@@ -96,7 +96,7 @@ export function PipelineNode(nodeProps: NodeProps<NodeData>) {
               </button>
             )}
             {changeableTypes.length > 0 && (
-              <div className="dropdown dropdown-top">
+              <div className="dropdown dropdown-bottom">
                   <button className="btn btn-xs join-item" aria-label="Change node type">
                       <i className="fa-solid fa-right-left"></i>
                   </button>
--- a/assets/javascript/apps/pipeline/PipelineNode.tsx
+++ b/assets/javascript/apps/pipeline/PipelineNode.tsx
@@ -1,6 +1,6 @@
 import {Node, NodeProps, NodeToolbar, Position} from "reactflow";
 import React, {ChangeEvent, MouseEvent} from "react";
-import {concatenate, formatDocsForSchema, getCachedData, nodeBorderClass} from "./utils";
+import {buildTypeChangeParams, concatenate, formatDocsForSchema, getCachedData, nodeBorderClass} from "./utils";
 import usePipelineStore from "./stores/pipelineStore";
 import useEditorStore from "./stores/editorStore";
 import {JsonSchema, NodeData} from "./types/nodeParams";
@@ -61,6 +61,23 @@ export function PipelineNode(nodeProps: NodeProps<NodeData>) {
     }));
   };
 
+  // Other user-addable types this node could become (#1452). Start/End/Passthrough are never
+  // offered, in either direction: a node whose own schema isn't `ui:can_add` gets no control at
+  // all, and this list excludes them as targets the same way ComponentList's palette does.
+  const changeableTypes = nodeSchema["ui:can_add"]
+    ? Array.from(getCachedData().nodeSchemas.values())
+        .filter((schema) => schema["ui:can_add"] && schema.title !== data.type)
+        .sort((a, b) => a["ui:label"].localeCompare(b["ui:label"]))
+    : [];
+
+  const changeNodeType = (newSchema: JsonSchema) => {
+    setNode(id, produce((next) => {
+      next.data.type = newSchema.title;
+      next.data.label = newSchema["ui:label"];
+      next.data.params = buildTypeChangeParams(newSchema, data.params);
+    }));
+  };
+
   const currentColor = data.params["color"] || NODE_COLORS[0].value;
   const nodeClasses = `${nodeBorderClass(hasErrors, selected)} ${currentColor}`;
 
@@ -77,6 +94,21 @@ export function PipelineNode(nodeProps: NodeProps<NodeData>) {
               <button className="btn btn-xs join-item" onClick={() => editNode()}>
                   <i className="fa fa-pencil"></i>
               </button>
+            )}
+            {changeableTypes.length > 0 && (
+              <div className="dropdown dropdown-top">
+                  <button className="btn btn-xs join-item" aria-label="Change node type">
+                      <i className="fa-solid fa-right-left"></i>
+                  </button>
+                  <ul tabIndex={0} className="dropdown-content z-[1] menu p-2 shadow bg-base-100 rounded-box w-52 max-h-60 overflow-y-auto">
+                    <li className="menu-title">Change type</li>
+                    {changeableTypes.map((schema) => (
+                      <li key={schema.title}>
+                        <a onClick={() => changeNodeType(schema)}>{schema["ui:label"]}</a>
+                      </li>
+                    ))}
+                  </ul>
+              </div>
             )}
             {nodeDocs && (
               <div className="dropdown dropdown-right">
--- a/assets/javascript/apps/pipeline/PipelineNode.tsx
+++ b/assets/javascript/apps/pipeline/PipelineNode.tsx
@@ -104,7 +104,7 @@ export function PipelineNode(nodeProps: NodeProps<NodeData>) {
                     <li className="menu-title">Change type</li>
                     {changeableTypes.map((schema) => (
                       <li key={schema.title}>
-                        <a onClick={() => changeNodeType(schema)}>{schema["ui:label"]}</a>
+                        <button type="button" onClick={() => changeNodeType(schema)}>{schema["ui:label"]}</button>
                       </li>
                     ))}
                   </ul>
--- a/assets/javascript/apps/pipeline/nodes/GetInputWidget.tsx
+++ b/assets/javascript/apps/pipeline/nodes/GetInputWidget.tsx
@@ -208,9 +208,10 @@ const getWidgetsGeneric = (
       return 0;
     }
   });
-   if (!Array.isArray(nodeData.params.keywords)) {
-    nodeData.params.keywords = [""]; // initialize keywords with size 1
-  }
+  // The "keywords" widget (widgets.tsx) already falls back to an empty list on its own when
+  // params.keywords is missing or not an array, so nothing here needs to lazily add it -- and
+  // mutating params in place is unsafe besides: a type change (#1452) can hand this an
+  // already-frozen object (immer auto-freezes every produce() result), which throws on write.
   return schemaProperties.map((name) => (
     <React.Fragment key={name}>
       {widgetGenerator({
--- a/assets/javascript/apps/pipeline/utils.tsx
+++ b/assets/javascript/apps/pipeline/utils.tsx
@@ -1,7 +1,7 @@
 import React from "react"
 import ShortUniqueId from "short-unique-id";
 import {NodeParameterValues, Option} from "./types/nodeParameterValues";
-import {JsonSchema, PropertySchema} from "./types/nodeParams";
+import {JsonSchema, NodeParams, PropertySchema} from "./types/nodeParams";
 
 declare global {
   interface Window {
@@ -75,6 +75,39 @@ export function getDocumentationLink(schema: JsonSchema) {
   return documentationLink;
 }
 
+
+/**
+ * The param values a fresh node of this schema starts with: each property's own default,
+ * falling back to the team's cached default-values map, or null when neither has one.
+ */
+export function getDefaultParamValues(schema: JsonSchema): NodeParams {
+  const {defaultValues} = getCachedData();
+  const defaults: NodeParams = {name: ""};
+  for (const name in schema.properties) {
+    const property = schema.properties[name];
+    defaults[name] = [property.default, defaultValues[name]].find((value) => value !== undefined && value !== null) ?? null;
+  }
+  return defaults;
+}
+
+/**
+ * The param values a node keeps when its type changes to `newSchema` (#1452).
+ *
+ * Every type-specific param resets to the new type's own defaults, never carried across by
+ * matching field names -- a name shared between two schemas is not proof it means the same
+ * thing in both (see `NodeUpdateSerializer` on the v2 API, which refuses a type change in place
+ * for exactly this reason). `name` and `color` are the only survivors: `name` is UI identity
+ * every schema declares, and `color` is UI-only state that is not a schema field on any node
+ * type at all, so it is carried separately rather than lost to the reset.
+ */
+export function buildTypeChangeParams(newSchema: JsonSchema, currentParams: NodeParams): NodeParams {
+  const params = getDefaultParamValues(newSchema);
+  params.name = currentParams.name;
+  if (currentParams.color !== undefined) {
+    params.color = currentParams.color;
+  }
+  return params;
+}
 
 export function concatenate(value: string | string[] | null | undefined): string {
   if (!value) return "";
--- a/assets/javascript/apps/pipeline/utils.tsx
+++ b/assets/javascript/apps/pipeline/utils.tsx
@@ -84,6 +84,10 @@ export function getDefaultParamValues(schema: JsonSchema): NodeParams {
   const {defaultValues} = getCachedData();
   const defaults: NodeParams = {name: ""};
   for (const name in schema.properties) {
+    // Every real node schema declares "name" as one of its own properties (it's a field on the
+    // shared base node), but it's never given a schema or cached default here -- it's set by
+    // the caller (a generated id on add, the preserved name on a type change).
+    if (name === "name") continue;
     const property = schema.properties[name];
     defaults[name] = [property.default, defaultValues[name]].find((value) => value !== undefined && value !== null) ?? null;
   }
--- a/apps/api/v2/pipeline_edit/graph_editor.py
+++ b/apps/api/v2/pipeline_edit/graph_editor.py
@@ -240,7 +240,3 @@ def _is_overtaken_end_node(node: FlowNode, new_node_x: float) -> bool:
     if node.type != REACT_FLOW_END_TYPE:
         return False
     return (node.position.get("x") or 0) <= new_node_x
-
-
-#: Moved to ``apps.pipelines.build_state`` (``output_handle_labels``, ``rewired_edges_for_node``)
-#: so ``patching.py``'s node-update path can rewire edges the same way, for #1452.
--- a/apps/api/v2/pipeline_edit/graph_editor.py
+++ b/apps/api/v2/pipeline_edit/graph_editor.py
@@ -6,12 +6,11 @@ from rest_framework import status
 from rest_framework.exceptions import APIException, NotFound
 
 from apps.api.v2.discovery.node_types import get_node_class, get_node_type_schema
-from apps.pipelines.build_state import output_handles
+from apps.pipelines.build_state import output_handle_labels, rewired_edges_for_node
 from apps.pipelines.flow import (
     REACT_FLOW_END_TYPE,
     EdgeDiff,
     Flow,
-    FlowEdge,
     FlowNode,
     FlowNodeData,
     NodeDiff,
@@ -25,10 +24,6 @@ from .facade import PipelineEdit, graph_diff
 from .ids import with_free_suffix
 from .node_params import node_params, writable_params
 from .references import check_references
-
-#: A node's output handles as ``{handle: branch label}``. The label is ``None`` for the single
-#: standard output, and a router's branch keyword otherwise.
-OutputHandles = dict[str, str | None]
 
 #: How far right of the rightmost node a new one is parked, and the row it is parked on. The step is
 #: about a node's width, so a parked node clears the one before it.
@@ -79,12 +74,13 @@ def plan_update(flow: Flow, team: Team, node_id: str, label: str | None, params:
 
     Params merge key by key rather than replacing the stored dict: the point of the façade is that
     changing one setting does not mean resending the whole node. An edit that changes which output
-    handles the node offers takes that node's edges with it -- see :func:`_rewired_edges`. Start and
-    End are refused outright, label-only edits included.
+    handles the node offers takes that node's edges with it -- see
+    :func:`~apps.pipelines.build_state.rewired_edges_for_node`. Start and End are refused outright,
+    label-only edits included.
     """
     node, content = find_node(flow, node_id)
     refuse_if_server_managed(content.type)
-    before = _output_handles(content)
+    before = output_handle_labels(content)
     if params:
         # 404s a type the API does not publish. Only when there are params to write: renaming a
         # node of such a type is not something the API has to withhold.
@@ -98,7 +94,8 @@ def plan_update(flow: Flow, team: Team, node_id: str, label: str | None, params:
         content.params = stored_params(content)
     if label is not None:
         content.label = label
-    edges = _rewired_edges(flow, node_id, before, _output_handles(content))
+    changed, deleted_ids = rewired_edges_for_node(flow.edges, node_id, before, output_handle_labels(content))
+    edges = EdgeDiff(update=changed, delete=deleted_ids)
     return PipelineEdit(diff=graph_diff(nodes=NodeDiff(update=[node]), edges=edges), written_ids=[node_id])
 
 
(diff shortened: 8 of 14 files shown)


apps/pipelines/patching.py
```diff
@@ -52,9 +53,11 @@ def _apply_node_diff(flow: Flow, diff: NodeDiff) -> None:
     for node_id in diff.delete:
         node_map.pop(node_id, None)
 
-    # Update: replace in-place
+    # Update: replace in-place, following or dropping edges whose output handles moved
     for updated in diff.update:
+        previous = node_map.get(updated.id)
         node_map[updated.id] = updated
+        _rewire_edges_for_update(flow, updated.id, previous, updated)
```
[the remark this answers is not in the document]
Davemaina1: Confirmed, fixed in ae63ba8d3. Kept the original apply order and added a guard in `_apply_edge_diff`: its update path now only replaces an edge id already present, so it can't resurrect one the node-update rewiring already dropped. A reorder was tried first but broke 3 v2 API tests via a double-transform, so this fix is narrower. Added a regression test for a combined node+stale-edge-update patch.

assets/javascript/apps/pipeline/PipelineNode.tsx
```diff
@@ -78,6 +95,21 @@ export function PipelineNode(nodeProps: NodeProps<NodeData>) {
...
+            {changeableTypes.length > 0 && (
+              <div className="dropdown dropdown-bottom">
+                  <button className="btn btn-xs join-item" aria-label="Change node type">
+                      <i className="fa-solid fa-right-left"></i>
+                  </button>
+                  <ul tabIndex={0} className="dropdown-content z-[1] menu p-2 shadow bg-base-100 rounded-box w-52 max-h-60 overflow-y-auto">
+                    <li className="menu-title">Change type</li>
+                    {changeableTypes.map((schema) => (
+                      <li key={schema.title}>
+                        <a onClick={() => changeNodeType(schema)}>{schema["ui:label"]}</a>
```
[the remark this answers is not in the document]
Davemaina1: Confirmed, fixed in ae63ba8d3 using the suggested `<button type="button">`.

assets/javascript/apps/pipeline/utils.tsx
```diff
@@ -76,6 +76,39 @@ export function getDocumentationLink(schema: JsonSchema) {
...
+/**
+ * The param values a fresh node of this schema starts with: each property's own default,
+ * falling back to the team's cached default-values map, or null when neither has one.
+ */
+export function getDefaultParamValues(schema: JsonSchema): NodeParams {
+  const {defaultValues} = getCachedData();
+  const defaults: NodeParams = {name: ""};
+  for (const name in schema.properties) {
+    const property = schema.properties[name];
+    defaults[name] = [property.default, defaultValues[name]].find((value) => value !== undefined && value !== null) ?? null;
```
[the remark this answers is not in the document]
Davemaina1: Confirmed, fixed in ae63ba8d3. Skipped `name` inside the loop instead of overwriting it after, since every node schema declares `name` as its own property. Added a regression test with a schema that declares a non-empty `name` default.

SmittieC (commented):
The "static router" and "update participant data" nodes are moved to the right, causing a scrollbar to be shown. See the screenshot:

Image

It would be great if we can keep all nodes in one list

assets/javascript/apps/pipeline/nodes/GetInputWidget.tsx
```diff
@@ -208,9 +208,10 @@ const getWidgetsGeneric = (
       return 0;
     }
   });
-   if (!Array.isArray(nodeData.params.keywords)) {
-    nodeData.params.keywords = [""]; // initialize keywords with size 1
-  }
+  // The "keywords" widget (widgets.tsx) already falls back to an empty list on its own when
```
SmittieC: Let's rather remove this comment. It's defending something that the dev wouldn't ask when viewing this code

Davemaina1:
> The "static router" and "update participant data" nodes are moved to the right, causing a scrollbar to be shown. See the screenshot:
> 
> Image
> It would be great if we can keep all nodes in one list


@SmittieC, fixed in a144fe8b. DaisyUI's `.menu` sets `flex-wrap` by default, so a long list wraps into a second column instead of scrolling. Added `flex-nowrap` to keep it one scrollable list.

change-type-dropdown-fix-cropped

apps/pipelines/models.py
```diff
@@ -630,6 +630,11 @@ def update_from_params(self):
                 custom_action_infos.append({"custom_action_id": custom_action_id, "operation_id": operation_id})
 
             set_custom_actions(self, custom_action_infos)
+        elif self.custom_action_operations.exists():
```
SmittieC: I wonder if this branch is necessary. If we remove the `if self.type == LLMResponseWithPrompt.__name__:` check, then `self.params.get("custom_actions") or []` default to the empty list and we'll have what we want?

apps/pipelines/build_state.py
```diff
@@ -156,5 +208,10 @@ def _router_output_map(
         fallback = dict(params)
         if isinstance(fallback.get("keywords"), list):
             fallback["keywords"] = [str(keyword).upper() for keyword in fallback["keywords"]]
+        else:
+            # Anything else -- missing, or explicitly None (what a type change (#1452) sends,
+            # since the frontend has no static default to offer for a default_factory field) --
+            # is "no keywords yet", not a value get_output_map() can enumerate().
+            fallback["keywords"] = []
```
SmittieC: thoughts on this?

```suggestion
keywords = fallback.get("keywords", [])
fallback["keywords"] = [str(keyword).upper() for keyword in keywords]
```

apps/pipelines/patching.py
```diff
@@ -69,16 +72,45 @@ def _apply_node_diff(flow: Flow, diff: NodeDiff) -> None:
         flow.edges = [edge for edge in flow.edges if edge.source not in deleted_ids and edge.target not in deleted_ids]
 
 
+def _rewire_edges_for_update(flow: Flow, node_id: str, previous: FlowNode | None, updated: FlowNode) -> None:
```
SmittieC: 1. Super nit: can we call this `_rewire_edges` instead? `_rewire_edges_for_update` reads a bit weird in my opinion.

2. `FlowNode` has the `node_id`, so no need to pass it as a separate parameter

assets/javascript/apps/pipeline/nodes/GetInputWidget.tsx
Davemaina1: The frozen-params (immer) part isn't obvious from this file so kept it, dropped the rest.

apps/pipelines/models.py
Davemaina1: Yes, kept a guard, just not that one. Drop it entirely and set_custom_actions(), a delete plus a refresh_from_db(), runs on every node's every save instead of only when there's something to sync or clear. Swapped the type check for a data-driven one, done in 256b716.

apps/pipelines/build_state.py
Davemaina1: That crashes on an explicit None, which is what a type change sends here, .get(key, default) only covers a missing key. Used or [] instead, in 256b71641.

apps/pipelines/patching.py
Davemaina1: Done, in 256b71641.

SmittieC (commented):
## 🤖 AI-generated code review

This review was produced by Claude (via Claude Code) at a maintainer's request, not by a human reviewer. Please weigh it accordingly and verify anything before acting on it.

Where a finding says **verified**, it means I ran it — mostly by driving `apply_pipeline_patch` directly against this branch, or by counting queries with `CaptureQueriesContext`. Findings about UI wiring and schema gating were verified by reading the relevant call sites and enumerating the live node schemas. Each inline comment states its own evidence.

---

### The theme

The change-type control itself is in good shape. The risk is concentrated in a decision the PR makes almost in passing: wiring `_rewire_edges` into `_apply_node_diff` imports the **v2 API's deliberate-edit edge policy into the builder's per-keystroke autosave**. Those two callers have very different assumptions about what a node update *means*.

On the v2 API, a `PATCH` of `keywords: [...]` is one considered edit, and there is genuinely no way to tell a rename from a reorder-and-replace — so treating a renamed branch as a removed branch is defensible. In the builder, `KeywordsWidget.updateKeyword` fires on **every keystroke** and `autoSaveCurrentPipline` debounces for 1s. The same policy applied there destroys user work silently.

Four of the inline findings (the two ordering bugs, the rename deletion, and the Ctrl+S bypass below) are really **one design question**: *who owns edge rewiring, and when does it run?* Running it once, **after** `_apply_edge_diff`, over the merged edge set — with **positional** identity on the builder path and **label** identity reserved for the v2 API — resolves all four and removes the need for the new `if updated.id in edge_map` guard entirely.

---

### One finding that could not be inlined

`apps/pipelines/views.py` is not part of this diff, so this one has no anchor line:

**Ctrl+S bypasses the feature entirely.** `_handle_pipeline_post` (`views.py:259-271`) runs `split_flow_data` straight into `Pipeline.data` and never calls `apply_pipeline_patch`, so none of the new rewiring applies. `manualSaveCurrentPipeline` (`Pipeline.tsx:130`, bound to Ctrl+S) calls `savePipeline`, which clears `saveTimeoutId` and POSTs the full graph. Change a node's type and press Ctrl+S inside the 1s autosave window, and edges pointing at `output_0`/`output_1` on a node that no longer offers them are persisted — exactly the state this PR set out to prevent. The same applies to pipeline import and `revert_to_version`, which also reconcile through `update_nodes_from_data` without the patch engine.

Related: `savePipeline` (`pipelineStore.ts:418`) sets `isSaving: true` without ever checking it — only `_flushSave` defers. So a Ctrl+S racing an in-flight PATCH re-posts the client's full local edge list and resurrects the edge the PATCH just dropped; both then bump `edit_revision` and one 409s. The comment at `pipelineStore.ts:52` already warns about drift between these two save paths.

---

### Style notes (`AGENTS.md`)

Minor, grouped here rather than inlined:

- **Comments narrating the diff** — `AGENTS.md`: *"A comment describes the code as it stands, not what moved, what it replaced, or which PR changed it."* See `patching.py:78` ("the pre-existing gap this closes"), `test_facade.py:234-238` ("Two independent moves since this was last pinned at 29 … Net: still 29"), `test_patching_handle_rewiring.py:4` ("until now") and `:149` ("same as before this feature existed"). `GetInputWidget.tsx:211-212` is the awkward case: it describes code the PR *deleted*, and now sits above an unrelated `return`.
- **Comments pre-rebutting alternatives** — `patching.py:80-82` ("neither is special-cased … just re-applies the same values afterward — harmless"), which the inline finding on line 112 shows is not actually true; also `build_state.py:244` ("unlike `dict.get(key, default)`") and `utils.tsx:103-108`.
- **Parametrize** — `test_patching_handle_rewiring.py:39/80/98/130` share an identical four-step body differing only in source type, params and expected surviving ids; two have byte-identical assertions. `AGENTS.md` asks for `pytest.mark.parametrize` with `pytest.param(..., id="...")` here.

---

### Also worth knowing

A few things I checked that turned out **fine**, recorded so nobody re-investigates them:

- `flow.errors` is never populated in `Pipeline.data`, so dropped edges leave no stale error keys behind.
- `update_nodes_from_data` does persist the changed `type`, and `compare_with_model` sees a type change — version diffing is not blind to it.
- `KeywordsWidget` and `NodeOutputs` both null-guard `params.keywords`, so removing the `GetInputWidget` seeding causes no client-side crash.
- `data.flowType` is stripped before persistence and `data.label` is never read by the frontend, so *not* updating the React Flow node type on a swap is inert today (all `can_add` schemas are `pipelineNode`). The missing `updateNodeInternals` is a separate matter and does still stand.

apps/pipelines/patching.py
```diff
@@ -52,9 +53,11 @@ def _apply_node_diff(flow: Flow, diff: NodeDiff) -> None:
     for node_id in diff.delete:
         node_map.pop(node_id, None)
 
-    # Update: replace in-place
+    # Update: replace in-place, following or dropping edges whose output handles moved
     for updated in diff.update:
+        previous = node_map.get(updated.id)
         node_map[updated.id] = updated
+        _rewire_edges(flow, previous, updated)
```
SmittieC: ### 🔴 Renaming a router keyword silently deletes that branch's edge

**Verified by running it.** This is the finding I'd most want a second opinion on, because it destroys user work with no warning and no undo step.

`handle_remap` matches old handles to new ones **by branch label**, so a rename reads as "branch removed" and the edge is dropped. That is sound for the v2 API. But wiring it in here also applies it to the builder, where `KeywordsWidget.updateKeyword` (`widgets.tsx:660`) fires on *every keystroke* and `autoSaveCurrentPipline` debounces 1s — so the PATCH arrives carrying `nodes.update` only, with no edge diff at all.

Driving `apply_pipeline_patch` directly on a `StaticRouterNode` with `keywords: ['SUPPORT','SALES']` and edges `e0 -> output_0`, `e1 -> output_1`:

```
rename "SUPPORT" -> "SUPPORT2":            surviving edges: [('e1','output_1')]   # e0 destroyed
one backspace mid-typing ("SUPPORT2"->"SUPPORT"): surviving edges: []             # destroyed
```

A pause longer than 1s in the middle of typing is enough. `_patchPipeline` (`pipelineStore.ts:471-481`) then replaces the store's edges from `response.data` **inside `withTemporalPaused`**, so the wire vanishes from the canvas mid-edit and leaves no undo entry to recover it. Typing the rest of the word does not bring it back.

The deletion also buys nothing here: a rename strands no edge, because `output_0` still exists and still points somewhere the user chose. Only **removal** and **reorder** actually need rewiring.

There is no test for the rename case — the new module covers drop, reorder, and both type-change directions.

**Suggested fix:** use positional identity on this path and keep label identity for the v2 API, or have `KeywordsWidget` commit on blur rather than on change.

apps/pipelines/patching.py
```diff
@@ -69,16 +72,45 @@ def _apply_node_diff(flow: Flow, diff: NodeDiff) -> None:
...
+    type change (#1452) the same way, since both are just an update whose handles differ before
+    and after -- neither is special-cased. Runs before the edge diff below, so a caller that also
+    sends its own correct edge diff for the same node (the v2 API always does) just re-applies the
+    same values afterward -- harmless. A deleted edge is the one thing that step must not bring
+    back; see _apply_edge_diff.
+    """
+    previous_data = previous.data if previous else None
+    if previous_data is None or updated.data is None:
+        return
+    before = output_handle_labels(previous_data)
```
SmittieC, line 101: ### 🟠 `patching.py` now contradicts its own module docstring

Line 4 of this file states: *"Never touches the database directly — the caller (the PATCH view) is responsible for persisting the merged graph."* That is no longer true.

`output_handle_labels` → `_router_output_map` → `node_class.model_validate(...)`, and `RouterNode` inherits `HistoryMixin`/`LLMResponseMixin`, whose validators call `ORMRepository().get_llm_provider_model()`. Note these two calls run **before** the `before == after` fast path on line 90 can skip anything.

**Verified by query counting.** A *position-only* drag of a single `RouterNode` through `apply_pipeline_patch`:

```
F5 QUERIES ON A POSITION-ONLY DRAG: 2
    SELECT "service_providers_llmprovidermodel"...
    SELECT "service_providers_llmprovidermodel"...
```

Zero before this PR, because the patch engine never instantiated a node class. It scales linearly with routers in a multi-select drag, and it all happens inside `select_for_update` on the pipeline row (`views.py:302-326`), lengthening the lock hold on an operation that should be pure in-memory bookkeeping.

There is a second-order risk too: `_router_output_map` catches only `(pydantic.ValidationError, PipelineNodeBuildError)`, so **any other** exception a validator raises now aborts the whole save from inside `transaction.atomic()` — a failure mode the patch engine previously could not have had.

**Suggested fix:** handles are a pure function of type and params, so gate the whole thing on `previous_data.type != updated.data.type or previous_data.params != updated.data.params`. That skips the common case (drags, selections, label edits) entirely and restores the docstring's claim for everything but a genuine param change.

apps/pipelines/patching.py
```diff
@@ -69,16 +72,45 @@ def _apply_node_diff(flow: Flow, diff: NodeDiff) -> None:
...
+    after = output_handle_labels(updated.data)
+    if before == after:
+        return
+    changed, deleted_ids = rewired_edges_for_node(flow.edges, updated.id, before, after)
+    if not changed and not deleted_ids:
+        return
+    changed_by_id = {edge.id: edge for edge in changed}
+    deleted = set(deleted_ids)
+    flow.edges = [changed_by_id.get(edge.id, edge) for edge in flow.edges if edge.id not in deleted]
```
SmittieC: ### 🟠 Ordering: an edge added in the same patch escapes the handle check

`_rewire_edges` runs from `_apply_node_diff`, which executes **before** `_apply_edge_diff`. So the rewiring only ever sees edges that already existed; anything the same patch adds lands afterwards with no handle validation, and can collide with an edge the rewiring just moved.

**Verified, both directions.**

Collision — router `['A','B']` with `e1` on `output_1`; patch removes keyword `'A'` (so `e1` legitimately moves to `output_0`) while adding a new edge on `output_0`:

```
[('e1','output_0'), ('eNew','output_0')]
```

Two wires from one branch — precisely the duplicate that `edge_editor.py`'s `_refuse_duplicate` exists to prevent.

Stranding — router `['A','B']` with `e0` on `output_0`; patch narrows to `['A']` while adding an edge on `output_1`:

```
[('e0','output_0'), ('new','output_1')]
```

`new` is permanently stranded on a handle the updated router no longer offers.

Both are reachable from the builder inside a single 1s autosave window — the debounce coalesces a keyword edit and a fresh connection into one PATCH.

**Suggested fix:** run the rewiring **after** `_apply_edge_diff`, over the merged edge set. That fixes this, fixes the sibling finding on line 112, and makes the new `if updated.id in edge_map` guard unnecessary.

apps/pipelines/patching.py
```diff
@@ -69,16 +72,45 @@ def _apply_node_diff(flow: Flow, diff: NodeDiff) -> None:
...
         edge_map.pop(edge_id, None)
 
-    # Update: replace in-place
+    # Update: replace in-place -- but never resurrect an id the node-update rewiring above
+    # already dropped from this same flow (a caller's edge diff can carry a stale value for
+    # an edge whose handle no longer exists once the node update lands; see
+    # _rewire_edges).
     for updated in diff.update:
-        edge_map[updated.id] = updated
+        if updated.id in edge_map:
```
SmittieC: ### 🟠 The guard only stops a *deleted* edge returning — a stale update still clobbers a *moved* one

The comment above correctly identifies the hazard, but the guard is narrower than the problem. It prevents resurrecting an id that `_rewire_edges` **deleted**; it does nothing when `_rewire_edges` **moved** an edge to a different handle and the caller's diff still carries the old value.

**Verified.** Keywords `['A','B']` → `['B','A']`, with an edge diff still carrying `e0` on `output_0` (exactly the stale-client case the comment anticipates). `_rewire_edges` correctly produces `e0 -> output_1` and `e1 -> output_0`; `_apply_edge_diff` then re-applies the stale `e0`:

```
[('e0','output_0'), ('e1','output_0')]
```

Both edges on `output_0`, and branch `'A'` has no wire at all. `test_an_explicit_edge_update_in_the_same_patch_does_not_resurrect_a_rewired_edge` covers only the delete case.

**Separately, this narrows a contract.** `EdgeDiff.update` was an upsert; it is now a silent no-op for a missing id, while the `add` branch three lines below still upserts. Verified: a patch carrying only `EdgeDiff(update=[<id not in the flow>])` now yields `edges == []` where it previously inserted, with a `200` and `dirty` cleared client-side — the caller is told it succeeded.

An id can go missing for reasons unrelated to `_rewire_edges`: `_apply_node_diff` culls edges attached to deleted nodes (line 72), and `diff.delete` pops ids earlier in this same function. So a patch that both deletes and updates an edge silently loses the update. No current caller triggers it — `edge_editor` only emits `add`/`delete`, and `plan_update`'s updates are copies of edges read out of `flow.edges` — so this is latent rather than live, but it fails silently when it does fire.

**Suggested fix:** reordering so the rewiring runs last (see the comment on line 97) preserves the upsert contract and makes this guard unnecessary. Failing that, track the ids `_rewire_edges` actually dropped and guard on *that* set rather than on membership.

apps/pipelines/build_state.py
```diff
@@ -172,6 +172,58 @@ def output_handles(node_type: str, params: dict, node_id: str, django_node: Node
...
+
+
+def handle_remap(before: dict[str, str | None], after: dict[str, str | None]) -> dict[str, str]:
+    """Where each handle a node used to offer has ended up, keyed by the handle it was.
+
+    A handle whose branch the edit removed is absent: its edge has nowhere to go. A rename counts
+    as a removal -- inheriting the old branch's target would wire the new one somewhere nobody
+    chose.
+    """
+    if _labels_are_distinct(before) and _labels_are_distinct(after):
```
SmittieC, line 187: ### 🟠 The two branches of `handle_remap` apply opposite policies to the same user action

The distinct-label path (line 187-189) **deletes** an edge whose branch was renamed. The duplicate-label fallback (line 193) **re-points** it onto whatever branch now occupies that position. Same user action, opposite outcome, three lines apart — and the second one silently redirects live traffic.

Duplicate keywords are genuinely storable: `update_nodes_from_data` writes params verbatim, and `ensure_keywords_exist` only raises during validation, which `_router_output_map` then falls back from to `model_construct`.

**Verified.** Router with `keywords: ['A','A','B']` and edges `e0`/`e1`/`e2` on `output_0`/`1`/`2`. The user fixes the duplicate to `['A','C','B']`:

```
[('e0','output_0','t0'), ('e1','output_1','t1'), ('e2','output_2','t2')]
```

`e1` still sits on `output_1`, which now serves branch `'C'`. Traffic the user deliberately wired to `'A'` is now routed to `'C'`, with nothing in the UI to indicate it moved.

The comment on 190-192 justifies positional fallback as the conservative choice ("which edge belongs to which of two identical branches is a guess"), which is fair as far as it goes — but the consequence is that fixing a duplicate keyword, the most likely reason a user would ever edit this state, quietly re-aims a branch.

Worth deciding what the intended policy actually is and applying it consistently. My own view is that positional is right for *both* paths on the builder side, which would also resolve the rename finding on `patching.py:60`.

apps/pipelines/models.py
```diff
@@ -619,13 +619,12 @@ def set_params(self, params):
 
     def update_from_params(self):
         """Callback to do DB related updates pertaining to the node params"""
-        from apps.pipelines.nodes.nodes import LLMResponseWithPrompt  # noqa: PLC0415 - circular: nodes.nodes→models
-
         self._sync_resource_fk_fields()
 
-        if self.type == LLMResponseWithPrompt.__name__:
+        custom_actions = self.params.get("custom_actions") or []
+        if custom_actions or self.custom_action_operations.exists():
```
SmittieC: ### 🟠 One DB round trip per node added to every graph save, and the gate now disagrees with its sibling

**Two separate problems with this line.**

**1. Performance.** `update_nodes_from_data` (line 198) calls `update_from_params()` on the instance returned by `Node.objects.update_or_create`, which carries no prefetch cache — so `NODE_RESOURCE_PREFETCHES` cannot serve this and `.exists()` always hits the database.

Measured on this branch: a 10-node `RenderTemplate` graph re-save issues exactly 10 `SELECT 1 ... FROM custom_actions_customactionoperation WHERE node_id = %s LIMIT 1` statements where it previously issued **0**. It scales 1:1 (30 nodes → 30 queries). This runs inside `transaction.atomic()` on the POST save and inside `select_for_update` on the PATCH. Note `QUERIES_UNDER_THE_LOCK = 29` cannot catch this — `_collect_node_data` maps every unwritten node to `None`, so that fixture reconciles a single node.

The old check was in-memory and free. Since `custom_actions` is declared on exactly one node class (`nodes.py:319`, `LLMResponseWithPrompt`), `self.has_parameter("custom_actions")` would restore that while still fixing the orphaned-rows bug this change is here for.

**2. The gate moved, but its siblings didn't.** Ownership of `CustomActionOperation` rows is now purely value-based, while `_get_version_details` (line 753) still reads `if self.type == LLMResponseWithPrompt.__name__ and self.params.get("custom_actions")`. Previously all three sites — row writes, `create_new_version`'s copy (line 600), and version display — agreed on one type.

Now any node whose params carry a truthy `custom_actions` gets rows written that version comparison **refuses to render** (its `case "custom_actions": continue` also skips the raw param), while `create_new_version` copies them type-agnostically. Reachable via a restored export, a `revert_to_version` onto a node whose type has since changed, or a client echoing params back.

The PR description says *"Added the missing `elif` branch"* — worth updating, since the change actually replaces the type check outright, which is a materially wider change than an added branch.

apps/pipelines/models.py
```diff
@@ -619,13 +619,12 @@ def set_params(self, params):
...
             custom_action_infos = []
-            for custom_action_operation in self.params.get("custom_actions") or []:
+            for custom_action_operation in custom_actions:
                 custom_action_id, operation_id = custom_action_operation.split(":")
```
SmittieC: ### 🟡 This unguarded unpack is now reachable from every node type

`custom_action_operation.split(":")` raises `ValueError` on any entry that doesn't contain exactly one colon — `"abc"` gives one element, `"a:b:c"` gives three, and both fail the two-value unpack.

Before this PR the enclosing branch was gated on `self.type == LLMResponseWithPrompt.__name__`, so the only values reaching here came from that node's own multiselect widget. With the gate now value-based (see the comment on line 625), any node type carrying a truthy `custom_actions` reaches this line.

That matters because of *where* it runs: inside `update_nodes_from_data`, within `transaction.atomic()`. A malformed entry from a hand-edited export or an older payload shape becomes an **unhandled 500 on save**, rolling back the user's whole graph, rather than a reported validation error against that one node.

**Suggested fix:** `str.partition(":")` with a skip on a missing separator, or gate on `self.has_parameter("custom_actions")` so only the type that defines the field can reach this loop.

assets/javascript/apps/pipeline/PipelineNode.tsx
```diff
@@ -62,6 +62,23 @@ export function PipelineNode(nodeProps: NodeProps<NodeData>) {
     }));
   };
 
+  // Other user-addable types this node could become (#1452). Start/End/Passthrough are never
+  // offered, in either direction: a node whose own schema isn't `ui:can_add` gets no control at
+  // all, and this list excludes them as targets the same way ComponentList's palette does.
+  const changeableTypes = nodeSchema["ui:can_add"]
```
SmittieC: ### 🟡 The `ui:can_add` gate hides this control on exactly the nodes that most need it

`NodeSchema.update_metadata_fields` (`nodes/base.py:606-608`) forces `can_add = False` whenever a type is `deprecated` or `removed`. Because this line gates the whole control on `nodeSchema["ui:can_add"]`, a node *of* a deprecated type gets no swap button at all.

Enumerating the live schemas, the non-addable `pipelineNode` types are:

```
BooleanNode   (deprecated)
LLMResponse   (deprecated)
Passthrough
AssistantNode (deprecated + removed)
```

`AssistantNode`'s own stub message (`node_metadata.py:281`) reads: *"OpenAI retired the Assistants API on 26 August 2026. **Delete this node and use an LLM node instead.**"* That is the delete-and-recreate workflow this PR exists to replace — and it is one of the three types where the new button never appears.

The comment above explains the gate in terms of Start/End/Passthrough and doesn't mention the deprecated case, which suggests this may be incidental rather than intended. Gating the *source* on `!schema["ui:removed"]` while keeping `ui:can_add` on the *target* list would let users migrate off deprecated nodes without ever letting them create new ones.

**Minor, same line:** this `Array.from` + `filter` + `sort(localeCompare)` has no `useMemo`, although `nodeSchemas` is a module-level cache that never changes. React Flow re-renders nodes on selection and throughout drags, so this rebuilds and re-sorts the full schema list canvas-wide during interactions that have nothing to do with node types. It also duplicates the identical sorted list at `ComponentList.tsx:17`, so the two menus can drift.

assets/javascript/apps/pipeline/PipelineNode.tsx
```diff
@@ -62,6 +62,23 @@ export function PipelineNode(nodeProps: NodeProps<NodeData>) {
...
+    ? Array.from(getCachedData().nodeSchemas.values())
+        .filter((schema) => schema["ui:can_add"] && schema.title !== data.type)
+        .sort((a, b) => a["ui:label"].localeCompare(b["ui:label"]))
+    : [];
+
+  const changeNodeType = (newSchema: JsonSchema) => {
```
SmittieC, line 76: ### 🟠 `changeNodeType` swaps the node's entire handle set without telling React Flow

`NodeOutputs` derives its `<Handle id>` set from `data.type`/`data.params` — `"output"` for a plain node, `"output_0..n"` for `RouterNode`/`BooleanNode`/`StaticRouterNode` — so converting an LLM node to a router changes both the **count** and the **ids** of its handles.

React Flow v11 caches `handleBounds` per node and only recomputes on mount or a dimension change. That is exactly why `KeywordsWidget` calls `updateNodeInternals` in both `addKeyword` (`widgets.tsx:657`) and `deleteKeyword` (`:685`). `changeNodeType` is the only handle-changing mutator in the app that omits it, so existing edges anchor to a handle that no longer exists, and a drag from a newly-created handle may not start a connection until the canvas remounts.

Unlike `deleteKeyword`, this also calls no `setEdges`, so the canvas renders edges on dead handles for the full 1s debounce plus the server round trip before the response corrects them.

**Suggested fix:** take `useUpdateNodeInternals()` at the top of the component (as `KeywordsWidget` does) and call `updateNodeInternals(id)` after `setNode` here.

assets/javascript/apps/pipeline/PipelineNode.tsx
```diff
@@ -62,6 +62,23 @@ export function PipelineNode(nodeProps: NodeProps<NodeData>) {
...
+    setNode(id, produce((next) => {
+      next.data.type = newSchema.title;
+      next.data.label = newSchema["ui:label"];
+      next.data.params = buildTypeChangeParams(newSchema, data.params);
```
SmittieC: ### 🟠 Reads the render-time prop instead of the immer draft, dropping a concurrent edit

`data` is the `NodeProps` value React last rendered. `setNode` (`pipelineStore.ts:151-156`) resolves its function argument against `get().nodes.find(...)` — the **live store**. zustand's `set()` is synchronous but the React re-render that follows is not, so there is a real window where `next` is current and `data` is stale.

This matters more than it might look, because `buildTypeChangeParams` reads exactly two keys off that stale object — `name` and `color` — and both have per-interaction writers in this very component:

- `updateParamValue` (line 48) fires on **every keystroke** in the node's name field
- `changeNodeColor` (line 60) fires on a colour pick

So: rename a node or recolour it, then click a type in the change-type menu in the same frame, and that edit is silently thrown away. Both sibling handlers write through `next.data.params`; only this new one reaches outside the draft.

```suggestion
      next.data.params = buildTypeChangeParams(newSchema, next.data.params);
```

assets/javascript/apps/pipeline/utils.tsx
```diff
@@ -79,6 +79,43 @@ export function getDocumentationLink(schema: JsonSchema) {
...
+export function getDefaultParamValues(schema: JsonSchema): NodeParams {
+  const {defaultValues} = getCachedData();
+  const defaults: NodeParams = {name: ""};
+  for (const name in schema.properties) {
+    // Every real node schema declares "name" as one of its own properties (it's a field on the
+    // shared base node), but it's never given a schema or cached default here -- it's set by
+    // the caller (a generated id on add, the preserved name on a type change).
+    if (name === "name") continue;
+    const property = schema.properties[name];
+    defaults[name] = [property.default, defaultValues[name]].find((value) => value !== undefined && value !== null) ?? null;
```
SmittieC: ### 🟠 `null` for `default_factory` fields — a fresh or converted router persists `keywords: null`

Pydantic emits no `"default"` for a `default_factory` field, and `get_node_default_values` (`node_metadata.py:251`) returns only `llm_provider_id`/`llm_provider_model_id`. So `keywords` finds neither source and falls through to `?? null`.

That null then reaches the server, and the two sides disagree about what the node offers. **Verified:**

```
output_handles('StaticRouterNode', {'keywords': None}, ...) == []
```

while `NodeOutputs.getOutputNames` still draws `output_0` (`Math.max(1, params.keywords?.length || 1)`). The canvas offers a handle the server does not believe exists — so an edge drawn from it reads as stranded, and the v2 wiring API refuses the same connection with `NoOutputHandles.NO_BRANCHES`.

Validation also degrades: `keywords: null` fails pydantic with *"Input should be a valid list"*, where the `keywords: ['']` that the removed `GetInputWidget` seeding used to produce gave the domain-level *"Keywords cannot be empty"*.

This is the root cause the PR patches downstream in `_router_output_map`'s new `or []`. Defaulting array-typed properties to `[]` here instead would fix all three symptoms in one place — the null param, the removed seeding, and the `or []` coercion. The codebase already carries five hand-written `None -> []` coercions (`nodes.py:385-395`) for the same underlying reason.

**Suggested fix:** when `property.type === "array"`, default to `[]` rather than `null`.

assets/javascript/apps/pipeline/utils.tsx
```diff
@@ -79,6 +79,43 @@ export function getDocumentationLink(schema: JsonSchema) {
...
+ * Every type-specific param resets to the new type's own defaults, never carried across by
+ * matching field names -- a name shared between two schemas is not proof it means the same
+ * thing in both (see `NodeUpdateSerializer` on the v2 API, which refuses a type change in place
+ * for exactly this reason). `name` and `color` are the only survivors: `name` is UI identity
+ * every schema declares, and `color` is UI-only state that is not a schema field on any node
+ * type at all, so it is carried separately rather than lost to the reset.
+ */
+export function buildTypeChangeParams(newSchema: JsonSchema, currentParams: NodeParams): NodeParams {
+  const params = getDefaultParamValues(newSchema);
+  params.name = currentParams.name;
```
SmittieC: ### 🟡 Unconditional assignment can overwrite the deliberately seeded empty string with `undefined`

`getDefaultParamValues` seeds `{name: ""}` and then **skips** the schema's own `name` property (line 93) specifically to protect that seed. This line then assigns over it with no fallback, so if `currentParams.name` is absent, `params.name` becomes `undefined`.

That is worse than it sounds, because `diffPipeline`'s `serializeNode` hands `data.params` straight to `JSON.stringify`, which **omits** `undefined` keys entirely. The PATCH body then carries no `name` at all, and the required `name: str` on `BasePipelineNode` reports *"Field required"* rather than the empty-name error the builder UI is designed around.

Every current path sets `name` (`addNode` assigns `data.params["name"] = newId`, and `to_flow_node` always includes it), so this is latent rather than live — but it is reachable for any node whose params were written by a path that did not go through `addNode`, and it quietly defeats the protection added one function above.

```suggestion
  params.name = currentParams.name ?? "";
```

apps/pipelines/tests/test_build_state.py
```diff
@@ -123,6 +123,19 @@ def test_router_with_dangling_provider_model_still_reports_handles(self):
             {"handle": "output_1", "label": "B"},
         ]
 
+    def test_router_with_none_keywords_still_reports_handles(self):
```
SmittieC: ### 🟡 This test asserts the opposite of its own name, and pins the wrong behaviour as approved

The name says *"still reports handles"*; the assertion on line 137 is `== []`, i.e. reports **no** handles. Its direct sibling `test_router_with_dangling_provider_model_still_reports_handles` (line 112) uses the identical phrasing and asserts two real handles, so the convention is established and this one breaks it.

This is more than a naming slip. `keywords: None` is exactly the state a freshly type-changed or freshly dragged router lands in (see the comment on `utils.tsx:95`), where the server reports zero output handles while `NodeOutputs` still draws `output_0`. That client/server divergence is now locked in by a **green test whose name reads as an endorsement** — the next person to touch this will reasonably conclude it was the intended design.

The comment body is accurate and useful; it's specifically the name that misleads.

**Suggested fix:** decide first whether zero handles is the behaviour you actually want here. If yes, rename to `test_router_with_none_keywords_reports_no_handles`. If no — and I'd argue it isn't, given the canvas draws a handle the server denies — make it assert the behaviour the feature needs, and fix the `null` at its source in `getDefaultParamValues`.

apps/pipelines/patching.py
Davemaina1: Positional fixes the rename but breaks a plain reorder, an edge wired to keyword 'a' ends up on whatever keyword now sits at that index. Haven't picked a policy, want your take before I do.

apps/pipelines/patching.py
Davemaina1: Gated the handle recompute on type/params actually changing, so a drag or selection never touches output_handle_labels. Test asserts zero calls on a position-only update.

apps/pipelines/patching.py
Davemaina1: Rewiring runs once now, after the edge diff, over the merged set. Added the collision and stranding cases as tests.

apps/pipelines/patching.py
Davemaina1: Reordering removed the need for the guard, EdgeDiff.update is a plain upsert again. Matches what it was before this feature touched the file.

apps/pipelines/build_state.py
Davemaina1: Same open question as patching.py:60. Positional fixes this but reopens the reorder case.

apps/pipelines/models.py
Davemaina1: Both fixed, differently than suggested. has_parameter gating the whole block breaks an existing test (clearing orphaned rows needs the exists() check to still run), so only the unpack loop is gated. For the query cost, update_nodes_from_data now checks which nodes have rows once for the whole batch instead of once per node. The _get_version_details swap to has_parameter is currently a no-op though, custom_actions is on exactly one class with no subclasses, so it agrees with the old type check for every type that exists today. Kept it since it ties both sites to the same check, not because it fixes anything live.

apps/pipelines/models.py
Davemaina1: Went with has_parameter. Test covers a malformed entry on a RenderTemplate node, ignored instead of throwing.

assets/javascript/apps/pipeline/PipelineNode.tsx
Davemaina1: Not with !ui:removed, that still misses AssistantNode since it's deprecated and removed. Used ui:can_add || ui:deprecated || ui:removed, covers all three types you listed. Verified live. Also memoized it and pulled the sorted list into a shared getCanAddNodeSchemas() so this and ComponentList can't drift.

assets/javascript/apps/pipeline/PipelineNode.tsx
Davemaina1: Matches KeywordsWidget's own pattern, useUpdateNodeInternals() at the top, called right after setNode.

assets/javascript/apps/pipeline/PipelineNode.tsx
Davemaina1: Used your diff as is.

assets/javascript/apps/pipeline/utils.tsx
Davemaina1: Array-typed properties default to [] now. Also renamed the Python test at test_build_state.py:126, that null case is the defensive fallback for an older client now, not the everyday path.

assets/javascript/apps/pipeline/utils.tsx
Davemaina1: Used your diff as is.

apps/pipelines/tests/test_build_state.py
Davemaina1: Renamed to test_router_with_none_keywords_reports_no_handles, per your first option.

apps/pipelines/patching.py
Davemaina1: The label-match/rename-drops-edge policy isn't new; apps/api/v2/pipeline_edit already does this for a param-only edit. What this PR added is reusing that same policy for the PATCH path too, since the type-change feature needs some handle reconciliation, and duplicating the logic instead of sharing it would be worse.

So going positional on this path only would mean the same edit, drop or rename a keyword is safe via one save path and destructive via the other, depending on which one happened to fire. That's a bigger inconsistency than the original bug.

Since it's already your call for the v2 API, should we keep both paths on label matching, or actually diverge them?

SmittieC (commented):
Looking at the amount of BE work involved in this change, I am wondering if we let the UI execute similar steps that a human might have, if we can't simplify a lot of these. Perhaps we wouldn't need any python code changes.

Here's my thinking:
When a node type changes, we reset the node in anycase, so we can:

1. delete the incoming edge + outgoing edges, but keep track of the source and target node(s). In a router's case, the first target should suffice.
2. delete the node, but keep track of its current position
3. Place the new node at this position
4. Create an edge between the source node and current node
5. Create an edge between the current node and the target node (or first target if the previous node was a router).

Thoughts on this approach? Sorry, I know you have been doing a lot of work already

Davemaina1:
Thanks for chiming in.

I agree; this is simpler and safer.

I went and checked the two things I was worried about. Custom actions and resource refs already get wiped on a type change; buildTypeChangeParams resets every type param to the new schema's defaults no matter what the old type had, so update_from_params clears them out either way. This is quite similar to your proposal.

Same with first target only, router keywords start empty on a fresh type so every edge already gets dropped on a type change involving a router. Keeping the first one is more than we do now.

The only catch is on the frontend, this needs to be one store update, not delete then add then reconnect as three separate calls. Otherwise undo breaks into pieces and autosave fires three times instead of once. deleteNode already bundles node and edges for that reason.

Most of the type-change handling in patching.py and build_state.py becomes dead weight if we do this. This is fine by me; I'll pull it out once the new path lands.

SmittieC:
@Davemaina1 I'm just checking in to see if you're going to look at the frontend-only change?

Davemaina1:
@SmittieC Yes, doing it. It's simpler and safer than what's built, same or better edge preservation, custom actions and resource refs already reset on any type change today so there's no regression there either, and it removes the whole class of bugs this review has been catching since there's no new diffing logic left to have them in. Picking it up now.

apps/pipelines/patching.py
Davemaina1: Decided this one. A reorder always moves at least two branches at once, a rename only ever changes one. So when the same branches survive an edit and exactly one changed label, that's a rename and the edge stays on it. Everything else, add, remove, a real reorder, duplicate labels, works the same as before.

Fixed in handle_remap, tests added on both the PATCH path and the v2 API's own keyword tests.

Also found graph_editor.py's own _handle_remap copy was dead, plan_update never called it, apply_pipeline_patch already covers it. Removed it.

apps/pipelines/build_state.py
Davemaina1: Same fix resolves this one. The duplicate-label fallback was already positional, the distinct-label path now agrees with it for a rename specifically, since a rename never actually needs label disambiguation, only positional identity. Reorders and removals still go through label matching on both paths.

assets/javascript/apps/pipeline/PipelineNode.tsx
Davemaina1: Update: this doesn't apply anymore. Type changes went through a rewrite since, changeNodeType is now delete-and-recreate rather than an in-place mutation, so there's no existing node whose handle set changes under it, a fresh node just mounts with its own handles. useUpdateNodeInternals is gone from this file entirely.

apps/pipelines/tests/test_build_state.py
Davemaina1: Update: this test moved during the upstream merge, output_handles logic relocated onto NodeType and the whole TestNodeOutputHandles class, this test included, went with it into test_node_type.py. Checked every case that class covered against what landed there, all had an equivalent or better replacement except this one, keywords:None. That gap wasn't just missing coverage, output_handles actually crashes on it now (TypeError, enumerate(None)). Fixed and retested.

Davemaina1:
@SmittieC I have made the frontend change and verified it. Delete old node and edges, add fresh one at the same position, reconnect source and first target, name and color carry over.

I also fixed a real regression from the NodeType rebase, the router with keywords:None was crashing, and pulled out some dead code in graph_editor.py the merge brought back.

SmittieC (commented):
I don't think the backend changes is still necessary anymore? I got claude to look at what I had in mind. See https://github.com/dimagi/open-chat-studio/pull/4595

Davemaina1:
Agreed, your PR should carry this. Should I close this?

Two things surfaced while working on this that are independent of type-change and still bugs: renaming a branch label on an existing Router or StaticRouter node currently drops its edge instead of keeping it, and a router with keywords:None crashes output_handles (a regression from the NodeType move). Want us to open a PR for those?

One thing on your implementation: it assigns the swapped node a fresh name, so any node referencing the old one's output by name (temp_state.outputs.<name>) will silently stop resolving after a type change.

SmittieC:
> Agreed, your PR should carry this. Should I close this?

Sure. Thanks for this PR though.


> Two things surfaced while working on this that are independent of type-change and still bugs: renaming a branch label on an existing Router or StaticRouter node currently drops its edge instead of keeping it, and a router with keywords:None crashes output_handles (a regression from the NodeType move). Want us to open a PR for those?
> 
> One thing on your implementation: it assigns the swapped node a fresh name, so any node referencing the old one's output by name (temp_state.outputs.) will silently stop resolving after a type change.

I'll look at these, thanks

PR #4477, opened 2026-09-10
Add undo/redo to the pipeline builder

### LINKED ISSUE #2006: Undo functionality for pipeline building
As a user I want to easily be able to "undo" edits on a pipeline that are currently not possible, for example, removing a node.

Davemaina1:
### Product Description
Adds undo/redo to the pipeline builder. Ctrl+Z / Ctrl+Shift+Z (or Ctrl+Y), plus toolbar buttons next to the existing zoom/fit-view controls, undo and redo node and edge edits: add, remove, move, connect, and param/label changes.

### Technical Description
Wraps the pipeline store's `nodes`/`edges` in zundo's `temporal` middleware, scoped via `partialize` so only those two keys are tracked.

History is scoped to real user edits, not every store write:
- `resetFlow` and the autosave/save response handlers pause tracking around themselves, reloading a pipeline or reconciling a server response isn't a user edit.
- `onNodesChange`/`onEdgesChange` filter out React Flow's own bookkeeping changes: a `dimensions` change as each node first measures itself, a `select` change on every click, and a `position` change with no `position` value, which React Flow sends on a plain click as its own drag-end signal.
- `equality: shallow` is needed because zundo otherwise pushes a history entry on every `set()` call regardless of whether `nodes`/`edges` actually changed.

`deleteNode` now issues one `set()` call instead of two (edges then nodes), since removing a node and its edges is one user action and undo history is recorded per `set()` call, splitting it made a single undo land on a broken intermediate state instead of the original.

`handleSet` is throttled so a node drag collapses into one history entry instead of one per pointer move.

zundo's `undo()`/`redo()` set `nodes`/`edges` through the store's raw `set()`, bypassing `setNodes`/`setEdges` the only places that call `autoSaveCurrentPipline()`. An undo restored the canvas but never told the server, so a reload right after would lose it. Added `undoLastChange`/`redoLastChange` store actions that call zundo's `undo()`/`redo()` then `autoSaveCurrentPipline()` explicitly, and pointed the hotkeys and toolbar buttons at them. Both guard on `readOnly`, matching every other mutator in this store.

Blast radius: touches `pipelineStore.ts`, which every pipeline builder interaction depends on. The scoped mutators are the same ones the UI already calls for node/edge changes, so wrapping them changes when history is recorded, not what they do. Verified against the full pipeline test suite and a manual pass in the browser: undo during and after an in-flight autosave, and undo triggering a real PATCH that survives a reload, including when the undo happens while a prior save is still in flight, which the existing deferred-save handling (`pendingSave`/`settleDeferredSave`) covers without changes.

Related but separate: `Pipeline.revert_to_version` already exists (reached through a chatbot's published version history) and resets a working pipeline to a prior published snapshot. Different granularity and trigger from this undo stack, so it doesn't overlap.

### Issue Link
Closes #2006

### Demo
Delete a node, click Undo (restores it and its edge), click Redo (re-deletes it):

pipeline-undo-redo-demo-final

### Docs and Changelog

New user-facing feature: undo/redo (Ctrl+Z / Ctrl+Shift+Z, or Ctrl+Y) in the pipeline builder canvas, with toolbar buttons next to the existing zoom/fit-view controls.

### Operator Impact

### DIFF
--- a/assets/javascript/apps/pipeline/Pipeline.tsx
+++ b/assets/javascript/apps/pipeline/Pipeline.tsx
@@ -1,9 +1,11 @@
 import "reactflow/dist/style.css";
 import "./styles.css"
 import React, {useCallback, useState, useEffect} from "react";
+import {useStore} from "zustand";
 import ReactFlow, {
   Background,
   BackgroundVariant,
+  ControlButton,
   Controls,
   EdgeTypes,
   FitViewOptions,
@@ -55,6 +57,8 @@ export default function Pipeline() {
   const currentPipeline = usePipelineStore((state) => state.currentPipeline);
   const autoSaveCurrentPipline = usePipelineStore((state) => state.autoSaveCurrentPipline);
   const savePipeline = usePipelineStore((state) => state.savePipeline);
+  const canUndo = useStore(usePipelineStore.temporal, (state) => state.pastStates.length > 0);
+  const canRedo = useStore(usePipelineStore.temporal, (state) => state.futureStates.length > 0);
   const { nodeSchemas } = getCachedData();
 
   const editingNode = useEditorStore((state) => state.currentNode);
@@ -130,6 +134,14 @@ export default function Pipeline() {
 
   useHotkeys(["backspace", "delete"], handleDelete);
   useHotkeys("ctrl+s", () => manualSaveCurrentPipeline(), {preventDefault: true});
+  useHotkeys(["ctrl+z", "meta+z"], () => {
+    if (readOnly) return;
+    usePipelineStore.temporal.getState().undo();
+  }, {preventDefault: true});
+  useHotkeys(["ctrl+shift+z", "meta+shift+z", "ctrl+y"], () => {
+    if (readOnly) return;
+    usePipelineStore.temporal.getState().redo();
+  }, {preventDefault: true});
 
   const onSelectionChange = useCallback(
     (flow: OnSelectionChangeParams): void => {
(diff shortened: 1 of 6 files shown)


SmittieC (commented):
@Davemaina1 this looks good. I noticed that when I undo, it doesn't trigger an auto-save

Davemaina1:
Undo/redo write straight to the store's raw set(), skipping the autosave trigger entirely. Fixed in 90d8063.

snopoke (commented):
This looks good but I think we can clean up the `pause` / `resume` flow a bit. There are two approaches we can take:

### Option 1: Use the  [using](https://www.totaltypescript.com/typescript-5-2-new-keyword-using) keyword

```typescript
const temporalPause = (shouldPause = true): Disposable => {
  const temporal = usePipelineStore.temporal.getState();
  
  if (shouldPause) {
    temporal.pause();
  }

  return {
    [Symbol.dispose]() {
      if (shouldPause) {
        temporal.resume();
      }
    }
  };
};

const shouldPause = true;

{
  using _ = temporalPause(shouldPause);

  // 🪄 Arbitrary code goes here
  // If shouldPause is true: it pauses before and resumes after.
  // If shouldPause is false: it runs smoothly without pausing.
}
```

### Option 2: A function wrapper
```typescript
function withTemporalPaused<T>(shouldPause: boolean, callback: () => T): T {
  const temporal = usePipelineStore.temporal.getState();
  
  if (shouldPause) {
    temporal.pause();
  }
  
  try {
    return callback();
  } finally {
    if (shouldPause) {
      temporal.resume();
    }
  }
}

const shouldPause = true;

withTemporalPaused(shouldPause, () => {
  // 🪄 Arbitrary code goes here
});
```

Davemaina1:
@snopoke, I went with the function wrapper. No `using` anywhere in this codebase yet, and it would need a tsconfig lib bump (`ESNext.Disposable`) to typecheck, felt like too much to bundle into an undo/redo PR.

Added a try/finally around the callback too: without it, a throw mid-callback leaves tracking paused for the rest of the session. In 4f18729.

snopoke (approved):

SmittieC (approved):

PR #4478, opened 2026-09-10
Fix pathological join fan-out in get_experiments_for_display

### LINKED ISSUE #4475: Participant detail page takes ~70s: get_experiments_for_display query is pathological and run twice
## Problem

A production request to the participant detail page (`participants:single-participant-home`) took **69.0s** with 107 queries. A single query, run twice, accounts for **67.3s** of that. Everything else on the page totals about 0.5s.

Source: django-silk capture, 2026-09-10 12:19 UTC, `GET /a/Vaccine_Coach/participants/42352`.

| | count | time |
|---|---|---|
| `Participant.get_experiments_for_display()` | 2 | 33.67s + 33.65s |
| all other queries | 105 | ~0.53s |

## Cause 1 — the "experiments for display" query is run twice

`single_participant_home_context` (`apps/participants/views.py:171`) already has the experiment list from `_get_participant_and_chatbots` (`views.py:77`). It then calls `_schedules_panel_context` (`views.py:116`), which calls `participant.get_schedules_for_experiments(as_dict=True, include_inactive=True)` with no `experiment_id`, and that method rebuilds the same list via `get_experiments_for_display()` (`apps/experiments/models.py:1229`).

Introduced by a90e98e55 (merge of `get_schedules_for_all_experiments` into `get_schedules_for_experiments`). The schedules-panel htmx view has the same shape, so the Schedules tab pays it too.

Fix: let `get_schedules_for_experiments` accept the already-loaded experiments (or a list of ids) instead of recomputing them. Halves the page time on its own.

## Cause 2 — the query itself is pathological

Generated SQL (`apps/experiments/models.py:1179-1200`):

```sql
SELECT DISTINCT experiments_experiment.*,
  (SELECT created_at FROM experiments_experimentsession WHERE participant_id = 42352 ORDER BY 1 LIMIT 1) AS joined_on,
  (SELECT m.created_at FROM chat_chatmessage m
     JOIN chat_chat c ON m.chat_id = c.id
     JOIN experiments_experimentsession s ON c.id = s.chat_id
    WHERE s.experiment_id = experiments_experiment.id AND s.participant_id = 42352 AND m.message_type = 'human'
    ORDER BY 1 DESC LIMIT 1) AS last_message
FROM experiments_experiment
LEFT OUTER JOIN experiments_experimentsession ON experiments_experiment.id = experiments_experimentsession.experiment_id
WHERE experiments_experimentsession.participant_id = 42352
   OR experiments_experiment.id IN (SELECT experiment_id FROM experiments_participantdata WHERE participant_id = 42352)
ORDER BY experiments_experiment.name ASC
```

Silk captured the plan (`EXPLAIN`, abridged; the full `EXPLAIN ANALYZE` is in the first comment):

```
Unique  (cost=3150543..3159116 rows=118245)
  ->  Sort  (rows=118245)
        ->  Hash Right Join  (cost=743..3123157 rows=118245)
              Hash Cond: (experimentsession.experiment_id = experiment.id)
              Filter: ((experimentsession.participant_id = 42352) OR (hashed SubPlan 3))
              ->  Seq Scan on experiments_experimentsession  (rows=236480)
              ->  Hash  ->  Seq Scan on experiments_experiment  (rows=7661)
              SubPlan 2   -- last_message, cost 26 per evaluation
                ->  Limit -> Sort -> Nested Loop (session by participant+experiment -> chat -> chatmessage by chat_id+message_type)
              SubPlan 3   -- participantdata lookup, hashed once
```

The mechanism:

1. **`LEFT JOIN` + `OR` defeats the participant filter.** `get_experiments_queryset` filters with `Q(sessions__participant=self) | Q(id__in=<participantdata>)`. The join runs against every session row (236k seq scan) and the participant predicate is applied only afterwards as a post-join `Filter`.
2. **The `OR` branch keeps every session of the participant's experiments.** For any experiment the participant has a `ParticipantData` row for, `experiment.id IN (...)` is true on *every* joined row of that experiment, i.e. for every session of every participant on that chatbot. The planner estimates 118,245 surviving rows; `EXPLAIN ANALYZE` (see comment below) measured 22,367, one per session on the chatbot. `DISTINCT` collapses them afterwards, but not before step 3.
3. **The correlated `last_message` subquery runs once per surviving row.** 22,367 evaluations at ~1.76ms each is the entire 33–40s. Each evaluation itself is cheap and well indexed; the count is the problem.
4. **`joined_on` is not correlated at all** (InitPlan, evaluated once). Harmless, but it does not need to be a subquery.

So the fix is structural, not an index: the per-experiment annotation must be driven from the participant's own sessions, not computed over a join that fans out to the whole chatbot's session table.

Scrubbed Silk export (headers, cookie, team slugs, membership/flag queries and the widget UUID removed; SQL, tracebacks and plans intact): https://gist.github.com/snopoke/7154b9ae6731095dd755d9cb21cf4736

## Proposed work

1. **Stop the double call** (Cause 1). Pass experiments into `get_schedules_for_experiments`.
2. **Rewrite `get_experiments_queryset`** to avoid the join: collect experiment ids from the participant's sessions and `ParticipantData` (two indexed lookups by `participant_id`, or a `UNION`), then `Experiment.objects.get_all().filter(id__in=...)`. No `LEFT JOIN`, no `DISTINCT`.
3. **Replace the correlated `last_message` subquery** with one grouped query driven from the participant side, e.g. `ChatMessage.objects.filter(chat__experiment_session__participant=p, message_type="human").values("chat__experiment_session__experiment_id").annotate(Max("created_at"))`, merged in Python. Alternatively use `ExperimentSession.last_activity_at`, which is already written only on participant messages (`models.py:1509`) and indexed, via `Max("sessions__last_activity_at")` filtered to the participant.
4. **Compute `joined_on` once** rather than as a per-row annotation (or `Min("sessions__first_activity_at")`).
5. Re-run `EXPLAIN (ANALYZE, BUFFERS)` on prod for participant 42352 after the change and add a query-count assertion to the participant view tests.

## Secondary finding (out of scope here, may deserve its own issue)

`django_tables2.TemplateColumn` renders each cell with `get_template(...).render(context, request=request)`, which re-runs every context processor per cell. On this page that produced 29 identical copies of the `user_teams` membership query (`apps/users/context_processors.py:19`), ~146ms. This affects every table with template columns, not just this page. Options: pass `request=None` via a project `TemplateColumn` subclass, or make the `user_teams` / `unread_notifications_count` processors memoise on the request.

The 15 `UserNotificationPreferences` + 15 `COUNT` queries are `unread_notifications_count_all_teams` on a cold cache for a user in 15 teams, about 150ms; expected behaviour, noted for completeness.

## Related

- #4231 (participant page redesign, closed) listed "improve page performance" as a goal
- #559 (closed) earlier report that this page was slow

preston-56:
### Product Description
The participant detail page no longer times out. Query time drops from ~70s in production to milliseconds, regardless of how many other participants use the same chatbot.

### Technical Description
`get_experiments_queryset` joined across every session on a chatbot; an `id__in` `OR` condition kept unrelated session rows alive as join matches. `get_experiments_for_display` then ran a correlated `last_message` subquery per surviving row, once per session on the chatbot instead of once per experiment, and was called twice per page load.

Fix: resolve experiment ids via two small indexed lookups instead of a join, compute `last_message` as one grouped `Max()`, compute `joined_on` once, and let `get_schedules_for_experiments` accept an already-loaded `experiments` list to avoid the duplicate call.

Closes #4475

### Migrations

### Demo
`get_experiments_for_display()` against 805 sessions (800 from other participants): 6.5ms, down from 33-40s in production at 22,367 sessions.

Full page load, same dataset: 1.3s, down from ~70s in production.

Added a query-count regression test guarding that the page's query count doesn't scale with unrelated sessions on the chatbot.

### Docs and Changelog

### Operator Impact

### DIFF
--- a/apps/experiments/models.py
+++ b/apps/experiments/models.py
@@ -21,6 +21,7 @@ from django.db.models import (
     Case,
     Count,
     F,
+    Max,
     OuterRef,
     Q,
     Subquery,
@@ -1176,28 +1177,46 @@ class Participant(BaseTeamModel):
         )
         return f"{url}#{experiment.id}"
 
-    def get_experiments_for_display(self):
-        """Used by the html templates to display various stats about the participant's participation."""
-        exp_scoped_human_message = ChatMessage.objects.filter(
-            chat__experiment_session__participant=self,
-            message_type="human",
-            chat__experiment_session__experiment__id=OuterRef("id"),
-        )
-        last_message = exp_scoped_human_message.order_by("-created_at")[:1].values("created_at")
-        joined_on = self.experimentsession_set.order_by("created_at")[:1].values("created_at")
-        return (
-            self.get_experiments_queryset(include_archived=True)
-            .annotate(
-                joined_on=Subquery(joined_on),
-                last_message=Subquery(last_message),
+    def get_experiments_for_display(self) -> list[Experiment]:
+        """Used by the html templates to display various stats about the participant's participation.
+
+        Each returned `Experiment` carries two extra attributes: `joined_on` (this participant's
+        single earliest session, the same value on every experiment, not per-experiment) and
+        `last_message` (their latest human message on that specific experiment, or `None`).
+
+        Computed as three small, independently-indexed queries (which experiments, joined_on,
+        last_message per experiment) rather than one query joined across every session on each
+        chatbot. That join let one `id__in` participant-data match keep every session row of the
+        whole chatbot alive as a join partner, so a correlated per-row subquery for `last_message`
+        ran once per session on the chatbot instead of once per experiment (#4475).
+        """
+        experiments = list(self.get_experiments_queryset(include_archived=True))
+        if not experiments:
+            return experiments
+
+        joined_on = self.experimentsession_set.order_by("created_at").values_list("created_at", flat=True).first()
+        last_message_by_experiment = dict(
+            ChatMessage.objects.filter(
+                chat__experiment_session__participant=self,
+                message_type="human",
+                chat__experiment_session__experiment_id__in=[e.id for e in experiments],
             )
-            .distinct()
+            .values("chat__experiment_session__experiment_id")
+            .annotate(last_message=Max("created_at"))
+            .values_list("chat__experiment_session__experiment_id", "last_message")
         )
+        for experiment in experiments:
+            experiment.joined_on = joined_on
+            experiment.last_message = last_message_by_experiment.get(experiment.id)
+        return experiments
 
     def get_experiments_queryset(self, include_archived=False):
         """Get the experiments that the participant has interacted with"""
+        session_experiment_ids = self.experimentsession_set.values_list("experiment_id", flat=True)
+        data_experiment_ids = self.data_set.values_list("experiment_id", flat=True)
+        experiment_ids = set(session_experiment_ids) | set(data_experiment_ids)
         query = Experiment.objects.get_all() if include_archived else Experiment.objects.all()
-        return query.filter(Q(sessions__participant=self) | Q(id__in=Subquery(self.data_set.values("experiment"))))
+        return query.filter(id__in=experiment_ids)
 
     def get_data_for_experiment(self, experiment_id) -> dict:
         try:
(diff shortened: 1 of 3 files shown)


preston-56:
Reverting to #4476.

PR #4489, opened 2026-09-11
refactor: collect node type introspection behind NodeType

SmittieC:
### Product Description
No change.

### Technical Description
"What does this node type do?" was answered in half a dozen places — free functions in `build_state.py`, a private schema reader in the write API, `resolve_node_class` plus an ad-hoc `can_delete` check for "is this the server's node", and `Node`/`FlowNodeData` each spelling the same questions their own way. `NodeType` (`apps/pipelines/node_type.py`) collects them behind one value object and the callers now all ask the same way.

Two design points worth a look:

- **It holds no node state.** Keyed only by the type string, so it is constructible from a stored row, an unwritten edit on the wire, or a bare string. Params are arguments to the methods that need them rather than fields, which keeps it from becoming a second source of truth for what a node holds (ADR-0046).
- **It is a null object.** `Node.type` is graph data with nothing validating it against the node classes, and `REMOVED_NODE_TYPES` makes an unresolvable type a supported state — so every accessor answers for one (empty / `None` / `False`) rather than raising. `exists` is there for the caller that has to tell the cases apart. Discovery's `get_node_class` keeps its own 404 because "published type" and "resolvable type" genuinely differ for the removed-type stubs.

It lives in its own module rather than `nodes/base.py`, and imports nothing from `apps.pipelines.nodes` at module level. That is what lets `models.py`, `flow.py`, `versioning.py` and the API modules share it without dragging in the `nodes → langgraph → apps.experiments.models` chain; the one lazy import is isolated in `_nodes_base()`.

The last commit is a real bug fix riding along, not a refactor: `Node.requires_attachment_tool()` read `collection_id` out of `params` while every other read of that reference uses the FK column. An archived node whose collection was later deleted (the delete guard only queries unarchived nodes) keeps the stale id in JSON while the column is nulled — so the node claimed to need the attachment tool for a collection that is gone. Reviewable on its own if you want to take it separately.

Review is easiest commit by commit; each one is self-contained and green.

### Issue Link
N/A

### Migrations
N/A — no schema change.

### Demo
N/A — no user-visible change. Behaviour is covered by `apps/pipelines/tests/test_node_type.py` plus the existing pipeline/API suites (1522 passing across `apps/pipelines`, `apps/api/v2`, `apps/chat/tests`).

### Docs and Changelog

### Operator Impact


apps/pipelines/models.py
```diff
@@ -760,8 +760,13 @@ def _get_version_details(self) -> VersionDetails:
         )
 
     def requires_attachment_tool(self) -> bool:
-        """When a collection is linked, the attachment tool is required"""
-        return self.params.get("collection_id") is not None
+        """When a collection is linked, the attachment tool is required.
+
+        Read off the FK column, which ``resource_params`` documents as the constraint-backed mirror
```
SmittieC: Make this docstring a 1 liner

apps/pipelines/node_type.py
```diff
@@ -0,0 +1,217 @@
...
+    #: A type naming no node class -- removed since, or never one. Its handles are unknowable.
+    UNKNOWN_TYPE = "unknown_type"
+    #: A router with no keywords yet: its handles *are* its branches, so it has none until they are set.
+    NO_BRANCHES = "no_branches"
+    #: Offers none for a reason this method does not recognise -- unreachable today.
+    UNDETERMINED = "undetermined"
+
+
+@dataclass(frozen=True)
+class NodeType:
```
SmittieC: perhaps this can be a `TextChoices` and we can make the type field on Node be one of the choices. All of this logic then stays in this class. Ways one should be able to get a type:

```python
NodeType("node_type") # instantiate directly
node.type_enum # established pattern in the codebase. Go through the instance
```

apps/api/v2/pipeline_edit/graph_editor.py
```diff
@@ -54,9 +55,11 @@ def plan_create(flow: Flow, node_type: str, label: str | None, params: dict[str,
...
     """
     # The types the `pipeline_node_list` endpoint serves are exactly the resolvable node classes,
-    # and
-    # `get_node_type_schema` has already refused any other name, so this cannot come back None.
-    node_class = cast(type[BasePipelineNode], resolve_node_class(node_type))
+    # and `get_node_type_schema` has already refused any other name, so neither of these can come
+    # back None.
+    resolved = NodeType(node_type)
+    node_class = cast(type[BasePipelineNode], resolved.node_class)
+    schema = cast(NodeSchema, resolved.schema)
```
SmittieC: why cast to NodeSchema if `resolved.schema` already does this?

apps/pipelines/models.py
```diff
@@ -580,11 +601,11 @@ def create_new_version(self, is_copy=False, new_flow_id=None, pipeline=None):  #
         if is_copy and new_flow_id:
             old_flow_id = new_version.flow_id
             new_version.flow_id = new_flow_id
-            if new_version.type not in ("StartNode", "EndNode") and new_version.params.get("name") == old_flow_id:
+            if not self.node_type.is_server_managed and new_version.params.get("name") == old_flow_id:
```
SmittieC: Should this not be

```suggestion
            if not self.node_type.is_server_managed and new_version.display_name == old_flow_id:
```

apps/pipelines/models.py
SmittieC: Fixed in e797010d1. Now one line: `"""True when a collection is linked, read off the FK column rather than the id in params."""` — the FK-vs-params point survives, the paragraph does not.

*— Claude Code*

apps/api/v2/pipeline_edit/graph_editor.py
SmittieC: `resolved.schema` casts the *value* type (pydantic types `json_schema_extra` as a loose JSON dict or callable), but it is declared `NodeSchema | None` because an unresolvable type has no schema. The outer cast only drops that `| None` — same job as the `node_class` line above it, which casts away the `| None` on `NodeType.node_class`.

Added a clause to the comment in a912fe93a so the next reader does not have to ask.

*— Claude Code*

apps/pipelines/models.py
SmittieC: Not that one — `display_name` returns `self.type` when the name equals the flow id, which is exactly the case this branch is testing for. It happens to work here only because `flow_id` was reassigned to `new_flow_id` on the line above, so the fallback does not fire; swap those two lines and the branch goes dead.

Took the smaller half of the suggestion in 1cf73f124 — `new_version.name == old_flow_id`. `Node.name` is the plain `params.get("name")` lookup, so it reads better with no change in behaviour.

*— Claude Code*

apps/api/v2/pipeline_edit/graph_editor.py
```diff
@@ -6,19 +6,11 @@
...
-    FlowNode,
-    FlowNodeData,
-    NodeDiff,
-    react_flow_node_type,
-)
+from apps.pipelines.const import REACT_FLOW_END_TYPE
+from apps.pipelines.flow import EdgeDiff, Flow, FlowEdge, FlowNode, FlowNodeData, NodeDiff
 from apps.pipelines.models import Node
-from apps.pipelines.nodes.base import BasePipelineNode, NodeSchema, resolve_node_class
+from apps.pipelines.node_type import NodeType
```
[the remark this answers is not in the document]
SmittieC: Not applicable — `get_node_class` is still imported (`graph_editor.py:8`) and still called at line 85. The diff removed `resolve_node_class` from the `nodes.base` import, not `get_node_class` from the `discovery.node_types` import. No `NameError`; `apps/api/v2/pipeline_edit/tests/test_node_update.py` covers the param-update path.

*— Claude Code*

apps/pipelines/models.py
```diff
@@ -539,6 +528,33 @@ def __str__(self):
...
+        """
+        name = self.name
+        if name is None or name == self.flow_id:
+            return self.type
+        return name
+
+    @property
+    def tool_names(self) -> list[str]:
+        """The built-in tools this node has selected."""
+        return self.params.get("tools") or []
```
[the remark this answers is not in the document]
SmittieC: Fixed in 8b7e895f9. Confirmed: `get_node_tools` (`apps/chat/agent/tools.py:770`) appends `ATTACH_MEDIA` to this list, so a collection-backed node was mutating its own `params["tools"]`. `tool_names` now returns `list(...)` and says so in its docstring. Pre-existing rather than new — the old inline `node.params.get("tools") or []` had the same problem — but the property made it easier to miss.

*— Claude Code*

SmittieC:
The idea behind this refactor is to have a centralized place that answers questions about a node type as apposed to the node's data - functions are scattered across the codebase that interrogates node types, which is bad. Node type answers are derived from the node schemata. Node record / row questions can / should be asked to the Node model. This refactor isn't comprehensive in it's changes, but I think it's a good start

apps/pipelines/models.py
```diff
@@ -562,10 +578,15 @@ def to_flow_node(self) -> FlowNode:
...
         """True if this node's type declares ``param_name`` as a param. Unknown types have none."""
-        from apps.pipelines.nodes.base import resolve_node_class  # noqa: PLC0415 - heavy: nodes→langgraph
+        return self.node_type.declares(param_name)
+
+    def output_handles(self) -> list[dict]:
+        """The output handles this row offers, as ``{handle, label}``.
 
-        node_class = resolve_node_class(self.type)
-        return node_class is not None and param_name in node_class.model_fields
+        The type decides; this supplies the row. ``django_node`` goes along so a router's branches
```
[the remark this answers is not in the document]
SmittieC: Fixed in 528d51d94. Now reads `new_version.node_type.is_server_managed`.

No behaviour change — `new_version` is a clone of `self`, so the two agree — but every other term in that condition (`new_version.flow_id`, `new_version.name`) reads the new version, and the pre-refactor line read `new_version.type`, so `self` was the odd subject out.

*— Claude Code*

apps/pipelines/tests/test_node_resource_fks.py
```diff
@@ -632,6 +632,28 @@ def test_backfill_drops_archived_collection_index():
     assert set(node.collection_indexes.values_list("id", flat=True)) == {valid_index.id}
 
 
+@pytest.mark.django_db()
```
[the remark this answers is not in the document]
SmittieC: Not applicable — this is a per-module cohesion metric applied to a test file, and the threshold (4 responsibilities) is one no enumerated test module can meet.

The PR adds 22 lines to a 639-line file: one `TestRequiresAttachmentTool` case covering `requires_attachment_tool` reading the FK column rather than the stale id in params. The file was already at ~40 test functions; the new test did not create the shape, it crossed the reporting line.

The module is cohesive by subject — every test in it is about node resource FKs — which is what CodeScene's name-clustering heuristic cannot see. The obvious alternative home, `test_models.py`, is itself flagged for file length, so moving it there trades one biomarker for another. Splitting the file is a refactor unrelated to this PR.

*— Claude Code*

snopoke (approved):

PR #4492, opened 2026-09-11
Answer participant-actionable errors instead of failing the task

SmittieC:
### Product Description

When a participant sends something the chatbot cannot use — an image in a format the LLM provider rejects, or a voice note to a chatbot with no transcription configured — they now get a reply explaining what to change, on every channel, and the occurrence is no longer recorded as a failure.

Before, only Telegram, WhatsApp, Slack and email delivered that reply, and even there the Celery task failed, the trace was marked errored and Sentry got an event for a message that had already been answered. The web chat page, the embedded widget and the API discarded the reply and showed the participant raw exception text instead.

A voice note now always stays in the chat history and on the trace, whatever went wrong with it. Previously the reply about a voice note could arrive with the note missing from the history, reading as a non-sequitur.

### Technical Description

`UserReportableError` was not handled anywhere in `MessageProcessingPipeline`, so it fell to the catch-all for unexpected exceptions: generate a reply, run terminal stages, re-raise. The re-raise unwound past the point in `ChannelBase.new_user_message` that would have returned the reply.

It is now `UserActionableError` and is handled as control flow — generate the reply via `EventBot`, set `ctx.early_exit_response`, run terminal stages, return normally. The rule for raising it, and the pipeline's side of the contract, are recorded in **ADR-0065**.

Two supporting changes:

- Suppressing the error signal takes three places, because the exception can be raised at any depth: the stage's own span, spans opened inside the bot run, and LangChain's chain/LLM/tool error callbacks. Without all three, an unsupported attachment still finished the trace as `ERROR` and fired a "Run Pipeline Failed" notification.
- **Every** exception raised during query extraction — no speech, no transcription configured, and genuine backend faults alike — is now deferred to a new `ErrorGuardStage` that runs after `ChatMessageCreationStage`. The participant's message is persisted in all cases, faults included; those still re-raise from the guard and still notify the team.

Unrelated drive-by: `inv runserver --public` silently bound to localhost when `portless` was installed.

### Issue Link

None — found while auditing how this exception was raised and caught.

### Migrations

No migrations.

### Demo

None. The behaviour is covered end to end through the real pipeline and `ChannelBase.new_user_message`; it has not been exercised against a live LLM provider.

### Docs and Changelog

### Operator Impact

### DIFF
--- a/apps/api/tests/test_chat_poll_api.py
+++ b/apps/api/tests/test_chat_poll_api.py
@@ -109,31 +109,11 @@ def test_chat_poll_task_response_processing_with_progress(mock_progress, api_cli
 
 
 @pytest.mark.django_db()
-def test_chat_poll_user_facing_error_returns_400(api_client, mock_session, mock_task_response):
-    """User-facing errors (user_facing_error=True) must return HTTP 400, not 500."""
-    mock_task_response.return_value = {
-        "complete": True,
-        "error_msg": "Voice transcription is not available for this chatbot.",
-        "user_facing_error": True,
-        "message": None,
-    }
-
-    url = reverse("api:chat:task-poll-response", kwargs={"session_id": TEST_SESSION_ID, "task_id": "test-task-3"})
-    response = api_client.get(url)
-
-    assert response.status_code == 400
-    data = response.json()
-    assert data["status"] == "error"
-    assert "transcription" in data["error"]
-
-
-@pytest.mark.django_db()
-def test_chat_poll_generic_error_returns_500(api_client, mock_session, mock_task_response):
-    """Non-user-facing errors (internal failures) must return HTTP 500."""
+def test_chat_poll_error_returns_500(api_client, mock_session, mock_task_response):
+    """A task that finished with an error is a server error."""
     mock_task_response.return_value = {
         "complete": True,
         "error_msg": "Something went wrong internally.",
-        "user_facing_error": False,
         "message": None,
     }
 
--- a/apps/channels/pipeline.py
+++ b/apps/channels/pipeline.py
@@ -8,7 +8,7 @@ from django.utils import timezone
 
 from apps.channels.exceptions import EarlyAbort, EarlyExitResponse
 from apps.chat.bots import EventBot
-from apps.chat.exceptions import ChatException, NoSpeechDetected, NoSpeechReason
+from apps.chat.exceptions import ChatException, NoSpeechDetected, NoSpeechReason, UserReportableError
 from apps.pipelines.exceptions import (
     CodeNodeRunError,
     NodeUserConfigRunError,
@@ -219,10 +219,13 @@ class MessageProcessingPipeline:
         5. If any raises NoSpeechDetected, reply with a message generated
            from the reason and set ctx.early_exit_response -- but do NOT
            re-raise (silence is user input, not a bug).
-        6. If any raises an unexpected exception, generate an error message
+        6. If any raises UserReportableError, reply with a message generated
+           from the error and set ctx.early_exit_response -- but do NOT
+           re-raise (the participant can act on it, so it is not a bug).
+        7. If any raises an unexpected exception, generate an error message
            and set ctx.early_exit_response.
-        7. Run terminal stages unconditionally (they always fire).
-        8. If there was an unexpected exception, re-raise it after terminal
+        8. Run terminal stages unconditionally (they always fire).
+        9. If there was an unexpected exception, re-raise it after terminal
            stages complete.
         """
         try:
@@ -275,6 +278,13 @@ class MessageProcessingPipeline:
         except NoSpeechDetected as e:
             logger.info("No speech detected in voice message: %s", e.reason)
             ctx.early_exit_response = self._user_message(ctx, self.NO_SPEECH_PROMPTS[e.reason], e)
+        except UserReportableError as e:
+            # The participant can act on this -- an unsupported attachment, a voice note
+            # on a chatbot that cannot transcribe. Answer them with the generated message
+            # and do NOT re-raise: the fix is theirs to make, so it must not fail the task
+            # or reach Sentry.
+            logger.info("Participant-actionable error: %s", e)
+            ctx.early_exit_response = self._generate_error_message(ctx, e)
         except Exception as e:
             ctx.early_exit_response = self._generate_error_message(ctx, e)
             ctx.processing_errors.append(str(e))
--- a/apps/channels/pipeline.py
+++ b/apps/channels/pipeline.py
@@ -8,7 +8,7 @@ from django.utils import timezone
 
 from apps.channels.exceptions import EarlyAbort, EarlyExitResponse
 from apps.chat.bots import EventBot
-from apps.chat.exceptions import ChatException, NoSpeechDetected, NoSpeechReason, UserReportableError
+from apps.chat.exceptions import ChatException, NoSpeechDetected, NoSpeechReason, UserActionableError
 from apps.pipelines.exceptions import (
     CodeNodeRunError,
     NodeUserConfigRunError,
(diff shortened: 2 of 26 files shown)


apps/channels/pipeline.py
```diff
@@ -275,6 +278,13 @@ def _run_core_stages(self, ctx: MessageProcessingContext) -> Exception | None:
         except NoSpeechDetected as e:
             logger.info("No speech detected in voice message: %s", e.reason)
             ctx.early_exit_response = self._user_message(ctx, self.NO_SPEECH_PROMPTS[e.reason], e)
+        except UserActionableError as e:
+            # The participant can act on this -- an unsupported attachment, a voice note
+            # on a chatbot that cannot transcribe. Answer them with the generated message
+            # and do NOT re-raise: the fix is theirs to make, so it must not fail the task
+            # or reach Sentry.
+            logger.info("Participant-actionable error: %s", e)
+            ctx.early_exit_response = self._generate_error_message(ctx, e)
```
SmittieC: Fixed in 7ae36fbab. Extracted the per-exception-type handling out of `_run_core_stages` into a new `_handle_stage_exception` method, bringing the cyclomatic complexity of `_run_core_stages` well under the threshold.

*— Claude Code*

apps/channels/stages/core.py
```diff
@@ -604,7 +604,7 @@ def _do_transcription(self, ctx: MessageProcessingContext, audio: BytesIO) -> st
             speech_service = ctx.experiment.voice_provider.get_speech_service()
             if speech_service.supports_transcription:
                 return speech_service.transcribe_audio(audio)
-        raise UserReportableError("Voice transcription is not available for this chatbot")
+        raise UserActionableError("Voice transcription is not available for this chatbot")
```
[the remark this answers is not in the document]
SmittieC: Not applicable — this is already handled: `QueryExtractionStage.process` has an `except UserActionableError: raise` clause (apps/channels/stages/core.py:582) before the broad `except Exception` handler, so `_do_transcription`'s `UserActionableError` is re-raised without triggering the failure notification or the "Voice transcription failed" processing error.

*— Claude Code*

apps/channels/stages/base.py
```diff
@@ -8,17 +8,24 @@
...
-_CONTROL_FLOW_SIGNALS = (EarlyExitResponse, EarlyAbort, GenerationCancelled, NoSpeechDetected)
+# silent abort, generation cancellation, a voice note with no speech in it, an
+# error the participant can act on) rather than indicate something broke. They
+# must not mark a stage's span as errored, which would also count them on the
+# operator error-rate charts.
+_CONTROL_FLOW_SIGNALS = (
+    EarlyExitResponse,
+    EarlyAbort,
+    GenerationCancelled,
+    NoSpeechDetected,
```
SmittieC: perhaps `NoSpeechDetected` should be excluded here and `NoSpeechGuardStage` should rather raise `UserReportableError` with NO_SPEECH_PROMPTS instead, because they do pretty much the same thing, just at different levels atm.

apps/api/tests/test_chat_poll_api.py
```diff
@@ -109,31 +109,11 @@ def test_chat_poll_task_response_processing_with_progress(mock_progress, api_cli
...
-    data = response.json()
-    assert data["status"] == "error"
-    assert "transcription" in data["error"]
-
-
-@pytest.mark.django_db()
-def test_chat_poll_generic_error_returns_500(api_client, mock_session, mock_task_response):
-    """Non-user-facing errors (internal failures) must return HTTP 500."""
+def test_chat_poll_error_returns_500(api_client, mock_session, mock_task_response):
+    """A task that finished with an error is a server error."""
```
SmittieC: Not sure about this

apps/channels/stages/core.py
```diff
@@ -579,6 +579,11 @@ def process(self, ctx: MessageProcessingContext) -> None:
                 # attachment carry the content.
                 ctx.user_query = ""
                 ctx.no_speech_reason = e.reason
+            except UserActionableError:
```
SmittieC: this will suffer from the same issue that `ctx.no_speech_reason` solves. Perhaps NoSpeechDetected could be replaced entirely with UserActionableError. Rename `no_speech_reason` to `error_reason` or something similar and rename `NoSpeechGuardStage` to something more suitable. This relates to https://github.com/dimagi/open-chat-studio/pull/4492#discussion_r4002702972

apps/channels/stages/base.py
SmittieC: Fixed in 82e6fe051. `NoSpeechDetected` stays the speech service's vocabulary and is now caught at `QueryExtractionStage`, which hands the framework its own `UserActionableError` carrying the wording for the reason. The framework only ever raises and catches `UserActionableError`, so `NoSpeechDetected` is out of `_CONTROL_FLOW_SIGNALS` and out of the pipeline's exception policy entirely.

Rather than a second prompt lookup, `NO_SPEECH_PROMPTS` became `NO_SPEECH_MESSAGES` on the stage and the wording goes into the exception message — `_generate_error_message` already feeds a `ChatException`'s message to the error bot.

*— Claude Code*

apps/channels/stages/core.py
SmittieC: Fixed in 87d749934 (on top of the rename in 82e6fe051). You were right that it had the same bug: a chatbot with no transcription-capable voice provider raised straight out of `QueryExtractionStage`, so `ChatMessageCreationStage` never ran and the voice note was missing from both the chat history and the trace. It now defers through the same path, and `ErrorGuardStage` raises it once the turn is recorded.

`ctx.no_speech_reason` is now `ctx.error_reason`, holding the deferred `UserActionableError` rather than an enum, and `NoSpeechGuardStage` is now `ErrorGuardStage`. Added a failing end-to-end test first (`test_voice_note_to_a_bot_without_transcription_records_the_turn`) — it reproduced the missing `ChatMessage` before the fix.

*— Claude Code*

apps/channels/stages/core.py
```diff
@@ -557,6 +557,18 @@ def _get_original_message(self, ctx: MessageProcessingContext) -> ChatMessage |
 # ---------------------------------------------------------------------------
 
 
+# Participant-facing wording for a voice note the transcriber found no words in. Written
```
SmittieC: remove this comment

apps/channels/stages/core.py
```diff
@@ -572,21 +584,30 @@ def process(self, ctx: MessageProcessingContext) -> None:
...
                 ctx.user_query = self._transcribe_voice(ctx)
             except NoSpeechDetected as e:
-                # Defer to NoSpeechGuardStage so ChatMessageCreationStage records the
-                # turn first: the voice note is real input and belongs in the history
-                # and on the trace, even though there are no words in it. The empty
-                # query is what makes that stage keep the text empty and let the
-                # attachment carry the content.
-                ctx.user_query = ""
-                ctx.no_speech_reason = e.reason
+                self._defer(ctx, UserActionableError(NO_SPEECH_MESSAGES[e.reason]))
```
SmittieC: rename `_defer` -> `_defer_exception`

SmittieC: Fixed in b5c4f9339.

*— Claude Code*

apps/channels/stages/core.py
SmittieC: Fixed in 9014d3076.

*— Claude Code*

tasks.py
```diff
@@ -205,7 +205,7 @@ def _get_portless_name(c: Context) -> str:
 def runserver(c: Context, public=False):
     """Start Django development server (alias: inv django)."""
     _disable_stdin_forwarding(c)
-    if _has_portless(c):
+    if _has_portless(c) and not public:
```
SmittieC: Fixed in 00e7a570d. The `and not public` guard tipped `runserver` to 9. Pulled the command building out into `_base_runserver_command` and `_with_ngrok_env`, so `runserver` is now down to one branch.

*— Claude Code*

snopoke (approved):

PR #4508, opened 2026-09-15
fix: stop service provider credentials leaking into Sentry via pydantic model repr

### LINKED ISSUE #4404: Service provider credentials leak into Sentry via pydantic model repr
## Problem

Sentry issue OPEN-CHAT-STUDIO-2C7 (`UserReportableError: Unable to transcribe audio`) has 24 production events, and every one of them carries the team's live Azure Speech subscription key in plaintext.

The leaking frame is `apps/service_providers/speech_service.py:67`. `attach_stacktrace=True` sends stack-frame locals, and the local `self` serializes as its pydantic repr:

```
self: "AzureSpeechService(azure_subscription_key='<redacted>', azure_region='eastus')"
```

Anyone with issue-read access on the Sentry project can read the credential.

## Why the existing scrubber doesn't catch it

`config/sentry.py` has an `EventScrubber` denylist that already includes `auth_token`, `access_token`, `secret_key` etc. It cannot help here: the scrubber matches by exact variable/dict-key name, but the secret is not a variable — it is embedded inside the repr *string* of `self`. Adding `azure_subscription_key` to `SENTRY_SECRET_VAR_DENYLIST` would not fix it.

## Scope

Not Azure-specific. Every service-provider pydantic model holds its credential as a plain `str`, so any exception raised in a method of one of these classes leaks that credential the same way:

- `apps/service_providers/speech_service.py` — `aws_secret_access_key`, `azure_subscription_key`, `openai_api_key` (x2), `elevenlabs_api_key`, `intron_api_key`, `minimax_api_key`
- `apps/service_providers/messaging_service.py` — `auth_token` (x2), `client_secret`, `access_token`, `app_secret`, `verify_token`
- `apps/service_providers/llm_service/main.py` — `openai_api_key` (x2), `anthropic_api_key`, `deepseek_api_key`, `google_api_key`, `voyage_api_key`

`SpeechService.transcribe_audio` / `synthesize_voice` wrap every provider call in `except Exception`, so the speech models are the most exposed.

## Proposed fix

1. Type these fields as `pydantic.SecretStr` so the repr masks the value. `apps/service_providers/auth_service/main.py` already uses this convention. Roughly 30 `self.<field>` read sites need `.get_secret_value()`.
2. Add a regression test asserting `repr()` of each service model contains no credential material.

Rotation is the urgent part; the typing change prevents recurrence.

## Follow-up (separate)

All 24 events are Azure `NoMatch` (15 `NoMatchReason.InitialSilenceTimeout`, 9 `NoMatchReason.NotRecognized`) — silent or unintelligible WhatsApp voice notes, not infrastructure failures. Worth handling as a user-facing "couldn't hear that" reply instead of an unhandled Sentry error. Two things spotted while reading `_transcribe_audio`, also worth their own issues: the recognition language is hardcoded `en-US` while the affected participants are on Indian numbers, and the incoming audio is written to a `.wav`-suffixed temp file regardless of its actual content type (`audio/ogg` in these events), which Azure sniffs by extension.

See https://github.com/dimagi/open-chat-studio/issues/4405

zandre-eng:
### Product Description
No user-visible change. Team credentials for speech, messaging and LLM providers were being published to Sentry in plaintext; this stops that and prevents it recurring.

### Technical Description
Sentry runs with `attach_stacktrace=True`, so stack-frame locals travel with every event. A pydantic service model sitting on the frame is serialized by its repr:

```
self: "AzureSpeechService(azure_subscription_key='<the real key>', azure_region='eastus')"
```

The `EventScrubber` in `config/sentry.py` cannot help: it matches by exact variable or dict-key name, and here the secret is not a variable — it is embedded inside the repr *string* of `self`. Adding `azure_subscription_key` to the denylist would change nothing.

**The fix** types the credential fields as `pydantic.SecretStr`, the convention `apps/service_providers/auth_service` already follows, and unwraps with `.get_secret_value()` at each client boundary — 19 fields, 27 read sites:

| File | Fields | Read sites |
|---|---|---|
| `speech_service.py` | 7 | 9 |
| `messaging_service.py` | 6 | 7 |
| `llm_service/main.py` | 6 | 10 |
| `channels/tasks.py` (Twilio signature validation) | — | 1 |

`GoogleVertexAILlmService.credentials_json` leaks a service account private key through the same channel. `SecretStr` cannot wrap a dict, so it is excluded from the repr instead (`Field(repr=False)`); it lands as its own commit since it was outside the original issue scope.

No construction sites change — every model is built by unpacking a config dict, so pydantic coerces the stored strings.

### Issue Link
Closes #4404

### Migrations
None — this PR adds no migrations.

### Demo
Two guards, in one parametrized test file:

- **repr guard** — discovers every production service model by subclass and asserts no credential material survives `repr()`/`str()`, plus a structural check that a credential-named field is `SecretStr` or excluded from the repr. New providers are covered without a new test.
- **boundary guard** — drives each service to its first provider call and asserts the credential arrives as a plain string. Nothing else in the suite checks this: a mocked SDK accepts `SecretStr('**********')` without complaint, so a missing `.get_secret_value()` stays green and fails only in production, as an auth error.

That last point is measured, not assumed. Stripping `.get_secret_value()` from a line and running the other 2697 tests:

```
SURVIVED  azure transcribe        (line IS executed by the existing suite)
SURVIVED  turn.io client          (line IS executed by the existing suite)
SURVIVED  anthropic chat model    (line IS executed by the existing suite)
SURVIVED  twilio client           (never executed)
SURVIVED  twilio signature check  (never executed)
```

All five survive — the suite goes green with Twilio auth broken. With the new tests in place, mutating each of the 27 read sites in turn: **27/27 caught.**

### Docs and Changelog

### Operator Impact
Self-hosted operators running any affected provider should assume credentials in their own Sentry project were exposed and **rotate them**. Deploying this change does not scrub events that were already sent — existing events must be deleted from Sentry separately.


config/sentry.py
```diff
@@ -13,6 +13,10 @@
 #
 # Anything holding raw secret/key material should be named to match one of these entries; prefer
 # the ``encryption_key`` convention for CommCare Connect per-participant keys.
+#
+# The denylist cannot reach a secret embedded in the repr *string* of a frame local, so a credential
+# held on a pydantic model needs ``pydantic.SecretStr`` instead (or ``Field(repr=False)`` where
+# SecretStr does not fit). ``apps/service_providers/tests/test_credential_repr.py`` enforces that.
```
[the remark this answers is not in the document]
zandre-eng: Resolved in [c2d254b](https://github.com/dimagi/open-chat-studio/pull/4508/commits/c2d254b7314b7f2528717f06f48af26a98630096).

apps/service_providers/tests/test_credentials.py
```diff
@@ -0,0 +1,255 @@
+"""Service provider credentials stay out of Sentry, and still reach their clients intact.
+
+``attach_stacktrace=True`` sends stack-frame locals with every event, so a pydantic service model
+on the frame is serialized by its repr. The ``EventScrubber`` in ``config/sentry.py`` matches by
+exact variable/dict-key name and cannot reach a secret embedded inside that repr string, which is
+how an Azure subscription key was published to 24 production events.
```
SmittieC: > which is
how an Azure subscription key was published to 24 production events.

this is probably not worth mentioning

SmittieC (dismissed):

apps/service_providers/tests/test_credentials.py
zandre-eng: Good point. Addressed in [326a8ae](https://github.com/dimagi/open-chat-studio/pull/4508/commits/326a8aef633af9ec5f63476e54b839bcec182afa).

SmittieC (approved):

PR #4520, opened 2026-09-16
Fix audit_archived_pipeline_references broken by assistant model removal

snopoke:
`main` is red: `audit_archived_pipeline_references` raises `FieldError: Cannot resolve keyword 'assistant'`. #4456 (audit command) and #4484 (removes `Node.assistant` from Django state, and the assistant spec from the versioning registry) merged concurrently and neither saw the other.

The command now derives its archived-reference filter from `all_versioned_param_specs()` rather than naming FK fields by hand, so it can't drift from the registry again.

The "reports nothing" test built an `AssistantNode` with an `assistant_id` param. `AssistantNode` is a removed node type with no FK mirror, so `update_from_params()` sets nothing and the test asserted only that an unrelated empty query is empty. Rewritten against a live, non-archived source material so it actually exercises the negative case.

### Issue Link
N/A

### Docs and Changelog

### Operator Impact


ocs-percy (approved):
Clean fix for the broken `main` — the derive-from-registry approach is strictly better than the hardcoded field list that caused this in the first place.

**Filter derivation** — `reduce(or_, ...)` over `all_versioned_param_specs()` replaces the hardcoded `Q(assistant__...) | Q(source_material__...) | ...`. The old list was already stale the moment `Node.assistant` was removed; now the command can't drift from the registry again. Adding `_M2M_FK_FIELDS` as an explicit constant is a nice touch — it was previously implicit and inline.

**`select_related` / `prefetch_related`** — `select_related` correctly keeps using `_SCALAR_FK_FIELDS` only; `prefetch_related("collection_indexes")` is still hardcoded but that's a pre-existing minor inconsistency, not a regression here.

**One theoretical note** — `reduce(or_, ...)` on an empty sequence raises `TypeError` at module load time. In practice impossible given the registry always has entries, and the original code had the same implicit assumption, so not blocking.

**Test fix** — the old `AssistantNode` test was hollow: `update_from_params()` on a removed node type set nothing, so it was asserting an unrelated empty query is empty. The rewrite against `LLMResponseWithPrompt` + `SourceMaterialFactory` actually exercises the negative case. Good cleanup.

PR #4521, opened 2026-09-16
chore(lint): enable RUF, PERF, PIE, RET, TC and C90 ruff rule sets

### LINKED ISSUE #4513: Enable additional Ruff rule sets (RUF, PERF, PIE, RET, TCH, C90)
## Goal

Add several Ruff rule sets to `pyproject.toml` to catch more categories of code issues. Each task adds one rule set, fixes (or suppresses with `# noqa`) any violations it reveals, and leaves the linter green. Tasks are designed to be reviewed independently.

## Context

- **File:** `pyproject.toml`, section `[tool.ruff.lint]`
- **Current `select`:** `["E", "F", "I", "UP", "DJ", "PT", "B", "SIM", "PLC0415", "TID253"]`
- **Preview a rule set before adding it:** `uv run ruff check --select <RULESET> .`
- **Auto-fix:** `uv run ruff check --fix --select <RULESET> .`
- **Full lint + format:** `uv run inv ruff`
- **Tests:** `uv run pytest -n auto --no-migrations`
- Parent tracking issue: #4257

### Task 1: Enable `RUF` (Ruff-specific rules)
- [ ] Add RUF to ruff select; fix violations

Append `"RUF"` to the `select` list and add `"RUF001"`, `"RUF002"`, `"RUF003"` to the `ignore` list (unicode-confusable checks — noisy for human-facing strings in this codebase).

Key rules this enables:
- `RUF006` — `asyncio.create_task()` result not stored; task silently garbage-collected
- `RUF012` — Mutable class-body default not annotated `ClassVar` (Django model footgun)
- `RUF019` — Redundant key existence check before dict access
- `RUF100` — Stale `# noqa` directives left behind after violations are fixed

Preview first: `uv run ruff check --select RUF .`

Auto-fix what's safe: `uv run ruff check --fix --select RUF .`

For remaining violations: fix manually, or suppress with `# noqa: RUFXXX` only where the rule is genuinely inapplicable (add a reason comment).

Run `uv run pytest -n auto --no-migrations` and `uv run inv ruff` before opening the PR.

### Task 2: Enable `PERF` (Perflint)
- [ ] Add PERF to ruff select; fix violations

Append `"PERF"` to the `select` list.

Key rules:
- `PERF203` — `try`/`except` inside a loop (exception-handling overhead per iteration)
- `PERF401` — Manual `for` loop building a list that should be a list comprehension

Preview: `uv run ruff check --select PERF .`
Auto-fix: `uv run ruff check --fix --select PERF .`

Run tests before opening PR.

### Task 3: Enable `PIE` (flake8-pie) and `RET` (flake8-return)
- [ ] Add PIE and RET to ruff select; fix violations

Append `"PIE"` and `"RET"` to the `select` list.

Key rules:
- `PIE790` — Unnecessary `pass` in a non-empty block
- `PIE810` — Multiple `isinstance` checks combinable into one tuple
- `RET504` — Unnecessary intermediate variable before `return`
- `RET505`–`RET508` — Superfluous `elif`/`else` after `return`/`break`/`continue`/`raise`

Preview: `uv run ruff check --select PIE,RET .`

If `RET505`–`RET508` produce many hits without clear benefit (some codebases prefer explicit `else` after `return` for readability), add them to `ignore` and keep only `RET504`. Document the decision in the PR.

Auto-fix: `uv run ruff check --fix --select PIE,RET .`
Run tests before opening PR.

### Task 4: Enable `TCH` (flake8-type-checking)
- [ ] Add TCH to ruff select; fix violations

Append `"TCH"` to the `select` list.

This rule set moves imports that are only used in type annotations into `if TYPE_CHECKING:` blocks. Reduces runtime import overhead and eliminates a category of circular import — consistent with the codebase's existing lazy-import policy (`TID253`).

Preview: `uv run ruff check --select TCH .`
Auto-fix: `uv run ruff check --fix --select TCH .`

**Important:** After auto-fix, run the full test suite — a common mistake is moving a `TYPE_CHECKING`-only import that is actually used at runtime (e.g. in a `isinstance` check or a default argument). If tests fail, move that import back and add `# noqa: TCH001` (or `TCH002`/`TCH003`) with a comment.

Also run `uv run inv typecheck` to confirm ty still passes.

### Task 5: Enable `C90` (McCabe complexity)
- [ ] Add C90 to ruff select with max-complexity = 15; suppress or refactor flagged functions

Append `"C90"` to the `select` list and add the threshold config:

```toml
[tool.ruff.lint.mccabe]
max-complexity = 15
```

Preview flagged functions: `uv run ruff check --select C90 .`

For each flagged function:
- If it can be cleanly broken into smaller helpers → refactor it (with tests).
- If the complexity is load-bearing (e.g. a parser, a state machine, a migration) → add `# noqa: C901` with a one-line comment explaining why.

Do not refactor functions just to pass the check if the refactoring makes the code harder to follow. Suppression with a reason is the right call in those cases.

Run tests and `uv run inv ruff` before opening PR.

snopoke:
### Product Description

No user-facing change. Static-analysis coverage only.

### Technical Description

Adds six ruff rule sets to `[tool.ruff.lint] select`, one commit per rule set so each can be reviewed independently:

| # | Commit | Rule set | Violations |
|---|---|---|---|
| 1 | `957c85bbc` | `RUF` | 154 |
| 2 | `54146065e` | `PERF` | 34 |
| 3 | `0a37c753a` | `PIE` + `RET` | 189 |
| 4 | `3a360106b` | `TC` | 80 |
| 5 | `7a5bb88bb` | `C90` @ 15 | 7 |

`TCH` is the deprecated alias in ruff 0.16 and selecting it warns, so the rule set is selected as `TC`.

#### A real bug this surfaced (`05b580cf7`)

`RUF015`'s `next(iter(x))` rewrite is not equivalent when `x` is a `QuerySet.iterator()`. `index_collection_files_task` and `migrate_vector_stores` pass `collection_files=QuerySet.iterator(100)` -- a server-side cursor -- so `next(iter(...))` read one row and left the cursor open. It was closed after the test's transaction had already rolled back, and the resulting error aborted the transaction of whichever test ran next in that worker. The three call sites now drain through a `_drain` helper whose docstring records why.

#### Rules ignored, with reasons

- `RUF001/2/3` (30) -- unicode confusables are intentional in human-facing strings/docstrings/comments (as the issue specified).
- `RUF012` (371) -- almost all Django/DRF declarative class config (`Meta.fields`, `widgets`, `list_display`). `ClassVar`-annotating each is ~370 lines of noise that catches no bugs.
- `RUF100` (47) -- conflicts with an enforced repo convention. `scripts/check_inline_imports.py` parses `# noqa: PLC0415 - <reason>` as a machine-read lazy-import justification (`NOQA_REASON`, line 51). TID253's `banned-module-level-imports` already exempts those modules from `PLC0415`, so ruff reads all 19 directives as unused and its fix deletes them, stripping the reason that checker requires. Its fix also corrupted `scripts/check_inline_imports.py:49` -- the comment documenting the convention -- because ruff parsed the comment's own `# noqa:` mention as a directive. The rest are `BLE001`/`S608`/`N803`/`D101` for rules this project has not enabled.
- `RET505`-`RET508` (63) -- the issue's stated escape hatch. All 63 only de-indent a block; an explicit `else`/`elif` after `return` is clearer where branches are parallel in meaning (dispatch on request method, node type, provider). `RET501`-`RET504` stay on and are fixed.

#### Config added

```toml
[tool.ruff.lint.mccabe]
max-complexity = 15

[tool.ruff.lint.flake8-type-checking]
runtime-evaluated-base-classes = ["pydantic.BaseModel"]
```

The `flake8-type-checking` setting keeps an import referenced by a pydantic field annotation at module level -- pydantic resolves those at model build time, unlike method signatures.

#### Per-task notes

- `PERF`: no `PERF203` -- there is no `try`/`except` inside a loop in the codebase. 8 auto-fixed; the 26 `PERF401`/`PERF403` hits had no available fix (ruff only offers one when the target starts empty at the loop) and were rewritten by hand. Ruff's `PERF403` suggestion for `apps/pipelines/nodes/base.py` would have rebound `output` instead of updating it, so that one uses `output.update(...)`.
- `TC`: `TC006` quotes the type in `typing.cast()`; `TC001`/`TC002`/`TC003` were only reported in modules that already carry `from __future__ import annotations`, so no annotation stops resolving at runtime and no new `__future__` import was introduced. The one `TC004` hit was a duplicate `Node` import in `apps/chat/agent/tools.py` (module level and inside `TYPE_CHECKING`); the duplicate is removed.
- `C90`: all seven over-threshold functions are the load-bearing cases the issue describes (two one-shot data migrations, three CSV parsers, two Langfuse migration functions), so each carries `# noqa: C901` with a reason rather than a refactor.

#### Verification

- `uv run ruff check .` and `uv run ruff format --check .` clean
- `uv run ty check apps/` clean
- `uv run pytest -n auto`: 8220 passed, 1 skipped, 0 failed
- `scripts/` repo-wide gates (`check_atomic_exception_handling`, `check_inline_imports`, `reconcile_models`): 237 passed
- No function-level import was added or removed, so the `check_inline_imports.py` CI gate is unaffected

### Issue Link

Closes #4513

### Migrations

No migrations.

### Demo

N/A -- no runtime surface.

### Docs and Changelog

### Operator Impact

---


SmittieC (dismissed):
lgtm

snopoke:
@claude please resolve conflicts

snopoke:
### Response to the OpenCodeReview findings

**`apps/service_providers/tests/mock_tracer.py` (critical) — not applicable.**

Annotations on attribute targets (`self.spans: dict[UUID, dict] = {}`) inside a function body are not evaluated at runtime; PEP 526 only evaluates the annotation for a complex target in a module or class body. Verified on 3.13.0, the version CI runs:

```
>>> class A:
...     def __init__(self):
...         self.x: dict[Undefined, dict] = {}
>>> A()
<__main__.A object at ...>
```

and directly: `MockTracer()` instantiates, `t.spans == {}`. `apps/service_providers/tests/test_tracing_service.py` passes.

**`apps/evaluations/views/evaluation_config_views.py` (medium) — fixed in f6da89c33.**

Real issue. Rather than adding `HttpResponseNotAllowed` (unused elsewhere in the repo), the view now carries `@require_http_methods(["GET", "POST"])`, matching `load_experiment_versions` in the same file, and the `elif request.method == "POST"` became `else` so there is no trailing return to write.

**`apps/utils/slug.py` (low) — fixed in 5bbd3c940.**

Agreed on the misleading `str | None` contract. The trailing `return None` is now a `raise ValueError`, which satisfies RET503 without widening the return type.

---

The two failing checks (`review`, `Automated review verdict`) are environmental: `Workflow initiated by non-human actor: claude (type: Bot)`. The `allowed_bots` guard keys off the push actor, and the previous push was the conflict-resolution merge made by the bot. The push above is from a human account, so they should go green on this run.

*— Claude Code*

SmittieC (approved):

PR #4522, opened 2026-09-16
Drop the OpenAI Assistants tables and FK columns (#4254)

pxwxnvermx:
### Product Description
The final step of the OpenAI Assistants removal: the `assistants` tables and the two FK columns pointing at them are physically dropped, along with the app's permissions. Nothing user-facing changes — the rows were already deleted in #4484 and no code has read them since.

**This must not merge until #4484 has been deployed.** It is the second half of a two-phase drop and is safe only once the release that removed these from Django state is live.

### Technical Description
#4484 deleted the data and removed everything from Django *state*, deliberately leaving the columns and tables in place so the previous release kept working through the deploy. This drops them for real.

**`pipelines/0032`** and **`custom_actions/0008`** drop `assistant_id` from `pipelines_node` and `custom_actions_customactionoperation`. Postgres takes each column's FK constraint with it, and on the operations table also the `assistant_or_node_required` check and the `unique_assistant_custom_action_operation` partial unique index, since both depend solely on that column — so neither needs removing by name and no `CASCADE` is involved. `0008` then creates `node_required` as a database-only operation, because #4484 already added it to state.

**`assistants/0017`** removes the two `ToolResources` relations, runs `DeleteModel` on `ToolResources` then `OpenAiAssistant`, and finally deletes the app's `ContentType` and `Permission` rows. That cleanup has to come *after* `DeleteModel`: `contenttypes` and `auth` recreate both on every `migrate` for any model still in Django state, so doing it earlier just has post_migrate put them back. The column drops are declared as dependencies — `DROP TABLE` would otherwise hit the inbound FK constraints.

`apps/assistants/models.py`, `apps/utils/factories/assistants.py` and the `0016` migration test go with the tables. The app stays in `INSTALLED_APPS` as a migrations-only package so Django keeps owning the historical graph, which is also why it joins the model-less apps that `test_missing_content_types` ignores. `CONTENT_TYPES["assistants"]`, the two `IGNORED_MODELS` entries and the package-map entry go too — removing the `CONTENT_TYPES` entry is what finally revokes `Super Admin`'s (already unreachable) assistants permissions, via `GroupDef.update_or_create`'s `group.permissions.set()`.

**Tests that used an assistant incidentally are repointed, not deleted**, so their coverage survives:
- The team-deletion auditing regression (deleting a team must not null another team's `llm_provider`) now uses `Collection` — audited, team-scoped, `SET_NULL` FK to `LlmProvider`.
- The unlinked-version rendering test also uses `Collection`: only chatbots set `has_version_specific_url`, so every other family reports versions as plain badges, and the assertion now pins that the row name is the one link.
- The archived-reference negative case uses a live source material.

Only genuinely assistant-specific tests were removed: `test_file_deletion.py` in full (every test was about `ToolResources` cascading to files), the usages "includes assistants" case, and a provider-model deletion test that `test_cannot_delete_provider_models_used_by_an_evaluator` already covers for the FK path.

**Verified against a real database**, not inferred from green tests — after a full `--create-db` run:

```
assistants_* tables ......... none
assistant_id columns ........ none, in any table
constraints ................. node_required  (the two assistant-era ones went with the column)
content types / permissions . 0 / 0
```

Full suite: **8250 passed, 1 skipped, 0 failures**. `makemigrations --check` reports no changes.

### Issue Link
#4254

### Migrations

Backwards compatible **given #4484 is deployed**, which is the condition this PR waits on. That release removed all of this from Django state, so the running code selects none of the dropped columns and reads none of the dropped tables.

Not reversible by revert — the tables are gone. The rows they held were already deleted in #4484.

### Demo
No UI. A pipeline holding an `AssistantNode` still renders it as a **Removed Node** and still cannot be built, unchanged.

### Docs and Changelog

`docs/architecture/package-map.md` is updated here. A changelog entry for the schema drop should go in the release this lands in.

### Operator Impact

The `assistants` tables are dropped permanently, as are `pipelines_node.assistant_id` and `custom_actions_customactionoperation.assistant_id`. The audit log retains the history of the rows themselves.

`Super Admin` loses its assistants permissions on the deploy after this merges — they granted access to nothing, since the models no longer exist.

If `retire_assistant_file_purpose` (shipped in #4484, run post-deploy) has not been run yet, run it **before** this deploys. It reclaims the storage behind files that only ever belonged to an assistant; it is still safe to run afterwards, but the rows are easier to reason about while the purpose value is still meaningful.


apps/custom_actions/migrations/0008_drop_operation_assistant_column.py
```diff
@@ -0,0 +1,31 @@
...
+        migrations.RunSQL(
+            sql="ALTER TABLE custom_actions_customactionoperation DROP COLUMN IF EXISTS assistant_id;",
+            reverse_sql=migrations.RunSQL.noop,
+        ),
+        migrations.SeparateDatabaseAndState(
+            state_operations=[],
+            database_operations=[
+                migrations.AddConstraint(
+                    model_name="customactionoperation",
+                    constraint=models.CheckConstraint(condition=Q(("node__isnull", False)), name="node_required"),
```
SmittieC, line 27: why is this needed?

pxwxnvermx: The constraint before depended on `assistant` and `node` columns, since the assistant column is getting dropped now the constraint needed to be modified as well.

snopoke: I'm with @SmittieC. I don't see the assistant column referenced on [main](https://github.com/dimagi/open-chat-studio/blob/main/apps/custom_actions/models.py#L155-L159) and there are no changes to the model file in this PR. It looks like the code change to the model was in https://github.com/dimagi/open-chat-studio/commit/f61b4df70b6ec88172eef267d3703eeba4894b8c which already included a 'state migration'. Should this be a SQL migration to 'apply' those changes?

pxwxnvermx: Yes — that's exactly what this is.

`f61b4df70` (already on main) added `node_required` to the model and to Django's migration **state only**: `custom_actions/0007` wraps it in `SeparateDatabaseAndState` with `database_operations=[]`. So on main today the model declares the constraint and the database does not have it. On a scratch DB migrated to `0007`:

```
          conname           |                     pg_get_constraintdef
----------------------------+---------------------------------------------------------------
 assistant_or_node_required | CHECK (((assistant_id IS NOT NULL) OR (node_id IS NOT NULL)))
```

`node_required` is absent. This migration is the SQL half that applies it — the same wrapper the other way round (`state_operations=[]`), so `AddConstraint` emits the DDL without re-adding to state. After `0008` the same query returns:

```
    conname    |     pg_get_constraintdef
---------------+-------------------------------
 node_required | CHECK ((node_id IS NOT NULL))
```

It can't just be dropped, because the old constraint is never removed by name anywhere. Postgres drops `assistant_or_node_required` and the partial unique index automatically along with `assistant_id`:

```sql
ALTER TABLE op DROP COLUMN assistant_id;
-- assistant_or_node_required and unique_assistant_op both disappear with it
```

So without this line the table would come out of the PR with no owner check at all, while `models.py` still claims one — and `node_id` is still nullable at the column level, so nothing else would stop an operation with no owner.

`AddConstraint` inside `database_operations` rather than raw `RunSQL` just lets Django generate the DDL instead of hand-writing it.

*— Claude Code*

snopoke: Right, thanks for the clarification. I actually missed that this was a database operation and not a state operation.

snopoke (approved):

PR #4528, opened 2026-09-16
Add OpenCodeReview PR review workflow

snopoke:
### Product Description
No change — CI tooling only.

### Technical Description
Adds [OpenCodeReview](https://github.com/alibaba/open-code-review) as a second automated reviewer alongside `claude-code-review.yml`.

It starts in shadow mode: `route_severity_below: critical` routes every finding, at any severity, into one sticky summary comment instead of onto the diff, so inline review comments still come only from Claude while we work out whether OCR's output is worth acting on. Deleting that one input switches it to inline.

- **Fork PRs are skipped.** Upstream's own example uses `pull_request_target` to reach secrets from forks; this uses `pull_request`, matching the fork posture the other review workflows already document. Changing that is a separate security call, not a workflow edit.
- **Reuses `ANTHROPIC_API_KEY`** — no new secret to provision.
- **Double-pinned.** The action installs the OCR CLI from npm at run time, so pinning the action SHA alone doesn't freeze behaviour; `ocr_version` is pinned to the matching release.
- `claude-sonnet-5` rather than Opus, for cost and latency. Upstream trades recall for precision deliberately — if the trial shows recall is the weak spot, the model is the lever.

Worth a closer look: the input set is essentially the whole change. I validated it against `action.yml` at the pinned commit (no unknown inputs, no missing required ones) and actionlint passes, but nothing has run against a real PR yet. The first run on this PR is the actual test, including whether the endpoint config authenticates.

Left out on purpose: [dimagi-rad/scout](https://github.com/dimagi-rad/scout/blob/main/.github/workflows/ocr.yml) stages OCR as a *gate* in front of the Claude review rather than beside it, which is a better end state than running both — but it needs calibration data on how often OCR reports high/critical on our PRs, which is what shadow mode produces. OCR also takes path-targeted rules in `.opencodereview/rule.json` that the AGENTS.md invariants would map onto closely; premature before seeing baseline output.

### Docs and Changelog

### Operator Impact


SmittieC (approved):
Cool

Issue #4531, opened 2026-09-16
Consolidate privilege elevation (sudo) behind typed grants and allauth re-auth


snopoke:
Privilege elevation ("sudo") is spread across `apps/web/superuser_utils.py`, two views in `apps/web/views.py`, `OcsAdminSite.admin_view`, `check_superuser_team_access` and a context processor. The core idea is sound; the problems are all at the edges. This consolidates it behind one typed grant model and hands the identity-proof step to allauth.

## Current state

| Part | Location |
|---|---|
| Store | `apps/web/superuser_utils.py` — `session["elevated_privileges"]`, list of `(grant, unix_expiry)`, 30 min TTL, max 5 |
| Acquire / release | `apps/web/views.py:54`, `apps/web/views.py:90` — password re-auth via `ConfirmIdentityForm` |
| Consumers | `OcsAdminSite.admin_view` (grant `admin_site`), `check_superuser_team_access` (grant = team slug), `get_team_membership_for_request` → `SuperuserMembership`, `context_processors.project_meta` → banner |
| Entry points | the `/django-admin/` redirect, an inline form on `templates/404.html`, or `/sudo/<slug>/` directly |

### Security / correctness

**A — `admin_site` shares a namespace with team slugs.** `ADMIN_SLUG = "admin_site"` and `slugify("admin_site") == "admin_site"`. A team with that slug means elevating into the team also unlocks `/django-admin/`, with no admin-site confirmation.

**B — `/admin/` (`apps.admin`) has no step-up at all.** `apps/admin/views.py:68` defines its own bare `is_staff` / `is_superuser` checks. That surface includes `update_flag` / `delete_flag` (global feature flags), `configuration` (`OcsConfiguration`), `team_metadata` (CSV import, `is_staff`), and `find_provider_by_key` (`is_staff`, decrypts every provider's secrets cross-team to match a pasted key). Meanwhile `/django-admin/` — mostly read — demands a password.

**B2 — staff skip elevation on `/django-admin/` entirely.** `AdminSite.has_permission` is `is_active and is_staff`, but `apps/web/admin.py:39` guards the elevation check with `if request.user.is_superuser`. A staff non-superuser gets in with no step-up.

**C — `ValueError` escapes into the request.** `apply_temporary_superuser_access` raises on the 6th concurrent grant (`superuser_utils.py:23`) and on a blank grant. Neither call site (`views.py:65`, `decorators.py:74`) catches it, so it is a 500 rather than "release one first".

[...]

PR #4532, opened 2026-09-16
Fix session churn and error handling in superuser elevation

snopoke:
### Product Description
No user-visible change, except that hitting the 5-grant elevation cap now says so instead of erroring.

### Technical Description
Phase 0 of #4531 — the fixes that don't depend on the refactor.

- `remove_expired_temporary_superuser_access` wrote the session unconditionally. It runs from the `project_meta` context processor on every rendered page, so every authenticated request marked the session modified: one `UPDATE django_session` per page view, and session expiry rolling indefinitely. It now filters the stored entries instead of rebuilding them and returns early when nothing was pruned. Rebuilding was the trap — the session's JSON serializer returns each entry as a list, so re-packed tuples never compare equal to what is stored.
- The 5-grant cap raised `ValueError` and neither call site caught it, so it was a 500. Now a `TooManyElevatedPrivileges` exception, surfaced as a form error on the acquire view and as a denial in the DEBUG branch of `check_superuser_team_access`. Logged at the raise site, since neither caller would otherwise leave a trace of the denial.
- `remove_temporary_superuser_access` logged a removal that may not have happened; it now reports whether the grant was held and logs only then.

Not included, against what #4531's Phase 0 says: honouring a validated `next` on release. Redirecting back to the page you just dropped access to lands you on the acquire prompt, which reads worse than the current redirect to `/`. Left for the phase that reworks the URLs.

### Issue Link
Part of #4531

### Migrations

### Demo
n/a

### Docs and Changelog

### Operator Impact


apps/web/superuser_utils.py
```diff
@@ -20,7 +24,15 @@ def apply_temporary_superuser_access(request, grant):
 
     elevated_privileges = request.session.get("elevated_privileges", [])
     if len(elevated_privileges) >= MAX_CONCURRENT_PRIVILEGES:
-        raise ValueError("Maximum number of concurrent privileges exceeded")
+        # Logged here rather than at the call sites: one turns this into a form error and the
+        # other falls through to deny, so neither would leave a trace of the denial.
```
snopoke:
```suggestion
```

snopoke: Fixed in 7b97f56. Agreed — it was defending a placement decision nobody had questioned.

*— Claude Code*

apps/web/views.py
```diff
@@ -62,11 +66,19 @@ def acquire_superuser_powers(request, slug):
...
+                    form.add_error(
+                        None,
+                        "You already hold the maximum number of elevated privileges. "
+                        "Release one of them and try again.",
+                    )
+                else:
+                    redirect_to = form.cleaned_data["redirect"]
+                    if not redirect_to or not url_has_allowed_host_and_scheme(redirect_to, allowed_hosts=None):
+                        redirect_to = "/"
+                    return HttpResponseRedirect(redirect_to)
```
snopoke: Fixed in 47f3e2d. Extracted `_safe_redirect` and `_grant_on_confirmed_identity`; the former also collapses the duplicated redirect validation in both branches. The method was already over threshold at 15 before this PR, so this takes it below rather than back to parity.

*— Claude Code*

zandre-eng (approved):

PR #4534, opened 2026-09-17
Feat: Adds Anthropic's sanitize tool

### LINKED ISSUE #4027: Sanitize tool/schema property names to satisfy Anthropic's ^[a-zA-Z0-9_.-]{1,64}$ constraint
## Problem

Anthropic requires all tool names and input schema property keys to match `^[a-zA-Z0-9_.-]{1,64}$`. When we generate tools or structured-output schemas dynamically from user-supplied names (e.g. evaluator output fields, OpenAPI parameter names), those names are passed through verbatim. Any user-defined string that contains spaces, parentheses, slashes, etc. causes a 400:

```
Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
  'message': "tools.0.custom.input_schema.properties: Property keys should match pattern '^[a-zA-Z0-9_.-]{1,64}$'"}}
```

### 1. LLM Evaluators — `apps/evaluations/`

`LLMEvaluator.evaluate()` (`evaluators.py`) calls `schema_to_pydantic_model(self.output_schema)` then `.with_structured_output(output_model)`. `schema_to_pydantic_model()` (`utils.py`) uses the `output_schema` dict keys directly as Pydantic field names:

```python
for field_name, field_def in schema.items():
    pydantic_fields[field_name] = (field_def.python_type, Field(...))
```

Those field names become the JSON schema `properties` keys that Anthropic rejects if they contain invalid characters (e.g. `"score (1-5)"`, `"my field"`).

**Fix:** sanitize keys in `schema_to_pydantic_model()` (or at the call site) before building the model. Keep a `sanitized → original` mapping so that result post-processing can restore the original names for storage/display.

Pydantic's `Field(alias=...)` or `model_config = ConfigDict(populate_by_name=True)` can help here — the sanitized name is the field name used in the schema, the original name is the alias used when deserialising the result.

### 2. Pipeline Router node — `apps/pipelines/nodes/mixins.py`

`RouterMixin.build_router_tool()` creates:
```python
create_model("RouterOutput", route=(Literal[tuple(self.keywords)], ...))
```
The property key `route` is hardcoded (fine), but the model/tool name `RouterOutput` and the `Literal` values are user-controlled keywords. Anthropic's constraint also applies to the **tool name itself** (`^[a-zA-Z0-9_.-]{1,64}$`), so if keywords ever influence the tool name this would break too. Worth auditing.

### 3. Custom Actions / OpenAPI tools — `apps/chat/agent/openapi_tool.py`

`openapi_spec_op_to_function_def()` builds a `StructuredTool` whose `name` and `args_schema` property keys come from user-supplied OpenAPI specs (operation IDs and parameter/body schema property names). Both the tool name and property keys are passed through without sanitization.

## Suggested approach

1. **Central sanitizer utility** — add a helper (e.g. `sanitize_tool_property_name(name: str) -> str`) that:
   - Replaces any char outside `[a-zA-Z0-9_.-]` with `_`
   - Collapses consecutive `_`
   - Truncates to 64 chars
   - Ensures the result is non-empty

2. **`schema_to_pydantic_model()`** — apply the sanitizer to each key, store the mapping, and thread it through so callers can map results back to original names.

3. **`openapi_spec_op_to_function_def()`** — sanitize the operation function name and all parameter names. OpenAPI names are usually already alphanumeric but edge cases exist.

4. **Validation** — add a validator on `LLMEvaluator.output_schema` (and the evaluator creation form/serializer) that rejects or warns when a field name would require sanitization, so users know upfront rather than getting a 400 mid-run.

## Notes

- This constraint is Anthropic-specific but OpenAI's tool schema is more permissive. The sanitization should be applied at the point where we build the schema destined for Anthropic, or made provider-aware.
- The 64-char length limit is also easy to silently violate with long user-defined names — the sanitizer should handle truncation in a collision-safe way (e.g. append a short hash suffix).

brian-mecha:
### Product Description

Fixes `tool/evaluator` setups that previously failed with a `400` error when a user-defined name (_an evaluator's output field, or a custom action's OpenAPI parameter_) contained characters like spaces, parentheses, or slashes e.g. an output field named score (1-5). 
These names are now automatically sanitized to a valid format before being sent to Anthropic, with no change to what users see: evaluation results and custom action API calls still use the original names exactly as configured.

### Issue Link

Resolves #4027 

### Docs and Changelog

### Operator Impact


apps/chat/agent/openapi_tool.py
```diff
@@ -363,7 +366,41 @@ def _get_basic_type(data_type: DataType) -> type:
...
+    taken: set[str] = set()
+
+    for prop_name, prop_value in properties.items():
+        sanitized_name = sanitize_property_name(prop_name, taken)
+        taken.add(sanitized_name)
+        param_name_mapping[sanitized_name] = prop_name
+        sanitized_properties[sanitized_name] = prop_value
+
+    model = create_model(_make_model_name(name), **sanitized_properties, **kwargs)
+    model.__param_name_mapping__ = param_name_mapping
```
SmittieC: can we rename this to something like `__ocs_param_name_mapping__` just to prevent potential name conflicts if pydantic decides to use a similar mapping in the future

apps/chat/agent/openapi_tool.py
```diff
@@ -363,7 +366,41 @@ def _get_basic_type(data_type: DataType) -> type:
...
+    return model
+
+
+def _dump_with_original_names(value: Any) -> Any:
```
SmittieC: It would be nice if we can make use of pydantic's [model_serializer](https://pydantic.dev/docs/validation/dev/concepts/serialization/#model-serializers) somehow, otherwise we'd need to remember to call `_dump_with_original_names` instead of the normal `model_dump` wherever we want this behavior. Perhaps we can build our own base class to achieve this?

(example from gemini)
```python
# 1. Define a generic class that implements the serialization override
class DynamicSerializerMixin(BaseModel):
    @model_serializer
    def custom_serialize(self):
        return {
            "message": f"Hello {getattr(self, 'name', 'User')}!",
            "original_fields": {k: v for k, v in self.__dict__.items() if not k.startswith('_')}
        }

# 2. Build your dynamic fields using the Mixin as the base
DynamicUser = create_model(
    'DynamicUser',
    name=(str, ...),
    age=(int, ...),
    __base__=DynamicSerializerMixin  # Inherit the compiled serializer
)
```

SmittieC (commented):
Looking good. Two comments

apps/evaluations/evaluators.py
```diff
@@ -148,7 +148,8 @@ def run(
         )
         with track_evaluator_usage(usage_context) as callbacks:
             result_model = llm_with_retry.invoke(formatted_prompt, config={"callbacks": callbacks})
-        result = result_model.model_dump()
+        field_name_mapping = getattr(output_model, "__field_name_mapping__", {})
```
SmittieC: Please also use the ocs namespace for this. `__ocs_field_name_mapping__`

apps/evaluations/evaluators.py
```diff
@@ -148,7 +148,8 @@ def run(
         )
         with track_evaluator_usage(usage_context) as callbacks:
             result_model = llm_with_retry.invoke(formatted_prompt, config={"callbacks": callbacks})
-        result = result_model.model_dump()
+        field_name_mapping = getattr(output_model, "__ocs_field_name_mapping__", {})
+        result = {field_name_mapping.get(name, name): value for name, value in result_model.model_dump().items()}
```
SmittieC: Almost there. If you use the `OriginalNameSerializerMixin` base class in [schema_to_pydantic_model](https://github.com/dimagi/open-chat-studio/blob/20c5848a25a8b17369c5f72d57aebdea2c6eb593/apps/evaluations/utils.py#L505-L541) as well, then you don't have to do this here.

(Looks like you missed [_create_router_schema](https://github.com/dimagi/open-chat-studio/blob/918129748147b13960c8f6f098270c3d794690f2/apps/pipelines/nodes/mixins.py#L338-L342))

Looks like there are 3 distinct places where we call pydantic's `create_model`. [The router](https://github.com/dimagi/open-chat-studio/blob/918129748147b13960c8f6f098270c3d794690f2/apps/pipelines/nodes/mixins.py#L338-L342), [evals](https://github.com/dimagi/open-chat-studio/blob/20c5848a25a8b17369c5f72d57aebdea2c6eb593/apps/evaluations/utils.py#L539-L541)  and [custom actions](https://github.com/dimagi/open-chat-studio/blob/12693163a1a162693e49d4ecd1fef7f5357c6af7/apps/chat/agent/openapi_tool.py#L383-L404) which seems to do mostly the same. Perhaps we should use a single helper method to build pydantic models from a schema? This helper method can live in the "utils" app or something.

SmittieC (approved):

PR #4537, opened 2026-09-17
Move apps/chatbots views into a views package

### LINKED ISSUE #4322: Move apps/chatbots views into a views package
`apps/chatbots` keeps its views in a single 990-line `views.py`, with `apps/chatbots/public_link.py` sitting beside it as a sibling module. Six apps already use a `views` package instead:

```
apps/annotations/views/       comment_views.py  tag_views.py
apps/api/views/               channels.py  chat.py  experiments.py  files.py  participants.py  sessions.py
apps/evaluations/views/       auto_population_views.py  dataset_views.py  evaluation_config_views.py  evaluator_views.py
apps/experiments/views/       chat.py  consent.py  experiment.py  prompt.py  source_material.py  utils.py
apps/human_annotations/views/ annotate_views.py  export_views.py  queue_views.py
apps/teams/views/             feature_flags.py  internal_metadata.py  invitation_views.py  membership_views.py
```

## Shape of the change

`apps/chatbots/views.py` becomes `apps/chatbots/views/`, the current module moves inside it, `public_link.py` moves in beside it, and `__init__.py` re-exports the view names the way `apps/annotations/views/__init__.py` does.

Non-test importers are `config/urls.py` (`from apps.chatbots.public_link import public_link_page`) and `apps/chatbots/urls.py`. `from apps.chatbots.views import ...` keeps working through the re-export.

## One thing to watch

`@patch` targets that name a view module attribute do not all behave the same way after the move.

These keep working, because the path resolves to the Celery task object and patches an attribute on it, and that object is shared however it is imported:

- `apps/chatbots/tests/test_chatbot_views.py:771` - `apps.chatbots.views.send_bot_message.delay`
- `apps/chatbots/tests/test_broadcast.py:259,278,314` - `apps.chatbots.views.send_broadcast_message.delay`

These stop intercepting, because they replace a name on the module object while the calling code looks up its own module global in the moved file:

- `apps/channels/tests/channels/test_disabled_channel.py:406` - `apps.chatbots.views.get_channel_class_for_platform`
- `apps/channels/tests/channels/test_disabled_channel.py:432` - `apps.chatbots.views.send_bot_message`
- `apps/channels/tests/channels/test_disabled_channel.py:473` - `apps.chatbots.views.send_experiment_invitation`

They fail rather than pass silently, but they need repointing at the moved module as part of the change.

## Optional

`views.py` covers several concerns at 990 lines and could be split further than a straight move. Worth deciding separately, since a plain move is reviewable on its own.

---

Raised in review of #4284, where `public_link.py` was added as a sibling module.

snopoke:
### Product Description
No user-facing change.

### Technical Description
`apps/chatbots/views.py` becomes a `views/` package, matching the six apps that already do this:

```
apps/chatbots/views/
  __init__.py        re-exports the view names, plus public_link_page
  chatbot_views.py   (was apps/chatbots/views.py, byte-identical)
  public_link.py     (was apps/chatbots/public_link.py, byte-identical)
```

Git records both files as 100% renames, so the diff is the new `__init__.py` plus import and patch-target updates. `__init__.py` uses explicit named imports with `# noqa: F401`, following `apps/annotations/views/__init__.py` and `apps/experiments/views/__init__.py`. `from apps.chatbots.views import ...` keeps working, so `apps/chatbots/urls.py` is untouched; `config/urls.py` now imports `public_link_page` through the re-export.

All six `@patch` targets that named a view module attribute are repointed at `apps.chatbots.views.chatbot_views.*`, including the two task-`.delay` ones the issue expected to survive -- `patch` resolves those as attributes on the package, and the re-export list carries view names, not imported task objects. The module path is also where the code under test looks the name up. `test_chatbot_views.py` now imports the private `_chatbot_chat_ui` from the module rather than the package.

The optional further split of the 1128-line module is left for a separate change, as the issue suggests.

Verified with `pytest apps/chatbots/tests apps/channels/tests/channels/test_disabled_channel.py apps/pipelines/tests/test_deprecated_llm_models.py` (224 passed), `ruff`, and `ty`.

### Issue Link
Closes #4322

### Migrations

### Demo
n/a -- pure refactor.

### Docs and Changelog

### Operator Impact


ocs-percy (approved):
Clean structural refactor, consistent with the six apps that already do this. Easy to verify because the only real content is the new `__init__.py` and the patch-target corrections.

**`__init__.py`** — re-exports the full public surface of `chatbot_views.py` with `# noqa: F401`, matching `apps/annotations/views/` and `apps/experiments/views/`. `from apps.chatbots.views import ...` in `urls.py` keeps working untouched, and `public_link_page` is re-exported too so the single `config/urls.py` update is sufficient.

**`@patch` target corrections** — all six repointed to `apps.chatbots.views.chatbot_views.*`. Correct: `patch` resolves the name where the code under test looks it up (its import site in `chatbot_views.py`), not where it's re-exported from `__init__.py`. Patching the re-export would silently not intercept the call. The two `.delay` patches follow the same rule.

**`_chatbot_chat_ui`** — private, correctly absent from the re-export list. Direct import from the module in the test is right.

**`CSP`** — never a view, never re-exported, correct to update to the module path directly.

The two renamed files show no patch content, consistent with byte-identical.

PR #4543, opened 2026-09-17
Raise Celery queue health check timeout to 3s

snopoke:
### Product Description
no change

### Technical Description
`CeleryQueueCheck` gave each of its two broadcast RPCs a 1 second budget. That is not enough when the broker is busy — notably while a deploy restarts workers — and a slow reply is indistinguishable from no worker at all, so `/status/` reported `No worker for Celery queue 'background'`.

The real gap during deploys is fixed in dimagi/ocs-deploy#46; this only stops a slow broker from looking like the same failure.

### Docs and Changelog

### Operator Impact


ocs-percy (approved):
Approving — the 1s → 3s bump is correct, the test improvement (using `check.timeout.total_seconds()` instead of hardcoding `1.0`) is a nice bonus, and the comment explaining *why* 1s wasn't enough is the right context to leave here.

Issue #4504, opened 2026-09-14
Make Node.position_x / position_y NOT NULL


snopoke:
Part of #4499. Deferred by ADR-0049.

`Node.position_x` / `position_y` were added nullable so migration `pipelines.0030_strip_node_data` could backfill them from the blob. The backfill has run everywhere, and every save now writes them. Make the columns `NOT NULL`.

- Confirm no rows have NULL positions in prod before writing the migration (ADR-0049 notes the drift guard skipped pipelines whose blob had content with no backing row).
- Deploy note: the migration must be safe against the previous release still inserting rows. Every current write path sets the columns, so a plain `AlterField` should be fine, but check `update_or_create` defaults in `update_nodes_from_data` and the `NodeFactory`.
- Drop the `position` property's `None` branch and the "Null until the row is saved" comment on the model.

Issue #4499, opened 2026-09-14
Make Node resource FKs the source of truth and strip resource ids from Node.params


snopoke:
## Context

`Node` grew FK/M2M columns for every resource its params reference (`llm_provider`, `llm_provider_model`, `source_material`, `collection`, `collection_indexes`, `assistant`, `synthetic_voice`; plan: `docs/superpowers/plans/2026-06-15-node-resource-fks.md`). ADR-0046 and ADR-0049 then moved node content and layout out of `Pipeline.data` onto the rows. That leaves the third leg of the original design unfinished: making the FKs the source of truth for resource references and getting the ids out of `Node.params`.

## Current state

**Reads trust the FKs.** `Node.to_flow_node` overlays `resource_params()` (read off the FK columns) onto params, and the inspect API serializers read `node.llm_provider` etc. directly.

**Writes still go through params.** `update_nodes_from_data` stores `content.params` verbatim, then `_sync_resource_fk_fields` derives the FK columns from the JSON. `set_params` does the same. The docstrings and the plan still describe the FKs as "a derived mirror" of an authoritative params blob, which is the inverse of the intended end state.

Still params-driven:

- `apps/pipelines/versioning.py`: `_NODE_PARAM_SPECS` is keyed by node type and rewrites `params[...]` on publish, revert and archive. The "Option B" redesign the plan listed as follow-up was never started (overlaps #4433).
- `apps/pipelines/build_state.py`: reads `llm_provider_model_id` from params (#4377 is the N+1 that falls out of this).
- Runtime nodes in `apps/pipelines/nodes/nodes.py` are Pydantic models whose resource ids are declared fields, so execution is fed from params.
- JSON-filter lookups: `apps/utils/deletion.py` (`get_related_pipelines_queryset*`, including the `__contains` variant that can false-match), `apps/service_providers/admin.py`, `LlmProviderModel.has_related_objects`, `check_llm_model_usage.py`, `apps/documents/models.py`. Of the 13 callsites in the plan's follow-up table only the three inspect API ones were converted.
- PATCH engine `stored_params` (`apps/api/v2/pipeline_edit/graph_editor.py`) only strips mirrored ids from node types that do not declare them; it is not the read/write split.

## Work

Tracked as sub-issues, in dependency order:

1. #4500 Write resource ids to the FK columns, not params (includes the data migration and docstring rewrite)
2. #4503 Decide how runtime pipeline nodes receive their resources (decision needed before #4500 lands)
[...]

PR #4545, opened 2026-09-17
Default Node positions to the origin instead of NULL

zandre-eng:
### Product Description
No change. Groundwork so the follow-up can add a NOT NULL constraint without the previous release being able to write a NULL underneath it.

### Technical Description
`Node.position_x` / `position_y` are nullable because migration 0030 had to backfill them from `Pipeline.data`. Making them NOT NULL is #4504, but the write paths are not quite ready for it: `node_position_fields` returns `{}` for a position that is missing or malformed, and `update_nodes_from_data` spreads that into its `update_or_create` defaults, so an insert could omit both columns. A positionless import file or a factory-built pipeline left the row NULL.

Both fields now carry `default=0`, which closes every insert path at once rather than each being found and fixed. The origin is already what reads serve for an unpositioned row (`to_flow_node`), so this moves an existing fallback from read time to write time.

`node_position_fields` keeps returning `{}`, deliberately: on an *update* that means "leave the row where it is", so a malformed position on a later save cannot drag a placed node back to the origin. That distinction previously had no test and now does — `test_unusable_position_never_moves_the_row` covers both halves.

This ships separately from the constraint so that by the time #4504's migration runs, the release still serving traffic cannot insert a NULL. The constraint itself is #4546, which must not be deployed until this one has rolled out.

Migration 0032 is state-only; `sqlmigrate` emits no SQL.

### Issue Link
Groundwork for #4504, part of #4499.

### Migrations

### Demo
N/A — no user-facing change.

### Docs and Changelog

### Operator Impact


apps/pipelines/tests/test_models.py
```diff
@@ -471,17 +471,31 @@ def test_position_is_written_to_the_row(self):
             pytest.param({"x": 1}, id="missing-axis"),
         ],
     )
-    def test_unusable_position_is_not_written(self, position):
-        """Raw import files bypass wire validation; a bad position must not crash the
-        save or write garbage — the row keeps its previous position columns."""
+    @pytest.mark.parametrize(
```
zandre-eng: Fixed in 527fd4c3c. The three position tests move out to `apps/pipelines/tests/test_node_position.py` (60 lines out of `test_models.py`), which puts the file back under the threshold.

The file was already close to the limit and this branch's +14 lines tipped it, so trimming those alone would have left it sitting on the line for the next change to trip. Position behaviour — what a save writes to the columns, what a read serves when there is none — is a coherent subject on its own, so it moves as a unit instead.

snopoke (approved):

PR #4551, opened 2026-09-17
Typed, namespaced privilege elevation grants

snopoke:
### Product Description
No change to how elevation is obtained — same prompt, same 30 minutes. The release banner now names what you hold ("Django admin", `Team "Acme"`) instead of a raw slug, and the elevation URLs move to `/sudo/django-admin/`, `/sudo/team/<slug>/` and `/sudo/release/<grant>/`.

### Technical Description
Phase 1 of #4531. `apps/web/superuser_utils.py` becomes `apps/web/elevation.py`, with `Grant` (a frozen dataclass that parses and serialises its own wire form and knows its minimum role) and `Elevation` (the session wrapper). The identity proof is still the password form — allauth re-auth is Phase 2, and the `requires_elevation` decorator that closes B and B2 is Phase 3.

What the diff doesn't show:

- **Namespacing is the point.** `admin_site` and a team slug shared one namespace, so a team slugged `admin_site` unlocked `/django-admin/` (problem A). Wire forms are now `django_admin`, `ocs_admin` and `team:<slug>`.
- **Session key rename drops in-flight elevations.** `elevated_privileges` (list of pairs) → `elevations` (dict of wire form → unix expiry, fixing H). Anyone elevated across the deploy is re-prompted; the TTL is 30 minutes and nobody is locked out. Entries the new code cannot parse are pruned on read like expired ones.
- **The acquire view now checks `grant.may_be_held_by(user)` rather than a blanket `is_superuser` test**, which is what lets staff elevate into the admin grants while team impersonation stays superuser-only. It grants staff nothing today — `OcsAdminSite.admin_view` still guards its elevation check with `if request.user.is_superuser`, and nothing reads `ocs_admin` at all until Phase 3. The deny is `Http404` rather than a login redirect, so the surfaces aren't enumerable.
- **`Grant.OCS_ADMIN` is defined ahead of its consumer** deliberately, so the grant type lands once.
- **Release honours a validated `next`.** That was listed under Phase 0 and didn't make it into #4532.

Worth review attention: `Grant.__post_init__` coerces `kind` to the enum member. `may_be_held_by` branches on `kind is GrantKind.TEAM`, and a `Grant` built with a raw string kind compares equal and serialises identically but fails that `is` check, so it would take the staff branch. No call site does that today; the coercion keeps it that way.

### Issue Link
Part of #4531.

### Demo
Driven against the dev server with a throwaway superuser: `/django-admin/` redirects to `/sudo/django-admin/?next=/django-admin/`, the password elevates, the admin header shows "Release Admin Access — Expires in 29 minutes", release sends you back to the redirect. Team flow likewise, including the `Team "test-team"` banner and its `/sudo/release/team:test-team/` link, plus wrong password, unknown team, and the old `admin_site` wire form all landing where they should.

### Docs and Changelog

### Operator Impact


apps/web/urls.py
```diff
@@ -15,8 +15,10 @@
     ),
     path("status/", views.HealthCheck.as_view()),
     path("status/<str:subset>/", views.HealthCheck.as_view()),
-    path("sudo/<slug:slug>/", views.acquire_superuser_powers, name="sudo"),
-    path("sudo/<slug:slug>/release/", views.release_superuser_powers, name="release_sudo"),
+    path("sudo/django-admin/", views.elevate_django_admin, name="elevate_django_admin"),
+    path("sudo/ocs-admin/", views.elevate_ocs_admin, name="elevate_ocs_admin"),
```
SmittieC, line 19: Will these stay unwired to buttons somewhere in the UI?

SmittieC (approved):
LGTM

apps/web/urls.py
snopoke: There is a button in the menu for 'ocs-admin' (if you have permission to see it). There's not button for Django admin and I don't intend to add one since I don't want to encourage it's use.

PR #4554, opened 2026-09-17
Run CI on script-only changes

### LINKED ISSUE #4517: Remove `scripts/**` from CI `paths-ignore` so script-only PRs trigger tests
## Goal

Remove `scripts/**` from the `paths-ignore` list in `.github/workflows/lint_and_test.yml` so that PRs touching only the `scripts/` directory still trigger CI — including the scripts' own test suites.

## Context

- **File:** `.github/workflows/lint_and_test.yml`
- The `scripts/` directory contains several quality-gate scripts with their own test files:
  - `scripts/check_atomic_exception_handling.py` + `scripts/test_check_atomic_exception_handling.py`
  - `scripts/check_inline_imports.py` + `scripts/test_check_inline_imports.py`
  - `scripts/diff_allauth_overrides.py` + `scripts/test_diff_allauth_overrides.py`
  - `scripts/reconcile_models.py` + `scripts/test_reconcile_models.py`
- Currently, a PR that only changes files under `scripts/` skips all CI jobs entirely, so those tests never run for script-only changes.
- `scripts/` is **not** excluded from ty's `[tool.ty.src]` exclude list — ty already checks scripts when run manually.
- Parent tracking issue: #4257

## Change

In `.github/workflows/lint_and_test.yml`, remove `'scripts/**'` from **both** `paths-ignore` blocks (the `pull_request` trigger and the `push` trigger). They are currently identical:

```yaml
# Before (both pull_request and push triggers)
paths-ignore:
  - 'components/**'
  - 'docs/**'
  - 'mkdocs.yml'
  - 'README.md'
  - 'CLAUDE.md'
  - 'LICENSE'
  - '.coderabbit.yaml'
  - 'locust/**'
  - 'docker-compose-dev.yml'
  - 'heroku.yml'
  - 'app.json'
  - 'scripts/**'    # ← Remove this line from both blocks
  - 'media/**'
```

Remove the `'scripts/**'` line from both `paths-ignore` blocks. No other changes.

The `python-tests` job already runs `uv run pytest -n auto ...` which recurses into the project root. Confirm that `scripts/test_*.py` files are discovered by checking pytest's `norecursedirs` in `pyproject.toml` — if `scripts/` is listed there, remove it too. Currently `norecursedirs` is: `".* build dist venv node_modules compose assets static"` — `scripts/` is not listed, so pytest will find the test files automatically.

snopoke:
### Product Description
No change.

### Technical Description
A PR touching only `scripts/` currently skips every CI job, so the quality-gate scripts' own test suites (379 tests) never run. Removing `'scripts/**'` from both `paths-ignore` blocks, and extending the `ty` job from `apps/` to `apps/ scripts/` — `[tool.ty.src]` never excluded `scripts/`, it just was not in the job's argument.

No `pyproject.toml` change needed for the tests: `norecursedirs` doesn't list `scripts/` and there's no `testpaths`, so a bare root-level `pytest` already collects all 379 of them.

Turning ty on surfaced two errors, fixed in the second commit:

- `mock_connect_server.py` passed the base64 HMAC digest to httpx as `bytes` in an otherwise-`str` header dict. Now `.decode()`d. httpx renders both to the same header bytes (checked), and `verify_hmac` re-encodes the header value anyway, so the mock still signs identically.
- `ocs_cost_estimator.py` rebinds the optional `anthropic` import to `None`. ty can't express "module or None" for an import binding, so the existing (ineffective) `# type: ignore[assignment]` became `# ty: ignore[invalid-assignment]`. Behaviour unchanged — the dependency stays optional.

Also suppressed `unresolved-import` for `flask` in `mock_connect_server.py`: it's a PEP 723 script declaring its own deps, so flask isn't in the venv ty runs against. Adding it as a project dev dependency for one dev script seemed worse than the one-line suppression, but say the word if you'd rather have that.

### Issue Link
Closes #4517

### Docs and Changelog

### Operator Impact


ocs-percy (approved):
Correct on all counts. Removing `scripts/**` from `paths-ignore` means the quality-gate tests (#4524) now actually run on CI. The `ty` extension and the `digest.decode()` fix are both right, and `# ty: ignore[unresolved-import]` for a PEP 723 script is the appropriate suppression rather than adding flask as a project dev dependency.

Issue #4529, opened 2026-09-16
Download Latest Results fails with 502 on large evaluation configs


ocs-percy:
## Bug report

Clicking **Download Latest Results** on a large evaluation config hangs for a long time and then fails with a **502 Bad Gateway** instead of producing a CSV.

## Steps to reproduce

1. Open an evaluation config with a large number of completed runs / dataset items.
2. Click **Download Latest Results**.
3. Observe the "Generating…" spinner for an extended period.
4. Page eventually shows a 502 Bad Gateway error rather than a download link.

## Technical context

The download flow is:

1. `POST evaluation_bulk_download_start` → dispatches `export_evaluation_bulk_results_task` and returns a `task_id` immediately.
2. The template polls `evaluation_bulk_download_link` every 2 s via HTMX until the task reports success.
3. `export_evaluation_bulk_results_task` (on the `BACKGROUND` queue) runs `_get_bulk_results_queryset` → `build_evaluation_table_data` → writes the full CSV into an in-memory `StringIO` → saves a `File` object.

Two likely causes:

**a) No time limit on the export task.**  
Unlike the evaluation batch tasks (which have `soft_time_limit=240`), `export_evaluation_bulk_results_task` has no `soft_time_limit` / `time_limit`. For a large dataset it can run long enough to hit the worker's default hard limit or an infrastructure-level kill, leaving the task in a permanent pending state. The HTMX poller then never gets a `complete=True` result — it keeps polling until something between the browser and the server cuts the connection (resulting in the 502).

**b) Entire result set assembled in memory.**  
`_get_bulk_results_queryset` fetches every `EvaluationResult` for every completed run of the config (with `select_related` + `prefetch_related`), passes it all through `build_evaluation_table_data`, and writes the whole CSV into a `StringIO` in one shot. For configs with many runs and large datasets this is an unbounded memory allocation with no streaming.

The `DISTINCT ON (message_id, evaluator_id)` in the queryset also requires PostgreSQL to materialise and sort the full result set before deduplication, which can be very slow when the number of result rows is large.

## Expected behaviour

Large exports should complete without error and produce a downloadable CSV.

## Possible fixes to investigate

- Add a `soft_time_limit` to `export_evaluation_bulk_results_task` so timeouts surface as a clean error rather than a hung task, and surface that error in the UI via the existing `export_error` template path.
- Stream the queryset in chunks (`iterator()`) and write rows incrementally rather than loading all results into memory at once.
[...]

PR #4557, opened 2026-09-17
Stream the bulk evaluation results export and show its progress

SmittieC:
### Product Description
An attempt at stopping **Download Latest Results** from failing on a large evaluation config. The export no longer holds the whole result set in memory, and the button reports a percentage instead of an open-ended spinner.

I have not reproduced the 502 against a config big enough to trigger it, so treat this as a mitigation of the cause described in #4529, not a confirmed fix.

### Technical Description
Two commits:

- **Stream the export.** The task built every row in memory and uploaded the CSV as one `ContentFile`. Results are now read in chunks and a row is written as each message's results arrive, spooled to a temp file that storage uploads in chunks — peak memory is flat in the number of results. Worth a look: the applied-tags `ArraySubquery` moved onto an *outer* query over the deduplicated ids. Postgres projects a correlated subquery below the `DISTINCT ON`, so annotating the inner query ran that subquery once per result of every run, not once per exported row.
- **Report progress.** The UI polled an HTMX endpoint that only knew "done / not done". That endpoint is gone; the button now uses `CeleryProgressBar` like the chat and team-files exports, fed by a `ProgressRecorder` the task drives in one-percent steps. The row count is taken up front — one extra query the streaming read would not otherwise make, which is what buys a percentage rather than a spinner.

#### How the spooling works, and why there are two files

The CSV header is the union of every row's keys. Context columns and evaluator output fields vary per message, so the header is not known until the last row has been read — but the header has to be the first line written. That is what forces two passes, and a spool for each:

1. **Row spool** (in `write_evaluation_csv`). Each row is written out as one JSON object per line as it arrives, while the header set accumulates. Only that set of column names stays resident across rows; the row dict itself is dropped once written.
2. **CSV spool** (`export_evaluation_csv_to_tempfile`). Once the last row is in, the header is ordered and written, then the row spool is re-read a line at a time and each row emitted in column order. This one is binary, so it can be handed straight to Django storage, which uploads it in chunks.

Both are `SpooledTemporaryFile(max_size=10 MiB)`: they stay in memory while small and only spill to disk past that threshold. A small export never touches disk; a large one never holds itself in RAM. Between the chunked queryset read and these two spools, the only thing that grows with the size of the export is the set of header names.

Note the export task returns its failure rather than raising, so a task that "succeeded" can still carry an error. The JS checks the result payload rather than relying on `onTaskError`.

#### Measured

Two configs, exported through a local benchmark command (not part of this PR), before and after. The CSV is byte-for-byte the same size either way, so this is the same output built differently. Before-figures repeated across runs to within 0.1 s.

| | Config 1 (15k messages) | Config 2 (messsage count unknown, but also a lot) |
|---|---|---|
| wall time | 12.91 s → **3.32 s** | 9.86 s → **2.90 s** |
| peak Python objects | 353.6 MiB → **4.0 MiB** | 724.6 MiB → **9.3 MiB** |
| peak RSS growth | 737.7 MiB → **7.2 MiB** | 1,129.9 MiB → **16.8 MiB** |
| peak RSS total | 1,076.8 MiB → **346.8 MiB** | 1,470.4 MiB → **356.4 MiB** |
| CSV size | 9.4 MiB (unchanged) | 8.9 MiB (unchanged) |

RSS growth is what the export itself adds to the worker: ~740 MiB and ~1.1 GiB of transient allocation become single-digit MiB. Neither of these configs was large enough to fail before, so this shows the allocation curve flattening rather than a failure going away. The wall-time drop is a side effect of the `ArraySubquery` move, not of the streaming.

### Issue Link
#4529

### Demo
Click **Download Latest Results** on an evaluation config with completed runs.

### Docs and Changelog

Bug fix: the bulk evaluation results export is built in a streaming fashion instead of in memory, and shows progress while it runs.

### Operator Impact


apps/evaluations/tasks.py
```diff
@@ -1498,52 +1504,91 @@ def report(created: int, skipped: int) -> None:
...
         EvaluationResult.objects.filter(
             run__config=config,
             run__status=EvaluationRunStatus.COMPLETED,
             run__type__in=[EvaluationRunType.FULL, EvaluationRunType.DELTA],
             team=team,
         )
-        .select_related("message__session__experiment", "evaluator", "session", "run")
-        .prefetch_related("applied_tags__tag")
+        .values("message_id")
+        .distinct()
```
snopoke: Is distinct needed here?

apps/evaluations/tasks.py
```diff
@@ -1498,52 +1504,91 @@ def report(created: int, skipped: int) -> None:
...
-            file_obj = File.objects.create(
-                name=filename,
-                team=team,
-                content_type="text/csv",
-                file=ContentFile(csv_buffer.getvalue().encode("utf-8"), name=filename),
-                purpose=FilePurpose.DATA_EXPORT,
-                expiry_date=timezone.now() + timedelta(days=7),
-            )
+
+            # Two spooled temp files: one holds the rows until the header is known, one the CSV.
```
snopoke: This comment seems misplaced

apps/evaluations/tasks.py
SmittieC: Yes. Considering that this is for the bulk export (merging results from multiple evaluators / runs), we want to know how many _unique_ evaluation messages are to be processed. We want to know this, because in `iter_evaluation_table_rows`, we do a `group_by(message_id)` so that we process them unique message at a time.

apps/evaluations/tasks.py
SmittieC: I removed it rather

snopoke (dismissed):
Up to you if you want to address the other review findings.

SmittieC:
Addressed one of the three, pushing back on the other two.

**1. `apps/evaluations/tasks.py` — duplicated filter: fixed** in b26bf6ae25. Extracted `_bulk_export_results(config, team)`; both `_count_bulk_export_rows` and `_get_bulk_results_queryset` now build from it, so the progress total and the exported rows can't drift apart.

**2. `bulk_download.html` — regex substitution on the reversed URL: not changing.**

The suggested replacement doesn't run. `files:base` is `path("<int:pk>/", ...)` ([`apps/files/urls.py:9`](https://github.com/dimagi/open-chat-studio/blob/main/apps/files/urls.py#L9)), so `{% url 'files:base' request.team.slug '__FILE_ID__' %}` raises `NoReverseMatch` — the `int` converter only matches digits. Reversing with `0` is precisely why the current form exists.

It's also the established convention rather than something this PR introduced. Three other templates do the same substitution:

- `templates/experiments/components/exports.html:93`
- `templates/teams/partials/download_files_progress.html:102`
- `templates/documents/partials/download_progress.html:102`

Changing only the new call site would make it the odd one out. The underlying point is fair — if `files:base` moves, four call sites break silently — but that's a shared helper across all four, which is a separate change from this PR.

**3. `bulk_download.html` — re-initialization / duplicate ids: not changing.**

The scenario can't occur as written:

- The partial is included once, on `templates/evaluations/evaluation_runs_home.html:27`, which renders one config per page.
- The script block only renders under `{% if task_id %}`, and in that same branch the button renders `disabled` (line 8). A disabled button fires no click, so htmx can't re-swap and re-run the script. `finish()` then hides the button on every terminal state.

The finding is phrased conditionally ("if the export button is ever clicked again", "future reuse of this partial"), so this would be defensive code for a case nobody has proposed.

SmittieC:
> Up to you if you want to address the other review findings.

Addressed one of them

snopoke (approved):

PR #4561, opened 2026-09-18
fix: report a model upstream serves only as non-chat as removed

SmittieC:
### Product Description
No change — internal tooling for the model auto-sync workflow.

### Technical Description
`translate` built its second catalogue (`everything`) from every key that mapped to an OCS provider regardless of mode, so a model LiteLLM still prices but no longer serves as chat stayed in that map and never showed up as `removed`. `groq/whisper-large-v3-turbo` is the live example: it moved to `mode: audio_transcription` upstream and would have sat in our catalogue indefinitely. Both catalogues are now mode-filtered, and `everything` is renamed `all_chat` since it no longer holds everything upstream lists. The two maps still differ on deprecation, which is the distinction that matters for keeping "retired" and "deprecated" as separate findings.

Worth a look: `MIN_UPSTREAM_MODELS` stays at 300, but the number it guards dropped from 558 to 386 now that non-chat modes are excluded. The margin is narrower than before, so a large upstream chat retirement is now closer to tripping the "table is unreadable" refusal.

### Docs and Changelog

### Operator Impact


SmittieC:
Basically all this does it to detect + remove non-chat models from our registry. We had a whisper model for groq (not sure why)

snopoke (approved):

PR #4566, opened 2026-09-18
python node improvements

### LINKED ISSUE #3061: Python node improvement ideas
* make the AI prompt always visible
* show the new code side by side
* update the prompt to say "don't change code unless necessary"
* allow clearing the AI generated code
* add a button to 'check the code for bugs' with a specific AI prompt
* review code gen prompt for missing context e.g. set_participant_data_key

brian-mecha:
### Product Description

Reworks the AI Help panel on Python nodes so the flow matches a code-review diff view instead of a one-shot suggestion box:

- The prompt stays visible after generating, so you can tweak it and refine without retyping.
- The AI's suggestion now renders as an inline diff against your current code, instead of a separate unrelated editor.
- New `"Check for bugs"` button reviews the current code without needing a prompt.
- New `"Clear"` action to discard a suggestion and start over, separate from "Reject" (which discards the suggestion but keeps your prompt).
- The AI is now told to make minimal, targeted edits rather than rewriting the whole function, and its instructions now cover five sandbox functions (`set_participant_data_key`, `append_to_participant_data_key`, `increment_participant_data_key`, `get_participant_schedules`, `end_session`) and the `http` client's typed exceptions that it previously didn't know existed — it should stop suggesting non-existent APIs or ignoring safer alternatives to a full data overwrite.

### Issue Link

Resolves #3061 

### Migrations

### Demo

python_node_ai_diff

### Docs and Changelog

### Operator Impact


apps/help/tests/test_help.py
```diff
@@ -389,6 +346,45 @@ def test_run_passes_trace_config_into_every_retry(self, mock_build, mock_get_tra
         for call in mock_agent.invoke.call_args_list:
             assert call.kwargs["config"]["callbacks"] == [callback]
 
+    def test_build_system_prompt_includes_minimal_diff_instruction_when_current_code_present(self):
```
SmittieC, line 355: Thoughts on removing this and `test_build_system_prompt_omits_minimal_diff_instruction_when_no_current_code`? They don't seem that useful to test?

apps/help/tests/test_help.py
```diff
@@ -389,6 +346,45 @@ def test_run_passes_trace_config_into_every_retry(self, mock_build, mock_get_tra
...
+
+    def test_build_system_prompt_omits_minimal_diff_instruction_when_no_current_code(self):
+        agent = CodeGenerateAgent(input=CodeGenerateInput(query="write hello world"))
+        prompt = agent._build_system_prompt("", error=None)
+        assert "Make the smallest possible edit" not in prompt
+
+    @pytest.mark.parametrize(
+        "name",
+        [
+            "set_participant_data_key",
```
SmittieC: can this list be derived from the output of `nodes._get_custom_functions`?

apps/help/code_generate_system_prompt.md
```diff
@@ -30,8 +30,23 @@ def get_participant_data() -> dict:
     Returns the current participant's data as a dictionary.
 
 def set_participant_data(data: dict) -> None:
-    Updates the current participant's data with the provided dictionary. This will overwrite any existing
-    data.
+    Merges the provided dictionary into the current participant's data. Only the keys present in `data`
+    are overwritten; other existing keys are left untouched.
+
+def set_participant_data_key(key_name: str, data: Any) -> None:
```
SmittieC: Can we derive the list of methods from `nodes._get_custom_functions` and their docstrings? Looks like there was similar logic in the deleted utils.py file

SmittieC (commented):
Looking good. A few small comments

snopoke (commented):
I test this locally and it all seems to be working. One think I did notice is that any comments outside the `main` function get removed. I know some bot builders add comments above the code for documentation.

apps/help/tests/test_help.py
brian-mecha: I'd keep them. They're the only automated coverage of the `minimal-diff instruction` part of this PR. Without them, a later refactor could silently regress the exact behavior I added (`e.g. someone "simplifies" _build_system_prompt by moving the instruction into the static .md template unconditionally, which the plan explicitly considered and rejected`)

apps/help/agents/code_generate.py
```diff
@@ -75,13 +75,20 @@ def _build_system_prompt(self, current_code: str, error: str | None) -> str:
...
+                "relevant to the requested change. Do not rewrite or reformat unrelated code."
+            )
         if error:
             prompt_context["error"] = f"\nThe current function has the following error. Try to resolve it:\n\n{error}"
 
         system_prompt = system_prompt.format(**prompt_context).strip()
         system_prompt += (
-            "\n\nIMPORTANT: Start your response with exactly"
-            " `def main(input: str, **kwargs) -> str:` and nothing else before it."
+            "\n\nIMPORTANT: Respond with only Python code — no markdown code fences and no prose"
```
SmittieC, line 89: This does not seem to preserve the comments above the code 🤔

brian-mecha: @SmittieC The fix is already there (commit 89b3c04ac). The "start your response with exactly `def main...` and nothing else before it" instruction was replaced with one that explicitly permits and preserves leading comment lines, while still forbidding markdown fences and prose.

snopoke: @brian-mecha did you also test the 'check for bugs' path?

brian-mecha: @snopoke I have added 2 eval cases to `apps/help/evals/fixtures/code_generate.yml` (`check_for_bugs_finds_and_fixes_a_real_bug`, `check_for_bugs_leaves_correct_code_unchanged`), both using the literal `CHECK_FOR_BUGS_PROMPT` text from `widgets.tsx` (_with a comment noting they must stay in sync_), covering both halves of that prompt's contract. This fixes a real bug when one exists, and leaves already-correct code alone.

snopoke (approved):

PR #4567, opened 2026-09-18
fix: use consent_forms for the export isolation test

snopoke:
### Product Description
no change

### Technical Description
`test_resource_isolates_other_teams_data` started failing on main. Not a flake — the auto-models migration merged in 588b9ec4c took the seeded global `LlmProviderModel` count to 103, and `llm_provider_models` serves global (`team IS NULL`) rows alongside the team's own. With the endpoint's default page size of 100 and a `pk` cursor ordered ascending, the row the test creates lands on page 2, so it is absent from `results`. A stale `--reuse-db` database hides it; `--create-db` reproduces it every time.

Switched the test to `consent_forms`, which is not in `GLOBAL_CONFIG` and so has no global rows to accumulate. Same `pk` cursor and same `team_scoped_queryset` path, so the view behaviour under test is unchanged.

The alternative — raising `limit` on the request — keeps the resource but stays dependent on a seeded row count that migrations keep growing, which is the thing that broke here.

This drops the only view-level test touching a resource with global rows. Left that way deliberately: the view adds no global-specific behaviour, and `test_team_scoped_queryset_isolates_teams_and_includes_globals` pins own-rows + globals + other-team-excluded at the queryset layer, unpaginated and therefore immune to the seeded count.

### Issue Link
N/A

### Demo
N/A

### Docs and Changelog

### Operator Impact


snopoke:
Resolved by https://github.com/dimagi/open-chat-studio/pull/4568

SmittieC (approved):
Thanks

PR #4568, opened 2026-09-18
test: page the export resource to exhaustion when checking team isolation

SmittieC:
### Product Description
None. Test-only change.

### Technical Description
`test_resource_isolates_other_teams_data` read only the first page of `/api/export/llm_provider_models/`.

That resource is team-scoped *plus* global rows — `team_scoped_queryset` adds `Q(team__isnull=True)` so the team-less rows the migrations seed are shared across servers. The default page size is 100 (`DEFAULT_LIMIT` in `apps/api/export/views.py`) and the resource pages by pk ascending, so the seeded global rows come first.

`DEFAULT_LLM_PROVIDER_MODELS` went from 96 to 103 entries in #4547 (migration `0082_auto_model_sync_20260917`). At 103 seeded rows the first page is entirely global, the team's own row is on page 2, and the assertion fails:

```
FAILED apps/api/export/tests/test_views.py::test_resource_isolates_other_teams_data
  assert 113 in [1, 2, 3, 4, 5, 6, ...]
```

The other half of the test was worse than broken — `assert theirs.id not in ids` passed vacuously, since the other team's row was off page 1 whatever the scoping did.

The test now pages to exhaustion through a `_collect_ids` helper, so it no longer depends on how many global rows the migrations seed. That matches the existing invariant in `AGENTS.md` that tests must not depend on migration-seeded global rows.

Raising the `limit` instead would work today and break again at `MAX_LIMIT`.

The failure only shows up against a freshly migrated database. The suite runs with `--reuse-db`, which skips migrations, so a local test database created before #4547 still has the old model list and the test passes there.

### Issue Link
None — failure observed in CI on `main`: https://github.com/dimagi/open-chat-studio/actions/runs/35337707096/job/105576294158

### Migrations

No migrations in this PR.

### Demo
Before, against a freshly created test database:

```
apps/api/export/tests/test_views.py::test_resource_isolates_other_teams_data FAILED
    assert mine.id in ids
E   assert 106 in [1, 2, 3, 4, 5, 6, ...]
```

After:

```
28 passed, 33 warnings in 2.42s
```

### Docs and Changelog

### Operator Impact


snopoke (dismissed):

snopoke (approved):

SmittieC:
Resolved by https://github.com/dimagi/open-chat-studio/pull/4567

PR #4570, opened 2026-09-18
Use allauth re-authentication for privilege elevation

snopoke:
### Product Description
Elevating to admin or team access now goes through the normal "Confirm Access" prompt instead of a bespoke password form. Accounts with two-factor authentication can confirm with an authenticator or recovery code, and SSO-only accounts with neither a password nor MFA are refused rather than being handed a form they cannot answer. The prompt names what is being elevated into, and repeated attempts are rate limited.

### Technical Description
Phase 2 of #4531.

`_acquire_elevation` hands the identity proof to allauth's `stash_and_reauthenticate`, with `apps.web.elevation.complete_elevation` as the resume callback. The callback re-checks staleness, the grant, and the minimum role before adding it to the session.

- `did_recently_authenticate` / `reauthentication_required` are deliberately not used. That timer is global and five minutes wide, so a user who logged in moments ago would elevate with no prompt, and it returns `True` unconditionally for users with no usable password and no MFA.
- allauth keeps the stash in the session until *some* re-authentication consumes it, so an abandoned elevation could otherwise be completed by an unrelated later prompt. `STASH_MAX_AGE` bounds it at 2 minutes, as agreed on the issue. Open to widening it: it bounds the whole round trip, and an MFA user who has to find a recovery code can exceed it, in which case they lose `next` and land on `/` with a message.
- The acquire step is now a GET that only stashes, so the prompt names the pending grant (`pending_elevation`, rendered by `account/snippets/pending_elevation.html`). The removed confirmation page showed the same thing; without it a planted stash would be indistinguishable from one the user started.

Removed: `ConfirmIdentityForm`, the `check_password` call, `templates/web/temporary_superuser_powers.html`, and the duplicate password form on `templates/404.html` (now a link to the acquire URL — that form's `next` was also silently empty, since `request_path` was never in the context).

Added: `templates/mfa/reauthenticate.html`, styled to match `templates/account/reauthenticate.html`. allauth ships one, but it renders with unstyled `{% element %}` tags. Both now include the alternatives list, which is how a user switches between password and authenticator.

Gained: `ACCOUNT_RATE_LIMITS["reauthenticate"]` at `10/m/user`, and MFA step-up.

Still open for phase 3: `/admin/` has no elevation, and `OcsAdminSite.admin_view` still guards on `is_superuser`, so staff skip step-up on `/django-admin/`.

### Issue Link
#4531

### Migrations

No migrations. The session shape is unchanged from phase 1.

### Demo
Driven against a local server, both branches end to end:

```
Password:
GET  /django-admin/                 302 /sudo/django-admin/?next=/django-admin/
GET  /sudo/django-admin/            302 /accounts/reauthenticate/
POST /accounts/reauthenticate/      302 /django-admin/
GET  /django-admin/                 200
GET  /sudo/release/django_admin/    302 /
GET  /django-admin/                 302 /sudo/django-admin/?next=/django-admin/

MFA:
GET  /sudo/django-admin/            302 /accounts/reauthenticate/
GET  /accounts/reauthenticate/      200  (offers "Use authenticator app or code")
POST /accounts/2fa/reauthenticate/  302 /django-admin/
GET  /django-admin/                 200
```

A wrong password, and an SSO-only account with no password and no MFA, both leave the grant unheld.

### Docs and Changelog

Docs are phase 5.

### Operator Impact


apps/web/elevation.py
```diff
@@ -175,3 +193,71 @@ def _parses(wire_form: str) -> bool:
...
+    stash = request.session.get(flows.reauthentication.STATE_SESSION_KEY) or {}
+    if stash.get("callback") != REAUTH_CALLBACK:
+        return None
+    state = stash.get("state") or {}
+    try:
+        grant = Grant.parse(state["grant"])
+    except (InvalidGrant, KeyError, TypeError):
+        return None
+
+    # Naming a request `complete_elevation` will refuse would promise access the user cannot get.
```
snopoke:
```suggestion
```

apps/web/views.py
```diff
@@ -101,19 +71,17 @@ def _acquire_elevation(request, grant):
...
-        if form.is_valid():
-            if response := _grant_on_confirmed_identity(request, grant, form):
-                return response
-    else:
-        redirect_to = _safe_redirect(request.GET.get("next", ""))
-        if grant.kind is GrantKind.TEAM and _is_team_member(request.user, grant.team_slug):
-            return HttpResponseRedirect(redirect_to)
+    next_url = safe_redirect_url(request.GET.get("next", ""))
+    elevation = Elevation(request)
+    if elevation.has(grant) or (grant.kind is GrantKind.TEAM and _is_team_member(request.user, grant.team_slug)):
```
SmittieC: thoughts on making the team membership check part of `elevation.has(grant)`? It has al the context to do that check already. Up to you

SmittieC (dismissed):
LGTM

apps/web/views.py
snopoke: `elevation.has` is currently a pure lookup of what has already been granted. The 2nd part of the conditional is a check to see whether a grant is necessary which is a separate concern so not going to fold it in.

snopoke: Updated to extract a helper function

snopoke: Fixed in 97d26897f. The held-grant check and the team-membership check are now one `if` each, so the conditional is no longer compound. `code_health_review` on the file reports 10.0 with no findings.

*— Claude Code*

SmittieC (approved):

